PR #11, opened 2023-03-03
HardWordWidsom.md: relate HCG and Stability Window

nfrisby:
Added an answer to "Why use the Honest Chain Growth window as the Ledger's Stability Window?"

### DIFF
--- a/docs/HardWonWisdom.md
+++ b/docs/HardWonWisdom.md
@@ -99,3 +99,16 @@ A couple more observations:
 
 So, either by mimicking the approach of the existing `ValidityInterval` validation logic or by altering the STS innards to allow short-circuiting failure, we could reify the `EpochInfo` failures into the `PredicateFailure` hierarchy, and thereby leverage the types to force each invocation of the Ledger API to independently handle the possibility of `PastHorizonException`s.
 But it's not obvious that that is definitely worth the extra complexity it would introduce on the Ledger code.
+
+## Why use the Honest Chain Growth window as the Ledger's Stability Window?
+
+Suppose we have selected a different chain than our peer and that our selected chain has L blocks after the intersection and their selected chain has R blocks after the intersection.
+
+REQ1: If k<L, we must promptly disconnect from the peer (b/c of Common Prefix violation and Limit on Rollback).
+
+REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their headers, because Praos requires us to fetch and select the longer chain, and validating those headers is the first step towards selecting those blocks.
+(This requirement ignores tiebreakers here because the security argument must hold even if the adversary wins a tiebreaker.)
+
+The most demanding case of REQ2 is L=k: at most we'll need to validate k+1 of the peer's headers.
+Thus using HCG window as Stability Window ensures that forecasting can't disrupt REQ2 when the peer is serving honest blocks.
+(We can tick the intersection ledger state to their first header, and then forecast 3k/f from there which by HCG will get us at least the remaining k headers if they're serving an honest chain.)


docs/HardWonWisdom.md
```diff
@@ -99,3 +99,16 @@ A couple more observations:
 
 So, either by mimicking the approach of the existing `ValidityInterval` validation logic or by altering the STS innards to allow short-circuiting failure, we could reify the `EpochInfo` failures into the `PredicateFailure` hierarchy, and thereby leverage the types to force each invocation of the Ledger API to independently handle the possibility of `PastHorizonException`s.
 But it's not obvious that that is definitely worth the extra complexity it would introduce on the Ledger code.
+
+## Why use the Honest Chain Growth window as the Ledger's Stability Window?
+
+Suppose we have selected a different chain than our peer and that our selected chain has L blocks after the intersection and their selected chain has R blocks after the intersection.
```
dnadales, line 105: Minor: I'd wrap up lines to 80 chars to minimize diffs and help readability.

docs/HardWonWisdom.md
```diff
@@ -99,3 +99,16 @@ A couple more observations:
...
+
+REQ1: If k<L, we must promptly disconnect from the peer (b/c of Common Prefix violation and Limit on Rollback).
```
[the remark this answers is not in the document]
dnadales: I wonder if we could add some ascii art to illustrate this, eg:

```text

|----------- R ------------|

          |------- k ------|

n ------- I ----------------
 \
  \-------------------------

|------------ R -----------|

where

- n: intersection point between our chain and our peer's.
- I: immutable point
```

docs/HardWonWisdom.md
```diff
@@ -99,3 +99,16 @@ A couple more observations:
...
+
+REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their headers, because Praos requires us to fetch and select the longer chain, and validating those headers is the first step towards selecting those blocks.
+(This requirement ignores tiebreakers here because the security argument must hold even if the adversary wins a tiebreaker.)
+
+The most demanding case of REQ2 is L=k: at most we'll need to validate k+1 of the peer's headers.
+Thus using HCG window as Stability Window ensures that forecasting can't disrupt REQ2 when the peer is serving honest blocks.
```
dnadales, line 113: We should define HCG (Honest Chain Growth) somewhere if it isn't already.

dnadales, line 113: I don't get this. What does it mean for forecasting to disrupt REQ2? How is this related to the peer serving honest blocks?

docs/HardWonWisdom.md
nfrisby: For Markdown files, I favor writing exactly one sentence per line in the file. It makes diffs drastically more legible.

It does harm readability, but I assume that people prioritizing readability can easily use a Markdown renderer when reading the document.

Seem reasonable?

docs/HardWonWisdom.md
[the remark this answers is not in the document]
nfrisby: > because the intersection point lies before the immutable tip.

That fact only requires a disconnection because it represents a Common Prefix violation and/or usefully communicating with this peer would require violating our Limit on Rollback.

So I wouldn't want to replace the text with your suggestion, but I wouldn't object to adding it (and incorporating the sentence I just wrote above this one).

Thoughts?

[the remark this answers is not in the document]
nfrisby: I hesitate to include a diagram in `HardWonWisdom`; seems out of scope/bloaty. But if this were integrated into a document with a wider intended audience, then I agree a diagram would make sense (but probably in a section explaining how we rely on Common Prefix to justify disconnections).

docs/HardWonWisdom.md
nfrisby: Agreed

nfrisby: > What does it mean for forecasting to disrupt REQ2?

If the forecasting horizon doesn't necessarily reach their `L+1`th header, then it would prevent us from validating them, which is what REQ2 requires we can do (if they're serving the honest chain.)

docs/HardWonWisdom.md
dnadales: Fair enough. The donwside is that reading the diff on GitHub is a bit harder.

docs/HardWonWisdom.md
dnadales: I see. Would it make sense to add this clarification to this document. You have to imagine people like me would be the audience :D

docs/HardWonWisdom.md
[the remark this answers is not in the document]
dnadales: > Thoughts?

Alright :) I was just pointing out that the situation you described is one in which the intersection point lies before the immutable tip.

dnadales (approved):

PR #15, opened 2023-03-09
Fix ouroboros-benchmark-workflow

dnadales:


.github/workflows/publish-mempool-benchmarks.yml
```diff
@@ -70,6 +69,15 @@ jobs:
           ghc-version: ${{ matrix.ghc }}
           cabal-version: ${{ matrix.cabal }}
 
+      - name: "Linux: Install build environment (apt-get)"
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install libsodium23 libsodium-dev
+          sudo apt-get -y install libsystemd0 libsystemd-dev
```
Jasagredo, line 77: This is not needed

.github/workflows/publish-mempool-benchmarks.yml
```diff
@@ -70,6 +69,15 @@ jobs:
           ghc-version: ${{ matrix.ghc }}
           cabal-version: ${{ matrix.cabal }}
 
+      - name: "Linux: Install build environment (apt-get)"
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install libsodium23 libsodium-dev
```
Jasagredo, line 76: `libsodium23` is not needed

.github/workflows/publish-mempool-benchmarks.yml
```diff
@@ -87,15 +95,17 @@ jobs:
           ref: master
           path: ouroboros-network
 
-      - name: Cabal update
+      - name: Create a cabal.project.local file
         run: |
-          cd ouroboros-network
-          cabal update
+          # We need this option to make cardano-crypto-praos use the libsodium
```
Jasagredo, line 100: Not exactly, we need it to tell cabal to include some C files in cardano-crypto-praos that allows us to work with an unmodified libsodium because the modified bits are provided by those c files.

Now that I worded it maybe I'm wrong, and it is precisely what you described, but in any case I'm leaving this here for your consideration 😄

Jasagredo (approved):

PR #24, opened 2023-04-12
Update style guide.

Jasagredo:
- Stop mentioning `RecordWildCards`
- No longer suggest to breakdown the import list in groups
- No longer suggest to import `as X` in re-export modules


docs/StyleGuide.md
```diff
@@ -419,14 +421,6 @@ the rules below, it is good practice to update the code's style to match them.
...
       ```
 
-      If the field names of the `Args` type are named appropriately, using
-      `RecordWildCards` is also acceptable, see also (17).
-
-      ```haskell
-      foo Args {..} =
-          ..
-      ```
-
```
jorisdral, line 429: Maybe we should mention somewhere that we prefer not to use `RecordWildCards`?

docs/StyleGuide.md
```diff
@@ -1046,9 +977,9 @@ the rules below, it is good practice to update the code's style to match them.
...
 
-    For records we often use `NamedFieldPuns` to make it convenient to
-    extract fields from the record. We also tend to use `RecordWildCards`
-    when the fields of the record are all of the form
+    For records we often use `NamedFieldPuns` to make it convenient to extract
+    fields from the record. We use the following convention when naming fields
+    to avoid duplicate record fields (we do not use `DuplicateRecordFields`):
 
     ```haskell
     data SomeRecord = SomeRecord {
```
jorisdral, line 985: There are actually a lot of places where we use an `srSomeField` naming convention, so mayb we should reconsider this section?

docs/StyleGuide.md
```diff
@@ -1107,6 +1033,8 @@ the rules below, it is good practice to update the code's style to match them.
     -Widentities
     -Wredundant-constraints
     -Wmissing-export-lists
+    -Wunused-packages
+    -Wno-unticked-promoted-constructors
```
jorisdral, line 1037: :heart:

jorisdral (approved):

docs/StyleGuide.md
```diff
@@ -1046,9 +977,9 @@ the rules below, it is good practice to update the code's style to match them.
 
 18. __Records__:
 
-    For records we often use `NamedFieldPuns` to make it convenient to
-    extract fields from the record. We also tend to use `RecordWildCards`
-    when the fields of the record are all of the form
+    For records we often use `NamedFieldPuns` to make it convenient to extract
+    fields from the record. We use the following convention when naming fields
+    to avoid duplicate record fields (we do not use `DuplicateRecordFields`):
```
dnadales, line 982: I second [@jorisdral's suggestion](https://github.com/input-output-hk/ouroboros-consensus/pull/24/files#r1164045079). Should we mention that we discourage the use of `RecordWildCards`?

docs/StyleGuide.md
dnadales: I agree. Maybe [here](https://github.com/input-output-hk/ouroboros-consensus/pull/24/files#r1166667492).

docs/StyleGuide.md
dnadales: In certain circumstances I prefer not prefixing the accessors if the module is intended to be imported qualified. But for this particular case perhaps we should mention that using the first letters of `SomeRecord` is also an option (for instance when the type name is too long.

dnadales (approved):
Thanks!

docs/StyleGuide.md
Jasagredo: Done

docs/StyleGuide.md
Jasagredo: Done

PR #39, opened 2023-04-25
Move CHANGELOGS to `extra-doc-files`

Jasagredo:
# Description
As suggested by `cabal check`, the CHANGELOG files should be `extra-doc-files` instead.


jorisdral (approved):

.github/workflows/checks.yml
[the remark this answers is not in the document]
Jasagredo: Done

.github/workflows/checks.yml
```diff
@@ -6,14 +6,15 @@ concurrency:
 
 on:
   pull_request:
-  merge_group:
```
amesgen, line 9: This PR won't merge as GHA is not triggered anymore on the relevant commit in the merge queue: https://github.com/input-output-hk/ouroboros-consensus/commit/2e65ccb9c5fb5ac569d734091bc6af5dbdba8b67

amesgen: Same for the other workflows.

PR #46, opened 2023-04-26
Several small tweaks to scripts and CI

Jasagredo:


.gitignore
```diff
@@ -65,6 +65,7 @@ launch-*
 # emacs
 .auctex-auto
 *_flymake.hs
+.dir-locals.el
```
jorisdral, line 68: Would we want to take this opportunity to add other ignores? Looking at https://github.com/github/gitignore/blob/main/Global/Emacs.gitignore, there could be a few, but I don't use emacs so you decide!

jorisdral (approved):

.gitignore
Jasagredo: Good remark, but any sufficiently configured emacs probably wouldn't create any of those in the project itself. Thanks in any case!

cabal.project
```diff
@@ -27,23 +27,53 @@ packages:
 tests: True
 benchmarks: True
 
-package cardano-ping
+package cardano-ledger-core
+  flags: +asserts
+
+package cardano-ledger-shelley
+  flags: +asserts
```
amesgen, line 34: Do we actually want this? E.g. it could be that this impacts performance significantly, in particular for db-analyser?

dnadales: I will add a note in the `db-analyser` and `db-synthesiser documentation stating that people should disable these asserts if they build these tools in our repo. If they build it outside our repo then this will not be a problem.

dnadales: https://github.com/input-output-hk/ouroboros-consensus/pull/51

PR #53, opened 2023-04-28
Make the LMDB backing store thread safe by using `RAWLock`s

### LINKED ISSUE #52: Make the LMDB backing store thread safe
We should investigate whether the LMDB backing store is thread safe, and make it thread safe where it is not. One part of the LMDB backing store that is not thread safe is the `bsClose` function:

```haskell
    bsClose :: m ()
    bsClose = do
      guardDbClosed dbClosed
      Trace.traceWith dbTracer $ TDBClosing dbFilePath
      openHandles <- IOLike.readTVarIO dbOpenHandles
      for_ openHandles HD.bsvhClose
      IOLike.atomically $ IOLike.modifyTVar dbClosed (const True)
      liftIO $ LMDB.closeEnvironment dbEnv
      Trace.traceWith dbTracer $ TDBClosed dbFilePath
```

In this definition, the list of open handles is read atomically, but is only until after we have closed each handle that we declare the whole of the backing store as closed. This means that new value handles can be created that will not be closed automatically, even though the backing store will become closed at some point. Value handles that are still open after the backing store is closed will throw errors if we try to submit reads to them, since the underlying LMDB environment/database will have been closed.

jorisdral:
# Description

Resolves #52.

We make the LMDB backing store thread safe by using `RAWLock`s for the backing store and each backing store value handle. The locks keep track of the `Status` of the resource, which can either be `Open` or `Closed`. Once such a `Status` becomes `Closed`, all subsequent operations that try to acquire a lock will fail. See `Ouroboros.Consensus.Storage.LedgerDB.BackingStore.LMDB.Status`.


jorisdral:
Having rebased the current branch on top of the branch that introduces a micro-benchmarking setup for backing store operations, we can gather some initial numbers. Firstly, running a benchmark that performs one write per 100 reads without the `RAWLock`ing behaviour shows us the following:
```sh
.../ouroboros-consensus$ cabal run backingstore-bench -- -p "oneWritePer100Reads LMDB"
All
  BackingStore
    oneWritePer100Reads LMDB  10_000: OK (5.48s)
      199  ms ±  17 ms

All 1 tests passed (5.49s)
```
Running the same benchmark with the `RAWLock` changes introduced in this PR, we see that the operations become ~80 ms slower
```sh
.../ouroboros-consensus$ cabal run backingstore-bench -- -p "oneWritePer100Reads LMDB"
All
  BackingStore
    oneWritePer100Reads LMDB  10_000: OK (6.19s)
      283  ms ± 2.1 ms
```
A performance regression was to be expected, since we now acquire/release locks for every backing store operation that we perform. However, we do get a reasonable guarantee of thread safety in return.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore/LMDB.hs
```diff
@@ -522,50 +524,59 @@ mkLMDBBackingStoreValueHandle db = do
...
   let
-    bsvhClose :: m ()
-    bsvhClose = do
-      Trace.traceWith tracer TVHClosing
-      guardDbClosed dbClosed
-      guardBsvhClosed vhId dbOpenHandles
+    -- | Clean up a backing store value handle by committing its transaction
+    -- handle.
+    cleanup :: Cleanup m
+    cleanup = Cleanup $ do
```
jasagredo, line 533:
```suggestion
    cleanup = Cleanup $ 
```

jasagredo (commented):
Looks good, are the results on benchmarking acceptable?

PR #56, opened 2023-04-28
Pending updates due to the addition of handshake queries

jprider63:
These were the pending updates from input-output-hk/ouroboros-network/pull/4256 to ouroboros-network before ouroboros-consensus was split out.

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
@@ -528,7 +528,7 @@ instance CardanoHardForkConstraints c
       , (NodeToClientV_15, CardanoNodeToClientVersion11)
       ]
 
-  latestReleasedNodeVersion _prx = (Just NodeToNodeV_10, Just NodeToClientV_14)
+  latestReleasedNodeVersion _prx = (Just NodeToNodeV_11, Just NodeToClientV_15)
 
 {-------------------------------------------------------------------------------
   ProtocolInfo
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -723,11 +723,13 @@ stdVersionDataNTN networkMagic diffusionMode peerSharing = NodeToNodeVersionData
     { networkMagic
     , diffusionMode
     , peerSharing
+    , query = False
     }
 
 stdVersionDataNTC :: NetworkMagic -> NodeToClientVersionData
 stdVersionDataNTC networkMagic = NodeToClientVersionData
     { networkMagic
+    , query = False
     }
 
 stdRunDataDiffusion ::


dnadales:
Thank you. Please note there are some conflicts that need to be resolved.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -528,7 +528,7 @@ instance CardanoHardForkConstraints c
       , (NodeToClientV_15, CardanoNodeToClientVersion11)
       ]
 
-  latestReleasedNodeVersion _prx = (Just NodeToNodeV_10, Just NodeToClientV_14)
+  latestReleasedNodeVersion _prx = (Just NodeToNodeV_11, Just NodeToClientV_15)
```
dnadales, line 531: Do we need to update this? If Conway is not imminent we should skip this change.

jprider63: That's probably a question for @coot. I'm not familiar with the release schedule.

coot: @dnadales I am ok with including it, but first we need to publish new version of `ouroboros-network` which includes input-output-hk/ouroboros-network#4256.  I merged it today, I'll release new version soon.

coot: I am confused why this PR didn't require a new version of `ouroboros-network-api`.

bolt12: Peer Sharing is on `V_11` so this would mean not needing to enable `TestEnableDevelopmentNetworkProtocols` flag to get it, right? (Of course one would still need to manually enable the `PeerSharing` flag)

coot: @bolt12 thanks for reminding me that, this means we need to use `V_12` for handshake changes, it will be safer as I am not sure if things will get into `8.0`.

coot: @jprider63 in input-output-hk/ouroboros-network#4538 I added `NodeToNodeV_12` and `NodeToClientV_16`, you'll need to change it here.

coot: You will also need to add:

*  `(NodeToNodeV_11, CardanoNodeToNodeVersion7)` to `supportedNodeToNodeVersions`
*  `(NodeToClientV_16, CardanoNodeToClientVersion11)` to `supportedNodeToClientVersions`

coot: If we use `ouroboros-network-api-0.5.0.0` (which is published to CHaP) then we don't need to change anything here.

coot: `ouroboros-network-api-0.5.0.0`, `ouroboros-network-0.5.0.2` were released in this PR:
https://github.com/input-output-hk/cardano-haskell-packages/pull/279

amesgen:
I included this change while updating Network in #85, so I think this can be closed.

PR #58, opened 2023-05-01
Expose a method for retrieving UTxO set size from the ChainDB

### LINKED ISSUE #57: Tracers: Determining UTxO set size in the UTxO HD era
Since the UTxO is no longer entirely stored in memory, we're losing direct means of estimation of its size.

Yet, we still need this information, in the long run -- for performance estimation/prediction reasons.

jorisdral:
# Description

Resolves #57.

We expose a method of retrieving the UTxO set size. More generally, the method can retrieve the size of any table. 

To achieve this, we do the following.
1. The backing store value handle interface is extended with a `bsvhStat` function, which is intended to provide statistics that are common for each backing store implementation, such as the current sequence number (last written slot number) and the number of entries stored on-disk.  This `bsvhStat` function provides the statistics for a consistent view of the database.
2. The changelog now counts the number of inserts and deletes of all diffs in the diff sequence. These numbers determine how the size of a `Data.Map` should increase if the diffs in the sequence are applied to it. This comes with some caveats relating to redundant inserts/deletes: inserting or deleting a key twice does not increase the size of a `Data.Map`. However, the UTxO access pattern is such that that there are most one insert and delete for each UTxO address, and a delete is always preceded by an insert. As such, we do not suffer from the caveats in our solution.
3. Summing the number of entries stored on-disk (step 1) and the total size effect of the changelog (step 2) gives us the total number of UTxO entries, corresponding to the logical current ledger state. This read-forward pattern is similar to the pattern we use when we retrieve the UTxO entries to use for various operations, like block application.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore.hs
@@ -25,6 +25,8 @@ module Ouroboros.Consensus.Storage.LedgerDB.BackingStore (
   , bsRead
   , castBackingStoreValueHandle
   , withBsValueHandle
+    -- * Statistics
+  , Statistics (..)
     -- * Ledger DB wrappers
   , LedgerBackingStore
   , LedgerBackingStore'
@@ -95,6 +97,8 @@ data BackingStoreValueHandle m keys values = BackingStoreValueHandle {
     -- Absent keys will merely not be present in the result instead of causing a
     -- failure or an exception.
   , bsvhRead      :: !(keys -> m values)
+    -- | Retrieve statistics
+  , bsvhStat      :: !(m Statistics)
   }
 
 castBackingStoreValueHandle ::
@@ -110,6 +114,7 @@ castBackingStoreValueHandle f g bsvh =
     , bsvhRangeRead = \(RangeQuery prev count) ->
         fmap f . bsvhRangeRead $  RangeQuery (fmap g prev) count
     , bsvhRead = fmap f . bsvhRead . g
+    , bsvhStat
     }
   where
     BackingStoreValueHandle {
@@ -117,6 +122,7 @@ castBackingStoreValueHandle f g bsvh =
       , bsvhAtSlot
       , bsvhRangeRead
       , bsvhRead
+      , bsvhStat
       } = bsvh
 
 data RangeQuery keys = RangeQuery {
@@ -164,6 +170,27 @@ withBsValueHandle store kont =
       (bsValueHandle store)
       bsvhClose
       kont
+
+{-------------------------------------------------------------------------------
+  Statistics
+-------------------------------------------------------------------------------}
+
+-- | Statistics for a key-value store.
+--
+-- Using 'bsvhStat' on a value handle only provides statistics for the on-disk
+-- state of a key-value store. Combine this with information from a
+-- 'DbChangelog' to obtain statistics about a "logical" state of the key-value
+-- store. See 'getStatistics'.
+data Statistics = Statistics {
+    -- | The last slot number for which key-value pairs were stored.
+    --
+    -- INVARIANT: the 'sequenceNumber' returned by using 'bsvhStat' on a value
+    -- handle should match 'bsvhAtSlot' for that same value handle.
+    sequenceNumber :: !(WithOrigin SlotNo)
+    -- | The total number of key-value pair entries that are stored.
+  , numEntries     :: !Int
+  }
+  deriving stock (Show, Eq)
 
 {-------------------------------------------------------------------------------
   Ledger DB wrappers
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore/LMDB.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore/LMDB.hs
@@ -37,6 +37,7 @@ import           Data.Functor (($>), (<&>))
 import           Data.Map (Map)
 import           Data.Map.Diff.Strict
 import qualified Data.Map.Strict as Map
+import           Data.Monoid (Sum (..))
 import qualified Data.Set as Set
 import qualified Data.Text as Strict
 import qualified Database.LMDB.Simple as LMDB
@@ -243,6 +244,11 @@ readDbStateMaybeNull ::
      LMDB.Database () DbState
   -> LMDB.Transaction mode (Maybe DbState)
 readDbStateMaybeNull db = LMDB.get db ()
+
+readDbState ::
+     LMDB.Database () DbState
+  -> LMDB.Transaction mode DbState
+readDbState db = readDbStateMaybeNull db >>= maybe (liftIO . throwIO $ DbErrNoDbState) pure
 
 withDbStateRW ::
      LMDB.Database () DbState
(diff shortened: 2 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -380,6 +380,9 @@ data ChainDB m blk = ChainDB {
                 (PointNotFound blk)
                 (LedgerTables (ExtLedgerState blk) ValuesMK))
 
+
+    , getCurrentLedgerTableSize :: m (Either (WithOrigin SlotNo, WithOrigin SlotNo) Int)
```
jorisdral, line 384: This needs some documentation

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -566,3 +568,17 @@ getLedgerTablesAtFor ::
 getLedgerTablesAtFor pt keys LgrDB{ varDB, lgrBackingStore } = do
   lgrDb <- readTVarIO varDB
   ReadsKeySets.getLedgerTablesAtFor pt keys lgrDb lgrBackingStore
+
+{-------------------------------------------------------------------------------
+  Temp
+-------------------------------------------------------------------------------}
```
jorisdral, line 574:
```suggestion
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/ReadsKeySets.hs
```diff
@@ -202,3 +205,29 @@ forwardTableKeySets ::
...
+{-------------------------------------------------------------------------------
+  Temp
+-------------------------------------------------------------------------------}
+
+getCurrentLedgerTableSize ::
+     (Monad m, HasLedgerTables l)
+  => LedgerDB l
+  -> LedgerBackingStore m l
+  -> m (Either (WithOrigin SlotNo, WithOrigin SlotNo) Int)
+getCurrentLedgerTableSize lgrDb lbs = do
```
jorisdral, line 218: I'm not sure in which module this function should be placed

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/ReadsKeySets.hs
```diff
@@ -202,3 +205,29 @@ forwardTableKeySets ::
             (LedgerTables l ValuesMK)
 forwardTableKeySets dblog =
   forwardTableKeySets' (changelogDiffAnchor dblog) (changelogDiffs dblog)
+
+{-------------------------------------------------------------------------------
+  Temp
+-------------------------------------------------------------------------------}
+
+getCurrentLedgerTableSize ::
```
jorisdral, line 213: Needs documentation

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -566,3 +568,17 @@ getLedgerTablesAtFor ::
...
+
+getCurrentLedgerTableSize ::
```
jorisdral, line 576: Needs documentation

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
dnadales: Indeed. If possible I think it's better to add it in this PR, otherwise, it'll become more difficult the more time it passes.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore.hs
```diff
@@ -62,6 +62,9 @@ data BackingStore m keys values diff = BackingStore {
   , bsValueHandle :: !(m (WithOrigin SlotNo, BackingStoreValueHandle m keys values))
     -- | Apply a valid diff to the contents of the backing store
   , bsWrite       :: !(SlotNo -> diff -> m ())
+    -- | Retrieve statistics: the sequence number, and the number of entries
```
dnadales, line 65: I'm just wondering if we want to use a generic name such as `stats` or if we want to make this more descriptive of what these stats actually are. I have no strong opinion.

dnadales (commented):
Besides adding more comments I wonder if we could test this new functionality somehow.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore/LMDB.hs
```diff
@@ -481,10 +482,19 @@ newLMDBBackingStoreInitialiser dbTracer limits sfs initFrom = do
                pure (dbsSeq, s {dbsSeq = At slot})
              Trace.traceWith dbTracer $ TDBWrite oldSlot slot
 
+           bsStat :: m (WithOrigin SlotNo, Int)
+           bsStat = do
+             guardDbClosed dbClosed
+             liftIO $ LMDB.readWriteTransaction dbEnv $ withDbStateRW dbState $ \s@DbState{dbsSeq} -> do
+               constn <- traverseLedgerTables (\(LMDBMK _ dbx) -> ConstMK <$> LMDB.size dbx) dbBackingTables
+               let n = getSum $ foldLedgerTables (Sum . getConstMK) constn
```
jasagredo, line 490: This means: the sum of all the items in LMDB in all the tables. This is fine because there is only one LMDBMK table, the UTxO table. But it will not be in the future. This probably deserves a second thought. 

I also wonder, why don't we use the `mdb_stats` thingie?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
jorisdral: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/DiffSeq.hs
```diff
@@ -125,9 +130,13 @@ newtype DiffSeq k v =
...
     -- | Cumulative length
-    rmLength :: {-# UNPACK #-} !Length
+    rmLength     :: {-# UNPACK #-} !Length
     -- | Cumulative diff
-  , rmDiff   :: !(Diff k v)
+  , rmDiff       :: !(Diff k v)
+    -- | Cumulative number of inserts
+  , rmNumInserts :: !(Sum Int)
+    -- | Cumulative number of deletes
+  , rmNumDeletes :: !(Sum Int)
```
jorisdral, line 139: Reconsider putting this in the internal measure instead

cabal.project
```diff
@@ -65,3 +65,12 @@ source-repository-package
...
     ouroboros-network-api
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/anti-diffs
+  tag: d03496f11d453d73baec4523621dc47b7c1e9313
+  --sha256: 07zfzfmfww9xg9y9kkb1gyqc1iwhkv50janiyxrymzq9ilv2lh8y
+  subdir:
+    diff-containers
+    fingertree-rm
```
jasagredo, line 76: I would favor a new release instead of a s-r-p for now

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/ReadsKeySets.hs
```diff
@@ -202,3 +205,34 @@ forwardTableKeySets ::
...
+-- current ledger state.
+getCurrentLedgerTableSize ::
+     (Monad m, HasLedgerTables l)
+  => LedgerDB l
+  -> LedgerBackingStore m l
+  -> m (Either (WithOrigin SlotNo, WithOrigin SlotNo) Int)
+getCurrentLedgerTableSize lgrDb lbs = do
+    Statistics{sequenceNumber = seqNo', numEntries = n} <- bsStat bs
+    if seqNo /= seqNo' then
+      pure $ Left (seqNo, seqNo')
```
jasagredo, line 223: This probably should be wrapped in some error type, it is always difficult to see which one is which in this tuples when you get an error

jasagredo (approved):
Looks good!

dnadales (approved):

dnadales:
This PR is currently waiting on the mempool rebase work.

jorisdral:
> This PR is currently waiting on the mempool rebase work.

Unblocked on the mempool rebase work.

The PR is now blocked on #138, which introduces a bug fix and some CI changes that are required. #138 should be merged first.

jorisdral:
@jasagredo @dnadales I've made a change that I would like your input on: 

Instead of in the `bsStat` function, I've moved the functionality into the value handle, so it is called `bsvhStat` instead. One optimisation that I haven't implemented yet is that each time we call `bsvhStat`, we go to disk to read statistics. However, since a value handle has consistent view of the database, this will return the same result every time (with respect to number of entries and sequence number, at least -- I'm not sure if other statistics can change for a value handle). For this reason, we could cache the result in something like an `MVar`, and only go to disk the first time this function is called.

jorisdral:
> @jasagredo @dnadales I've made a change that I would like your input on:
> 
> Instead of in the `bsStat` function, I've moved the functionality into the value handle, so it is called `bsvhStat` instead. One optimisation that I haven't implemented yet is that each time we call `bsvhStat`, we go to disk to read statistics. However, since a value handle has consistent view of the database, this will return the same result every time (with respect to number of entries and sequence number, at least -- I'm not sure if other statistics can change for a value handle). For this reason, we could cache the result in something like an `MVar`, and only go to disk the first time this function is called.

EDIT: Or: retrieve those statistics in `bsValueHandle`, such that retrieving the statistics afterwards is always a pure operation. This adds a little bit of overhead to `bsValueHandle` though.

jasagredo (approved):

PR #59, opened 2023-05-01
draft ComponentDiagram.md

nfrisby:
Drafts the `ComponentDiagram.md` file in the documentation.


nfrisby:
This PR migrates from the old repo's PR input-output-hk/ouroboros-network#4368

nfrisby:
@fraser-iohk as a newer member, do you have any feedback regarding this PR? Thanks!

nfrisby:
Rendered: https://github.com/input-output-hk/ouroboros-consensus/blob/nfrisby/consensus-component-diagram/docs/website/docs/ComponentDiagram.md

fraser-iohk:
the only comment I really have is about the diagram: is it possible to modify the edges so they don't need a separate legend? I imagine that it starts to look really cramped, though

nfrisby:
Ah. Yeah, I think it'd be too cramped if I tried to keep the same amount of information. And, unsurprisingly, in my opinion that is the right amount of information to put in the diagram. But maybe we could have a few variants that strip out some of that information. For starters, for example, the edge would just be labeled with T, B, L, H(B), or H(L).

abailly-iohk:
I would keep the explicite labelling of edges, perhaps slightly abbreviated, otherwise it makes it hard at a glance to understand what's going on. Do I understand this some "desired state" diagram? If not, then I would stick to use the current names of the components as they appear in the code in order to not confuse the somewhat informed reader who knows the main concepts of the consensus.

nfrisby:
OK, I updated the box labels to favor the established names but also relate them to the better conceptual names I original used in this diagram.

I think this is ready for merge. Any objections? @dnadales

dnadales (approved):
Thank you Nick!

PR #62, opened 2023-05-02
HardWordWidsom.md: spell out the forecast horizon

nfrisby:
Added an answer to "Which slots/times/epochs can we translate?"

### DIFF
--- a/docs/website/docs/HardWonWisdom.md
+++ b/docs/website/docs/HardWonWisdom.md
@@ -95,7 +95,7 @@ A couple more observations:
 
 - At least one part of the ledger _does_ use the `EpochInfo Either` as a test: the validation of the `ValidityInterval` of a transaction that contains a Plutus script.
   The code here accommodates the inflexibility of the `Predicate` rule by using an additional `whenFailureFree` combinator to skip over invalid tx, thus avoiding the computation that would require the result of the `EpochInfo` that instead returns `Left`.
-  ... I wonder if every use of the `epochInfo` could do that same thing.
+  (TODO ... Frisby wonders if every use of the `epochInfo` could do that same thing).
 
 So, either by mimicking the approach of the existing `ValidityInterval` validation logic or by altering the STS innards to allow short-circuiting failure, we could reify the `EpochInfo` failures into the `PredicateFailure` hierarchy, and thereby leverage the types to force each invocation of the Ledger API to independently handle the possibility of `PastHorizonException`s.
 But it's not obvious that that is definitely worth the extra complexity it would introduce on the Ledger code.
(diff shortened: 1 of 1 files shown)


nfrisby:
Rendered https://github.com/input-output-hk/ouroboros-consensus/blob/nfrisby/hardwonwisdom-hfc-horizon-details/docs/website/docs/HardWonWisdom.md#which-slotstimesepochs-can-we-translate

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,123 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+
+- These are ultimately used for every slot/epoch/time translation: `EpochInfo` passed to the ledger, the leadership check in a block-producing node, and all Consesus queries (see `ouroboros-consensus:Ouroboros.Consensus.HardFork.Combinator.Ledger.Query.GetInterpreter`).
+
+- Every downstream use, such as the `query-slot` CLI command ultimately boils down to issuing the `GetInterpreter` query to the node itself and then invoking `interpretQuery`.
+
+- Crucially, that interpretation can fail when the given interpreter cannot answer the given query, which is precisely the topic of the Office Hours question.
+
+- Remark.
+  A significantly fresher `Interpreter` will be able to answer more queries than a stale `Interpreter`.
+  The `ouroboros-consensus:Ouroboros.Consensus.HardFork.History.Caching` module captures that idea.
```
amesgen, line 227: Maybe "captures" -> "makes use of", as the most direct way to interpret that remark is to always use the `Interpreter` made from the ledger state of the tip of the current selection.

Also, I am not quite interested in how expensive creating a `Summary`/an `Interpreter` from a `LedgerState` actually is; intuitively, it seems very cheap :thinking: OTOH, this abstraction is super lightweight, so no need to change anything even if it is super cheap.

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,123 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+On `mainnet` Cardano, only blocks in an era can cause that era to end.
+This also implies each era with a known end time on a `mainnet` chain must contain some blocks on that chain.
+On the other hand, the Consensus Layer configuration inputs do permit overriding the era transition rules for the sake of testing, in which case a ledger state could anticipate the end time of several subsequent eras and eras could be empty (eg main use case is skipping from Byron to some later Shelley era ASAP, in which case all the intervening eras have no blocks but also have no slots!).
+
+There is one last notable caveat at this abstract level.
+The Hard Fork Combinator safe zone for each `mainnet` era is exactly the minimum number of slots of forewarning before the end of that era.
+One might naturally expect this is the stability window (see the above entry in this file for why this is `3k/f` in Praos), but it is instead twice that.
+The best resource we have for motivating this decision is [Edsko de Vries' IOG seminar from 2020 June on the HFC and time](https://drive.google.com/file/d/1QIJ-VBlj-txB6K6E7DIEnY5TzaD89qQm/view).
```
amesgen, line 197: I think something is being mixed up here; we set the `eraSafeZone` field in `EraParams` (one of which for each era live in `Shape` via `HardForkConsensusConfig`) to $3k/f$ and not $6k/f$ for all Shelley-based eras:

https://github.com/input-output-hk/ouroboros-consensus/blob/e19a58d078b882e5180d9ef591a1f4e9a1987fe0/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L134-L140

(for reference: [`SL.computeStabilityWindow`](https://github.com/input-output-hk/cardano-ledger/blob/180271602640bcac1214084b6de61d0468332f00/eras/shelley/impl/src/Cardano/Ledger/Shelley/StabilityWindow.hs#L19-L20))

I also skimmed the linked talk again and couldn't find where "safe zone = two stability windows" is stated.

Maybe the difference is between the number of slots it takes for an proposal to come into effect (which is $6k/f$, and I think is what you are trying to describe here, in particular with the reference to the `PUPP` ledger rule) compared to the number of slots ahead of a ledger state in which an era transition is guaranteed not to occur if it is not already known (cf. which is the definition of "safe zone" in [section 17.3.3 in the report](https://input-output-hk.github.io/ouroboros-network/pdfs/report#subsection.17.3.3))?

nfrisby: Hmm. Yes. I think I did the conflation you were talking about.

1) How early does the ledger state determine when the transition will happen, _assuming_ it does happen? At least 6k/f before the epoch transition.

2) How early does the ledger state determine when the transition will happen _and_ that it definitely will happen? At least 3k/f before the epoch transition.

I'll think through this more carefully and rewrite.

nfrisby: I rewrote it.

nfrisby: I added another couple paragraphs; spelling out the fundamentals of Header-Body Split's requirements on the stability of header validity. Phew!

dnadales (approved):

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+No Cardano era is specified to never end, and we currently do not foresee every doing so.
+The behavior is only supported for tests and for reuse by other blockchain implementors that are not concerned with extensibility.
+
+### The Wrinkle with Time
+
+For interpreting slot/time/epoch translation queries, the key motivating fact is that each era can have a different size of epoch (in slots) and/or a different length of slot (a time duration).
+Thus, we cannot handle a translation query unless we know which era its argument is in and the durations of all preceding eras, or, equivalently, the start times of all eras up-to-and-including the argument is in.
+
+Each query is handled by some piece of code and with respect to some ledger state.
+The code (including config files) determines the supported eras (including their epoch size and slot length), and the ledger state determines which prefix of eras has known start times.
```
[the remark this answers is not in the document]
nfrisby: Yep. Each ledger state declares what era it is in, which is therefore the current one, which thereby determines which eras are the preceding ones.

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+The known start times are enough to recognize when the given argument is during some era with a known end time, in which case the translation succeeds.
+Otherwise, we need more information to know whether the given argument is during the last era with a known start time, ie the first era with an unknown end time.
+Even if the query argument is after that start time, it _might_ instead be in a future era with a different epoch size/slot length, and so the correct translation may differ from the hypothetical response derived from the assumption the query argument precedes the first unknown end time.
```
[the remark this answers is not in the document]
nfrisby: I'll double-check, bu I think: throughout this right up, the query argument is either a slot, an epoch number, or a UTC time: it's one of those that you're trying to translate to one of the others.

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+And, at that same time, all valid Cardano chains and their prefixes (aka all valid Cardano blocks) will inhabit some prefix of that sequence---in the absence of catastrophic release and/or operator failure.
+
+A Hard Fork Combinator ledger state provides enough information to determine which era it is in.
+It must also determine (ie record) when that era and every preceding era started.
+Trivially, this also determines when all the _preceding_ eras ended (ie when the successor started).
+The ledger state _might_ also determine when the ledger state's own era will end.
+Thus, every ledger state determines the start time of N eras and the end time of the preceding N-1 eras (note: there's one more start time than end times), where the ledger state inhabits either the N-1th or the Nth era.
+Subsequent ledger states may have greater Ns, but the invariant remains true.
+Every N will be at least 1, at most the length of the sequence of eras the code supports, and subsequent ledger states on a chain cannot have a lesser N.
+
```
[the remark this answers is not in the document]
amesgen: We could also add [KaTeX to Docusaurus](https://docusaurus.io/docs/markdown-features/math-equations).

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+  So the `d` parameter is something that even the laxer rule could not accomodate (even when assuming it can't increase, because the its overall influence on honest headers is not monotonic).
+
+- As of epoch 257, `d=0` (ie no slots are overlaid).
+  Since then, no change to the Cardano protocol parameters (including era transitions) has changed the (average) number of slots any given issuer leads during any given interval of slots.
+  Also since then, every change that has at all affected the validity of headers has been an era transition.
+  Therefore, _none of those subsequent changes truly required a stability window of warning_.
+  Honest headers after such change would correctly identify themselves as such by claiming the new era, and therefore would be correctly validated.
+  An adversarial header, in this hypothetical, could have falsely claimed to be from the new era, but the crucial insight is that doing so wouldn't have (TODO "significantly" or "at all"?) increased the chance the adversary could create an apparently-valid chain that is long enough to cause problems.
+
+Recall that that discussion was merely theoretical, to better delineate the fundamental requirements.
```
amesgen, line 235: I think it would make sense to specify upfront what invariant that is present in the current design/implementation, but which you are temporarily discarding here for the (very useful!) sake of exploring the design space, i.e. "assume the safe zone is empty".

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+The real Cardano ledger always has and still does ensure that header validity is fully determined at least one stability window in advance.
+We have had no motivating reason to eliminate that simplification, as of yet; so far we don't anticipate one arising soon.
+(
+Currently, the protocol parameter changes are currently even known two stabilty windows early!
+But some factors are still just singly stable, eg the stake distribution is only frozen one stability window before the next epoch.
+)
+
+TODO What about the config file overrides that result in empty epochs?
```
amesgen, line 243:
```suggestion
TODO What about the config file overrides that result in empty eras?
```
?

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
...
+  So the `d` parameter is something that even the laxer rule could not accomodate (even when assuming it can't increase, because the its overall influence on honest headers is not monotonic).
+
+- As of epoch 257, `d=0` (ie no slots are overlaid).
+  Since then, no change to the Cardano protocol parameters (including era transitions) has changed the (average) number of slots any given issuer leads during any given interval of slots.
+  Also since then, every change that has at all affected the validity of headers has been an era transition.
+  Therefore, _none of those subsequent changes truly required a stability window of warning_.
+  Honest headers after such change would correctly identify themselves as such by claiming the new era, and therefore would be correctly validated.
+  An adversarial header, in this hypothetical, could have falsely claimed to be from the new era, but the crucial insight is that doing so wouldn't have (TODO "significantly" or "at all"?) increased the chance the adversary could create an apparently-valid chain that is long enough to cause problems.
```
amesgen, line 233: This seem like a very interesting kind of "trap" headers, i.e. headers where it is impossible to follow up with a matching valid block, right?

docs/website/docs/HardWonWisdom.md
amesgen: Thanks, that is a very useful thing to explore, I left a few comments.

Also, this might be sufficiently separate from the overall topic to move to its own section, or even put it in [section 24.5 ("Eliminating safe zones") in the report](https://input-output-hk.github.io/ouroboros-network/pdfs/report/report.pdf#section.24.5) (or the succeeding section).

docs/website/docs/HardWonWisdom.md
```diff
@@ -112,3 +112,196 @@ REQ2: If L≤k (see REQ1) and L<R, we must validate at least L+1 of their header
 The most demanding case of REQ2 is L=k: at most we'll need to validate k+1 of the peer's headers.
 Thus using HCG window as Stability Window ensures that forecasting can't disrupt REQ2 when the peer is serving honest blocks.
 (We can tick the intersection ledger state to their first header, and then forecast 3k/f from there which by HCG will get us at least the remaining k headers if they're serving an honest chain.)
+
+## Which slots/times/epochs can we translate?
+
+TODO Confirm and emphasize that this exact same logic is used by the node itself to determine which slot the current wall clock is in.
```
amesgen, line 118: You mean whether `hardForkBlockchainTime` uses that mechanism? In that case, the answer is positive.

nfrisby:
ef51415ab07446e657c284a45b88d8a07a108516 tips a series of TOSQUASH commits that resolved all of the remaining comments.

350a5492b9e4b7b64660e29e199e489829b686b7 squashes that down into a single commit, which I am now merging.

Note: I intentionally left some TODOs in the text: they'd have to be resolved before this content gets promoted to some more mature document.

PR #69, opened 2023-05-09
Update to `diff-containers-1.0.0.0` and `fingertree-rm-1.0.0.0`

jorisdral:
# Description

Update the UTxO HD branch to use the latest version of the anti-diffs packages.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/DiffSeq.hs
jorisdral, line 1: The module-level comment still refers to the `Group` class. We should update the documentation there and in other places that still refer to the `Group` class

PR #70, opened 2023-05-09
Add db-truncater tool, for truncating an ImmutableDB at a specified slot number

fraser-iohk:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/4512

# Description

Adds a new tool, `db-truncater`, that truncates the ImmutableDB at a specified point. This code heavily reuses command line parsers and mechanisms designed for `db-analyser`, and performs the truncate using the internal `deleteAfter` ImmutableDB function. In order to find the correct block for truncation without requiring that the user provide the hash of that block, it uses the iterator provided by `streamAll`, and truncates at the first block inhabiting the latest slot whose slot number is less than the target truncate point. 

-----

the code here is mostly complete, so worth reviewing, but I still plan to make some stylistic changes to conform to the style guide

potential improvements, either for now or for later:
- [x] add support for truncating at a specific block number, rather than specifically slot number
- [ ] make it faster by not streaming every single block from the DB

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -463,6 +463,8 @@ library cardano-tools
     Cardano.Tools.DBSynthesizer.Orphans
     Cardano.Tools.DBSynthesizer.Run
     Cardano.Tools.DBSynthesizer.Types
+    Cardano.Tools.DBTruncater.Run
+    Cardano.Tools.DBTruncater.Types
     Cardano.Tools.ImmDBServer.Diffusion
     Cardano.Tools.ImmDBServer.MiniProtocols
 
@@ -550,6 +552,21 @@ executable db-synthesizer
     , ouroboros-consensus
 
   other-modules:  DBSynthesizer.Parsers
+
+executable db-truncater
+  import:         common-exe
+  hs-source-dirs: app
+  main-is:        db-truncater.hs
+  build-depends:
+    , base
+    , cardano-crypto-wrapper
+    , ouroboros-consensus
+    , optparse-applicative
+    , ouroboros-consensus-cardano:{ouroboros-consensus-cardano, cardano-tools}
+
+  other-modules:
+    DBAnalyser.Parsers,
+    DBTruncater.Parsers,
 
 executable immdb-server
   import:         common-exe
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -560,13 +560,13 @@ executable db-truncater
   build-depends:
     , base
     , cardano-crypto-wrapper
-    , ouroboros-consensus
     , optparse-applicative
+    , ouroboros-consensus
     , ouroboros-consensus-cardano:{ouroboros-consensus-cardano, cardano-tools}
 
   other-modules:
-    DBAnalyser.Parsers,
-    DBTruncater.Parsers,
+    DBAnalyser.Parsers
+    DBTruncater.Parsers
 
 executable immdb-server
   import:         common-exe
--- a/ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
+++ b/ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
@@ -29,7 +29,7 @@ import           Prelude hiding (truncate)
 import           System.IO
 
 truncate ::
-     forall block . (Node.RunNode block, HasProtocolInfo block)
+     forall block. (Node.RunNode block, HasProtocolInfo block)
   => DBTruncaterConfig
   -> Args block
   -> IO ()
@@ -62,8 +62,8 @@ truncate DBTruncaterConfig{ dbDir, truncateAfter, verbose } args = do
         Nothing ->
           putStrLn $ mconcat
             [ "Unable to find a truncate point. This is likely because the tip "
-            , "of the ImmutableDB has a slot number less than the intended "
-            , "truncate point"
+            , "of the ImmutableDB has a slot number or block number less than "
+            , "the intended truncate point"
             ]
         Just (header, isEBB) -> do
           let HeaderFields slotNo blockNo hash = getHeaderFields header
@@ -79,7 +79,7 @@ truncate DBTruncaterConfig{ dbDir, truncateAfter, verbose } args = do
 -- | Given a 'TruncateAfter' (either a slot number or a block number), and an
 -- iterator, find the last block whose slot or block number is less than or
 -- equal to the intended new chain tip.
-findNewTip :: forall m blk c .
+findNewTip :: forall m blk c.
               (HasHeader (Header blk), Monad m)
            => TruncateAfter
            -> Iterator m blk (Header blk, c)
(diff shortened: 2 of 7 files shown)


ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -545,6 +547,21 @@ executable db-synthesizer
...
+  build-depends:
+    , base
+    , cardano-crypto-wrapper
+    , ouroboros-consensus
+    , optparse-applicative
+    , ouroboros-consensus-cardano:{ouroboros-consensus-cardano, cardano-tools}
+
+  other-modules:
+    DBAnalyser.Parsers,
+    DBTruncater.Parsers,
```
jasagredo, line 563: `cabal-fmt` will probably complain. Use `./scripts/ci/run-cabal-fmt.sh`

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
```diff
@@ -0,0 +1,109 @@
...
+          { immTracer = immutableDBTracer
+          , immRegistry = registry
+          , immCheckIntegrity = nodeCheckIntegrity (configStorage config)
+          , immCodecConfig = configCodec config
+          , immChunkInfo = chunkInfo
+          }
+
+    withDB immutableDBArgs $ \(immutableDB, internal) -> do
+      iterator <- ImmutableDB.streamAll immutableDB registry
+        ((,,) <$> GetSlot <*> GetBlock <*> GetIsEBB)
```
amesgen, line 55: > make it faster by not streaming every single block from the DB

Yeah, I imagine this would be nice; of course, the "dream" would be constant time truncation; but maybe, once you only read from the primary/secondary index (so removing `GetBlock`, it is already fast enough for e.g. the mainnet chain?

fraser-iohk: I've swapped out `GetBlock` for `GetHeader` (I'd missed `getHeaderFields`) -- am I correct in thinking that the headers are accessible entirely from the index?

nfrisby: I also anticipate that runtime linear in the length of the chain is fine for the anticipated uses of this tool.

On the other hand, it does seem relatively easy to reach logarithmic time: do a binary search over the block numbers (modulo EBBs :'() to find the greatest block number whose slot is below the requested truncation threshold.

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Types.hs
```diff
@@ -0,0 +1,14 @@
...
+import           Ouroboros.Consensus.Block.Abstract
+
+data DBTruncaterConfig = DBTruncaterConfig {
+    dbDir         :: FilePath
+  , truncatePoint :: TruncatePoint
+  , blockType     :: BlockType
+  , verbose       :: Bool
+  }
+
+newtype TruncatePoint = TruncatePoint SlotNo
```
nfrisby, line 13: I think this would be a good place to add some Haddock that clarifies whether the requested truncation slot is inclusive or exclusive.

nfrisby, line 13: `Point` is a fundamental type in our database. I recommend trying to avoid use of the word "point" when unless you're actually talking about `Point`.

Hence I would suggest renaming this type to `TruncationSlot`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/API.hs
```diff
@@ -397,7 +397,8 @@ data MissingBlock blk
     -- | There is no block in the slot of the given point.
   = EmptySlot (RealPoint blk)
     -- | The block and/or EBB in the slot of the given point have a different
-    -- hash.
+    -- hash. We return the 'HeaderHash' for each block we found with the
```
nfrisby, line 400: Nice. Thank you very much for this line.

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
amesgen: > am I correct in thinking that the headers are accessible entirely from the index?

No, the (secondary) index just stores the header offset+size as well as the header hash, see [Section 8.2.2 in the report](https://input-output-hk.github.io/ouroboros-network/pdfs/report/report.pdf#subsection.8.2.2). But using `GetHeader` instead of `GetBlock` should already result in way less disk IO as you don't have to read the full block (on mainnet, we have 88KiB max block size and 1.1KB header size).

> On the other hand, it does seem relatively easy to reach logarithmic time

Conceptually, constant time is simple, as the ImmDB uses a uniform slot chunk size (see https://github.com/input-output-hk/ouroboros-network/issues/4512), so finding the respective chunk file is just a division; the only obstacle is whether this is very annoying to implement given the existing API/how much one would have to expose in addition.

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Types.hs
```diff
@@ -0,0 +1,18 @@
+module Cardano.Tools.DBTruncater.Types (module Cardano.Tools.DBTruncater.Types) where
```
jasagredo, line 1: This export list is not in sync with other export lists. Please export everything explicitly

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
```diff
@@ -0,0 +1,121 @@
...
+    withDB immutableDBArgs $ \(immutableDB, internal) -> do
+      iterator <- ImmutableDB.streamAll immutableDB registry
+        ((,) <$> GetHeader <*> GetIsEBB)
+
+      mTruncatePoint <- findNewTip truncateAfter iterator
+      case mTruncatePoint of
+        Nothing ->
+          putStrLn $ mconcat
+            [ "Unable to find a truncate point. This is likely because the tip "
+            , "of the ImmutableDB has a slot number less than the intended "
```
jasagredo, line 65:
```suggestion
            , "of the ImmutableDB has a slot number/block number less than the intended "
```

jasagredo: I think?

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
```diff
@@ -0,0 +1,121 @@
...
+import qualified Ouroboros.Consensus.Storage.ImmutableDB as ImmutableDB
+import           Ouroboros.Consensus.Storage.ImmutableDB.Impl
+import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Consensus.Util.ResourceRegistry (runWithTempRegistry,
+                     withRegistry)
+import           Prelude hiding (truncate)
+import           System.IO
+
+truncate ::
+     forall block . (Node.RunNode block, HasProtocolInfo block)
```
jasagredo, line 32:
```suggestion
     forall block. (Node.RunNode block, HasProtocolInfo block)
```

jasagredo (approved):
Approved with some comments. Also the question whether doing a binary search or sth else still remains, but I think we can act on that one once the tool is being used and if requested?

ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBTruncater/Run.hs
fraser-iohk: yup, ty

PR #76, opened 2023-05-11
CI bits

amesgen:
- Split up Cicero into independent jobs such that the logs of different platforms are no longer interleaved, easing log inspection
 - We temporarily bumped the GHC on GHA to 9.2 in #46 as it started to randomly fail with weird out-of-memory errors on 8.10. We still have to support compiling for 8.10 for ~one month, so we are reverting that change here.

### DIFF
--- a/.github/workflows/cabal.project.local.Windows
+++ b/.github/workflows/cabal.project.local.Windows
@@ -1,5 +1,6 @@
 benchmarks:     True
-documentation:  True
+-- causes weird out-of-memory errors on 8.10
+documentation:  False
 reorder-goals:  True
 tests:          True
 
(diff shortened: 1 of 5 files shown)


jasagredo:
Will cicero build with docs enabled? i.e. an error in the haddocks will be caught somewhere?

amesgen:
> Will cicero build with docs enabled? i.e. an error in the haddocks will be caught somewhere?

Not yet, will change that :+1:

amesgen:
Actually, adding `-haddock` no longer makes the GHC parser more strict since GHC 9.0 (see https://gitlab.haskell.org/ghc/ghc/-/issues/23379). We could enable `-Winvalid-haddock`, which currently generates lots of warnings (e.g. all Haddock-ish comments on non-toplevel bindings), but they are not blocking Haddock generation.

I will try to enable Haddock again for GHA, maybe it works again :shrug:

jasagredo (approved):
:shipit:

PR #81, opened 2023-05-12
Build and publish the Consensus report

amesgen:
The previous URL in ouroboros-network is no longer accessible as they switched to the non-cumulative way of deploying to GitHub Pages.


jasagredo (approved):
Nice, thanks @amesgen !!

PR #82, opened 2023-05-12
Update the haskell-build workflow to run the benchmarks

dnadales:
This patch also removes the publish-mempool-benchmarks as it is no longer needed.


.github/workflows/haskell-build.yml
```diff
@@ -122,6 +129,22 @@ jobs:
     - name: Build projects [build]
       run: cabal build all -j
 
+      - name: Run benchmarks
```
dnadales, line 132: TODO: run only if we merge a PR to `main` or if the run was manually dispatched.

jasagredo:
What do we need to close this one?

jasagredo (approved):
:shipit:

jorisdral (approved):

PR #83, opened 2023-05-12
place upper bound on cardano-ledger-core

JaredCorduan:
# Description

In order to release `cardano-ledger-core` 1.2.0 on CHaPs, a revision in `ouroboros-consensus-cardano` was required, placing an upper bound. This commit mirrors the revision exactly.

See https://github.com/input-output-hk/cardano-haskell-packages/pull/254#discussion_r1191769688

We can remove this bound soon if we like, as I will be integrating the latest ledger packages into consensus now.


JaredCorduan:
I believe we can close this PR now, without merging it, since we merged https://github.com/input-output-hk/ouroboros-consensus/pull/85. yea?

PR #84, opened 2023-05-12
Use TraceLabelPeer for blockfetch and chainsync server

karknu:
Tracking which clients downloads blocks is very useful.


dnadales:
Thanks Karl! Could you [add a changelog entry](https://github.com/input-output-hk/ouroboros-consensus/blob/main/docs/website/docs/ReleaseProcess.md#adding-a-changelog-fragment) because we have breaking changes.

PR #85, opened 2023-05-12
update dependencies

JaredCorduan:
# Description

## CHaPs

This PR updates the CHaPs index state, and includes some integration of the new packages:

* `DPState` is now called `CertState`, and record names have changed as well
* `DState` and `PState` are now parametrized by `era`, not `crypto`.
* `Cardano.Ledger.UMapCompact` has been renamed `Cardano.Ledger.UMap`
* The conway serialization is still changing frequently, so corresponding golden files have been changed. Moreover, the old `DPState` in the ledger state now includes a new field, `VState`
* The dependency on `ouroboros-network-protocols` had to be set to exactly `0.5.0.0` (`0.5.0.1` was breaking the build)

## Fix registerGenesisStaking

In order for the ledger to bring back the TICKF [optimizations](https://github.com/input-output-hk/cardano-ledger/blob/115e140c733240d700a1fcfd817f0b0270505994/docs/adr/2022-12-12_007-optimize-ledger-view.md) a bug in `registerGenesisStaking` had to be fixed.
    
The function `registerGenesisStaking` is a hacky function, to be used only in testing. It sets up the ledger state in a way that is fraught with dragons :dragon:, since it bypasses our usual protections.
    
Indeed, a recent change to the ledger resulted in breaking this function. In particular, the ledger now memoizes the per-stake-pool stake distribution in a way that needed to be accounted for inside `registerGenesisStaking`.
    
The bug was caught by tests in the cardano-node which relied on `registerGenesisStaking`.


ouroboros-consensus-diffusion/changelog.d/20230512_142120_jaredcorduan_update_ledger.md
```diff
@@ -0,0 +1,5 @@
+### Patch
+
+- Set ouroboros-network-protocols dependency to ==0.5.0.0
+
+-->
```
JaredCorduan, line 5:
```suggestion
```

jasagredo (approved):
This looks good. The golden artifacts are still shown in the git diff but it is not a blocker.

amesgen:
> The golden artifacts are still shown in the git diff but it is not a blocker.

#86 only caused the diffs to no longer be expanded (i.e. you now see "Some generated files are not rendered by default."), but AFAIK there is no way to tell GitHub to completely hide the files from the "Files changed" view.

PR #90, opened 2023-05-16
Use a minimum of 10s reconnection delay

karknu:
Use a minimum of 10s reconnection delay. Non-p2p nodes will disconnect from a p2p-node when chainsync exits. By waiting 10s to reconnect we are less likely to try and promote the peer to hot while the same peer is asynchronously being demoted to cold.

# Description

this PR might resolve.


jasagredo:
This should include a minimal changelog entry mentioning the change in behavior

dnadales (approved):

PR #95, opened 2023-05-22
Restore en-/decoding compatibility for `GetCurrentPParams`

amesgen:
https://github.com/input-output-hk/ouroboros-network/pull/4349 changed/fixed the encoding of `PParams`, which broke compatibility of older clients with Node 8.0. This PR restores compatibility, by making the en-/decoding version-dependent.

See the commit message for some implementation details. Also, note how the golden files changed due to this PR:

 - Pre-Alonzo serialization did not change.
 - Alonzo and Babbage changed, but only for `CardanoNodeToClientVersion <= 10`; these are enabled by `NodeToClient <= 14`, which are the currently released node-to-client versions.
 - Note that no golden files changed for `CardanoNodeToClientVersion{11,12}` (which are enabled by `NodeToClientV_{15,16}`). `NodeToClientV_15` will be released in Node 8.1, and indeed, we want to use the new and fixed encoding when this version is negotiated.


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -15,6 +15,7 @@
 {-# LANGUAGE TypeFamilies               #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE FlexibleContexts           #-}
```
dnadales, line 18: Do we want to move this up?

PR #102, opened 2023-05-22
force new strict MVar value before calling putTMVar in updateMVar

fraser-iohk:
Part of [ouroboros-network#4006](https://github.com/input-output-hk/ouroboros-network/issues/4006)

Adds a strictness annotation in `Ouroboros.Consensus.Util.MonadSTM.StrictMVar.updateMVar` to force the value we're putting into the `StrictMVar` (`'a`) before checking its invariant. This function isn't commonly used in ouroboros-consensus, but was causing `Ouroboros.Consensus.Mock.Node.Praos.praosBlockForging` to violate its NoThunks invariant during tests.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
@@ -168,6 +168,17 @@ isEmptyMVar StrictMVar { tmvar } = atomically $ Lazy.isEmptyTMVar tmvar
 
 updateMVar :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> (a, b)) -> m b
 updateMVar StrictMVar { tmvar, tvar, invariant } f = do
+    -- it's not unreasonable to assume that forcing !(!a', b) inside the
+    -- atomically block will force the new value before putting it into the
+    -- MVar, but there's actually an additional closure constructed (with a
+    -- dependency on a') that will only force a' when *it's* evaluated! in
+    -- order to ensure that we're forcing the value inside the MVar before
+    -- calling checkInvariant, we need an additional bang outside the
+    -- atomically block, which will correctly force a' before checkInvariant
+    -- looks to see if it's been evaluated or not. without this change, it's
+    -- possible to put a lazy value inside a StrictMVar (though it's unlikely
+    -- to occur in production environments because this intermediate unforced
+    -- closure is optimized away at -O1 and above)
     (!a', b) <- atomically $ do
         a <- Lazy.takeTMVar tmvar
         let !(!a', b) = f a
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
@@ -170,20 +170,24 @@ updateMVar :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> (a, b)) -> m
 updateMVar StrictMVar { tmvar, tvar, invariant } f = do
     -- it's not unreasonable to assume that forcing !(!a', b) inside the
     -- atomically block will force the new value before putting it into the
-    -- MVar, but there's actually an additional closure constructed (with a
-    -- dependency on a') that will only force a' when *it's* evaluated! in
-    -- order to ensure that we're forcing the value inside the MVar before
-    -- calling checkInvariant, we need an additional bang outside the
-    -- atomically block, which will correctly force a' before checkInvariant
-    -- looks to see if it's been evaluated or not. without this change, it's
-    -- possible to put a lazy value inside a StrictMVar (though it's unlikely
-    -- to occur in production environments because this intermediate unforced
-    -- closure is optimized away at -O1 and above)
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/test/consensus-test/Test/Consensus/Util/MonadSTM/NormalForm.hs
```diff
@@ -0,0 +1,58 @@
...
+import           GHC.Generics
+import           NoThunks.Class
+import           Ouroboros.Consensus.Util.MonadSTM.NormalForm (MonadSTM,
+                     newMVar, updateMVar)
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+
+-- Note that all of the tests here are only significant with compiler
+-- optimizations turned off!
```
jorisdral, line 18: I understand what you mean to say here, but maybe elaborate on why we are adding these tests. This will make it clear for future readers why running the tests without compiler optimisations turned off could be a problem

jorisdral (changes requested):

ouroboros-consensus/test/consensus-test/Test/Consensus/Util/MonadSTM/NormalForm.hs
fraser-iohk: I've added a larger description of what the tests do and why they aren't useful with compiler optimizations turned on 👍

jorisdral (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -168,7 +168,18 @@ isEmptyMVar StrictMVar { tmvar } = atomically $ Lazy.isEmptyTMVar tmvar
 
 updateMVar :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> (a, b)) -> m b
 updateMVar StrictMVar { tmvar, tvar, invariant } f = do
-    (a', b) <- atomically $ do
+    -- it's not unreasonable to assume that forcing !(!a', b) inside the
+    -- atomically block will force the new value before putting it into the
+    -- MVar, but there's actually an additional closure constructed (with a
+    -- dependency on a') that will only force a' when *it's* evaluated! in
```
jasagredo, line 174:
```suggestion
    -- MVar, but although the value in the tuple is forced, there's actually 
    -- a thin closure constructed that just points to the forced value and 
    -- that one is the one GHC returns in the constructed tuple (so it is 
    -- actually a thunk albeit a trivial one!). in
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -168,7 +168,18 @@ isEmptyMVar StrictMVar { tmvar } = atomically $ Lazy.isEmptyTMVar tmvar
...
+    -- calling checkInvariant, we need an additional bang outside the
+    -- atomically block, which will correctly force a' before checkInvariant
+    -- looks to see if it's been evaluated or not. without this change, it's
+    -- possible to put a lazy value inside a StrictMVar (though it's unlikely
+    -- to occur in production environments because this intermediate unforced
+    -- closure is optimized away at -O1 and above)
+    (!a', b) <- atomically $ do
         a <- Lazy.takeTMVar tmvar
         let !(!a', b) = f a
         Lazy.putTMVar tmvar a'
```
jasagredo, line 185: I would perhaps add a last comment mentioning:
```
-- To exactly see what we mean, compile this module with `-ddump-stg-final`
-- and look for the definition of the closure that is placed as the first
-- item in the returned tuple here
```

jasagredo (changes requested):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -168,7 +168,23 @@ isEmptyMVar StrictMVar { tmvar } = atomically $ Lazy.isEmptyTMVar tmvar
...
+    -- inside the MVar before calling checkInvariant, we need an additional
+    -- bang outside the atomically block, which will correctly force a' before
+    -- checkInvariant looks to see if it's been evaluated or not. without this
+    -- change, it's possible to put a lazy value inside a StrictMVar (though
+    -- it's unlikely to occur in production environments because this
+    -- intermediate unforced closure is optimized away at -O1 and above).
+    --
+    -- To exactly see what we mean, compile this module with `-ddump-stg-final`
+    -- and look for the definition of the closure that is placed as the first
+    -- item in the returned tuple here
```
jasagredo, line 186: As we say "the returned tuple here" I would put this next to 
```
return (a', b)
```
instead of here. Sorry that I didn't say it on my previous comment

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
fraser-iohk: ah! I read "the returned tuple here" to mean "the tuple returned from the `atomically` block", I'll change the comment slightly to "the tuple returned here" and put it before the last line of the `atomically`

jasagredo (approved):

PR #103, opened 2023-05-23
Golden tests: don't repeatedly test the same version combinations

amesgen:
With this, running e.g.
```console
cabal run shelley-test -- -p Golden
```
only reports 92 instead of 172 tests.

### DIFF
--- a/ouroboros-consensus/src/consensus-testlib/Test/Util/Serialisation/Golden.hs
+++ b/ouroboros-consensus/src/consensus-testlib/Test/Util/Serialisation/Golden.hs
@@ -445,8 +445,10 @@ goldenTest_SerialiseNodeToClient ::
   -> TestTree
 goldenTest_SerialiseNodeToClient codecConfig goldenDir Examples {..} =
     testGroup "SerialiseNodeToClient" [
-        testVersion (nodeToClientVersionToQueryVersion version, blockVersion)
-      | (version, blockVersion) <- nub $ Map.toList $ supportedNodeToClientVersions $ Proxy @blk
+        testVersion (queryVersion, blockVersion)
+      | (queryVersion, blockVersion) <-
+          nub . fmap (first nodeToClientVersionToQueryVersion) . Map.toList $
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/consensus-testlib/Test/Util/Serialisation/Golden.hs
```diff
@@ -445,8 +445,10 @@ goldenTest_SerialiseNodeToClient ::
   -> TestTree
 goldenTest_SerialiseNodeToClient codecConfig goldenDir Examples {..} =
     testGroup "SerialiseNodeToClient" [
-        testVersion (nodeToClientVersionToQueryVersion version, blockVersion)
-      | (version, blockVersion) <- nub $ Map.toList $ supportedNodeToClientVersions $ Proxy @blk
+        testVersion (queryVersion, blockVersion)
+      | (queryVersion, blockVersion) <-
+          nub . fmap (first nodeToClientVersionToQueryVersion) . Map.toList $
```
amesgen, line 450: `nub` is always suspicious, but this is only test code, and it would actually be the first place where we require `Ord` for `BlockNodeToClientVersion`.

dnadales (approved):

ouroboros-consensus/src/consensus-testlib/Test/Util/Serialisation/Golden.hs
fraser-iohk: `Data.List.nub` is on `Eq`, right? so it still won't require an `Ord` constraint on `BlockNodeToClientVersion`?

amesgen: Yes, I am just always strongly considering using e.g. [`nubOrd`](https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Containers-ListUtils.html#v:nubOrd) whenever I see `nub` :smile:

PR #104, opened 2023-05-23
Fix performance regression of `GetFilteredDelegationsAndRewardAccounts`

lehins:
# Description

Addition of deposits made `rewView` extremly expensive, because it now iterates the full staking credentials Map. The proper way to do it is to restrict the Map first and then do the conversion.

This will very likely fix the timeout issue in  https://github.com/CardanoSolutions/ogmios/issues/313


lehins:
>Can you add a [changelog fragment](https://github.com/input-output-hk/ouroboros-consensus/blob/main/docs/website/docs/ReleaseProcess.md#adding-a-changelog-fragment)?

@amesgen Done

PR #113, opened 2023-05-29
Remove the `ledger-db-backends-checker` executable.

jorisdral:
# Description

The executable is subsumed by the model-based tests for `BackingStore` implementations. These tests also give a strong indication that the in-memory and LMDB backing stores do not differ in their behaviour, since both implementations must match the same model.


dnadales (approved):
Thanks! :broom:

PR #115, opened 2023-05-29
Rename `StrictMVar` to `StrictSVar`.

jorisdral:
# Description

The `StrictMVar` identifier from `Ouroboros.Consensus.Util.MonadSTM.StrictMVar` clashes with `StrictMVar` from the new `strict-mvar` package. Furthermore, the former is not truly an `MVar`, since it does not offer the same fairness guarantees that an actual `MVar` does offer. For this reason, we rename the former `StrictMVar` to `StrictSVar`, which means "Strict Stale Var". "Stale" refers to the fact that we can read the last ("stale") value from the variable if it is currently empty.

NOTE: it is probably useful to review the diff per commit. The full diff of the PR shows that `StrictMVar` was deleted, and that `StrictSVar` is a completely new file, even though I used `git mv` to rename the file.

### DIFF
--- a/ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBAnalyser/Run.hs
+++ b/ouroboros-consensus-cardano/src/tools/Cardano/Tools/DBAnalyser/Run.hs
@@ -52,7 +52,7 @@ analyse ::
   -> IO (Maybe AnalysisResult)
 analyse DBAnalyserConfig{analysis, confLimit, dbDir, selectDB, validation, verbose} args =
     withRegistry $ \registry -> do
-      lock           <- newMVar ()
+      lock           <- newSVar ()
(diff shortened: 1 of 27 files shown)


jorisdral:
Current dependencies on/for this PR:
* main
  * **PR #115** <a href="https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/115" target="_blank">Graphite</a>  👈
    * **PR #116** <a href="https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/116" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/115?utm_source=stack-comment).

jasagredo (approved):

PR #116, opened 2023-05-29
Add `strict-mvar` dependency and `StrictMVar` with NoThunks invariants.

jorisdral:
# Description

We depend on an unreleased version of `strict-mvar` and we create a new module for `StrictMVar`s with `NoThunks` invariants. We also add a flag `checkmvarnothunks` that acts as a switch for whether we want to run the invariant checks or not. By default, these checks are off, since we do not want to perform these checks in production.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -29,6 +29,11 @@ flag expose-sublibs
   manual:      True
   default:     False
 
+flag checkmvarinvariants
+  description: Enable runtime invariant checks on StrictMVars
+  manual:      True
+  default:     False
+
 common common-lib
   default-language: Haskell2010
   ghc-options:
@@ -41,6 +46,9 @@ common common-lib
     ghc-options: -fno-ignore-asserts
     cpp-options: -DENABLE_ASSERTIONS
 
+  if flag(checkmvarinvariants)
+    cpp-options: -DCHECK_MVAR_INVARIANTS
+
 common common-test
   import:      common-lib
   ghc-options: -threaded -rtsopts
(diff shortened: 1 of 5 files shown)


jorisdral:
Current dependencies on/for this PR:
* main
  * **PR #115** <a href="https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/115" target="_blank">Graphite</a> 
    * **PR #116** <a href="https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/116" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/input-output-hk/ouroboros-consensus/116?utm_source=stack-comment).

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -41,6 +47,9 @@ common common-lib
     ghc-options: -fno-ignore-asserts
     cpp-options: -DENABLE_ASSERTIONS
 
+  if flag(checkmvarnothunks)
+    cpp-options: -DCHECK_MVAR_NOTHUNKS
```
jorisdral, line 51: Rename to `CHECK_MVAR_INVARIANTS`

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -49,6 +58,7 @@ library
   import:           common-lib
   hs-source-dirs:   src/ouroboros-consensus
   exposed-modules:
+    Control.Concurrent.Class.MonadMVar.Strict.NoThunks
```
jasagredo, line 61: I'm a bit unsure of adding this kind of "general" modules to Consensus, but I guess we are bridging NoThunks and strict-mvar here, so maybe there is not a better place

jasagredo (approved):
Approving with a comment

ouroboros-consensus/ouroboros-consensus.cabal
jorisdral: I agree, but I'm also unsure about adding another `Util` module. We can keep it here for now, and move it somewhere better if we find a better place

PR #120, opened 2023-05-30
Use `cryptoInit` everywhere

amesgen:
Follow-up to #119 that makes sure that we call `cryptoInit` before all test suites and executables.

As for why we didn't notice that this was missing for e.g. db-analyser so far: According to the cardano-base maintainers, not calling `cryptoInit` does at least not cause random segfaults, but may impact performance and cryptographic security.

### DIFF
--- a/ouroboros-consensus/src/consensus-testlib/Test/Util/TestEnv.hs
+++ b/ouroboros-consensus/src/consensus-testlib/Test/Util/TestEnv.hs
@@ -8,6 +8,7 @@ module Test.Util.TestEnv (
   , defaultTestEnvConfig
   ) where
 
+import           Cardano.Crypto.Init (cryptoInit)
 import           Data.Proxy (Proxy (..))
 import           Options.Applicative (metavar)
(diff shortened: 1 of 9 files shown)


ouroboros-consensus/src/consensus-testlib/Test/Util/TestEnv.hs
```diff
@@ -17,9 +18,10 @@ import           Test.Tasty.QuickCheck
 
 -- | 'defaultMain' extended with 'iohkTestEnvIngredient'
 defaultMainWithTestEnv :: TestEnvConfig -> TestTree -> IO ()
-defaultMainWithTestEnv testConfig testTree =
-    defaultMainWithIngredients (testEnvIngredient : defaultIngredients) $ withTestEnv testConfig testTree
-
+defaultMainWithTestEnv testConfig testTree = do
+    cryptoInit
```
dnadales, line 22: :100:

dnadales (approved):
Thanks!

PR #138, opened 2023-06-06
Add `splitAtSlot`

jasagredo:
# Description


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -33,17 +36,72 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
...
+       f (SeqDiffMK prunedSeq) (SeqDiffMK extendedSeq) =
+         SeqDiffMK $
+         if DS.minSlot prunedSeq == DS.minSlot extendedSeq
+         then extendedSeq
+         else
+           let (imm, _) = DS.splitAtSlot (fromWithOrigin 0 $ pointSlot $ getTip $ AS.anchor newVolatile) prunedSeq
+               (_, rest) = DS.splitAtFromEnd (AS.length newVolatile) extendedSeq
+           in DS.append imm rest
+     in
+       modifyTVar v (\pruned ->
```
dnadales, line 98: Perhaps we should give `v` a more descriptive name. 

Also, I wonder why we call the changelog inside `v` `pruned`. Is to denote that we expect this changelog to be pruned already?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -33,17 +36,72 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
...
+setCurrent v dblog = \case
+  Prune ->
+    updatePruning (changelogLastFlushedState dblog)
+  Extend ->
+    updateExtending (changelogVolatileStates dblog) (changelogDiffs dblog)
+ where
+   updatePruning :: ExtLedgerState blk EmptyMK -> STM m ()
+   updatePruning lf =
+     let
+       f :: (Ord k, Eq v) => SeqDiffMK k v -> SeqDiffMK k v
```
dnadales, line 67: I wonder if we could give `f` a more descriptive name or comment what's happening here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -33,17 +36,72 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
 import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog.Update
 import           Ouroboros.Consensus.Storage.LedgerDB.ReadsKeySets
 import           Ouroboros.Consensus.Util.IOLike
+import qualified Ouroboros.Network.AnchoredSeq as AS
+
+data LedgerDBUpdate = Prune | Extend
```
dnadales, line 41: Maybe this is related to [Joris' comment](https://github.com/input-output-hk/ouroboros-consensus/pull/138/files#r1219698948). I wonder if we should split `setCurrent` into two functions: one for extending and one for pruning.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -33,17 +36,72 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
...
+  -> LedgerDBUpdate
   -> STM m ()
-setCurrent = writeTVar
+setCurrent v dblog = \case
+  Prune ->
+    updatePruning (changelogLastFlushedState dblog)
+  Extend ->
+    updateExtending (changelogVolatileStates dblog) (changelogDiffs dblog)
+ where
+   updatePruning :: ExtLedgerState blk EmptyMK -> STM m ()
```
dnadales, line 64: Can two `updatePruning` operations race with each other? Are there any implicit assumptions here? For instance, I wonder if we are assuming that we prune only in one place (in the copy and snapshot thread).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -33,17 +36,72 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
...
+         if DS.minSlot prunedSeq == DS.minSlot extendedSeq
+         then extendedSeq
+         else
+           let (imm, _) = DS.splitAtSlot (fromWithOrigin 0 $ pointSlot $ getTip $ AS.anchor newVolatile) prunedSeq
+               (_, rest) = DS.splitAtFromEnd (AS.length newVolatile) extendedSeq
+           in DS.append imm rest
+     in
+       modifyTVar v (\pruned ->
+                       DbChangelog {
+                          changelogLastFlushedState = changelogLastFlushedState pruned
```
dnadales, line 100: Perhaps we should explain what happens if:

```
changelogLastFlushedState dblog /= changelogLastFlushedState pruned
```

I think that is the problem this PR is trying to solve right, races like:

```
(take cl; flush cl) || (take cl; modify cl; setCurrent cl)
```

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -1075,7 +1075,7 @@ runDB standalone@DB{..} cmd =
         (toFlush, bs) <- atomically $ do
           db <- readTVar dbChlog
           bs <- readTVar dbBackingStore
-          let (toFlush, db') = DbChangelog.splitForFlushing FlushAllImmutable db
```
dnadales, line 1078: How do we make sure now that we flush all the immutable DB here?

dnadales (commented):
If we're trying to prevent race conditions on code paths like the following:

```haskell
cl <- takeCL ldb
cl' <- mutate cl
setCL cl'
```

I wonder if we could implement some checks on the relation between `cl` and `cl'`. For instance, if we flush and then set the changelog, then the changelog we are setting should not have been flushed in the meantime.

PR #152, opened 2023-06-12
Port UTxO-HD docs from the `ouroboros-network` repository.

### LINKED ISSUE #151: Port UTxO-HD documents to the new feature branch
Documentation related to UTxO-HD currently still only lives on [the old branch in `ouroboros-network`](https://github.com/input-output-hk/ouroboros-network/pull/3863). We should port those documents to the feature branch in `ouroboros-consensus`.

jorisdral:
# Description

Resolves #151.

This PR ports the UTxO-HD docs from https://github.com/input-output-hk/ouroboros-network/pull/3863.


jasagredo (approved):

dnadales (changes requested):
We should probably take some time to discuss where we want this documentation to land. We currently have `report` and `website` inside `docs` and maybe we want to keep this structure. Also, I don't know if we want to expand the report or leave it as a historical reference. The tex files could be put into our `website` directory in the latter case.

jorisdral:
> We should probably take some time to discuss where we want this documentation to land. We currently have `report` and `website` inside `docs` and maybe we want to keep this structure. Also, I don't know if we want to expand the report or leave it as a historical reference. The tex files could be put into our `website` directory in the latter case.

We discussed this, and the consensus is that we want to move most of the (contents of) the docs to the website (See the `Structure` section of
https://github.com/input-output-hk/ouroboros-consensus/issues/126). I've done exactly that for a few files in this PR, but I would suggest we incrementally integrate the remaining files in the website in follow-up PRs. I've added a `NOTE.md` file to the `docs/utxo-hd` folder, which records that we are supposed to remove the folder at some point, i.e., when we have integrated everything we want to integrate in the documentation website.

dnadales (approved):

PR #158, opened 2023-06-15
Refactor ledger tables (data structure and combinators)

jorisdral:
# Description

Ledger tables are barbie-types (see `barbies` package), though unfortunately we can not implement classes like `FunctorB` for ledger tables because the class expects a type that is indexed over a _(uni-)functor_. Ledger tables are indexed over _bifunctors_ (mapkinds), so the kinds do not match.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables.hs
```diff
@@ -13,13 +15,23 @@
...
-  , MapKind
+    -- * Re-exports
+    module Ouroboros.Consensus.Ledger.Tables.Combinators
+  , module Ouroboros.Consensus.Ledger.Tables.Common
     -- * Basic LedgerState classes
   , CanStowLedgerTables (..)
   , HasLedgerTables (..)
-  , HasTickedLedgerTables (..)
+  , HasTickedLedgerTables
+    -- * TODO: To be removed
```
dnadales, line 25: What do we need to remove these?

dnadales (approved):
Nice work! 

Did we see any difference in the mempool micro-benchmarks?

jorisdral:
> Nice work!
> 
> Did we see any difference in the mempool micro-benchmarks?

I will run these tomorrow

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables.hs
jorisdral: Not much, but it mostly inflates the diff because identifiers have to be replaced. I can make a follow-up PR to remove these

jorisdral:
> > Nice work!
> > Did we see any difference in the mempool micro-benchmarks?
> 
> I will run these tomorrow

The mempool benchmarks seemed to be slightly faster on this branch compared to `feature/utxo-hd`. So, I'll go ahead and merge this once CI passes!

PR #222, opened 2023-07-11
ChainSync client: simplify a correctness argument

amesgen:
This is a single-commit PR; see its message.

Migrated from https://github.com/input-output-hk/ouroboros-network/pull/4313

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -562,7 +562,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg
           -- ('rollBackward'), so we have nothing to do.
           return $ Just kis
 
-        | Just intersection <- AF.intersectionPoint ourFrag' theirFrag ->
+        | Just (intersection, trimmedCandidateFrag) <- cross ourFrag' theirFrag
           -- Our current chain changed, but it still intersects with candidate
           -- fragment, so update the 'ourFrag' field and trim to the
           -- candidate fragment to the same anchor point.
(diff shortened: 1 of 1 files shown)


amesgen:
For direct reference, the comments from https://github.com/input-output-hk/ouroboros-network/pull/4313:

---

@nfrisby:

> Hmm. Actually: does "average-case = worst-case" apply here, meaning that we don't want to share the selection's prefix amongst all the ChainSync clients' theirFrag prefixes?

---

@amesgen

> I think this is a nice change, I could even see defining an abstract function like
> ```haskell
> cross ::
>      forall blk. HasHeader blk
>   => AnchoredFragment blk
>   -> AnchoredFragment blk
>   -> Maybe (Point blk, AnchoredFragment blk)
> cross c1 c2 = do
>     (p1, _p2, _s1, s2) <- AF.intersect c1 c2
>     let crossed = fromMaybe (error "impossible") $ AF.join p1 s2
>     pure (AF.anchorPoint s2, crossed)
> ```
> to make it more clear that we don't rely on any domain-specific properties of the ChainSync fragments here. Of course, we should keep the fairbairn threshold in mind.
> 
> W.r.t. performance, this new implementation should have the same worst-case performance; and I don't immediately see how it might have a better average-case performance (`splitAfterPoint` and `join` both perform sharing with previous fragments). In any case, it should be easy to write a small benchmark (and if we use the `cross` function above, it is easy to swap out the implementation if we intentionally want to make the average case slower again).

---

@nfrisby

> > `splitAfterPoint` and `join` both perform sharing with previous fragments
> 
> `cross` causes all clients to share the selection's prefix. `join` causes each client to reuse its own copy of that prefix. So `splitAfterPoint` causes the clients to share less, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -572,29 +572,15 @@ chainSyncClient mkPipelineDecision0 tracer cfg
...
-           --   then there would be no intersection anymore).
-           -- + This means the new anchor point must be between the old anchor
-           --   point and the new intersection point.
-           -- + Since we know both the old anchor point and the new
-           --   intersection point exist on their fragment, the new anchor
-           --   point must also.
-           Nothing -> error
-               "anchor point must be on candidate fragment if they intersect"
-           Just (_, trimmedCandidateFrag) -> return $ Just $
+          case ourPre `AF.join` theirSuf of
```
amesgen, line 575: We talked about this in a call and came to the conclusion that whatever improvement of the best/average case this is, it is likely irrelevant for the decision of a user to downgrade their machine specs.

Concretely, a Cardano header has a maximum size of `1.1 kB`, and `ourPre` contains at most `k = 2160` headers, so we get a maximum size decrease of `1.1 kB × 2160 × 100 = 237.6 MB` very conservatively assuming 100 ChainSync clients, which we think is dwarfed by the overall memory usage by the node.

nc6 (approved):
We reviewed this in a call, and agree that it makes sense, and that the increased sharing in a good case is unlikely to be significant, especially compared to the ledger state.

nfrisby (approved):
The local definition is a good balance 👍

PR #228, opened 2023-07-12
create-release.sh -> create-release.hs

fraser-iohk:
(mostly) a port of the `create-release.sh` script to haskell (using turtle)

how it handles version bumps (especially w.r.t. dependencies) could definitely use a look

### DIFF
--- a/scripts/release/create-release.hs
+++ b/scripts/release/create-release.hs
@@ -1,0 +1,265 @@
+#!/usr/bin/env cabal 
+{- project:
+  with-compiler: ghc-9.4
+-}
+{- cabal:
+  build-depends:
+    base,
+    commonmark,
+    containers,
+    filepath,
+    foldl,
+    text,
+    turtle ^>=1.6.0,
+-}
+{-# OPTIONS_GHC -Wall -Wextra #-}
+{-# LANGUAGE GHC2021 #-}
+{-# LANGUAGE BlockArguments, LambdaCase, OverloadedStrings, ScopedTypeVariables #-}
+
+import Commonmark
+import Control.Monad
+import Data.Map (Map)
+import Data.Monoid (First(..))
+import Data.Semigroup (Max(..))
+import Data.Version
+import System.FilePath
+import Turtle hiding (fp, d, o)
+import qualified Data.Map as Map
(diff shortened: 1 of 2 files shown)


scripts/release/create-release.hs
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env cabal 
+{- project:
+  with-compiler: ghc-9.4
```
jasagredo, line 3: I think (I at least) we use 8.10 officially, 9.2 in every day's work and we aim to support 9.6 in the future. Why did you choose 9.4?

scripts/release/create-release.hs
```diff
@@ -0,0 +1,265 @@
...
+    commonmark,
+    containers,
+    filepath,
+    foldl,
+    text,
+    turtle ^>=1.6.0,
+-}
+{-# OPTIONS_GHC -Wall -Wextra #-}
+{-# LANGUAGE GHC2021 #-}
+{-# LANGUAGE BlockArguments, LambdaCase, OverloadedStrings, ScopedTypeVariables #-}
```
jasagredo, line 17: I think this didn't pass through stylish Haskell. Perhaps we can format it?

scripts/release/create-release.hs
```diff
@@ -0,0 +1,265 @@
...
+        createGitCommit packageName next
+
+-- | Pairs of packages with a list of their dependencies. We use the
+--   dependencies to calculate which version changes should require a version
+--   bump in a package's dependencies as well.
+--
+--   BEWARE! This list should always be ordered such that any given package's
+--   dependencies are located BEFORE that package in the list!
+packages :: [(FilePath, [FilePath])]
+packages =
```
jasagredo, line 90: Can't this list be auto-generated? `stack dot` I think is able to generate a graph of deps so there must be some way to do it

scripts/release/create-release.hs
fraser-iohk: done! should I add this to the `check-stylish` script too

jasagredo: I would do so, yeah

jasagredo (approved):

PR #235, opened 2023-07-14
Test Cardano translations with an empty UTxO

### LINKED ISSUE #212: Test cardano translations with empty UTxO tables
For completeness, we should test translations with the empty UTxO. See `Test.Consensus.Cardano.Translation`

jorisdral:
# Description

Resolves #212

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Translation.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Translation.hs
@@ -86,11 +86,15 @@ tests = testGroup "UpdateTablesOnEraTransition"
   [ testTablesTranslation "Byron to Shelley"
                           byronToShelleyLedgerStateTranslation
                           byronUtxosAreInsertsInShelleyUtxoDiff
-                          (\st -> cover 50 (      nonEmptyUtxosByron st) "UTxO set is not empty"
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -87,8 +87,10 @@ tests = testGroup "UpdateTablesOnEraTransition"
                           byronToShelleyLedgerStateTranslation
                           byronUtxosAreInsertsInShelleyUtxoDiff
                           (\st -> cover 50 (      nonEmptyUtxosByron st) "UTxO set is not empty"
-                                -- TODO: #4473 we should test with empyt UTxO!
-                                -- . cover 1  (not $ nonEmptyUtxosByron st) "UTxO set is empty"
+                                . cover 1  (not $ nonEmptyUtxosByron st) "UTxO set is empty"
```
jorisdral, line 90: Adding this makes the tests very slow in some cases, because the probability of generating an empty UTxO is still relatively low. Do we still want to add this `cover`?

jorisdral: Seems to be solved by requiring `0.1` coverage instead of `1`.

PR #238, opened 2023-07-17
General ledger table classes for the `HardForkBlock`

jorisdral:
# Description

* General instances of ledger table classes (`HasLedgerTables`, `CanStowLedgerTables`, `LedgerTablesAreTrivial`) for a `HardForkBlock`.
* `HasTickedLedgerTables` is now a total class, and should mainly be used to partially apply the composition of `HasLedgerTables` and `Ticked1`. Such a composition may be necessary when using `sop-core` functions like `hcmap`, where one could now use `All (Compose HasTickedLedgerTables LedgerState) xs)`.


jasagredo (approved):
Looks perfect. I have no corrections. Thanks!!

PR #240, opened 2023-07-18
consensus-testlib: add chain generators

amesgen:
These generators are the first step towards a new set of tests, that will target the upcoming Ouroboros Genesis implementation.
One of the earlier milestones, however, will be some tests that apply even before Ouroboros Genesis.

The tip of the iceberg is the `Test.Ouroboros.Consensus.ChainGenerator.Tests` module. You can chase dependencies from there.

This PR was migrated from https://github.com/input-output-hk/ouroboros-network/pull/4438

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -286,6 +286,14 @@ library unstable-consensus-testlib
   visibility:      public
   hs-source-dirs:  src/unstable-consensus-testlib
   exposed-modules:
+    Test.Ouroboros.Consensus.ChainGenerator.Adversarial
+    Test.Ouroboros.Consensus.ChainGenerator.BitVector
+    Test.Ouroboros.Consensus.ChainGenerator.Counting
+    Test.Ouroboros.Consensus.ChainGenerator.Honest
+    Test.Ouroboros.Consensus.ChainGenerator.Params
+    Test.Ouroboros.Consensus.ChainGenerator.RaceIterator
+    Test.Ouroboros.Consensus.ChainGenerator.Slot
+    Test.Ouroboros.Consensus.ChainGenerator.Some
     Test.Util.BoolProps
     Test.Util.ChainDB
     Test.Util.ChainUpdates
@@ -362,6 +370,7 @@ library unstable-consensus-testlib
     , time
     , tree-diff
     , utf8-string
+    , vector
 
 library unstable-mock-block
   import:          common-lib
@@ -507,16 +516,24 @@ test-suite infra-test
   main-is:        Main.hs
   other-modules:
     Ouroboros.Consensus.Util.Tests
+    Test.Ouroboros.Consensus.ChainGenerator.Tests
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.Adversarial
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.BitVector
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.Counting
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.Honest
     Test.Util.ChainUpdates.Tests
     Test.Util.Schedule.Tests
     Test.Util.Split.Tests
 
   build-depends:
     , base
+    , mtl
     , ouroboros-consensus:{ouroboros-consensus, unstable-consensus-testlib}
     , QuickCheck
+    , random
     , tasty
     , tasty-quickcheck
+    , vector
 
 test-suite storage-test
   import:         common-test
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -1,0 +1,572 @@
+{-# LANGUAGE BangPatterns              #-}
+{-# LANGUAGE DataKinds                 #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts          #-}
+{-# LANGUAGE MultiWayIf                #-}
+{-# LANGUAGE NamedFieldPuns            #-}
+{-# LANGUAGE PolyKinds                 #-}
+{-# LANGUAGE ScopedTypeVariables       #-}
+{-# LANGUAGE TypeApplications          #-}
+
+-- TODO rename to .Alternative?
+module Test.Ouroboros.Consensus.ChainGenerator.Adversarial (
+    -- * Generating
+    AdversarialRecipe (AdversarialRecipe, arHonest, arParams, arPrefix)
+  , CheckedAdversarialRecipe (UnsafeCheckedAdversarialRecipe, carHonest, carParams, carWin)
+  , NoSuchAdversarialChainSchema (NoSuchAdversarialBlock, NoSuchCompetitor, NoSuchIntersection)
+  , SomeCheckedAdversarialRecipe (SomeCheckedAdversarialRecipe)
+  , checkAdversarialRecipe
+  , uniformAdversarialChain
+    -- * Testing
+  , AdversarialViolation (BadAnchor, BadCount, BadRace)
+  , AnchorViolation (HonestActiveMustAnchorAdversarial)
+  , ChainSchema (ChainSchema)
+  , RaceViolation (AdversaryWonRace, rvAdv, rvHon)
+  , checkAdversarialChain
+  ) where
+
+import           Control.Applicative ((<|>))
+import           Control.Monad (void, when)
+import qualified Control.Monad.Except as Exn
+import           Data.Proxy (Proxy (Proxy))
+import qualified System.Random.Stateful as R
+import qualified Test.Ouroboros.Consensus.ChainGenerator.BitVector as BV
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Counting as C
+import           Test.Ouroboros.Consensus.ChainGenerator.Honest
+                     (ChainSchema (ChainSchema))
+import           Test.Ouroboros.Consensus.ChainGenerator.Params (Asc,
+                     Delta (Delta), Kcp (Kcp), Scg (Scg))
+import qualified Test.Ouroboros.Consensus.ChainGenerator.RaceIterator as RI
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Slot as S
+import           Test.Ouroboros.Consensus.ChainGenerator.Slot
+                     (E (ActiveSlotE, EmptySlotE, SlotE))
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Some as Some
+
+-----
+
+data AnchorViolation =
+    -- | An honest active slot must immediately precede the adversarial interval
+    HonestActiveMustAnchorAdversarial
+  |
+    -- | The were not exactly 'arPrefix' many active slots preceding @adv@
+    WrongNumberOfHonestPredecessors
+  deriving (Eq, Read, Show)
+
+-- | A violation of Races Before Acceleration Assumption
+--
+-- INVARIANT: @'C.windowLast' 'rvAdv' < 'C.windowLast' 'rvHon' + 'Delta'@
+--
+-- INVARIANT: @'C.windowStart' 'rvHon' <= 'C.windowStart' 'rvAdv'@
+data RaceViolation hon adv = AdversaryWonRace {
+    -- | The adversarial race window
+    rvAdv :: !(RI.Race adv)
+  ,
+    -- | The honest race window
+    rvHon :: !(RI.Race hon)
+  }
+  deriving (Eq, Read, Show)
+
+data AdversarialViolation hon adv =
+    BadAnchor !AnchorViolation
+  |
+    -- | The schema does not contain a positive number of active slots
+    BadCount
+  |
+    BadRace !(RaceViolation hon adv)
+  deriving (Eq, Read, Show)
+
+-- | Check the chain matches the given 'AdversarialRecipe'.
+--
+-- * It must intersect the honest chain at an active slot
+-- * It must satisfy the Races Before Acceleration Assumption
+--   (which implies the Length of Competing Chains Assumption)
+--
+-- Definition of a /Praos Race Window/ of a chain. It is an interval of slots
+-- that contains at least @k+1@ blocks of the chain and exactly 'Delta' slots
+-- after the @k+1@st block.
+--
+-- Definition of the /Length of Competing Chains Assumption/. We assume every adversarial
+-- chain contains at most @k@ blocks in the Praos Race Window anchored at the
+-- the intersection.
+--
+-- Definition of /Acceleration Lower Bound of an Adversarial Chain/. It is
+-- the lowest slot at which an adversary could speed up its elections on a
+-- given adversarial chain. It is defined as the 'Scg'-th slot after the first
+-- adversarial block or the 'Delta'-th slot after the @k+1@st honest block after the
+-- intersection, whichever is greater.
+--
+-- Definition of the /Races Before Acceleration Assumption/. Every adversarial
+-- chain has at most @k@ blocks in every Praos Race Window of the honest chain
+-- that starts after the intersection and ends before the acceleration lower
+-- bound.
+--
+-- Definition of the /Genesis Implication Conjecture/. We conjecture that
+-- assuming that the Genesis window length is no greater than the Stability
+-- Window and that every possible adversarial chain satisfies the Races
+-- in Stability Windows Assumption, then the honest chain strictly wins every possible density
+-- comparison from the Ouroboros Genesis paper. The key intuition is that a less
+-- dense chain would have to lose at least one Praos race within the Genesis window.
+checkAdversarialChain ::
+  forall base hon adv.
+     AdversarialRecipe base hon
+  -> ChainSchema base adv
+  -> Exn.Except (AdversarialViolation hon adv) ()
+checkAdversarialChain recipe adv = do
+    checkStart
+    checkCount
+    checkRaces
+  where
+    AdversarialRecipe {
+        arHonest = ChainSchema winH vH
+      ,
+        arParams = (Kcp k, Scg s, Delta d)
+      ,
+        arPrefix
+      } = recipe
+
+    ChainSchema winA vA = adv
+
+    checkStart = do
+        let startA       = C.windowStart winA :: C.Index base SlotE
+            intersection = startA C.- 1       :: C.Index base SlotE
+
+        -- The intersection must be at an active slot in the honest chain.
+        case C.toWindow winH intersection of
+            Nothing -> do
+              -- the genesis block is the only permissible anchor outside of @hon@
+              when (startA   /= C.Count 0) $ Exn.throwError $ BadAnchor HonestActiveMustAnchorAdversarial
+              when (arPrefix /= C.Count 0) $ Exn.throwError $ BadAnchor WrongNumberOfHonestPredecessors
+
+            Just i -> do
+                let _ = i :: C.Index hon SlotE
+
+                when (BV.testV S.inverted vH i) $ do
+                    Exn.throwError $ BadAnchor HonestActiveMustAnchorAdversarial
+
+                C.SomeWindow Proxy precedingSlots <-
+                    pure $ C.withWindowBetween (C.windowSize winH) (C.Lbl @"foo") (C.Count 0) i
+                let pc = BV.countActivesInV S.notInverted (C.sliceV precedingSlots vH)
+
+                -- arPrefix must correctly count the active slots in the part of
+                -- the chain upto the intersection
+                when (C.fromWindowVar precedingSlots (C.toVar pc) /= arPrefix) $ do
+                    Exn.throwError $ BadAnchor WrongNumberOfHonestPredecessors
+
+    checkCount = do
+        let pc = BV.countActivesInV S.notInverted vA
+        when (C.toVar pc <= 0) $ Exn.throwError BadCount
+
+    -- the youngest slot in which the adversarial schedule cannot have accelerated
+    --
+    -- (IE @s@ past the first active adversarial slot, or @d@ past the @k+1@st
+    --  slot after the intersection)
+    youngestStableA :: C.Index base SlotE
+    youngestStableA =
+        let sYoungest = case BV.findIthEmptyInV S.inverted vA (C.Count 0) of
+              BV.JustFound firstActiveA ->
+                -- if s=0, then the slot of their block is the youngest stable slot
+                C.fromWindow winA $ firstActiveA C.+ s
+              BV.NothingFound           ->
+                -- the rest of the function won't force this since there are no
+                -- adversarial active slots
+                error "dead code"
+            kPlus1stYoungest = case BV.findIthEmptyInV S.inverted vH (C.toIndex arPrefix C.+ k) of
+              BV.JustFound kPlus1st -> C.fromWindow winH kPlus1st
+              BV.NothingFound       ->
+                -- If the honest fork didn't reach @k+1@ before it ended, then
+                -- the conservative assumption that all slots after 'Len' are
+                -- active on the honest chain implies the @k+1@st block is in
+                -- slot @C.windowLast winH + (k+1-x)@, where the honest chain
+                -- has @x@ remaining blocks such that @x<k+1@.
+                let honestBlocksAfterIntersection =
+                      C.getCount (BV.countActivesInV S.notInverted vH) - C.getCount arPrefix
+                in
+                    C.windowLast winH
+                  C.+
+                    (k + 1 - honestBlocksAfterIntersection)
+         in
+            max sYoungest (kPlus1stYoungest C.+ d)
+
+    checkRaces = do
+        let iterH =
+                maybe (RI.initConservative vH) id
+              $ RI.init (Kcp k) vH
+        case RI.init (Kcp (k - 1)) vA >>= RI.next vA of
+            Nothing    -> pure ()   -- there are <= k total adversarial active slots
+            Just iterA ->
+                -- TODO optimization: how to skip all the honest Race Windows that
+                -- don't reach beyond the intersection? Perhaps skip to i - s + d?
+                go iterH iterA
+
+    -- INVARIANT iterH spans k+1 active slots (possibly conservatively)
+    --
+    -- INVARIANT iterA spans k active slots (actually, not conservatively)
+    go !iterH !iterA = do
+        C.SomeWindow Proxy raceWinH <- pure $ let RI.Race x = iterH in x
+        C.SomeWindow Proxy raceWinA <- pure $ let RI.Race x = iterA in x
+
+        -- lift both windows to @base@ so that they're comparable
+        let raceWinH' = C.joinWin winH raceWinH
+            raceWinA' = C.joinWin winA raceWinA
+
+        if
+
+          -- any Race Window that ends /after/ the adversary can accelerate is unconstrained
+          | youngestStableA < C.windowLast raceWinH' C.+ d -> pure ()
+
+          -- advance the adversarial Race Window if its start is <= the honest Race Window's start
+          | C.windowStart raceWinA' <= C.windowStart raceWinH' ->
+            case RI.next vA iterA of
+                Just iterA' -> go iterH iterA'
+                Nothing     -> pure ()   -- there are < k remaining adversarial active slots
+
+          -- fail if the adversary won or tied the race
+          | C.windowLast raceWinA' <= C.windowLast raceWinH' C.+ d ->
+                -- iterA contains exactly k active slots, but A) it's anchored
+                -- in an active slot and B) iterH contains that anchor. Thus
+                -- adv has k+1 in iterH.
+                Exn.throwError $ BadRace AdversaryWonRace {
+                    rvAdv = iterA
+                  ,
+                    rvHon = iterH
+                  }
+
+          -- advance the honest Race Window
+          | otherwise -> case RI.next vH iterH <|> RI.nextConservative vH iterH of
+                Just iterH' -> go iterH' iterA
+                Nothing     -> pure ()   -- there are no remaining honest active slots
+                                         --
+                                         -- TODO hpc shows this never executes
+
+-----
+
+-- | Named arguments for 'checkAdversarialRecipe'
+data AdversarialRecipe base hon =
+    AdversarialRecipe {
+        -- | The honest chain to branch off of
+        arHonest :: !(ChainSchema base hon)
+      ,
+        -- | protocol parameters
+        arParams :: (Kcp, Scg, Delta)
+      ,
+        -- | Where to branch off of 'arHonest'
+        --
+        -- It is the amount of blocks shared by the honest and the adversarial
+        -- chain. In other words, the 0-based index of their intersection
+        -- in blocks, such that
+        --
+        -- * @0@ identifies the genesis block
+        -- * @1@ identifies the first block in arHonest
+        -- * @2@ identifies the second block in arHonest
+        -- * etc
+        arPrefix :: !(C.Var hon ActiveSlotE)
+      }
+  deriving (Eq, Read, Show)
+
+-- | See 'CheckedAdversarialRecipe'
+data SomeCheckedAdversarialRecipe base hon =
+    forall adv.
+    SomeCheckedAdversarialRecipe
+        !(Proxy adv)
+        !(CheckedAdversarialRecipe base hon adv)
+
+instance Show (SomeCheckedAdversarialRecipe base hon) where
+    showsPrec p (SomeCheckedAdversarialRecipe adv car) =
+        Some.runShowsPrec p
+      $ Some.showCtor SomeCheckedAdversarialRecipe "SomeCheckedAdversarialRecipe"
+            `Some.showArg` adv
+            `Some.showArg` car
+
+instance Read (SomeCheckedAdversarialRecipe base hon) where
+    readPrec =
+        Some.runReadPrec
+      $ Some.readCtor SomeCheckedAdversarialRecipe "SomeCheckedAdversarialRecipe"
+            <*> Some.readArg
+            <*> Some.readArg
+
+-- | Image of 'checkAdversarialRecipe' when it accepts the recipe
+data CheckedAdversarialRecipe base hon adv =
+    UnsafeCheckedAdversarialRecipe {
+        -- | The honest chain to branch off of
+        carHonest :: !(ChainSchema base hon)
+      ,
+        -- | protocol parameters
+        carParams :: (Kcp, Scg, Delta)
+      ,
+        -- | The window starting at the first slot after the intersection
+        -- and ending at the last slot of 'carHonest'.
+        --
+        -- INVARIANT: there is at least one active honest slot in @adv@
+        --
+        -- In other words, the adversarial leader schedule does not /extend/ the
+        -- chain represented by 'carHonest', it competes with it.
+        carWin    :: !(C.Contains SlotE hon adv)
+      }
+  deriving (Eq, Read, Show)
+
+-- | Image of 'checkAdversarialRecipe' when it rejects the recipe
+data NoSuchAdversarialChainSchema =
+    -- | There is no slot the adversary can lead
+    --
+    -- Two possible reasons, where @X@ is the slot of 'arPrefix' and Y is the youngest slot of 'arHonest'.
+    --
+    --    * The interval @[X, Y)@ is empty.
+    --
+    --    * @k=0@
+    --
+    -- Suppose k=0 and slot x and slot y are two honest active slots with only
+    -- honest empty slots between them.
+    --
+    -- Length of Competing Chains prohibits the adversary from leading in the interval (x,y].
+    --
+    -- In fact, by induction, the adversary can never lead: suppose the same y
+    -- and a slot z are two honest active slots with only honest empty slots
+    -- between them...
+    NoSuchAdversarialBlock
+  |
+    -- | @not ('arPrefix' < C)@ where @C@ is the number of active slots in 'arHonest'
+    NoSuchCompetitor
+  |
+    -- | @not (0 <= 'arPrefix' <= C)@ where @C@ is the number of active slots in 'arHonest'
+    NoSuchIntersection
+  deriving (Eq, Show)
+
+-----
+
+-- | The suffix of the slots starting strictly /after/ the intersection
+data AdvLbl
+
+-- | The interval of slots that have most recently attained their final value
+data SettledLbl
+
+-- | Reject a bad 'AdversarialRecipe'
+checkAdversarialRecipe ::
+  forall base hon.
+     AdversarialRecipe base hon
+  -> Exn.Except
+         NoSuchAdversarialChainSchema
+         (SomeCheckedAdversarialRecipe base hon)
+checkAdversarialRecipe recipe = do
+    when (0 == k) $ Exn.throwError NoSuchAdversarialBlock
+
+    -- validate 'arPrefix'
+    firstAdvSlot <- case compare arPrefix 0 of
+        LT -> Exn.throwError NoSuchIntersection
+        EQ -> pure $ C.Count 0
+        GT -> case BV.findIthEmptyInV S.inverted vH $ C.toIndex $ arPrefix - 1 of
+            BV.NothingFound -> Exn.throwError NoSuchIntersection
+            BV.JustFound x  -> do
+                when (x == C.lastIndex (C.windowSize winH)) $ Exn.throwError NoSuchAdversarialBlock
+                pure (x C.+ 1)
+
+    C.SomeWindow Proxy winA <- pure $ C.withSuffixWindow (C.windowSize winH) (C.Lbl @AdvLbl) firstAdvSlot
+
+    -- there must be at least one honest active slot in @adv@
+    case BV.findIthEmptyInV S.inverted (C.sliceV winA vH) (C.Count 0) of
+        BV.NothingFound -> Exn.throwError NoSuchCompetitor
+        BV.JustFound{}  -> pure ()
+
+    pure $ SomeCheckedAdversarialRecipe Proxy $ UnsafeCheckedAdversarialRecipe {
+        carHonest = arHonest
+      ,
+        carParams = arParams
+      ,
+        carWin    = winA
+      }
+  where
+    AdversarialRecipe {
+        arHonest
+      ,
+        arParams
+      ,
+        arPrefix
+      } = recipe
+
+    (Kcp k, _scg, _delta) = arParams
+
+    ChainSchema winH vH = arHonest
+
+-----
+
+data RaceAssumptionLbl
+data UntouchableLbl
+data TouchableLbl
+
+-- | Generate an adversarial 'ChainSchema' that satifies 'checkExtendedRaceAssumption'
+--
+-- The distribution this function samples from is not simple to describe. It
+-- begins by drawing a sample of length 'adv' from the Bernoulli process
+-- induced by the active slot coefficient 'Asc'. (IE 'adv' many i.i.d. samples
+-- from @Uniform(f)@). Then it visits the Extended Praos Race Windows that
+-- overlap with the prefix of that leader schedule that ends one stability
+-- window after the first (remaining) active adversarial slot in it. In each
+-- such Window, it removes the youngest active slots from the adversarial
+-- leader schedule until it loses that Race.
+uniformAdversarialChain ::
+  forall g base hon adv.
+     R.RandomGen g
+  => Maybe Asc   -- ^ 'Nothing' means @1@
+  -> CheckedAdversarialRecipe base hon adv
+  -> g
+  -> ChainSchema base adv
+{-# INLINABLE uniformAdversarialChain #-}
+uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
+    g <- R.newSTGenM g0
+
+    let sz = C.windowSize carWin :: C.Size adv SlotE
+
+    -- randomly initialize the bitstring
+    mv <- C.replicateMV sz $ case mbAsc of
+        Nothing  -> pure $ S.mkActive S.notInverted
+        Just asc -> S.genS asc `R.applySTGen` g
+
+    -- ensure the adversarial leader schedule is not empty
+    do  void $ BV.fillInWindow
+            S.notInverted
+            (BV.SomeDensityWindow (C.Count 1) (C.windowSize carWin))
+            g
+            mv
+
+    -- find the slot of the k+1 honest block
+    let kPlus1st :: C.Index adv SlotE
+        kPlus1st = case BV.findIthEmptyInV S.inverted (C.sliceV carWin vH) (C.Count k) of
+          BV.NothingFound -> maybe (error "dead code") id $ C.toWindow carWin $ C.windowLast carWin
+          BV.JustFound x  -> x
+
+
+    -- ensure the adversarial leader schedule does not win any of the races for
+    -- which the honest Race Window fits within the Stability Window anchored at
+    -- the first adversarial active slot
+    --
+    -- TODO Why is it ok to skip early honest races, some of which overlap with
+    -- @adv@? Is it because not having >k in some prefix [0, n] of adv ensures
+    -- you can't have >k in the interval [0, C.frWindow adv n] either?
+    let iterH :: RI.Race adv
+        iterH =
+            maybe (RI.initConservative vA) id
+          $ RI.init kcp vA
+    unfillRaces kPlus1st (C.Count 0) UnknownYS iterH g mv
+
+    pure mv
+  where
+    UnsafeCheckedAdversarialRecipe {
+        carHonest
+      ,
+        carParams = (kcp, scg, delta)
+      ,
+        carWin
+      } = recipe
+
+    wrap v = ChainSchema (C.joinWin winH carWin) v
+
+    Kcp   k = kcp
+    Scg   s = scg
+    Delta d = delta
+
+    ChainSchema winH vH = carHonest
+
+    vA = C.sliceV carWin vH
+
+    -- ensure the adversary loses this 'RI.Race' and each subsequent race that ends before it can accelerate
+    unfillRaces kPlus1st !scope !mbYS !iter !g !mv = when (withinYS delta mbYS iter) $ do
+        C.SomeWindow Proxy rwin <- pure $ let RI.Race x = iter in x
+
+        C.SomeWindow (Proxy :: Proxy skolem) win <-
+            pure
+          $ C.withWindowBetween
+                (C.windowSize carWin)
+                (C.Lbl @RaceAssumptionLbl)
+                (C.windowStart rwin)
+                (C.windowLast  rwin C.+ d)   -- rwin ends in a block, so if d=0
+                                             -- then the slot after that block
+                                             -- is unconstrained; hence no +1
+
+        -- INVARIANT: @win@ contains @scope@
+        let _ = scope :: C.Index adv SlotE
+
+        -- remove adversarial active slots as needed
+        --
+        -- But only remove them from /after/ @scope@ (ie do not remove them
+        -- from slots in the previous Race Window).
+        do  untouchZeroCount :: C.Var adv EmptySlotE <- do
+                C.SomeWindow Proxy untouch <-   -- untouchable slots
+                    pure
+                  $ C.withWindowBetween
+                        (C.windowSize carWin)
+                        (C.Lbl @UntouchableLbl)
+                        (C.windowStart rwin)
+                        (scope C.- 1)   -- window will be empty if scope is 0
+                C.fromWindowVar untouch <$> BV.countActivesInMV S.inverted (C.sliceMV untouch mv)
+
+            C.SomeWindow (Proxy :: Proxy skolem2) touch <-   -- touchable slots
+                pure
+              $ C.withWindowBetween
+                    (C.windowSize carWin)
+                    (C.Lbl @TouchableLbl)
+                    scope
+                    (C.windowLast rwin C.+ d)
+
+
+            let
+              maxActive :: C.Var (C.Win RaceAssumptionLbl skolem) ActiveSlotE
+              maxActive = C.Count k
+
+              -- at most k can be active in this race window, so at least size - k must be empty
+              minEmpty :: C.Var (C.Win RaceAssumptionLbl skolem) EmptySlotE
+              minEmpty = S.complementActive S.notInverted (C.windowSize win) maxActive
+
+            -- Discount that basic requirement by the number of zeros already
+            -- in the untouchable portion of this race window.
+            let touchableEmpty =
+                    max 0
+                  $ C.fromWindowVar win minEmpty - untouchZeroCount
+                  :: C.Var adv EmptySlotE
+
+            void $ BV.fillInWindow
+                S.inverted
+                (BV.SomeDensityWindow (C.toWindowVar touch touchableEmpty) (C.windowSize touch))
+                g
+                (C.sliceMV touch mv)
+
+        case RI.next vA iter <|> RI.nextConservative vA iter of
+            Nothing -> pure ()   -- there are no remaining honest active slots
+
+            Just iter' -> do
+                C.SomeWindow Proxy rwin' <- pure $ let RI.Race x = iter' in x
+                mbYS' <- case mbYS of
+                    KnownYS{} -> pure mbYS
+                    UnknownYS -> do
+                        -- check whether the slots that are settled as of just
+                        -- now contain the first adversarial active slot
+                        C.SomeWindow Proxy settledSlots <-
+                            pure
+                          $ C.withWindowBetween
+                                (C.windowSize carWin)
+                                (C.Lbl @SettledLbl)
+                                (C.windowStart rwin)
+                                (C.windowStart rwin' C.- 1)
+                        mbFound <- BV.findIthEmptyInMV S.inverted (C.sliceMV settledSlots mv) (C.Count 0)
+                        case mbFound of
+                            BV.NothingFound -> pure UnknownYS
+                            BV.JustFound x  ->
+                                -- x is the first settled adversarial slot, so
+                                -- the adversary can accelerate its growth as
+                                -- of x+s+1 (If s were 0, it could accelerate
+                                -- in the very next slot, thus the plus 1.)
+                                pure $! KnownYS $!
+                                  max
+                                    (kPlus1st C.+ d C.+ 1)
+                                    (C.fromWindow settledSlots x C.+ s C.+ 1)
+                unfillRaces kPlus1st (C.windowLast win C.+ 1) mbYS' iter' g mv
+
+-- | The youngest stable slot
+--
+-- If @x@ is the first adversarial active slot, then @x+s@ (see 'Scg') is the youngest stable slot.
+data MaybeYS base = UnknownYS | KnownYS !(C.Index base SlotE)
+  deriving (Eq, Read, Show)
+
+-- | Does the Race Window end in a stable slot?
+withinYS :: Delta -> MaybeYS base -> RI.Race base -> Bool
+withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
+    KnownYS ys -> C.windowLast win C.+ d < ys
+    UnknownYS  -> True   -- Honest Chain Growth ensures every Race Window is at most @'Scg' - 'Delta'@ slots wide
--- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
+++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
@@ -1,0 +1,477 @@
+{-# LANGUAGE DataKinds                 #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE LambdaCase                #-}
+{-# LANGUAGE NamedFieldPuns            #-}
+{-# LANGUAGE TypeApplications          #-}
+
+module Test.Ouroboros.Consensus.ChainGenerator.Tests.Adversarial (tests) where
+
+import           Control.Applicative ((<|>))
+import qualified Control.Monad.Except as Exn
+import           Data.Functor ((<&>))
+import           Data.Functor.Identity (runIdentity)
+import           Data.IORef (modifyIORef', newIORef, readIORef, writeIORef)
+import           Data.Proxy (Proxy (Proxy))
+import           Data.Word (Word8)
+import qualified System.Random as R
+import qualified System.Random.Stateful as R
+import qualified System.Timeout as IO (timeout)
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Adversarial as A
+import qualified Test.Ouroboros.Consensus.ChainGenerator.BitVector as BV
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Counting as C
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Honest as H
+import           Test.Ouroboros.Consensus.ChainGenerator.Params (Asc,
+                     Delta (Delta), Kcp (Kcp), Len (Len), Scg (Scg),
+                     ascFromBits)
+import qualified Test.Ouroboros.Consensus.ChainGenerator.RaceIterator as RI
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Slot as S
+import           Test.Ouroboros.Consensus.ChainGenerator.Slot
+                     (E (ActiveSlotE, SlotE))
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Some as Some
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Tests.Honest as H
+import qualified Test.QuickCheck as QC
+import           Test.QuickCheck.Random (QCGen)
+import qualified Test.Tasty as TT
+import qualified Test.Tasty.QuickCheck as TT
+
+-----
+
+tests :: [TT.TestTree]
+tests = [
+    TT.testProperty "prop_adversarialChain" prop_adversarialChain
+  ,
+    TT.localOption (TT.QuickCheckMaxSize 14) $ TT.testProperty "prop_adversarialChainMutation" prop_adversarialChainMutation
+  ]
+
+-----
+
+data SomeTestAdversarial =
+    forall base hon.
+    SomeTestAdversarial
+        !(Proxy base)
+        !(Proxy hon)
+        !(TestAdversarial base hon)
+
+instance Show SomeTestAdversarial where
+    showsPrec p (SomeTestAdversarial base hon test) =
+        Some.runShowsPrec p
+      $ Some.showCtor SomeTestAdversarial "SomeTestAdversarial"
+            `Some.showArg` base
+            `Some.showArg` hon
+            `Some.showArg` test
+
+instance Read SomeTestAdversarial where
+    readPrec =
+        Some.runReadPrec
+      $ Some.readCtor SomeTestAdversarial "SomeTestAdversarial"
+            <*> Some.readArg
+            <*> Some.readArg
+            <*> Some.readArg
+
+data TestAdversarial base hon = TestAdversarial {
+    testAscA     :: !Asc
+  ,
+    testAscH     :: !Asc
+  ,
+    testRecipeA  :: !(A.AdversarialRecipe base hon)
+  ,
+    testRecipeA' :: !(A.SomeCheckedAdversarialRecipe base hon)
+  ,
+    testRecipeH  :: !H.HonestRecipe
+  ,
+    testRecipeH' :: !(H.CheckedHonestRecipe base hon)
+  ,
+    testSeedH    :: !QCGen
+  }
+  deriving (Read, Show)
+
+genArPrefix :: H.ChainSchema base hon -> QC.Gen (C.Var hon ActiveSlotE)
+genArPrefix schedH =
+    if C.toVar numChoices < 2 then pure (C.Count 0) {- can always pick genesis -} else do
+        g <- QC.arbitrary
+        pure $ C.toVar $ R.runSTGen_ (g :: QCGen) $ C.uniformIndex numChoices
+  where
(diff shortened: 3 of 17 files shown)


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -544,16 +554,25 @@ test-suite infra-test
...
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.BitVector
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.Counting
+    Test.Ouroboros.Consensus.ChainGenerator.Tests.Honest
     Test.Util.ChainUpdates.Tests
     Test.Util.Schedule.Tests
     Test.Util.Split.Tests
 
   build-depends:
     , base
+    , hmatrix
```
nfrisby, line 568: `hmatrix` is the oddest of the new dependencies. It's used in `Test.Ouroboros.Consensus.ChainGenerator.Honest.genFirstWindowPopCount`. There are many large comments nearby that motivate the question it's used to solve.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+--
+-- Definition of /window/ and /anchored window/. A window is a contiguous
+-- sequence of slots. A window anchored at a block starts with the slot
+-- immediately after that block.
+--
+-- Definition of /Praos Chain Growth Assumption/. We assume the honest chain
+-- contains at least @k@ blocks in every window that contains @s@ slots.
+--
+-- Definition of /Praos Enriched Chain Growth Assumption/. We assume the honest
+-- chain contains at least @k+1@ blocks in every window that contains @s - Δ@
```
nfrisby, line 605: After discussing it with Facundo, I think the subtraction by delta in the denominator is likely unnecessary. It's only necessary if we assume the adversary can (significantly) accelerate immediately after the stability window.

The adversarial chain generator et al does allow for that to happen. But it's unlikely to be true on any real chain.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Some.hs
```diff
@@ -0,0 +1,131 @@
+{-# LANGUAGE DataKinds                  #-}
```
nfrisby, line 1: This module is a utility, much more general than `ChainGenerators`, but relied upon heavily by it.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
```diff
@@ -0,0 +1,341 @@
+{-# LANGUAGE ExistentialQuantification  #-}
```
nfrisby, line 1: This module is a utility, much more general than `ChainGenerators`, but relied upon heavily by it.

facundominguez (commented):
I'm sharing some comments. I have plenty to read yet.

I started looking at the honest tests and dived into some of the data type definitions.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+-----
+
+-- | The leader schedule of an honest /chain/
+--
+-- The represented chain grows by one block per active slot. A different data
+-- type would represent a leader schedule in which leaders to not extend a block
+-- from the previous active slot.
+--
+-- INVARIANT: at least one active slot
+data ChainSchedule base inner =
```
facundominguez, line 126: Maybe `ChainSchedule` needs to be renamed to `HonestChainSchema` to follow the test plan terminology.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,144 @@
...
+
+-- | Many functions in this library are parameterized over a type of kind @Pol@
+--
+-- If the type is 'Inverted', then the function will treat all 'S' values as if they were first complemented.
+--
+-- The 'PreImage' family does the corresponding parameterization of 'ActiveSlotE' and 'EmptySlotE' at the type level.
+--
+-- NOTE: No 'S' value is ever actually complemented because of 'Inverted', but
+-- the functions parameterized by 'POL' will treat them as if they were.
+data Pol = Inverted | NotInverted
```
facundominguez, line 107: I still fail to see the motivation to parameterize types with polarity. A pointer, or a description would be helpful.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+each window is only missing one active slot (eg need 4 actives per 5 slots),
+then toggling B would require three toggles in total, whereas toggling A and C
+would solve the problem with just two toggles.
+
+-}
+
+-- | A 'ChainSchedule' that satisfies 'checkHonestChain'
+--
+-- The distribution this function samples from is not simple to describe. It
+-- begins by drawing a sample of length 'Len' from the Bernoulli process
```
facundominguez, line 257:
```suggestion
-- The distribution this function samples from
-- begins by drawing a sample of length 'Len' from the Bernoulli process
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+-- induced by the active slot coefficient 'Asc'. (IE 'Len' many i.i.d. samples
+-- from @Uniform(f)@). Then it slides an @s-d@ window along that leader
+-- schedule, toggling as many empty slots as necessary to ensure every such
+-- window has at least @k+1@ active slots, ie Enriched Praos Chain
+-- Growth---HOWEVER, it is a difficult computation to add a /strictly minimal/
+-- number of active slots to the output of the Bernoulli process (it's an
+-- integer programming optimization problem as far as we can tell). We have
+-- settled for a relatively simple algorithm that approaches the minimal number
+-- of toggles (TODO calculate how tight the bounds are).
```
facundominguez, line 266: Could this note explain or point to a description of what the simple algorithm is?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+-- concatentation of a few long periodic and minimal runs with the occasional
+-- window with @k+2@ active slots demarcating the runs.
+--
+-- TODO sample from a disjunction of generators that explore interesting
+-- boundary conditions. EG one would fill in each sparse window by setting the
+-- /youngest/ empty slots in it.
+uniformTheHonestChain ::
+  forall base hon g.
+     R.RandomGen g
+  => Maybe Asc   -- ^ 'Nothing' means @0@, which induces a periodic chain
```
facundominguez, line 284: Is it meaningful to have `Asc == 0`? I would imagine in such case no slot would be active.

Moreover, are `Nothing` and `Just 0` two representation of the same value in the domain?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/BitVector.hs
```diff
@@ -0,0 +1,183 @@
...
+            <*> Some.readArg
+
+fillInWindow ::
+  forall proxy pol base g s.
+     (POL pol, R.StatefulGen g (ST s))
+  => proxy pol
+  -> SomeDensityWindow pol
+  -> g
+  -> C.MVector base SlotE s S
+  -> ST s (C.Var base (PreImage pol ActiveSlotE))   -- ^ the count after filling
```
facundominguez, line 140: Could we have an explanation of what this function is expected to do?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: `Asc` cannot be `0`, by construction. (See its Haddock.)

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: I agree 100%

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Ah, then the documentation of the `Maybe Asc` parameter makes no sense. `Nothing` can't mean `0`.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
```diff
@@ -0,0 +1,341 @@
...
+(.+) :: Int -> Int -> Int
+(.+) = (Prelude.+)
+
+(.-) :: Int -> Int -> Int
+(.-) = (Prelude.-)
+
+-----
+
+-- | Indexed by what you're counting from, what you're counting, and which of them you're counting.
+newtype Count (base :: Type) (elem :: kelem) (which :: kwhich) = Count Int
```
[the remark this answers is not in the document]
facundominguez: Addressed in cfaf6e6

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+-----
+
+data HonestRecipe = HonestRecipe !Kcp !Scg !Delta !Len
+  deriving (Eq, Read, Show)
+
+data CheckedHonestRecipe base hon = UnsafeCheckedHonestRecipe {
+    chrEhcgDensity :: !(BV.SomeDensityWindow S.NotInverted)
+  ,
+    chrWin         :: !(C.Contains SlotE base hon)
+  }
```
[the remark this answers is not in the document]
facundominguez: Documented in 5d60adb.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,647 @@
...
+                     Delta (Delta), Kcp (Kcp), Len (Len), Scg (Scg), ascVal)
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Slot as S
+import           Test.Ouroboros.Consensus.ChainGenerator.Slot
+                     (E (ActiveSlotE, SlotE), S)
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Some as Some
+
+-----
+
+data HonestRecipe = HonestRecipe !Kcp !Scg !Delta !Len
+  deriving (Eq, Read, Show)
```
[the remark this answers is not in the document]
facundominguez: I think this should be called something like `HonestSchemaSpec`.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,144 @@
...
+
+    bool = q < ascVal asc
+
+showS :: S -> ShowS
+showS (S bool) = showChar $ if bool then '1' else '0'
+
+-----
+
+-- | The different kinds of element counted in this library
+data E = ActiveSlotE | EmptySlotE | SlotE
```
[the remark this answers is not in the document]
amesgen: Your description for `ActiveSlotE` and `EmptySlotE` are correct, and `SlotE` is just for when you want to count slots without making a statement whether they are active or empty.

Eg `HonestChainViolation` has a constructor `BadLength !(C.Size hon SlotE)` which reports the actual size of the honest chain (unequal to the requested size from the recipe).

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,144 @@
...
+  -> C.Count base (PreImage pol ActiveSlotE) which
+  -> C.Count base (PreImage pol EmptySlotE ) which
+complementActive _pol (C.Count n) (C.Count i) = C.Count (n - i)
+
+-- | Every slot is either active or empty
+complementEmpty ::
+     proxy pol
+  -> C.Size base SlotE
+  -> C.Count base (PreImage pol EmptySlotE ) which
+  -> C.Count base (PreImage pol ActiveSlotE) which
```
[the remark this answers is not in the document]
amesgen: When `s ~ SlotE`, considering the complement is maybe a bit too weird ("please count how many of these slots are no slots"), so one needs some evidence that `s` is not `SlotE`. One could write something with almost the type signature you suggested (extra `proxy` argument to avoid ambiguous types):
```haskell
class Complementable s where
  complement ::
       proxy' s
    -> proxy pol
    -> C.Size base SlotE
    -> C.Count base (PreImage pol s)                     which
    -> C.Count base (PreImage pol (PreImage Inverted s)) which

instance Complementable ActiveSlotE where
  complement _ = complementActive

instance Complementable EmptySlotE where
  complement _ = complementEmpty
```
I guess the reason why it doesn't exist is that the two separate functions are easy enough to use?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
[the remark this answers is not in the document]
amesgen: Slots do not have a polarity, this is an implementation detail to allow functions to work both active and empty slot. As an example, consider 
```haskell
findIthEmptyInMV ::
     POL   pol
  => proxy pol
  -> C.MVector base SlotE s S
  -> C.Index base (PreImage pol EmptySlotE)
  -> ST s (MaybeFound base)
```
As the information whether an `Index` is counting active or empty slots is tracked in the types by a phantom type, but the implementation has to be different depending on whether we want to find active or empty slots, we need some evidence (here: the `POL` instance) to decide how to test whether a given slot is active or inactive.

With simpler types (but generalized to any `a` with `Eq a`, but we only need it for `a ~ Bool` or newtypes of `Bool`), this would be
```haskell
-- | Return the index of the @i@-th occurrence of @a@ in the list.
findIthOccurrence :: 
     Eq a
  => a                    -- corresponds to `pol`
  -> [a]                  -- corresponds to `C.MVector base SlotE s S`
  -> Int                  -- `C.Index base (PreImage pol EmptySlotE)`
  -> Maybe Int
findIthOccurrence a as i =
    [ j | (j, b) <- zip [0..] as, a == b ] !? i
```
This is of course much simpler, as there it does not try to track what "kind" of index the input/output `Int`s are, but one also does not get any protection against eg using the returned `Int` incorrectly.

amesgen: Maybe https://github.com/input-output-hk/ouroboros-consensus/pull/240#discussion_r1288332318 gives some motivation?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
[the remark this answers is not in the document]
facundominguez: Oh, indeed, `s` would be ambiguous without the @proxy' s@ parameter. Thanks for the elaboration.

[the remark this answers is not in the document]
facundominguez: Added a note in 049bcc4e

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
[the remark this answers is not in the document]
facundominguez: Thanks @amesgen.

I documented `E` in 600eabb7.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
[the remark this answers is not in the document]
facundominguez: I elaborated a bit more the documentation of Pol in 8a9363577
Thanks @amesgen!

facundominguez: It does!

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,144 @@
...
+    basicLength (V_S v)               = VG.basicLength v
+    basicUnsafeSlice i l (V_S v)      = V_S (VG.basicUnsafeSlice i l v)
+    basicUnsafeIndexM (V_S v) i       = coerce <$> VG.basicUnsafeIndexM v i
+    basicUnsafeCopy (MV_S mv) (V_S v) = VG.basicUnsafeCopy mv v
+    elemseq (V_S v) x y               = VG.elemseq v (coerce x) y
+instance V.Unbox S
+
+-----
+
+genAsc :: R.RandomGen g => Asc -> g -> (S, g)
```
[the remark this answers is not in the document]
facundominguez: Renamed to `genS` in 037f95be

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/BitVector.hs
```diff
@@ -0,0 +1,183 @@
...
+    SomeDensityWindow
+        !(C.Var  slidingWindow (PreImage pol ActiveSlotE))
+        !(C.Size slidingWindow SlotE)
+
+instance Eq (SomeDensityWindow pol) where
+    SomeDensityWindow l1 l2 == SomeDensityWindow r1 r2 =
+        Some.runEq
+      $ Some.eqCtor (SomeDensityWindow @pol) (SomeDensityWindow @pol)
+            `Some.eqArg` (C.forgetBase, l1, C.forgetBase, r1)
+            `Some.eqArg` (C.forgetBase, l2, C.forgetBase, r2)
```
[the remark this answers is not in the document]
facundominguez: Adopted in 45cb4c9.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Renamed ChainSchedule to ChainSchema in bb2d522bf. I could not prefix the name with "Honest" because the PR seems to use the same type for the adversarial schema.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: "Praos Enriched Chain Growth Assumption" should be changed to "Length of Competing Chains" as in the test plan. The concern about subtracting or not subtracting $\Delta$ then becomes irrelevant, I think.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,657 @@
...
+
+    when (l <= 0) $ Exn.throwError BadLen
+
+    when (k < 0 || s - d <= k) $ Exn.throwError BadKcp
+
+    C.withTopWindow (C.Lbl @HonestLbl) l $ \base topWindow -> do
+        C.SomeWindow Proxy slots <- pure topWindow
+
+        pure $ SomeCheckedHonestRecipe base Proxy UnsafeCheckedHonestRecipe {
+            chrEhcgDensity = BV.SomeDensityWindow (C.Count (k + 1)) (C.Count (s - d))
```
facundominguez, line 121:
```suggestion
            chrEhcgDensity = BV.SomeDensityWindow (C.Count k) (C.Count s)
```
It is unclear why we should not test `k/s`. Whatever adversarial acceleration happens after the stability window, it should be contained by the /common prefix/ property. The generation of adversarial chains won't produce k+1-race-winning fragments either.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,657 @@
...
+     HonestRecipe
+  -> ChainSchema base hon
+  -> Exn.Except (HonestChainViolation hon) ()
+checkHonestChain recipe sched = do
+    when (C.getCount sz /= l) $ Exn.throwError $ BadLength sz
+
+    do  let pc = countChainSchema sched
+        when (C.toVar pc <= 0) $ Exn.throwError BadCount
+
+    -- every slot is the first slot of a unique EHCG window
```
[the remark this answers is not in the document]
facundominguez: Probably means "Enriched Honest Chain Growth" and EHCG window is a window of length $s-\Delta$. But I think we can get rid of such concept and stick with a prosaic stability window.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,521 @@
...
+-- Definition of the /Praos Race Assumption/. We assume every adversarial chain
+-- contains at most @k@ blocks in the Praos Race Window anchored at the
+-- intersection.
+--
+-- Definition of the /Extended Praos Race Assumption/. Beyond the Praos Race
+-- Assumption, we assume that every adversarial chain also has at most @k@
+-- blocks in every subsequent Praos Race Window that ends within 'Delta' of the
+-- Stability Window anchored at the first adverarial block. (Crucially, these
+-- are the Praos Race Windows in which the the adversary could not yet have
+-- drastically accelerated its rate of election.)
```
facundominguez, line 90: And this is the property that I think we should replace with "Length of competing chains". The two properties are similar, but LoCC does not care about the stability window (that is the sole concern of Chain Growth instead).

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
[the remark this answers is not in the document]
amesgen: You already found it, just linking it here FTR: https://github.com/input-output-hk/ouroboros-consensus/pull/240#discussion_r1290439479

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,657 @@
...
+                ehcgvPopCount = pc
+              ,
+                ehcgvWindow   = ehcg
+              }
+
+  where
+    HonestRecipe (Kcp k) (Scg s) (Delta d) (Len l) = recipe
+
+    -- the general EHCG window contains @'Scg' - 'Delta'@ slots
+    ehcgWidth = s - d :: Int
```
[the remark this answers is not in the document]
amesgen: I think this is the same discussion as in https://github.com/input-output-hk/ouroboros-consensus/pull/240#discussion_r1266960323

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
amesgen: Why can't `Nothing` mean `0`? The intention of feeding `Nothing` is to generate a chain where no block can be removed without violating any of the safety properties.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
amesgen: :+1: on the general impetus to use the terminology from the test plan also here in the code, and indeed they are similar.

A difference is that the LoCC is concerned just with the first race right after an intersection, while the EPRA is concerned with additional races beyond that (that is also why the EPRA has to talk about the stability window in contrast to the LoCC, as that constraints the adversary in "speeding up"). As you say, this is to rule out "improbable" leader schedules. We could either incorporate that into the LoCC in the test plan, or have an additional property that captures the guarantees that the EPRA currently gives on top of LoCC+DoCC.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: `Nothing` cannot meant `Asc==0` because the meaning of `Asc` is not defined when it is `0`.  

> The intention of feeding Nothing is to generate a chain where no block can be removed without violating any of the safety properties.

But I think this "intention" is an alright meaning for `Nothing`. It just needs to be documented :)

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
```diff
@@ -0,0 +1,378 @@
...
+--
+-- Note that the window can spill either on the right if @i + innerSz > outerSz@,
+-- or it can spill on the left if @i < 0@, or it can spill on both sides
+-- simultaneously.
+--
+withWindow :: Size outer elem -> Lbl lbl -> Index outer elem -> Size x elem -> SomeWindow lbl outer elem
+withWindow (Count n) _lbl (Count i) (Count m) =
+    SomeWindow Proxy $ UnsafeContains (Count i') (Count m')
+  where
+    i' = min n (max 0 i)
```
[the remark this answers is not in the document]
amesgen: Hmm, when we have
```haskell
C.SomeWindow Proxy (C.Contains (C.Count i') (C.Count m')) =
  C.withWindow (C.Count n) (C.Lbl :: C.Lbl "test") (C.Count i) (C.Count m)
```
and `n >= 0` and `i >= n`, then we get `i' >= n` which I guess you want to avoid using `min n` as it is beyond the containing window. But we also have `m' == 0`, so the resulting window is empty.

I think there are two options:

 - All empty windows are considered identically (and are in particular vacuously contained in all other windows) regardless of what their index is.
 - Adapt the `prop_withWindow` test to only generate `i < n`, eg naively like this
   ```diff
   diff --git a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs
   index a32a4eff3..ae329fc6d 100644
   --- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs
   +++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs
   @@ -19,6 +19,7 @@ tests = [
    
    prop_withWindow :: QC.NonNegative Int -> Int -> Int -> QC.Property
    prop_withWindow (QC.NonNegative n) i m =
   +    (i < n) QC.==>
        case C.withWindow (C.Count n) (C.Lbl :: C.Lbl "test") (C.Count i) (C.Count m) of
            C.SomeWindow Proxy (C.Contains (C.Count i') (C.Count m')) ->
                QC.counterexample (show (i', m'))
   ```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Ah, I think I'm just understanding EPRA enough to recognize that it is accomplishing the same as the strengthened LoCC.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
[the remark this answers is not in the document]
facundominguez: I am thinking of 
```diff
--- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs
+++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Counting.hs
@@ -25,7 +25,9 @@ prop_withWindow (QC.NonNegative n) i m =
           $ QC.conjoin [
                 if i < 0
                 then QC.counterexample "neg i" $ i' QC.=== 0
-                else QC.counterexample "nonneg i" $ i' QC.=== i
+                else if i > n
+                     then QC.counterexample "nonneg i" $ i' QC.=== n
+                     else QC.counterexample "too large i" $ i' QC.=== i
               , if m < 0
                 then QC.counterexample "neg m" $ m' QC.=== 0
                 else QC.counterexample "nonneg m" $ min (n - 1) i' + m' <= n
```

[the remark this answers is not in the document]
facundominguez: > I am thinking of ...

Adopted this in ccc5a40b for now.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: In c06d54a I renamed EPRA to Adversarial chains in stability windows.

I'm changing the name to make it more precise, and also because it doesn't obviously imply the Praos Race Assumption. In particular the implication is unclear when the Praos Race Window is larger than the stability window.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Actually, the check of the adversarial chain was checking "Length of Competing Chains" although with a different name: "Praos Race Assumption".

In cd5aff46 I replaced the EHCG window with the stability window. The concerns about races are only checked for the schemas of the adversarial chains.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,521 @@
...
+--
+-- Definition of a /Praos Race Window/ of a chain. It is an interval of slots
+-- that contains at least @k+1@ blocks of the chain and exactly 'Delta' slots
+-- after the @k+1@st block.
+--
+-- Definition of the /Length of Competing Chains/. We assume every adversarial
+-- chain contains at most @k@ blocks in the Praos Race Window anchored at the
+-- the intersection.
+--
+-- Definition of the /Adversarial Chains in Stability Windows/.
```
facundominguez, line 88: Maybe "Races in Stability Windows" is shorter and more evocative.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,650 @@
...
+
+    when (l <= 0) $ Exn.throwError BadLen
+
+    when (k < 0 || s - d <= k) $ Exn.throwError BadKcp
+
+    C.withTopWindow (C.Lbl @HonestLbl) l $ \base topWindow -> do
+        C.SomeWindow Proxy slots <- pure topWindow
+
+        pure $ SomeCheckedHonestRecipe base Proxy UnsafeCheckedHonestRecipe {
+            chrScgDensity = BV.SomeDensityWindow (C.Count (k + 1)) (C.Count s)
```
facundominguez, line 121: Changing the density here to `k` instead of `k+1` breaks the tests, and I'm guessing that it will require fixing the generation of the first stability window of the schema. 

I would need to have the generation simplified before doing this change, or I would need to pair with someone to do it.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,521 @@
...
+    AdversarialRecipe {
+        -- | The honest chain to branch off of
+        arHonest :: !(ChainSchema base hon)
+      ,
+        -- | protocol parameters
+        arParams :: (Kcp, Scg, Delta)
+      ,
+        -- | The number of (real) honest active slots as of the intersection
+        --
+        -- @0@ means the genesis block and @i > 0@ means @i - 1 :: C.Index hon ActiveSlotE@
```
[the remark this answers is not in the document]
nfrisby:
```
        -- | Where to branch off of 'arHonest'
        --
        -- Specifically, how many blocks should the honest and adversarial chain share. In other words, the 0-based index of their intersection, such that `0` identifies the genesis block.
        arPrefix :: !(C.Var hon ActiveSlotE)
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -73,29 +73,29 @@ data AdversarialViolation hon adv =
...
+-- Chains.
 --
--- Definition of a /Praos Race Window/. Such a window includes exactly 'Delta'
--- slots after the @k+1@st honest block in the window.
+-- Definition of a /Praos Race Window/ of a chain. It is an interval of slots
+-- that contains at least @k+1@ blocks of the chain and exactly 'Delta' slots
+-- after the @k+1@st block.
 --
--- Definition of the /Praos Race Assumption/. We assume every adversarial chain
--- contains at most @k@ blocks in the Praos Race Window anchored at the
```
nfrisby: Minor: "anchored" is a well-established term in the rest of the Consensus code base and documentation. I generally suggest not replacing uses of it.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
nfrisby: "Definition of _the_ /.../" would require a noun phrase to be the ellipses, such as Assumption, Property, etc.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -73,29 +73,29 @@ data AdversarialViolation hon adv =
...
 --
--- Definition of the /Extended Praos Race Assumption/. Beyond the Praos Race
--- Assumption, we assume that every adversarial chain also has at most @k@
--- blocks in every subsequent Praos Race Window that ends within 'Delta' of the
--- Stability Window anchored at the first adverarial block. (Crucially, these
--- are the Praos Race Windows in which the the adversary could not yet have
+-- Definition of the /Adversarial Chains in Stability Windows/.
+-- Every adversarial chain has at most @k@ blocks in every Praos Race Window of
+-- the honest chain and ending within 'Delta' blocks of the Stability Window
+-- starting at the slot of the first adversarial block. (Crucially, these
```
nfrisby, line 91: I think there's a typo in the current text. Is the following suggestion accurate?

```suggestion
-- Every adversarial chain has at most @k@ blocks in every Praos Race Window
-- that ends at least 'Delta' slots before the Stability Window
-- anchored at the first adversarial block. (Crucially, these
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Do you mean window anchoring? It took me a bit to find the definition of that. But I don't feel strong. I'll reintroduce it.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Got it.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: I edited your suggestion as I think it would make it accurate. I don't think anchoring will make it easy to express the same stability window.

The definition I have of anchored window is:
> A window anchored at a block starts with the slot immediately after that block.

... from `Honest.hs`.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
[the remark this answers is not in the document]
facundominguez: Adopted in 94570155

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Adopted in 491a9d9

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -73,29 +73,29 @@ data AdversarialViolation hon adv =
...
--- Definition of a /Praos Race Window/. Such a window includes exactly 'Delta'
--- slots after the @k+1@st honest block in the window.
+-- Definition of a /Praos Race Window/ of a chain. It is an interval of slots
+-- that contains at least @k+1@ blocks of the chain and exactly 'Delta' slots
+-- after the @k+1@st block.
 --
--- Definition of the /Praos Race Assumption/. We assume every adversarial chain
--- contains at most @k@ blocks in the Praos Race Window anchored at the
--- intersection.
+-- Definition of the /Length of Competing Chains/. We assume every adversarial
```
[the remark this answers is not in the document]
facundominguez: Adopted in https://github.com/input-output-hk/ouroboros-consensus/commit/491a9d94c30e61475f389e71c941ed8e5d6d7b24

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: 491a9d9 corrects the substitution of blocks with slots.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: I just realized this occurrence of anchored is preserved in the current formulation of PRA as LoCCA. Please, let me know if any action is needed to address this comment.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,135 @@
...
+
+    pure $! UnsafeRace $ C.withWindowBetween
+        sz
+        (C.Lbl @RaceLbl)
+        (next0 C.+ 1)
+        nextK
+  where
+    sz = C.lengthV v
+
+nextConservative ::
```
facundominguez, line 109: It doesn't look like `nextConservative` or `initConservative` would yield race windows (they might contain more or less than k+1 active slots).

These functions need to be documented.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,525 @@
...
+    checkRaces = do
+        let iterH =
+                maybe (RI.initConservative (Scg s) (Delta d) winH) id
+              $ RI.init (Kcp k) vH
+        case RI.init (Kcp (k - 1)) vA >>= RI.next vA of
+            Nothing    -> pure ()   -- there are <= k total adversarial active slots
+            Just iterA ->
+                -- TODO optimization: how to skip all the honest Race Windows that
+                -- don't reach beyond the intersection? Perhaps skip to i - s + d?
+                go iterH iterA
```
facundominguez, line 173: Addressing the TODO would require counting blocks backwards from the intersection. Is that ideal?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,525 @@
...
+                    Exn.throwError $ BadAnchor WrongNumberOfHonestPredecessors
+
+    checkCount = do
+        let pc = BV.countActivesInV S.notInverted vA
+        when (C.toVar pc <= 0) $ Exn.throwError BadCount
+
+    -- the youngest slot in the adversarial schedule that cannot have accelerated
+    --
+    -- (IE s past the first active adversarial slot.)
+    youngestStableA :: C.Index adv SlotE
```
facundominguez, line 153: The first slot after the stability window can have a leader resulting from an accelerated leader schedule IIUC. So something is off with the name of this variable, or its value needs to be corrected.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,151 @@
...
+-- @Nothing@.
+--
+initConservative :: Scg -> Delta -> C.Contains SlotE outer inner -> Race inner
+initConservative (Scg s) (Delta d) win =
+    UnsafeRace
+  $ C.withWindow
+        (C.windowSize win)
+        (C.Lbl @RaceLbl)
+        (C.Count 0)
+        (C.Count (s - d))
```
facundominguez, line 71: The size of the window needs to extend past the container size in order for the conservative Race window to have `k+1` active slots. Moreover, it should be counted how many active slots there are in the overlapping portion of the window, otherwise the race window might be too small.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
facundominguez: Done in dc7707eb.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Adopted in 37b643d9.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -292,7 +292,11 @@ would solve the problem with just two toggles.
 uniformTheHonestChain ::
   forall base hon g.
      R.RandomGen g
-  => Maybe Asc   -- ^ 'Nothing' means @0@, which induces a periodic chain
+  => Maybe Asc   -- ^ When @Nothing@, the generated schema has a minimal amount
+                 -- of active slots. Deactivating any of them would violate
+                 -- safety properties. The generated schema shows a periodic
+                 -- structure, but this is more an artifact of the implementation
+                 -- than a feature.
```
amesgen, line 299: This sounds a bit like that there might be non-periodic minimally dense schemas, but the implementation is just not generating them. Maybe
```suggestion
                 -- safety properties. The generated schema shows a periodic
                 -- structure which is determined by the first stability window.
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,151 @@
...
+            Just
+         $! UnsafeRace
+         $  C.withWindowBetween
+                (C.lengthV v)
+                (C.Lbl @RaceLbl)
+                (C.Count 0)
+                kPlus1st
+
+-- | @initConservative@ is like @init@, but assumes that any Race window
+-- spilling the vector length has all the excedent slots active.
```
amesgen, line 59:
```suggestion
-- spilling the vector length has all the exceeding slots active.
```
(also for `nextConservative`)

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,151 @@
...
+--
+-- This function is only safe to call when we know that 'init' would return
+-- @Nothing@.
+--
```
amesgen, line 63: I think that `initConservative` is actually assuming that the chain is as sparse as still allowed by chain growth (here, currently still via EHCG, ie `k+1` blocks in `s-d` slots); in opposition to assuming that every slot is active (which would be maximally dense, ie the size of the race window would be exactly `k+1`).

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -87,12 +87,16 @@ data AdversarialViolation hon adv =
 -- chain contains at most @k@ blocks in the Praos Race Window anchored at the
 -- the intersection.
 --
--- Definition of the /Races in Stability Windows Assumption/.
+-- Definition of /Acceleration Lower Bound of an Adversarial Chain/. It is
+-- the lowest slot at which an adversary could speed up its elections on a
+-- given adversarial chain. It is defined as the last slot of the Stability
+-- Window anchored at the slot of the first adversarial block or the slot
+-- after the k+1st honest block after the intersection (whichever happens later).
```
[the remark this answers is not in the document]
facundominguez: Adopted in 16d00b3d3

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -13,6 +13,22 @@ module Test.Ouroboros.Consensus.ChainGenerator.RaceIterator (
...
+
+The first window is produced by 'init', which unconditionally starts the window at the first slot.
+This window can then be passed to 'next', which starts the new window after the first active slot.
+
+@
+---X--X--X--X
+^ start of window 1 from 'init'
+    ^ start of window 2 from 'next'
+       ^ start of window 3 from 'next'
+@
```
amesgen, line 26: Maybe explicitly state `k=1` for this example (otherwise, window 3 can not be returned by `next` as it contains only two active slots), or extend the schema with dots to indicate that it does not necessarily end after the fourth active slot?

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -13,6 +13,22 @@ module Test.Ouroboros.Consensus.ChainGenerator.RaceIterator (
...
+
+Valid windows must have @k+1@ active slots.
+If the vector doesn't have sufficient slots to meet this condition, 'init' and 'next' return 'Nothing' and we fall back
+to 'initConservative' and 'nextConservative', which pad the window with only active slots.
```
amesgen, line 30: What do you mean by "pad the window with only active slots"? Also see https://github.com/input-output-hk/ouroboros-consensus/pull/240#discussion_r1295818408

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,536 @@
...
+                -- in an active slot and B) iterH contains that anchor. Thus
+                -- adv has k+1 in iterH.
+                Exn.throwError $ BadRace AdversaryWonRace {
+                    rvAdv = iterA
+                  ,
+                    rvHon = iterH
+                  }
+
+          -- advance the honest Race Window
+          | otherwise -> case RI.next vH iterH <|> RI.nextConservative (Scg s) (Delta d) vH iterH of
```
facundominguez, line 215: Since "Races before acceleration lower bound" doesn't require checking windows that spill over the end of time, conservative race windows aren't necessary here.

facundominguez: This counter-example is produced if we just remove conservative race windows from adversarial schema generation.

k=5, s=14, d=1
```
0111100101000001111
        11111111110
```

We need to propose how to change the generation to stop examples like these from being produced, and we still need to check that they aren't produced.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/BitVector.hs
facundominguez: Done in f63c3a0e

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
tek: oops, wrote that from the wrong perspective

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
tek: ✅

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
amesgen: > If the vector doesn't have sufficient slots to meet this condition, `init` and `next` return `Nothing` and we fall back to `initConservative` and `nextConservative`, which create a window of size `scg - delta` unconditionally, which is later filled in with active slots to have maximum density within the chain growth limit.

Hmm, I am still confused, where do we later fill conservative race windows with active slots of maximum density? For comparison, WDYT about this:

> If the vector doesn't have sufficient slots to meet this condition, `init` and `next` return `Nothing` and we fall back to `initConservative` and `nextConservative`, which create a window of size `scg - delta` unconditionally, in order to constrain adversarial active slots there to satisfy the Length of Competing Chains Assumption.

tek: hmmm right I interpreted this:
```
the vector length spans the full stability window (minus delta) and has all the exceeding slots active.
```

and your
```
 as sparse as still allowed by chain growth
```

as implying that `uniformAdversarialChain` treats the spillover part specially, but it doesn't

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,164 @@
...
+@
+---X--X--X--X-- ...
+^ start of window 1 from 'init'
+    ^ start of window 2 from 'next'
+       ^ start of window 3 from 'next'
+@
+
+Valid windows must have @k+1@ active slots.
+If the vector doesn't have sufficient slots to meet this condition, 'init' and 'next' return 'Nothing' and we fall back
+to 'initConservative' and 'nextConservative', which create a window of size @scg - delta@ unconditionally, in order to
```
facundominguez, line 30: I'm afraid this is still not correct. The conservative functions always clip the race windows to the length of the containing sequence, so in practice they don't extend beyond the end.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,151 @@
...
+    pure $! UnsafeRace $ C.withWindowBetween
+        sz
+        (C.Lbl @RaceLbl)
+        (next0 C.+ 1)
+        nextK
+  where
+    sz = C.lengthV v
+
+-- | @nextConservative v r@ yields a window anchored at the first
+-- activel slot of @r@ if it exists, and extending until the end of @v@.
```
tek, line 133:
```suggestion
-- active slot of @r@ if it exists, and extending until the end of @v@.
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
facundominguez: Addressed in 5e0cc0797

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: I don't think we will remove the conservative race windows after more discussion.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -108,22 +105,20 @@ data NoSuchHonestChainSchema =
 
 checkHonestRecipe :: HonestRecipe -> Exn.Except NoSuchHonestChainSchema SomeCheckedHonestRecipe
 checkHonestRecipe recipe = do
-    when (s <= d) $ Exn.throwError BadDeltaScg
-
     when (l <= 0) $ Exn.throwError BadLen
 
-    when (k < 0 || s - d <= k) $ Exn.throwError BadKcp
+    when (k < 0 || s < k) $ Exn.throwError BadKcp
```
facundominguez, line 110: Maybe this should be `<=` to be consistent with the description of `BadKcp`.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -153,19 +153,25 @@ checkAdversarialChain recipe adv = do
...
+              BV.JustFound firstActiveA ->
                 -- if s=0, then the slot of their block is the youngest stable slot
-                firstActiveA C.+ s
-            BV.NothingFound           ->
+                C.fromWindow winA $ firstActiveA C.+ s
+              BV.NothingFound           ->
                 -- the rest of the function won' force this since there are no
                 -- adverarial active slots
                 error "dead code"
+            k1Plus1stYoungest = case BV.findIthEmptyInV S.inverted vH (C.toIndex arPrefix C.+ k) of
```
facundominguez, line 170: Mispelling! *kPlus1stYoungest* please.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -407,6 +413,13 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
             g
             mv
 
+    -- find the slot of the k+1 honest block
+    let kPlus1st :: C.Index adv SlotE
+        kPlus1st = case BV.findIthEmptyInV S.inverted (C.sliceV carWin vH) (C.Count k) of
```
facundominguez, line 418: Consider using the same mechanism as in `youngestStableA`.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Changed the description of `BadKcp` in 94f71d434.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Honest.hs
```diff
@@ -68,8 +68,8 @@ data TestHonest = TestHonest {
 genKSD :: QC.Gen (Kcp, Scg, Delta)
 genKSD = sized1 $ \sz -> do
     d <- QC.choose (0, div sz 4)
-    k <- QC.choose (0, 2 * sz)
-    s <- (\x -> x + d + k) <$> QC.choose (1, 3 * sz)   -- ensures @k + 1 / s - d <= 1@
+    k <- QC.choose (1, 2 * sz)
+    s <- (\x -> x + k) <$> QC.choose (1, 3 * sz)   -- ensures @k / s <= 1@
```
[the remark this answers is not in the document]
facundominguez: Adopted in a854186dd.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Done in c127eab20.

ouroboros-consensus/ouroboros-consensus.cabal
facundominguez: `hmatrix` and `genFirstWindowPopCount` were removed in 414e480. We deemed the implementation effort not worth the benefits for the time being, and we thought of simpler alternatives like generating longer chains and dropping their first windows.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,482 @@
...
+-- NOTE: when @'Asc' = 0@ the resulting leader schedule is periodic and has
+-- exactly @k@ active slots in every @s@ window.
+--
+-- NOTE: when @'Asc' << k/s@ the sample is likely to be the
+-- concatentation of a few long periodic and minimal runs with the occasional
+-- window with @k+1@ active slots demarcating the runs.
+--
+-- TODO sample from a disjunction of generators that explore interesting
+-- boundary conditions. EG one would fill in each sparse window by setting the
+-- /youngest/ empty slots in it.
```
facundominguez, line 283: I tweaked the expressions in this documentation after removing EHCG. But I'm not sure what all of these are saying. Maybe it is a good fragment to review together in a call.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Done in f0e51da8

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: I expanded this documentation in f0e51da8

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,523 @@
...
+
+-- | A 'ChainSchema' that satisfies 'checkHonestChain'
+--
+-- The distribution this function samples from
+-- begins by drawing a sample of length 'Len' from the Bernoulli process
+-- induced by the active slot coefficient 'Asc'. (IE 'Len' many i.i.d. samples
+-- from @Uniform(f)@). Then there is a second phase of slot activations.
+-- This phase slides a window along that leader
+-- schedule, toggling as many empty slots as necessary to ensure every such
+-- window has enoght active slots to achieve the given density
```
facundominguez, line 265:
```suggestion
-- window has enough active slots to achieve the given density
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -259,28 +259,69 @@ would solve the problem with just two toggles.
...
+-- > 000000000000000000000000 -- initial sample
+-- > 101100101100101100101100 -- second phase
+--
+-- > k=3, s=6
+-- >
+-- > 000000000000000000000000 -- initial sample
+-- > 000111000111000111000111 -- second phase
+--
+-- NOTE: When @'Asc'@ is close to 0, the sample will have a few active
+-- slots appart. Every time the sliding window encounters non-empty
```
tek, line 312:
```suggestion
-- slots apart. Every time the sliding window encounters non-empty
```

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Addressed when removing the EHCG.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Solved in db26c4e and the immediately following commits.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
facundominguez: This code has been rewritten since this comment was made and it no longer applies.

ouroboros-consensus/src/consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
facundominguez: Thanks @amesgen. You've helped clarifying this code since we had this conversation.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -0,0 +1,558 @@
...
+              BV.JustFound firstActiveA ->
+                -- if s=0, then the slot of their block is the youngest stable slot
+                C.fromWindow winA $ firstActiveA C.+ s
+              BV.NothingFound           ->
+                -- the rest of the function won' force this since there are no
+                -- adverarial active slots
+                error "dead code"
+            kPlus1stYoungest = case BV.findIthEmptyInV S.inverted vH (C.toIndex arPrefix C.+ k) of
+              BV.JustFound kPlus1st -> C.fromWindow winH $ kPlus1st C.+ d
+              BV.NothingFound       -> C.Count d
```
nfrisby, line 172: This case seems wrong to me. It would only be accurate if `k+1` were `0`, wouldn't?

I think `C.windowLast winH` would be better. Still inaccurate, but I think it'd at least lead to the right behavior.

Accurate would be: `C.windowLast winH C.d (k+1 - x) C.+ d` where `x` is how many blocks the honest chain does have remaining (where `x < k +1` since we're not in the `JustFound` alterantive).

nfrisby:
I pushed up several commits. Most are small. Most of them are rewordings of comments. But there is at least one big comment rewrite, on `uniformTheHonestChain`. cc @amesgen @facundominguez

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Agreed and adopted in e1135c0.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
```diff
@@ -0,0 +1,477 @@
...
+        <|>
+            ((,) True <$> RI.nextConservative vH iter)
+
+    go' = \case
+        Nothing -> []
+        Just (cons, iter@(RI.Race (C.SomeWindow Proxy win)))
+          | A.SomeCheckedAdversarialRecipe Proxy recipeA' <- someRecipeA'
+          , A.UnsafeCheckedAdversarialRecipe { A.carWin } <- recipeA'
+          , C.windowStart win < C.windowStart carWin
+          -> go' (next iter)
```
nfrisby, line 471: d7626941b78800b831f4e7a397ed91183e258c30 changed this to not print the race windows after the first.

But isn't it possible that the "problem" might be in one of those later race windows?

facundominguez: I don't think that reasoning works. There are two arguments to make against printing all of the windows.

1. The honest and the adversarial schemas are printed over and over in exactly the same way. The only things that change in the reprints are the indications of where the windows start and end, I think the developer can reasonably infer the windows on her own if the parameters are small. If the parameters are large, the developer won't bother in checking each of the many windows whether printed or not.
2. In mutation tests, the printed examples might not have any problem at all. The example is arbitrarily picked from all examples that pass the schema checks. If the error message could point at an example that has passed the schema check when it shouldn't, printing a verbose output would make more sense.

nfrisby: > If the parameters are large, the developer won't bother in checking each of the many windows whether printed or not.

I don't fully agree with this, but I do sympathize with it.

Perhaps the best answer is that the renderer should only print whichever race has the identified violation in it?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
nfrisby: I pushed d2de74c37b8e45b4a64064034af4f2470cbd6f0c, which I think does the "exact" calculation and isn't too onerous.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
nfrisby: Ah ha! Nevermind. I didn't realize this code was _only_ used for mutation tests.

facundominguez: > Perhaps the best answer is that the renderer should only print whichever race has the identified violation in it?

All race windows are equal in this regard: none of them causes the adversarial schema check to fail.

nfrisby:
@dnadales @jasagredo @fraser-iohk 

Our Tweag team working on the Genesis Statement of Work is making two claims about this PR.

- It is worthwhile.
- It is completely additive: the only way it could disrupt the Consensus codebase is if the tests we're adding here end up being flaky.

Please inspect this PR enough to feel comfortable with this claims and Approve it.

If you doubt the worthwhileness, please consider Approving anyway, and we can discuss whether or not to back it out after the fact. Thank you!

dnadales (approved):

dnadales:
I can confirm:

- It is worthwhile.
- It is completely additive: the only way it could disrupt the Consensus codebase is if the tests we're adding here end up being flaky.

PR #242, opened 2023-07-18
Add a multi-repl-enabled cabal to shell

amesgen:
This adds a multi-repl-enabled (see https://github.com/haskell/cabal/pull/8726) cabal (using a recent commit from `master`) to the shell.

When using GHC >=9.4, one can then run e.g.
```console
cabal-multi-repl repl ouroboros-consensus ouroboros-consensus-diffusion
```
to get a REPL with all components in `ouroboros-consensus` and `ouroboros-consensus-diffusion` :tada: 
`ghcid` also works, e.g.
```console
ghcid -c 'cabal-multi-repl repl ...'
```

Beware that this feature is still somewhat experimental, e.g. `cabal-multi-repl repl ouroboros-consensus-cardano` (or also `-protocol`) are stalling for me, and selecting individual components involving sublibraries fails with this message:
```console
 $ cabal-multi-repl repl ouroboros-consensus:consensus-testlib ouroboros-consensus:consensus-test
Error:
    Dependency on unbuildable library 'consensus-testlib' from ouroboros-consensus
```

Still, this might already be useful, and should not have any risks/maintenance burden.

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -6,6 +6,7 @@ in
 hsPkgs.shellFor {
   nativeBuildInputs = [
     pkgs.cabal
+    pkgs.cabal-multi-repl
     pkgs.fd
     pkgs.nixpkgs-fmt
     pkgs.stylish-haskell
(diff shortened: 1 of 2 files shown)


tek:
tried it out, appears to work! ran a test and changed something in the main lib and it recompiled.

PR #244, opened 2023-07-19
Minor tweaks in Shelley HFC protocol updates logic

amesgen:
See the commit messages

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
@@ -21,12 +21,9 @@ import qualified Cardano.Ledger.Shelley.API as SL
 import qualified Cardano.Ledger.Shelley.Core as Core
 import qualified Cardano.Ledger.Shelley.PParams as SL
 import           Control.Monad
-import           Data.List (sortBy)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe)
-import           Data.Ord (comparing)
-import           Data.Tuple (swap)
 import           Data.Void
 import           Data.Word (Word64)
 import           Lens.Micro ((^.))
@@ -37,7 +34,6 @@ import           Ouroboros.Consensus.Ledger.Inspect
 import           Ouroboros.Consensus.Shelley.Eras (EraCrypto)
 import           Ouroboros.Consensus.Shelley.Ledger.Block
 import           Ouroboros.Consensus.Shelley.Ledger.Ledger
-import           Ouroboros.Consensus.Util
 import           Ouroboros.Consensus.Util.Condense
 
 data ProtocolUpdate era = ProtocolUpdate {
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
```diff
@@ -121,14 +117,13 @@ protocolUpdates genesis st = [
             , proposalReachedQuorum = length votes >= fromIntegral quorum
             }
         }
-    | (proposal, votes) <- proposalsInv
```
nfrisby, line 124: For functional correctness, I think the only real question is whether `proposalsInv` could possible differ from its replacement.

The simple test below shows that the content of the lists is the same, but it does fail if I remove the two final `sort`s.

So: it's the same, except the votes will be in a different order. I don't think that matters in this context, so 👍.

```
{-# LANGUAGE LambdaCase #-}

module DoubleCheck where

import           Data.List (sort, sortBy)
import qualified Data.Map as Map
import           Data.Map (Map)
import           Data.Ord (comparing)
import           Data.Tuple (swap)
import qualified Test.QuickCheck as QC

main :: IO ()
main = QC.quickCheckWith QC.stdArgs{QC.maxSuccess=1000000} (prop @Int @Char)

prop :: forall a b. (Ord a, Ord b, Show a, Show b) => Map a b -> QC.Property
prop proposals =
     map (fmap sort) new QC.=== map (fmap sort) old
  where
    old = proposalsInv
    new = Map.toList $ invertMap proposals

    proposalsInv :: [(b, [a])]
    proposalsInv =
          groupSplit id
        . sortBy (comparing fst)
        $ map swap (Map.toList proposals)

    invertMap :: Ord b => Map a b -> Map b [a]
    invertMap = Map.fromListWith (<>) . fmap swizzle . Map.toList
      where
        swizzle (a, b) = (b, [a])

groupOn :: forall a b. Eq b => (a -> b) -> [a] -> [(b, [a])]
groupOn f = groupSplit (\a -> (f a, a))

groupSplit :: forall a b c. Eq b => (a -> (b, c)) -> [a] -> [(b, [c])]
groupSplit f = \case
    []   -> []
    a:as -> let (b, c) = f a in go b [c] as
  where
    go :: b -> [c] -> [a] -> [(b, [c])]
    go b acc []     = [(b, reverse acc)]
    go b acc (a:as)
        | b' == b   = go b (c:acc) as
        | otherwise = (b, reverse acc) : go b' [c] as
      where
        (b', c) = f a
```

nfrisby: For non-functional correctness, I would imagine the time- and space- performance changes, but 1) it's not obvious it'll get worse 2) it'd have to get much much worse to matter in this particular context.

Hmm. Does (2) remain true once Voltaire decentralizes governance? We'll have to ask that question even for the old definition...

nfrisby (approved):
Nice insight 👍

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
amesgen: Thanks for double-checking! We could even consider changing the type of `proposalVotes` to `Set (SL.KeyHash 'SL.Genesis (EraCrypto era))` instead of `[SL.KeyHash 'SL.Genesis (EraCrypto era)]`, but I don't think this is buying us that much.

> Hmm. Does (2) remain true once Voltaire decentralizes governance? We'll have to ask that question even for the old definition...

I'd *guess* that Voltaire will use some different mechanism, and not just basically increase the number of (Genesis) keys. At the very least, the type will have to change away from always using genesis keys, so we will have to revisit this then (though the current code will stay for old eras presumably).

PR #249, opened 2023-07-21
Expand documentation on HF ticking

### LINKED ISSUE #211: Expand documentation and tests on HF ticking in UTxO-HD
The documentation for `extendToSlot` should be expanded with an example that explains why it is a no-op if we are in one era and otherwise it translates Diffs

Also we should implement some unit test that illustrates the intended flow

jasagredo:
Closes #211


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State.hs
```diff
@@ -168,7 +168,40 @@ epochInfoPrecomputedTransitionInfo shape transition st =
...
+--
+-- 2. keep the Shelley diffs aside, and translate the Shelley ledger state
+--    without ledger tables, which produces a Mary ledger state together with a
+--    set of Mary differences.
+--
+-- 3. Translate the Shelley diffs to Mary differences, and prepend them to the
+--    ones created in the step 2.
+--
+-- 4. Attach the diffs resulting from step 3 to the Mary ledger state from step
+--    2, and return it.
```
jorisdral, line 204: I'm unsure how much we should talk about `Cardano` eras here. Maybe we can rewrite this example to mention generic `era1`, `era2`, `era3` or something?

jorisdral (approved):

jorisdral:
Almost forgot. #211 mentions that we should have unit tests for `extendToSlot`. Are we still going to do that?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State.hs
jasagredo: Sounds reasonable

PR #253, opened 2023-07-25
Refactor `LedgerTables` translation 

### LINKED ISSUE #144: Address partiality of `applyDistribLedgerTables`
`applyDistribLedgerTables` is partial. If we try to project into `Allegra` tables that have entries for `Alonzo` it will fail. This won't happen in absence of consensus bugs, but we might want to make this explicit and fail the query instead of killing the node with an error.

jorisdral:
# Description

Resolves #144


jorisdral:
Superseded by #262, #263, #264

PR #254, opened 2023-07-26
Add support for creating git tags to create-release.hs

fraser-iohk:
# Description

- Add support for creating git tags to create-release.hs
- Add a deprecation warning to create-release.sh (should it just be removed?)

### DIFF
--- a/scripts/release/create-release.sh
+++ b/scripts/release/create-release.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+echo "DEPRECATED!!! USE create-release.hs INSTEAD!!!"
+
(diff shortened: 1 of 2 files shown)


scripts/release/create-release.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+echo "DEPRECATED!!! USE create-release.hs INSTEAD!!!"
```
dnadales, line 3: Nice. Although can you remind me why we did not remove this script altogether?

fraser-iohk: no particular reason, I'm more than happy to remove it in this PR instead of just adding a warning

jasagredo (approved):

PR #269, opened 2023-08-01
Put strict bounds on `fs-api` and `fs-sim` dependencies

jorisdral:
# Description

Releasing a new version of `fs-api` or `fs-sim` breaks builds for older consensus versions. This is be mirrored by revisions in CHaP: see https://github.com/input-output-hk/cardano-haskell-packages/pull/417


jorisdral:
The benchmarks have become too slow and thus CI fails, but this PR doesn't change any of the dependencies since the bounds do not change which versions of `fs-api` and `fs-sim` are used for the build. I'm unsure what's wrong

dnadales (approved):

PR #280, opened 2023-08-08
Improve GHA caching

jorisdral:
# Description

I've done a bit of research into programmatic cache deletion through GitHub actions since our Linux cabal store cache had grown to 1.5GB. So far I could not find a satisfying, automatic solution to deleting caches. I looked into deleting caches based on a size threshold, but I realised that only deleting caches that are above a size threshold means we still keep around caches that are large but just under the size threshold. In addition, a tool like https://github.com/actions/gh-actions-cache/ does not provide filtering for cache sizes. It looks to me like hard "resets" like updating the `CABAL_CACHE_VERSION` in the workflow file is still the way to go, because it makes older caches unrestorable and therefore these caches are eventually evicted automatically if they are old enough. However, this requires manual intervention. One other option would be to use a similar approach to how we used to cache the `dist-newstyle` folder, using something like the month number, invalidating caches at the start of each month. @input-output-hk/ouroboros-consensus what do you think?

EDIT: One advantage of the month-number approach is that long-lived branches do not keep large caches alive. If `CABAL_CACHE_VERSION` is updated on `main`, only caches created on branches that branch off after that point will inherit the updated version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,16 +69,28 @@ jobs:
         | sort \
         | uniq > dependencies.txt
 
+    # This is used for caching the cabal store.
+    - name: Store month number as environment variable
+      run:  echo "MONTHNUM=$(/usr/bin/date -u '+%m')" >> $GITHUB_ENV
+
     # Restore the cabal store cache. See `save-cabal-store`.
-    - uses: actions/cache/restore@v3
+    #
+    # When we restore a previous cache and store a new key, we store both files
+    # that were already in the cache, and new files that were created. To
+    # prevent the cache from growing to quickly, we reset the cache each month.
+    #
+    # NOTE: it's vital that all restore-keys include the month number.
+    # Otherwise, we could restore a cache from a previous month, which makes
+    # caches grow unboundedly.
+    - name: "Restore cache for cabal-store"
+      uses: actions/cache/restore@v3
       id: restore-cabal-store
-      name: "Restore cache for cabal-store"
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store }}
-        key: cabal-store-${{ env.CABAL_CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+        key: cabal-store-${{ env.CABAL_CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ env.MONTHNUM }}-${{ hashFiles('dependencies.txt') }}
         restore-keys: |
-          cabal-store-${{ env.CABAL_CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cache-cleanup.yml
jorisdral, line 1: I'm unsure about this cache cleanup workflow. As can be read in the note, stacked branches will lose access to parent's caches as soon as the parent is merged, which is the major downside in my opinion.

.github/workflows/ci.yml
```diff
@@ -55,20 +55,17 @@ jobs:
...
         sudo apt-get -y install libsodium-dev
+        sudo apt-get -y install autoconf automake libtool
 
     - name: Configure Cabal to use libsodium
       run: |
         cp .github/workflows/cabal.project.local .
 
-    - name: "LINUX: Install build environment (for secp256k1)"
-      run: sudo apt-get -y install autoconf automake libtool
-
```
jorisdral, line 71: INFO: I merged this with the `Install build environment` step

.github/workflows/ci.yml
```diff
@@ -78,42 +75,44 @@ jobs:
     - name: Update Cabal (Hackage and CHaP)
       run: cabal clean; cabal update
 
-    - name: Generate dist-newstyle/cache/plan.json
-      run: cabal build all --dry-run --minimize-conflict-set
```
jorisdral, line 82: INFO: We're also doing this in the next step, so I removed it.

.github/workflows/ci.yml
```diff
@@ -128,8 +127,23 @@ jobs:
...
     - name: Build dependencies
+      id: build-dependencies
       run: cabal build --only-dependencies all -j
 
+    # Save the `cabal-store` cache if the dependencies were succesfully built,
+    # even if subsequent consensus builds/tests/benchmarks could fail. This
+    # should shave off some running time because dependencies don't have to be
+    # rebuilt if any of the later steps fail and the job has to be re-run.
+    #
+    # See https://github.com/actions/cache/tree/v3/save#always-save-cache.
```
jorisdral, line 138: INFO: We save built dependencies even if the remainder of the job fails. I hope this will speed up the runs a bit.

.github/workflows/cache-cleanup.yml
jasagredo: I think we can remove it, yeah

.github/workflows/ci.yml
jasagredo: Should not be necessary after #288

jasagredo (approved):
Looks good

.github/workflows/ci.yml
```diff
@@ -51,58 +51,82 @@ jobs:
...
-        cabal build all --dry-run
+        cabal build all --dry-run --minimize-conflict-set
         cat dist-newstyle/cache/plan.json \
         | jq '.["install-plan"][].id' \
         | sort \
         | uniq > dependencies.txt
 
-    - uses: actions/cache@v3
-      name: "Cache `cabal store`"
+    # This is used for caching the cabal store.
```
dnadales, line 72: Here it might be a good idea to comment on why we chose the month number.

dnadales (approved):

PR #282, opened 2023-08-09
Consolidate protocol parameters across eras

### LINKED ISSUE #275: Consolidate `ProtocolParams` across eras
Right now, the protocol parameters (`ProtocolParamsByron`, `ProtocolParamsShelley`, etc.) for different eras are unrelated data types. The goal of this ticket is to introduce a type/data family abstracting over these, and make use of it where appropriate.

As a concrete example, consider the type signature of `protocolInfoCardano`:
https://github.com/input-output-hk/ouroboros-network/blob/0aff5cf05f09bb56d4ac40f0090c636d5893dfb2/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L503-L517
After this change, this function would take sth like `NP ProtocolParamsFamily (CardanoEras c)` instead of the individual `ProtocolParams*` arguments.

A concrete benefit of this change would be that we could automatically extract the latest protocol version from this `NP`, making the approach of input-output-hk/ouroboros-network#4111 even more robust (also see https://github.com/input-output-hk/ouroboros-network/pull/4111#discussion_r1005815258).

jorisdral:
# Description

Closes #275.

This PR adds a new `ProtocolParams` data family, with instances for all Byron and Shelley blocks, and an instance for the Cardano block. This simplies `protocolInfoCardano` because per-era protocol parameter arguments are now bundled in the Cardano `ProtocolParams` type.

Tip: hiding whitespace should make the reviewable diff smaller.

See https://github.com/input-output-hk/ouroboros-consensus/pull/282#pullrequestreview-1571502484, which discusses possible follow-up issues for other improvements to `protocolInfoCardano`.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20230815_110046_joris_275_consolidate_protocolparams_across_eras.md
+++ b/ouroboros-consensus-cardano/changelog.d/20230815_110046_joris_275_consolidate_protocolparams_across_eras.md
@@ -1,0 +1,33 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Add a `ProtocolTransitionParams` data family, and provide instances for
+  transitions from Byron to Shelley and Shelley-based eras to Shelley-based
+  eras.
+- Add a data instance of `ProtocolParams` for the Cardano block.
+- Provide a `CardanoProtocolParams` type synonym and associated pattern synonym
+  (with record syntax).
+- Export all `ProtocolParams` and `ProtocolTransitionParams` instances from
+  `Ouroboros.Consensus.Cardano` and `Ouroboros.Consensus.Cardano.Node`.
+
+### Breaking
+
+- Refactor `ProtocolParamsByron` to a data instance of `ProtocolParams`.
+- Refactor protocol parameters for Shelley eras (e.g, `ProtocolParamsAlonzo` and `ProtocolParamsBabbage`) to data instances of `ProtocolParams`.
+- Export all Shelley `ProtocolParams` data instances from `Ouroboros.Consensus.Shelley.Node`.
+- Remove the `ProtocolTransitionParamsShelleyBased` datatype in favour of
+  `ProtocolTransitionParams`.
+- Make `protocolInfoCardano` require a `CardanoProtocolParams` type as its only
+  argument, instead of a long list of arguments.
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node.hs
@@ -8,18 +8,19 @@
 
 module Ouroboros.Consensus.Byron.Node (
     PBftSignatureThreshold (..)
-  , ProtocolParamsByron (..)
   , blockForgingByron
   , byronBlockForging
-  , defaultPBftSignatureThreshold
-  , mkByronConfig
-  , protocolClientInfoByron
-  , protocolInfoByron
     -- * Secrets
   , ByronLeaderCredentials (..)
   , ByronLeaderCredentialsError
   , mkByronLeaderCredentials
   , mkPBftCanBeLeader
+    -- * ProtocolInfo
+  , ProtocolParams (..)
+  , defaultPBftSignatureThreshold
+  , mkByronConfig
+  , protocolClientInfoByron
+  , protocolInfoByron
   ) where
 
 import qualified Cardano.Chain.Delegation as Delegation
@@ -151,7 +152,7 @@ mkPBftCanBeLeader (ByronLeaderCredentials sk cert nid _) = PBftCanBeLeader {
     }
 
 blockForgingByron :: Monad m
-                  => ProtocolParamsByron
+                  => ProtocolParams ByronBlock
                   -> [BlockForging m ByronBlock]
 blockForgingByron ProtocolParamsByron { byronLeaderCredentials      = mLeaderCreds
                                       , byronMaxTxCapacityOverrides = maxTxCapacityOverrides
@@ -169,7 +170,7 @@ defaultPBftSignatureThreshold :: PBftSignatureThreshold
 defaultPBftSignatureThreshold = PBftSignatureThreshold 0.22
 
 -- | Parameters needed to run Byron
-data ProtocolParamsByron = ProtocolParamsByron {
+data instance ProtocolParams ByronBlock = ProtocolParamsByron {
       byronGenesis                :: Genesis.Config
     , byronPbftSignatureThreshold :: Maybe PBftSignatureThreshold
     , byronProtocolVersion        :: Update.ProtocolVersion
@@ -178,7 +179,7 @@ data ProtocolParamsByron = ProtocolParamsByron {
     , byronMaxTxCapacityOverrides :: Mempool.TxOverrides ByronBlock
     }
 
-protocolInfoByron :: ProtocolParamsByron
+protocolInfoByron :: ProtocolParams ByronBlock
                   -> ProtocolInfo ByronBlock
 protocolInfoByron ProtocolParamsByron {
                       byronGenesis                = genesisConfig
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano.hs
@@ -8,12 +8,8 @@ module Ouroboros.Consensus.Cardano (
   , ProtocolCardano
   , ProtocolShelley
     -- * Abstract over the various protocols
-  , ProtocolParamsAllegra (..)
-  , ProtocolParamsAlonzo (..)
-  , ProtocolParamsByron (..)
-  , ProtocolParamsMary (..)
-  , ProtocolParamsShelley (..)
-  , ProtocolTransitionParamsShelleyBased (..)
+  , ProtocolParams (..)
+  , ProtocolTransitionParams (..)
   , module X
   ) where
 
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
@@ -1,27 +1,34 @@
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE FlexibleInstances     #-}
-{-# LANGUAGE LambdaCase            #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE NamedFieldPuns        #-}
-{-# LANGUAGE OverloadedStrings     #-}
-{-# LANGUAGE PatternSynonyms       #-}
-{-# LANGUAGE RankNTypes            #-}
-{-# LANGUAGE ScopedTypeVariables   #-}
-{-# LANGUAGE TypeApplications      #-}
-{-# LANGUAGE TypeFamilies          #-}
-{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE DataKinds                #-}
+{-# LANGUAGE FlexibleContexts         #-}
+{-# LANGUAGE FlexibleInstances        #-}
+{-# LANGUAGE LambdaCase               #-}
+{-# LANGUAGE MultiParamTypeClasses    #-}
+{-# LANGUAGE NamedFieldPuns           #-}
+{-# LANGUAGE OverloadedStrings        #-}
+{-# LANGUAGE PatternSynonyms          #-}
+{-# LANGUAGE RankNTypes               #-}
+{-# LANGUAGE ScopedTypeVariables      #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE TypeApplications         #-}
+{-# LANGUAGE TypeFamilies             #-}
+{-# LANGUAGE TypeOperators            #-}
 {-# OPTIONS_GHC -Wno-orphans
                 -Wno-incomplete-patterns
                 -Wno-incomplete-uni-patterns
                 -Wno-incomplete-record-updates
                 -Wno-overlapping-patterns #-}
+
+-- TODO: this is required for ghc-8.10.7, because using NamedFieldPuns and
+-- PatternSynonyms with record syntax results in warnings related to shadowing.
+-- This can be removed once we drop ghc-8.10.7.
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+
 module Ouroboros.Consensus.Cardano.Node (
     CardanoHardForkConstraints
+  , CardanoProtocolParams
   , MaxMajorProtVer (..)
-  , ProtocolParamsAllegra (..)
-  , ProtocolParamsMary (..)
-  , ProtocolTransitionParamsShelleyBased (..)
+  , ProtocolParams (.., CardanoProtocolParams, paramsByron, paramsShelleyBased, paramsShelley, paramsAllegra, paramsMary, paramsAlonzo, paramsBabbage, paramsConway, transitionParamsByronToShelley, transitionParamsShelleyToAllegra, transitionParamsAllegraToMary, transitionParamsMaryToAlonzo, transitionParamsAlonzoToBabbage, transitionParamsBabbageToConway)
+  , ProtocolTransitionParams (..)
   , TriggerHardFork (..)
   , protocolClientInfoCardano
   , protocolInfoCardano
@@ -62,6 +69,7 @@ import qualified Codec.CBOR.Encoding as CBOR
 import           Control.Exception (assert)
 import qualified Data.ByteString.Short as Short
 import           Data.Functor.These (These1 (..))
+import           Data.Kind (Type)
 import qualified Data.ListMap as ListMap
 import qualified Data.Map.Strict as Map
 import           Data.SOP.Counting
@@ -81,6 +89,8 @@ import           Ouroboros.Consensus.Cardano.CanHardFork
 import           Ouroboros.Consensus.Cardano.ShelleyBased
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.HardFork.Combinator
+import           Ouroboros.Consensus.HardFork.Combinator.AcrossEras
+                     (PerEraProtocolParams (..))
 import           Ouroboros.Consensus.HardFork.Combinator.Embed.Nary
 import           Ouroboros.Consensus.HardFork.Combinator.Serialisation
 import qualified Ouroboros.Consensus.HardFork.History as History
@@ -102,8 +112,6 @@ import           Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion
 import           Ouroboros.Consensus.Shelley.Node
 import           Ouroboros.Consensus.Shelley.Node.Common (ShelleyEraWithCrypto,
                      shelleyBlockIssuerVKey)
-import           Ouroboros.Consensus.Shelley.Node.Praos
-                     (ProtocolParamsBabbage (..), ProtocolParamsConway (..))
 import qualified Ouroboros.Consensus.Shelley.Node.Praos as Praos
 import qualified Ouroboros.Consensus.Shelley.Node.TPraos as TPraos
 import           Ouroboros.Consensus.Storage.Serialisation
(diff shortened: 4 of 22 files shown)


ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node.hs
```diff
@@ -8,18 +8,19 @@
...
   , ByronLeaderCredentials (..)
   , ByronLeaderCredentialsError
   , mkByronLeaderCredentials
   , mkPBftCanBeLeader
+    -- * ProtocolInfo
+  , ProtocolParams (..)
+  , defaultPBftSignatureThreshold
+  , mkByronConfig
+  , protocolClientInfoByron
+  , protocolInfoByron
```
jorisdral, line 23: INFO: introduced some sectioning here to group definitions related to `ProtocolInfo`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -14,10 +14,7 @@
 
 module Ouroboros.Consensus.Shelley.Node (
     MaxMajorProtVer (..)
-  , ProtocolParamsAllegra (..)
-  , ProtocolParamsAlonzo (..)
-  , ProtocolParamsMary (..)
-  , ProtocolParamsShelley (..)
+  , ProtocolParams (..)
```
jorisdral, line 17: INFO: protocol parameters for all shelley eras are now exported by this module. It should now be sufficient import `O.C.Shelley.Node` instead of a combination of `O.C.Shelley.Node`, `O.C.Shelley.Node.Praos` and `O.C.Shelley.Node.TPraos`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Praos.hs
```diff
@@ -1,18 +1,22 @@
...
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE TypeOperators       #-}
 
+
```
jorisdral, line 13: Remove redundant newline

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -48,15 +45,15 @@ import           Ouroboros.Consensus.Shelley.Eras (EraCrypto)
 import           Ouroboros.Consensus.Shelley.Ledger
 import           Ouroboros.Consensus.Shelley.Ledger.Inspect ()
 import           Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion ()
+import           Ouroboros.Consensus.Shelley.Node.Praos
```
jorisdral, line 48: INFO: previously, we couldn't import this module because it would introduce a cyclic dependency.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Praos.hs
```diff
@@ -114,14 +118,12 @@ praosSharedBlockForging
   ProtocolInfo
 -------------------------------------------------------------------------------}
 
--- | Parameters needed to run Babbage
-data ProtocolParamsBabbage c = ProtocolParamsBabbage {
+data instance ProtocolParams (ShelleyBlock (Praos c) (BabbageEra c)) = ProtocolParamsBabbage {
```
jorisdral, line 121: I should probably put the constructor on the next line. Idem for all the other `ProtocolParams` instances.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Praos.hs
```diff
@@ -30,9 +35,8 @@ import           Ouroboros.Consensus.Shelley.Eras (BabbageEra, ConwayEra,
                      EraCrypto, ShelleyBasedEra (shelleyBasedEraName))
 import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock,
                      ShelleyCompatible, forgeShelleyBlock)
-import           Ouroboros.Consensus.Shelley.Node
-                     (ShelleyLeaderCredentials (..))
-import           Ouroboros.Consensus.Shelley.Node.Common (ShelleyEraWithCrypto)
+import           Ouroboros.Consensus.Shelley.Node.Common (ShelleyEraWithCrypto,
+                     ShelleyLeaderCredentials (..))
```
jorisdral, line 39: INFO: this fixes the cyclic dependency mentioned in `O.C.Shelley.Node`.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano.hs
```diff
@@ -8,11 +8,7 @@ module Ouroboros.Consensus.Cardano (
   , ProtocolCardano
   , ProtocolShelley
     -- * Abstract over the various protocols
-  , ProtocolParamsAllegra (..)
-  , ProtocolParamsAlonzo (..)
-  , ProtocolParamsByron (..)
-  , ProtocolParamsMary (..)
-  , ProtocolParamsShelley (..)
+  , ProtocolParams (..)
```
jorisdral, line 11: INFO: Exports all `ProtocolParams` instances for Byron and Shelley eras if they are in scope.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -560,6 +560,54 @@ data ProtocolTransitionParamsShelleyBased era = ProtocolTransitionParamsShelleyB
     , transitionTrigger            :: TriggerHardFork
     }
 
+-- | Parameters needed to run Cardano.
+--
+-- TODO(jdral): could we get rid of 'shelleyBasedProtocolParams' somehow by
+-- including it in the 'ProtocolParams' of one or multiple eras?
```
jorisdral, line 566: I'm unsure what the best way to do this would be.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -560,6 +560,54 @@ data ProtocolTransitionParamsShelleyBased era = ProtocolTransitionParamsShelleyB
...
+data instance ProtocolParams (CardanoBlock c) = ProtocolParamsCardano {
+    cardanoProtocolParamsPerEra :: PerEraProtocolParams (CardanoEras c)
+  , shelleyBasedProtocolParams  :: ProtocolParamsShelleyBased (ShelleyEra c)
+  }
+
+-- TODO(jdral): move to Cardano.Block
+type CardanoProtocolParams c = ProtocolParams (CardanoBlock c)
+
+-- TODO(jdral): move to Cardano.Block
+pattern CardanoProtocolParams ::
```
jorisdral, line 576: Does this make sense?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -657,8 +665,55 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
...
+        } = paramsBabbage
+    ProtocolParamsConway {
+          conwayProtVer                = protVerConway
+        , conwayMaxTxCapacityOverrides = maxTxCapacityOverridesConway
+        } = paramsConway
+
     -- The major protocol version of the last era is the maximum major protocol
     -- version we support.
+    --
+    -- TODO: use index of CardanoProtocolParams NP
```
jorisdral, line 716: Currently, protocol versions are unique, named fields in each `ProtocolParams` data family instance. Should we introduce a class to extract protocol versions, such that we can do something like extracting the protocol version of the last era in the `NP` of protocol parameters?

jorisdral (commented):
I left some INFO comments, some remarks for me to resolve, and some questions for reviewers to consider

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -570,14 +618,7 @@ data ProtocolTransitionParamsShelleyBased era = ProtocolTransitionParamsShelleyB
...
-  -> ProtocolParamsShelley c
-  -> ProtocolParamsAllegra c
-  -> ProtocolParamsMary    c
-  -> ProtocolParamsAlonzo  c
-  -> ProtocolParamsBabbage c
-  -> ProtocolParamsConway  c
+  => CardanoProtocolParams c
   -> ProtocolTransitionParamsShelleyBased (ShelleyEra c)
   -> ProtocolTransitionParamsShelleyBased (AllegraEra c)
   -> ProtocolTransitionParamsShelleyBased (MaryEra c)
```
jorisdral, line 624: INFO: protocol transition parameters should probably also be put in `ProtocolParamsCardano`, in the form of some `InPairs ProtocolTransitionParams (CardanoEras c)` field. However, this requires a new type/data family, so it can probably be done in a follow-up PR. If reviewers agree, I'll open a separate issue for this.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -687,7 +742,7 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
           , topLevelConfigBlock    = blockConfigByron
           }
       , pInfoInitLedger = initExtLedgerStateByron
-      } = protocolInfoByron protocolParamsByron
+      } = protocolInfoByron paramsByron
```
jorisdral, line 745: INFO: the `protocolInfo*` functions can probably abstracted over too, into something like a `MakeProtocolInfo` class. Cardano tools already provide some form of abstraction for this, so it probably makes sense to do it here too, and then cardano tools can reuse the machinery. If reviewers agree, I'll create a separate issue for this.

jorisdral (commented):
Included some notes about follow-up issues.

jorisdral:
This PR is missing a changelog entry.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
dnadales: Makes sense!

dnadales (approved):
LGTM!

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -560,6 +560,54 @@ data ProtocolTransitionParamsShelleyBased era = ProtocolTransitionParamsShelleyB
...
+     ProtocolParams ByronBlock
+  -> ProtocolParamsShelleyBased (ShelleyEra c)
+  -> ProtocolParams (ShelleyBlock (TPraos c) (ShelleyEra c))
+  -> ProtocolParams (ShelleyBlock (TPraos c) (AllegraEra c))
+  -> ProtocolParams (ShelleyBlock (TPraos c) (MaryEra c))
+  -> ProtocolParams (ShelleyBlock (TPraos c) (AlonzoEra c))
+  -> ProtocolParams (ShelleyBlock (Praos c)  (BabbageEra c))
+  -> ProtocolParams (ShelleyBlock (Praos c)  (ConwayEra c))
+  -> CardanoProtocolParams c
+pattern CardanoProtocolParams
```
nfrisby, line 586: Would we want this pattern to use record label syntax?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: Hmm.

```
-- | Parameters needed to transition to a Shelley era.
data ProtocolTransitionParamsShelleyBased era = ProtocolTransitionParamsShelleyBased {
      transitionTranslationContext :: Core.TranslationContext era
    , transitionTrigger            :: TriggerHardFork
    }
```

Nothing about that seems particularly Shelley-specific to me.

Unless `TranslationContext` has some Shelley-specific aspect I'm not remembering.

Or unless we want to explicitly bake-in the fact that we've only ever considered the HFC in the context of Shelley eras.

(Also see my reply to your comment on versions.)

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: Yes, versions have been the easiest point of confusion. I think we definitely would benefit from constraining their structure. Like some types to ensure that the version is always increasing, perhaps?

nfrisby: In fact, I wonder if the right approach to a HFC block is to start with the details of the first era and then each subsequent era requires a "Transition" argument, which would itself include the protocol params of the new era...

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: It's not obvious to me that we'd benefit from abstracting over the way to create protocol infos for different block types. Their differences seem too "ad-hoc".

ouroboros-consensus-cardano/src/tools/Cardano/Node/Protocol/Cardano.hs
```diff
@@ -140,96 +139,98 @@ mkSomeConsensusProtocolCardano NodeByronProtocolConfiguration {
...
-          -- is in the Conway era.
-          Praos.conwayProtVer =
-            if npcTestEnableDevelopmentHardForkEras
-            then ProtVer (natVersion @9) 0  -- Advertise we can support Conway
-            else ProtVer (natVersion @8) 0, -- Otherwise we only advertise we know about Babbage
-          Praos.conwayMaxTxCapacityOverrides =
-            Mempool.mkOverrides Mempool.noOverridesMeasure
-        }
+            -- This is /not/ the Byron protocol version. It is the protocol
+            -- version that this node will use in blocks it creates. It is used
```
nfrisby, line 149: > It is the protocol version that this node will use in blocks it creates
            
I think this sentence has cause so much confusion over the years. If I recall correctly, "use in blocks it creates" means "There's a header field that will be set to this value. Nothing else about the block depends on that field's value, not even validation. It's merely for out-of-band signaling".

I wonder if we could refine our API to make this oddness much more obvious and less of a hazard.

nfrisby: Maybe that comment is more relevant to the hypothetical Issue you proposed about versions.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
jorisdral: I also don't think it looks Shelley-specific, but I also don't know the details about `TranslationContext`. Some experimentation with `data family` again:

```haskell
-- | Parameters needed to transition from one era to the next.
type ProtocolTransitionParams :: Type -> Type -> Type
data family ProtocolTransitionParams x y

-- | Parameters needed to transition from Byron to Shelley.
data instance ProtocolTransitionParams ByronBlock (ShelleyBlock proto era) =
    ProtocolTransitionParamsByronToShelley {
        transitionByronToShelleyTranslationContext :: Core.TranslationContext era
      , transitionByronToShelleyTrigger            :: TriggerHardFork
      }

-- | Parameters needed to transition from any Shelley Era to any Shelley
data instance ProtocolTransitionParams (ShelleyBlock proto era) (ShelleyBlock proto' era') =
    ProtocolTransitionParamsIntraShelley {
        transitionIntraShelleyTranslationContext :: Core.TranslationContext era'
      , transitionIntraShelleyTrigger            :: TriggerHardFork
      }

-- | Parameters needed to run Cardano.
--
-- TODO(jdral): could we get rid of 'shelleyBasedProtocolParams' somehow by
-- including it in the 'ProtocolParams' of one or multiple eras?
data instance ProtocolParams (CardanoBlock c) = ProtocolParamsCardano {
    cardanoProtocolParamsPerEra     :: PerEraProtocolParams (CardanoEras c)
  , shelleyBasedProtocolParams      :: ProtocolParamsShelleyBased (ShelleyEra c)
  , cardanoProtocolTransitionParams :: InPairs ProtocolTransitionParams (CardanoEras c)
  }
```

The nice thing about this is that we can again re-use some `sop-core` machinery: `InPairs`. But maybe we don't need the `data family`, and we can make it a single datatype with the first parameter being a phantom parameter.

Also something I was thinking of: would we want a (phantom) type parameter on `TriggerHardFork`?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
jorisdral: There are also some things that look less composable at a glance, like how `initExtLedgerStateCardano` is created, and how Byron `BlockForging` is merged with Shelley `BlockForging`.

jorisdral: Maybe we can't define a `protocolInfoHardForkBlock` for the general `HardForkBlock`, but we could try refactoring `protocolInfoCardano` a bit so it is more in the style of "make a protocol info for each individual block, and then do the ad-hoc composing"

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
jorisdral: Also, see https://github.com/input-output-hk/ouroboros-network/pull/4111#discussion_r1005815258

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
jorisdral: ^ the code is here: https://github.com/input-output-hk/ouroboros-consensus/commit/98ed9447fc35936a18d389f53d8d62f4aa196e1b

ouroboros-consensus-cardano/src/tools/Cardano/Node/Protocol/Cardano.hs
amesgen: Just recording what we briefly talked about in the Office Hours: The version of the *last* era will also be used to set the `MaxMajorProtVer`. The Shelley ledger specs (see Figure 74, function `chainChecks`) require that all blocks are considered invalid if the on-chain protocol version is larger than the `MaxMajorProtVer`, quoting:

> The constant `MaxMajorPV` provides a mechanism for halting outdated nodes. Once the
major component of the protocol version in the protocol parameters exceeds this value, every subsequent block is invalid.

In our code, since Praos, this is realised here:

https://github.com/input-output-hk/ouroboros-consensus/blob/421a7eaa0860bd9d07dfeca5390bf2ad4ad7dba8/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs#L66-L77

Somewhat confusingly, this check does *not* depend on any data from the header (`maxpv` is a constant, and `m` comes from the ledger view/state), we have https://github.com/input-output-hk/ouroboros-network/issues/4110 to improve on this.

jorisdral:
I've just pushed a commit, which also puts transition parameters inside the cardano protocol parameters. Let me know what you think @nfrisby @dnadales @amesgen

ouroboros-consensus-cardano/changelog.d/20230815_110046_joris_275_consolidate_protocolparams_across_eras.md
```diff
@@ -0,0 +1,33 @@
...
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
```
nfrisby, line 14: Huh. It hadn't occurred to me that you could have multiple of these sections. But I guess `scriv` does call them "categories", which suggests "tags", which suggests multiplicity.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: I think your last comment would be a good first step.

ouroboros-consensus-cardano/src/tools/Cardano/Node/Protocol/Cardano.hs
nfrisby: > The version of the _last_ era will also be used to set the `MaxMajorProtVer`.

The "last era" is the last "HFC era", right? Because ledger also has a notion of "intra-era" discrepancies. Just something to double-check: if ledger introduces another "ledger era" that corresponds to the same "HFC era" as the previous "ledger era", might that mean that `MaxMajorProtVer` should be greater than "the version of the last HFC era"?

nfrisby: To be more concrete:

https://github.com/input-output-hk/cardano-node/blob/4f310a8688bb96b93e4bab56501464136491d2b0/cardano-node/src/Cardano/Node/Protocol/Cardano.hs#L195-L217

Suppose the latest thing ledger had done was create the intra-era hard fork inside of Alonzo. In that case, the chain should be on major version 6, despite the version of the last HFC era (ie Mary's in this case) being 5.

nfrisby: I opened https://github.com/input-output-hk/ouroboros-consensus/issues/324 for this.

PR #285, opened 2023-08-10
Fix docusaurus' prism config.

dnadales:
We configured `prism` in two places, which overrode part of our settings.


amesgen (approved):
Can confirm that this fixes syntax highlighting locally

PR #294, opened 2023-08-15
Consolidate Cardano protocol versions

### LINKED ISSUE #276: Consolidate protocol versions in `cardano-tools`
Currently, the Cardano protocol versions are defined twice in `ouroboros-consensus-cardano-tools`:

 - For db-analyser in `Cardano.Tools.DBAnalyser.Block.Cardano`
   
   Here, e.g. `alonzoProtVer` is correctly set to `7` since input-output-hk/ouroboros-network#4377 (note that, somewhat confusingly, this is *not* the Alonzo protocol version, see [this comment](https://github.com/input-output-hk/cardano-node/blob/363fc0ff6e1c90a0f20dca05de08211ea4b661ef/cardano-node/src/Cardano/Node/Protocol/Cardano.hs#L180-L183)):
   https://github.com/input-output-hk/ouroboros-network/blob/eddc225de3cb7a5e803ccf83e32242fa3e131e55/ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Block/Cardano.hs#L329

 - For db-synthesizer in `Cardano.Node.Protocol.Cardano`
   
   Here, e.g. `alonzoProtVer` is incorrect:
   https://github.com/input-output-hk/ouroboros-network/blob/eddc225de3cb7a5e803ccf83e32242fa3e131e55/ouroboros-consensus-cardano-tools/src/Cardano/Node/Protocol/Cardano.hs#L210

The goal of this ticket is to correct these discrepancies by consolidating them. This can serve as the baseline to extract this into a library that can also be reused in the node [here](https://github.com/input-output-hk/cardano-node/blob/363fc0ff6e1c90a0f20dca05de08211ea4b661ef/cardano-node/src/Cardano/Node/Protocol/Cardano.hs), cf. input-output-hk/ouroboros-network#3925.

---

For your convenience: See [this table from CIP-0059](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0059/feature-table.md) for an overview of the different protocol versions.

jorisdral:
# Description

Closes #276.

Currently, protocol versions for Cardano eras are redefined in a number of places, among which `cardano-node`, `db-analyser` and `db-synthesiser`. This is not ideal, since mistakes often creep into the code that redefines these versions (see the description of #276).

This PR provides the `Ouroboros.Consensus.Cardano.Node.ProtocolVersions` module, where protocol versions for each of the Cardano eras are consolidated in a single `CardanoProtocolVersions` datatype. `cardanoMainnetProtocolVersions` is a constant that defines the protocol versions of the Cardano eras on mainnet. `cardanoMaxProtocolVersion` can be used to extract the maximum protocol version across the Cardano eras (equivalently, it is the protocol version of the last era), which is information that is required by the `TPraos` and `Praos` protocols.

Since protocol versions are currently defined as part of `ProtocolParams` datatypes for each era, downstream code would have to pattern match on the `CardanoProtocolVersions` datatype, retrieving protocol versions for each era, and filling in the protocol versions in the `ProtocolParams` for each era. To facilitate easily using `CardanoProtocolVersions`, the `Ouroboros.Consensus.Cardano.Node.HasPartialProtocolParams` module provides a `HasPartialProtocolParams` class: instances of this class can create a `ProtocolParams` datatype from a `ProtocolVersion` and `PartialProtocolParams` (i.e., a `ProtocolParams` without any `ProtocolVersion`). As such, partial protocol parameters for Cardano and protocol versions for each era can be passed in separately, and combined using the `completeProtocolParams` function. `protocolInfoCardano'` is a variant of `protocolInfoCardano` that automates this. A usage example of `protocolInfoCardano'` can be found in the `Cardano.Tools.DBAnalyser.Block.Cardano` module.

### Todos
- [ ] Rename commits
- [ ] Add changelog fragments
- [ ] Re-evaluate public exports. What should marked as internal?
- [ ] Re-evaluate module hierarchy. Can some functionality be moved to `ouroboros-consensus`?

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/HasPartialProtocolParams.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/HasPartialProtocolParams.hs
@@ -1,0 +1,114 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeFamilies      #-}
+
+-- | Complete protocol parameters from partial protocol parameters.
+module Ouroboros.Consensus.Cardano.Node.HasPartialProtocolParams (
+    HasPartialProtocolParams (..)
+  , PartialProtocolParams (..)
+  ) where
+
+import qualified Cardano.Chain.Genesis as Genesis
+import qualified Cardano.Chain.Update as Update
+import           Data.Kind (Type)
+import           Ouroboros.Consensus.Byron.Ledger.Block (ByronBlock)
+import           Ouroboros.Consensus.Byron.Node (ByronLeaderCredentials,
+                     PBftSignatureThreshold, ProtocolParams (..))
+import           Ouroboros.Consensus.Cardano.Node.ProtocolVersions
+                     (ProtocolVersion)
+import qualified Ouroboros.Consensus.Mempool.Capacity as Mempool
+import           Ouroboros.Consensus.Protocol.Praos (Praos)
+import           Ouroboros.Consensus.Protocol.TPraos (TPraos)
+import           Ouroboros.Consensus.Shelley.Eras (AllegraEra, AlonzoEra,
+                     BabbageEra, ConwayEra, MaryEra, ShelleyEra)
+import           Ouroboros.Consensus.Shelley.Ledger.Block (ShelleyBlock)
+import           Ouroboros.Consensus.Shelley.Node (ProtocolParams (..))
+
+{-------------------------------------------------------------------------------
+  Class
+-------------------------------------------------------------------------------}
+
+class HasPartialProtocolParams blk where
+  data PartialProtocolParams blk :: Type
+
+  completeProtocolParams ::
+       ProtocolVersion blk
+    -> PartialProtocolParams blk
+    -> ProtocolParams blk
+
+{-------------------------------------------------------------------------------
+  Instances
+-------------------------------------------------------------------------------}
+
+instance HasPartialProtocolParams ByronBlock where
+  data instance PartialProtocolParams ByronBlock =
+    PartialProtocolParamsByron {
+        pbyronGenesis                :: Genesis.Config
+      , pbyronPbftSignatureThreshold :: Maybe PBftSignatureThreshold
+      , pbyronSoftwareVersion        :: Update.SoftwareVersion
+      , pbyronLeaderCredentials      :: Maybe ByronLeaderCredentials
+      , pbyronMaxTxCapacityOverrides :: Mempool.TxOverrides ByronBlock
+      }
+
+  completeProtocolParams byronProtVer partialParamsByron =
+      ProtocolParamsByron {
+          byronGenesis = pbyronGenesis partialParamsByron
+        , byronPbftSignatureThreshold = pbyronPbftSignatureThreshold partialParamsByron
+        , byronProtocolVersion = byronProtVer
+        , byronSoftwareVersion = pbyronSoftwareVersion partialParamsByron
+        , byronLeaderCredentials = pbyronLeaderCredentials partialParamsByron
+        , byronMaxTxCapacityOverrides = pbyronMaxTxCapacityOverrides partialParamsByron
+        }
+
+instance HasPartialProtocolParams (ShelleyBlock (TPraos c) (ShelleyEra c)) where
+  newtype instance PartialProtocolParams (ShelleyBlock (TPraos c) (ShelleyEra c)) =
+    PartialProtocolParamsShelley {
+        pshelleyMaxTxCapacityOverrides :: Mempool.TxOverrides (ShelleyBlock (TPraos c) (ShelleyEra c))
+      }
+
+  completeProtocolParams protVer (PartialProtocolParamsShelley txOverrides) =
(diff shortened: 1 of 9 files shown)


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/ProtocolVersions/Internal.hs
```diff
@@ -0,0 +1,187 @@
...
+      , protVerShelley = SL.ProtVer (SL.natVersion @3) 0
+      , protVerAllegra = SL.ProtVer (SL.natVersion @4) 0
+      , protVerMary    = SL.ProtVer (SL.natVersion @5) 0
+      , protVerAlonzo  = SL.ProtVer (SL.natVersion @7) 0
+      , protVerBabbage = if experimentalHardForksEnabled then
+                           SL.ProtVer (SL.natVersion @9) 0
+                         else
+                           SL.ProtVer (SL.natVersion @8) 0
+      , protVerConway  = SL.ProtVer (SL.natVersion @9) 0
+      }
```
amesgen, line 101: What about adding some documentation here what these numbers mean and where they are coming from?

One step could be to define these using the relevant ledger functions:
```suggestion
cardanoMainnetProtocolVersions :: Bool -> CardanoProtocolVersions c
cardanoMainnetProtocolVersions experimentalHardForksEnabled =
    CardanoProtocolVersions {
        protVerByron   = Byron.Update.ProtocolVersion
                           (SL.getVersion (eraProtVerLow @Shelley))
                           0 0
      , protVerShelley = SL.ProtVer (eraProtVerLow @Allegra) 0
      , protVerAllegra = SL.ProtVer (eraProtVerLow @Mary) 0
      , protVerMary    = SL.ProtVer (eraProtVerLow @Alonzo) 0
      , protVerAlonzo  = SL.ProtVer (eraProtVerLow @Babbage) 0
      , protVerBabbage = if experimentalHardForksEnabled then
                           SL.ProtVer (eraProtVerLow @Conway) 0
                         else
                           SL.ProtVer (eraProtVerHigh @Babbage) 0
      , protVerConway  = SL.ProtVer (eraProtVerHigh @Conway) 0
      }
```
But I think it is still useful to see what these numbers are explicitly, so if we do this, we could copy the current definition just below and have a test that asserts that they are the same. WDYT?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/ProtocolVersions/Internal.hs
```diff
@@ -0,0 +1,187 @@
...
+
+instance HasProtocolVersion (ShelleyBlock (Praos c) (ConwayEra c)) where
+  extractProtocolVersion = conwayProtVer
+
+{-------------------------------------------------------------------------------
+  Value for /each/ era
+-------------------------------------------------------------------------------}
+
+type PerEraProtocolVersion :: [Type] -> Type
+newtype PerEraProtocolVersion xs = PerEraProtocolVersion {
```
amesgen, line 153: Is this intermediate type buying us something here, or is it just there for consistency with other "per-era" `NP`s?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/HasPartialProtocolParams.hs
```diff
@@ -0,0 +1,114 @@
...
+                     BabbageEra, ConwayEra, MaryEra, ShelleyEra)
+import           Ouroboros.Consensus.Shelley.Ledger.Block (ShelleyBlock)
+import           Ouroboros.Consensus.Shelley.Node (ProtocolParams (..))
+
+{-------------------------------------------------------------------------------
+  Class
+-------------------------------------------------------------------------------}
+
+class HasPartialProtocolParams blk where
+  data PartialProtocolParams blk :: Type
```
amesgen, line 31: What is nice about this is that we don't have to change the era-specific `ProtocolParams*` types at all, but eventually, it seems less ad-hoc/boilerplate-y to remove the protocol versions from `ProtocolParams*` and to instead e.g. have some `WithProtocolVersion` wrapper type (the eg all of the instances below would become unnecessary). Maybe another thing for #319, or even for this PR if it is not too invasive? :smile:

amesgen (commented):
Very nice, downstream will love this!

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node/ProtocolVersions/Internal.hs
nfrisby: > we could copy the current definition just below and have a test that asserts that they are the same.

If I recall correctly, ledger defines these values at the type-level. So you could design it to incur a type error instead of a test error when the indirected definition and the concrete definition disagree.

dnadales:
This might be superseded by the work on https://github.com/IntersectMBO/ouroboros-consensus/issues/324

jorisdral:
Superseded by #1224

PR #295, opened 2023-08-15
Allow to build `db-analyser` without asserts via Nix

amesgen:
See the commit description for more details.

TL;DR: You can now get a `db-analyser` without asserts via

    nix build .#db-analyser


asserts.cabal
```diff
@@ -0,0 +1,55 @@
+--
```
dnadales, line 1: I wonder if it'd make sense to write a comment about why we have this file.

dnadales (approved):

asserts.cabal
amesgen: Will do :+1:

PR #305, opened 2023-08-21
Avoid trying to remove empty lists of transactions

jasagredo:
# Description

By increasing the opcert twice on a testing cluster, we found out that this code path could be executed with an empty list of transactions leading to an exception. This change make sure we only try to remove transactions if there are any in the block, and in fact, only of the error of the block was due to the body.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -68,6 +68,7 @@ import           Ouroboros.Consensus.Storage.LedgerDB.API
 import           Ouroboros.Consensus.Storage.LedgerDB.BackingStore
 import           Ouroboros.Consensus.Storage.LedgerDB.DbChangelog
 import qualified Ouroboros.Consensus.Storage.LedgerDB.DbChangelog.Query as LedgerDB
+import           Ouroboros.Consensus.Util (whenJust)
 import           Ouroboros.Consensus.Util.EarlyExit
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Orphans ()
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -451,15 +451,25 @@ forkBlockForging IS{..} blockForging =
...
+            -- will remove all the transactions in that block from the mempool
+            -- as a defensive programming measure. Otherwise we'll run the risk
+            -- of forging the same invalid block again. This means that we'll
+            -- throw away some good transactions in the process.
+            case reason of
+              ChainDB.ValidationError (ExtValidationErrorLedger _) ->
+                maybe
+                  (pure ())
+                  (lift . removeTxs mempool)
+                  $ NE.nonEmpty (map (txId . txForgetValidated) txs)
```
amesgen, line 471:
```suggestion
                whenJust
                  (NE.nonEmpty (map (txId . txForgetValidated) txs))
                  (lift . removeTxs mempool)
```

amesgen (approved):

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
jasagredo: Done

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -451,15 +452,26 @@ forkBlockForging IS{..} blockForging =
...
+            -- will remove all the transactions in that block from the mempool
+            -- as a defensive programming measure. Otherwise we'll run the risk
+            -- of forging the same invalid block again. This means that we'll
+            -- throw away some good transactions in the process.
+            case reason of
+              ChainDB.ValidationError (ExtValidationErrorLedger _) ->
+                whenJust
+                  (NE.nonEmpty (map (txId . txForgetValidated) txs))
+                  (lift . removeTxs mempool)
+              _ -> pure ()
```
nfrisby, line 474:
```suggestion
              ChainDB.InFutureExceedsClockSkew{} -> pure ()
              ChainDB.ValidationError err -> case err of
                ExtValidationErrorHeader{} -> pure ()
                ExtValidationErrorLedger{} ->
                  whenJust
                    (NE.nonEmpty (map (txId . txForgetValidated) txs))
                    (lift . removeTxs mempool)
```

nfrisby (approved):
I Approved, but I also made a suggestion that I feel somewhat strongly about.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
jasagredo: Done

PR #304, opened 2023-08-21
Fixup the mid August releases

nfrisby:
- Removed the `expose-sublibs` cabal flag, since Cabal/Nix handled it poorly.
- Instead, added a `unstable-` prefix to the name of each sublibrary, to strongly indicate that we ignore them when evolving the package's version.

PR #308, opened 2023-08-21
Patch the latest releases

nfrisby:
The recent changes to our `expose-sublibs` cabal flag prevented a CHaP release.

Instead of embellishing CHaP, we've simplified our packaging, removing the flag. See PR #304.


amesgen (approved):

nfrisby:
The unpatched versions will never be on CHaP. In, hypothetically, they were, then we really should be bumping the lower-bounds here. So, if they ever do get released (accidentally?), we could consider a revision.

I'm skipping this step because it's arduous to find all the necessary places, it's not obviously necessary (see above paragraph), and I'm trying to unblock downstream teams with this PR.

PR #312, opened 2023-08-22
Fixup tags in create-release.hs

nfrisby:
The first commit adds the `release-` prefix to the tag names the tool makes.

The second commit removes the ability to make tags from the script. As-is, this script tags the commits that it creates. However, we don't want to tag those commits, we want to instead tag the merge of the PR that adds those commits to the `origin/main`.


jasagredo (approved):

fraser-iohk:
yep, more than happy to make this change -- I didn't use this when I did the last release, so no objection to removing it at all

fraser-iohk (approved):

PR #321, opened 2023-08-24
Move SOP modules to a separate package

jasagredo:
# Description

Extract the SOP modules to a separate package, as a preparation step to move them to a separate repository.


sop-extras/src/Data/SOP/Counting.hs
```diff
@@ -68,64 +54,31 @@ import           Data.SOP.Strict
 newtype Exactly xs a = Exactly { getExactly :: NP (K a) xs }
 
 -- | At most one value for each type level index
-data AtMost :: [Type] -> Type -> Type where
+type AtMost :: [Type] -> Type -> Type
```
amesgen, line 57: Do we get type inference problems when this is polykinded?

sop-extras/src/Data/SOP/Counting.hs
```diff
@@ -68,64 +54,31 @@ import           Data.SOP.Strict
...
+data AtMost xs a where
```
amesgen, line 58: About naming: I find the name and documentation counterintuitive, ie "at most one value for each type level index" seems to imply for me that if we have
```haskell
xs ~ [A, B, C, D]
```
then e.g. the following five values would be possible inhabitants of the type `AtMost xs Int`:
```haskell
[A, B, C, D]
 -  -  -  -
 0  -  1  -
 -  -  -  5
 -  1  2  3
 8  9  -  -
```
But actually, only the first and last one can be represtend, namely via `AtMostNil` and `AtMostCons 8 (AtMostCons 9 AtMostNil)`, if we have no value for e.g. the index `B`, then we can also have no values for `C` and `D`.

A more intuitive name for me would be `Prefix`. WDYT? Is this a sufficient reason to change the name?

sop-extras/src/Data/SOP/NonEmpty.hs
```diff
@@ -0,0 +1,175 @@
...
+import           Data.Proxy
+import           Data.SOP.Sing
+
+{-------------------------------------------------------------------------------
+  NonEmpty
+-------------------------------------------------------------------------------}
+
+-- | Non-empty variation on 'AtMost'
+type NonEmpty :: [Type] -> Type -> Type
+data NonEmpty xs a where
```
amesgen, line 44: Similar comments as for `AtMost` apply, an IMO more fitting name would be `NonEmptyPrefix`.

Another observation: `NonEmpty xs a` is isomorphic to `Telescope (K a) (K a) xs`/`SimpleTelescope (K a) xs`, but I am not sure if that should inform our choice of name (`VerySimpleTelescope` doesn't sound very helpful).

sop-extras/src/Data/SOP/NonEmpty.hs
```diff
@@ -0,0 +1,175 @@
...
+nonEmptyInit (NonEmptyOne  x)    = (Nothing, x)
+nonEmptyInit (NonEmptyCons x xs) =
+    case nonEmptyInit xs of
+      (Nothing  , final) -> (Just (NonEmptyOne  x)     , final)
+      (Just xs' , final) -> (Just (NonEmptyCons x xs') , final)
+
+-- | Build a 'NonEmpty' from a list. Returns 'Nothing' when the list is empty
+-- or when it's longer than @xs@.
+nonEmptyFromList :: forall xs a. SListI xs => [a] -> Maybe (NonEmpty xs a)
+nonEmptyFromList = go (sList @xs)
```
jorisdral, line 94: A total version of this function, i.e. something returning just `NonEmpty xs a` seems straightforward, since we have the `IsNonEmpty` class. 

Also something for a later PR.

sop-extras/src/Data/SOP/Counting.hs
jorisdral: Looks to me like it would be fine if `xs` were a list of polykinds

strict-sop-core/strict-sop-core.cabal
```diff
@@ -0,0 +1,51 @@
+cabal-version:   3.0
+name:            strict-sop-core
+synopsis:        Strict replacement for NS and NP.
+description:     Strict replacement for NS and NP from @sop-core@.
+version:         0.1.0.0
+license:         Apache-2.0
+license-file:    LICENSE
```
jorisdral, line 7: A `NOTICE` file should be added and referenced here

strict-sop-core/strict-sop-core.cabal
```diff
@@ -0,0 +1,51 @@
...
+    Data.SOP.Strict.NP
+    Data.SOP.Strict.NS
+
+  build-depends:
+    , base      >=4.14 && <4.19
+    , nothunks  ^>=0.1
+    , sop-core  ^>=0.5
+
+  default-language:   Haskell2010
+  default-extensions:
```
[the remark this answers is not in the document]
jasagredo: I forgot to remove this

sop-extras/src/Data/SOP/NonEmpty.hs
jasagredo: Phew that type :sweat_smile: 

In any case, I think we are in agreement about:
- `AtMost` -> `Prefix`
- `NonEmpty` -> `NonEmptyPrefix`

But I shall do (and we can discuss further) these renamings in a separate PR.

sop-extras/src/Data/SOP/Telescope.hs
```diff
@@ -69,14 +74,6 @@ import           Prelude hiding (scanl, sequence, zipWith)
 -- A telescope is an extension of an 'NS', where every time we "go right" in the
 -- sum we have an additional value.
 --
--- Blockchain intuition: think of @g@ as representing some kind of past state,
--- and @f@ some kind of current state. Then depending on how many hard fork
--- transitions we have had, we might either have, say
---
--- > TZ currentByronState
--- > TS pastByronState $ TZ currentShelleyState
--- > TS pastByronState $ TS pastShelleyState $ TZ currentGoguenState
```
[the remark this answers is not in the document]
jasagredo: I thought of adding them to the `HardForkState` definition, which uses a telescope

```
newtype HardForkState f xs = HardForkState {
      getHardForkState :: Telescope (K Past) (Current f) xs
    }
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
```diff
@@ -37,6 +37,19 @@ import           Prelude
...
 --
 -- This is used both for the consensus state and the ledger state.
+--
+-- By using a telescope, we will keep track of 'Past' information for eras
+-- before the current one:
+--
+-- > TZ currentByronState
+-- > TZ pastByronState $ TZ currentShelleyState
+-- > TZ pastByronState $ TS pastShelleyState $ TZ currentAllegraState
+-- > ...
```
jorisdral, line 47: We should explain that we take `f ~ LedgerState` as an example here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
```diff
@@ -37,6 +37,19 @@ import           Prelude
...
+--
+-- Using 'Telescope.extend' and 'Telescope.retract' we are able to advance
+-- (_translate_) a @LedgerState@ from one era to the next/previous one. Using
+-- 'Telescope.align' or 'Telescope.alignExtend', we will also put in sync
+-- multiple 'Telescope's.
```
jorisdral, line 52: I'd maybe copy the blockchain intuitions for `extend`, `retract` and `align` literally, and put them in sections. Something like:

```
-- === extend
--
-- ...
--
-- === retract
--
-- ...
--
-- === align
--
-- ...
```

sop-extras/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2023 Input Output (Hong Kong) Ltd.
```
jorisdral, line 1:
```suggestion
Copyright 2023 Input Output Global Inc (IOG).
```

strict-sop-core/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2023 Input Output (Hong Kong) Ltd.
```
jorisdral, line 1:
```suggestion
Copyright 2023 Input Output Global Inc (IOG).
```

```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
jasagredo: Done

jorisdral (approved):

PR #329, opened 2023-08-29
Isolate the backing-store tracing

jasagredo:
# Description

Create a new tracer to provide for the backing store events.


mgmeier (approved):
Thank you @jasagredo

PR #340, opened 2023-09-07
First tick across the era boundary when translating the ledger/chain-dep state

### LINKED ISSUE #339: Change implementation of ticking in the HFC to tick-then-translate-then-tick
### Status quo

Currently, when ticking a ledger state across an era boundary in the HFC, we use the **translate-then-tick** strategy.

<details>
<summary>Click to see current implementation details</summary>

HFC ticking is implemented here:
https://github.com/input-output-hk/ouroboros-consensus/blob/0ca9ca08f41b04619dc9ed1df692102ef9ba3c0e/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs#L121-L154

Note that we first translate in `extendToSlot` if the target slot is in a new era:
https://github.com/input-output-hk/ouroboros-consensus/blob/0ca9ca08f41b04619dc9ed1df692102ef9ba3c0e/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State.hs#L212
https://github.com/input-output-hk/ouroboros-consensus/blob/0ca9ca08f41b04619dc9ed1df692102ef9ba3c0e/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State.hs#L226-L228

</details>

This means that when we tick a ledger state `s` in era `FromEra` (which arose by applying a block with slot `x`) to a slot `y` in the next era `ToEra`, we

 - first translate the `FromEra` ledger state `s` to a `ToEra` ledger state `s'` (`s'` is now a `FromEra` ledger state, even though its tip slot `x` hasn't moved, so is still *before* the era boundary), and
 - then tick `s'` to slot `y`,

yielding the ledger state `s'''` we can use to validate blocks in slot `y`.

In the following diagram, this means that we start at `s`, first go right and then down.
```
╔══════════════╦═════════════════════════════════════╗
║ Time \ Era   ║ FromEra ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌⇢ ToEra ║
╠══════════════╬═════════════════════════════════════╣
║  Start - x   ║     s ──── translate ────────→ s'   ║
║      ┊       ║     │╲                         │    ║
║      ┊       ║  tick ╲tick                    tick ║
║   Boundary ╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌ ║
║      ┊       ║     │   s' ── translate ─→ s'' │    ║
║      ┊       ║     │                      ╲   │    ║
║      ┊       ║     │                   tick╲  │    ║
║      ┊       ║     │                        ╲ │    ║
║      ⇣       ║     ↓                         ↘↓    ║
║  Target - y  ║     s' ─── translate ────────→ s''' ║
╚══════════════╩═════════════════════════════════════╝
```

### Why the status quo is problematic

The current behavior is causing https://github.com/input-output-hk/cardano-ledger/issues/3491; we need some context how Ledger handles updates to the on-chain protocol parameters before Voltaire (ie in all Shelley-based eras before Conway):

 - At every epoch boundary, the protocol parameters can change if sufficiently many Genesis keys submit the same update proposal.
 - This logic is handled by the [UPEC rule](https://github.com/input-output-hk/cardano-ledger/blob/a0db238724037e4875516ab87f0d1a1b61ce3ec0/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Upec.hs), which is executed when eg ticking.

This mechanism changed completely in Conway; in particular, there is no direct analogue to update proposals signed by Genesis keys. This means that all such proposals are [*discarded*](https://github.com/input-output-hk/cardano-ledger/blob/c7b6ce12a869838eefe18d72420cbf7f965fa9ee/eras/conway/impl/src/Cardano/Ledger/Conway/Translation.hs#L157-L170) when translating a Babbage ledger state to a Conway ledger state.

Hence, as we currently first translate and then tick, the Conway logic responsible for ticking has no way of knowing that it should update the protocol version (or any other updateable parameter).

### Proposed change

In this issue, we suggest to change the HFC cross-era ticking to the **tick-then-translate-then-tick** approach, ie first tick to the epoch boundary, then translate the ledger state, and then tick to the requested slot.

In more detail: In order to tick a ledger state `s` (with tip `x`) in era `FromEra` to a slot `y` in the next era `ToEra`, we

 - first tick the ledger state `s` to the first slot in `FromEra`, yielding a ledger state `s'`,
 - then translate the `FromEra` ledger state `s'` to a `ToEra` ledger state `s''`, and
 - finally tick the ledger state `s''` to slot `y`

again yielding the desired ledger state `s'''`.

In the diagram above, we again start at `s`, tick by moving diagonally to `s'`, then translate by moving horizontally to `s''`, and finally tick again by moving diagonally to `s'''`.

This way, we use the `FromEra` logic to tick across the era/epoch boundary, which can then execute the UPEC rule in the example above.

Slogan: Cross-era ticking is fundamentally something that happens at the *end* of an era, so it should be done using the logic of that ending era.

### Alternatives

We considered the following alternatives:

 1. Use the **tick-then-translate** apprach, ie tick directly to the target slot, and then translate.
   
    While this would probably work fine with existing any likely also with future eras, it seems wrong to use the ticking logic of the old era to let time pass across slots that lie purely within the new era. Also, we again have (just before translating) a ledger state in the "wrong" era as with the current **translate-then-tick** approach.

 2. Don't change the HFC logic at all, but rather introduce an ad-hoc field in the Conway ledger state that records the Babbage update proposals, such that they can be preserved by translating.

    This seems quite ugly, ie this field would only be present in the first Voltaire era, and it would require ledger rule logic that is purely related to era transitions, which is something that the ledger rules usually do not have to handle explicitly.

### Remarks

Note that the HFC mechanism that detects whether we should transition in the first place (see `singleEraTransition`) is not affected by this bug; we still properly transition from Babbage to Conway.

amesgen:
Closes #339, closes https://github.com/input-output-hk/cardano-ledger/issues/3491

This PR is intended to be reviewed commit by commit, which should all compile and pass all tests individually.

On a high level, this PR changes the way we do cross-era ticking in the HFC from **translate-then-tick** to **tick-then-translate-then-tick** for both the ledger state and, for consistency, also for the chain-dep/protocol state. For motivation, see #339, for further directions in this area, see #345.

The first commit enriches an existing test to showcase the problem, it is then fixed in a later commit.

TODO:

 - [ ] Determine whether we need more test, in particular for `ChainDepState` cross-era ticking. We could do this using actual Cardano blocks.
 - [ ] Do we need more documentation? (Maybe a GHC-style note explaining the overall HFC cross-era ticking approach?)

### DIFF
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
@@ -89,27 +89,30 @@ import           Ouroboros.Consensus.Util (ShowProxy (..), (..:))
 -------------------------------------------------------------------------------}
 
 data instance LedgerState ByronBlock = ByronLedgerState {
-      byronLedgerTipBlockNo :: !(WithOrigin BlockNo)
-    , byronLedgerState      :: !CC.ChainValidationState
+      byronLedgerState      :: !CC.ChainValidationState
     , byronLedgerTransition :: !ByronTransition
     }
   deriving (Eq, Show, Generic, NoThunks)
 
 -- | Information required to determine the transition from Byron to Shelley
-data ByronTransition =
-    -- | Per candidate proposal, the 'BlockNo' in which it became a candidate
-    --
-    -- The HFC needs to know when a candidate proposal becomes stable. We cannot
-    -- reliably do this using 'SlotNo': doing so would mean that if we were to
-    -- switch to a denser fork, something that was previously deemed stable is
-    -- suddenly not deemed stable anymore (although in actuality it still is).
-    -- We therefore must do this based on 'BlockNo' instead, but unfortunately
-    -- the Byron ledger does not record this information. Therefore, we record
-    -- it here instead.
-    --
-    -- Invariant: the domain of this map should equal the set of candidate
-    -- proposals.
-    ByronTransitionInfo !(Map Update.ProtocolVersion BlockNo)
+data ByronTransition = ByronTransitionInfo {
+      -- | Per candidate proposal, the 'BlockNo' in which it became a candidate
+      --
+      -- The HFC needs to know when a candidate proposal becomes stable. We
+      -- cannot reliably do this using 'SlotNo': doing so would mean that if we
+      -- were to switch to a denser fork, something that was previously deemed
+      -- stable is suddenly not deemed stable anymore (although in actuality it
+      -- still is). We therefore must do this based on 'BlockNo' instead, but
+      -- unfortunately the Byron ledger does not record this information.
+      -- Therefore, we record it here instead.
+      --
+      -- Invariant: the domain of this map should equal the set of candidate
+      -- proposals.
+      perCandidateProposals :: !(Map Update.ProtocolVersion BlockNo)
+    , -- | Block number of the last applied block, used to compute when a
+      -- proposal became stable.
+      byronLedgerTipBlockNo :: !(WithOrigin BlockNo)
+    }
   deriving (Eq, Show, Generic, NoThunks)
 
 instance UpdateLedger ByronBlock
@@ -121,8 +124,7 @@ initByronLedgerState :: Gen.Config
                      -> LedgerState ByronBlock
 initByronLedgerState genesis mUtxo = ByronLedgerState {
       byronLedgerState      = override mUtxo initState
-    , byronLedgerTipBlockNo = Origin
-    , byronLedgerTransition = ByronTransitionInfo Map.empty
+    , byronLedgerTransition = ByronTransitionInfo Map.empty Origin
     }
   where
     initState :: CC.ChainValidationState
@@ -249,7 +251,7 @@ instance LedgerSupportsProtocol ByronBlock where
   --
   -- To create a forecast, take the delegation state from the given ledger
   -- state, and apply the updates that should be applied by the given slot.
-  ledgerViewForecastAt cfg (ByronLedgerState _tipBlkNo st _) = Forecast at $ \for ->
+  ledgerViewForecastAt cfg (ByronLedgerState st _) = Forecast at $ \for ->
       toTickedPBftLedgerView <$> if
         | for == lastSlot ->
           return $ CC.getDelegationMap st
@@ -358,19 +360,23 @@ applyABlock validationMode cfg blk blkHash blkNo TickedByronLedgerState{..} = do
             aux candidate = (UPE.cpuProtocolVersion candidate, blkNo)
 
         transition' :: ByronTransition
-        transition' =
-            case untickedByronLedgerTransition of
-              ByronTransitionInfo oldEntries -> ByronTransitionInfo $
+        transition' = ByronTransitionInfo {
+              perCandidateProposals =
                 -- Candidates that have /just/ become candidates
                 let newEntries :: Map Update.ProtocolVersion BlockNo
                     newEntries = ifNew `Map.difference` oldEntries
 
                 -- Remove any entries that aren't candidates anymore
                 in (oldEntries `Map.intersection` ifNew) `Map.union` newEntries
+            , byronLedgerTipBlockNo = NotOrigin blkNo
+            }
+          where
+            ByronTransitionInfo {
+                perCandidateProposals = oldEntries
+              } = untickedByronLedgerTransition
 
     return ByronLedgerState {
-          byronLedgerTipBlockNo = NotOrigin blkNo
-        , byronLedgerState      = st'
+          byronLedgerState      = st'
         , byronLedgerTransition = transition'
         }
 
@@ -386,9 +392,10 @@ applyABoundaryBlock :: Gen.Config
 applyABoundaryBlock cfg blk blkNo TickedByronLedgerState{..} = do
     st' <- CC.validateBoundary cfg blk tickedByronLedgerState
     return ByronLedgerState {
-        byronLedgerTipBlockNo = NotOrigin blkNo
-      , byronLedgerState      = st'
-      , byronLedgerTransition = untickedByronLedgerTransition
+        byronLedgerState      = st'
+      , byronLedgerTransition = untickedByronLedgerTransition {
+            byronLedgerTipBlockNo = NotOrigin blkNo
+          }
       }
 
 {-------------------------------------------------------------------------------
@@ -412,7 +419,7 @@ encodeByronHeaderState = encodeHeaderState
     encodeByronChainDepState
     encodeByronAnnTip
 
--- | Encode transition info
+-- | Encode candidate proposals (in 'ByronTransition').
 --
 -- We encode the absence of any info separately. This gives us a bit more
 -- wiggle room to change our mind about what we store in snapshots, as they
@@ -422,8 +429,8 @@ encodeByronHeaderState = encodeHeaderState
 -- inclusion of a list length. We didn't, so the decoder is a bit awkward :/
 --
 -- TODO: If we break compatibility anyway, we might decide to clean this up.
-encodeByronTransition :: ByronTransition -> Encoding
-encodeByronTransition (ByronTransitionInfo bNos)
+encodeByronCandidateProposals :: Map Update.ProtocolVersion BlockNo -> Encoding
+encodeByronCandidateProposals bNos
   | Map.null bNos = CBOR.encodeWord8 0
   | otherwise     =
          CBOR.encodeListLen (fromIntegral (Map.size bNos))
@@ -438,13 +445,13 @@ encodeByronTransition (ByronTransitionInfo bNos)
         , encode bno
         ]
 
--- | Decode Byron transition info
+-- | Decode candidate proposals (in 'ByronTransition')
 --
 -- See comments for 'encodeByronTransition'.
-decodeByronTransition :: Decoder s ByronTransition
-decodeByronTransition = do
+decodeByronCandidateProposals :: Decoder s (Map Update.ProtocolVersion BlockNo)
+decodeByronCandidateProposals = do
     ttype <- CBOR.peekTokenType
-    fmap ByronTransitionInfo $ case ttype of
+    case ttype of
       CBOR.TypeUInt -> do
         tag <- CBOR.decodeWord8
         case tag of
@@ -470,16 +477,19 @@ encodeByronLedgerState ByronLedgerState{..} = mconcat [
       encodeListLen 3
     , encode byronLedgerTipBlockNo
     , encode byronLedgerState
-    , encodeByronTransition byronLedgerTransition
+    , encodeByronCandidateProposals perCandidateProposals
     ]
+  where
+    ByronTransitionInfo{..} = byronLedgerTransition
 
 decodeByronLedgerState :: Decoder s (LedgerState ByronBlock)
 decodeByronLedgerState = do
     enforceSize "ByronLedgerState" 3
-    ByronLedgerState
-      <$> decode
-      <*> decode
-      <*> decodeByronTransition
+    tipBlockNo         <- decode
+    ledgerState        <- decode
+    candidateProposals <- decodeByronCandidateProposals
+    let transitionInfo = ByronTransitionInfo candidateProposals tipBlockNo
+    pure $ ByronLedgerState ledgerState transitionInfo
 
 encodeByronQuery :: BlockQuery ByronBlock result -> Encoding
 encodeByronQuery query = case query of
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -22,6 +22,7 @@ module Ouroboros.Consensus.Cardano.CanHardFork (
   , ShelleyPartialLedgerConfig (..)
   , forecastAcrossShelley
   , translateChainDepStateAcrossShelley
+  , translateLedgerStateAcrossShelley
   ) where
 
 import qualified Cardano.Chain.Common as CC
@@ -273,11 +274,11 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
   hardForkEraTranslation = EraTranslation {
       translateLedgerState   =
           PCons translateLedgerStateByronToShelleyWrapper
-        $ PCons translateLedgerStateShelleyToAllegraWrapper
-        $ PCons translateLedgerStateAllegraToMaryWrapper
-        $ PCons translateLedgerStateMaryToAlonzoWrapper
-        $ PCons translateLedgerStateAlonzoToBabbageWrapper
-        $ PCons translateLedgerStateBabbageToConwayWrapper
+        $ PCons translateLedgerStateAcrossShelley
+        $ PCons translateLedgerStateAcrossShelley
+        $ PCons translateLedgerStateAcrossShelley
+        $ PCons translateLedgerStateAcrossShelley
+        $ PCons translateLedgerStateAcrossShelley
         $ PNil
     , translateChainDepState =
           PCons translateChainDepStateByronToShelleyWrapper
@@ -313,8 +314,8 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
                     translateTxAllegraToMaryWrapper
                     translateValidatedTxAllegraToMaryWrapper
               )
-      $ PCons (RequireBoth $ \_cfgMary cfgAlonzo ->
-                let ctxt = getAlonzoTranslationContext cfgAlonzo
+      $ PCons (RequireBoth $ \_cfgMary (WrapLedgerConfig cfgAlonzo) ->
+                let ctxt = shelleyLedgerTranslationContext cfgAlonzo
                 in
                 Pair2
                   (translateTxMaryToAlonzoWrapper          ctxt)
@@ -327,8 +328,8 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
                   (translateTxAlonzoToBabbageWrapper          ctxt)
                   (translateValidatedTxAlonzoToBabbageWrapper ctxt)
               )
-      $ PCons (RequireBoth $ \_cfgBabbage cfgConway ->
-                let ctxt = getConwayTranslationContext cfgConway
+      $ PCons (RequireBoth $ \_cfgBabbage (WrapLedgerConfig cfgConway) ->
+                let ctxt = shelleyLedgerTranslationContext cfgConway
                 in
                 Pair2
                   (translateTxBabbageToConwayWrapper          ctxt)
@@ -383,22 +384,22 @@ translateLedgerStateByronToShelleyWrapper ::
      )
   => RequiringBoth
        WrapLedgerConfig
-       (Translate LedgerState)
+       (TickedTranslate LedgerState)
        ByronBlock
        (ShelleyBlock (TPraos c) (ShelleyEra c))
 translateLedgerStateByronToShelleyWrapper =
     RequireBoth $ \_ (WrapLedgerConfig cfgShelley) ->
-    Translate   $ \bound ledgerByron ->
+    Translate   $ \bound (Comp ledgerByron) ->
       ShelleyLedgerState {
         shelleyLedgerTip =
           translatePointByronToShelley
-            (ledgerTipPoint ledgerByron)
-            (byronLedgerTipBlockNo $ byronLedgerTransition ledgerByron)
+            (castPoint $ getTip ledgerByron)
+            (byronLedgerTipBlockNo $ untickedByronLedgerTransition ledgerByron)
       , shelleyLedgerState =
           SL.translateToShelleyLedgerState
             (toFromByronTranslationContext (shelleyLedgerGenesis cfgShelley))
             (boundEpoch bound)
-            (byronLedgerState ledgerByron)
+            (tickedByronLedgerState ledgerByron)
       , shelleyLedgerTransition =
           ShelleyTransitionInfo{shelleyAfterVoting = 0}
       }
@@ -504,18 +505,6 @@ crossEraForecastByronToShelleyWrapper =
   Translation from Shelley to Allegra
 -------------------------------------------------------------------------------}
 
-translateLedgerStateShelleyToAllegraWrapper ::
-     (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
-  => RequiringBoth
-       WrapLedgerConfig
-       (Translate LedgerState)
-       (ShelleyBlock (TPraos c) (ShelleyEra c))
-       (ShelleyBlock (TPraos c) (AllegraEra c))
-translateLedgerStateShelleyToAllegraWrapper =
-    ignoringBoth $
-      Translate $ \_bound ->
-        unComp . SL.translateEra' () . Comp
-
 translateTxShelleyToAllegraWrapper ::
      (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
   => InjectTx
@@ -531,22 +520,6 @@ translateValidatedTxShelleyToAllegraWrapper ::
        (ShelleyBlock (TPraos c) (AllegraEra c))
 translateValidatedTxShelleyToAllegraWrapper = InjectValidatedTx $
     fmap unComp . eitherToMaybe . runExcept . SL.translateEra () . Comp
-
-{-------------------------------------------------------------------------------
-  Translation from Shelley to Allegra
--------------------------------------------------------------------------------}
-
-translateLedgerStateAllegraToMaryWrapper ::
-     (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
-  => RequiringBoth
-       WrapLedgerConfig
-       (Translate LedgerState)
-       (ShelleyBlock (TPraos c) (AllegraEra c))
-       (ShelleyBlock (TPraos c) (MaryEra c))
-translateLedgerStateAllegraToMaryWrapper =
-    ignoringBoth $
-      Translate $ \_bound ->
-        unComp . SL.translateEra' () . Comp
 
 {-------------------------------------------------------------------------------
   Translation from Allegra to Mary
@@ -572,24 +545,6 @@ translateValidatedTxAllegraToMaryWrapper = InjectValidatedTx $
   Translation from Mary to Alonzo
 -------------------------------------------------------------------------------}
 
-translateLedgerStateMaryToAlonzoWrapper ::
-     (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
-  => RequiringBoth
-       WrapLedgerConfig
-       (Translate LedgerState)
-       (ShelleyBlock (TPraos c) (MaryEra c))
-       (ShelleyBlock (TPraos c) (AlonzoEra c))
-translateLedgerStateMaryToAlonzoWrapper =
-    RequireBoth $ \_cfgMary cfgAlonzo ->
-      Translate $ \_bound ->
-        unComp . SL.translateEra' (getAlonzoTranslationContext cfgAlonzo) . Comp
-
-getAlonzoTranslationContext ::
-     WrapLedgerConfig (ShelleyBlock (TPraos c) (AlonzoEra c))
-  -> SL.TranslationContext (AlonzoEra c)
-getAlonzoTranslationContext =
-    shelleyLedgerTranslationContext . unwrapLedgerConfig
-
 translateTxMaryToAlonzoWrapper ::
      (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
   => SL.TranslationContext (AlonzoEra c)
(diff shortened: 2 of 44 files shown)


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -683,9 +683,28 @@ translateLedgerStateBabbageToConwayWrapper ::
...
+      Translate $ \bound ->
+          unComp
+          -- then translate to Conway
+        . SL.translateEra' (getConwayTranslationContext cfgConway)
+        . Comp
+        . untickShelleyLedgerState
+          -- first tick to the boundary
+        . applyChainTick
+            (unwrapLedgerConfig cfgBabbage)
+            (boundSlot bound)
```
amesgen, line 698: Move this "into" the HFC

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
```diff
@@ -109,13 +110,15 @@ sequenceHardForkState (HardForkState tel) =
...
 -------------------------------------------------------------------------------}
 
--- | Translate @f x@ to @f y@ across an era transition
+-- | Translate @f x@ to @g y@ across an era transition
 --
--- Typically @f@ will be 'LedgerState' or 'WrapChainDepState'.
-newtype Translate f x y = Translate {
-      translateWith :: EpochNo -> f x -> f y
+-- Typically @f@/@g@ will be 'LedgerState' or 'WrapChainDepState'.
+newtype Translate' f g x y = Translate {
```
nfrisby, line 116: Could we hardcode `Ticked :.: f` instead of introducing `g`?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -712,3 +718,17 @@ translateValidatedTxBabbageToConwayWrapper ::
...
+
+{-------------------------------------------------------------------------------
+  Auxiliary
+-------------------------------------------------------------------------------}
+
+-- TODO abstract this out more
+untickShelleyLedgerState ::
+     TickedLedgerState (ShelleyBlock proto era)
+  -> LedgerState       (ShelleyBlock proto era)
+untickShelleyLedgerState st = ShelleyLedgerState {
```
nfrisby, line 730: Would you pattern match on `st` in a way that makes it locally obvious that this is half of an isomorphism?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -18,10 +18,12 @@
 {-# LANGUAGE TypeFamilies               #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE TypeOperators              #-}
```
nfrisby, line 21: This should move up two lines

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -205,10 +206,11 @@ injectInitialExtLedgerState cfg extLedgerState0 =
 
     targetEraLedgerState :: LedgerState (HardForkBlock (x ': xs))
     targetEraLedgerState =
-        HardForkLedgerState $
+        -- TODO discarding ledger events, probably fine?
```
nfrisby, line 209: I would also guess it's fine, but it's slightly tricky since this whole function is an (important) optimization (for the benchmarking team)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -732,6 +746,103 @@ shiftUpdate = go
           (eraIndexSucc ix)
           (eraIndexSucc ix')
 
+{-------------------------------------------------------------------------------
+  Extending
+-------------------------------------------------------------------------------}
+
+-- | Extend the telescope until the specified slot is within the era at the tip
+--
+-- TODO documentation (in particular that this now does ticking!)
```
nfrisby, line 755: TODO agreed!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -732,6 +746,103 @@ shiftUpdate = go
...
+extendToSlot ::
+     forall xs. CanHardFork xs
+  => HardForkLedgerConfig xs
+  -> SlotNo
+  -> HardForkState LedgerState xs
+  -> LedgerResult (LedgerState (HardForkBlock xs)) (HardForkState LedgerState xs)
+extendToSlot  ledgerCfg@HardForkLedgerConfig{..} slot ledgerSt@(HardForkState st) =
```
nfrisby, line 762: I anticipate we should be ticking N times if we do N translations. Do you agree?

And I think I see that natural pairing in `howExtend`. Would you add a comment there highlighting it?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
nfrisby: If so, perhaps call it `TranslateAndUntick`?


... although ... will the translation logic ever make use of the extra info in the `Ticked` wrapper? I'd need to look and think more carefully

nfrisby (commented):
First pass 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Types.hs
amesgen: Nice, just pushed, should be addressed there :smile:

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
amesgen: Hmm, will this always be true? In general, I think ticked ledger states could contain strictly more information than unticked ones.

Or are you only saying that this is true **ATM**, and it would be neat to locally highlight that?

nfrisby: Yeah, the second one.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -226,7 +228,7 @@ injectInitialExtLedgerState cfg extLedgerState0 =
         State.align
           (InPairs.requiringBoth
             (hmap (WrapConsensusConfig . configConsensus) cfgs)
-            (translateChainDepState hardForkEraTranslation))
+            (InPairs.hmap (error "TODO") $ translateChainDepState hardForkEraTranslation))
```
nfrisby, line 231: This is the `ChainDepState` bit I didn't handle.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/TPraos.hs
```diff
@@ -355,47 +360,51 @@ instance SL.PraosCrypto c => ConsensusProtocol (TPraos c) where
 
       SL.GenDelegs dlgMap = SL.lvGenDelegs lv
 
-  tickChainDepState cfg@TPraosConfig{..}
-                    (TickedPraosLedgerView lv)
-                    slot
-                    (TPraosState lastSlot st) =
+  projectHorizonView _cfg _lv = TickedTrivial
+
+  -- TODO comment explaining why we're not using the 'SL.tickChainDepState' function
```
nfrisby, line 365: Also TODO: open a PR upstream deleting that function?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
amesgen: Implemented this using `extendToHorizon` as added in the latest push.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
nfrisby: (Just a note to self: GH labels this Outdated, but the comment isn't there yet.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
amesgen: I left a justification.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/TPraos.hs
amesgen: I opened #350 for the natural larger task.

We could also go the 'opposite" route and add a property test that our vendored tick function behaves just as the one from ledger (for an interim period until #350 is done).
I locally tested that this works (and that introducing bugs in TPraos `tickChainDepState` cause the test to fail), but I am not sure if it is worth it:
<details>
<summary>Click to see compatibility test for TPraos ticking</summary>

```haskell
{-# LANGUAGE FlexibleInstances  #-}
{-# LANGUAGE NamedFieldPuns     #-}
{-# LANGUAGE StandaloneDeriving #-}

module Test.Consensus.Protocol.TPraos.LedgerCompat (tests) where

import qualified Cardano.Ledger.BaseTypes as SL
import qualified Cardano.Ledger.Chain as SL
import           Cardano.Ledger.Crypto
import qualified Cardano.Protocol.TPraos.API as SL
import           Cardano.Slotting.EpochInfo.API (epochInfoEpoch,
                     generalizeEpochInfo)
import           Data.Functor.Identity
import           Ouroboros.Consensus.Block
import           Ouroboros.Consensus.HardFork.History.EpochInfo
                     (toPureEpochInfo)
import           Ouroboros.Consensus.HardFork.History.Util (addSlots)
import           Ouroboros.Consensus.Protocol.Abstract
import           Ouroboros.Consensus.Protocol.TPraos
import           Test.Cardano.Ledger.Shelley.Utils (testGlobals)
import           Test.Cardano.Protocol.TPraos.Arbitrary ()
import           Test.QuickCheck
import           Test.Tasty
import           Test.Tasty.QuickCheck

tests :: TestTree
tests = testGroup "TPraos Ledger compat"
    [ testProperty "tick" testTickCompat
    ]

data TickSetup p = TickSetup {
    tsConsensusConfig  :: Blind (ConsensusConfig p)
  , tsTickedLedgerView :: Ticked (LedgerView p)
  , tsSlotNo           :: SlotNo
  , tsChainDepState    :: ChainDepState p
  }

instance Crypto c => Arbitrary (TickSetup (TPraos c)) where
  arbitrary = do
      initialNonce <- arbitrary
      let tpraosParams  = TPraosParams {
              tpraosSlotsPerKESPeriod = SL.slotsPerKESPeriod                 globals
            , tpraosLeaderF           = SL.activeSlotCoeff                   globals
            , tpraosSecurityParam     = SecurityParam $ SL.securityParameter globals
            , tpraosMaxKESEvo         = SL.maxKESEvo                         globals
            , tpraosQuorum            = SL.quorum                            globals
            , tpraosMaxMajorPV        = MaxMajorProtVer $ SL.maxMajorPV      globals
            , tpraosMaxLovelaceSupply = SL.maxLovelaceSupply                 globals
            , tpraosNetworkId         = SL.networkId                         globals
            , tpraosInitialNonce      = initialNonce
            , tpraosSystemStart       = SL.systemStart                       globals
            }
          tsConsensusConfig = Blind TPraosConfig {
              tpraosParams
            , tpraosEpochInfo = generalizeEpochInfo $ SL.epochInfoPure globals
            }
      chainCheckPParams   <-
        SL.ChainChecksPParams <$> arbitrary <*> arbitrary <*> arbitrary
      tsTickedLedgerView  <- fmap TickedPraosLedgerView $
        SL.LedgerView <$> arbitrary <*> arbitrary <*> arbitrary <*> arbitrary <*> pure chainCheckPParams
      slotAlreadyTickedTo <- frequency
        [ (9 , NotOrigin <$> chooseEnum (0, 1000))
        , (1, pure Origin)
        ]
      slotOffset          <- chooseEnum (0, 200)
      chainDepState       <- SL.ChainDepState <$> arbitrary <*> arbitrary <*> arbitrary
      let tsChainDepState = TPraosState {
              tpraosStateLastSlot      = slotAlreadyTickedTo
            , tpraosStateChainDepState = chainDepState
            }
          tsSlotNo = slotOffset `addSlots` fromWithOrigin 0 slotAlreadyTickedTo
      pure TickSetup {
          tsConsensusConfig
        , tsTickedLedgerView
        , tsSlotNo
        , tsChainDepState
        }
    where
      globals :: SL.Globals
      globals = testGlobals

testTickCompat :: Blind (TickSetup (TPraos StandardCrypto)) -> Property
testTickCompat (Blind ts) =
    consensusImpl === ledgerImpl
  where
    consensusImpl =
          tickedTPraosStateChainDepState
        $ tickChainDepState cfg tlv tsSlotNo cds

    ledgerImpl =
        SL.tickChainDepState
          (mkShelleyGlobals cfg)
          (getTickedPraosLedgerView tlv)
          tickingToNewEpoch
          (tpraosStateChainDepState cds)

    tickingToNewEpoch =
        epochNo (NotOrigin tsSlotNo) /= epochNo (tpraosStateLastSlot cds)
      where
        ei      = toPureEpochInfo $ tpraosEpochInfo cfg
        epochNo = runIdentity . epochInfoEpoch ei . fromWithOrigin 0

    TickSetup {
        tsConsensusConfig  = Blind cfg
      , tsTickedLedgerView = tlv
      , tsSlotNo
      , tsChainDepState    = cds
      } = ts
```

</details>

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -138,7 +139,8 @@ prop_simple_hfc_convergence testSetup@TestSetup{..} =
     counterexample ("eraSizeA: " <> show eraSizeA) $
     tabulate "epochs in era A" [labelEraSizeA] $
     prop_general args testOutput .&&.
-    prop_allExpectedBlocks
+    prop_allExpectedBlocks .&&.
+    prop_finalProtVers
```
dnadales, line 144: Just to confirm: we don't add an expected failure here because even if this commit won't be squashed, when bisecting we don't care about compiling the tests? Or do we? :thinking:

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -89,27 +89,30 @@ import           Ouroboros.Consensus.Util (ShowProxy (..), (..:))
...
-    --
-    -- Invariant: the domain of this map should equal the set of candidate
-    -- proposals.
-    ByronTransitionInfo !(Map Update.ProtocolVersion BlockNo)
+data ByronTransition = ByronTransitionInfo {
+      -- | Per candidate proposal, the 'BlockNo' in which it became a candidate
+      --
+      -- The HFC needs to know when a candidate proposal becomes stable. We
+      -- cannot reliably do this using 'SlotNo': doing so would mean that if we
+      -- were to switch to a denser fork, something that was previously deemed
```
dnadales, line 103: denser or sparser?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -280,6 +282,57 @@ forecastAcrossShelley cfgFrom cfgTo transition forecastFor ledgerStateFrom
                (SL.stabilityWindow (shelleyLedgerGlobals cfgFrom))
                (SL.stabilityWindow (shelleyLedgerGlobals cfgTo))
 
+translateLedgerStateAcrossShelley ::
```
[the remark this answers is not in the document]
dnadales: In particular, contributors might be puzzled about the unticking part. Perhaps it's worth replicating here what was discussed in the issue that motivated this change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -738,17 +752,37 @@ shiftUpdate = go
...
+--
+-- This function relies on 'singleEraTransition' to find out when a specific era
+-- is ending. If so, we:
+--
+--  * first tick just across the era boundary using the logic of the old era,
+--
+--  * then translate the ledger state into the new era (using
+--    'translateLedgerState').
+--
+-- The final tick to the given slot does /not/ happen here.
```
dnadales, line 765: I feel a bit uneasy about the fact that this function has to mention what happens elsewhere. I guess there's no way to avoid it given the current design, but it seems a very delicate piece of logic is scattered around.

dnadales (approved):

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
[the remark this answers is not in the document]
amesgen: Good point, will add a reference to the comment on `translateLedgerState` for context, and there also #339.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -116,6 +117,16 @@ class ( Show (ChainDepState   p)
...
 
+  -- | A projection of 'LedgerView' containing only what is needed for ticking
+  --
+  -- For single-era protocols, there are further constraints on this type, see
+  -- 'Ouroboros.Consensus.HardFork.Combinator.Abstract.SingleEraBlock.eraTransitionHorizonView'.
+  -- Usually, 'HorizonView' will be a singleton type.
+  --
+  -- For the 'Ouroboros.Consensus.HardFork.Combinator.Basics.HardForkProtocol'
+  -- combinator, this is the same as the 'LedgerView'.
+  type family HorizonView p :: Type
```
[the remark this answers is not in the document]
amesgen: The second bullet point is enabled  by #354.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/TPraos.hs
```diff
@@ -247,6 +248,14 @@ newtype instance Ticked (SL.LedgerView c) = TickedPraosLedgerView {
       getTickedPraosLedgerView :: SL.LedgerView c
     }
 
+newtype TPraosHorizonView = TPraosHorizonView {
```
nfrisby, line 253: ... want to make this empty?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/TPraos.hs
```diff
@@ -361,27 +370,42 @@ instance SL.PraosCrypto c => ConsensusProtocol (TPraos c) where
...
-              st
-      shelleyGlobals = mkShelleyGlobals cfg
+      newEpoch = isNewEpoch
+                   (History.toPureEpochInfo tpraosEpochInfo)
+                   lastSlot
+                   slot
+
+      -- We can't use 'SL.tickChainDepState' as it unnecessarily takes the
+      -- entire 'SL.LedgerView' as an argument. Hence, we inline it here; future
+      -- work should include moving the TPraos logic entirely to Consensus.
```
nfrisby, line 419:
```suggestion
      -- work should include moving the TPraos logic entirely to the ouroboros-consensus repository.
```

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Translate.hs
```diff
@@ -18,21 +18,35 @@ import qualified Cardano.Protocol.TPraos.Rules.Prtcl as SL
...
-                     TPraosState (tpraosStateChainDepState, tpraosStateLastSlot))
+import           Ouroboros.Consensus.Protocol.TPraos (TPraos, TPraosParams (..))
 import qualified Ouroboros.Consensus.Protocol.TPraos as TPraos
 import           Ouroboros.Consensus.Protocol.Translate (TranslateProto (..))
 
 {-------------------------------------------------------------------------------
   Translation from transitional Praos
 -------------------------------------------------------------------------------}
 
+instance (c1 ~ c2) => TranslateProto (TPraos c1) (TPraos c2) where
```
nfrisby, line 37: I wonder if the error message would actually be easier to understand if the head was `TranslateProto (TPraos c) (TPraos c)`?

And same for `Praos` below.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/TPraos.hs
```diff
@@ -253,7 +255,7 @@ newtype TPraosHorizonView = TPraosHorizonView {
     }
 
 newtype instance Ticked TPraosHorizonView = TickedTPraosHorizonView {
-      tickedTpraosHorizonViewExtraEntropy :: SL.Nonce
+      tickedTPraosHorizonViewExtraEntropy :: SL.Nonce
```
nfrisby, line 258: Minor: preferably this diff is squashed into the commit that introduces `TickedTPraosHorizonView`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
nfrisby: I could have sworn that I commented in this thread in my Review just now...

My recommendation was ultimately to rename the function `extendToEraOfSlot` or something like that.

nfrisby (changes requested):
I think https://github.com/input-output-hk/ouroboros-consensus/pull/340/commits/a82b126d78d7b0afcbc43a5da489b023c67a31fa is important.

We should squash it down.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -353,6 +358,25 @@ prop_simple_hfc_convergence testSetup@TestSetup{..} =
...
 
+    prop_finalProtVers :: Property
+    prop_finalProtVers =
+        counterexample ("final protocol versions: " <> show finalProtVers) $
+        -- TODO This property is showcasing a problem with the HFC: even though
+        -- we will definitely end up in era B (and hence, the protocol version
+        -- should be @'succ' 'initProtVer'@), this is currently not the case.
+        -- Subsequent commits will fix this.
+        all (== initProtVer) finalProtVers
+
```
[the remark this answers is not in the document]
amesgen: No, will remove :+1:

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -211,7 +216,27 @@ instance IsLedger (LedgerState BlockA) where
...
+        | targetSlot >= firstSlotNextEpoch -- cross-epoch tick
+        , Just (_, confirmationDepth) <- getConfirmationDepth st
+          -- block with transition tx is doubly stable
+        , confirmationDepth >= safeFromTipA k + stabilityWindowA k
+        = \s -> s { lgrA_protVer = succ (lgrA_protVer s) }
+        | otherwise = id
+
+      k = lcfgA_k cfg
+
+      firstSlotNextEpoch = runIdentity $ do
```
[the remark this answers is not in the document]
amesgen: I only know of the functions in [`Ouroboros.Consensus.Ledger.Protocol.Util`](https://github.com/input-output-hk/ouroboros-consensus/blob/f743b8fe9e3e8f24822ba503ecf43eab13bce66f/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/Util.hs) (in particular `isNewEpoch`), but we can't use them here without shuffling things around as ouroboros-consensus-diffusion does not depend on ouroboros-consensus-protocol. I will keep it like this for now, but we could also start some `EpochInfo`-related utilities in `ouroboros-consensus` such that they can be used in both `oc-diffusion` and `oc-protocol`.

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -89,27 +89,30 @@ import           Ouroboros.Consensus.Util (ShowProxy (..), (..:))
...
+      -- stable is suddenly not deemed stable anymore (although in actuality it
+      -- still is). We therefore must do this based on 'BlockNo' instead, but
+      -- unfortunately the Byron ledger does not record this information.
+      -- Therefore, we record it here instead.
+      --
+      -- Invariant: the domain of this map should equal the set of candidate
+      -- proposals.
+      perCandidateProposals :: !(Map Update.ProtocolVersion BlockNo)
+    , -- | Block number of the last applied block, used to compute when a
+      -- proposal got stable.
```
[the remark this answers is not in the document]
amesgen: TIL :+1:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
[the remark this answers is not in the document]
amesgen: > Maybe a Haddock comment that at least highlights that the below works perfectly fine when `eraTo ~ eraFrom`.

It doesn't as that would contradict `SL.PreviousEra eraTo ~ eraFrom`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
amesgen: FTR: You are probably referring to https://github.com/input-output-hk/ouroboros-consensus/pull/340#discussion_r1331794291

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -738,17 +752,38 @@ shiftUpdate = go
...
+--
+--  * first tick just across the era boundary using the logic of the old era,
+--
+--  * then translate the ledger state into the new era (using
+--    'translateLedgerState').
+--
+-- The final tick to the given slot does /not/ happen here, but rather in
+-- 'applyChainTickLedgerResult' above.
+--
+-- It can be the case that the telescope is extended multiple times, for example
```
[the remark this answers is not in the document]
amesgen: Good point, for the example here, I will make it specifically about the zero-epoch eras for benchmarking as that is the most relevant case today.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Translate.hs
amesgen: I would have to play around to find this out, I *think* the way how it is written here can sometimes be beneficial because it causes instances selection to pick this instance even if it does not (yet) know `c1 ~ c2` (a similar example is discussed [here](https://youtu.be/XfIlhJFmw3c?t=930)).

amesgen:
Closing for now, see https://github.com/input-output-hk/ouroboros-consensus/issues/339#issuecomment-1758097242

PR #346, opened 2023-09-15
docs: draft HardForkCombinatorRollbackMonotonicityProposal.md

nfrisby:
Draft PR


nfrisby:
This is a Draft PR. But I'd appreciate your thoughts already.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Question 1
+
+Which aspects of the protocol and ledger's future behavior must this node be able to predict from a given ledger state and why?
+
+- Assume it's well-connected and running in a healthy net.
+- Assume k > 0.
+- Let D be the node's currently selected chain.
+- Let C be the current best valid chain in the net.
+- Let L be the ledger state resulting from the youngest block that is on both C and D.
```
dnadales, line 16: Is the youngest block different from the intersection block? (that corresponds to the intersection point)

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+- Let R count the number of blocks after L on D.
+  Observe that there must be more than R blocks after L on C, since C is better than D.
+  (This isn't necessarily true, due to tiebreakers.
+  However, even with tiebreakers, this remains true in the worst case, which is what ultimately matters.)
```
dnadales, line 20: It's not clear to me what the worst case is here, and why this is true in that case.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+- Common Prefix ensures R <= k.
+- L must contain enough information for the node to correctly judge the validity of the k+1th block after L on C.
+- Otherwise the Header-Body Split would prevent the node from adopting C, even though doing so requires at most rolling back k blocks.
```
dnadales, line 23: Perhaps the "Otherwise" bullet can be merged with the previous one, or indented one level.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+That is a complete answer.
+In particular, there is nothing else this simplified node needs to predict.
+
+If we assume Chain Growth in addition to Common Prefix, then we can further refine the answer.
+
+- Chain Growth ensures that every window of scg slots will contain at least k blocks on C.
```
dnadales, line 30: `scg`?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+- Because of the Header-Body Split, L must contain enough information for the node to correctly judge the validity of any header that is no more than 3k/f after the first block on C.
```
dnadales, line 31: The first block on `C` or the first block in the intersection? 

Also, `3k/f` appears suddenly here. Perhaps we could say something about its relationship to k (or create a new `.md` file and link to it).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+Note that, if necessary, L can be ticked in order to validate the first header after L on C, without requiring any predictions.
+In the absence of the corresponding block body---ie because of the Header-Body Split---the assessment of subsequent headers on C will require predictions.
+Thus a prediction range of at least 3k/f ensures at least k blocks after the first header, and hence k+1 blocks in total, as necessary.
```
dnadales, line 35: Why `k+1` in total? Are we counting the header of the block that led to `L`?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Question 2
+
+What behaviors could arise only if Chain Growth were violated?
+
+- For example, suppose R=k and the k+1th block after L on C was more than 3k/f slots after the first block after L on C.
+- In this case, the node might not be able to switch from D to C despite R <= k.
+  It can't switch to C without downloading the R+1 blocks, which it can't do without validating the R+1 headers, which it can't do without predicting more than 3k/f slots after the first header, which it might not be able to do.
+
+For example, suppose that L is the result of a block in relative slot number 7k/f - 2 within some epoch.
```
dnadales, line 45: Why did we choose `7k/ f - 2`?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+Specifically, the node will need to know if the chain will actually be in whichever era that header claims to be in, and it wouldn't necessarily be able to answer that question more than a safe zone after the first header after L on C.
+
+The answer to Question 2 does change: there is one additional change in behavior possibly enabled by a Chain Growth violation.
+Unlike the ledger, the hard fork combinator actually counts blocks instead of only slots.
+If a hypothetical roll back of k-many blocks could possibly change the ledger-based governance decision that was intended to trigger the era transition---which is true if Chain Growth was (severely) violated in the last two stability windows before the intended era transition---then the hard fork combinator will not transition to the next era.
+(For this reason, the voting deadline was moved from 7k/f to 4k/f in the ledger governance rules, so that the hard fork combinator's final decision would be known at least a stability window (and so the safe zone) before the epoch transition.)
```
dnadales, line 80: moved to `4k/f` after the beginning of the epoch? Or before it ends?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Back to Reality: the Cardano Node
+
+First, instead of merely pure Praos, consider a node running Cardano's hard fork combinator and its history of eras from Byron through Babbage.
+The answer to Question 1 actually does not change.
+One crucial reason is because Cardano sets the hard fork combinator safe zone of each era to that era's stability window.
+If the safe zone were instead less than one stability window, then L does not necessarily contain enough information to validate the R+1th header after L on C, _even without_ a Chain Growth violation.
+Specifically, the node will need to know if the chain will actually be in whichever era that header claims to be in, and it wouldn't necessarily be able to answer that question more than a safe zone after the first header after L on C.
+
+The answer to Question 2 does change: there is one additional change in behavior possibly enabled by a Chain Growth violation.
```
dnadales, line 77: I think a concrete example would be great here.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+        GetStakeDelegDeposits                   :: Set (StakeCredential (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (StakeCredential (EraCrypto era)) Coin)
+        GetConstitution                         :: BlockQuery (ShelleyBlock proto era) (Maybe (LC.Constitution era))
+        GetGovState                             :: BlockQuery (ShelleyBlock proto era) (LC.GovState era)
+        GetDRepState                            :: Set (SL.Credential 'DRepRole (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (SL.Credential 'DRepRole (EraCrypto era)) (SL.DRepState (EraCrypto era)))
+        GetDRepStakeDistr                       :: Set (LC.DRep (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (LC.DRep (EraCrypto era)) Coin)
+        GetCommitteeState                       :: BlockQuery (ShelleyBlock proto era) (SL.CommitteeState era)
+
+The queries marked with `*` and/or `+` are the only queries whose responses exhibit weak monotinicity as the node improves its selection by switching/extending chains.
```
dnadales, line 136: Could we define, explain and exemplify "weak monotonicity"  in this context?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Suppose Y is the first block after L on C. The forecasting allows for k blocks after Y on C. So that's k+1 total.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Because 7k/f - 2 is "slightly before one stability window before the end of the epoch".

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: 4k/f in the epoch = ie 6k/f before it ends = two stability windows before it ends

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: It's the same. I'd tried to be as concrete as possible, despite those on our team having established that "the intersection (block)" is what I mean here.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Yes, this parenthetical is admittedly handwaivy.

C can be better than D in two ways: it can be the same length and have a better tiebreaker or it can be longer. When it's longer requires a deeper rollback from us. That extra depth is the "worst case" aspect here.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,190 @@
...
+- Let L be the ledger state resulting from the youngest block that is on both C and D.
+- Let R count the number of blocks after L on D.
+  Observe that there must be more than R blocks after L on C, since C is better than D.
+  (This isn't necessarily true, due to tiebreakers.
+  However, even with tiebreakers, this remains true in the worst case, which is what ultimately matters.)
+- Common Prefix ensures R <= k.
+- L must contain enough information for the node to correctly judge the validity of the k+1th block after L on C.
+- Otherwise the Header-Body Split would prevent the node from adopting C, even though doing so requires at most rolling back k blocks.
+
+That is a complete answer.
```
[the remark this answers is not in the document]
nfrisby: > What constitutes a complete answer in this context?

The next sentence was intended to clarify that.

> how do we convince ourselves this is one of them?

Conversation, property testing, etc :D

nfrisby:
Just a point of reference: the node CLI query for the leader schedule is perhaps also in the same gray area as time translations. Should that query refuse to answer until a rollback by <= k blocks could not change it?

dnadales (dismissed):

nfrisby:
I also pushed up some adjustments for the smaller comments you made.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+  However, it's also almost certain there would be hundreds of blocks in every window of 3k/f slots.
+  And, since the remaining nodes are still sufficiently well-connected, Praos's natural short forks would indeed remain short.
+  Thus R would remain much smaller than k, such that any node that needed to switch from D to C would still be able to validate R+1 headers after L on C despite not being able to predict more than 3k/f after the first header after L on C.
+
+## Back to Reality: the Cardano Node
+
+First, instead of merely pure Praos, consider a node running Cardano's hard fork combinator and its history of eras from Byron through Babbage.
+The answer to Question 1 actually does not change.
+One crucial reason is because Cardano sets the hard fork combinator safe zone of each era to that era's stability window.
+If the safe zone were instead less than one stability window, then L does not necessarily contain enough information to validate the R+1th header after L on C, _even without_ a Chain Growth violation.
```
amesgen, line 169: On the Byron->Shelley transition, it is also relevant that forecasting from a Byron blocks within 2k slots of the era transition into Shelley has a range much *larger* than one Byron stability window, otherwise this property would be violated.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+Unlike the ledger, the hard fork combinator actually counts blocks instead of only slots.
+If a hypothetical roll back of k-many blocks could possibly change the ledger-based governance decision that was intended to trigger the era transition---which is true if Chain Growth was (severely) violated in the last two stability windows before the intended era transition---then the hard fork combinator will not transition to the next era.
+(For this reason, the voting deadline was moved from 7k/f to 4k/f (ie two stablity windows before the end) in the ledger governance rules, so that the hard fork combinator's final decision would be known at least a stability window (and so the safe zone) before the epoch transition.)
+Awkwardly, the ledger will still update the major version of the protocol version protocol parameter, but the chain will remain in the previous era, which many users and developers would find counter-intuitive.
+
+So this incoherence can arise only due to a Chain Growth violation (and a specific and severe one, at that), which means the hard fork combinator itself changes the answer to Question 2.
```
amesgen, line 178: I think we weren't yet aware of this when this text was written, but I think it would be good to mention that a less severe chain growth violation (when the k+1-st still occurs before the epoch end, but only in the last 3k/f slots) still has pretty weird behavior (https://github.com/input-output-hk/ouroboros-consensus/issues/385).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+        GetStakeDelegDeposits                   :: Set (StakeCredential (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (StakeCredential (EraCrypto era)) Coin)
+        GetConstitution                         :: BlockQuery (ShelleyBlock proto era) (Maybe (LC.Constitution era))
+        GetGovState                             :: BlockQuery (ShelleyBlock proto era) (LC.GovState era)
+        GetDRepState                            :: Set (SL.Credential 'DRepRole (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (SL.Credential 'DRepRole (EraCrypto era)) (SL.DRepState (EraCrypto era)))
+        GetDRepStakeDistr                       :: Set (LC.DRep (EraCrypto era)) -> BlockQuery (ShelleyBlock proto era) (Map (LC.DRep (EraCrypto era)) Coin)
+        GetCommitteeState                       :: BlockQuery (ShelleyBlock proto era) (SL.CommitteeState era)
+
+The queries marked with `*` and/or `+` are the only queries whose responses exhibit weak monotinicity as the node improves its selection by switching/extending chains.
```
amesgen, line 231:
```suggestion
The queries marked with `*` and/or `+` are the only queries whose responses exhibit weak monotonicity as the node improves its selection by switching/extending chains.
```

amesgen, line 231: I guess you are implicitly assuming here that the client always acquires the latest ledger state? Otherwise, these monotonicity properties don't hold of course.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- `**` means the monotonicity could only be spoiled if something like data corruption caused the node to roll back some blocks without simultaneously adding at least that many blocks.
+- `*+` means the monotonicity could only be spoiled if the node somehow performed a switch that rolled back more than k blocks.
+   Call this _k-rollback-monotonicity_.
+
+If the hard fork combinator were to always agree with the ledger---thereby avoiding the awkward surprise arising possible after a Chain Growth violation---then it would have to count only slots as the ledger does and not blocks.
+As a result, `GetEraStart` and `GetInterpreter` would no longer exhibit weak monotonicity.
+Specifically, a particular Chain Growth violation could cause `GetEraStart` to yield `Just x` in one moment and then `Nothing` or `Just y` where `x /= y` at a later moment, despite the node never rolling back more than k blocks.
+(`GetInterpreter` would be more intricate to describe, but the nub is exactly the same.)
+
+It is remarkable that every query without a marker is not weakly monotonic, and no one expects them to be!
```
amesgen, line 245: I agree, but part of that is that "the UTxO is monotonic" seems incoherent even (what would be the (partial) order) :+1:

amesgen (approved):

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
+# DRAFT NUMBER 2
+
+# Introduction
+
+The Babbage->Conway transition on an earlier iteration of SanchoNet revealed a bug in the Consensus layer by which the Hard Fork Combinator transitioned to Conway because the relevant Babbage protocol parameter update proposal received sufficient votes but then that same proposal wasn't actually enacted: the major protocol version remained unchanged even after Conway started.
+
+Troubleshooting this bug and moreover determining how to fix it in such a way that makes similar mistakes less likely in the future has lead to the following mini-framework for understanding the HFC's behavior on a particular chain.
+
+So many dimensions arose from investiating such a relatively simple misbehavior for two reasons.
+First, the mechanisms implementing these policies are---at least in the current implementation---somewhat interdependent.
```
dnadales, line 10: Which policies are we referring to here? The tick/translate policies?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+Second, the HFC (including its instantiation for Cardano) remains underdocumented since back when it was originaly developed under time-pressure, and our team no longer includes its inventor.
+The Consensus Team learns new things and unfortunately re-learns some old things every time they dive in to alter it.
+(Optimistically, Nicolas Frisby feels they currently understand it now better than they have ever before.)
+
+## Dimensions
+
+The following discrete variables characterize dimensions of the HFC design space.
+In this context, the _determination_ is the `singleEraTransition` method: it's how an era's ledger state determines when the era ends.
```
dnadales, line 18: I'm a bit confused by the reference to `singleEraTransition`? I would have expected something like: "the determination is the (single?) era transition method"

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+- DST - Determination is Sensitive to Ticks  - whether ticking the ledger state can change its determination of the end of its era: Sensitive vs Insensitive
+
+- CCT - Criss-Crossed Ticks - whether an era's tick rules might be used to tick across slots that are in _other_ eras: Allowed vs Disallowed
+
+- PIL - Potential to Ignore the Ledger - whether the HFC can ignore the ledger's hard fork-related governance: Possible vs Impossible
+
+- DSR - Determination is Sensitive to Rollback - for chains C and D where C < D and len(C) <= len(C /\ D) + k, whether the current era of C can have a determined end on C but an undetermined or different end on D: Sensitive vs Insensitive
```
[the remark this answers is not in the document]
dnadales: I feel I miss the units in `len (C ^ D) + k`. Is it in blocks or slots?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+## Dimensions
+
+The following discrete variables characterize dimensions of the HFC design space.
+In this context, the _determination_ is the `singleEraTransition` method: it's how an era's ledger state determines when the era ends.
+
+- DST - Determination is Sensitive to Ticks  - whether ticking the ledger state can change its determination of the end of its era: Sensitive vs Insensitive
+
+- CCT - Criss-Crossed Ticks - whether an era's tick rules might be used to tick across slots that are in _other_ eras: Allowed vs Disallowed
+
+- PIL - Potential to Ignore the Ledger - whether the HFC can ignore the ledger's hard fork-related governance: Possible vs Impossible
```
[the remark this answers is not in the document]
amesgen: Yes, due to "block counting", the current HFC implementation will ignore the ledger's hard fork-related governance (concretely: updating the ledger protocol version based on voting by the Genesis keys) when less than $k$ blocks occurr between the $4k/f$ voting deadline and the end of the epoch (note that this is an extremely severe chain growth violation). Here, "ignore" means that the ledger will update the protocol version to `X` at the epoch boundary, but the HFC will not perform an era transition even if `TriggerHardForkAtVersion X` has been configured.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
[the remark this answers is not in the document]
dnadales: If `len (C)` would be `> len (C ^ D) + k` then we would not be able to roll back to `D`, right? I wonder if we need to mention this at all: we could simply say `D` is a chain we could potentially switch to (regardless of the conditions under which this switch is possible).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- ZEE - Zero Epoch Eras - whether an era (in its entirety) can contain zero epochs: Allowed vs Disallowed
+
+- ZBE - Zero Block Eras - whether an era (in its entirety) can contain zero blocks: Allowed vs Disallowed
+
+- QSR - Queries are Sensitivity to Rollback - same as DSR but for HFC-related Node-To-Client queries instead of for validity of blocks: Sensitive vs Insensitive
+
+Note the following relations.
+
+- ZEE=Allowed implies ZBE=Allowed.
+- DSR=Insensitive implies QSR=Insensitive.
```
dnadales, line 37: I don't get this. What about queries that are not related to era transitions?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Table for the Current Cardano HFC
+
+The following table and associated notes classifies the current HFC implementation in terms of these dimensions.
+
+| DST | CCT | PIL | DSR | ZEE | ZBE | QSR |
+| --- | --- | --- | --- | --- | --- | --- |
+| I   | A   | P 1 | I   | 2   | 3   | I   |
+
+- (1) It's Possible, but only if there's a Chain Growth violation at a particular time during the epoch.
```
dnadales, line 47: What does (1) refer to?

dnadales: I would need an example to understand this :grimacing:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales, line 24: On a second (third) read, I realize an example of this would help me understand what we mean by ignore the ledger HF related gov.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
amesgen: It refers to the `1` in the PIL column right above (maybe it could also use (1), or the GH markdown footnote syntax?).

> I would need an example to understand this 😬

See the description in https://github.com/input-output-hk/ouroboros-consensus/pull/346#discussion_r1354586938 (less than $k$ blocks in the last $6k/f$ slots of an epoch)

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
amesgen: (see https://github.com/input-output-hk/ouroboros-consensus/pull/346#discussion_r1354586938)

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- ZEE=Allowed implies ZBE=Allowed.
+- DSR=Insensitive implies QSR=Insensitive.
+
+## Table for the Current Cardano HFC
+
+The following table and associated notes classifies the current HFC implementation in terms of these dimensions.
+
+| DST | CCT | PIL | DSR | ZEE | ZBE | QSR |
+| --- | --- | --- | --- | --- | --- | --- |
+| I   | A   | P 1 | I   | 2   | 3   | I   |
```
dnadales, line 45: About `DST: I`: I would have expected that it is sensitive to ticks since ticking might cause votes to be counted and a hardfork could be decided as a consequence of this. Would it be possible to explain why this is not the case?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+- (1) It's Possible, but only if there's a Chain Growth violation at a particular time during the epoch.
+
+- (2) The current HFC's behavior is inconsistent without making further assumptions.
+  Overall, ZEE is allowed, but it might cause some unexpected behavior.
+  In particular, forecasts always anticipate that there is at least one safe zone of slots in an era (which is not true for eras with zero epochs).
+  That mismatch could manifest as incorrect time-slot translations eg (within ledger rules execution, but not within Node-To-Client query reponses).
```
dnadales, line 52: Could we elaborate an example of this?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+  Moreover, today's node would not even mint blocks, since the leadership check (at least currently) relies on forecasting.
+
+## Table for an Particular Hypothetical Cardano HFC
+
+The following table and associated notes classifies an hypothetical HFC implementation in terms of these dimensions.
+Nicolas Frisby anticipates this implementation would be easier to understand, less likely to surprise users, more general (ie place fewer requirements on the block types), and not require excessive code changes.
+
+| DST | CCT | PIL | DSR | ZSE | ZBE | QSR |
+| --- | --- | --- | --- | --- | --- | --- |
+| S   | D   | I   | S 2 | D   | D 1 | I 3 |
```
dnadales, line 65: The proposed options seem intuitive to me :+1:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+- (1) ZBE would be achievable for a prefix of eras when constructing the initial HFC ledger state (in the `ProtocolInfo` passed to the Consensus layer), but otherwise the HFC as a block type does not need to support ZBE at all.
```
dnadales, line 67: Yes, I think this is quite important for starting in a given Era, which is crucial for testing and benchmarking.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+- (2) Esgen points out that perhaps DSR could/should still be Insensitive, and that wouldn't obviously lose too much simplicity.
```
dnadales, line 69: I wonder how that could even be possible. :thinking: Do you have an example?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+- (3) Having QSR=Insensitive despite DSR=Sensitive requires an additional mechanism.
+  In this design point, that is anticipated to be constraining these queries to only be answered via an immutable ledger state.
+
+## Disclaimer
+
+There may be other points in the design space worth considering.
+(Or even other dimensions, of course.)
+In particular, DST=Insensitive and PIL=Impossible could be achieved by changing the Cardano ledger's governance to match the existing Cardano HFC logic.
```
dnadales, line 78: I can't imagine/understand why this would be the case. Would it be possible to illustrate/expand on this explanation?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+Specifically, if the ledger did not enact governance actions unless there have actually been k+1 blocks since they were ratified even despite possible Chain Growth violations, then the Cardano HFC would never override the Cardano ledger.
```
dnadales, line 79: I fail to see the relation between chain growth violations and k + 1 blocks.

nfrisby: Right now governance actions are enacted if they both 1) get enough votes and 2) get those votes before a slot-based deadline. The specified choice of _slot_ for the deadline has so far ensured that the final Yes-vs-No status of the governance action would be determined (ie predictable) more than k _blocks_ before the change was actually made (aka "enacted")---unless there was a Chain Growth violation (since Chain Growth ensures at least k blocks of growth per a fixed number of slots).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: I had intended the next sentence to do so. So I suppose we should expand that one.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Yeah. Extremely high-level summary: it's because of PIL=Possible. The ticking does make the votes happening, but if there hasn't been enough blocks, the Cardano HFC logic will "veto" the decision and simply not transition to the next era.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: And https://github.com/input-output-hk/ouroboros-consensus/pull/346#discussion_r1355048349

(I don't know why I'm reviewing these PR comments bottom-up 🤷 )

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: > What about queries that are not related to era transitions?

I wasn't considering them at all here. So this document doesn't constraint them at all.

I should say that explicitly in the definition of QSR then.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: I don't think I understand what you mean by this GitHub comment.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: That word choice was a bit handwavey. I was thinking of eg "PIL=Impossible" versus "PIL=Possible" as two possible policies.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: We could. It should be pretty simple, but it will require a lot of words, just because of the context. If it doesn't require a lot of works, then it would remain pretty abstract, which kind of defeats the purpose of the example.

I left it out for now because it seems like a lot of words. Maybe we could consolidate all of the examples into an "example" section that shares the necessary context amongst a few pointed examples.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: I think @amesgen would more easily be able to explain it than I can.

amesgen: Right now, the criterion for `singleEraTransition` to report the era transition for Shelley-based eras is:

 - Were there at least $k$ blocks after the voting deadline ($6k/f$ before an epoch boundary)?

We could change that to

 - Were there at least $k$ blocks after the voting deadline *or* was the last applied block within the last $3k/f$ slots of the epoch?

This choice would have DST=Sensitive, PIL=Impossible and DSR=Insensitive (apart from chain growth violations in the $3k/f$ slots after the voting deadline).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+---
+
+# DRAFT NUMBER 1
+
+## Introduction
+
+This document motivates a simplification of the HFC, that would also alleviate the "double stability" constraint on the ledger's governance rules.
+
+Specifically, it suggests that the HFC should not be vetoing the ledger's decision to transition to the next era while that decision is not yet immutable.
+This veto can only happen in the presence of a Chain Growth violation, but it's confusing none-the-less and requires extra implementation complexity in the HFC.
```
dnadales, line 92: Could we add an example of how the CG violation would lead to the HFC vetoing the ledger's decision? I thought that the requirement of 2-stability-windows requirement was due to rollbacks.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Question 1
+
+Which aspects of the protocol and ledger's future behavior must this node be able to predict from a given ledger state and why?
+
+- Assume it's well-connected and running in a healthy net.
+- Assume k > 0.
+- Let C be the current best valid chain in the net.
+- Let D be the node's currently selected chain, and assume it's worse than C.
+- Let L be the ledger state resulting from the youngest block that is on both C and D, aka their intersection.
```
dnadales, line 111: Now I realize that I'm confused about the mention of "youngest block", doesn't block at the intersection suffice?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- Let R count the number of blocks after L on D.
+  Observe that there must be more than R blocks after L on C, since C is better than D.
+  (This isn't necessarily true, due to tiebreakers.
+  However, even with tiebreakers, this remains true in the worst case, which is what ultimately matters.)
+- Common Prefix ensures R <= k.
+- L must contain enough information for the node to correctly judge the validity of the k+1th block after L on C.
```
dnadales, line 117: Not related to this PR, but why do we require `k+1` instead of k? Because otherwise we won't be able to switch to `C`?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- Otherwise the Header-Body Split would prevent the node from adopting C, even though doing so requires at most rolling back k blocks.
+
+That is a complete answer.
+In particular, there is nothing else this simplified node needs to predict.
+
+If we assume Chain Growth in addition to Common Prefix, then we can further refine the answer.
+
+- Chain Growth ensures that every window of scg (ie "the s parameter of the papers' CG property", aka "one stability window") slots will contain at least k blocks on C.
+- Because of the Header-Body Split, L must contain enough information for the node to correctly judge the validity of any header that is no more than 3k/f after the first block after L on C.
```
dnadales, line 126: Why do we require that `L` can judge the validity of at least `k + 1` blocks on any chain that forks from it, but judge the validity of `3k / f` slots only? CG ensures we have at least `k` blocks, but we could have *exactly* `k`. Does that mean that we can only switch to chains that have `> k` blocks in `3k/f` slots?

dnadales: Hmmm, I think this is described below:

>  and hence k+1 blocks in total, as necessary.

Are we counting the intersection block as well? If so I'm still confused about the requirement that `L` can validate `k + 1` blocks *after* the intersection :grimacing:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+Note that, if necessary, L can be ticked in order to validate the first header after L on C, without requiring any predictions.
+In the absence of the corresponding block body---ie because of the Header-Body Split---the assessment of subsequent headers on C will require predictions.
+Thus a prediction range of at least 3k/f ensures at least k blocks after the first header, and hence k+1 blocks in total, as necessary.
+
+### Question 2
+
+What behaviors could arise only if Chain Growth were violated?
+
+- For example, suppose R=k and the k+1th block after L on C was more than 3k/f slots after the first block after L on C.
+- In this case, the node might not be able to switch from D to C despite R <= k.
```
dnadales, line 137: I've drawn some pictures. I find they help me greatly in following the arguments presented here. Maybe we can add them to our documentation (but might not be needed for this PR).

image

And perhaps we should unify the notation for these diagrams :smile:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+  It can't switch to C without downloading the R+1 blocks, which it can't do without validating the R+1 headers, which it can't do without predicting more than 3k/f slots after the first header, which it might not be able to do.
+
+For example, suppose that L is the result of a block in relative slot number 7k/f - 2 within some epoch (ie "slightly more than scg before the end of the epoch").
```
dnadales, line 140: Here is a bit of context I'm missing: do we assume a certain duration of the epochs in terms of `3k/f` (or `k/f`)?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+Because there might be a block in relative slot 7k/f - 1---which would change the nonce used for the next epoch---L is unable to validate headers in the next epoch.
+If the first header after L on C is indeed in slot 7k/f - 1, then Chain Growth ensures the k+1th header after L on C will still be in this same epoch, and so L can validate it.
```
dnadales, line 142: Without a drawing, it's quite hard for me to follow what `L` and `C` are. (I had to make one myself and I go to it back and forth, maybe we can spare the reader having to make her own drawing).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+If the first header after L on C is instead after slot 7k/f - 1, then ticking L to that first header will tick it past slot 7k/f - 1, and so determines the next epoch's nonce.
```
dnadales, line 143: We would also need some more context/background info on when nonces are determined in relation to the slot numbers or division of epoch into chunks of `k/f`.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+That's likely sufficient for the next k headers, unless there is an extremely severe Chain Growth violation.
+(In the non-simplified node, the prediction is always restricted to exactly 3k/f slots, despite the data in question actually advancing an entire epoch at a time.
+Thus regardless of where the first header is, any Chain Growth violation in the 3k/f slots after it would prevent switch to C when R=k.)
+
+That is a complete answer, but there are a few particulars worth emphasizing.
+
+- No other behavior of the simplified node will change due to a Chain Growth violation.
+  (This is actually false, but only because we use predictions in the leadership check.
+  One option to make this true is to instead use ticking in the leadership check.
```
dnadales, line 152: Why ticking would make this true?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+  Another option is to also characterize here how Chain Growth violations affect the predication-based leadership check, which is even milder than the above: it only manifests when there are zero blocks in some 3k/f slot window, not merely any number of blocks less than k.)
```
dnadales, line 153: I don't get wahy this is the case. Would it be possible to illustrate this?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: Let me try again :) Why do we refer to the `singleEraTransition` identifier? Would my sentence above still be correct if we removed this?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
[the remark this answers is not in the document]
dnadales: So is the problem in a nutshell that the HFC counts blocks whereas the ledger counts slots when evaluating an era change? Are we describing a situation like this one? 

image

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: Thanks for the link to the comment. It's clear now (provided I understood what said comment tried to explain :)).

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: Yes, apparently three of my questions had the same answer :)

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: > The ticking itself does make the vote-counting actually happen

Oh, I thought this was the case. Do you know how we currently determine if we should transition to the next era based on the ledger info? I guess that we do not rely on this rule on epoch change. Do we roll out our own logic (that totally ignores the ledger?).

image

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: Yeah, I think that would help the readers greatly. But I acknowledge this is not intended to be a document used for mass consumption, right? So perhaps this is also out of scope.

dnadales: Yeah, I think that would help the readers greatly. But I acknowledge this is not intended to be a document used for mass consumption, right? So perhaps this is also out of scope.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: > Were there at least $k$ blocks after the voting deadline ($6k/f$ before an epoch boundary)?

But don't we require 2k blocks after the voting deadline to actually transition to the next era?

dnadales: > This choice would have DST=Sensitive

Why? Because we would be counting slots? 

> PIL=Impossible

Because we would be counting slots, like the ledger does?

dnadales: > was the last applied block within the last $3k/f$ slots of the epoch?

and I don't get why this will ensure that a rollback won't cause a different era-transition decision (even if there are no CG violations).

(sorry for the many questions :pray: )

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: Let me see if I get this: if $b^d_1$ is the block at the voting deadline, where the votes are counted, then in the following chain:

$b^d_1 \rightarrow b^d_2 \rightarrow \ldots \rightarrow b^d_k \rightarrow b^d_{k+1}$

The vote counting is not stable until we reach/add? block $b^d_{k+1}$, right?

And does the insensitivity to ticks stem from the fact that we're counting blocks instead of slots?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Because you can only forecast so far, but you can tick indefinitely (in theory, at least). EG if there's a block in the slot 0 of an epoch, but there no block in the next 3k/f slots, then today's node wouldn't be able to lead even if it were elected in the next slot (ie slot 3k/f+1 of the epoch), since it couldn't forecast far enough to see that. On the other hand, it can tick to that slot.

(In theory ticking has no limit, but I think the Cardano ledger _currently_ assumes you can't tick across more than one epoch boundary.)

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: It's currently acquired 3k/f before the end of the epoch. IOG Researchers want us to move it slightly earlier, ie 4k/f before the end.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Yes: Byron was 10k and every era after has been 10k/f slots per epoch. It could vary to other values in later eras, but no one within IOG has yet suggested it should.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: We can tick L up to the next block on their chain. And then we can forecast 3k/f from there.

I think that's actually exactly _one slot short_ 🤔 of ensuring L can validate k+1 headers on a chain that does not violate without having to apply any more blocks to L, unless the sparsity of the k headers after the first violate Chain Growth.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Yep, that's why. Praos promises we'll never need to rollback more than `k`. And it's possible that the tiebreakers align such that we'd need to rollback exactly `k` in order to a longer chain, ie a chain with at least k+1 blocks after the intersection.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Yeah; see our previous convo https://github.com/input-output-hk/ouroboros-consensus/pull/346#discussion_r1333010441

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: 2-stability-windows only suffices to prevent the veto if the chain satisfies Chain Growth.

Without the Chain Growth property, nothing prevents a rollback from rolling back (up to `k`) blocks that collectively span more than one stability window of slots. Or two stability windows. Or any number of them for that matter.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: The deadline is actually a boundary between slots. There's a slot that's the last slot you could sneak in a vote, and the next slot you can't. The boundary between them is the deadline.

If $b^d_1$ is first block immediately before that deadline or anywhere after it, then I think that diagram is correct. Because rolling back up to k blocks would still not allow the votes to change.

---

> The vote counting is not stable until we reach/add? block $b^d_{k+1}$, right?

Hmm. The word "stable" is tricky, since it is often used to refer to something that's one stability window old. But in the presence of CG violations, that could still be subject to rollback.

So it's "actually stable" only once block `k+1` arrives, yes.

> And does the insensitivity to ticks stem from the fact that we're counting blocks instead of slots?

Yep.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: > > The ticking itself does make the vote-counting actually happen
> 
> Oh, I thought this was the case.

I sounds like maybe you read "doesn't" because I wrote "does", which is a bit unusual.

> Do we roll out our own logic (that totally ignores the ledger?)

Here's the entrypoint to the Consensus-side logic https://github.com/input-output-hk/ouroboros-consensus/blob/51da3876c01edc2eec250fdc998f6cb33cdc4367/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs#L130-L164

There's probably plenty of redundancy that could be removed from this Consensus-side logic (in the "`Inspect`" module), but it's only exactly this one line that "ignores the ledger" https://github.com/input-output-hk/ouroboros-consensus/blob/51da3876c01edc2eec250fdc998f6cb33cdc4367/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs#L160 `shelleyAfterVoting` is from the ledger state; it's incremented ([here](https://github.com/input-output-hk/ouroboros-consensus/blob/51da3876c01edc2eec250fdc998f6cb33cdc4367/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L428-L432)) whenever we apply a block.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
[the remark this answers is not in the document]
nfrisby: Yes, I think that's an accurate summary.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: I think "the (single?) era transition method" is too ambiguous.

If you want to remove the identifier, then I would say:

> In this context, the _determination_ is the logic that maps a ledger state in some era to `Just e` if possible, where `e` is the `EpochNo` of the first epoch after that era.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: Claim: "No other behavior of the simplified node will change due to a Chain Growth violation."

It's claiming that the intersection of A and B is empty, where A is "behaviors that only arise in the presence of CG violations" and B is "behaviors that haven't already been discussed".

- If we change the leadership check to use ticking, then CG violations wouldn't affect it, and so the above claim would actually be true, since the leadership check is no longer member of A.

- If we instead also explained that (very extreme) CG violations can affect the leadership check, then the above claim would actually be true, since the leadership check is no longer a member of B.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: > I think "the (single?) era transition method" is too ambiguous.

What are the possible interpretations? (that makes this ambiguous)

I like the alternative phrasing you proposed  :+1:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: To double check: is it true that to judge the validity of headers we need to be within the forecast range?

dnadales: Also, according to my notes, it must always be the case that

```
forecast range <= safe zone
```

This makes sense since we cannot forecast if we can't guarantee a transition won't happen. 

What I still don't get (but it's outside the scope of the PR, but might be related to Alex's comment) is what happens to the forecast range around an era transition.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+The answer to Question 2 does change: there is one additional change in behavior possibly enabled by a Chain Growth violation.
+Unlike the ledger, the hard fork combinator actually counts blocks instead of only slots.
+If a hypothetical roll back of k-many blocks could possibly change the ledger-based governance decision that was intended to trigger the era transition---which is true if Chain Growth was (severely) violated in the last two stability windows before the intended era transition---then the hard fork combinator will not transition to the next era.
+(For this reason, the voting deadline was moved from 7k/f to 4k/f (ie two stablity windows before the end) in the ledger governance rules, so that the hard fork combinator's final decision would be known at least a stability window (and so the safe zone) before the epoch transition.)
+Awkwardly, the ledger will still update the major version of the protocol version protocol parameter, but the chain will remain in the previous era, which many users and developers would find counter-intuitive.
```
dnadales, line 176: This section makes more sense after I read your answers to my questions. Also the second draft helped me a lot.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+This decision may seem even more reasonable when you consider that the opposite question is handled the same exact way: the node will not let the user translate a UTC time to a slot until that answer is also k-rollback-monotonic.
+(Frisby's inference is that UTC time seems so fundamental that it's easy to wrongly assume that the translation back and forth between slots and UTC times does not depend on the evolution of the chain.)
+Requiring this monotonicity for `GetEraStart` and `GetInterpreter` avoids that confusion from manifesting.
+
+## Proposal: Best of Both Worlds
+
+Consider the following alteration of the hard fork combinator.
+
+- Remove the hard fork combinator's second trigger constraint, such that it transitions to the next era exactly when the ledger governance updates the major version of the protocol version protocol parameter (mutatis mutandi as of Conway).
```
dnadales, line 258: What is the first trigger constraint? :grimacing:

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- Only answer `GetEraStart` and `GetInterpreter` queries using an immutable ledger state.
+- (The ledger rules could then return the voting deadline to 7k/f; one stability window before the epoch transition would be sufficient again.)
+
+This would have the exact same answers to Question 1 and Question 2 as does the simplified node, and the Note-To-Client time translation queries would still exhibit k-rollback-monotonicity.
+
+There are three main benefits for the Consensus Team.
+
+- The hard fork combinator would no longer need the block-counting logic.
```
dnadales, line 266: Does this mean that, when checking era transitions we'd be counting slots instead?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
...
+- The Consensus Team would no longer have to justify the "double-stability" for the ledger governance rules.
```
dnadales, line 267: I don't follow this. Even if we rely on the ledger's decision, we'd still need a double stability to ensure time-translations are sound. If not, a rollback could give a different answer to a time-translation.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
dnadales: > I sounds like maybe you read "doesn't" because I wrote "does", which is a bit unusual.

I can't read :facepalm: Sorry about that.

dnadales: Thanks for the links to the relevant snippets :pray: 

> but it's only exactly this one line that "ignores the ledger"

"Ignores the ledger when there aren't `k` blocks after a transition to an era was voted on the ledger, right?

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
nfrisby: That the ledger governance decided the protocol major version would change.

docs/website/contents/for-developers/HardForkCombinatorRollbackMonotonicityProposal.md
```diff
@@ -0,0 +1,285 @@
+# DRAFT NUMBER 2
```
nfrisby, line 1: We suspect we should rename the `.md` file, as of this second draft

dnadales (approved):

Issue #319, opened 2023-08-23
Improve `protocolInfoCardano`


jorisdral:
This issue lists the different steps taken towards improving the `protocolInfoCardano` function

- [x] #275
- [x] #276
- [x] #317 
- [x] #318
- [x] Rename `ProtocolVersion` to `SupportedProtocolVersion` or something similar, since the name has often caused confusion.
- [x] Encode the invariant that protocol versions should be strictly increasing (except for the last era).
- [x] #324 
- [x] https://github.com/input-output-hk/cardano-ledger/issues/3682

amesgen:
FTR: There have been some changes to `protocolInfoCardano` in #377 due to the new `TransitionConfig` concept from Ledger.

dnadales:
@amesgen, when you have time, we should discuss the remaining item in this issue and whether we can close it, as I don't see a sensible renaming.

dnadales:
As discussed out-of-band with @amesgen, with [1224](https://github.com/IntersectMBO/ouroboros-consensus/pull/1224), creating a separate type is no longer necessary.

PR #283, opened 2023-08-09
cardano: update ledger deps

nfrisby:
Updated the ledger deps in `ouroboros-consensus-cardano` to gain the latest Conway changes.

The ledger definitions changed the CBOR encoding of all era's ledger states, not only Conway. Hence there are so much updated golden test files.

This PR also passes through the latest Conway queries supported by ledger.

Fixes #292.

PR #225, opened 2023-07-11
Add golden transactions test to the local transaction submission server

dnadales:
This PR introduces a test to detect errors in the Ledger code, [like this one](https://github.com/input-output-hk/ouroboros-consensus/issues/137), early on. 

Other changes include:

- Addition of a new test library to mock a mempool.
- Addition of a module to elaborate a `ProtocolInfo` for testing purposes. This protocol info is used when opening a Cardano mempool.

PR #352, opened 2023-09-19
Update to ledger-api-1.6 and ledger-conway-1.9

amesgen:
Next major ledger release should be soon-ish and will likely be not too different from what is being integrated here.

 - Most of the diff is due to integrating the new `TransitionConfig` concept from Ledger, see https://github.com/input-output-hk/cardano-ledger/pull/3737.
    - Every Shelley-based era `XEra c` has a `TransitionConfig (XEra c)` that contains all Genesis information up until this era. In particular, one can get all previous `TranslationContext`s and `TransitionConfig`s up until this era from it.
    - They can be used to register initial funds and staking (benchmarking features whose implementation has now been moved to Ledger), and to create initial Shelley ledger states for pure Shelley-based chains.
    - We could keep the current interface to `protocolInfoCardano` the same (by assembling a Conway `TransitionConfig` from the given `TranslationContext`s), but Ledger wants to push the construction of the `TransitionConfig` downwards to Node etc (eg in order to be able to add more information to it, or even consolidate all Genesis files into one).
      This also actually makes things easier on our end, as this allows us to simplify the `ProtocolTransitionParams`.
 - Ledger state serialization and the Conway Genesis file format changed.


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -564,48 +564,58 @@ instance CardanoHardForkConstraints c
...
+    getCardanoHardForkTriggers :: NP (K TriggerHardFork) (CardanoShelleyEras StandardCrypto)
+  }
 
--- | Parameters needed to transition from a Shelley era to the next.
-data instance ProtocolTransitionParams (ShelleyBlock proto era) (ShelleyBlock proto' era') =
-    ProtocolTransitionParamsIntraShelley {
-        transitionIntraShelleyTranslationContext :: Core.TranslationContext era'
-      , transitionIntraShelleyTrigger            :: TriggerHardFork
-      }
+pattern CardanoHardForkTriggers' ::
```
dnadales, line 571: In general I find the `'` problematic when it comes to constructor because it's hard to remember which one is which. But I don't have a good suggestion for an alternative so feel free to ignore this. (Perhaps sth like `MkCardanoHardForkTriggers`?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -56,8 +57,9 @@ module Ouroboros.Consensus.Cardano.Node (
 
 import           Cardano.Binary (DecoderError (..), enforceSize)
 import           Cardano.Chain.Slotting (EpochSlots)
+import qualified Cardano.Ledger.Api.Era as L
```
dnadales, line 60: I wonder if it'd make sense to import this `as Ledger`. Maybe `L` is clear enough, although I do think `Ledger.LatestKnownEra` reads a bit better than `L.LatestKnownEra`.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -726,30 +732,15 @@ protocolInfoCardano paramsCardano
...
-    ProtocolTransitionParamsIntraShelley {
-          transitionIntraShelleyTranslationContext = transCtxtBabbage
-        , transitionIntraShelleyTrigger            = triggerHardForkBabbage
-        } = transitionParamsAlonzoToBabbage
-    ProtocolTransitionParamsIntraShelley {
-          transitionIntraShelleyTranslationContext = transCtxtConway
-        , transitionIntraShelleyTrigger            = triggerHardForkConway
-        } = transitionParamsBabbageToConway
+    transitionConfigs :: NP WrapTransitionConfig (CardanoShelleyEras c)
+    transitionConfigs@(  WrapTransitionConfig transitionConfigShelley
```
dnadales, line 736: Would it make sense to have a pattern similar to `CardanoHardForkTriggers'` here as well?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -1161,24 +1152,52 @@ protocolClientInfoCardano epochSlots = ProtocolClientInfo {
...
+              (transitionConfig ^. L.tcShelleyGenesisL)
+              (transitionConfig ^. L.tcTranslationContextL)
               -- 'completeLedgerConfig' will replace the 'History.dummyEpochInfo'
               -- in the partial ledger config with the correct one.
               History.dummyEpochInfo
               maxMajorProtVer
         , shelleyTriggerHardFork = shelleyTriggerHardFork
         }
+
+class
```
dnadales, line 1173: Could these new classes use some comments?

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/ProtocolInfo.hs
```diff
@@ -307,34 +306,20 @@ mkTestProtocolInfo
...
+              triggerHardForkShelley = hfSpecTransitionTrigger Shelley hardForkSpec
+            , triggerHardForkAllegra = hfSpecTransitionTrigger Allegra hardForkSpec
+            , triggerHardForkMary    = hfSpecTransitionTrigger Mary    hardForkSpec
+            , triggerHardForkAlonzo  = hfSpecTransitionTrigger Alonzo  hardForkSpec
+            , triggerHardForkBabbage = hfSpecTransitionTrigger Babbage hardForkSpec
+            , triggerHardForkConway  = hfSpecTransitionTrigger Conway  hardForkSpec
             }
+          ( L.mkLatestTransitionConfig
+              shelleyGenesis
+              -- tests using this ProtocolInfo do not even reach Allegra
```
dnadales, line 319: If why know why these tests do not reach `Allegra` perhaps we can explain why.

dnadales (approved):

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
lehins: Why do you need `transitionConfigs` at all? 

You ever only need `transitionConfig` for the latest era, since it will contain `transitionConfig` for all previous eras

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
lehins: In my opinion this whole collection of classes is totally unnecessary, but I might be missing something.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
lehins: Why not use a classic Haskell98 record type here? There is nothing special about `TriggerHardFork` at the type level. This would simplify initialization of protocolInfo in db-analyzer and cardano-api 10 fold.
Does is it really bring any more type safety when used using generics sop approach?
In my understanding of this whole setup this does not protect me at all from reading the wrong thing from json config file and mixing up the order.

amesgen: > This would simplify initialization of protocolInfo in db-analyzer and cardano-api 10 fold.

What would get simpler? From a usage perspective, exposing this pattern synonym means that it is exactly as convenient as a plain old record type, users don't have to be aware of the SOP stuff at all. See eg the integration in [cardano-api](https://github.com/input-output-hk/cardano-api/pull/270/files#diff-501ffc6173492d93fcdd4a112b067f1bffb106547a6ca798187c84012599be09R839-R847).

---

Using an `NP` to represent this allows us to know statically what era a trigger is for, which means we can eg use `eraProtVerLow` to figure out the trigger versions automatically, as is done eg in db-analyser since this PR:
https://github.com/input-output-hk/ouroboros-consensus/blob/29c591b9eb6cd2c2bdb803ad3c63cdda7f6f3ed1/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs#L226

This means that in the future, people might not even have to specify the trigger versions (unless they want to override them). Also see #319 for broader goals in making `protocolInfoCardano` simpler to use.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: > Would it make sense to have a pattern similar to CardanoHardForkTriggers' here as well?

We could do that, but it would be only used here, and it is not a type that third parties will use.

> Why do you need `transitionConfigs` at all?
> 
> You ever only need `transitionConfig` for the latest era, since it will contain `transitionConfig` for all previous eras

Sure, but you need the `TransitionConfig` of a specific era in order to register the initial funds/staking for that era, which is where we use `transitionConfigs`:

https://github.com/input-output-hk/ouroboros-consensus/blob/29c591b9eb6cd2c2bdb803ad3c63cdda7f6f3ed1/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L1031

Eg it might be the case that people configure their node to skip ahead directly to say Allegra *and* they specify initial funds/staking. In that case, we need to register those, but as we don't statically know to which era they skip, we need to keep an `NP` of corresponding functions around.

We could also create this `NP` more manually, without the helper classes at the bottom of this file; the current state prevents boilerplate mistakes and automatically adapts to new eras, but it requires a bit of typeclass juggling.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: See the last paragraph in https://github.com/input-output-hk/ouroboros-consensus/pull/352#discussion_r1337399253

Will push a commit that shows how it looks like without these helper classes. I think we could eventually remove most per-era boilerplate in this file by adding even more such classes, but it actually seems more consistent to go the "pedestrial" route for now. :smile:

amesgen: See fdc687428b61a27bec27848e01c6e1b5bd38d270

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: > And you can always get the transitionConfig for that era from the latest one

Sure, that is what we are doing here (a bit more manually since fdc687428b61a27bec27848e01c6e1b5bd38d270).

> you don't need to keep them all around.

I am not exactly sure what you mean by that; for each era, we need a registration function (we don't know which, if any, we have to call), and each of those require a `TransitionConfig` for that era. So each of these functions closures has to refer to one of those `TransitionConfig`s, so they are kept around until initial registration has been performed. Does that make sense or were you talking about sth different?

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/ProtocolInfo.hs
amesgen: Oh thanks for making me double-check that; you did originally factor out this module in #225, but the comment was added in 3df408dc1435b584e72d2565e2fdfd9417172536 (#283), but I think it is incorrect as your [Cardano LocalTxSubmissionServer test](https://github.com/input-output-hk/ouroboros-consensus/blob/a70f0f35594923dc2412edcc76a03bb92c50583c/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/MiniProtocol/LocalTxSubmission/Server.hs) actually go all the way to Conway (in contrast to the ThreadNet test, which indeed never fork into Alonzo). Will correct this comment and make it more precise :+1:

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: Yeah, our qualified names for ledger imports are quite ad-hoc; we even have a ticket for that: https://github.com/input-output-hk/ouroboros-network/issues/3105

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
lehins: I am always in favor of more readable code, even if it results in a little bit more boiler plate. That's why I pushed back on this general solution that was removed in: https://github.com/input-output-hk/ouroboros-consensus/commit/fdc687428b61a27bec27848e01c6e1b5bd38d270

That being said, it is not my place to tell you how to write code, so feel free to choose the solution that you prefer.

amesgen: No, thanks for raising that point, it is especially useful to get feedback on the parts of our API that get used by other teams/projects! What I like about these "general" solutions is that they eliminate potentially error-prone boilerplate, but I agree that it wasn't really worth it here in the scope of this PR.

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/ProtocolInfo.hs
amesgen: 9751e958013a996b7f346a60947ffbf7b9f3827a

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: Fully agreed that the second one is much clearer for someone without sop-core/Consensus experience and should defnitely be preferred in contexts where that is not a given (eg in node, that is also why I did not use this style in eg cardano-api as linked above); but eg one can't rely on types to catch mistakes like accidentally swapping `@Alonzo` and `@Allegra` (don't want to say that that is a huge deal here).

But more generally, the changes in this PR and not requiring anyone to write code like in the first approach (see the cardano-api integration), they can just use the convenient pattern synonym, so I do not see any barrier. We definitely want to make invoking `protocolInfoCardano` as simple as possible to use (#319), and I think in this PR is a step in that direction (could remove quite a few now-redundant lines in the `protocolInfoCardano` invocation in cardano-api).

PR #354, opened 2023-09-19
Move ticked LedgerView out of ticked ChainDepState

nfrisby:
While reviewing PR #340, we realized the following change might make the `ConsensusProtocol` class (and how the HFC depends on it) clearer.

The key idea: let `Ticked ChainDepState` be exactly a ticked `ChainDepState` and nothing beyond that. This requires also introducing `PreparedChainDepState` (and sadly `PreparedHeaderState`) to bear the burden that was formerly on `Ticked ChainDepState`. While we're at it, we also fold the other "determined" argument (`SlotNo`) into the `Prepared*` types, so make the `ConsensusProtocol` methods even simpler.


nfrisby:
The first commit is the core change and everything necessary so that `lib:ouroboros-consensus` builds. The second commit is propagating the change to the rest of the repository.

nfrisby:
The GHC 9.6 warnings in `WithEarlyExit` seem to be https://gitlab.haskell.org/ghc/ghc/-/issues/23143

amesgen:
Summary from my PoV: I think there are a few possible motivations for this PR:

 1. It is a bit weird that before this PR `Ticked (ChainDepState p)` contains `Ticked (LedgerView p)`

This is true, but also not a super important point IMO. Indeed, if we would get rid of `Ticked` as argued above, it would be even less motivating.

 2. The introduction of `HorizonView` in `SingleEraProtocol` becomes very simple with this change.

This was the original motivation to consider this change, but there are alternatives for this purposes, namely the last two bullet points in https://github.com/input-output-hk/ouroboros-consensus/pull/354#issuecomment-1728269639.

 3. This PR makes it hard(er) to provide inconsistent `LedgerView`/`SlotNo` arguments to `tickChainDepState` and then `checkIsLeader`/`(re)updateChainDepState`.

This point is unaffected by whether or not we remove `Ticked`. Is it enough to justify this PR?

PR #355, opened 2023-09-19
Require that LedgerView family is empty

nfrisby:
Enforcing this property as in this PR is not actually necessary, but we suspect it may prevent a good deal of confusion.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/PBFT.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/PBFT.hs
@@ -1,9 +1,12 @@
 {-# LANGUAGE DeriveAnyClass            #-}
 {-# LANGUAGE DeriveGeneric             #-}
+{-# LANGUAGE EmptyCase                 #-}
+{-# LANGUAGE EmptyDataDecls            #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts          #-}
 {-# LANGUAGE FlexibleInstances         #-}
 {-# LANGUAGE MultiParamTypeClasses     #-}
+{-# LANGUAGE LambdaCase                #-}
 {-# LANGUAGE NamedFieldPuns            #-}
 {-# LANGUAGE PatternSynonyms           #-}
 {-# LANGUAGE RecordWildCards           #-}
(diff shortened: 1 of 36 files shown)


nfrisby:
This relatively large diff emerges from merely adding the `invariantLedgerViewEmpty :: Proxy p -> LedgerView p -> a` method to the `ConsensusProtocol` class. Everything else is a consequence of that---no other major decisions were made.

I had originally planned to declare `data LedgerView p` itself as an empty data type. However, our protocol combinators (eg `ModChainSel p s`) currently rely on `LedgerView (ModChainSel p s) ~ LedgerView p`, and that would be impossible with an injective `LedgerView`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -188,27 +191,24 @@ forgePBftFields contextDSIGN PBftIsLeader{..} toSign =
   Information PBFT requires from the ledger
 -------------------------------------------------------------------------------}
 
-newtype PBftLedgerView c = PBftLedgerView {
-      -- | ProtocolParameters: map from genesis to delegate keys.
```
nfrisby, line 192: I should have kept this haddock comment

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -438,7 +440,7 @@ append PBftConfig{} PBftWindowParams{..} =
 -- avoid space leaks.
 data PBftValidationErr c
   = PBftInvalidSignature !Text
-  | PBftNotGenesisDelegate !(PBftVerKeyHash c) !(PBftLedgerView c)
+  | PBftNotGenesisDelegate !(PBftVerKeyHash c) !(Ticked (PBftLedgerView c))
```
nfrisby, line 443: This is the only aspect of the diff that I anticipate trickling down into eg `cardano-node`.

nfrisby (commented):
I think the major alternative to this approach is to replace all occurrences of `Ticked (LedgerView p)` with `LedgerView p`.

Perhaps renaming it to `TickedLedgerView p`?

amesgen:
Another idea: push the evidence that `LedgerView` is empty into a superclass constraint instead of into a new member function, eg a simple `Generic`-based assertion that it has no constructors, or reusing sth like the existing [`Absurd`](https://hackage.haskell.org/package/boring-0.2.1/docs/Data-Boring.html#t:Absurd) class (already has `Generic` deriving).

nfrisby:
I much prefer https://github.com/input-output-hk/ouroboros-consensus/pull/355#pullrequestreview-1633985585, so I'm Closing this.

PR #358, opened 2023-09-21
Cross era ticking

### LINKED ISSUE #345: [FEAT] - Generalize how the HFC handles era transitions
## Background

Today's hard fork combinator design handles transitions from one era to the next (eg Byron to Shelley, Alonzo to Babbage, etc) in a way that has so-far supported the needs of Cardano.

However, Issue #339 results from the current HFC design clashing with the current Conway design. (To be clear: the Conway design is not at fault, in my opinion.)

As part of our work on Issue 339, we've also been considering which assumptions of today's HFC design need to change. EG The draft PR #340 changes Edsko's original `translate*-then-tick` scheme (which made plenty of sense for Byron-to-Shelley!) to a new `(tick-then-translate)*-then-tick` scheme. We think the latter better matches the very reasonable intuition that the current (😞) Ledger and Consensus team members have about "when" era-to-era translations happen. However, we've finally also realized that this new scheme is still very specific to Cardano.

In these discussions around Issue 339, @dnadales early on suggested that the HFC should make even less assumptions. I've come around to that agree with him: the HFC should be as reasonably-general as possible---even beyond Cardano---and we should include helper combinators (eg `(tick-then-translate)*-then-tick`) for instantiating it conveniently for the Cardano use case.

## Task

This Issue is to propose a new generalized interface about how the HFC handles era transitions. We don't necessarily need to implement it, but at least having "_the_ general interface" semi-formalized will help us navigate these kinds of questions, both Issue 339 now and similar issues in the future.

Also, we should discuss that "idealized" interface with at least the Ledger Team, since they ultimately own any design decisions that are required in order to implement each concrete era transition.

nfrisby:
This PR replaces the ledger state translation function and chain-dependent state translation function with corresponding _tick_ functions instead.

The major improvement is that the Hard Fork Combinator no longer imposes a particular scheme for ticking across era boundaries (eg it was only ever using the new era's rules to do so). In particular, this makes it straight-forward to adjust the code to prevent https://github.com/input-output-hk/cardano-ledger/issues/3491.

Closes #345.

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -21,7 +21,7 @@ module Ouroboros.Consensus.Cardano.CanHardFork (
     -- * Re-exports of Shelley code
   , ShelleyPartialLedgerConfig (..)
   , forecastAcrossShelley
-  , translateChainDepStateAcrossShelley
+  , crossEraTickChainDepStateAcrossShelley
   ) where
 
 import qualified Cardano.Chain.Common as CC
@@ -50,6 +50,7 @@ import           Data.SOP.InPairs (RequiringBoth (..), ignoringBoth)
 import           Data.SOP.Strict (hpure)
 import           Data.SOP.Tails (Tails (..))
 import qualified Data.SOP.Tails as Tails
+import           Data.Void (Void)
 import           Data.Word
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
@@ -69,7 +70,6 @@ import           Ouroboros.Consensus.Ledger.SupportsProtocol
                      (LedgerSupportsProtocol)
 import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.Protocol.PBFT (PBft, PBftCrypto)
-import           Ouroboros.Consensus.Protocol.PBFT.State (PBftState)
 import qualified Ouroboros.Consensus.Protocol.PBFT.State as PBftState
 import           Ouroboros.Consensus.Protocol.Praos (Praos)
 import qualified Ouroboros.Consensus.Protocol.Praos as Praos
@@ -80,6 +80,7 @@ import           Ouroboros.Consensus.Shelley.Ledger
 import           Ouroboros.Consensus.Shelley.Node ()
 import           Ouroboros.Consensus.Shelley.Protocol.Praos ()
 import           Ouroboros.Consensus.Shelley.ShelleyHFC
+import           Ouroboros.Consensus.Ticked (WhetherTickedOrNot (..))
 import           Ouroboros.Consensus.TypeFamilyWrappers
 import           Ouroboros.Consensus.Util (eitherToMaybe)
 import           Ouroboros.Consensus.Util.RedundantConstraints
@@ -132,15 +133,17 @@ import           Ouroboros.Consensus.Util.RedundantConstraints
 
 byronTransition :: PartialLedgerConfig ByronBlock
                 -> Word16   -- ^ Shelley major protocol version
-                -> LedgerState ByronBlock
+                -> WhetherTickedOrNot (LedgerState ByronBlock)
                 -> Maybe EpochNo
-byronTransition ByronPartialLedgerConfig{..} shelleyMajorVersion state =
+byronTransition ByronPartialLedgerConfig{..} shelleyMajorVersion wtState =
       takeAny
     . mapMaybe isTransitionToShelley
     . Byron.Inspect.protocolUpdates byronLedgerConfig
-    $ state
+    $ wtState
   where
-    ByronTransitionInfo transitionInfo = byronLedgerTransition state
+    ByronTransitionInfo transitionInfo = case wtState of
+      NoTicked  st -> byronLedgerTransition st
+      YesTicked st -> untickedByronLedgerTransition st
 
     genesis = byronLedgerConfig
     k       = CC.Genesis.gdK $ CC.Genesis.configGenesisData genesis
@@ -193,9 +196,14 @@ byronTransition ByronPartialLedgerConfig{..} shelleyMajorVersion state =
     isReallyStable (BlockNo bno) = distance >= CC.unBlockCount k
       where
         distance :: Word64
-        distance = case byronLedgerTipBlockNo state of
+        distance = case rootBno of
                      Origin                  -> bno + 1
                      NotOrigin (BlockNo tip) -> tip - bno
+
+    rootBno :: WithOrigin BlockNo
+    rootBno = case wtState of
+        NoTicked  st -> byronLedgerTipBlockNo st
+        YesTicked st -> untickedByronLedgerTipBlockNo st
 
     -- We only expect a single proposal that updates to Shelley, but in case
     -- there are multiple, any one will do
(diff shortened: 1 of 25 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -280,22 +284,19 @@ forecastAcrossShelley cfgFrom cfgTo transition forecastFor ledgerStateFrom
...
-      Translate $ \_epochNo (WrapChainDepState chainDepState) ->
-        -- Same protocol, same 'ChainDepState'. Note that we don't have to apply
-        -- any changes related to an epoch transition, this is already done when
-        -- ticking the state.
-        WrapChainDepState $ Proto.translateChainDepState @protoFrom @protoTo chainDepState
+crossEraTickChainDepStateAcrossShelley =
+    CrossEraTickChainDepState
+  $ \Proxy Proxy _cfg cfg' lv' _eno sno cdst ->
+      tickChainDepState cfg' lv' sno
+    $ Proto.translateChainDepState @protoFrom @protoTo cdst
```
nfrisby, line 299: TODO the `translateChainDepState` method should also change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -198,20 +199,22 @@ injectInitialExtLedgerState cfg extLedgerState0 =
...
+          ( State.epochInfoLedger (configLedger cfg)
+          $ hmap (Comp . NoTicked)
+          $ hardForkLedgerStatePerEra targetEraLedgerState
+          )
           cfg
 
     targetEraLedgerState :: LedgerState (HardForkBlock (x ': xs))
     targetEraLedgerState =
         HardForkLedgerState $
+          undefined cfgs {-
```
nfrisby, line 211: TODO

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -222,14 +225,16 @@ injectInitialExtLedgerState cfg extLedgerState0 =
 
     targetEraChainDepState :: HardForkChainDepState (x ': xs)
     targetEraChainDepState =
+        undefined InPairs.requiringBoth firstEraChainDepState {-
```
nfrisby, line 228: TODO

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ticked.hs
```diff
@@ -37,6 +40,8 @@ data instance Ticked () = TickedTrivial
 
 newtype instance Ticked (K a x) = TickedK { getTickedK :: Ticked a }
 
+data WhetherTickedOrNot x = YesTicked !(Ticked x) | NoTicked !x
```
nfrisby, line 43: The use of this type is the most pervasive change.

sop-extras/src/Data/SOP/InPairs.hs
```diff
@@ -99,6 +105,15 @@ hcpure _ f =
   RequiringBoth
 -------------------------------------------------------------------------------}
 
+infixr 2 :..:
```
nfrisby, line 108: I think these are more flexible and easier to read/track than `Requiring` etc.

TODO we should replace all remaining usings of `Requiring` etc

sop-extras/src/Data/SOP/Telescope.hs
```diff
@@ -285,151 +276,78 @@ fromTZ :: Telescope g f '[x] -> f x
...
 -- | Extend the telescope
 --
 -- We will not attempt to extend the telescope past its final segment.
-extend :: forall m h g f xs. Monad m
-       => InPairs (Requiring h (Extend m g f)) xs -- ^ How to extend
-       -> NP (f -.-> Maybe :.: h) xs              -- ^ Where to extend /from/
-       -> Telescope g f xs -> m (Telescope g f xs)
-extend = go
+--
+-- We will not attempt to extend the telescope more than one segment.
```
nfrisby, line 288: This is a significant change.

I think it's justified---at least right now---essentially because the HFC is currently unable to forecast across more than one era boundary. Being able to tick ledger states across multiple era boundaries would technically be possible, but that's not enough for a whole network to keep running. EG Right now even the leadership check relies on forecasting (though it could be changed to use ticking).

It's necessary for this PR, because it allows for the era-transition logic of `extend` (and `align` is similar) to change the type when transitioning. If the function must be able to extend across multiple boundaries, then that argument type must be an endomorphism.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -185,51 +192,70 @@ instance Inject ExtLedgerState where
...
-          (InPairs.requiringBoth
-            (hmap (WrapConsensusConfig . configConsensus) cfgs)
-            (translateChainDepState hardForkEraTranslation))
-          (hpure (fn_2 (\_ st -> st)))
-          (hardForkLedgerStatePerEra targetEraLedgerState)
-          firstEraChainDepState
-
-    targetEraHeaderState :: HeaderState (HardForkBlock (x ': xs))
-    targetEraHeaderState = genesisHeaderState targetEraChainDepState
+    go :: Ext.ExtLedgerState (HardForkBlock (x ': xs)) -> Ext.ExtLedgerState (HardForkBlock (x ': xs))
```
nfrisby, line 213: I think this loop works as-is. It relies on unticking, which is worrying, but relatively harmless, because of the isolation via `LockBox`.

The only rough alternative I've imagined that would likely be preferable is an `NP (ExtLedgerState ByronBlock -.-> ExtLedgerState) xs` argument, I think.

sop-extras/src/Data/SOP/Telescope.hs
amesgen: FTR: We talked about an idea where `tickLedgerState`/`tickChainDepState` would be split into an "initial" part and an "endomorphism" part, ie (names are horrible)
```haskell
initTicking :: SlotNo ->         LedgerState  -> Ticked (LedgerState)
moreTicking :: SlotNo -> Ticked (LedgerState) -> Ticked (LedgerState)
```
subject to a few laws (group action-like), like `moreTicking b = moreTicking b . moreTicking a` for `b >= a` (which allows us to actually state intuitive notions like "ticking to an intermediate slot and then to the final slot is the same as ticking all the way). (Also, `initTicking` should probably be implemented by ticking by one slot, but that might be a bit annoying with EBBs.)

In any case, if we had sth like this, then the cross-era tick functions could take a *ticked* state as input.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
nfrisby: A SanchNet sync would validate whether this works for the interesting multi-era case.

amesgen: > A SanchNet sync would validate whether this works for the interesting multi-era case.

Did so using a db-analyser pass with full block application. (FTR: Sanchonet is configured to hardfork into Alonzo at Genesis).

Also, this validated that the protocol version bug is still present in this PR (as expected, as this is not the job of this PR, it just enables a very simple fix).

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -272,28 +280,28 @@ type CardanoHardForkConstraints c =
...
-        $ PCons translateLedgerStateMaryToAlonzoWrapper
-        $ PCons translateLedgerStateAlonzoToBabbageWrapper
-        $ PCons translateLedgerStateBabbageToConwayWrapper
+      crossEraTickLedgerState =
+          PCons crossEraTickLedgerStateByronToShelley
+        $ PCons crossEraTickLedgerStateAcrossTPraos
+        $ PCons crossEraTickLedgerStateAcrossTPraos
+        $ PCons crossEraTickLedgerStateAcrossTPraos
+        $ PCons crossEraTickLedgerStateTPraosToPraos   -- TODO is there an Alonzo->Babbage bug wrt extraEntropy?
+        $ PCons crossEraTickLedgerStateAcrossPraos   -- TODO fix Babbage->Conway bug
```
amesgen: Why do the `crossEraTickLedgerState*` functions mention the `ConsensusProtocol`s TPraos and Praos? It would make sense for the `crossEraTickChainDepState` functions, but I think the cross-era ledger state ticking should be orthogonal to what `ConsensusProtocol` is used, right?

I pushed 2713ab6c28b809f77258babd48a2f05d2e493cb6 unifying these.

nfrisby: The cast is only needed when the protocol differs, and I considered the concrete types easier to read than the constraints on the polymorphic one.

I was aware that you had unified them on your other PR. I only have a minor preference for the un-unified approach.

PR #360, opened 2023-09-22
Publish `utxo-db.pdf` and `utxo-db-api.pdf` docs on the website

jorisdral:


docs/pdfs/report/.gitignore
dnadales, line 1: What about using `reports` or `tech-reports` for our technical reports. The name `pdfs` might not be very representative of that folder's contents, and would become a misnomer if we write a report in other formats (although it's unlikely I think).

docs/website/sidebars.js
```diff
@@ -23,6 +23,7 @@ const sidebars = {
       label: 'About Ouroboros',
       items: [
         'about-ouroboros/index',
+        'about-ouroboros/utxo-hd',
```
dnadales, line 26: Aren't the reports too low-level for this section? I would expect these to be linked from the "For Developers" section.

docs/pdfs/report/.gitignore
jorisdral: Done! Moved the files

docs/website/sidebars.js
jorisdral: Also, these are only references. The contents of the `.md` file shouldn't be too technical, but I think it's fine to link documents for further reading, so I don't think they have to move

dnadales: > The main report is also listed in `about-ouroboros/References`. Should we move that one too then?

I think that would make sense.

dnadales: In other words my main objection is that it feels strange to have a top level (sidebar) link for a very specific feature.

dnadales (approved):

PR #361, opened 2023-09-22
Babbage->Conway ledger state ticking: use translate-then-tick

### LINKED ISSUE #339: Change implementation of ticking in the HFC to tick-then-translate-then-tick

amesgen:
Closes https://github.com/input-output-hk/cardano-ledger/issues/3491 (tested via a modified db-analyser pass that this fixes the Sanchonet protocol version bug).

Also see #339

Depends on #358

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -285,7 +285,7 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
         $ PCons crossEraTickLedgerStateViaTranslateThenTick
         $ PCons crossEraTickLedgerStateViaTranslateThenTick
         $ PCons crossEraTickLedgerStateViaTranslateThenTick   -- TODO is there an Alonzo->Babbage bug wrt extraEntropy?
-        $ PCons crossEraTickLedgerStateViaTranslateThenTick -- TODO fix Babbage->Conway bug
+        $ PCons crossEraTickLedgerStateViaTickThenTranslate   -- TODO justify
         $ PNil
     , crossEraTickChainDepState =
           PCons crossEraTickChainDepStateByronToShelley
@@ -543,6 +543,30 @@ crossEraTickLedgerStateViaTranslateThenTick =
         , shelleyLedgerState      = nes
         , shelleyLedgerTransition = st
         }
+
+-- TODO tick-then-translate-then-tick would also work, but needs the slot right
+-- after the boundary (probably by passing in the full bound instead of just the
+-- epochNo).
+crossEraTickLedgerStateViaTickThenTranslate ::
+     forall fromEra toEra proto. -- could also be made proto-changing
+     ( ShelleyBasedEra fromEra
+     , SL.TranslateEra toEra ((Ticked :.: LedgerState) :.: ShelleyBlock proto)
+     , fromEra ~ SL.PreviousEra toEra
+     )
+  => CrossEraTickLedgerState
+       (ShelleyBlock proto fromEra)
+       (ShelleyBlock proto   toEra)
+crossEraTickLedgerStateViaTickThenTranslate =
+    CrossEraTickLedgerState
+  $ \cfg cfg' _epochNo sno ->
+      -- Somewhat unintuitively: the second era's ledger rule for ticking across
+      -- the epoch transition is not invoked.
+        fmap ( pureLedgerResult
+             . unComp . unComp
+             . SL.translateEra' (shelleyLedgerTranslationContext cfg')
+             . Comp . Comp
+             )
+      . applyChainTickLedgerResult cfg sno
 
 {-------------------------------------------------------------------------------
   Translation from Shelley to Allegra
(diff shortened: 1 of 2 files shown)


amesgen:
https://github.com/input-output-hk/ouroboros-consensus/pull/358#issuecomment-1758092221

PR #363, opened 2023-09-22
consensus: replace Ticked LedgerView with LedgerView

nfrisby:
Single commit PR; see its message.

This entire diff results from simply replacing every occurrence of `Ticked (LedgerView T)` with `LedgerView T` and then fixing type errors and updating any identifiers that referred to a "ticked ledger view".

### DIFF
--- a/docs/website/contents/for-developers/AbstractProtocol.md
+++ b/docs/website/contents/for-developers/AbstractProtocol.md
@@ -201,7 +201,7 @@ Classes connected to ledgers:
 
   -- | Link protocol to ledger
   class (BlockSupportsProtocol b, UpdateLedger b, ValidateEnvelope b) => LedgerSupportsProtocol b where
-    protocolLedgerView   :: lc → Ticked l → Ticked ledvw   -- 'ledvw' ('LedgerView (BlockProtocol b)') extracted from the ledger
+    protocolLedgerView   :: lc → Ticked l → ledvw          -- 'ledvw' ('LedgerView (BlockProtocol b)') extracted from the ledger
     ledgerViewForecastAt :: lc → l → Forecast ledvw        -- get a forecast (of future 'ledvw's) from a given ledger state.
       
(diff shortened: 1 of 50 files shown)


nfrisby:
If merged, this PR supplants https://github.com/input-output-hk/ouroboros-consensus/pull/355

amesgen (approved):
Thanks, this makes our code base simpler than the approach in #355, and I think that outweighs any (if even any) conceivable added clarity by having the syntactic `Ticked` be wrapped around `LedgerView` :+1: 

---

For context/posterity: In the PR https://github.com/input-output-hk/ouroboros-network/pull/2037 that introduced using `Ticked` for `LedgerView`, there is the following comment:

> > I'm also not really sure what the semantics really are meant to be of a "ticked" ledger view? E.g there is a comment saying "forecasting is equivalent to ticking", which is true.. but it's not clear what slot it should be ticked to here - is this a view from a ticked ledger state, or a ticked view (I guess, a view for a specific slot?)
>
> Not sure I follow this. It's the ledger view protected forward to a particular slot, which by that lemma is equivalent to what the chain tick is doing? The slot is clear no? It's the slot we protected to... should I rephrase this somehow?

https://github.com/input-output-hk/ouroboros-network/pull/2037#issuecomment-623443344

Note that back then, `Ticked` still defined like this (slightly simplified):
```haskell
data Ticked l = Ticked l SlotNo
```
Hence, the main motivation of this PR ("unticked"/"raw" `LedgerView`s not being used anywhere) didn't apply back then.

---

I rebased, fixed a redundant import warning and pushed a commit removing mentions of ledger views being ticked in a few more places (I didn't go through the report). If that looks good, LGT squashed and merged :+1:

nfrisby:
Thanks for finding those sneakier occurrences!

I squashed and rebased and also further strengthened the `Ticked` Haddock. I'm now clicking Merge when ready.

PR #369, opened 2023-09-26
HardWonWisdom: the HFC interface

nfrisby:
We're still working through the most subtle parts, but this seems like a broadly useful artifact to end up with.

Also see the single commit's message.

### DIFF
--- a/docs/website/contents/for-developers/HardWonWisdom.md
+++ b/docs/website/contents/for-developers/HardWonWisdom.md
@@ -348,3 +348,165 @@ The key corresponding parts of the implementation are as follows.
 - The interpretation only depends on the eras' start and end bounds, not on the safe zone.
   The safe zone is instead used in an preceding step to safely determine the least possible value of the first unknown end bound.
   This determination is implemented by the `ouroboros-consensus:Ouroboros.Consensus.HardFork.Combinator.State.reconstructSummaryLedger` function.
+
+## The hard fork combinator interface
+
+To run the Consensus Layer on a block type `blk` requires the following (whether or not it's a `HardForkBlock`).
+
+- Values for various tracers, ChainDB options, etc, as required by `Ouroboros.Consensus.Node.run`.
+- A `ProtocolInfo blk` value, which is `TopLevelConfig blk` as well as initial state of the ledger and protocol.
+- Instances of `ConsensusProtocol (BlockProtocol blk)`, `LedgerSupportsProtocol blk`, and so on (see the `RunNode blk` class).
+
+Suppose there are two blocks types `F` and `G` for which the above is already available (everything will generalize to cases with more than two eras via simple replication).
+To run the Consensus Layer instead on the hard fork combinator that sequences `F` into `G` requires the following additional information about `F` and `G`.
+
+- A value of `EraParams` for `F` and for `G` (ie `Shape [F, G]`), which decomposes as the following for `F` and `G` individually.
+    - The duration of every slot in the era as a [`NominalDiffTime`](https://hackage.haskell.org/package/time-1.12.2/docs/Data-Time-Clock.html#t:NominalDiffTime).
+    - The duration of every epoch in the era as a number of slots.
+    - The era's _safe zone_ is intimately related to the `singleEraTransition` function introduced below, as explained in the next subsection.
+      The safe zone cannot be 0 because---as discussed elsewhere---that would imply a stability window of 0, which is impossible.
+- An instance of `CanHardFork [F, G]`, which decomposes as follows.
+    - A function `LedgerConfig F -> LedgerConfig G -> LedgerState F -> Bound {- starting slot&epoch&time-since-genesis of G -} -> SlotNo -> Except OutsideForecastRange (LedgerView G)`.
+      This function will be used for forecasting the ledger view across the era transition; see "How does cross-era forecasting work?" above.
+    - A function `ConsensusConfig F -> ConsensusConfig G -> EpochNo {- start of G -} -> ChainDepState F -> ChainDepState G`.
+      This function will be applied to the chain-dependent state immediately after the last `F` header.
+      Then the protocol rules for `G` will be used to tick that state to the slot of the first `G` header.
+      This use of `G`'s ticking rules instead of `F`'s to cross the end of the last epoch of `F` is subtle and somewhat counter-intuitive.
+      (We haven't yet had any concrete issues here, unlike with the `LedgerState` issue linked just below.
+      However, we did only realize recently that the extra entropy parameter of Transitional Praos should have been considered when translating to Praos at the Alonzo->Babbage transition.)
+    - A function `LedgerConfig F -> LedgerConfig G -> EpochNo {- start of G -} -> LedgerState F -> LedgerState G`.
+      This function will be applied to the ledger state immediately after the last `F` block.
+      Then the ledger rules for `G` will be used to tick that state to the slot of the first `G` block.
+      This use of `G`'s ticking rules instead of `F`'s to cross the end of the last epoch of `F` is subtle and somewhat counter-intuitive (eg see Issue <https://github.com/input-output-hk/cardano-ledger/issues/3491>).
+    - Instances of `SingleEraBlock` for `F` and `G`, which decomposes as the following for `F` and `G` individually.
+        - Types that represent the ledger and protocol configuration _without_ any `EpochInfo` (see `HasPartialLedgerConfig` and `HasPartialConsensusConfig`).
+        - A function `singleEraTransition :: PartialLedgerConfig blk -> EraParams -> Bound {- era's start -} -> LedgerState blk -> Maybe EpochNo` which must return `Just e` if any extension of this ledger state will end between epoch `e` and its predecessor (ie epoch `e-1` when `0<e` and "genesis" otherwise).
+            - As an immediate observation: note that this function is monotonic along the ledger states of an era on a single chain.
+              It cannot return `Just` for some ledger state on a chain and then `Nothing` for a later ledger state in the same era on the same chain.
+            - Similarly: it must not return an epoch `e` which is less than the epoch in which the era started.
+              (See the third subsection for the corner case in which it may return the starting epoch.)
+            - The logic used in this function must be invariant with respect to ticking.
+              In other words, if the end of the era is unknown or uncertain after the application of a block, then it cannot become certainly known before the next block.
+                - This constraint justifies that the function's type signature does not support applying it to a ticked ledger state; if it hypothetically did, then its output would exactly agree with the pre-image of that ticked ledger state.
+                  (Implementation detail: a ticked HFC ledger state must include `TransitionInfo` derived from the non-ticked ledger state only because the signature of `singleEraTransition` would not otherwise allow a ticked HFC ledger state's contents to determine the `TransitionInfo`.)
+                - Notably, this constraint is trivially true if the function is actually counting blocks in order to achieve its monotonicity with respect to maximum rollbacks without relying on a Chain Growth property.
+            - Recall that an era's `singleEraTransition` is intimately related to its safe zone; see the next subsection.
+
+Some notable components are not listed above because they are determined by existing data about `F` and `G`.
+
+- The content of `TopLevelConfig (HardForkBlock [F, G])` other than the `EraParams` values can be completely determined from the `TopLevelConfig F` and `TopLevelConfig G`.
+- The initial state of the ledger and protocol can be completely determined from the same for `F`.
+
+### The Safe Zone Semantics
+
+An era's safe zone constrains how its `singleEraTransition` can vary as the chain is extended in that era.
+The precisely-worded high-level rule for an era whose safe zone is the positive integer `d` is as follows.
+
+> If a chain's configuration data and blocks so far together determine when the era starts but not when it ends, then at least `d` slots after the latest block (non-EBB) will be in that era.
+> (If the latest block is from an earlier era, then those `d` slots begin with the first of the era.
+> Otherwise, the latest block is in the era and the `d` slots begin immediately after it.)
+
+Consider an arbitrary `[F, G]` chain with 200 `F` blocks followed by 30 `G` blocks (these counts are only real blocks, excluding any EBBs).
+Each block on that chain results in a ledger state in the same era as the block.
+In addition, however, there is also a `LedgerState` at the beginning of each era that is not the result of a block from that era.
+For `F`, it's the initial ledger state (resulting from the genesis "block") and for `G` it's the translation of the last `F` ledger state to `G`.
+Thus, this chain will induce a sequence of 201 `F` ledger states and a sequence of 31 `G` ledger states.
+The `F` safe zone constrains the output of `singleEraTransition @F` when it is applied to that each `LedgerState F` in that sequence, and similarly for `G`.
+(Note that those sequences _exclude_ ledger states that arise from any EBBs interspersed among the real blocks.)
+
+Specifically, for the sequence of ledger states within an era, if the era has a safe zone of `d`, then its `singleEraTransition` function cannot return `Nothing` for a ledger state in that era and `Just e` for the next ledger state in that era unless the epoch `e` is still far enough in the future.
+Note that the second ledger state in that specification necessarily arises from a block in the era, since only the first ledger state in an era does not arise from a block in that era.
+Let `y` be that block; the exact safe zone constraint is `slot(y) + d < firstSlot(e)`.
+(Lastly, recall once more that the ledger state sequence in question is intentionally excluding any arising from EBBs.)
+
+That interpretation of the safe zone constraint explicitly only considers the evolution of `singleEraTransition` _within a single era_.
+In particular, it did not at all constrain the output of `singleEraTransition` for the initial ledger state of the era.
+The rest of this subsection discusses that, including notably that that initial output is always `Nothing` in the fundamental use case, in which the end of an era cannot be known before any of its block exist.
+
+In the fundamental motivating use case of the HFC, `singleEraTransition` will return `Nothing` for the first ledger state of each era, and subsequent governance events within that era on the chain will cause it to return `Just` for the remaining suffix of that chain's ledger states in this era.
+However, there are a total of three possible trajectories of the `singleEraTransition` within an era on some chain.
+
+- EXTENSIBLE trajectory.
+  It starts with `Nothing` and ends with `Just`, as in the fundamental motivating use case.
+  Note that once it becomes `Just` it will always remain `Just` (with the same epoch number `e`) for the rest of this chain, by the monotonicity of `singleEraTransition` within an era of some chain.
+  This is the only case in which the above single-era interpretation of the safe zone applies, because it's the only case that involves both `Nothing` and `Just` in the same era.
+- NEVER trajectory.
+  It starts a `Nothing` and will necessarily remain `Nothing` on the entire chain.
+  For example, `singleEraTransition _cfg _params _start _st = Nothing` is a compatible definition.
+  In this case, any safe zone value would satisfy the high-level rule, because the era never ends---the implementation names this an _indefinite_ safe zone.
+- IMMEDIATE trajectory.
+  It starts as `Just e`.
+  Note that it will always remain `Just e` for the rest of this chain, by the monotonicity of `singleEraTransition` within an era of some chain.
+  The IMMEDIATE trajectory can arise in two ways.
+    - Case IMMEDIATE-INDEPENDENT.
+      It's possible that the configuration data determines the end of the era, entirely independent of the preceding chain.
+      For example, `singleEraTransition cfg _params _start _st = Just $ foo cfg + 42` is a valid definition.
+      In this case, any safe zone value would satisfy the high-level rule, because the end of the era is determined even before the start is.
+      (TODO the HFC is "technically incorrect" because it assumes the safe zone applies even in this IMMEDIATE-INDEPENDENT trajectory, but only in the corner case when the era does not contain even one safe zone of sots, eg it contains no slots, as in the next subsection.)
+    - Case IMMEDIATE-DEPENDENT.
+      It's possible that the preceding chain before the era determines its end.
+      No known chain does this.
+      It is technically allowed by today's HFC, but only by accident.
+      There are two notable subcases; see the next subsection.
+        - Subcase IMMEDIATE-DEPENDENT-BOUND.
+          The determination instead depends only when the previous era ended.
+          This is less obviously exotic.
+          For example, the era could specify that it lasts exactly 10 epochs, regardless of when it starts.
+          In this case, any safe zone value would satisfy the high-level rule, because the end of the era is determined exactly when the start is.
+          (TODO the HFC is "technically incorrect" because it assumes the safe zone applies even in this IMMEDIATE-DEPENDENT-BOUND trajectory, but only in the corner case when the era does not contain even one safe zone of sots, eg it contains no slots, as in the next subsection.)
+        - Subcase IMMEDIATE-DEPENDENT-TRANSLATION.
+          If the determination depends on the result of translating some parts of the last ledger state in the previous era, then the blocks of one era are directly determining the end of a later era.
+          This seems incompatible with the fundamental motivation of the HFC.
+          This case has an intricate relationship with the safe zone.
+          Either the output epoch is at least a safe zone after the start of the era, or else this era starts in epoch 0 and all previous eras both start and end in epoch 0, since that guarantees that the initial ledger state is immediately available, so that the era's end will be determined when the start is, thereby trivially satisfying the high-level safe zone rule.
+          Note that this prefix of eras therefore contains no slots, which is the topic of the next subsection.
+
+Every era of Cardano mainnet today uses the EXTENSIBLE trajectory.
+However, other Cardano chains used for testing and/or benchmarking instead use the IMMEDIATE era for some prefix of eras in order to "skip" them, starting the chain in a later era (see the next subsection).
+Lastly, due to bugs, some mainnet eras were previously using the NEVER trajectory by accident (eg see PR <https://github.com/input-output-hk/ouroboros-network/pull/3754>).
+
+The intended use of each trajectory is as follows.
+
+- Use EXTENSIBLE for eras that are expected to end on the chain, even if they could not actually end today (eg even if their current `singleEraTransition` implementation will never return `Just`).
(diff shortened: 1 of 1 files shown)


docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+It's unclear if it's even desirable for the HFC to allow eras that contain no slots, with one exception: the Cardano system-level benchmarks already rely on this behavior in order to skip a prefix of the chain's eras.
+Indeed, skipping a prefix of eras seems like the most natural use case, since the "genesis block" itself is already not an actual block.
+But there is likely a more direct way to support that use case.
+And so it seems possible and desirable to simplify the HFC interface so that it only applies `singleEraTransition` to a ledger state arising from a block in the current era.
+Something alone the lines of "if an era's end is known before its first block exists, then the era must end before slot 0".
+
+TODO the current code (and Edsko's chapter in the report) sometimes applies the safe zone to the first slot of the epoch.
+How does that relate here?
+Perhaps it suggests that the translated-but-not-actually-ticked state is in fact considered to have been "ticked" up to the beginning of the next era.
+And so `singleEraTransition` is interpreting translated state from the "previous era" as a kind of summary of everything that came before, a la the "genesis block"?
```
nfrisby, line 424: Maybe a big improvement would simply be renaming `translateLedgerState` and `Translate` to `convertToInitialLedgerState`/`reinterpretAsGenesisBlock`/etc.

Would that have made it more obvious that the current era's tick rule won't be used for the end of the epoch?

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+
+- Values for various tracers, ChainDB options, etc, as required by `Ouroboros.Consensus.Node.run`.
+- A `ProtocolInfo blk` value, which is `TopLevelConfig blk` as well as initial state of the ledger and protocol.
+- Instances of `ConsensusProtocol (BlockProtocol blk)`, `LedgerSupportsProtocol blk`, and so on (see the `RunNode blk` class).
+
+Suppose there are two blocks types `F` and `G` for which the above is already available.
+To run the Consensus Layer instead on the hard fork combinator that sequences `F` into `G` requires the following additional information about `F` and `G`.
+
+- A value of `EraParams` for `F` and for `G` (ie `Shape [F, G]`), which decomposes as follows.
+    - The duration of every slot in the epoch as a [`NominalDiffTime`](https://hackage.haskell.org/package/time-1.12.2/docs/Data-Time-Clock.html#t:NominalDiffTime).
```
dnadales, line 364: In the epoch of `F` and `G`?

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+    - The duration of every epoch in the era as a number of slots.
+    - The era's _safe zone_, which depends on the `singleEraTransition` function; see below.
+        - If the era's safe zone is `d`, then `singleEraTransition` cannot return `Just e` for the ledger state resulting from some block `x` (possibly the "genesis block") (possibly from an earlier era, and not even necessarily the previous era) unless there are at least `d` slots between `x` and `e`, ie `slot(x) + d < firstSlot(e)`.
```
dnadales, line 367: What does `e` represent here?

amesgen, line 367: > If the era's safe zone is `d`, then `singleEraTransition` cannot return `Just e` for the ledger state resulting from some block `x` (possibly the "genesis block") (possibly from an earlier era, and not even necessarily the previous era) unless there are at least `d` slots between `x` and `e`, ie `slot(x) + d < firstSlot(e)`.

Hmm, isn't it completely fine for `singleEraTransition` to return `Just e` within one safe zone before `e` starts as long as it was also known before that (ie a ledger state with tip before one safe zone before `e` started also caused `singleEraTransition` to return `Just e`)? Eg our implementation for eg Shelley will happily do that.

I guess one could fix that by talking about the *first time* `singleEraTransition` reports `Just e` (this must happen at least one stability window before `e` starts).

Maybe the definition from the report could be repeated here again (to keep it more high-level):
> **Definition 17.3** (Safe zone)**.** A safe zone is a period of time ahead of a ledger’s tip in which an era transition is guaranteed not to occur if it is not yet known.

Maybe we could try to translate this into a law for `singleEraTransition`, but it would have to talk about rollbacks, which you already do further below.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+        - The fundamental motivation for the HFC combinator is to enable a chain to eventualy seemlessly transition into eras that do not yet exist.
```
amesgen, line 368:
```suggestion
        - The fundamental motivation for the HFC combinator is to enable a chain to eventually seamlessly transition into eras that do not yet exist.
```
Also, is this actually *the* fundamental motivation and not just *one* of them (eg the HFC would still be very useful even if we knew that there won't be any era transitions in the foreseeable future)?

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+- An instance of `CanHardFork [F, G]`, which decomposes as follows.
+    - A function `LedgerConfig F -> LedgerConfig G -> LedgerState F -> Bound {- starting slot&epoch&time-since-genesis of G -} -> SlotNo -> Except OutsideForecastRange (LedgerView G)`.
+      This function will be used for forecasting the ledger view across the era transition; see "How does cross-era forecasting work?" above.
+    - A function `ConsensusConfig F -> ConsensusConfig G -> EpochNo {- start of G -} -> ChainDepState F -> ChainDepState G`.
+      This function will be applied to the chain-dependent state immediately after the last `F` header.
+      Then the protocol rules for `G` will be used to tick that state to the slot of the first `G` header.
+      This use of `G`'s ticking rules instead of `F`'s to cross the end of the last epoch of `F` is subtle and somewhat counterintuitive (eg see Issue <https://github.com/input-output-hk/cardano-ledger/issues/3491>).
```
amesgen, line 380: Maybe remove the link to that issue here, as it only affects ledger states and is repeated just below? Instead, we could write that we haven't yet encountered any problems with the translate-then-tick scheme for `ChainDepState`s (except maybe a incorrect (as in: not-as-specified) handling of non-neutral extra entropy at the TPraos->Praos depending on how you read the ledger specs).

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+                - This constraint justifies that the function's type signature does not support applying it to a ticked ledger state; if it hypothetically did, then its output would exactly agree with the pre-image of that ticked ledger state.
+                  (Implementation detail: a ticked HFC ledger state must include `TransitionInfo` derived from the non-ticked ledger state only because the signature of `singleEraTransition` would not otherwise allow a ticked HFC ledger state's contents to determine the `TransitionInfo`.)
+                - Notably, this constraint is trivially true if the function is actually counting blocks in order to achieve its montonicity with respect to maximum rollbacks without relying on a Chain Growth property.
+
+See also "Which slots/times/epochs can we translate?" above.
+
+Some notable components are not listed above because they are determined by existing data about `F` and `G`.
+
+- The `TopLevelConfig (HardForkBlock [F, G])` can be completely determined from the same for `F` and `G`.
+- The initial state of the ledger and protocol can be completely determined from the same for `F`.
```
amesgen, line 400: Exception here is the skip-ahead benchmarking functionality, which relies on `singleEraTransition` for `F`.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+Consider an era sequence `[F, G, H]`.
+If `G` has no slots, then the latest its end could have became known is once `F`'s end became known.
+Thus, the safe zone would seem to be respected as long as `G`'s safe zone is no larger than `F`'s.
+However, it may be too subtle that the block `x` in the definition of the safe zone is not necessarily from the current era!
+
+It's unclear if it's even desirable for the HFC to allow eras that contain no slots, with one exception: the Cardano system-level benchmarks already rely on this behavior in order to skip a prefix of the chain's eras.
+Indeed, skipping a prefix of eras seems like the most natural use case, since the "genesis block" itself is already not an actual block.
+But there is likely a more direct way to support that use case.
+And so it seems possible and desirable to simplify the HFC interface so that it only applies `singleEraTransition` to a ledger state arising from a block in the current era.
+Something alone the lines of "if an era's end is known before its first block exists, then the era must end before slot 0".
```
amesgen, line 419: :+1:, or phrased froma a different angle: the HFC has to support start from a ledger state that is not in the first era if all skipped eras contain zero slots.

docs/website/contents/for-developers/HardWonWisdom.md
amesgen: I guess it would definitely have helped to signal (at least to me) that something weird is going on (in contrast to "translate", which sounds more principled/generic); but I don't know if these specific names would have helped, eg these functions don't reset the ledger state tip to `Origin` as one might expect in a vacuum.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: That's not actually contradictory to what I wrote. I wrote "can". Sneaky.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: One for each, yes; I'll update the text to be more explicit.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,77 @@ The key corresponding parts of the implementation are as follows.
...
+                - This constraint justifies that the function's type signature does not support applying it to a ticked ledger state; if it hypothetically did, then its output would exactly agree with the pre-image of that ticked ledger state.
+                  (Implementation detail: a ticked HFC ledger state must include `TransitionInfo` derived from the non-ticked ledger state only because the signature of `singleEraTransition` would not otherwise allow a ticked HFC ledger state's contents to determine the `TransitionInfo`.)
+                - Notably, this constraint is trivially true if the function is actually counting blocks in order to achieve its montonicity with respect to maximum rollbacks without relying on a Chain Growth property.
+
+See also "Which slots/times/epochs can we translate?" above.
+
+Some notable components are not listed above because they are determined by existing data about `F` and `G`.
+
+- The `TopLevelConfig (HardForkBlock [F, G])` can be completely determined from the same for `F` and `G`.
```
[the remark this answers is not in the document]
nfrisby: Yes, I should have said "the rest of TopLevelConfig can be"...

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: Well, the HFC was created in the first place in order to solve that problem for Byron-to-Shelley.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: I think maybe I pasted the wrong link here. Let me check.

docs/website/contents/for-developers/HardWonWisdom.md
amesgen: Right, then maybe it was just unclear to me what an era "existing" means, ie you could say that during the development preparing for the Byron-to-Shelley HF, Shelley did already exist (and was fully implemented), but no block in the Shelley era was yet on mainnet. In contrast, at that time, Conway didn't even exist as a concept. Please ignore if you think this is just hairsplitting.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: Done.

docs/website/contents/for-developers/HardWonWisdom.md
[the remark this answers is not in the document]
nfrisby: Done.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: I added a new sentence that mentions extra entropy, for example.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: > eg the HFC would still be very useful even if we knew that there won't be any era transitions in the foreseeable future

What's the use case you're thinking of here? Specifically, why are there multiple separately-implemented eras to being with? Perhaps just for re-use, I guess?

amesgen: Suppose that eg in some future era, everyone is super happy with Cardano and there is no majority that is willing to fund work on a new era that might only bring marginal improvements. Then the HFC is still very useful to allow syncing to this chain from Genesis.

But that probably only occurs to me because I can look back now at several eras; I agree that it definitely was the fundamental motivation for the HFC back then :+1:

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,79 @@ The key corresponding parts of the implementation are as follows.
...
+- An instance of `CanHardFork [F, G]`, which decomposes as follows.
+    - A function `LedgerConfig F -> LedgerConfig G -> LedgerState F -> Bound {- starting slot&epoch&time-since-genesis of G -} -> SlotNo -> Except OutsideForecastRange (LedgerView G)`.
+      This function will be used for forecasting the ledger view across the era transition; see "How does cross-era forecasting work?" above.
+    - A function `ConsensusConfig F -> ConsensusConfig G -> EpochNo {- start of G -} -> ChainDepState F -> ChainDepState G`.
+      This function will be applied to the chain-dependent state immediately after the last `F` header.
+      Then the protocol rules for `G` will be used to tick that state to the slot of the first `G` header.
+      This use of `G`'s ticking rules instead of `F`'s to cross the end of the last epoch of `F` is subtle and somewhat counterintuitive.
+      (We haven't yet had any concrete issues here, unlike with the `LedgerState` issue linked just below.
+      However, we did only realize recently that the extra entropy parameter of Transitional Praos should have been considered when translating to Praos at the Alonzo->Babbage transition.)
```
amesgen, line 382: Are we really sure if the Shelley ledger specs make a statement here, ie if the occurrences of epochs starting/ending are intended to imply a specific semantics at era transitions?

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: OK. I reorganized and I think it's all technically correct now.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: I would so "no". However, this question wasn't even asked originally, and I think it should have been. IE it should have been "considered" as I wrote in this text.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: (And I do discuss the skipping later.)

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: Ah ha! I fixed one problem, but not the whole thing. I'm on it now.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,162 @@ The key corresponding parts of the implementation are as follows.
...
+  In this case, any safe zone value would satisfy the high-level rule, because the era never ends---the implementation names this an _indefinite_ safe zone.
+- IMMEDIATE trajectory.
+  It starts as `Just e`.
+  Note that it will always remain `Just e` for the rest of this chain, by the monotonicity of `singleEraTransition` within an era of some chain.
+  The IMMEDIATE trajectory can arise in two ways.
+    - Case IMMEDIATE-INDEPENDENT.
+      It's possible that the configuration data determines the end of the era, entirely independent of the preceding chain.
+      For example, `singleEraTransition cfg _params _start _st = Just $ foo cfg + 42` is a valid definition.
+      In this case, any safe zone value would satisfy the high-level rule, because the end of the era is determined even before the start is.
+      TODO but `singleEraTransition` cannot be applied without the translate state, so what is the current code doing in this case?
```
nfrisby, line 445: I explored this TODO with the `StaticTransition` type in PR https://github.com/input-output-hk/ouroboros-consensus/pull/364

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,162 @@ The key corresponding parts of the implementation are as follows.
...
+      It's possible that the preceding chain before the era determines its end.
+      No known chain does this.
+      It is technically allowed by today's HFC, but only by accident.
+      There are two notable subcases; see the next subsection.
+        - Subcase IMMEDIATE-DEPENDENT-BOUND.
+          The determination instead depends only when the previous era ended.
+          This is less obviously exotic.
+          For example, the era could specify that it lasts exactly 10 epochs, regardless of when it starts.
+          In this case, any safe zone value would satisfy the high-level rule, because the end of the era is determined exactly when the start is.
+          TODO but `singleEraTransition` cannot be applied without the translate state, so what is the current code doing in this case?
```
nfrisby, line 456: I explored this TODO with the `StaticTransition` type in PR https://github.com/input-output-hk/ouroboros-consensus/pull/364

nfrisby:
Alright. Phew! Esgen pointed out a mistake (and had to twice, even). My fix for it involves a lot more text.

But I think the set of concepts being discussed now are all good ones to establish: a precise wording of the high-level safe zone semantics (matching Edsko's from the report, but slightly more concrete), its extremely concrete interpretation within a single era, the three possible trajectories of transition-knowledge within an era, and lastly the idea that we should remove one of those trajectories (aka `TriggerHardForkAtEpoch`).

In particular, I think this text (modulo the TODO about EBBs and the TODO about refining the semantics of `LedgerState` "translation") is now a complete discussion of the existing HFC interface's types as well as contracts that cover everything those types let you express (even slightly more than that, actually, but it's a small fix to actually establish harmony among the types, ie PR 364's `StaticTransition` type).

docs/website/contents/for-developers/HardWonWisdom.md
amesgen: Ah, I somehow misinterpreted "the extra entropy should have been considered" to take a stance that is should have been used (ie "mixed in" at the TPraos->Praos era/epoch boundary, which would have made a difference if it was non-neutral there); fully agree that the question was not, but should have been considered :+1:

docs/website/contents/for-developers/HardWonWisdom.md
amesgen: I am not exactly sure what "current code" you are referring to. The HFC cross-era ticking logic can just use the translated state, so you probably don't mean that.
Hence, you probably mean forecasting/`Summary` construction etc. as these currently do not translate ledger states, so they use `TransitionUnkown` and hence construct the upper bound of a IMMEDIATE-INDEPENDENT era (before having ticked into it) only based on the safe zone, even though the end is already known statically.

amesgen (commented):
This looks really good, only have the last section left to review.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,165 @@ The key corresponding parts of the implementation are as follows.
...
+    - Instances of `SingleEraBlock` for `F` and `G`, which decomposes as the following for `F` and `G` individually.
+        - Types that represent the ledger and protocol configuration _without_ any `EpochInfo` (see `HasPartialLedgerConfig` and `HasPartialConsensusConfig`).
+        - A function `singleEraTransition :: PartialLedgerConfig blk -> EraParams -> Bound {- era's start -} -> LedgerState blk -> Maybe EpochNo` which must return `Just e` if and only if the arguments determine that only a roll back of more than `k` blocks (which is impossible without disk corruption or manual intervention) could prevent this era from ending between epoch `e` and its predecessor (ie epoch `e-1` when `0<e` and "genesis" otherwise).
+            - As an immediate observation: note that this function is monotonic along the ledger states of an era on a single chain.
+              It cannot return `Just` for some ledger state on a chain and then `Nothing` for a later ledger state in the same era on the same chain.
```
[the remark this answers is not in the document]
nfrisby: I'm not certain that the Consensus Layer _inherently_ requires this property, and so I hesitate to include it here. Thanks for mentioning it, though.

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,165 @@ The key corresponding parts of the implementation are as follows.
...
+
+- The content of `TopLevelConfig (HardForkBlock [F, G])` other than the `EraParams` values can be completely determined from the `TopLevelConfig F` and `TopLevelConfig G`.
+- The initial state of the ledger and protocol can be completely determined from the same for `F`.
+
+### The Safe Zone Semantics
+
+An era's safe zone constrains how its `singleEraTransition` can vary as the chain is extended in that era.
+The precisely-worded high-level rule for an era whose safe zone is the positive integer `d` is as follows.
+
+> If a chain's configuration data and blocks so far together determine when the era starts but not when it ends, then at least `d` slots after the latest block (TODO non-EBB?) will be in that era.
```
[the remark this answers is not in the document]
nfrisby: The HFC is used to validate parts of historical chains that include EBBs. So I don't think we can _completely_ disregard them.

However, I do think we can say something like "in the Byron era, `singleEraTransition` must always return `Nothing` for every ledger state in an epoch that began with an EBB" and leave it at that.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: That's 100% what I meant, yep.

Arguably, this is a bug in the HFC, currently. It contains logic that assumes future eras will contain at least one safe zone's worth of slots, but that's not necessarily true if those eras are configured via `TriggerHardForkAtEpoch` (which induces the IMMEDIATE-INDEPENDENT trajectory).

docs/website/contents/for-developers/HardWonWisdom.md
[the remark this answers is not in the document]
amesgen: We briefly talked about this yesterday, and we should then probably also remove/tweak the mention of rollbacks just above:
> A function `singleEraTransition :: PartialLedgerConfig blk -> EraParams -> Bound {- era's start -} -> LedgerState blk -> Maybe EpochNo` which must return `Just e` if and only if the arguments determine that only a roll back of more than `k` blocks (which is impossible without disk corruption or manual intervention) could prevent this era from ending between epoch `e` and its predecessor (ie epoch `e-1` when `0<e` and "genesis" otherwise).

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,165 @@ The key corresponding parts of the implementation are as follows.
...
+
+### Eras that contain no slots
+
+The HFC interface and contract together permit `singleEraTransition cfg eraParams start st` to return `Just (boundEpoch start)`.
+In that case, the era contains no slots on the chain.
+Since it therefore contains no blocks, `st` must be the era's initial ledger state, and so this equality rules out the EXTENSIBLE trajectory.
+It also cannot be the NEVER trajectory, since `singleEraTransition` is returning a `Just` at all.
+Therefore this equality is only possible for an era using the IMMEDIATE trajectory.
+
+As explained above, the safe zone is trivially satisfied for the IMMEDIATE-INDEPENDENT and IMMEDIATE-DEPENDENT-BOUND cases.
```
amesgen, line 493:
```suggestion
As explained above, any safe zone is trivially valid for the IMMEDIATE-INDEPENDENT and IMMEDIATE-DEPENDENT-BOUND cases.
```

docs/website/contents/for-developers/HardWonWisdom.md
```diff
@@ -348,3 +348,165 @@ The key corresponding parts of the implementation are as follows.
...
+In the IMMEDIATE-DEPENDENT-TRANSLATION case, however, `singleEraTransition` cannot be applied without the initial ledger state of the era, and so the end can't be determined before that initial ledger state is.
+Since in general some chains in the preceding non-empty era will determine the start of this era before they determine its initial ledger state, the IMMEDIATE-DEPENDENT-TRANSLATION case only necessarily satisfies the safe zone when some prefix of the era sequence contains no slots.
+That's when the era's initial ledger state necessarily available as soon as the its start is determined (since it's just a sequence of translations starting from the chain's genesis "block").
```
amesgen, line 496:
```suggestion
That's when the era's initial ledger state is necessarily available as soon as its start is determined (since it's just a sequence of translations starting from the chain's genesis "block").
```
(maybe also 'starting from the initial era's ledger state after the chain's genesis "block"')

amesgen (approved):

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: Done.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: Done.

docs/website/contents/for-developers/HardWonWisdom.md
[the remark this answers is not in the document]
nfrisby: I left it as-is---which is fine for HWW but would need to be addressed when it gets promoted out to a better document.

docs/website/contents/for-developers/HardWonWisdom.md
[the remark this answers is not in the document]
nfrisby: Fixed.

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: I rephrased the TODO text to be much much more precise

docs/website/contents/for-developers/HardWonWisdom.md
nfrisby: I'm leaving this TODO in--its appropriate content for `HardWonWisdom.md`

PR #323, opened 2023-08-25
Test that consensus emits valid CBOR

dnadales:
- Check that the encoders used in the serialization roundtrip are valid.
- Add a unit test to check that the encoders used to encode the Cardano examples are valid.

Needs: https://github.com/well-typed/cborg/issues/324.

Closes https://github.com/input-output-hk/ouroboros-network/issues/3099.

PR #377, opened 2023-09-27
Cut releases for Node 8.5

amesgen:
Also inlcudes changelog entries for #366 and the first commit in #323 (second commit was reverted in #362).


ouroboros-consensus-protocol/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 # Changelog entries
 
+<a id='changelog-0.5.0.7'></a>
+## 0.5.0.7 — 2023-09-27
+
+### Non-Breaking
```
nfrisby, line 8: Isn't this a `Patch` change, not `Non-Breaking`?

ouroboros-consensus-diffusion/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 # Changelog entries
 
+<a id='changelog-0.8.0.1'></a>
+## 0.8.0.1 — 2023-09-27
+
+### Non-Breaking
```
nfrisby, line 8: Also, `Patch`, right? Just based on the version number you chose.

Or maybe the version bump is too gentle.

nfrisby (approved):
Approved. Disclaimer:

- I pointed out that the changelog labels and the actual version changes seem incoherent (patch versus non-breaking), but I trust you to resolve that without another round of review.

- I see that some of the changelog content here is not arriving via `scriv`. I did not check whether what you wrote covers _all_ PRs that were missing their changelog fragments.

ouroboros-consensus-protocol/CHANGELOG.md
amesgen: Ah thanks, this should indeed be `Patch`, will change :+1:

ouroboros-consensus-diffusion/CHANGELOG.md
amesgen: https://github.com/input-output-hk/ouroboros-consensus/pull/377#discussion_r1339685322

amesgen:
> I see that some of the changelog content here is not arriving via `scriv`. I did not check whether what you wrote covers *all* PRs that were missing their changelog fragments.

Thanks for pointing that out, I updated the PR description to mention that.

PR #380, opened 2023-09-29
HardWonWisdom.md: bugfix in HFC time-translation explanation

nfrisby:
See single commit message.


amesgen (approved):
We discussed this on a call.

PR #382, opened 2023-10-02
Rename some types and functions in the Peer simulator of genesis tests

facundominguez:


Niols (approved):
Very neat improvement!

PR #383, opened 2023-10-02
Updates ouroboros-network to fix PeerSharing

bolt12:
# Description

This PR accommodates changes from https://github.com/input-output-hk/ouroboros-network/pull/4644.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
+++ b/ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
@@ -1,0 +1,15 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Updates dependency on `ouroboros-network-api` to `0.6.0`.
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -278,12 +278,12 @@ defaultCodecs :: forall m blk addr.
                 )
               => CodecConfig       blk
               -> BlockNodeToNodeVersion blk
-              -> (addr -> CBOR.Encoding)
-              -> (forall s . CBOR.Decoder s addr)
+              -> (NodeToNodeVersion -> addr -> CBOR.Encoding)
+              -> (NodeToNodeVersion -> forall s . CBOR.Decoder s addr)
               -> NodeToNodeVersion
               -> Codecs blk addr DeserialiseFailure m
                    ByteString ByteString ByteString ByteString ByteString ByteString ByteString
-defaultCodecs ccfg version encAddr decAddr _nodeToNodeVersion = Codecs {
+defaultCodecs ccfg version encAddr decAddr nodeToNodeVersion = Codecs {
       cChainSyncCodec =
         codecChainSync
           enc
@@ -325,7 +325,7 @@ defaultCodecs ccfg version encAddr decAddr _nodeToNodeVersion = Codecs {
 
     , cKeepAliveCodec = codecKeepAlive_v2
 
-    , cPeerSharingCodec = codecPeerSharing encAddr decAddr
+    , cPeerSharingCodec = codecPeerSharing (encAddr nodeToNodeVersion) (decAddr nodeToNodeVersion)
     }
   where
     p :: Proxy blk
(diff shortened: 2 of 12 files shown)


ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
```diff
@@ -0,0 +1,20 @@
...
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Changes `NoPeerSharing` to `PeerSharingDisabled` ([#4644](https://github.com/input-output-hk/ouroboros-network/pull/4644))
```
nfrisby, line 15: - `PeerSharingDisabled` was renamed to `NoPeerSharing` upstream. I don't think that needs to be mentioned in the Consensus change log.
- On the other hand, I would imagine it does mean that the `build-depends` field should be updated with an greater lower bound?

ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Breaking
+
+- Adds `NodeToNodeVersion` to the arguments of `immDBServer` function.
```
nfrisby, line 19: `immDBServer` is not technically part of our interface, since it's in an "unstable-*" package. I therefore think this change should instead be at most a Patch-level item, or even just not mentioned in the changelog.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -278,12 +278,12 @@ defaultCodecs :: forall m blk addr.
...
               -> BlockNodeToNodeVersion blk
-              -> (addr -> CBOR.Encoding)
-              -> (forall s . CBOR.Decoder s addr)
+              -> (NodeToNodeVersion -> addr -> CBOR.Encoding)
+              -> (NodeToNodeVersion -> forall s . CBOR.Decoder s addr)
               -> NodeToNodeVersion
               -> Codecs blk addr DeserialiseFailure m
                    ByteString ByteString ByteString ByteString ByteString ByteString ByteString
-defaultCodecs ccfg version encAddr decAddr _nodeToNodeVersion = Codecs {
+defaultCodecs ccfg version encAddr decAddr nodeToNodeVersion = Codecs {
```
nfrisby, line 286: Looks like this is only called in the `ouroboros-consensus` repo's code in the `ouroboros-consensus-diffusion`'s `Ouroboros.Consensus.Node` module and the ThreadNet tests.

This diff alters the ThreadNet call, but it doesn't alter the `Ouroboros.Consensus.Node` call, since that one is using arguments from the Diffusion Layer packages whose type has recently changed, and which is what motivates the change here in the first place.

nfrisby (changes requested):
My only requested changes are:

- I think a couple of the changelog fragments are incorrect, see my comments there.

- I suspect at least one `build-depends` lower bound needs to be increased.

ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
bolt12: Got it, thanks!

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
bolt12: What are you trying to say?

ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
bolt12: Yes, but should I do that before network patches get a release?

nfrisby: I think so, but I don't know if any such policy is technically established.

A related thought: the PR cannot merge until CI passes, and CI won't pass until such a network release is on CHaP.

bolt12: I think that is alright, the PR is small and can wait to be merged until everything is in place

bolt12:
I believe I need to wait for `plutus` to make a release using most recent version of `nothunks`, before merging

jasagredo (approved):

ouroboros-consensus-cardano/changelog.d/20231002_154149_armandoifsantos_fix_peerSharing_handshake.md
jasagredo: Ouroboros-network is released, and I modified the changelog fragments. I will dismiss this request for changes.

PR #384, opened 2023-10-02
Chain switch trace events: record old and new `SelectView` 

amesgen:


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -565,7 +566,8 @@ data TraceOpenEvent blk =
   | OpenedLgrDB
   deriving (Generic, Eq, Show)
 
--- | Information about the new tip of the current chain.
+-- | Information on having switched our selection to a chain with a
```
nfrisby, line 569: This data type is also used when merely adding a block to our selection. So I think "switch" is a potential misnomer here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -589,8 +591,19 @@ data NewTipInfo blk = NewTipInfo {
       -- chain being A and having a disconnected C lying around, adding B will
       -- result in A -> B -> C as the new chain. The trigger B /= the new tip
       -- C.
+    , newTipSelectView  :: SelectView (BlockProtocol blk)
+      -- ^ The 'SelectView' of the new tip. It is guaranteed that
+      --
+      -- >>> Just newTipSelectView > oldTipSelectView
+      -- True
+    , oldTipSelectView  :: Maybe (SelectView (BlockProtocol blk))
```
nfrisby, line 599: This field name does have some friction with various idioms/expectations.

My gut tells me there are two options for the "right" fix.

OPTION 1 rename `NewTipInfo` to something like `SelectionChangeInfo`. I have at least confirmed that the type only occurs in the `AddedToCurrentChain` and `SwitchedToAFork` tracer constructors.

OPTION 2 define a matching `OldTipInfo` and include it as a second argument in both `AddedToCurrentChain` and `SwitchedToAFork`.

I suppose there is a third option:

OPTION 3 rename `oldTipSelectView` to something like `newTipDefeatedSelectView`, but that is pressing the boundaries of the concept of "new tip info".

nfrisby (commented):
I'm not Approving only because (as you hinted in your Slack message), I think the field name is maybe a little too confusing.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
amesgen: Right, will change to "change" :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
amesgen: I think I will go with OPTION 1 :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -636,7 +649,7 @@ data TraceAddBlockEvent blk =
     -- chain (second fragment).
   | AddedToCurrentChain
       [LedgerEvent blk]
-      (NewTipInfo blk)
+      (SelectionChangedInfo blk)
```
nfrisby, line 652: Once this diff context appear it occurred to me that perhaps the use cases could just get the `SelectView` from the two header chain arguments.

However, Esgen pointed out the call site would then need to also have the `BlockConfig` in scope. (That's certainly always _possible_, but likely often onerous).

nfrisby (approved):
Nice 👍

PR #396, opened 2023-10-03
consensus: rename TriggerHardForkNever to TriggerHardForkNotDuringThisExecution

nfrisby:
See the single commit's message.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Simple.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Simple.hs
@@ -9,6 +9,10 @@ import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 
 -- | The trigger condition that will cause the hard fork transition.
+--
+-- This type is only intended for use as part of a
+-- 'Ouroboros.Consensus.Ledger.Basics.LedgerCfg', which means it is "static":
+-- it cannot change during an execution of the node process.
 data TriggerHardFork =
     -- | Trigger the transition when the on-chain protocol major version (from
     -- the ledger state) reaches this number.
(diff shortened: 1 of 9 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Simple.hs
```diff
@@ -19,6 +19,14 @@ data TriggerHardFork =
...
+    -- | Ledger states in this era cannot determine when the hard fork
+    -- transition will happen.
+    --
+    -- It's crucial to note this era could eventually end via a hard fork
+    -- transition. But that could only happen once it is no longer configured
+    -- to use this setting (eg in a subsequent software release). Therefore,
+    -- this option does /not/ imply that "the era will never end", merely that
+    -- "the era cannot end within this node process before it restarts with
+    -- different software and/or configuration for this era".
+  | TriggerHardForkCannotBeKnown
```
amesgen, line 31: Naming bikeshidding: My first thought on seeing the new name was "why is it not `TriggerHardForkUnknown`?". Is there some connotation of `Unknown` you want to avoid that `CannotBeKnown` does not have? I could see differences if this were a "dynamic" value (part of some mutable state), where `CannotBeKnown` feels like it should be a terminal state, but one might have an `Unknown -> Known` state transition. But `TriggerHardFork` is static configuration data.

In your PR description, you write

> See the single commit's message.

Right now, the commit description is empty, maybe that would be a place to clarify?

amesgen (approved):
Thanks, this is a strict improvement, and might even have prevented bugs such as the one fixed by https://github.com/input-output-hk/ouroboros-network/pull/3754

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Simple.hs
nfrisby: Doh! Thanks for pointing out I forgot the commit message. Fixed.

nfrisby: And I changed the name to `TriggerHardForkNotDuringThisExecution`. Verbose, but very concrete.

PR #411, opened 2023-10-10
Cleanup various things

Niols:
_This PR only makes sense for the Genesis project._

This PR includes various cleanup operations. The commits are somewhat independent and they should speak for themselves. This PR should probably be rebased.


ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PointSchedule.hs
```diff
@@ -151,34 +155,40 @@ data Tick =
 instance Condense Tick where
   condense Tick {active} = condense active
 
-data PointSchedule =
-  PointSchedule { ticks :: [Tick], frags :: Peers TestFrag }
+newtype PointSchedule =
+  PointSchedule {ticks :: NonEmpty Tick}
```
tek, line 159: Initially, we also required each individual peer to have nonempty schedules, but we weren't sure that it was required (also it was painful to work with).
Is there a reason we should choose either of those? An empty fork doesn't make any sense, but the chain generator probably prevents that anyway.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Tests/LongRangeAttack.hs
```diff
@@ -40,9 +40,7 @@ import           Test.Util.TestBlock hiding (blockTree)
 import           Test.Util.Tracer (recordingTracerTVar)
 
 tests :: TestTree
-tests = testGroup "Genesis tests"
-    [ testProperty "blargh" prop_syncGenesis
-    ]
+tests = testProperty "long range attack" prop_syncGenesis
```
facundominguez: :tada:

facundominguez (approved):
Thanks @Niols. The small decoupled commits made these PR rather easy to review.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PointSchedule.hs
tek: The change would be to disallow peer fragments that only consist of the genesis point. Right now that's accepted. The benefit might be that the point schedule generator would get nonempty chains as input, which could make some of the types easier, but maybe some of them more convoluted. I'm unsure what would be better.

PR #421, opened 2023-10-11
Remove the ConnectionThread data type

facundominguez:
This PR removes ConnectionThread and replaces the use async with forkLinkedThread

### DIFF
--- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PeerSimulator/Run.hs
+++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PeerSimulator/Run.hs
@@ -4,7 +4,10 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 
-module Test.Ouroboros.Consensus.PeerSimulator.Run (runPointSchedule) where
+module Test.Ouroboros.Consensus.PeerSimulator.Run (
+    ChainSyncException (..)
+  , runPointSchedule
+  ) where
 
 import           Control.Monad.Class.MonadAsync
                      (AsyncCancelled (AsyncCancelled))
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PeerSimulator/Run.hs
```diff
@@ -288,12 +252,22 @@ runPointSchedule k asc pointSchedule tracer blockTree =
...
     runScheduler tracer pointSchedule resources
-    res <- killAll (Map.elems chainSyncThreads)
+    chainSyncExceptions <- collectExceptions (Map.elems chainSyncRess)
     b <- atomically $ ChainDB.getCurrentChain chainDb
-    pure $ maybe (Right b) Left res
+    pure $ maybe (Right b) Left chainSyncExceptions
   where
     config = defaultCfg k
 
+    collectExceptions :: [StrictTVar m (Maybe SomeException)] -> m (Maybe (NonEmpty SomeException))
```
tek, line 261: I think it would be great if we could bundle the `PeerId` with the exception. today we got a `TimeoutExceeded` and could have used the info in the message!

tek (changes requested):
nice!

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PeerSimulator/Run.hs
facundominguez: Done in d6a6e87

tek: awesome!

tek (approved):

tek:
looks good to merge!

Issue #420, opened 2023-10-11
Proposals for second iteration of the HFC


amesgen:
The most intricate parts of the HFC related to how time/era transitions are performed have been largely unmodified since they were first introduced[^1]. Recently, investigating https://github.com/input-output-hk/cardano-ledger/issues/3491 lead us to scrutinize these aspects, resulting in improving our understanding of the status quo and giving rise to potential improvements.

As the aforementioned issue has been resolved in https://github.com/input-output-hk/ouroboros-consensus/pull/366 (albeit in a relatively ad-hoc fashion), there is no immediate need to make any large HFC changes. Instead, the purpose of this issue is to give an overview over the HFC-related changes that came out of this effort.

### Documentation

 - [x] There is now a [HWW entry for the status quo of the HFC](https://input-output-hk.github.io/ouroboros-consensus/docs/for-developers/HardWonWisdom#the-hard-fork-combinator-interface).
 - [ ] https://github.com/IntersectMBO/ouroboros-consensus/issues/921

### HFC behavior

 - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/345
 - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/416
 - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/417
 - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/413 (likely part of implementing alternatives to block counting)
 - "Block counting" related:
    - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/385
    - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/389 (might supersede the previous point)
 - [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/419

### HFC implementation/testing

 - [ ] https://github.com/IntersectMBO/ouroboros-consensus/issues/1336

[^1]: This is largely due to the fact that they did their job well, with one notable exception: https://github.com/input-output-hk/ouroboros-network/pull/3754

nfrisby:
Alexey and I just had a nice chat, which included the potential of upstreaming the HFC's block counting into the Ledger's governance (only regarding HardFork initiation). He agreed it seems plausible and relatively straight-forward to implement --- and preferable to the HFC ever overriding the Ledger's governance.

[...]

Issue #413, opened 2023-10-11
Reconsider HFC requirement that ticking make a transition become known


amesgen:
Right now, the HFC makes use of the fact that ticking can't let a transition become known:

https://github.com/input-output-hk/ouroboros-consensus/blob/106ce89c51e4b6d144b15d02bf78513e947817e6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs#L130-L131

Right now, this is true due to "block counting" (see eg `ShelleyTransition`). On the other hand, alternatives to block counting (see #389) likely require relaxing this assumption.

For more context, see:
 - [This HWW section](https://input-output-hk.github.io/ouroboros-consensus/docs/for-developers/HardWonWisdom#the-hard-fork-combinator-interface) ("The logic used in this function must be invariant with respect to ticking. [...]").
 - The dimension "DST" in the HWW section added in #346
 - #365 for a draft implementation of this idea. At a high level, `singleEraTransition` now takes a slot to which the ledger state will be ticked next.

PR #428, opened 2023-10-12
Haddocks for certain aspects of the HFC status quo

amesgen:
Related to #420, in particular #413 and #345

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
@@ -18,20 +18,20 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
 
 data EraTranslation xs = EraTranslation {
       -- | For each pair @(x, y)@ of subsequent eras, describe how to construct
-      -- the initial ledger state for @y@ from the last ledger state in @x@.
+      -- the initial ledger state for @y@ from the ledger state resulting from the last block in @x@.
       --
       -- When ticking across an era boundary, the HFC will first invoke this and
       -- then tick the resulting ledger state (in @y@) to the requested slot.
       --
       -- The resulting ledger state must summarize every relevant aspect of what
       -- came before the new era. This is intentionally vague; for example,
-      -- ticking in @y@ might work rather differntly than in @x@, and so certain
+      -- ticking in @y@ might work rather differently than in @x@, and so certain
       -- aspects of the ticking logic of @x@ might need to happen as part of
       -- 'translateLedgerState'. For a concrete example in Cardano, see
       -- 'translateLedgerStateBabbageToConwayWrapper'.
       translateLedgerState   :: InPairs (RequiringBoth WrapLedgerConfig    (Translate LedgerState))       xs
       -- | For each pair @(x, y)@ of subsequent eras, describe how to construct
-      -- the initial chain-dependent state for @y@ from the last chain-dep state
+      -- the initial chain-dependent state for @y@ from the chain-dep state after the last header
       -- in @x@.
       --
       -- When ticking across an era boundary, the HFC will first invoke this and


amesgen:
I can split out the first commit if desired

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
```diff
@@ -17,7 +17,25 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
...
+      -- 'translateLedgerState'. For a concrete example in Cardano, see
+      -- 'translateLedgerStateBabbageToConwayWrapper'.
       translateLedgerState   :: InPairs (RequiringBoth WrapLedgerConfig    (Translate LedgerState))       xs
+      -- | For each pair @(x, y)@ of subsequent eras, describe how to construct
+      -- the initial chain-dependent state for @y@ from the last chain-dep state
+      -- in @x@.
+      --
+      -- When ticking across an era boundary, the HFC will first invoke this and
+      -- then tick the resulting chain-dep state (in @y@) to the requested slot.
     , translateChainDepState :: InPairs (RequiringBoth WrapConsensusConfig (Translate WrapChainDepState)) xs
```
amesgen, line 39: Do we want to rename these? I am a bit unsure whether sth like `constructInitialLedgerState` are really more clear, but OTOH, `translateLedgerState` is maybe too generic?

jasagredo: I would be in favor of renaming this to something more clear, that conveys not only in what context this will be used (as it is now with `translateX`) but also what it actually does (as it is in your proposed `constructInitialLedgerState`).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -113,6 +113,42 @@ deriving anyclass instance
...
+-- @
+--
+-- and we always have
+--
+-- >>> ixSt <= ixTickedSt
+-- True
+--
+-- with inequality exactly if we ticked across an era boundary.
+--
+-- == Ticking alone can't reveal era transitions
```
nfrisby, line 143: The wording of this subheading seems off to me:

> Ticking alone can't reveal era transitions

It seems like there are two things worth explaining/referencing:

- The transition must have been predicted by `singleEraTransition` at least one safe zone before it happened.

- Ticking is currently unable to change the output of `singleEraTransition`.

Right now, the text of this subheading could be interpreted as either of those two things. Is that ambiguity intentional?

Also, the text seems to call the first one a "reflection" of the second one, which I hesitate to agree with.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
```diff
@@ -17,7 +17,25 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
 -------------------------------------------------------------------------------}
 
 data EraTranslation xs = EraTranslation {
+      -- | For each pair @(x, y)@ of subsequent eras, describe how to construct
+      -- the initial ledger state for @y@ from the last ledger state in @x@.
```
nfrisby, line 21:
```suggestion
      -- the initial ledger state for @y@ from the ledger state resulting from the last block in @x@.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
```diff
@@ -17,7 +17,25 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
...
+      --
+      -- When ticking across an era boundary, the HFC will first invoke this and
+      -- then tick the resulting ledger state (in @y@) to the requested slot.
+      --
+      -- The resulting ledger state must summarize every relevant aspect of what
+      -- came before the new era. This is intentionally vague; for example,
+      -- ticking in @y@ might work rather differntly than in @x@, and so certain
```
nfrisby, line 28:
```suggestion
      -- ticking in @y@ might work rather differently than in @x@, and so certain
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
```diff
@@ -17,7 +17,25 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
...
+      -- aspects of the ticking logic of @x@ might need to happen as part of
+      -- 'translateLedgerState'. For a concrete example in Cardano, see
+      -- 'translateLedgerStateBabbageToConwayWrapper'.
       translateLedgerState   :: InPairs (RequiringBoth WrapLedgerConfig    (Translate LedgerState))       xs
+      -- | For each pair @(x, y)@ of subsequent eras, describe how to construct
+      -- the initial chain-dependent state for @y@ from the last chain-dep state
```
nfrisby, line 34:
```suggestion
      -- the initial chain-dependent state for @y@ from the chain-dep state after the last header
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Translation.hs
nfrisby: Yes, I also think renaming to something much more concrete than "translate" is important.

nfrisby (changes requested):
I'm Requesting Changes because:

- I don't think we should delay renaming the `translate*` functions. 
- I think we can eliminate some ambiguity in the new Haddock.
- The above renaming reminded me about the `Ouroboros.Consensus.Protocol.Translate` module; I figure it needs the same treatment you're giving `CanHardFork` here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
amesgen: > Ticking is currently unable to change the output of `singleEraTransition`.

This is what I wanted to talk about here in this subsection, but I didn't phrase it like this as you can't apply `singleEraTransition` to the output of ticking.

I didn't intend to talk about safe zones here (but that could definitely happen in a different subsection, will add that :+1:). In my mind, these two topics are orthogonal:

 - This subsection is about *how* era transitions can become known (currently, only via block application, not via ticking).
 - Safe zones are about *when* an era transition can become known.

> Also, the text seems to call the first one a "reflection" of the second one, which I hesitate to agree with.

Yeah, I didn't want to imply that, hopefully, separating safe zones to its own section will make that more clear.

PR #430, opened 2023-10-12
Enable the generation of multiple adversarial chains for the peer simulator

tek:
Refactors some parts of ChainGenerator to be reused by PeerSimulator

Creates individual functions for the steps involved and removes most of the general logic from the property test

### DIFF
--- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
+++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
@@ -1,0 +1,78 @@
+{-# LANGUAGE NamedFieldPuns  #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Test.Ouroboros.Consensus.Genesis.Setup.Classifiers (
+    Classifiers (..)
+  , classifiers
+  ) where
+
+import           Ouroboros.Consensus.Block.Abstract (SlotNo (SlotNo),
+                     withOrigin)
+import           Ouroboros.Consensus.Config
+import           Ouroboros.Network.AnchoredFragment (anchor, anchorToSlotNo,
+                     headSlot)
+import qualified Ouroboros.Network.AnchoredFragment as AF
+import           Ouroboros.Network.AnchoredFragment.Extras (slotLength)
+import           Test.Ouroboros.Consensus.BlockTree (BlockTree (..),
+                     BlockTreeBranch (..))
+import           Test.Ouroboros.Consensus.PointSchedule
+import           Test.Util.Orphans.IOLike ()
+
+{- | Interesting classification predicates used by the test:
+
+1. the immutable tip should ALWAYS be honest
+2. the selection should be honest IF there is at least a Genesis window after the intersection.
+
+It is possible that, at the end of a test, the selection might be dishonest.
+Consider the following example, where k = 5 and scg = 12:
+
+    slots:  0  1  2  3  4  5  6  7  8  9  10  11  12
+    trunk:  0──1──2──3──4──5──6──7──8──────9──10──11
+                     ╰──4──5──6──7──8──9──────────10
+
+Because of the Limit on Eagerness, the selection is stuck at slot 8 (k-th
+block after the intersection).
+Since the test chains only have 7 and 8 blocks after the intersection, Genesis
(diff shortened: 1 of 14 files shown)


ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/PointSchedule.hs
```diff
@@ -366,6 +390,28 @@ splitFragmentAtSlotNo _ (Empty anchor) =
   (Empty anchor, Empty anchor)
 
 -- | Returns the block number of the given slot.
--- REVIEW: might just be 'blockSlot'?
+-- REVIEW: might just be 'blockSlot'? yes!
 blockSlotNo :: HasHeader b => b -> WithOrigin SlotNo
 blockSlotNo = getTipSlotNo . tipFromHeader
+
+data ScheduleType =
```
facundominguez, line 397: Here or elsewhere it needs to be documented what FastAdversary and Banal types of schedule are.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
```diff
@@ -0,0 +1,78 @@
...
+    slots:  0  1  2  3  4  5  6  7  8  9  10  11  12
+    trunk:  0──1──2──3──4──5──6──7──8──────9──10──11
+                     ╰──4──5──6──7──8──9──────────10
+
+Because of the Limit on Eagerness, the selection is stuck at slot 8 (k-th
+block after the intersection).
+Since the test chains only have 7 and 8 blocks after the intersection, Genesis
+will not be able to make a final decision, since it needs to have enough information
+to determine that one of the fragments is definitely denser within 12 slots, and here
+both fragments have the potential to win the race.
```
facundominguez, line 38: I don't get it. At slot 12, it is clear that the honest chain is denser, so it should be selected.

Moreover, the example violates Length of Competing Chains.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
```diff
@@ -0,0 +1,78 @@
...
+This means that a test will terminate while the selection contains an arbitrary fragment
+whose tip is the block in slot 8.
+
+Without Genesis, if the blocks are sent so that the N+1st slot's blocks arrive after all
+blocks in slot N have arrived, the adversarial fragment will be selected every time because
+it will have k+1 blocks first.
```
facundominguez, line 44: I think it is unclear in which order the different blocks are sent.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
Niols: > Moreover, the example violates Length of Competing Chains.

Indeed. It would be great to have an example without this issue!

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
```diff
@@ -0,0 +1,78 @@
...
+
+So in our test we classify the input data according to the presence of an entire Genesis
+window after the intersection of at least one adversarial chain with the honest chain,
+on both chains.
+-}
+data Classifiers =
+  Classifiers {
+    existsSelectableAdversary :: Bool,
+    genesisAfterIntersection  :: Bool
```
Niols, line 53:
```suggestion
    -- | There are more than k blocks in the alternative chain after the intersection
    existsSelectableAdversary :: Bool,
    -- | There are at least scg slots after the intesection on the honest and in the
    -- alternative chain
    genesisAfterIntersection  :: Bool
```

facundominguez (approved):
Will address my comments in a follow up PR.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/Genesis/Setup/Classifiers.hs
tek: it took us like 15 minutes to write this sentence because we couldn't figure out how to express this clearly 😅

PR #431, opened 2023-10-12
Move Genesis tests to `ouroboros-consensus-diffusion:consensus-test`

Niols:
This PR crucially builds on top of #430 because #430 removes the dependency of Genesis tests into `Test....ChainGenerator.Tests.*` which was blocking a move to a different test suite.


facundominguez (approved):
LGTM

PR #432, opened 2023-10-12
Send `AwaitReply` to prevent timeouts.

Niols:
This PR builds on top of #431.

When a ChainSync server is done serving its chain, it now sends an `AwaitReply` message to the client so as to not get disconnected from too fast (typically, clients wait for 10s for a response, or 80-150s for a response after `AwaitReply`).

As a hack, servers also now return `AwaitReply` when they fall behind the intersection of the client due to the `PointSchedule`'s structure. It is not clear to us why this is accepted by a client.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
@@ -29,10 +29,9 @@ mkCdbTracer tracer =
   Tracer $ \case
     ChainDB.Impl.TraceAddBlockEvent event ->
       case event of
-        AddedToCurrentChain _ NewTipInfo {newTipPoint} _ newFragment -> do
+        AddedToCurrentChain _ NewTipInfo {newTipPoint} _ _ -> do
           trace "Added to current chain"
           trace $ "New tip: " ++ condense newTipPoint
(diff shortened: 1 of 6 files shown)


facundominguez:
I think I need some description of the problem that this PR is trying to solve.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
```diff
@@ -29,10 +29,9 @@ mkCdbTracer tracer =
   Tracer $ \case
     ChainDB.Impl.TraceAddBlockEvent event ->
       case event of
-        AddedToCurrentChain _ NewTipInfo {newTipPoint} _ newFragment -> do
+        AddedToCurrentChain _ NewTipInfo {newTipPoint} _ _ -> do
           trace "Added to current chain"
           trace $ "New tip: " ++ condense newTipPoint
-          trace $ "New fragment: " ++ condense newFragment
```
facundominguez, line 35: Is this change intended?

facundominguez (approved):
LGTM

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
tek: yes, that was too noisy and mostly redundant

PR #433, opened 2023-10-12
Documentation adjustments and some renamings

facundominguez:


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/Classifiers.hs
```diff
@@ -17,46 +17,43 @@ import           Test.Consensus.Network.AnchoredFragment.Extras (slotLength)
...
+otherwise the Genesis node has no chance to advance the immutable tip past
+the Limit on Eagerness.
 
 It is possible that, at the end of a test, the selection might be dishonest.
 Consider the following example, where k = 5 and scg = 12:
 
     slots:  0  1  2  3  4  5  6  7  8  9  10  11  12
     trunk:  0──1──2──3──4──5──6──7──8──────9──10──11
-                     ╰──4──5──6──7──8──9──────────10
+                     ╰──4──5──6──7─────────8───9──10
```
tek, line 31: Our motivation for this configuration was that in the test where we slow down the honest peer, or at least the blocks in each slot arrive synchronously (the part further down that you asked about in the other PR), we are guaranteed that the adversary wins due to having k+1 blocks earlier than the honest peer, if Genesis is disabled (as is currently the case in the test). So we aimed to provide an example that illustrates a point for both situations.

I guess this docstring is now detached from the test after we moved things around.

tek: This is a very contrived example, so I guess it does fail to bring the point across in any case.

But the idea was that we think about a point schedule, since we're motivating a classifier for a point schedule test. In particular, a point schedule where we know that block 9 of the adversary arrives definitely before block 9 of the honest peer, since the adversary sends it in an earlier slot, and the point schedule makes sure that it gets computation time to fully send the block before slot 10, where the honest peer sends its block 9.
Since block 9 is the k+1st block, the immutable tip will advance onto the adversary's fork.

Now with Genesis, the immutable tip wouldn't advance, but the selection would be stuck at block 9 at the end of the test, since a) LoE and b) we don't have a full Genesis window.

We were imagining that just seeing the numbers lined up like that, it would be intuitively apparent that the order of events would break the test as soon as we switch on Genesis, and that's why we're discarding and classifying.
But it seems that doesn't work as well as we hoped, since you were including the Length of competing chains criterion in your analysis 😅 So I'm unsure what the best example would be, if you think it's more useful with the new configuration, I'm fine with it.

facundominguez: Ok, I'm removing the example in 41920cb for now.

Issue #296, opened 2023-03-09
Inventorize ouroboros-consensus symbols used by clients


bartfrenk:
In the effort of defining a more formal interface for the ouroboros-consensus package it it useful to understand which symbols are actually used by users of the package. This issue is a meta-issue to keep track of which for which users this has been inventorized.

- [ ] db-analyser
- [ ] Galois tutorials
- [ ] node
- [ ] db-sync
- [ ] hydra

dnadales:
It's wise to do this after UTxO-HD is merged, otherwise, we'll have a yet bigger changeset to merge.

fraser-iohk:
There's now a (basic) tool for detecting which `ouroboros-consensus` symbols are used in a given package: https://github.com/fraser-iohk/reverse-dep-finder

At the moment, it just generates three basic CSVs: 
1. a list of every symbol usage in every built module 
2. a list of every module imported by every built module
3. a list of every symbol exported by every built module

Once we've decided what we want to do with this information, we can extend the tool to handle those use cases.

Issue #61, opened 2023-05-02
Add hardfork logic support for Voltaire


dnadales:
We should add harfork logic support for Voltaire. The new feature should be merged to `main`. This issue requires the following tasks:

- [ ] Specify what Voltaire governance actions are
    - This will allow us to understand to which actions consensus needs to react. We know we need at least to act on hardfork actions.
- [ ]  Specify (with the ledger team) how/if the ledger state needs to be inspected to decide governance actions.
    - Such a specification should land in a consensus design document.
- [ ] Understand and document how hard-forks are decided
    - We need to understand the current hardfork logic to understand what needs to be changed.
    - In particular, we need to understand what's the role of `proposalReachedQuorum`.
    - As a result of this activity we should improve our current documentation.
- [ ] Understand how [getProposedPPUpdates](https://github.com/input-output-hk/cardano-ledger/blob/c99488882e2231f2442d138ad6ec6e1809dc29c5/eras/shelley/impl/src/Cardano/Ledger/Shelley/Governance.hs#L57-L60) relates to the hardfork mechanics (in particular to `extendToSlot`).
- [ ] Look for tests that exercise `proposalReachedQuorum`.
    - We will need to test the new logic, hence the importance of this item. 
    - We might want to update the documentation.
- [ ] What is the relation between `votedValue` and `protocolUpdates.proposalReachedQuorum`?
   - Are they doing the same thing? Is `votedValue` or the result computed by this function used at all?
- [ ] Understand if there is any relation between `votedValue` and Consensus hard-fork logic.
    - ... or the field that `votedValue` affects, namely `futureProposals`.
- [ ] https://github.com/input-output-hk/ouroboros-consensus/issues/60

JaredCorduan:
* The only governance action that consensus will need to care about, as far as I can tell, is the hard fork. I do not think that consensus cares about any of the other protocol parameters, and the other governance actions are definitively only ledger specific.
[...]

PR #435, opened 2023-10-12
Integrate cardano-ledger for 8.6

teodanciu:
# Description

Updating ledger  (commit: [ddb5a36981d78b59a6ea57525c549836855aaf5f](https://github.com/input-output-hk/cardano-ledger/commit/ddb5a36981d78b59a6ea57525c549836855aaf5f)) for 8.6 release: 
cardano-ledger-conway  1.10.0
cardano-ledger-core  1.8.0.0
cardano-ledger-alonzo- 1.5.0.0
cardano-ledger-babbage 1.5.0.0
cardano-ledger-shelley 1.7.0.0
cardano-protocol-tpraos 1.0.3.6

Main Changes: 
*  Account for the new  `ProtVerAtMost era 8` constraint of ppuProtocolVersionL, by never hard forking in Conway
* `GetCommitteeMemberState` query replacing `GetCommitteeState` - type adjustments 
* The new Drep Pulser changed the LedgerState, so had to  regenerate LedgerState_Conway and ExtLedgerState_Conway for the golden tests to pass


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -161,6 +163,18 @@ class ( Core.EraSegWits era
...
          )
 
+  -- | Get the protocol version out of a 'Core.PParamsUpdate', used to detect
+  -- whether we should perform a HF. This will likely be removed/changed once we
+  -- implement HF enactment in Conway (see
+  -- <https://github.com/input-output-hk/ouroboros-consensus/issues/61>).
+  --
+  -- For now, this always returns 'Nothing' for Conway (see the instance below).
+  getProposedProtocolVersion :: Core.PParamsUpdate era -> Maybe ProtVer
+  default getProposedProtocolVersion ::
```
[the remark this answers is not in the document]
amesgen: I wasn't aware of that, we have lots of existing default method impls: `rg -t hs '  default '` returns 26 matches. Might make sense to add it to the style guide

In any case, this should be purely temporary until #61; I just went this route as it resulted in the smallest diff, but I agree that it is usually best to avoid default method impls.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -401,8 +404,8 @@ instance (ShelleyCompatible proto era, ProtoCrypto proto ~ crypto) => QueryLedge
           SL.queryDRepState st drepCreds
         GetDRepStakeDistr dreps ->
           SL.queryDRepStakeDistr st dreps
-        GetCommitteeState ->
```
teodanciu, line 404: `queryCommitteeState` is now deprecated in cardano-ledger. 

I did read the comment about not removing queries, however, I thought that since this is not in mainnet, it's maybe acceptable to replace it in this way (also I'll be integrating this change in cardano node). 

I'm happy to change it to do it properly, if you think it's necessary.

amesgen: As the query is guarded by an experimental version, it is fine to just remove it :+1:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -196,34 +192,51 @@ defaultApplyShelleyBasedTx globals ledgerEnv mempoolState _wti tx =
       mempoolState
       tx
 
+defaultGetProposedProtocolVersion ::
+  ( EraPParams era,
+    ProtVerAtMost era 8
+  ) => Core.PParamsUpdate era -> Maybe ProtVer
+defaultGetProposedProtocolVersion proposal =
+    strictMaybeToMaybe $ proposal ^. ppuProtocolVersionL
```
teodanciu, line 200: I tried to emulate the formatting of the function above (`defaultApplyShelleyBasedTx`), though I don't exactly understand the rules, and `run-stylish.sh` seems to have to effect no matter the indentation I use.  Please let me know if it's wrong (it looks weird to me)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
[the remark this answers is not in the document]
teodanciu: I added a commit which removes the default implementation - but I can remove it  if it seems not worth it (what with the fix being temporary).

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
amesgen: Looks perfect, thanks!

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -759,7 +762,11 @@ decodeShelleyQuery = do
       (1, 24) -> return $ SomeSecond GetGovState
       (2, 25) -> SomeSecond . GetDRepState <$> fromCBOR
       (2, 26) -> SomeSecond . GetDRepStakeDistr <$> LC.fromEraCBOR @era
-      (1, 27) -> return $ SomeSecond GetCommitteeState
+      (4, 27) -> do
+                  coldCreds <- fromCBOR
+                  hotCreds <- fromCBOR
+                  statuses <- LC.fromEraCBOR @era
+                  return $ SomeSecond $ GetCommitteeMembersState coldCreds hotCreds statuses
```
amesgen, line 769: Nitpick: removing some indentation
```suggestion
        coldCreds <- fromCBOR
        hotCreds <- fromCBOR
        statuses <- LC.fromEraCBOR @era
        return $ SomeSecond $ GetCommitteeMembersState coldCreds hotCreds statuses
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes     #-}
 {-# LANGUAGE DataKinds               #-}
+{-# LANGUAGE DefaultSignatures       #-}
```
amesgen, line 3:
```suggestion
```

amesgen (commented):
Looks good, can you add the corresponding [changelog fragments](https://github.com/input-output-hk/ouroboros-consensus/blob/main/docs/website/contents/for-developers/ReleaseProcess.md#adding-a-changelog-fragment)?

ouroboros-consensus-cardano/changelog.d/20231026_110631_teodora.danciu_cardano_ledger_for_8_6.md
```diff
@@ -0,0 +1,30 @@
...
+### Non-Breaking
+
+ - Updated to newer `cardano-ledger-*` packages:
+    * `cardano-ledger-api-1.7`
+    * `cardano-ledger-core-1.8`
+    * `cardano-ledger-shelley-1.7`
+    * `cardano-ledger-babbage-1.5`
+    * `cardano-ledger-conway-1.10`
+    * `cardano-protocol-tpraos-1.0.3.6`
+- Updated to `typed-protocols-0.1.1`
```
teodanciu, line 23: @amesgen  Added this changelog. 
I'm assuming that he changes needed to make things work with ledger don't need to be included here, since they don't seem to be touching "The surface", but I'm not sure... For example: not triggering the HF in Conway. Would you say this should be added here?

amesgen: > I'm assuming that he changes needed to make things work with ledger don't need to be included here, since they don't seem to be touching "The surface", but I'm not sure...

"The surface" currently still is everything (until we have to time do sth like #296), so in particular, this fragment should be "Breaking" instead of "Non-Breaking" as it changes the types of eg a few queries, which will require changes in `cardano-api`.

> For example: not triggering the HF in Conway.

Not mentioning this is fine I think, you currently can't HF out of Conway anyway as there is no successor era :smile:

teodanciu: Makes sense, I forgot about the breaking change to the query, though I integrated it in cardano-api myself :facepalm: 

I updated now, moving the ledger updates to "breaking" and mentioning the change in the query type

amesgen (approved):

PR #436, opened 2023-10-12
Add a test to check that timeouts are indeed enforced.

Niols:
Builds on top of #432.

Work in progress.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
@@ -1,0 +1,48 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Test.Consensus.PeerSimulator.Tests.Timeouts (tests) where
+
+import           Control.Monad.IOSim (runSimOrThrow)
+import           Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Data.Map as Map
+import           Ouroboros.Consensus.Block (getHeader)
+import           Ouroboros.Consensus.Util.Condense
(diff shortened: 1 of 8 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -0,0 +1,48 @@
...
+    dullSchedule :: TestFrag -> PointSchedule
+    dullSchedule (AF.Empty _) = error "requires a non-empty block tree"
+    dullSchedule (_ AF.:> tipBlock) =
+      let tipPoint = TipPoint $ tipFromHeader tipBlock
+          headerPoint = HeaderPoint $ getHeader tipBlock
+          blockPoint = BlockPoint tipBlock
+          state = Peer HonestPeer $ NodeOnline $ AdvertisedPoints tipPoint headerPoint blockPoint
+          tick = Tick { active = state, peers = Peers state Map.empty }
+      in
+      PointSchedule (tick :| replicate 1000000 tick)
```
tek, line 48: are you sure that's enough

Niols: Damn, you're right! Wonder what would be a good value here; maybe the number of photons in the extragalactic background light (4×10^84)?

Niols: Joke aside, it is actually harmful to put too big a value here because the scheduler and ChainSyncServer will run until the end, not necessarily doing much but still wasting computation power and time. It would be good to figure out a way to get tighter numbers here.

PR #440, opened 2023-10-15
pr-402.md: first draft

nfrisby:
On Thursday Oct 12, we discussed PR #402 during the Consensus Office Hours.

This document summarizes the discussion.

Edit: the document was moved into PR 402 itself, but some of the discussion here is still ongoing.

### DIFF
--- a/docs/DONOTMERGE/pr-402.md
+++ b/docs/DONOTMERGE/pr-402.md
@@ -37,6 +37,7 @@ The primary desiderata are as follows.
 Today's primary options are as follows.
 
 - Use ChainSync to follow the chain, and maintain a second copy of the ledger state.
+    - REQUIREMENT VIOLATION: This can only be implemented in Haskell.
     - REQUIREMENT VIOLATION: This violates the desideratum regarding holding second copy of the ledger state in memory.
       As of UTxO HD, the major portions of those two ledger states would be on disk instead of memory, but there'd still need to be two copies (later development could _maybe_ allow sharing the disk part).
     - To handle rollbacks, the client would also need to maintain its own copy of Consensus's `LedgerDB` abstraction.
@@ -60,6 +61,7 @@ The envisioned longer-term solutions are as follows.
     - This solution would very explicitly violate the design principles of the node, since this is non-trivial logic in the node that is absolutely not required for the core functionality of the blockchain itself.
 
 - If a client is already maintaing the UTxO map (as does `db-sync`), then the release of UTxO HD might allow that app to avoid the UTxO-related heap costs of the ledger state---it could implement its own UTxO HD `BackingStore`.
+    - REQUIREMENT VIOLATION: This can only be implemented in Haskell.
 
 - Release a minimal proxy that uses the existing ChainSync approach above in order to maintain a database mapping blocks (ie slot and hash) to ledger events.
     - This technically violates the second copy of ledger state in memory requirement.
--- a/docs/DONOTMERGE/pr-402.md
+++ b/docs/DONOTMERGE/pr-402.md
@@ -23,7 +23,7 @@ This document captures the key points of the discussion.
 Any complete solution for the needs of the relevant client applications _must_ handle all three of the following tasks.
 
 - staying synced with a node when its selection gains some blocks
-- staying synced with a node when its selection losts some blocks, aka rollbacks
+- staying synced with a node when its selection loses some blocks, aka rollbacks
 - becoming synced with a node
 
 The primary desiderata are as follows.
--- a/docs/DONOTMERGE/pr-402.md
+++ b/docs/DONOTMERGE/pr-402.md
@@ -1,0 +1,104 @@
+# How to provide ledger events to client apps
+
+## Introduction
+
+On Thursday Oct 12, we discussed [PR #402](https://github.com/input-output-hk/ouroboros-consensus/pull/402) during the Consensus Office Hours.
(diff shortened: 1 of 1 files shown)


nfrisby:
I am unable to request review from @KtorZ @kderme @koslambrou, so please let this ping suffice to ask you to see if I got anything wrong versus what we discussed in the call.

cc @dnadales @amesgen

nfrisby:
@abailly-iohk @dcoutts @jasagredo please comment with any questions that this summary piques from you---it may be that we discussed them and I forgot to include it in this first draft.

docs/DONOTMERGE/pr-402.md
```diff
@@ -0,0 +1,108 @@
...
+    - Then the ChainSync approach above could simply serve ledger events alongside the blocks.
+    - This solution would very explicitly violate the design principles of the node, since this is non-trivial logic in the node that is absolutely not required for the core functionality of the blockchain itself.
+    - Also: what happens if the specification of ledger events for some block changes (eg a new ledger event is added)?
+
+- If a client is already maintaing the UTxO map (as does `db-sync`), then the release of UTxO HD might allow that app to avoid the UTxO-related heap costs of the ledger state---it could implement its own UTxO HD `BackingStore`.
+    - REQUIREMENT VIOLATION: This can only be implemented in Haskell.
+
+- Release a minimal proxy that uses the existing ChainSync approach above in order to maintain a database mapping blocks (ie slot and hash) to ledger events.
+    - This technically violates the second copy of ledger state in memory requirement.
+      However, the database should be reused among clients, which would amortize that RAM cost.
```
kderme, line 69: There is an argument to be made about the second copy of ledger, which applies not only to this solution, but to all related. In the context of UTxO-HD and for a client that only reapplies blocks, its version of the UTxO doesn't need to be complete. In principle and if we're lucky in code already, calculating the ledger state and the events only ever needs the stake credentials of the UTxO.

I believe this observation needs further investigation with the ledger team.

nfrisby: > calculating the ledger state and the events only ever needs the stake credentials of the UTxO

This seems incorrect me. Eventually, calculating a ledger state requires traversing the entire UTxO map in order to compute the new stake distribution, in order to compute the upcoming leader schedule. I don't see a way around that.

Maybe I'm overlooking something you implied with the "In the context of UTxO-HD" qualifier on your statement?

docs/DONOTMERGE/pr-402.md
```diff
@@ -0,0 +1,108 @@
...
+
+- staying synced with a node when its selection gains some blocks
+- staying synced with a node when its selection loses some blocks, aka rollbacks
+- becoming synced with a node
+
+The primary desiderata are as follows.
+
+- The client can be implemented in any language.
+- The client has access to the per-block ledger events, or at least the ones known to be desirable (eg calculated rewards and calculated stake distributions).
+- The client need not hold its own copy of the ledger state in memory.
```
amesgen, line 33: Maybe we should already state here to what extent this is addressed by UTxO HD (probably just that is is *partially* addressed).

docs/DONOTMERGE/pr-402.md
kderme: The key observation here is that it needs to compute the **stake** distribution, which implies the need only of the staking part.

docs/DONOTMERGE/pr-402.md
```diff
@@ -0,0 +1,114 @@
...
+      The only workaround at the moment is to restart the node and force it to re-apply whichver historical blocks the client needs to process.
+      Moreover, currently, the only way to force the node to do so is to delete certains parts of its on-disk state!
+    - To handle rollbacks, the client would also need to maintain something akin to Consensus's `LedgerDB` abstraction, but more lightweight since its elements need not be entire Cardano ledger states.
+    - @kderme also remarked that `db-sync` only uses the ledger events for so much, and so still relies on the ledger state itself for the rest; this suggests that the blocks and their ledger events alone might not suffice for clients.
+
+The envisioned longer-term solutions are as follows.
+
+- Enrich the node to permanently store ledger events alongside blocks.
+    - Then the ChainSync approach above could simply serve ledger events alongside the blocks.
+    - This solution would very explicitly violate the design principles of the node, since this is non-trivial logic in the node that is absolutely not required for the core functionality of the blockchain itself.
```
koslambrou, line 64: Can't we make the same argument with the state-query mini protocol? Seems to me like something that is also not required for the core functionality of the blockchain.

amesgen: That is a good point, I think the word "non-trivial" in 

> this is non-trivial logic in the node that is absolutely not required for the core functionality of the blockchain itself

is doing the heavy lifting here, with the main difference between on-disk storage of ledger events and the local state query protocol being that the latter does not require any additional state/storage beyond what the node already maintains for other reasons. 

Definitely makes sense to expand a bit on this :+1:, additional considerations might include: 

 - Maybe certain information available via the state query protocol are crucial to get directly from a node for SPOs (eg the leadership schedule indirectly).
 - Make "non-trivial logic" more precise, eg the state query protocol still has *some* maintenance costs, like ensuring backwards-compatible serialization, and hence is not "trivial", but still "easy enough" to be acceptable into the Node. But maybe it could be even simpler?

nfrisby:
This file was moved into PR #440 so as to stop confusing people.

The above comments from Kostas and Konstantin were not integrated before that happened.

PR #445, opened 2023-10-23
Update Cabal cache for CI to pass

jasagredo:
- Update Cabal cache
- Update badges in README and add a ci check so that they are always up to date


amesgen (approved):
Thanks, let's hope that this fixes the CI issue on `main`; as @jorisdral says, we could even completely disable the `dist-newstyle` caching if this comes up again and again.

Also thanks for the badge check; it is too easy to forget this otherwise!

PR #450, opened 2023-10-23
Convert FastAdversary to a frequency-based generator

tek:


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
```diff
@@ -23,17 +23,21 @@ import           Test.Util.Orphans.IOLike ()
...
   unsafeMapSuchThatJust do
     gt <- genChains numAdversaries
     pure $ ((gt,) <$> genSchedule scheduleType (gtBlockTree gt))
 
-prop_longRangeAttack :: QC.Gen QC.Property
-prop_longRangeAttack = do
-  (genesisTest, schedule) <- genChainsAndSchedule 1 FastAdversary
+prop_longRangeAttack :: Int -> [Int] -> QC.Gen QC.Property
+prop_longRangeAttack honestFreq advFreqs = do
+  (genesisTest, schedule) <- genChainsAndSchedule (fromIntegral (length advFreqs)) (Frequencies freqs)
```
Niols, line 40: Haskell question: why is `fromIntegral` necessary here?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -186,22 +185,28 @@ prettyPrint blockTree = do
...
+
+    branch :: SlotNo -> AF.AnchoredFragment blk -> (String, SlotNo, SlotNo)
+    branch oSlotNo frag =
+      (blocks, aoSlotNo, ahSlotNo)
+      where
+        blocks =
+          frag
+            & AF.toOldestFirst
+            & map (\block -> (fromIntegral (unSlotNo (blockSlot block) - unSlotNo aoSlotNo - 1), Just (unBlockNo (blockNo block))))
+            & Vector.toList . (Vector.replicate (fromIntegral (unSlotNo ahSlotNo - unSlotNo aoSlotNo)) Nothing Vector.//)
```
Niols, line 204: Frankly, I'm not even exactly sure what this is doing; it would be worth documenting, or even rewriting completely.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -149,19 +150,17 @@ findPath source target blockTree = do
...
-  let (aoSlotNo, _) = slotAndBlockNoFromAnchor $ AF.anchor advFragment
-  let (ahSlotNo, _) = slotAndBlockNoFromAnchor $ AF.headAnchor advFragment
+      (advBranches, aoSlotNos, ahSlotNos) = unzip3 (branch oSlotNo . btbSuffix <$> btBranches blockTree)
 
-  let firstSlotNo = min oSlotNo aoSlotNo
-  let lastSlotNo = max hSlotNo ahSlotNo
+  let firstSlotNo = minimum (oSlotNo : aoSlotNos)
+  let lastSlotNo = maximum (hSlotNo : ahSlotNos)
 
   -- FIXME: only handles two fragments at this point. not very hard to make it
```
Niols, line 165: This comment is out of date. (Sorry, I can't make a suggestion on those lines as the last one isn't in the diff.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -149,19 +150,17 @@ findPath source target blockTree = do
 --                      ╰─────3──4─────5
 --
 -- Returns a list of strings intended to be catenated with a newline.
-prettyPrint :: AF.HasHeader blk => BlockTree blk -> [String]
+prettyPrint :: forall blk . AF.HasHeader blk => BlockTree blk -> [String]
```
Niols, line 153: I see we both generalised this. The techniques are different but the result is the same, I believe. I personally don't care as long as it is generalised!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
```diff
@@ -23,17 +23,21 @@ import           Test.Util.Orphans.IOLike ()
 import           Test.Util.TestBlock (TestBlock, unTestHash)
 
 tests :: TestTree
-tests = testProperty "long range attack" prop_longRangeAttack
+tests =
+  testGroup "long range attack" [
+    testProperty "one adversary" (prop_longRangeAttack 1 [10]),
+    testProperty "three adversaries" (prop_longRangeAttack 1 [2, 5, 10])
+  ]
```
Niols, line 30: What about one adversary with frequency 2? Or 1.5? How do they behave? Are they sufficient to make the test pass?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -174,6 +177,13 @@ data Peer a =
 instance Functor Peer where
   fmap f Peer {name, value} = Peer {name, value = f value}
 
+instance Foldable Peer where
+  foldr step z (Peer _ a) = step a z
+
+instance Traversable Peer where
+  sequenceA (Peer name fa) =
+    fa <&> Peer name
```
Niols, line 185: Nice!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -261,6 +271,22 @@ pointSchedulePeers :: PointSchedule -> NonEmpty PeerId
...
+-- | Construct 'Peers' from values, adding adversary names based on the default schema.
+-- A single adversary gets the ID @adversary@, multiple get enumerated as @adversary N@.
+mkPeers :: a -> [a] -> Peers a
+mkPeers h as =
+  Peers (Peer HonestPeer h) (Map.fromList (mkPeer <$> advs as))
+  where
+    mkPeer (pid, a) = (pid, Peer pid a)
+    advs [a] = [("adversary", a)]
+    advs _   = zip enumAdvs as
+    enumAdvs = (\ n -> PeerId ("adversary " ++ show n)) <$> [1 :: Int ..]
```
Niols, line 288: Also nice!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -420,41 +413,28 @@ banalPointSchedule blockTree =
...
--- It uses 'banalStates' to convert each fragment to a schedule (which advances by one block per
--- tick).
-fastAdversaryPointSchedule ::
-  BlockTree TestBlock ->
-  Maybe PointSchedule
-fastAdversaryPointSchedule blockTree =
-  peer2Point frags (foldGenPeers frags trans)
+-- This is implemented by initializing each peer's schedule with 'banalStates' (which advances by
+-- one block per tick) and assigning interval lengths to each peer tick based on the frequency
+-- config in the first argument, then sorting the resulting absolute times.
```
Niols, line 418: I do not understand this. What is the “length of a tick”? What is an “interval length”?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
tek: `genChainsAndSchedule` takes a `Word`

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
tek: 😳 wtf 😅

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
tek: yeah I thought I'll leave it because the "in a clean way" part was still true 😅

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
tek: forgot to change the `<&>` back to `<$>` though 😁

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
tek: yeah I'm not so happy about the whole `PeerId` thing, that invariant for formatting the adversary name is depended upon in multiple locations

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
tek: discussed in sync meeting: they have no meaning beyond ordering the ticks in the generator, but we could use the new tick duration config from the other PR to unify the concept

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
Niols: Ah dayum.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
Niols: I'd assume that too, but they probably need the prefix to be long enough or something and that might be a function of `k`?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: Maybe we should just make it an `Int`?

Niols:
After discussing in a pairing session, we realised that the current design actually means that several adversaries may prevent each other from succeeding. Here is an example.

Take `k = 2`. Consider the following block tree with one trunk and two alternative branches:

```
slots:   0  1  2  3  4  5  6  7  8
trunk:   0─────1──2─────3─────4──5
alt. 1:        │  ╰──3─────4
alt. 2:        ╰─────2─────3─────4
```

Assume now that you have one honest node serving the trunk, a first adversary twice faster than the honest node serving the first alternative chain and second adversary four times faster than the honest node. Then we might still end up selecting the trunk (aka. honest chain):

- The first adversary serves their chain very fast, such that we end up selecting block 4 at slot 6.

- The second adversary serves their chain pretty fast, thus serving block 4 at block 8 reasonably fast, but we cannot select it because it would require rolling back by more than `k = 2`.

- The honest node finally serves their chain. It would be too late to roll back from the second alternative chain but it is not to roll back from the first alternative chain; we therefore end up selecting the trunk again.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
tek: that would be better.
it still provides no means of safely referencing the "right" peer with the frequencies, since peers are generated randomly.
don't think we can build a good general framework with this interface.
probably should be a function `Peers TestFrag -> Peers Int` to determine the frequencies, so we don't have to validate afterwards

tek: but also we shouldn't sink much time into this, since the slow-honest test variant isn't that useful for us right now

tek:
let's merge this with the new test disabled, agreed?

FTR we fixed a bug in `genChains` where forks off the same block would have the same fork number.
the fix is very ad-hoc and unsafe, but not more so than it already was due to the incompatibility of `TestBlock`'s intended constructors and our `BlockTree` variant, so I think it's find until we devise a more suitable data type.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -85,14 +86,16 @@ data SchedulerConfig =
 
     -- | Config shared with point schedule generators.
     , scSchedule        :: PointScheduleConfig
+    , scDebug           :: Bool
```
facundominguez, line 89: Document the effect of `scDebug`?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -449,41 +440,28 @@ banalPointSchedule blockTree =
...
+frequencyPointSchedule :: Peers Int -> BlockTree TestBlock -> Maybe PointSchedule
```
facundominguez, line 446: The `freqs` parameter can use more of an explanation. How does the integer of each peer affect the resulting point schedule?

facundominguez (approved):
I didn't try to understand the changes in detail, but added some documentation requests.

PR #452, opened 2023-10-24
Add a test to check that timeouts are indeed enforced

Niols:
aka the return of https://github.com/input-output-hk/ouroboros-consensus/pull/436

This pull requests introduces a test to check that timeouts are indeed enforced. This induces some other work needed to achieve this goal:

- We introduce a notion of “peer simulator state view”, which allows to inspect the state of the peer simulator. For now, this view only shows the selected chain (already used in the long range attack test) and the `ChainSync` exceptions (needed in the new timeout tests). We will add there all additional information that we need eventually. This state view is currently only used at the very end of the test but we could imagine (LTL-like?) properties that rely on the state at other moments in the execution. It could also serve as a basis to build “dynamic” point schedule generators.

- We generalise block tree printing to support an arbitrary number of branch. It work previously on only exactly one branch, which wasn't great for tests featuring more branches and would crash for tests featuring no branches (only a trunk) such as the timeout test. This change isn't really related and could be extracted to another PR.

- We split `runTest` in two variants, one that requires to define everything (in particular it does not force exceptions to be failures) and one that is currently used in the long range attack test. We might refine this second one or decide that it is in fact useless later in the future. We shall see when adding more tests. We also generalise `SchedulerConfig` to hold timeouts instead of just a boolean and we expose it to the tests, leaving them in charge of deciding which timeouts they want. Again, we might want to add helpers for those in the future.

- Finally, we add a timeout test. It features only one peer and a dull schedule that advertises the whole chain from the start but has just too many ticks, causing a timeout. It is actually fun to add a `- 1` to the number of ticks and see that the test then gets falsified.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -27,6 +27,7 @@ import           Test.Util.Tracer (recordingTracerTVar)
 import Test.Consensus.Genesis.Setup.GenChains
 import Test.Consensus.PeerSimulator.StateView
 import Ouroboros.Network.Protocol.ChainSync.Codec (ChainSyncTimeout(..))
+import Test.Consensus.PeerSimulator.Trace (traceLinesWith)
 
 runTest ::
   (IOLike m, MonadTime m, MonadTimer m, Testable a) =>
@@ -38,15 +39,16 @@ runTest ::
 runTest schedulerConfig genesisTest schedule makeProperty = do
     (tracer, getTrace) <- recordingTracerTVar
 
-    traceWith tracer $ "Security param k = " ++ show gtSecurityParam
-    traceWith tracer $ "Honest active slot coefficient asc = " ++ show gtHonestAsc
-    traceWith tracer $ "Genesis window scg = " ++ show gtGenesisWindow
-
-    traceWith tracer $ "SchedulerConfig:"
-    traceWith tracer $ "  ChainSyncTimeouts:"
-    traceWith tracer $ "    canAwait = " ++ show (canAwaitTimeout scChainSyncTimeouts)
-    traceWith tracer $ "    intersect = " ++ show (intersectTimeout scChainSyncTimeouts)
-    traceWith tracer $ "    mustReply = " ++ show (mustReplyTimeout scChainSyncTimeouts)
+    traceLinesWith tracer [
+      "Security param k = " ++ show gtSecurityParam,
+      "Honest active slot coefficient asc = " ++ show gtHonestAsc,
+      "Genesis window scg = " ++ show gtGenesisWindow,
+      "SchedulerConfig:",
+      "  ChainSyncTimeouts:",
+      "    canAwait = " ++ show (canAwaitTimeout scChainSyncTimeouts),
+      "    intersect = " ++ show (intersectTimeout scChainSyncTimeouts),
+      "    mustReply = " ++ show (mustReplyTimeout scChainSyncTimeouts)
+      ]
 
     mapM_ (traceWith tracer) $ BT.prettyPrint gtBlockTree
 
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -6,8 +6,9 @@
 {-# LANGUAGE NamedFieldPuns #-}
 
 module Test.Consensus.Genesis.Setup
-  ( module Test.Consensus.Genesis.Setup,
-    module Test.Consensus.Genesis.Setup.GenChains
+  ( module Test.Consensus.Genesis.Setup.GenChains,
+    runTest,
+    runTest'
   )
 where
 
@@ -29,11 +30,10 @@ runTest ::
   (IOLike m, MonadTime m, MonadTimer m) =>
   GenesisTest ->
   PointSchedule ->
-  (TestFragH -> Property) ->
+  (StateView -> Property) ->
   m Property
 runTest genesisTest@GenesisTest {gtBlockTree, gtHonestAsc} schedule makeProperty = do
     (tracer, getTrace) <- recordingTracerTVar
-    -- let tracer = debugTracer
 
     traceWith tracer $ "Honest active slot coefficient: " ++ show gtHonestAsc
 
@@ -42,14 +42,26 @@ runTest genesisTest@GenesisTest {gtBlockTree, gtHonestAsc} schedule makeProperty
     finalStateView <- runPointSchedule schedulerConfig genesisTest schedule tracer
     trace <- unlines <$> getTrace
 
-    pure
-      $ counterexample trace
-      $ case svChainSyncExceptions finalStateView of
-          [] ->
-            let fragment = svSelectedChain finalStateView
-            in counterexample ("result: " <> condense fragment) (makeProperty fragment)
-          exns ->
-            counterexample ("exceptions: " <> show exns) False
+    pure $ counterexample trace $ makeProperty finalStateView
 
     where
       schedulerConfig = SchedulerConfig {enableTimeouts = False}
+
+-- | Same as 'runTest' except it fails the test in case of exception.
+runTest' ::
+  (IOLike m, MonadTime m, MonadTimer m) =>
+  GenesisTest ->
+  PointSchedule ->
+  (StateView -> Property) ->
+  m Property
+runTest' genesisTest schedule makeProperty =
+  runTest
+    genesisTest
+    schedule
+    $ \stateView ->
+    case svChainSyncExceptions stateView of
+      [] ->
+        counterexample ("result: " <> condense (svSelectedChain stateView)) $
+          makeProperty stateView
+      exns ->
+        counterexample ("exceptions: " <> show exns) False
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -30,11 +30,12 @@ import Ouroboros.Network.Protocol.ChainSync.Codec (ChainSyncTimeout(..))
 
 runTest ::
   (IOLike m, MonadTime m, MonadTimer m) =>
+  SchedulerConfig ->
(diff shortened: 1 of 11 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
```diff
@@ -23,31 +24,61 @@ import           Test.QuickCheck
...
   PointSchedule ->
-  (TestFragH -> Property) ->
+  (StateView -> Property) ->
   m Property
-runTest genesisTest@GenesisTest {gtBlockTree, gtHonestAsc} schedule makeProperty = do
+runTest schedulerConfig genesisTest schedule makeProperty = do
     (tracer, getTrace) <- recordingTracerTVar
-    -- let tracer = debugTracer
 
-    traceWith tracer $ "Honest active slot coefficient: " ++ show gtHonestAsc
```
tek, line 37:
```haskell
traceLinesWith tracer = traverse_ (traceWith tracer)
```

?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -0,0 +1,85 @@
...
+          (btTrunk $ gtBlockTree genesisTest)
+
+  pure $ withMaxSuccess 10 $ runSimOrThrow $
+    runTest schedulerConfig genesisTest schedule $ \stateView ->
+      case svChainSyncExceptions stateView of
+        [] ->
+          counterexample ("result: " ++ condense (svSelectedChain stateView)) False
+        [exn] ->
+          case fromException $ cseException exn of
+            Just (ExceededTimeLimit _) -> counterexample "REVIEW" True
```
tek, line 54: what's the issue here? shouldn't it just be "timed out"?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -0,0 +1,85 @@
...
+          headerPoint = HeaderPoint $ getHeader tipBlock
+          blockPoint = BlockPoint tipBlock
+          state = Peer HonestPeer $ NodeOnline $ AdvertisedPoints tipPoint headerPoint blockPoint
+          tick = Tick { active = state, peers = Peers state Map.empty }
+          maximumNumberOfTicks = fromIntegral $ roundDiffTimeToSeconds $ timeout / tickDuration
+      in
+      PointSchedule (tick :| replicate maximumNumberOfTicks tick)
+
+    roundDiffTimeToSeconds :: DiffTime -> Integer
+    roundDiffTimeToSeconds = (`div` 1000000000000) . (+ 500000000000) . diffTimeToPicoseconds
```
tek, line 85:
```
ghci> 5.555 :: DiffTime
5.555s
ghci> round (5.555 :: DiffTime)
6
```

😅

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -0,0 +1,85 @@
...
+  where
+    -- FIXME: Should ideally not be hardcoded, especially because it is now
+    -- duplicated between here and within the PeerSimulator.
+    tickDuration :: DiffTime
+    tickDuration = 0.100 -- seconds
+
+    -- FIXME: Should ideally not be hardcoded, especially because it is now
+    -- duplicated between here and within the PeerSimulator.
+    slotLength :: SlotLength
+    slotLength = slotLengthFromSec 20
```
tek, line 68: in `Config`, it says:
```
REVIEW: Make it 1s or a parameter?
```

1s sounds pretty good, but it would still be unsafe I guess…maybe we should put it in `SchedulerConfig` and pass that to our `TopLevelConfig` constructor (which needs to be made more deliberate as well)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
Niols: Fixed in https://github.com/input-output-hk/ouroboros-consensus/pull/452/commits/c09d8d2feb87ccf01916ffe601812221ffae54e5.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
Niols: Oh yeah! Let's use just `round` then :p

tek (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -134,8 +133,7 @@ startChainSyncConnectionThread registry tracer cfg chainDbView fetchClientRegist
               traceUnitWith tracer ("ChainSyncClient " ++ condense srPeerId) "Terminating because of time limit exceeded."
             Nothing ->
               pure ()
-          throwIO exn
-        Right res' -> pure res'
+        Right _ -> pure ()
```
facundominguez, line 169: What difference does it make not rethrowing?

Niols: Rethrowing kills the scheduler and the tests. What we want (at least for TimeExceeded) is to register them and then process them in the QuickCheck property.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -0,0 +1,85 @@
...
+    -- contains just one too many ticks, therefore reaching the timeouts.
+    dullSchedule :: DiffTime -> TestFrag -> PointSchedule
+    dullSchedule _ (AF.Empty _) = error "requires a non-empty block tree"
+    dullSchedule timeout (_ AF.:> tipBlock) =
+      let tipPoint = TipPoint $ tipFromHeader tipBlock
+          headerPoint = HeaderPoint $ getHeader tipBlock
+          blockPoint = BlockPoint tipBlock
+          state = Peer HonestPeer $ NodeOnline $ AdvertisedPoints tipPoint headerPoint blockPoint
+          tick = Tick { active = state, peers = Peers state Map.empty }
+          maximumNumberOfTicks = fromIntegral $ roundDiffTimeToSeconds $ timeout / tickDuration
```
facundominguez: This expression is odd. `timeout / tickDuration` is no longer a DiffTime, but the number of ticks that fit in the timeout. So `roundDiffTimeToSeconds` is really rounding numbers of ticks to seconds ??

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -58,24 +62,16 @@ prop_timeouts = do
...
+    dullSchedule :: PointScheduleConfig -> DiffTime -> TestFrag -> PointSchedule
+    dullSchedule _ _ (AF.Empty _) = error "requires a non-empty block tree"
+    dullSchedule scheduleConfig timeout (_ AF.:> tipBlock) =
       let tipPoint = TipPoint $ tipFromHeader tipBlock
           headerPoint = HeaderPoint $ getHeader tipBlock
           blockPoint = BlockPoint tipBlock
           state = Peer HonestPeer $ NodeOnline $ AdvertisedPoints tipPoint headerPoint blockPoint
           tick = Tick { active = state, peers = Peers state Map.empty }
-          maximumNumberOfTicks = fromIntegral $ roundDiffTimeToSeconds $ timeout / pscTickDuration scSchedule
+          maximumNumberOfTicks = fromIntegral $ round $ timeout / pscTickDuration scheduleConfig
```
facundominguez, line 75: I think `fromIntegral` is redundant since `round` can return `Int`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
facundominguez: Later commits get rid of `roundDiffTimeToSeconds` which causes me confusion.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
Niols: It was removed later on.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
facundominguez: hm, I didn't expect exceptions to kill the tests if thrown inside forked threads.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
Niols: `DiffTime` has a [Fractional](https://hackage.haskell.org/package/base-4.16.3.0/docs/Prelude.html#t:Fractional) instance so `timeout / tickDuration` is still a `DiffTime`. Morally, a `DiffTime` should be an amount of time, but clearly `timeout / tickDuration` is a number of ticks (seconds divided by seconds per tick).

Niols: In general, I don't really like `DiffTime`; it can behave differently (as seconds or picoseconds) depending on the context.

Niols: It does behave as a number of seconds with respect to `round` and therefore `roundDiffTimeToSeconds` could simply be replaced by `round`.

Issue #448, opened 2023-10-23
Prototype of Lightweight Checkpointing


nfrisby:
This is for validating what we're writing in Issue #447.

dnadales:
We keep this open for now because there is a prototype for one approach, but we might want to prototype other approaches.

PR #453, opened 2023-10-24
Prototype lightweight checkpointing (non-combinator approach)

amesgen:
Prototype of lightweight checkpointing, see #448

To be merged, this would need more polish (Haddocks, addressing the inline TODOs) and tests.

Notably, this PR adds a new field to `ProtocolParams (CardanoBlock c)`:
```haskell
data instance ProtocolParams (CardanoBlock c) = ProtocolParamsCardano {
  ...
  , cardanoCheckpoints            :: Map BlockNo (HeaderHash (CardanoBlock c))
  }
```
with the semantics described in https://github.com/input-output-hk/ouroboros-consensus/issues/447#issuecomment-1775300165.

Note that this is *not* the approach described in https://github.com/input-output-hk/ouroboros-consensus/issues/447#issuecomment-1775306156; the approach of this PR is likely less invasive/has a smaller diff, but less elegant than the combinator-based appraoch.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20240111_195320_facundo.dominguez_lightweight_checkpointing.md
+++ b/ouroboros-consensus-cardano/changelog.d/20240111_195320_facundo.dominguez_lightweight_checkpointing.md
@@ -1,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+### Breaking
+
+- Implement lightweigth checkpointing [#449](https://github.com/IntersectMBO/ouroboros-consensus/issues/449).
+  A validation to help Genesis nodes follow the historical chain.
+
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
@@ -11,8 +11,9 @@ module Ouroboros.Consensus.Byron.Ledger.HeaderValidation (
   ) where
 
 import qualified Cardano.Chain.Slotting as CC
-import           Control.Monad (when)
+import           Control.Monad (unless, when)
 import           Control.Monad.Except (throwError)
+import qualified Data.Map.Strict as Map
 import           Data.Word
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
@@ -23,6 +24,7 @@ import           Ouroboros.Consensus.Byron.Ledger.Orphans ()
 import           Ouroboros.Consensus.Byron.Ledger.PBFT ()
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.HeaderValidation
+import           Ouroboros.Consensus.Util (whenJust)
 
 {-------------------------------------------------------------------------------
   Envelope
@@ -35,6 +37,7 @@ instance HasAnnTip ByronBlock where
 
 data ByronOtherHeaderEnvelopeError =
     UnexpectedEBBInSlot !SlotNo
+  | InvalidCheckpoint -- TODO args
   deriving (Eq, Show, Generic, NoThunks)
 
 instance BasicEnvelopeValidation ByronBlock where
@@ -56,9 +59,13 @@ instance BasicEnvelopeValidation ByronBlock where
 instance ValidateEnvelope ByronBlock where
   type OtherHeaderEnvelopeError ByronBlock = ByronOtherHeaderEnvelopeError
 
-  additionalEnvelopeChecks cfg _ledgerView hdr =
+  additionalEnvelopeChecks cfg _ledgerView hdr = do
       when (fromIsEBB newIsEBB && not (canBeEBB actualSlotNo)) $
         throwError $ UnexpectedEBBInSlot actualSlotNo
+      unless (fromIsEBB newIsEBB) $ -- TODO fine to ignore EBBs?
+        whenJust (Map.lookup (blockNo hdr) checkpoints) $ \checkpoint ->
+          unless (checkpoint == blockHash hdr) $
+            throwError InvalidCheckpoint
     where
       actualSlotNo :: SlotNo
       actualSlotNo = blockSlot hdr
@@ -75,3 +82,6 @@ instance ValidateEnvelope ByronBlock where
         . byronEpochSlots
         . configBlock
         $ cfg
+
+      checkpoints =
+        byronCheckpoints . configBlock $ cfg
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
@@ -37,7 +37,7 @@ instance HasAnnTip ByronBlock where
 
 data ByronOtherHeaderEnvelopeError =
     UnexpectedEBBInSlot !SlotNo
-  | InvalidCheckpoint -- TODO args
+  | CheckpointMismatch -- TODO args
   deriving (Eq, Show, Generic, NoThunks)
 
 instance BasicEnvelopeValidation ByronBlock where
@@ -65,7 +65,7 @@ instance ValidateEnvelope ByronBlock where
       when (not (fromIsEBB newIsEBB)) $ -- TODO fine to ignore EBBs?
         whenJust (Map.lookup (blockNo hdr) checkpoints) $ \checkpoint ->
           when (checkpoint /= blockHash hdr) $
-            throwError InvalidCheckpoint
+            throwError CheckpointMismatch
     where
       actualSlotNo :: SlotNo
       actualSlotNo = blockSlot hdr
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
@@ -11,7 +11,7 @@ module Ouroboros.Consensus.Byron.Ledger.HeaderValidation (
   ) where
 
 import qualified Cardano.Chain.Slotting as CC
-import           Control.Monad (unless, when)
+import           Control.Monad (when)
 import           Control.Monad.Except (throwError)
 import qualified Data.Map.Strict as Map
 import           Data.Word
@@ -62,9 +62,9 @@ instance ValidateEnvelope ByronBlock where
   additionalEnvelopeChecks cfg _ledgerView hdr = do
       when (fromIsEBB newIsEBB && not (canBeEBB actualSlotNo)) $
         throwError $ UnexpectedEBBInSlot actualSlotNo
-      unless (fromIsEBB newIsEBB) $ -- TODO fine to ignore EBBs?
+      when (not (fromIsEBB newIsEBB)) $ -- TODO fine to ignore EBBs?
         whenJust (Map.lookup (blockNo hdr) checkpoints) $ \checkpoint ->
-          unless (checkpoint == blockHash hdr) $
+          when (checkpoint /= blockHash hdr) $
             throwError InvalidCheckpoint
     where
       actualSlotNo :: SlotNo
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
@@ -331,6 +331,7 @@ protocolInfoTPraosShelleyBased ProtocolParamsShelleyBased {
           protVer
           genesis
           (shelleyBlockIssuerVKey <$> credentialss)
+          mempty -- TODO propagate?
 
     storageConfig :: StorageConfig (ShelleyBlock (TPraos c) era)
     storageConfig = ShelleyStorageConfig {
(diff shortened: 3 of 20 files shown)


ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
```diff
@@ -56,9 +59,13 @@ instance BasicEnvelopeValidation ByronBlock where
 instance ValidateEnvelope ByronBlock where
   type OtherHeaderEnvelopeError ByronBlock = ByronOtherHeaderEnvelopeError
 
-  additionalEnvelopeChecks cfg _ledgerView hdr =
+  additionalEnvelopeChecks cfg _ledgerView hdr = do
       when (fromIsEBB newIsEBB && not (canBeEBB actualSlotNo)) $
         throwError $ UnexpectedEBBInSlot actualSlotNo
+      unless (fromIsEBB newIsEBB) $ -- TODO fine to ignore EBBs?
+        whenJust (Map.lookup (blockNo hdr) checkpoints) $ \checkpoint ->
+          unless (checkpoint == blockHash hdr) $
```
nfrisby, line 67: Minor: I think the prevailing idiom is to prefer `when` over `unless` when the guard logic is easy to read either way.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
```diff
@@ -297,6 +297,7 @@ protocolInfoTPraosShelleyBased ProtocolParamsShelleyBased {
           protVer
           genesis
           (shelleyBlockIssuerVKey <$> credentialss)
+          mempty -- TODO propagate?
```
nfrisby, line 300: What's the hypothesis for why we wouldn't need to propagate?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -48,6 +49,7 @@ data PraosEnvelopeError
   = ObsoleteNode Version Version
   | HeaderSizeTooLarge Natural Natural
   | BlockSizeTooLarge Natural Natural
+  | InvalidCheckpoint -- TODO args
```
nfrisby, line 52: I suggest `CheckpointMismatch` or `CheckpointConflict` instead of "invalid checkpoint"

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -63,14 +65,17 @@ instance PraosCrypto c => ProtocolHeaderSupportsEnvelope (Praos c) where
...
+  envelopeChecks cfg checkpoints lv hdr = do
     unless (m <= maxpv) $ throwError (ObsoleteNode m maxpv)
     unless (fromIntegral (bhviewHSize bhv) <= maxHeaderSize) $
       throwError $
         HeaderSizeTooLarge (fromIntegral $ bhviewHSize bhv) maxHeaderSize
     unless (bhviewBSize bhv <= maxBodySize) $
       throwError $
         BlockSizeTooLarge (bhviewBSize bhv) maxBodySize
+    whenJust (Map.lookup (pHeaderBlock hdr) checkpoints) $ \checkpoint ->
+      unless (checkpoint == pHeaderHash hdr) $
```
nfrisby, line 77: same comment as other guard

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs
```diff
@@ -30,11 +39,18 @@ import           Ouroboros.Consensus.Shelley.Protocol.Abstract (ProtoCrypto,
...
                      ShelleyProtocol, ShelleyProtocolHeader, protocolHeaderView)
+import           Ouroboros.Consensus.Util (whenJust)
 
 type instance ProtoCrypto (TPraos c) = c
 
 type instance ShelleyProtocolHeader (TPraos c) = SL.BHeader c
 
+data TPraosEnvelopeError
+  = ChainPredicateFailure ChainPredicateFailure
+  | InvalidCheckpoint -- TODO args
```
nfrisby, line 50: same comment as other new constructor name

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs
```diff
@@ -44,13 +60,17 @@ instance PraosCrypto c => ProtocolHeaderSupportsEnvelope (TPraos c) where
...
+  type EnvelopeCheckError _ = TPraosEnvelopeError
+
+  envelopeChecks cfg checkpoints lv hdr = do
+    liftEither . first ChainPredicateFailure $
+      SL.chainChecks
+        maxPV
+        (SL.lvChainChecks lv)
+        (SL.makeHeaderView $ protocolHeaderView @(TPraos c) hdr)
+    whenJust (Map.lookup (pHeaderBlock hdr) checkpoints) $ \checkpoint ->
+      unless (checkpoint == pHeaderHash hdr) $
```
nfrisby, line 72: same comment as other guard

nfrisby (commented):
First impressions

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
amesgen: This function isn't ever called by `protocolInfoCardano` (which is the only place actually interesting to us), so I just did the lazy thing and didn't propagate it further for this prototype PR.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
amesgen: Yeah, I am always paralyzed by the choice between `when` and `unless`... I used `unless` here as it is also used in all preceeding checks (even though it would be easy to rewrite these places from eg `unless (m <= maxpv)` to `when (m > maxpv)`), and then also used it in all other `additionalEnvelopeChecks` for consistency.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs
amesgen: https://github.com/input-output-hk/ouroboros-consensus/pull/453#discussion_r1371394977

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
amesgen: https://github.com/input-output-hk/ouroboros-consensus/pull/453#discussion_r1371394977

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs
facundominguez: Switched unless to when in 5c93fa0

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/HeaderValidation.hs
facundominguez: Switched unless to when in 5c93fa0

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
facundominguez: Renamed to CheckpointMismatch in 8c48082

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs
facundominguez: Renamed to CheckpointMismatch in 8c48082

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
facundominguez: Switched unless to when in 5c93fa0

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
facundominguez: I would consider addressing this if necessary for testing.

ouroboros-consensus-cardano/changelog.d/20240111_195320_facundo.dominguez_lightweight_checkpointing.md
```diff
@@ -0,0 +1,23 @@
...
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
```
facundominguez, line 13: Should this be Breaking do to the changes to APIs in the code?

I initially chose based on whether this would break the behavior of honest nodes :)

amesgen: FTR: This should indeed be based on the changes to the code based on PVP :+1:

facundominguez: I changed this to the Breaking category.

facundominguez:
The plan now is to rearrange the PR so `Ouroboros.Consensus.HeaderValidation.validateEnvelope` invokes the validation of checkpoints. Let's call this function
```Haskell
validateIfCheckpoint :: TopLevelConfig blk -> Header blk -> Except (HeaderEnvelopeError blk) ()
```

This has the advantage of only needing to call the validation code at one place for ALL types of blocks.

This in turn requires getting the map of checkpoints at this point in the code. This can be done by adding the map as a method to the class `ValidateEnvelope`:

```Haskell
class ... => ValidateEnvelope blk where 
  ...
  -- | Checkpoints to validate. See 'validateIfCheckpoint'.
  checkpoints :: TopLevelConfig blk -> Map BlockNo (HeaderHash blk)
```

I wouldn't use a default method for this. It is better to signal to authors of instances that they need to define this method if they care about checkpointing, and they can easily return the empty map when they don't care.

Then the instances of `ValidateEnvelope` for `ShelleyBlock` and `ByronBlock` would need to return the checkpoints in their respective top-level configs.

As testing plan, we would only test `validateIfCheckpoint` for `TestBlock`. In particular, we wouldn't write any tests to test the instances for `ShelleyBlock` and `ByronBlock`.

nfrisby:
I suggest we take the "universality"-based approach even farther, and simply add a field `topLevelConfigCheckpoints :: Map BlockNo (HeaderHash blk)` to `TopLevelConfig`.

https://github.com/IntersectMBO/ouroboros-consensus/blob/982eff4ac03192d0685a9bc04431074d0943e7cb/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Config.hs#L38-L45

amesgen:
Just ensured that there are no roadblocks for this approach: https://github.com/IntersectMBO/ouroboros-consensus/pull/898 (diff could get much smaller with a compat layer/pattern synonym, but also not really necessary).

For completeness: A (we think relatively minor) disadvantage of #898 is that it adds checkpointing as a built-in concept to the abstract Consensus layer, despite this not being strictly necessary. (It is of course always possible to not do checkpointing by simply not providing any checkpoints.)

A small subtlety I noticed: In the HFC, we often create `TopLevelConfig`s for each era. In #898, these per-era `TopLevelConfig`s will contain no checkpoints, and this won't be caught by tests as `validateEnvelope` is only called at the HF layer.

nfrisby:
>  I noticed: In the HFC, we often create `TopLevelConfigs` for each era

I'm not seeing the "in the HFC" aspect of this in the PR 898 diff. I see `topLevelConfigCheckpoints = mempty` in plenty of `protocolInfo*` functions, for example, and some testing functions. But those are not specific to the HFC, right?

> ... `validateEnvelope` is only called at the HF layer.

Does that imply that it's harmless for the per-era configs to have an empty checkpoints field? Since our intended semantics is for the checkpoints field to _only_ affect `validateEnvelope`? Or am I oversimplifying something?

amesgen:
> > I noticed: In the HFC, we often create `TopLevelConfigs` for each era
> 
> I'm not seeing the "in the HFC" aspect of this in the PR 898 diff. I see `topLevelConfigCheckpoints = mempty` in plenty of `protocolInfo*` functions, for example, and some testing functions. But those are not specific to the HFC, right?

It happens here: https://github.com/IntersectMBO/ouroboros-consensus/blob/982eff4ac03192d0685a9bc04431074d0943e7cb/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs#L199-L211
`mkTopLevelConfig` is touched in #898 and currently always uses `mempty` (just was the most direct thing to get things to compile). 
 
> > ... `validateEnvelope` is only called at the HF layer.
> 
> Does that imply that it's harmless for the per-era configs to have an empty checkpoints field? Since our intended semantics is for the checkpoints field to _only_ affect `validateEnvelope`? Or am I oversimplifying something?

Yes, that is exactly right :+1: (Theoretically, some concrete block could of course depend on them in some non-trivial way, but I don't have a realistic use case in mind.)

amesgen:
Superseded by #898

PR #458, opened 2023-10-25
Add a test to check that rollbacks actually happen

Niols:
This is only meant as a sanity check for the implementation on rollback.

Should be squashed.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
@@ -37,9 +37,10 @@ import           Text.Printf (printf)
 --
 -- INVARIANT: @btbFull == fromJust $ AF.join btbPrefix btbSuffix@
 data BlockTreeBranch blk = BlockTreeBranch {
-    btbPrefix :: AF.AnchoredFragment blk,
-    btbSuffix :: AF.AnchoredFragment blk,
-    btbFull   :: AF.AnchoredFragment blk
+    btbPrefix      :: AF.AnchoredFragment blk,
+    btbSuffix      :: AF.AnchoredFragment blk,
+    btbTrunkSuffix :: AF.AnchoredFragment blk,
+    btbFull        :: AF.AnchoredFragment blk
   }
   deriving (Show)
 
@@ -79,7 +80,7 @@ mkTrunk btTrunk = BlockTree { btTrunk, btBranches = [] }
 -- block in common with an existingbranch.
 addBranch :: AF.HasHeader blk => AF.AnchoredFragment blk -> BlockTree blk -> Maybe (BlockTree blk)
 addBranch branch BlockTree{..} = do
-  (_, btbPrefix, _, btbSuffix) <- AF.intersect btTrunk branch
+  (_, btbPrefix, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
   -- NOTE: We could use the monadic bind for @Maybe@ here but we would rather
   -- catch bugs quicker.
   let btbFull = fromJust $ AF.join btbPrefix btbSuffix
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
@@ -80,7 +80,7 @@ mkTrunk btTrunk = BlockTree { btTrunk, btBranches = [] }
 -- block in common with an existing branch.
 addBranch :: AF.HasHeader blk => AF.AnchoredFragment blk -> BlockTree blk -> Maybe (BlockTree blk)
 addBranch branch BlockTree{..} = do
-  (_, btbPrefix, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
+  (btbPrefix, _, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
   -- NOTE: We could use the monadic bind for @Maybe@ here but we would rather
   -- catch bugs quicker.
   let btbFull = fromJust $ AF.join btbPrefix btbSuffix
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
@@ -77,7 +77,7 @@ mkTrunk btTrunk = BlockTree { btTrunk, btBranches = [] }
 -- the trunk.
 --
 -- FIXME: we should enforce that the new branch' suffix does not contain any
--- block in common with an existingbranch.
+-- block in common with an existing branch.
 addBranch :: AF.HasHeader blk => AF.AnchoredFragment blk -> BlockTree blk -> Maybe (BlockTree blk)
 addBranch branch BlockTree{..} = do
   (_, btbPrefix, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
@@ -31,11 +31,12 @@ import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Text.Printf (printf)
 
 -- | Represent a branch of a block tree by a prefix and a suffix. The full
--- fragment (the prefix and suffix catenated) is provided for practicality.
+-- fragment (the prefix and suffix catenated) and the trunk suffix (the rest of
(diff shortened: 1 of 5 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -76,7 +77,7 @@ mkTrunk btTrunk = BlockTree { btTrunk, btBranches = [] }
 -- block in common with an existingbranch.
 addBranch :: AF.HasHeader blk => AF.AnchoredFragment blk -> BlockTree blk -> Maybe (BlockTree blk)
 addBranch branch BlockTree{..} = do
-  (_, btbPrefix, _, btbSuffix) <- AF.intersect btTrunk branch
+  (_, btbPrefix, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
```
Niols, line 80: Actually, shouldn't it be
```suggestion
  (btbPrefix, _, btbTrunkSuffix, btbSuffix) <- AF.intersect btTrunk branch
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -37,9 +37,10 @@ import           Text.Printf (printf)
 --
 -- INVARIANT: @btbFull == fromJust $ AF.join btbPrefix btbSuffix@
```
Niols, line 38: This comment is out of date. It should mention `btbTrunkSuffix`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -0,0 +1,85 @@
...
+
+tests :: TestTree
+tests = testGroup "rollback" [
+  testProperty "can rollback" (prop_rollback True),
+  testProperty "cannot rollback" (prop_rollback False)
+  ]
+
+-- | The boolean parameter specifies whether we are interested in the property
+-- when we can rollback.
+prop_rollback :: Bool -> QC.Gen QC.Property
```
facundominguez, line 33: I think the documentation should explain with more precision what is the property being tested. I don't get it yet.

Niols: The goal for this is to be a sanity check for the PeerSimulator, not a proper consensus test (hence the location in `Consensus.PeerSimulator.Tests.*` and not `Consensus.Genesis.Tests.*` or something). It just aims at checking that the rollback is properly implemented by the mocked ChainSync server and properly followed by the client. Shall I add this to this property? Or maybe the whole module?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -0,0 +1,85 @@
...
+          branch = btbSuffix $ head $ btBranches blockTree
+          states = banalStates trunk ++ banalStates branch
+          peers = peersOnlyHonest states
+          pointSchedule = balanced peers
+       in fromJust pointSchedule
+
+    -- | Whether it is possible to roll back from the trunk after having served
+    -- it fully, that is whether there is an alternative chain that forks of the
+    -- trunk less than 'k' blocks from the tip and that is longer than the
+    -- trunk.
```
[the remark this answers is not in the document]
Niols: This is just as true for the part about `k`. The important property is indeed not about the tip but what last point you served: it has to be no more than `k` from the intersection and there has to be more blocks to serve on the alternative branch. For now, we only serve the tip, though it could be an option to serve less than this. Would you want me to make this clearer here?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
Niols: The chain generator is a bit of a degenerated case where both fragments share the same anchor. However, the interface of `BlockTree` does not make such assumptions and, in the event that it is reused in the future in a different way (by us or others), I think this would make it more robust, considering that the documented pre-condition for `addBranch` is only that the fragment has to be anchored in the trunk, not necessarily at the trunk's anchor.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -0,0 +1,85 @@
...
+    schedulerConfig = noTimeoutsSchedulerConfig defaultPointScheduleConfig
+
+    -- A schedule that advertises all the points of the trunk, then switches to
+    -- the first alternative chain of the given block tree.
+    --
+    -- PRECONDITION: Block tree with at least one alternative chain.
+    rollbackSchedule :: BlockTree TestBlock -> PointSchedule
+    rollbackSchedule blockTree =
+      let trunk = btTrunk blockTree
+          branch = btbSuffix $ head $ btBranches blockTree
```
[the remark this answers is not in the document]
Niols: This gives a warning “pattern not exhaustive” which the CI rejects. I'd rather not disable such a warning in the file so I will just revert this commit.

Niols:
Rebased on `genesis/point-scheduler`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -0,0 +1,85 @@
...
+          states = banalStates trunk ++ banalStates branch
+          peers = peersOnlyHonest states
+          pointSchedule = balanced peers
+       in fromJust pointSchedule
+
+    -- | Whether it is possible to roll back from the trunk after having served
```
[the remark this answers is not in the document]
Niols: I suppose this rejoins the general discussion on whether to discard inputs using pre-conditions or by changing the generators, although the solutions you suggest are extremely lightweight so I could easily be convinced!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
facundominguez: I'd like it to say something like
```
@prop_rollback True@ tests that the selection of the node under test changes branches when sent a rollback to a block no older than k blocks before the current selection.

@prop_rollback False@ tests that ...
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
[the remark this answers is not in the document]
facundominguez: Upon rereading the documentation, it looks sufficiently clear to me now. I still think that the name of the function could be made more precise.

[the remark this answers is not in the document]
Niols: Fine by me! I will take care of this in an hour or so.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
Niols: Sounds great! Will do that.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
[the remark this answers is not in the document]
Niols: Done in f9a9df975.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
Niols: WDYT about 291df5f1d?

facundominguez: That helps. Thanks!

facundominguez (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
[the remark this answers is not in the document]
tek: in any case, to be implemented later

PR #459, opened 2023-10-25
Discard the test cases _before_ running the peer simulator if possible?

Niols:
This PR is intended only to spark a discussion. Currently, we have:

```haskell
runTest .... $ \stateView ->
  testPremise ==> testConclusion
```

which means that we run the whole schedule before actually checking whether the premise holds. In my example, if `testPremise` depends on `stateView`, we can't really do anything different. However, if it doesn't (like in our current long range attack test case), we can also go for:

```haskell
testPremise
==>
runTest .... $ stateView -> testConclusion
```

which only runs the schedule when the premise holds. This is faster, and potentially a lot faster if a lot of tests are discarded (in the current state of the long range attack test case, we discard a bit less than 2/3 of the test cases). Is there a reason why not to do this? I already used this pattern in https://github.com/input-output-hk/ouroboros-consensus/pull/458 but can of course change that.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
@@ -38,12 +38,20 @@ prop_longRangeAttack = do
   (genesisTest, schedule) <- genChainsAndSchedule scheduleConfig 1 FastAdversary
   let Classifiers {..} = classifiers genesisTest
 
-  pure $ withMaxSuccess 10 $ runSimOrThrow $
-    runTest (noTimeoutsSchedulerConfig scheduleConfig) genesisTest schedule $ exceptionCounterexample $ \StateView{svSelectedChain} ->
-        classify genesisWindowAfterIntersection "Full genesis window after intersection"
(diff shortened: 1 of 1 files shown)


Niols:
It might be that the discarded tests would trigger invariant violations and therefore the second option would miss them.

Niols:
Idea: it could be good to have a “quick mode” of running tests that actually discards before and a “slow mode” of running tests that discards after. The former would typically be used locally while the latter would be run in CI. (Those modes could also change the number/size of tests.)

Niols:
We could also introduce our own operator (say `===>`) that behaves as an implication or as a noop depending on the mode. And then write something like:

```haskell
testPremise
===> 
runTest .... $ \stateView ->
  testPremise ==> testConclusion
```

(or we use the “approximate case” exception of QuickCheck or something).

facundominguez:
I'd rather have generators of which we can constrain the length of the chains. But we could deffer this conversations until profiling a bit the generator and the execution of the tests.

Niols:
Those two changes are not incompatible. Are you saying you would rather not make the change in this PR and change the generators instead?

facundominguez:
> Those two changes are not incompatible.

Adding parameters to control the chain lengths is going to eliminate the need to discard inputs in the example that we have. But it is fine to merge this in the meantime.

facundominguez (approved):

Niols:
Indeed!

PR #460, opened 2023-10-25
A few `HTrans` instances

jorisdral:
# Description

`HTrans` instances for `Mismatch` and `OptNP`, which we'll need later for #344

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
@@ -1,17 +1,21 @@
-{-# LANGUAGE ConstraintKinds      #-}
-{-# LANGUAGE DataKinds            #-}
-{-# LANGUAGE DerivingStrategies   #-}
-{-# LANGUAGE EmptyCase            #-}
-{-# LANGUAGE FlexibleContexts     #-}
-{-# LANGUAGE GADTs                #-}
-{-# LANGUAGE LambdaCase           #-}
-{-# LANGUAGE PolyKinds            #-}
-{-# LANGUAGE RankNTypes           #-}
-{-# LANGUAGE StandaloneDeriving   #-}
-{-# LANGUAGE TypeApplications     #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE TypeOperators        #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE DerivingStrategies    #-}
+{-# LANGUAGE EmptyCase             #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE InstanceSigs          #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PolyKinds             #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE StandaloneDeriving    #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
 -- | Intended for qualified import
 --
@@ -37,8 +41,11 @@ module Data.SOP.Match (
   ) where
 
 import           Data.Bifunctor
+import           Data.Coerce (coerce)
+import           Data.Constraint (Dict (..))
 import           Data.Functor.Product
 import           Data.Kind (Type)
+import           Data.SOP.Classes (Same)
 import           Data.SOP.Strict
 import           Data.SOP.Telescope (Telescope (..))
 import qualified Data.SOP.Telescope as Telescope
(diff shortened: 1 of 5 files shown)


ouroboros-consensus/src/ouroboros-consensus/Data/SOP/OptNP.hs
```diff
@@ -1,18 +1,24 @@
...
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE StandaloneDeriving    #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
jasagredo, line 19: Instead of using this, I prefer writing
```
{- HLINT ... -}
```
which will not be parsed as a pragma

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/OptNP.hs
```diff
@@ -1,18 +1,24 @@
...
+{-# LANGUAGE PolyKinds             #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE StandaloneDeriving    #-}
+{-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
jasagredo, line 18: Why do we need this?

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
```diff
@@ -109,6 +116,53 @@ instance HAp (Mismatch f) where
...
+       forall f g xs ys. AllZipN (Prod (Mismatch p)) (LiftedCoercible f g) xs ys
+    => Mismatch p f xs
+    -> Mismatch p g ys
+  hcoerce = \case
+      ML fx gy -> ML (coerce fx) $ hcoerce gy
+      MR fy gx | Dict <- tailDict -> MR (hcoerce fy) $ coerce gx
+        where
+          tailDict :: Dict (AllZip (LiftedCoercible p p) (Tail xs) (Tail ys))
+          tailDict = replaceAllZipConstraint (Proxy @(LiftedCoercible f g))
+      MS m     -> MS $ hcoerce m
```
amesgen, line 150: To avoid code duplication, we could also define
```haskell
defaultHcoerce ::
     forall h1 h2 f g xs ys.
     (HTrans h1 h2, AllZipN (Prod h1) (LiftedCoercible f g) xs ys)
  => h1 f xs
  -> h2 g ys
defaultHcoerce = htrans (Proxy @(LiftedCoercible f g)) coerce
```
and then write `hcoerce = defaultHcoerce` (also for `OptNP`).

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
```diff
@@ -109,6 +116,53 @@ instance HAp (Mismatch f) where
...
+  => Proxy c -> Dict (AllZip c' xs ys)
+replaceAllZipConstraint _ = go sList sList
+  where
+    go ::
+         forall as bs. AllZip c as bs
+      => SList as -> SList bs
+      -> Dict (AllZip c' as bs)
+    go SNil  SNil  = Dict
+    go SCons SCons = case go (sList @(Tail as)) (sList @(Tail bs)) of
+        Dict -> Dict
```
amesgen, line 164: As discussed in the past on Slack, this could be generalized to
```haskell
impliesAllZip ::
     forall c c' xs ys.
     (AllZip c xs ys, forall x y. c x y => c' x y)
  => Proxy c -> Dict (AllZip c' xs ys)
impliesAllZip _ = go sList sList
  where
    go ::
         forall as bs. AllZip c as bs
      => SList as -> SList bs
      -> Dict (AllZip c' as bs)
    go SNil  SNil  = Dict
    go SCons SCons = case go (sList @(Tail as)) (sList @(Tail bs)) of Dict -> Dict
```
(just to make sure that future users of this function are not confused :smile:)

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
jorisdral: Good point -- I forgot about this

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
jorisdral: Do you mean the name could be generalised? The code is the same as far as I can see

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/OptNP.hs
jorisdral: I forgot to remove this, because a previous version had `coerce_OptNP = unsafeCoerce`, in which case we still need some `AllZip (LiftedCoercible ...` constraints  to ensure that the unsafe coerce can not go wrong

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
jorisdral: Of course, we could also follow the `sop-core` convention and use `unsafeCoerce` instead. Do we want that?

amesgen: I have no intuition about how important/safe using `unsafeCoerce` here is.

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
amesgen: The quantified constraint changed from
```haskell
forall x y. c' x y
```
to
```haskell
forall x y. c x y => c' x y
```
where the latter is less general (easier to satisfy).

jorisdral: You're right, I overlooked that!

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/Match.hs
jorisdral: My intuition is that it `unsafeCoerce` is faster because it saves traversing the structure, but of course we should be reasonably sure that `unsafeCoerce` is "safe" within this context

amesgen: Let's write microbenchmarks for `hcoerce` :trollface:

ouroboros-consensus/src/ouroboros-consensus/Data/SOP/OptNP.hs
jorisdral: My editor complains. Does it not for you?

image

jasagredo: You can write it without the hashtags

amesgen (approved):

Issue #461, opened 2023-10-25
Investigate 8.5 regressions


dnadales:
# Regressions

1. Block context acquisition  (Issue of TVar contention?) regression needs to resolved - note the absolute value of the regression is the same for both the value and the plutus workloads. Regression present since 8.2.1. 
2. mempool snapshotting regression needs to be resolved - note that the absolute value is significantly different between the value and plutus workload. after 8.2.1

# Versions to compare

See manifests for details.

- For regression 1: 8.1.2 vs 8.2.1
- For regression 2: 8.2.1 vs 8.5.0

nfrisby:
Block context acquisition is the time between these two tracers https://github.com/input-output-hk/ouroboros-consensus/blob/e466cc1ce0844ffaf18cd1087c11ac5b5f36464c/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs#L279-L295

Remarks:

- The STM transaction could be replaced by `readTVarIO`.
- The `TVar` being read is written in exactly two places: when our selection changes (slightly more often than active slot coefficient `f`) and when the immutable tip advances (exactly as often as `f`).
- The `mkCurrentBlockContext` is not forced within the `atomically`.
- The vast majority of the time, the `mkCurrentBlockContext` logic determines the `Left/Right`ness after merely forcing the slot of the youngest header in the selection.
- None of that code changed between these two benchmarked versions.

doyougnu:
> Block context acquisition is the time between these two calls to trace.

This seems like a good candidate for inserting cost centers by hand then: https://downloads.haskell.org/ghc/latest/docs/users_guide/profiling.html#inserting-cost-centres-by-hand

and then with `-prof` you could observe a change between compiler versions.

doyougnu:
> A separate variance analysis demonstrated that that calculation varies by no more than 1ms.

This can only be true if the distribution of samples is found to be normal. If its not normal than a non-parametric test is required,  when I worked on the ledger regression the timing samples were highly, highly, non-normal. Which means that any report which relies on a `mean` as a summary statistic is going to be skewed by extreme outliers and may put the mean where no data points actually exist in the distribution.

mgmeier:
@doyougnu Thanks for pointing that out. Here is what we record with our statistics for a single run (this is the data for the 8.5 run above):
```
|---------------------------------+---------+--------+--------+--------+--------+--------|
[...]

PR #462, opened 2023-10-25
Add `getCurrentChainIO`, use it for `BlockContext` acquisition

amesgen:
Motivation: see if this has any impact on performance, cf #461

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE MultiWayIf          #-}
+{-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE RecordWildCards     #-}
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -60,6 +62,20 @@ import           Ouroboros.Network.Block (MaxSlotNo, maxSlotNoFromWithOrigin)
 -- In the latter case, we don't take blocks already in the ImmutableDB into
 -- account, as we know they /must/ have been \"immutable\" at some point, and,
 -- therefore, /must/ still be \"immutable\".
+getCurrentChain'
```
nfrisby, line 65: Minor: If you defined `getCurrentChain'` last, then would the diff be shorter/better aligned?

nfrisby (commented):
I agree that this diff does what we had discussed on the phone call today re Issue 461.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
amesgen: Right, that seems to be the case, will change :+1:

amesgen:
This is no longer relevant, see #513

PR #463, opened 2023-10-26
Lift `HTrans`-like functions to `HardForkBlock`-related datatypes

jorisdral:
# Description

Something that legacy blocks rely on is that we can "easily" map to and from `LegacyBlock`-related types. For example, the following types are coercible:

```haskell
newtype LegacyBlock (blk :: Type) = LegacyBlock { getLegacyBlock :: blk }
newtype instance BlockConfig (LegacyBlock blk) = LegacyBlockConfig (BlockConfig blk)
```

The `LegacyCardanoBlock` is however not just a `LegacyBlock` wrapper around a `CardanoBlock`, it also maps all individual blocks in the `HardForkBlock` to `LegacyBlock`s too.

```haskell
type LegacyCardanoEras c =  LegacyBlock ByronBlock
                         ': LegacyCardanoShelleyEras c
type LegacyCardanoShelleyEras c =
  '[ LegacyBlock (ShelleyBlock (TPraos c) (ShelleyEra c))
   , LegacyBlock (ShelleyBlock (TPraos c) (AllegraEra c))
   , LegacyBlock (ShelleyBlock (TPraos c) (MaryEra c))
   , LegacyBlock (ShelleyBlock (TPraos c) (AlonzoEra c))
   , LegacyBlock (ShelleyBlock (Praos c)  (BabbageEra c))
   , LegacyBlock (ShelleyBlock (Praos c)  (ConwayEra c))
   ]
type LegacyCardanoBlock c = LegacyBlock (HardForkBlock (LegacyCardanoEras c))
```

This means that, to map between, for example, `BlockConfig (LegacyCardanoBlock c)` and `BlockConfig (CardanoBlock c)`, we have to `htrans` or `hcoerce` SOP-like structures, since most `HardForkBlock`-related datatypes are SOP-like structures under the hood. For example:

```haskell
newtype instance BlockConfig (HardForkBlock xs) = HardForkBlockConfig {
      hardForkBlockConfigPerEra :: PerEraBlockConfig xs
    }
newtype PerEraBlockConfig xs = PerEraBlockConfig { getPerEraBlockConfig :: NP BlockConfig xs }
```

As such, we lift `htrans` and `hcoerce` functions from `sop-core` structures to `HardForkBlock`-related datatypes. Note that we can't write `HTrans` instances for these types, because the class expects structures to be parameterised by a functor, whereas something like `BlockConfig` has a fixed functor under the hood.

```haskell
htrans_BlockConfig ::
     AllZip c xs ys
  => proxy c
  -> (forall x y. c x y => BlockConfig x -> BlockConfig y)
  -> BlockConfig (HardForkBlock xs)
  -> BlockConfig (HardForkBlock ys)
htrans_BlockConfig p f = {- elided -}

hcoerce_BlockConfig ::
     AllZip (LiftedCoercible BlockConfig BlockConfig) xs ys
  => BlockConfig (HardForkBlock xs)
  -> BlockConfig (HardForkBlock ys)
hcoerce_BlockConfig = {- elided -}
```

The list of implemented functions is not exhaustive, though it should be enough to run a node with legacy blocks (see #344).

NOTE: `hcoerce_*` functions could be implemented in terms of `htrans_*` functions to reduce code duplication, but this requires some type fiddling to get the class constraints right. Furthermore, I'm not sure we are going to implement legacy blocks in this exact way if we want them in production, so I'd like to hold off on perfecting/optimising the code before we are sure we want to keep to the current approach.

### DIFF
--- a/ouroboros-consensus-cardano-legacy-block/src/Ouroboros/Consensus/HardFork/Trans.hs
+++ b/ouroboros-consensus-cardano-legacy-block/src/Ouroboros/Consensus/HardFork/Trans.hs
@@ -1,0 +1,866 @@
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE LambdaCase            #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Use camelCase" #-}
+
+module Ouroboros.Consensus.HardFork.Trans (
+    -- * Header
+    hcoerce_Header
+    -- * NestedCtxt_
+  , hcoerce_NestedCtxt_
+  , htrans_NestedCtxt_
+    -- * ProtocolClientInfo
+  , hcoerce_ProtocolClientInfo
+  , htrans_ProtocolClientInfo
+    -- * TopLevelConfig
+  , hcoerce_TopLevelConfig
+  , htrans_TopLevelConfig
+    -- * ConsensusConfig
+  , hcoerce_ConsensusConfig
+  , htrans_ConsensusConfig
+    -- * LedgerConfig
+  , hcoerce_LedgerConfig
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-cardano-legacy-block/src/Ouroboros/Consensus/HardFork/Trans.hs
```diff
@@ -0,0 +1,866 @@
...
+    }
+  where
+    ProtocolInfo {
+        pInfoConfig
+      , pInfoInitLedger
+      } = pinfo
+
+
+
+
```
jorisdral, line 866: Unnecessary newlines :no_good:

PR #466, opened 2023-10-26
Generative point schedule

Niols:
This PR introduces a mechanism for point schedules to be extended while running them. The commits are meant to be read one by one. This PR is not ready to be merged yet as the actual extension is untested.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
@@ -85,6 +85,13 @@ data SchedulerConfig =
 
     -- | Config shared with point schedule generators.
     , scSchedule        :: PointScheduleConfig
+
+    -- | A function that can add more ticks once the evaluation came to an end.
+    -- It returns 'Nothing' if it does not wish to extend the point schedule or
+    -- 'Just' a point schedule otherwise. It receive the whole state view which
+    -- contains the point schedule so far as well as other information on the
+    -- current simulation.
+    , scExtender        :: StateView -> Maybe PointSchedule
   }
 
 defaultSchedulerConfig :: PointScheduleConfig -> Asc -> SchedulerConfig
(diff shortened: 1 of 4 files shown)


Niols:
I now added a variant of the timeout test that uses the new feature. This is of course not the most interesting case but it's a start. Surprisingly, the dynamic variant is systematically faster than the static one, and I don't understand why. Maybe we made a mistake when computing the necessary size for the `dullSchedule` and we are actually sometimes running too many ticks?

Niols:
Naming question. How would we call this feature? (This question is useful even if we decide not to implement this feature in the end.)

- Generative point schedules,
- Dynamic point schedules,
- .....?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -286,13 +256,39 @@ runPointSchedule schedulerConfig GenesisTest {gtSecurityParam = k, gtBlockTree}
...
+          svChainSyncExceptions <- collectExceptions (Map.elems chainSyncRess)
+          svSelectedChain <- atomically $ ChainDB.getCurrentChain chainDb
+          let svPointSchedule = maybe pointSchedule (<> pointSchedule) pastPointSchedule
+              stateView = StateView {
+                svSelectedChain,
+                svChainSyncExceptions,
+                svPointSchedule
+                }
+          -- Extend and loop or return.
+          case scExtender schedulerConfig stateView of
```
tek, line 284: Do you have a use case in mind where the extender would continue in the presence of exceptions? otherwise we could do the exception check unconditionally here.

tek:
"extensible schedule" would match your existing terminology 😉

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
Niols: I don't think that'd be right. One peer can raise an exception (eg. because of a timeout) while the other peers still function properly.

Niols:
Indeed! Simple and efficient, I like it.

Niols:
However, my understanding was that we would maybe not go towards extensible point schedules after all, so maybe we should just trash the whole PR.

Niols:
We decided to not follow this route in the end. Closing.

PR #468, opened 2023-10-26
ReleaseProcess.md: recommend opening an early CHaP PR

amesgen:


jasagredo (approved):
Shall we mention that the script is not failproof and one should check it did what it was expected to do?

jasagredo (approved):

PR #473, opened 2023-10-27
Finalise Cardano Legacy Block

### LINKED ISSUE #344: Run a node with legacy blocks

jorisdral:
# Description

Resolves #344

### DIFF
--- a/ouroboros-consensus-cardano-legacy-block/src/Legacy/Byron/Node.hs
+++ b/ouroboros-consensus-cardano-legacy-block/src/Legacy/Byron/Node.hs
@@ -1,0 +1,59 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE InstanceSigs          #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeApplications      #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Legacy.Byron.Node () where
+
+import           Cardano.Chain.Byron.API (ApplyMempoolPayloadErr)
+import           Codec.CBOR.Decoding (Decoder)
+import           Codec.CBOR.Encoding (Encoding)
+import           Data.Coerce (coerce)
+import           Ouroboros.Consensus.Legacy.Block (CodecConfig (..),
+                     LegacyBlock)
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion
+                     (HasNetworkProtocolVersion (..))
+import           Ouroboros.Consensus.Node.Serialisation
+                     (SerialiseNodeToClient (..))
+import           Ouroboros.Consensus.Protocol.PBFT.State (PBftState)
+import           Ouroboros.Consensus.Storage.Serialisation (DecodeDisk (..),
+                     EncodeDisk (..))
+
+{-------------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


ouroboros-consensus-cardano-legacy-block/src/Legacy/Byron/Node.hs
jasagredo, line 1: I don't see this file as necessarily Byron only, am I wrong?

ouroboros-consensus-cardano-legacy-block/src/Legacy/Cardano/Node.hs
```diff
@@ -0,0 +1,345 @@
...
+  latestReleasedNodeVersion _ =
+      latestReleasedNodeVersion (Proxy @(CardanoBlock c))
+
+{-------------------------------------------------------------------------------
+  ProtocolInfo
+-------------------------------------------------------------------------------}
+
+-- | Only used in 'protocolInfoLegacyCardano'.
+class (ToLegacyBlock x y, CanStowLedgerTables (LedgerState x)) => C x y
+instance CanStowLedgerTables (LedgerState x) => C x (LegacyBlock x)
```
jasagredo, line 215: I don't see this being used.

ouroboros-consensus-cardano-legacy-block/src/Legacy/Shelley.hs
jasagredo, line 1: I'm dubious about how useful are these gathering modules. I think it should be fine because it is only instances. If they exported stuff I would not like it.

ouroboros-consensus-cardano-legacy-block/src/Legacy/Shelley/Node.hs
```diff
@@ -0,0 +1,76 @@
...
+  encodeNodeToClient ::
+       CodecConfig (LegacyBlock blk)
+    -> BlockNodeToClientVersion (LegacyBlock blk)
+    -> ApplyTxError era -> Encoding
+  encodeNodeToClient ccfg version = coerce $
+      encodeNodeToClient @blk @(ApplyTxError era) (coerce ccfg) version
+
+  decodeNodeToClient ::
+       CodecConfig (LegacyBlock blk)
+    -> BlockNodeToClientVersion (LegacyBlock  blk)
```
jasagredo, line 73:
```suggestion
    -> BlockNodeToClientVersion (LegacyBlock blk)
```

jasagredo (approved):
I left some comments, but nothing too serious. It looks good

ouroboros-consensus-cardano-legacy-block/src/Legacy/Cardano/Node.hs
jorisdral: It's used here: https://github.com/input-output-hk/ouroboros-consensus/pull/473/files/199d9e308113dc32970549ffe533b967f6e82fe2#diff-165cfbdc83522aa96e78a011a6891f030987b36a63cea4300f1c10ae1e3ed130R253

ouroboros-consensus-cardano-legacy-block/src/Legacy/Shelley.hs
jorisdral: It's similar to having a `Ouroboros.Consensus.Cardano` module in `ouroboros-consensus-cardano`. Would you not like it because of re-exports, or?

ouroboros-consensus-cardano-legacy-block/src/Legacy/Byron/Node.hs
jorisdral: Hmm, I suppose maybe it isn't. It's specific to Byron's protocol, not the byron block. However, this module follows the same structure as the `ouroboros-consensus-cardano:byron` package

PR #477, opened 2023-10-30
Change the minimum chain schema length

facundominguez:
This PR causes chain schemas to have more than k blocks after intersections.

Having more than k blocks after intersections makes the inputs interesting for the long range attack, and for testing that rollbacks of more than `k` can't happen in caught up nodes.

This guarantee comes from the generation algorithm, which ensures that every s window has exactly `k` blocks in the honest chain.

For alternative chains, we ensure that the chains have more than `s+k+d+1` slots, and we choose the first active slot before this many slots from the end. Then the acceleration bound is before the last `k` blocks of the alternative chain, which we always set to active.

Race windows do not push the acceleration bound to the last `k` blocks because the intersections with the honest chain are always behind the last block preceding the last `s+k+d+1` slots.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -15,6 +15,7 @@ import qualified Control.Monad.Except as Exn
 import           Data.List (foldl')
 import           Data.Proxy (Proxy (Proxy))
 import qualified Data.Vector.Unboxed as Vector
+import           Data.Word (Word8)
 import           Ouroboros.Consensus.Block.Abstract hiding (Header)
 import           Ouroboros.Consensus.Protocol.Abstract
                      (SecurityParam (SecurityParam))
@@ -42,8 +43,8 @@ import           Test.Util.TestBlock hiding (blockTree)
 -- | Random generator for an honest chain recipe and schema.
 genHonestChainSchema :: QC.Gen (Asc, H.HonestRecipe, H.SomeHonestChainSchema)
 genHonestChainSchema = do
-  honestRecipe <- H.genHonestRecipe
   asc <- genAsc
+  honestRecipe <- H.genHonestRecipe
 
   H.SomeCheckedHonestRecipe Proxy Proxy honestRecipe' <-
     case Exn.runExcept $ H.checkHonestRecipe honestRecipe of
@@ -73,7 +74,7 @@ genAlternativeChainSchema (testRecipeH, arHonest) =
       A.arHonest
     }
 
-    alternativeAsc <- genAscWithKcp kcp scg
+    alternativeAsc <- ascFromBits <$> QC.choose (1 :: Word8, maxBound - 1)
 
     case Exn.runExcept $ A.checkAdversarialRecipe testRecipeA of
       Left e -> case e of
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -15,7 +15,6 @@ import qualified Control.Monad.Except as Exn
 import           Data.List (foldl')
 import           Data.Proxy (Proxy (Proxy))
 import qualified Data.Vector.Unboxed as Vector
-import           Data.Word (Word8)
 import           Ouroboros.Consensus.Block.Abstract hiding (Header)
 import           Ouroboros.Consensus.Protocol.Abstract
                      (SecurityParam (SecurityParam))
@@ -43,8 +42,8 @@ import           Test.Util.TestBlock hiding (blockTree)
 -- | Random generator for an honest chain recipe and schema.
 genHonestChainSchema :: QC.Gen (Asc, H.HonestRecipe, H.SomeHonestChainSchema)
 genHonestChainSchema = do
-  asc <- genAsc
   honestRecipe <- H.genHonestRecipe
+  asc <- genAsc
 
   H.SomeCheckedHonestRecipe Proxy Proxy honestRecipe' <-
     case Exn.runExcept $ H.checkHonestRecipe honestRecipe of
@@ -74,7 +73,7 @@ genAlternativeChainSchema (testRecipeH, arHonest) =
       A.arHonest
     }
 
-    alternativeAsc <- ascFromBits <$> QC.choose (1 :: Word8, maxBound - 1)
+    alternativeAsc <- genAscWithKcp kcp scg
 
     case Exn.runExcept $ A.checkAdversarialRecipe testRecipeA of
       Left e -> case e of
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -66,7 +66,7 @@ genAlternativeChainSchema (testRecipeH, arHonest) =
     let H.HonestRecipe kcp scg delta _len = testRecipeH
 
     (seedPrefix :: QCGen) <- QC.arbitrary
-    let arPrefix = genPrefixBlockCount seedPrefix arHonest
+    let arPrefix = genPrefixBlockCount seedPrefix testRecipeH arHonest
 
     let testRecipeA = A.AdversarialRecipe {
       A.arPrefix,
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
@@ -81,19 +81,20 @@ prop_cannotRollback = do
   where
     schedulerConfig = noTimeoutsSchedulerConfig defaultPointScheduleConfig
 
--- A schedule that advertises all the points of the trunk up until the kth
+-- | A schedule that advertises all the points of the trunk up until the nth
 -- block after the intersection, then switches to the first alternative
 -- chain of the given block tree.
 --
 -- PRECONDITION: Block tree with at least one alternative chain.
 rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
-rollbackSchedule k blockTree =
+rollbackSchedule n blockTree =
   let branch = head $ btBranches blockTree
-      trunk = btTrunk blockTree
-      prefixLen = AF.length $ btbPrefix branch
-      trunkPrefix = AF.takeOldest (k + prefixLen) trunk
-      branchSuffix = btbSuffix branch
-      states = banalStates trunkPrefix ++ banalStates branchSuffix
+      trunkSuffix = AF.takeOldest n (btbTrunkSuffix branch)
+      states = concat
+        [ banalStates (btbPrefix branch)
+        , banalStates trunkSuffix
+        , banalStates (btbSuffix branch)
+        ]
       peers = peersOnlyHonest states
       pointSchedule = balanced peers
    in fromJust pointSchedule
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
@@ -39,12 +39,17 @@ prop_rollback :: Bool -> QC.Gen QC.Property
 prop_rollback wantRollback = do
   genesisTest <- genChains 1
 
-  let schedule = rollbackSchedule (gtBlockTree genesisTest)
+  let SecurityParam k = gtSecurityParam genesisTest
+      schedule =
+        if wantRollback then rollbackSchedule (fromIntegral k) (gtBlockTree genesisTest)
+        else rollbackSchedule (fromIntegral (k + 1)) (gtBlockTree genesisTest)
 
   -- | We consider the test case interesting if we want a rollback and we can
   -- actually get one, or if we want no rollback and we cannot actually get one.
   pure $
-    wantRollback == canRollbackFromTrunkTip (gtSecurityParam genesisTest) (gtBlockTree genesisTest)
+    alternativeChainIsLongEnough (gtSecurityParam genesisTest) (gtBlockTree genesisTest)
+      &&
+    (wantRollback || honestChainIsLongEnough (gtSecurityParam genesisTest) (gtBlockTree genesisTest))
     ==>
       runSimOrThrow $ runTest schedulerConfig genesisTest schedule $ \StateView{svSelectedChain} ->
         let headOnAlternativeChain = case AF.headHash svSelectedChain of
@@ -58,33 +63,43 @@ prop_rollback wantRollback = do
   where
     schedulerConfig = noTimeoutsSchedulerConfig defaultPointScheduleConfig
 
-    -- A schedule that advertises all the points of the trunk, then switches to
-    -- the first alternative chain of the given block tree.
+    -- A schedule that advertises all the points of the trunk up until the kth
+    -- block after the intersection, then switches to the first alternative
+    -- chain of the given block tree.
     --
     -- PRECONDITION: Block tree with at least one alternative chain.
-    rollbackSchedule :: BlockTree TestBlock -> PointSchedule
-    rollbackSchedule blockTree =
-      let trunk = btTrunk blockTree
-          branch = btbSuffix $ head $ btBranches blockTree
-          states = banalStates trunk ++ banalStates branch
+    rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
+    rollbackSchedule k blockTree =
+      let branch = head $ btBranches blockTree
+          trunk = btTrunk blockTree
+          prefixLen = AF.length $ btbPrefix branch
+          trunkPrefix = AF.takeOldest (k + prefixLen) trunk
+          branchSuffix = btbSuffix branch
+          states = banalStates trunkPrefix ++ banalStates branchSuffix
           peers = peersOnlyHonest states
           pointSchedule = balanced peers
        in fromJust pointSchedule
 
-    -- | Whether it is possible to roll back from the trunk after having served
-    -- it fully, that is whether there is an alternative chain that forks of the
-    -- trunk no more than 'k' blocks from the tip and that is longer than the
-    -- trunk.
+    -- | Whether the honest chain has more than 'k' blocks after the
+    -- intersection with the alternative chain.
     --
     -- PRECONDITION: Block tree with exactly one alternative chain, otherwise
     -- this property does not make sense. With no alternative chain, this will
     -- even crash.
-    --
-    -- REVIEW: Why does 'existsSelectableAdversary' get to be a classifier and
-    -- not this? (or a generalised version of this)
-    canRollbackFromTrunkTip :: SecurityParam -> BlockTree TestBlock -> Bool
-    canRollbackFromTrunkTip (SecurityParam k) blockTree =
-      let BlockTreeBranch{btbSuffix, btbTrunkSuffix} = head $ btBranches blockTree
-          lengthSuffix = AF.length btbSuffix
+    honestChainIsLongEnough :: SecurityParam -> BlockTree TestBlock -> Bool
+    honestChainIsLongEnough (SecurityParam k) blockTree =
+      let BlockTreeBranch{btbTrunkSuffix} = head $ btBranches blockTree
           lengthTrunkSuffix = AF.length btbTrunkSuffix
-       in lengthTrunkSuffix <= fromIntegral k && lengthSuffix > lengthTrunkSuffix
+       in lengthTrunkSuffix > fromIntegral k
+
+    -- | Whether the alternative chain has more than 'k' blocks after the
+    -- intersection with the honest chain.
+    --
+    -- PRECONDITION: Block tree with exactly one alternative chain, otherwise
+    -- this property does not make sense. With no alternative chain, this will
+    -- even crash.
+    alternativeChainIsLongEnough :: SecurityParam -> BlockTree TestBlock -> Bool
+    alternativeChainIsLongEnough (SecurityParam k) blockTree =
+      let BlockTreeBranch{btbSuffix} = head $ btBranches blockTree
+          lengthSuffix = AF.length btbSuffix
+       in lengthSuffix > fromIntegral k
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
@@ -63,43 +63,43 @@ prop_rollback wantRollback = do
   where
     schedulerConfig = noTimeoutsSchedulerConfig defaultPointScheduleConfig
 
-    -- A schedule that advertises all the points of the trunk up until the kth
-    -- block after the intersection, then switches to the first alternative
-    -- chain of the given block tree.
-    --
-    -- PRECONDITION: Block tree with at least one alternative chain.
-    rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
-    rollbackSchedule k blockTree =
-      let branch = head $ btBranches blockTree
-          trunk = btTrunk blockTree
-          prefixLen = AF.length $ btbPrefix branch
-          trunkPrefix = AF.takeOldest (k + prefixLen) trunk
-          branchSuffix = btbSuffix branch
-          states = banalStates trunkPrefix ++ banalStates branchSuffix
-          peers = peersOnlyHonest states
-          pointSchedule = balanced peers
-       in fromJust pointSchedule
+-- A schedule that advertises all the points of the trunk up until the kth
+-- block after the intersection, then switches to the first alternative
+-- chain of the given block tree.
+--
+-- PRECONDITION: Block tree with at least one alternative chain.
+rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
+rollbackSchedule k blockTree =
+  let branch = head $ btBranches blockTree
+      trunk = btTrunk blockTree
+      prefixLen = AF.length $ btbPrefix branch
+      trunkPrefix = AF.takeOldest (k + prefixLen) trunk
+      branchSuffix = btbSuffix branch
+      states = banalStates trunkPrefix ++ banalStates branchSuffix
+      peers = peersOnlyHonest states
+      pointSchedule = balanced peers
+   in fromJust pointSchedule
 
-    -- | Whether the honest chain has more than 'k' blocks after the
-    -- intersection with the alternative chain.
-    --
-    -- PRECONDITION: Block tree with exactly one alternative chain, otherwise
-    -- this property does not make sense. With no alternative chain, this will
-    -- even crash.
-    honestChainIsLongEnough :: SecurityParam -> BlockTree TestBlock -> Bool
-    honestChainIsLongEnough (SecurityParam k) blockTree =
-      let BlockTreeBranch{btbTrunkSuffix} = head $ btBranches blockTree
-          lengthTrunkSuffix = AF.length btbTrunkSuffix
-       in lengthTrunkSuffix > fromIntegral k
+-- | Whether the honest chain has more than 'k' blocks after the
+-- intersection with the alternative chain.
+--
+-- PRECONDITION: Block tree with exactly one alternative chain, otherwise
+-- this property does not make sense. With no alternative chain, this will
+-- even crash.
+honestChainIsLongEnough :: SecurityParam -> BlockTree TestBlock -> Bool
+honestChainIsLongEnough (SecurityParam k) blockTree =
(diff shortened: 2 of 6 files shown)


facundominguez:
At the moment, the rollback tests are failing because they want to do rollbacks from the tip of the honest chain, which has more than k blocks after intersections.

tek:
my understanding is that this ensures that chains are at least two genesis windows long and that all forks happen before the k+1st newest block – how does it cause alternative chains to have maximum active slots?

facundominguez:
> how does it cause alternative chains to have maximum active slots?

This is done already by the implementation by removing active slots from the alternative chain in the race windows. The description of the PR is probably not the most accurate.

tek:
ah, that's much clearer. so it's unlikely we'll manage to get minimum length for fork fragments?

facundominguez:
> so it's unlikely we'll manage to get minimum length for fork fragments?

The number of active slots on the alternative chain schemas depend on Delta and when they fork from the honest schema. We could ensure that they have k+1 active slots if we do more computation, but maybe the complexity isn't worth the trouble. We can see how much we discard with these changes and reconsider.

tek:
```haskell
C.SomeWindow Proxy winA <- pure $ C.withSuffixWindow (C.windowSize winH) (C.Lbl @AdvLbl) firstAdvSlot
```

do I understand correctly that this guarantees that the slot count in the fork fragment matches the honest chain?
And if we have the maximum number of active slots in that fragment, does it mean that we have at least k, since the honest chain now has at least k+1?

facundominguez:
> do I understand correctly that this guarantees that the slot count in the fork fragment matches the honest chain?

The alternative schema starts at `firstAdvSlot` and ends in the same slot as the honest chain schema.

> And if we have the maximum number of active slots in that fragment, does it mean that we have at least k, since the honest chain now has at least k+1?

It depends on the value of Delta. If Delta is large, it may be the case that the generation algorithm assumes the adversary can fill every slot after the stability widow, and therefore, less than k slots might be active in the alternative chain.

facundominguez:
With the latest changes, I think it is guaranteed that there are k+1 blocks in both the honest and the alternative schemas after the intersection. Thanks @amesgen for the final insight.

This PR needs a PBT to check that the generated chains do actually have k+1 blocks.

tek:
looks reasonable!

facundominguez:
At the moment `prop_adversarialChain` fails with the following output
```
...
testRecipeH = HonestRecipe (Kcp 2) (Scg 4) (Delta 0) (Len 9)
---
            [  )rHon
      011001110                                                                
      [        )H
      [        )A
      000000111
             [ )rAdv'
      [        )stabWin'
...
Use --quickcheck-replay=732461 to reproduce.
```
The problem is that the first active slot of the alternative schema is too close to the end of time. The verification of the alternative schema expects the acceleration of the leader schedule to happen only after `s` slots after this first active slot, and it complains when finding the acceleration earlier.

There are two solutions I can think of to still guarantee `k+1` slots in every alternative schema:
1. Require the first slot of the alternative schema to be active. Thus there are sufficient slots where acceleration of the leader schedule is allowed.
2. Have the verification of the alternative branch ignore the last `k+1` slots of the alternative branch (they are always more than sgen+delta slots after the intersection).

Let me know if you have other suggestions.

@nfrisby, you might have opinions on how to resolve this.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -592,8 +602,13 @@ genPrefixBlockCount g (HonestRecipe (Kcp k) (Scg s) (Delta d) _len) schedH
   where
     ChainSchema _slots v = schedH
 
-    -- 'H.uniformTheHonestChain' ensures 0 < pc
-    numChoices = BV.countActivesInV S.notInverted $
+    -- 'H.uniformTheHonestChain' ensures there is at least one block in the
+    -- first s slots.
+    -- By leaving on block after the intersection in the first s slots, we
+    -- ensure there is are k+1 active slots in the honest chain after the
```
tek, line 608:
```suggestion
    -- ensure there are k+1 active slots in the honest chain after the
```

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -592,8 +602,13 @@ genPrefixBlockCount g (HonestRecipe (Kcp k) (Scg s) (Delta d) _len) schedH
   where
     ChainSchema _slots v = schedH
 
-    -- 'H.uniformTheHonestChain' ensures 0 < pc
-    numChoices = BV.countActivesInV S.notInverted $
+    -- 'H.uniformTheHonestChain' ensures there is at least one block in the
+    -- first s slots.
+    -- By leaving on block after the intersection in the first s slots, we
```
tek, line 607:
```suggestion
    -- By leaving one block after the intersection in the first s slots, we
```

facundominguez:
@amesgen proposes that instead of setting the $k+1$ youngest slots in the alternative schemas, we instead ensure higher active slot coefficients are used. This doesn't ensure that there will be $k+1$ active slots, but should still decrease significantly the discarded inputs.

It looks simple enough that I'd like to change the PR to adopt it after giving people some time to react.

tek:
sounds good!

facundominguez:
Alright, I think this one is good to merge.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -24,67 +24,100 @@ import           Test.Util.TestBlock (TestBlock, unTestHash)
...
+--
+-- PRECONDITION: Block tree with at least one alternative chain.
+rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
+rollbackSchedule k blockTree =
+  let branch = head $ btBranches blockTree
+      trunk = btTrunk blockTree
+      prefixLen = AF.length $ btbPrefix branch
+      trunkPrefix = AF.takeOldest (k + prefixLen) trunk
+      branchSuffix = btbSuffix branch
+      states = banalStates trunkPrefix ++ banalStates branchSuffix
```
Niols, line 96: I think I would prefer to see it written in three parts: the prefix, then `k` blocks on the trunk, then the other branch. Something like (untested):
```
let prefix = btbPrefix branch
    trunkSuffix = AF.takeOlders k $ btbTrunkSuffix branch
    branchSuffix = btbSuffix branch
    states = banalStates prefix ++ banalStates trunkSuffix ++ banalStates branchSuffix
```
I think I understand better what's supposed to happen. What do you think?

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
```diff
@@ -42,6 +42,8 @@ import qualified Test.Tasty.QuickCheck as TT
 
 tests :: [TT.TestTree]
 tests = [
+    TT.testProperty "k+1 blocks after the intersection" prop_kPlus1BlocksAfterIntersection
```
Niols, line 45: Nice, thank you for adding this!

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
```diff
@@ -42,6 +42,8 @@ import qualified Test.Tasty.QuickCheck as TT
...
+  ,
     TT.testProperty "prop_adversarialChain" prop_adversarialChain
   ,
     TT.localOption (TT.QuickCheckMaxSize 14) $ TT.testProperty "prop_adversarialChainMutation" prop_adversarialChainMutation
```
Niols, line 49: While we're at it, do you happen to know what `prop_adversarialChain` and `prop_adversarialChainMutation` are supposed to test? Could we find a more description test name here? (Only if you can do that from the top of your head!)

Niols (approved):
LGTM. A few nitpicking comments. Frankly, I understand at best half of what's happening in the chain generator. If you want a deeper review, I can spend the time to try and understand in more details.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
```diff
@@ -43,8 +42,8 @@ import           Test.Util.TestBlock hiding (blockTree)
 -- | Random generator for an honest chain recipe and schema.
 genHonestChainSchema :: QC.Gen (Asc, H.HonestRecipe, H.SomeHonestChainSchema)
 genHonestChainSchema = do
-  asc <- genAsc
   honestRecipe <- H.genHonestRecipe
+  asc <- genAsc
```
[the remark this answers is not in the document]
facundominguez: This was accidental. It is a left over from a failed solution.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -24,67 +24,100 @@ import           Test.Util.TestBlock (TestBlock, unTestHash)
...
   ]
 
--- | @prop_rollback True@ tests that the selection of the node under test
+-- | @prop_rollback@ tests that the selection of the node under test
 -- changes branches when sent a rollback to a block no older than 'k' blocks
 -- before the current selection.
---
--- @prop_rollback False@ tests that the selection of the node under test *does
+prop_rollback :: QC.Gen QC.Property
+prop_rollback = do
```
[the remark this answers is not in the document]
facundominguez: Yep, there is duplication. I think it still reads better than dealing with the boolean wantRollback, but could undo this if controversial.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -24,67 +24,100 @@ import           Test.Util.TestBlock (TestBlock, unTestHash)
...
   genesisTest <- genChains 1
 
-  let schedule = rollbackSchedule (gtBlockTree genesisTest)
+  let SecurityParam k = gtSecurityParam genesisTest
+      schedule = rollbackSchedule (fromIntegral (k + 1)) (gtBlockTree genesisTest)
 
-  -- | We consider the test case interesting if we want a rollback and we can
-  -- actually get one, or if we want no rollback and we cannot actually get one.
+  -- We consider the test case interesting if it allows to rollback even if
+  -- the implementation doesn't
```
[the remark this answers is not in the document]
facundominguez: We expect the implementation to not rollback, but in order to test that it refrains, the input needs to give it the chance to make a rollback.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -24,67 +24,100 @@ import           Test.Util.TestBlock (TestBlock, unTestHash)
...
-          peers = peersOnlyHonest states
-          pointSchedule = balanced peers
-       in fromJust pointSchedule
+-- A schedule that advertises all the points of the trunk up until the kth
+-- block after the intersection, then switches to the first alternative
+-- chain of the given block tree.
+--
+-- PRECONDITION: Block tree with at least one alternative chain.
+rollbackSchedule :: Int -> BlockTree TestBlock -> PointSchedule
+rollbackSchedule k blockTree =
```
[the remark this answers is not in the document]
facundominguez: Happy to rename to `n`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
facundominguez: Fine by me. I'll rewrite.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
facundominguez: `prop_adversarialChain` checks that the density is lower than in the honest chain, and also checks Length of competing chains.

`prop_adversarialChainMutation` checks that when decreasing the security parameter or when increasing the message delay, an alternative chain generated with the original values might fail the checks in `prop_adversarialChain`.

I'll try to rename.

facundominguez (commented):
Thanks for the review. Please, let me know if there are other changes that are not sufficiently explained.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
facundominguez: Renamed in 617fff17

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
facundominguez: Addressed in 87cc585ba

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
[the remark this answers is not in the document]
Niols: No, that's a good point. Let's keep it that way.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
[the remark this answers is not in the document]
Niols: I prefer this, thank you! :-) -- funny how `k` really stopped being just a random letter and really means “security parameter” in my head.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
```diff
@@ -93,12 +94,20 @@ ascFromBits w = ascFromDouble $ toEnum (fromEnum w) / (2 ^ B.finiteBitSize w)
 ascVal :: Asc -> Double
 ascVal (Asc x) = x
 
+-- | Provides an Asc with high probability to produce k slots in
+-- a window of size s. Therefore, we chose @Asc >= 3k/s@.
```
amesgen, line 98: Did you choose this as we have `s = 3k/f` on mainnet? I think it is fine to keep it like this for now (especially as the tests show that this usually results in $k$ slots active out of $s$ with the currently generated values for $k$ and $s$).

In a a bit more principled fashion, I guess one could set a target false positive rate $p$ (eg $p = 1/1000$) and then choose $\textrm{Asc}$ such that $F(k-1,s,\mathrm{Asc}) \le p$ (where $F$ is the cumulative binomial distribution function, eg from `statistics`). Doesn't seem too complicated, but also not strictly necessary for now.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
```diff
@@ -144,6 +146,42 @@ instance QC.Arbitrary SomeTestAdversarial where
                     testSeedH
                   }
 
+-- | Both the honest and the alternative schema have k+1 blocks after the
+-- intersection.
```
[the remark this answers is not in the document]
facundominguez: > IIUC, this is rather testing that if the alternative schema has at least k + 1 blocks after the intersection, then the honest schema does so too.

:+1:

> Also, we should "keep an eye" that not too many test cases are skipped here, right?

I think so. Can we automate that?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
facundominguez: > Did you choose this as we have s = 3k/f on mainnet?

That's indeed the case. I'll add a note about the choice.

I'd say the choice of mechanism to compute $Asc$ depends on whether we do want to test with less dense chains, as $3k/s$ might be produce sparse chains with too little probability.

Maybe we should aim for two generators, or a generator parameterized by $Asc$ or $p$.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -637,15 +637,29 @@ withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
 --
 -- The count will be strictly smaller than the number of active slots in the given 'ChainSchema'.
 --
+-- The following precondition ensures that there are enough slots to produce an
+-- alternative chain schema with at least k+1 slots.
+--
+-- PRECONDITION: @schemaSize schedH >= s + k + d + 1@
```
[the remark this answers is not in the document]
facundominguez: `genHonestRecipe` choses a length that is at least `2*s + d + 1`.

The assertion is a convoluted way to say that the intersection block must be chosen in the first `s-k` slots to make more probable that there will be $k + 1$ slots afterwards.

It might be too much of an assumption depending on how $Asc$ has been chosen. I didn't try to justify `k` over a smaller or higher value to accomplish that. It could be another value to adjust based on a chosen rate for false positives.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
amesgen: > Maybe we should aim for two generators, or a generator parameterized by $Asc$ or $p$.

Yeah, this feels similar to the second point in https://github.com/input-output-hk/ouroboros-consensus/pull/477#discussion_r1400438813.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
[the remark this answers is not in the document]
amesgen: Right, so $s + k + d + 1$ you chose because it will likely contain $k + 1$ active slots? That makes sense; any sequence of less than $2s-k + 1$ (maybe off by one) is technically not *guaranteed* to contain at least $k+1$ active slots just due to chain growth; but for the use case here, using an ad-hoc value seems fine, but I think it should be documented.

ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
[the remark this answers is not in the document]
amesgen: > > Also, we should "keep an eye" that not too many test cases are skipped here, right?
> 
> I think so. Can we automate that?

One way would be to use `cover` + `checkCoverage`:

```diff
diff --git a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
index bcf207898..098016e04 100644
--- a/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
+++ b/ouroboros-consensus/test/infra-test/Test/Ouroboros/Consensus/ChainGenerator/Tests/Adversarial.hs
@@ -168,9 +168,13 @@ prop_kPlus1BlocksAfterIntersection someTestAdversarial testSeedA = runIdentity $
     C.SomeWindow Proxy stabWin <- do
         pure $ calculateStability scg schedA
 
+    let altSchemaHasKP1AfterIntersection =
+          BV.countActivesInV S.notInverted vA >= C.Count (k + 1)
+        honSchemaHasKP1AfterIntersection =
+          BV.countActivesInV S.notInverted vH
+            >= C.toSize (C.Count (k + 1) + A.arPrefix testRecipeA)
+
     pure $
-      (BV.countActivesInV S.notInverted vA >= C.Count (k + 1))
-      QC.==>
       (QC.counterexample (unlines $
                             H.prettyChainSchema schedH "H"
                             ++ H.prettyChainSchema schedA "A"
@@ -178,8 +182,10 @@ prop_kPlus1BlocksAfterIntersection someTestAdversarial testSeedA = runIdentity $
       $ QC.counterexample ("arPrefix = " <> show (A.arPrefix testRecipeA))
       $ QC.counterexample ("stabWin  = " <> show stabWin)
       $ QC.counterexample ("stabWin' = " <> show (C.joinWin winA stabWin))
-      $ BV.countActivesInV S.notInverted vH
-          >= C.toSize (C.Count (k + 1) + A.arPrefix testRecipeA)
+      $ QC.checkCoverage
+      $ QC.cover 99 altSchemaHasKP1AfterIntersection
+          "Alternative schema has k+1 after intersection"
+      $ honSchemaHasKP1AfterIntersection
       )
 
 -- | No seed exists such that each 'A.checkAdversarialChain' rejects the result of 'A.uniformAdversarialChain'
```
This way, the test will fail if the coverage ever drops below 99%. See [`checkCoverage`](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#v:checkCoverage) for context/caveats (might be flaky, might make the test run longer).

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -637,15 +637,29 @@ withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
 --
 -- The count will be strictly smaller than the number of active slots in the given 'ChainSchema'.
 --
+-- The following precondition ensures that there are enough slots to produce an
+-- alternative chain schema with at least k+1 slots.
```
nfrisby, line 641: Did you mean "active slots"?

```suggestion
-- alternative chain schema with at least k+1 active slots.
```

nfrisby: Why do we want to ensure the alternative chain has at least k+1 (active) slots?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -104,10 +104,16 @@ data NoSuchHonestChainSchema =
...
+  (Kcp k, Scg s, Delta d) <- genKSD
+  -- Ensure that there are k + 1 slots in the chain:
+  -- 2s has 2k blocks, but if the windows overlap as in 2s-k, then
+  -- only k blocks might be present.
+  -- Therefore we enlarge the schema by one slot to get the size 2s-k+1
+  --
+  -- If we then add k+d slots, we get enough room for alternative chains
+  -- to have k+1 blocks when they branch of blocks in the first s slots.
+  l <- (+ (2*s + d + 1)) <$> QC.choose (0, 5 * sz)
+  pure $ HonestRecipe (Kcp k) (Scg s) (Delta d) (Len l)
```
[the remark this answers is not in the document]
nfrisby: (Same question as on my other comment: Why do we want to ensure the alternative chain has at least k+1 (active) slots?)

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
nfrisby: Why do we want this behavior?

If we use it for the honest schema, then it's unlikely we'll generate "minimally dense" honest chains, which seems like a mistake.

If we use it for alternative schemas, then it's unlikely we'll generate sparse alternative chains. That seems less like a mistake, but it's also not obvious that we never want to test with those, is it?

nfrisby (commented):
It's been so long since you asked me to review that I don't recall the original motivation. Sorry.

I suspect that's why my initial comments here are mostly just asking why we want the new schema property you're ensuring here.

I can imagine why we'd want to enforce them _sometimes_. But:

- Do we always want them? Seems like a potential gap in our search space.
- From the properties I've seen so far, my instinct is that there's a more straight-forward way to ensure them... eg a call to `fillInWindow` rather than eg choosing higher values of `asc`. But maybe that's insufficient (eg if the honest schema is minimally dense then the only way to also have k+1 blocks after the intersection on the alternative schema is to extend the end of time 🤔 ).

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
amesgen: > Why do we want this behavior?

From the PR description:

> Having more than k blocks after intersections makes the inputs interesting for the long range attack, and for testing that rollbacks of more than k can't happen in caught up nodes.

For long range attacks, it ensures that the node (at least if implemented correctly) did commit to a chain at the end of the test, which is what is tested for. I don't remember the exact reason motivation in the context of the rollback schedules, maybe @Niols can expand on that?

---

> If we use it for the honest schema, then it's unlikely we'll generate "minimally dense" honest chains, which seems like a mistake.

It is only used for alternative schemas, not for honest ones.

> If we use it for alternative schemas, then it's unlikely we'll generate sparse alternative chains. That seems less like a mistake, but it's also not obvious that we never want to test with those, is it?

Yeah, this is a good point, I think this might be indicative that parameterizing the generators over such desiderata might be useful (in order to make sure we can satisfy them when necessary, but still test with the full exhaustiveness in other cases), also see https://github.com/input-output-hk/ouroboros-consensus/pull/477#discussion_r1400438813.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
amesgen: (see https://github.com/input-output-hk/ouroboros-consensus/pull/477#discussion_r1401765035)

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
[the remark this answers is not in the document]
amesgen: (see https://github.com/input-output-hk/ouroboros-consensus/pull/477#discussion_r1401765035)

amesgen:
> From the properties I've seen so far, my instinct is that there's a more straight-forward way to ensure them... eg a call to fillInWindow rather than choosing higher values of asc. But maybe that's insufficient (eg if the honest schema is minimally dense then the only way to also have k+1 blocks after the intersection on the alternative schema is to extend the end of time 🤔 ).

FTR: @facundominguez first tried some things by patching up the output of the schema generator, but it was surprisingly subtle, see https://github.com/input-output-hk/ouroboros-consensus/pull/477#issuecomment-1788263214.

facundominguez:
I think @amesgen answers well.

About parametrization of the generators, I think it is still not totally justified for two factors.

1. If we want to test alternative chains with less than k+1 blocks, we can generate point schedules (or block trees) that ignore the suffixes.
2. If we want to test sparser alternative chains, we can improve by increasing the probability of having k+1 blocks in the whole chain (even if it is long), instead of doing what we currently do: make density k/s probable over the whole alternative chain.
3. I think trying to use `fillInWindow` can be done easily enough if we generate longer alternative chains (sometimes `3*s` long). I was reluctant to enlarge our chains, but perhaps it is still to be preferred by simplicity.

facundominguez:
I brought `fillInWindow` back to activate the last k slots of alternative chain schemas. The chains are guaranteed to be long enough that the last k slots are after the acceleration bound.

The first active slot of an alternative chain schema is guaranteed to be before $s+k+d+1$ from the end, which allows ensuring the acceleration bound is always before the end.

Race windows do not push the acceleration bound to the last $k$ blocks because the intersections with the honest chain are always behind the last block preceding the last $s+k+d+1$ slots.

If this looks good to you I'll squash and merge in `genesis/milestone6`.

nfrisby (approved):
I have no objections to this PR.

We've talked about it a lot over the past couple weeks. We have other ideas we'd prefer (notably: generate the adv first such that it definitely has k+1 after the intersection, then generate the honest based on that), but they will require a bigger change.

So this approach to avoiding the onerous `discard`s is sufficient for now. The downsides (eg longer chains than strictly necessary) are acceptable in the short-term.

facundominguez:
Looks like this PR was closed accidentally by deleting the genesis/point-scheduler branch.

PR #483, opened 2023-11-01
Fix changelog scripts for CI

jasagredo:
# Description

The CI check will fail if the latest version in the changelog is different from the version in the cabal file.

It will also fail if the PR doesn't have the `no changelog` tag and there are haskell/cabal changes but no corresponding fragment.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   consensus-changelog:
 
-    name: Changelog sanity check
+    name: Changelogs
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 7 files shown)


dnadales (approved):

.github/workflows/checks.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   consensus-changelog:
 
-    name: Changelog sanity check
+    name: Changelogs
```
amesgen, line 16: I think the branch protection rules have to be adapted due to these renamings.

jasagredo: Will do

PR #489, opened 2023-11-02
Use the tick duration when waiting in the scheduler, not the slot length

tek:
I've made a terrible mistake


Niols (approved):
Haven't we all, at some point in our lifes?

PR #493, opened 2023-11-06
Decouple db-analyser internals from cardano details (`BlockType`)

jorisdral:
# Description

Currently, the `DBAnalyserConfig` datatype has a `BlockType` filed, but `BlockType` is specific to Cardano: it can either be a Byron, Shelley or Cardano blocktype. This means we can not use db-analyser with other types of blocks. This PR removes the `BlockType` field from `DBAnalyserConfig` (and the `DBTruncatorConfig`), since that field was largely unused anyway.

### DIFF
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE CPP           #-}
 
-module DBAnalyser.Parsers (parseCmdLine, blockTypeParser) where
+module DBAnalyser.Parsers (parseCmdLine, blockTypeParser, BlockType (..)) where
 
 import           Cardano.Crypto (RequiresNetworkMagic (..))
 import           Cardano.Tools.DBAnalyser.Block.Byron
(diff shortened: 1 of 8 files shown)


ouroboros-consensus-cardano/app/db-analyser.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
```
jorisdral, line 3: To delete

ouroboros-consensus-cardano/app/db-truncater.hs
```diff
@@ -1,9 +1,11 @@
+{-# LANGUAGE RecordWildCards #-}
+
```
jorisdral, line 2: To delete

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -159,6 +160,18 @@ pMaybeOutputFile =
       <> completer (bashCompleter "file")
       )
 
+parseMaybe ::  Parser a -> Parser (Maybe a)
+parseMaybe parser = asum [Just <$> parser, pure Nothing]
```
amesgen, line 164: Minor: As you are touching this: This could be deleted and replaced by [`optional`](https://hackage.haskell.org/package/base-4.19.0.0/docs/Control-Applicative.html#v:optional) at call sites.

amesgen (approved):

PR #495, opened 2023-11-07
Integration work for node 8.7.0

jasagredo:
# Description

- Update the ledger packages and collect changelogs.
- Expose `srnChainSyncTimeout`.


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -280,7 +280,7 @@ library
     , time
     , transformers
     , typed-protocols              ^>=0.1.1
-    , vector                       ^>=0.12
+    , vector                       ^>=0.12    || ^>=0.13
```
amesgen, line 283: Hmm, maybe we want to actually drop 0.12 support? Or is it still required upstream? As we talked about recently, we might otherwise accidentally drop support for (ie make compilation fail with) 0.12 without noticing it.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -807,6 +807,8 @@ data StdRunNodeArgs m blk (p2p :: Diffusion.P2P) = StdRunNodeArgs
   , srnMaybeMempoolCapacityOverride :: Maybe MempoolCapacityBytesOverride
     -- ^ Determine whether to use the system default mempool capacity or explicitly set
     -- capacity of the mempool.
+  , srnChainSyncTimeout             :: Maybe NTN.ChainSyncTimeout
+    -- ^ A custom timeout for ChainSync.
```
amesgen, line 811: Given that `stdChainSyncTimeout` uses `IO` for randomisation
```haskell
stdChainSyncTimeout :: IO ChainSyncTimeout
```
do we want to take a `Maybe (m NTN.ChainSyncTimeout)` here? Or is it clear that this extra generality is unnecessary?

amesgen (approved):
LGTM after CI in a corresponding CHaP PR suceeded :raised_hands:

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
amesgen: See #509

PR #496, opened 2023-11-08
Adjust number of test cases

Niols:
For now, we have relied on `withMaxSuccess` to reduce the number of tests. This does not allow for an increase of the number of tests in CI or nightly. This PR uses the proper mechanism.


Niols:
The PR seems to make the “can rollback” test flaky because it discards way too many objects. I don't think we should merge it until that is taken care of.

facundominguez (approved):
LGTM

PR #499, opened 2023-11-09
Use `bracketChainSync` in the peer simulator

tek:
This requires us to pass around the candidate fragments as a TVar of TVars.

This will become more significant in terms of code duplication once the governor is merged.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
@@ -26,7 +26,8 @@ import qualified Data.Map.Strict as Map
 import           Data.Traversable (for)
 import           Ouroboros.Consensus.Config (TopLevelConfig (..))
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client (ChainDbView,
-                     Consensus, chainSyncClient, defaultChainDbView)
+                     Consensus, bracketChainSyncClient, chainSyncClient,
+                     defaultChainDbView)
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -158,36 +160,41 @@ startChainSyncConnectionThread ::
...
-              traceUnitWith tracer ("ChainSyncClient " ++ condense srPeerId) "Terminating because of time limit exceeded."
-            Nothing ->
-              pure ()
-          case fromException exn of
-            Just ThreadKilled ->
-              traceUnitWith tracer ("ChainSyncClient " ++ condense srPeerId) "Terminated by GDD governor."
-            _ ->
-              pure ()
-        Right _ -> pure ()
+      bracketChainSyncClient nullTracer chainDbView varCandidates srPeerId ntnVersion $ \ varCandidate -> do
```
facundominguez, line 169: I'm not too concerned with the change. Any hint of what `bracketChainSyncClient` does and why it wasn't needed so far?

facundominguez (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
tek: it adds and removes the peer to/from the candidates (and later handles) vars. we just inlined it so far.
it also starts some block validity monitor thing, I don't think it does much for us

PR #500, opened 2023-11-09
Normalise adjustment of number of QuickCheck tests

Niols:
_This PR has been initially devised as part of the Genesis project, in https://github.com/input-output-hk/ouroboros-consensus/pull/496. It now targets `main` because we believe it might benefit the project as a whole._

### Description

This PR aims at normalising the adjustments of the number of QuickCheck tests in the whole codebase.

`ouroboros-consensus:unstable-consensus-testlib` provides utilities in a module `Test.Util.TestEnv` to help describing a difference of behaviour of tests when ran in a `dev` (default), `ci` or `nightly` environment. This is typically used to change the number of test cases that QuickCheck generates. This number will be 100 in `dev`, 10,000 in `ci` and 100,000 in `nightly` by default.

Now, for some tests, 100 is too much and so we end up wanting to run ten times less than that (or any other ratio, really). The proper way to do this without hard-coding a value (say 10 tests) which would then be the same independently of the environment is by adjusting the value option `QuickCheckTests`, with something like:
```haskell
adjustOption (\(QuickCheckTests n) -> QuickCheckTests (1 `max` (div n 10)))
  :: TestTree -> TestTree
```

This is what is being done in various places of the code base in slightly different ways. This PR introduces a helper `adjustQuickCheckTests` and uses it consistently across the code base. This helper allows writing “let's run 10⨉ less tests” as
```haskell
adjustQuickCheckTests (`div` 10)
  :: TestTree -> TestTree
```

### Bonus questions

I have two bonus questions which I think could be taken care of in this PR.

First, I have seen in various place a helper:
```haskell
adjustTestEnv :: TestTree -> TestEnv -> TestTree
adjustTestEnv tree = \case
  Nightly -> tree
  _       -> adjustQuickCheckTests (`div` 10) tree
```
I have the impression that this could be simplified as simply ``adjustQuickCheckTests (`div` 10)``. What do you think?

Second, a lot of places hardcode a number of tests with `localOption (QuickCheckTests 14)` for instance. I would be tempted to replace this by something like ``adjustQuickCheckTests (`div` 7)``, which would give a similar result but would behave better when test environments are indeed used in CI or nightly. What do you think?

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
@@ -143,15 +143,6 @@ instance Arbitrary TestSetup where
 
   -- TODO shrink
 
--- | Run relatively fewer tests
---
--- These tests are slow, so we settle for running fewer of them in this test
--- suite since it is invoked frequently (eg CI for each push).
-oneTenthTestCount :: QuickCheckTests -> QuickCheckTests
-oneTenthTestCount (QuickCheckTests n) = QuickCheckTests $
-    if 0 == n then 0 else
-    max 1 $ n `div` 10
-
 tests :: TestTree
 tests = testGroup "ShelleyAllegra ThreadNet" $
     [ let name = "simple convergence" in
@@ -163,7 +154,10 @@ tests = testGroup "ShelleyAllegra ThreadNet" $
       adjustTestMode :: TestTree -> TestEnv -> TestTree
       adjustTestMode tree = \case
         Nightly -> tree
-        _       -> adjustOption oneTenthTestCount tree
+        _       ->
+          -- These tests are slow, so we settle for running fewer of them in this test
+          -- suite since it is invoked frequently (eg CI for each push).
+          adjustQuickCheckTests (`div` 10) tree
 
 prop_simple_shelleyAllegra_convergence :: TestSetup -> Property
 prop_simple_shelleyAllegra_convergence TestSetup
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
@@ -154,10 +154,7 @@ tests = testGroup "ShelleyAllegra ThreadNet" $
       adjustTestMode :: TestTree -> TestEnv -> TestTree
       adjustTestMode tree = \case
         Nightly -> tree
-        _       ->
-          -- These tests are slow, so we settle for running fewer of them in this test
-          -- suite since it is invoked frequently (eg CI for each push).
-          adjustQuickCheckTests (`div` 10) tree
+        _       -> adjustQuickCheckTests (`div` 10) tree
 
 prop_simple_shelleyAllegra_convergence :: TestSetup -> Property
 prop_simple_shelleyAllegra_convergence TestSetup
--- a/ouroboros-consensus-cardano/test/byron-test/Test/ThreadNet/Byron.hs
+++ b/ouroboros-consensus-cardano/test/byron-test/Test/ThreadNet/Byron.hs
@@ -77,6 +77,7 @@ import           Test.Util.HardFork.Future (singleEraFuture)
 import           Test.Util.Orphans.Arbitrary ()
 import           Test.Util.Slots (NumSlots (..))
 import qualified Test.Util.Stream as Stream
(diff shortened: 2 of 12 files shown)


dnadales:
Thank you :raised_hands: 

> I have the impression that this could be simplified as simply adjustQuickCheckTests (`div` 10). What do you think?

```haskell
adjustTestEnv :: TestTree -> TestEnv -> TestTree
adjustTestEnv tree = \case
  Nightly -> tree
  _       -> adjustQuickCheckTests (`div` 10) tree
``` 

I don't understand, sorry. Don't we still want to run the unaltered number of tests in the `Nightly` case? 

>  I would be tempted to replace this by something like adjustQuickCheckTests (`div` 7)

Sounds sensible to me :+1:

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
```diff
@@ -163,7 +154,10 @@ tests = testGroup "ShelleyAllegra ThreadNet" $
       adjustTestMode :: TestTree -> TestEnv -> TestTree
       adjustTestMode tree = \case
         Nightly -> tree
-        _       -> adjustOption oneTenthTestCount tree
+        _       ->
+          -- These tests are slow, so we settle for running fewer of them in this test
```
dnadales, line 158: I see this comment repeated multiple times. Perhaps we can mention somewhere else why one would want to use `adjustQCTests` (using this comment), and use line 160 as an example. Then one might not need to add these comments. What do you think?

dnadales (approved):
Looks great. Thanks :heart:

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
Niols: I was also finding it a bit silly to move the same comment over and over again. I suppose a good place could be the documentation of `adjustQuickCheckTests`? Would that be sufficient? Or would you imagine something else elsewhere?

dnadales: Yes, also think `adjustQuickCheckTests` is a good place :+1:

Niols: What do you think of 8413f17b9?

dnadales: Looks good. Consider adding an example of how to use it, eg: 

```haskell
-- > adjustQuickCheckTests (`div` 10) 
-- 
-- will reduce the default number of tests by 10
```

Or something similar. But it's also fine as it is.

Niols:
> > I have the impression that this could be simplified as simply adjustQuickCheckTests (`div` 10). What do you think?
> 
> ```haskell
> adjustTestEnv :: TestTree -> TestEnv -> TestTree
> adjustTestEnv tree = \case
>   Nightly -> tree
>   _       -> adjustQuickCheckTests (`div` 10) tree
> ```
> 
> I don't understand, sorry. Don't we still want to run the unaltered number of tests in the `Nightly` case?

Reading my own comment, I was very unclear; sorry about that. There are three possible test environments at the moment: `Dev`, `Ci` and `Nightly`. The default is already to run 100 tests in `Dev`, 10,000 in `Ci` and 100,000 in `Nightly`. I suppose those numbers come from  What this `adjustTestEnv` function here does (even before my change) is to change it to 10 tests in `Dev`, 1,000 in `Ci` but keep it at 100,000 in `Nightly`. Replacing the whole thing by simply ``adjustQuickCheckTests (`div` 10)`` would change it to 10 tests in `Dev`, 1,000 in `Ci` and 10,000 in `Nightly`, which sounds good enough to me (also, I would be worried that maybe 100,000 would just be too much time). I have little context so I don't really know what these tests are supposed to be, so this was really meant as a very naive question and I am obviously very fine keeping it as-is!

> > I would be tempted to replace this by something like adjustQuickCheckTests (`div` 7)
> 
> Sounds sensible to me 👍

Done in ff2a88395.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
Niols: Great idea! Done in 6b55de0fb.

dnadales:
> Done in [ff2a883](https://github.com/input-output-hk/ouroboros-consensus/commit/ff2a88395663ca61f5fa069d85c2ff50a4e86633).

Perfect, thanks!

PR #501, opened 2023-11-09
Generate schedules for single peers

facundominguez:
Do not review yet

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer.hs
@@ -1,0 +1,324 @@
+-- | This module contains functions for generating random point schedules.
+--
+-- A point schedule is a set of tables, having one table per simulated peer.
+-- Each row of a table correspond to a point in time, also called a tick.
+--
+-- Each tick has a timestamp and a state to set for the corresponding peer at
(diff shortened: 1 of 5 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer.hs
```diff
@@ -0,0 +1,324 @@
...
+mergeOn :: Ord b => (a -> b) -> [a] -> [a] -> [a]
+mergeOn _f [] ys = ys
+mergeOn _f xs [] = xs
+mergeOn f xxs@(x:xs) yys@(y:ys) =
+    if f x <= f y
+      then x : mergeOn f xs yys
+      else y : mergeOn f xxs ys
+
+zipMany :: [a] -> [[b]] -> [[(a, b)]]
+zipMany xs0 = snd . mapAccumL (go []) xs0
```
tek, line 320: should you assert that the final `acc` is `[]`?

facundominguez:
Squashed to `genesis/point-scheduler-squashed`

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer.hs
facundominguez: Do you mean `fst . mapAccumL (go []) xs0`? Maybe.

PR #512, opened 2023-11-16
Drop GHC 9.2, make 9.6 the default

amesgen:
GHC 9.2 shows runtime performance regressions in the systme-level benchmarks, but the numbers look much better for 9.6, so it is very unlikely Node releases will ever be built with GHC 9.2. Hence, we can drop it here, and switch the default to GHC 9.6.

Also update various tools.


amesgen:
Currently waiting for https://github.com/input-output-hk/ouroboros-network/pull/4716

dnadales (approved):

PR #514, opened 2023-11-17
Add flag to LongRangeAttack for running in IO

tek:
This is only to experiment with better performance for development. It still idles about 60% of the time because we're using static threadDelay instead of waiting for completion of the clients.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
@@ -2,16 +2,20 @@
 {-# LANGUAGE DerivingStrategies  #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LongRangeAttack.hs
```diff
@@ -79,4 +85,13 @@ prop_longRangeAttack honestFreq advFreqs = do
     isHonestTestHeaderHash :: HeaderHash TestBlock -> Bool
     isHonestTestHeaderHash = all (0 ==) . unTestHash
 
-    scheduleConfig = defaultPointScheduleConfig
+    scheduleConfig = PointScheduleConfig {pscTickDuration}
+
+    pscTickDuration | useIOSim = 0.1
+                    | otherwise = 0.001
```
facundominguez, line 91:
```suggestion
    scheduleConfig | useIOSim = defaultPointScheduleConfig
                   | otherwise = defaultPointScheduleConfig {pscTickDuration=0.001}
```

facundominguez (approved):
LGTM

Issue #230, opened 2023-07-13
Establish a pipeline for feeding ledger ops data to P&T analysis


dnadales:
Required for https://github.com/input-output-hk/ouroboros-consensus/issues/223

- [ ] Generate input data for the benchmark-ledger-ops analysis.
- [x] Agree on the output format.
- [x] Have the benchmark-ledger-ops analysis to output data in said format.
- [ ] Feed the output data to the P&T toolchain for variance analysis.

dnadales:
Next steps, as discussed with @mgmeier. 

- [ ] Consider adding `deepseq` to [this line](https://github.com/input-output-hk/ouroboros-consensus/blob/87f86cd9fb33127be9af2512f570dff9a14adc1e/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs#L520).
- [ ] `DataPoint` should be output as JSON.
- [ ]`DataPoint` should be imported as a library so that the P&T tooling can use it.
- [ ] We will use a synthesized chain as input data.
- [ ] We need to annotate the output file with metadata:
    - RTS flags used.
    - Compiler version.
    - Package versions (`cardano-ledger` and `ouroboros-consensus`) or the `git` commit hash that correspond to the `db-analyser` version built.
- [ ] We need to set the same RTS flags as `cardano-node` (to constraint resources in the same way).

We'll pass the result of running the `ledger-ops` analysis to @mgmeier for variance analysis. We'll output a JSON file in the following format:

```javascript
{ metadata: { ... }
, datapoints: [ ... ]
}
```

dnadales:
We do not want to add `deepseq` to [this line](https://github.com/input-output-hk/ouroboros-consensus/blob/87f86cd9fb33127be9af2512f570dff9a14adc1e/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs#L520) because:

- it would inflate the measured time because we would be traversing the whole structure (`r` in the snippet above), and crucially, we do not use `deepseq` in our actual code.
- our types should be fully strict (with the exception of fingertree-like internal lazyness) such that `deepseq` would be a no-op (sematically, not performance-wise).

dnadales:
As discussed with @mgmeier, the variance analysis shows that `db-analyser` gives reproducible results. A PR with further improvements to `beacon` will be submitted by @mgmeier.

PR #518, opened 2023-11-21
db-analyser: add option to output results of "benchmark ledger ops " as JSON

dnadales:
Part of #230 

TODOs:

- [x] Add roundtrip test for `BenchmarkLedgerOps.Metadata`.
- [x] Reference the latest commit hash (see https://github.com/input-output-hk/ouroboros-consensus/pull/519).
- [x] Decide if any implementation is necessary to ensure we use the same RTS flags as `cardano-node` (perhaps we need to modify the `cabal` file or the `nix` derivation that builds `db-analyser`.
- [x] Output: `minor_gcs = GC.gcs - GC.major_gcs`.
- [x] Output `GC.allocated_bytes`
- [x] Remove roundtripping test and keep them in a branch (delete after ~1 year).

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/Metadata.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/Metadata.hs
@@ -1,0 +1,56 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+-- | Functions related to obtaining information about the 'db-analyser' run.
+--
+-- Metadata includes information such as:
+--
+-- - RTS flags.
+-- - Compiler version.
+-- - OS and architecture.
+--
+-- See 'Metadata' and 'getMetadata' for more details.
+--
+module Cardano.Tools.DBAnalyser.Analysis.BenchmarkLedgerOps.Metadata (
+    Metadata (..)
+  , getMetadata
+  ) where
+
+import           Cardano.Tools.GitRev (gitRev)
+import           Data.Aeson (ToJSON)
+import qualified Data.Aeson as Aeson
+import qualified Data.Text as T
+import qualified Data.Version
+import           Data.Word (Word32, Word64)
+import           GHC.Generics (Generic)
+import qualified GHC.RTS.Flags as RTS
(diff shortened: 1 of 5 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/Metadata.hs
```diff
@@ -0,0 +1,58 @@
...
+  , machineArchitecture         :: String
+  , gitRevison                  :: String
+  } deriving (Generic, Show, Eq)
+
+instance ToJSON Metadata where
+  toEncoding = Aeson.genericToEncoding Aeson.defaultOptions
+
+instance FromJSON Metadata
+
+getMetadata :: IO Metadata
```
mgmeier, line 45: ❤️

mgmeier (approved):
Thank you so much @dnadales, this is great.

ouroboros-consensus-cardano/changelog.d/20231123_120121_damian.only_230_add_option_to_output_data_points_as_JSON.md
jasagredo, line 1: I'd argue that by our conventions this entry should _not_ exist because it refers to an `unstable-` component and we don't record those. From the README:

```
The Consensus sublibraries---which are used for our internal testing---are visible and buildable by default only because the appropriate Cabal and/or Nix infrastructure to guard them behind an explicit opt-in is currently immature and fragile. WARNING breaking changes to the these libraries are not reflected in the package versions. That is why they all have the unstable- prefix in their name; use at your own risk, and please reach out to us if this policy of ours is an excessive burden on your use case.
```

Now that I read it, I'm unsure if this entry should exist. Perhaps it should but it should not influence the bumping of the version? that sounds wrong

dnadales: Yeah, good one. The thing is that this change affects downstream consumers. So maybe it should not be categorized as unstable? But .... on the other hand we want to prevent `db-analyser` from being built by `cardano-node`. No idea what's the best solution here.

mgmeier (approved):
Thanks @dnadales

PR #519, opened 2023-11-21
cardano-tools: allow to reference latest git commit hash

amesgen:
This PR enables to reference the commit hash in any of our cardano-tools. This works both when building via Cabal and Nix.

As an example, run one of

    cabal run db-analyser -- -h
    nix run .#db-analyser -- -h

and observe the commit hash at the bottom.

When the commit hash can't be obtained (eg because there is no `git` in the PATH, or the libraries were built in a context where the `.git` directory is not available), then the commit hash will be all zeroes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,6 +51,7 @@
             inputs.iohkNix.overlays.crypto
             inputs.haskellNix.overlay
             inputs.iohkNix.overlays.haskell-nix-crypto
+            inputs.iohkNix.overlays.haskell-nix-extra
             (import ./nix/tools.nix inputs)
             (import ./nix/haskell.nix inputs)
             (import ./nix/pdfs.nix)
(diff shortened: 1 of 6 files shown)


dnadales:
Thank you Alex! 

> then the commit hash will be all zeroes

Perhaps we could replace that by a message that explains exactly this:

> the commit hash can't be obtained (eg because there is no git in the PATH, or the libraries were built in a context where the .git directory is not available),

amesgen:
> > then the commit hash will be all zeroes
> 
> Perhaps we could replace that by a message that explains exactly this:
> 
> > the commit hash can't be obtained (eg because there is no git in the PATH, or the libraries were built in a context where the .git directory is not available),

Good point, added a more informative message in that case :+1:
Also included an indicator whether the git repo was dirty.

dnadales (approved):

PR #521, opened 2023-11-24
UTxO-HD: LedgerDB API rewrite and V1 implementation

### LINKED ISSUE #510: Prototype LedgerDB/BackingStore redesign to accommodate LSM-tree backend
To integrate the LSM-tree backend for UTxO-HD we need to redesign the Consensus storage layer, and this new design might enable us to implement an alternative in-memory backing store that will have a very similar performance and resource requirements as the current Cardano node.

We should prototype the redesign of the `LedgerDB` and `BackingStore` components so that they can be integrated with the LSM-tree backend. 

Definition of Done: make a prototype and present it to the consensus team

jorisdral:
# Description

Resolves #510

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
@@ -1,0 +1,789 @@
+{-# LANGUAGE CPP                      #-}
+{-# LANGUAGE ConstraintKinds          #-}
+{-# LANGUAGE DataKinds                #-}
+{-# LANGUAGE DeriveAnyClass           #-}
+{-# LANGUAGE DeriveGeneric            #-}
+{-# LANGUAGE DerivingStrategies       #-}
+{-# LANGUAGE FlexibleContexts         #-}
+{-# LANGUAGE FlexibleInstances        #-}
+{-# LANGUAGE FunctionalDependencies   #-}
+{-# LANGUAGE GADTs                    #-}
+{-# LANGUAGE LambdaCase               #-}
+{-# LANGUAGE QuantifiedConstraints    #-}
+{-# LANGUAGE RankNTypes               #-}
+{-# LANGUAGE ScopedTypeVariables      #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE TypeApplications         #-}
+{-# LANGUAGE TypeFamilies             #-}
+{-# LANGUAGE TypeOperators            #-}
+{-# LANGUAGE UndecidableInstances     #-}
+
+-- | The Ledger DB is responsible for the following tasks:
+--
+-- - __Maintaining the in-memory ledger state at the tip__: When we try to
+--     extend our chain with a new block fitting onto our tip, the block must
+--     first be validated using the right ledger state, i.e., the ledger state
+--     corresponding to the tip.
+--
+-- - __Maintaining the past \(k\) in-memory ledger states__: we might roll back
+--     up to \(k\) blocks when switching to a more preferable fork. Consider the
+--     example below:
+--
+--     <<docs/haddocks/ledgerdb-switch.svg>>
+--
+--     Our current chain's tip is \(C_2\), but the fork containing blocks
+--     \(F_1\), \(F_2\), and \(F_3\) is more preferable. We roll back our chain
+--     to the intersection point of the two chains, \(I\), which must be not
+--     more than \(k\) blocks back from our current tip. Next, we must validate
+--     block \(F_1\) using the ledger state at block \(I\), after which we can
+--     validate \(F_2\) using the resulting ledger state, and so on.
+--
+--     This means that we need access to all ledger states of the past \(k\)
+--     blocks, i.e., the ledger states corresponding to the volatile part of the
+--     current chain. Note that applying a block to a ledger state is not an
+--     invertible operation, so it is not possible to simply /unapply/ \(C_1\)
+--     and \(C_2\) to obtain \(I\).
+--
+--     Access to the last \(k\) ledger states is not only needed for validating
+--     candidate chains, but also by the:
+--
+--     - __Local state query server__: To query any of the past \(k\) ledger
+--       states.
+--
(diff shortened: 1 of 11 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -0,0 +1,789 @@
...
+--
+-- Thrown upon incorrect use: invalid input.
+data LedgerDbError blk =
+      -- | The LedgerDB is closed.
+      --
+      -- This will be thrown when performing some operations on the LedgerDB. The
+      -- 'CallStack' of the operation on the LedgerDB is included in the error.
+      ClosedDBError PrettyCallStack
+      -- | A Forker is closed.
+    | ClosedForkerError PrettyCallStack
```
jasagredo, line 268: I'm dubious these CallStacks are useful. Have you tried throwing an error on purpose? I think it will only say where it was thrown, not where the ledgerdb function was called, unless you add the constraint to the methods (and even in that case, it might only report the place where it was created?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -0,0 +1,789 @@
...
+       -> [Ap m l blk c]  -- ^ New blocks to apply
+       -> m (Either ExceededRollback (Forker bm l blk))
+switch ldb cfg numRollbacks trace newBlocks = do
+  foEith <- liftBase $ getForkerAtFromTip ldb numRollbacks
+  case foEith of
+    Left rbExceeded -> pure $ Left rbExceeded
+    Right fo -> do
+      case newBlocks of
+        [] -> pure ()
+        -- no blocks to apply to ledger state, return current DbChangelog
```
jasagredo, line 507: This is not a dbchangelog anymore

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -0,0 +1,789 @@
...
+toRealPoint :: HasHeader blk => Ap m l blk c -> RealPoint blk
+toRealPoint (ReapplyVal blk) = blockRealPoint blk
+toRealPoint (ApplyVal blk)   = blockRealPoint blk
+toRealPoint (ReapplyRef rp)  = rp
+toRealPoint (ApplyRef rp)    = rp
+toRealPoint (Weaken ap)      = toRealPoint ap
+
+-- | Apply block to the current tip of the db changelog
+--
+-- We take in the entire 'DbChangelog' because we record that as part of errors.
```
jasagredo, line 562: Same here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -0,0 +1,789 @@
...
+  where
+    withValues :: blk -> (l ValuesMK -> m (l DiffMK)) -> m (l DiffMK)
+    withValues blk f = do
+        l <- liftBase $ atomically $ forkerGetLedgerState fo
+        vs <- withLedgerTables l
+              <$> liftBase (forkerReadTables fo (getBlockKeySets blk))
+        f vs
+
+-- | If applying a block on top of the ledger state at the tip is succesful,
+-- extend the DbChangelog with the resulting ledger state.
```
jasagredo, line 596: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -0,0 +1,789 @@
...
+  deriving (Show, Eq)
+
+newtype PushGoal blk = PushGoal { unPushGoal :: RealPoint blk }
+  deriving (Show, Eq)
+
+newtype Pushing blk = Pushing { unPushing :: RealPoint blk }
+  deriving (Show, Eq)
+
+data TraceValidateEvent blk =
+    -- | Event fired when we are about to push a block to the DbChangelog
```
jasagredo, line 779: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API/DiskPolicy.hs
```diff
@@ -0,0 +1,127 @@
...
+  ) where
+
+import           Control.Monad.Class.MonadTime.SI
+import           Data.Time.Clock (secondsToDiffTime)
+import           Data.Word
+import           GHC.Generics
+import           NoThunks.Class
+import           Ouroboros.Consensus.Config
+
+-- | Length of time that has to pas after which a snapshot is taken.
```
jasagredo, line 19:
```suggestion
-- | Length of time that has to pass after which a snapshot is taken.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/BackingStore.hs
```diff
@@ -107,6 +107,7 @@ data BackingStoreSelector m where
 -- in use.
 data BackingStoreTraceByBackend = LMDBTrace BackingStoreTrace
                                 | InMemoryTrace BackingStoreTrace
+                                | InitEvent TraceBackingStoreInitEvent
```
jasagredo, line 110: This will have to be reworked later on

jasagredo (approved):
Haddocks should link to each other, but we can probably do this later on.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
jorisdral: I haven't -- I'm just copying what the ChainDB is doing

PR #525, opened 2023-11-28
Do not even propagate future headers

nfrisby:
@njd42 and I came up with this idea a couple weeks ago, during a chat about the vagaries of NTP/clocks/etc.

See the commit message of the 'consensus: do not even propagate future headers' commit.

The main concrete benefits:

- It's a step towards Chronos and that helps simplify/ground our story around future blocks.
- A network-load optimization of sorts: if someone with a horribly out of sync clock sends a block 60 seconds early, the direct neighbors won't even propagate it.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -213,7 +213,7 @@ mkHandlers
   -- ^ Peer Sharing result computation callback
   -> Handlers       m addrNTN           blk
 mkHandlers
-      NodeKernelArgs {keepAliveRng, miniProtocolParameters}
+      NodeKernelArgs {chainSyncFutureCheck, keepAliveRng, miniProtocolParameters}
       NodeKernel {getChainDB, getMempool, getTopLevelConfig, getTracers = tracers}
       computePeers =
     Handlers {
@@ -224,6 +224,7 @@ mkHandlers
               (chainSyncPipeliningHighMark miniProtocolParameters))
             (contramap (TraceLabelPeer peer) (Node.chainSyncClientTracer tracers))
             getTopLevelConfig
+            chainSyncFutureCheck
             (defaultChainDbView getChainDB)
       , hChainSyncServer = \peer _version ->
           chainSyncHeadersServer
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -55,6 +55,8 @@ import           Ouroboros.Consensus.Ledger.SupportsPeerSelection
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.Mempool
 import qualified Ouroboros.Consensus.MiniProtocol.BlockFetch.ClientInterface as BlockFetchClientInterface
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck
+                     (HeaderInFutureCheck)
 import           Ouroboros.Consensus.Node.Run
 import           Ouroboros.Consensus.Node.Tracers
 import           Ouroboros.Consensus.Protocol.Abstract
@@ -132,6 +134,7 @@ data NodeKernelArgs m addrNTN addrNTC blk = NodeKernelArgs {
     , btime                   :: BlockchainTime m
     , chainDB                 :: ChainDB m blk
     , initChainDB             :: StorageConfig blk -> InitChainDB m blk -> m ()
+    , chainSyncFutureCheck    :: HeaderInFutureCheck m blk
     , blockFetchSize          :: Header blk -> SizeInBytes
     , mempoolCapacityOverride :: MempoolCapacityBytesOverride
     , miniProtocolParameters  :: MiniProtocolParameters
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Fragment/InFuture.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Fragment/InFuture.hs
@@ -16,8 +16,9 @@ module Ouroboros.Consensus.Fragment.InFuture (
     -- * Clock skew
   , clockSkewInSeconds
   , defaultClockSkew
-    -- ** opaque
+    -- ** not exporting the constructor
   , ClockSkew
+  , unClockSkew
     -- * Testing
   , dontCheck
   , miracle
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -36,7 +36,8 @@ module Ouroboros.Consensus.MiniProtocol.ChainSync.Client (
   , TraceChainSyncClientEvent (..)
   ) where
 
-import           Control.Monad.Except
+import           Control.Monad (join)
+import           Control.Monad.Except (runExcept, throwError)
 import           Control.Tracer
 import           Data.Kind (Type)
 import           Data.Map.Strict (Map)
@@ -51,12 +52,16 @@ import           NoThunks.Class (unsafeNoThunks)
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.Forecast
+import           Ouroboros.Consensus.HardFork.History
+                     (PastHorizonException (PastHorizon))
 import           Ouroboros.Consensus.HeaderStateHistory
                      (HeaderStateHistory (..), validateHeader)
 import qualified Ouroboros.Consensus.HeaderStateHistory as HeaderStateHistory
 import           Ouroboros.Consensus.HeaderValidation hiding (validateHeader)
+import           Ouroboros.Consensus.Ledger.Basics (LedgerState)
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.Storage.ChainDB (ChainDB,
@@ -64,6 +69,7 @@ import           Ouroboros.Consensus.Storage.ChainDB (ChainDB,
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
 import           Ouroboros.Consensus.Util
 import           Ouroboros.Consensus.Util.Assert (assertWithMsg)
+import qualified Ouroboros.Consensus.Util.EarlyExit as EarlyExit
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.STM (Fingerprint, Watcher (..),
                      WithFingerprint (..), withWatcher)
@@ -426,6 +432,7 @@ chainSyncClient
     => MkPipelineDecision
     -> Tracer m (TraceChainSyncClientEvent blk)
     -> TopLevelConfig blk
+    -> InFutureCheck.HeaderInFutureCheck m blk
     -> ChainDbView m blk
     -> NodeToNodeVersion
     -> ControlMessageSTM m
@@ -433,6 +440,13 @@ chainSyncClient
     -> StrictTVar m (AnchoredFragment (Header blk))
     -> Consensus ChainSyncClientPipelined blk m
 chainSyncClient mkPipelineDecision0 tracer cfg
+                InFutureCheck.HeaderInFutureCheck
+                { -- these fields in order of use
+                  proxyArrival        = Proxy :: Proxy arrival
+                , recordHeaderArrival
+                , judgeHeaderArrival
+                , handleHeaderArrival
+                }
                 ChainDbView
                 { getCurrentChain
                 , getHeaderStateHistory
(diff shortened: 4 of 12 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/InFutureCheck.hs
```diff
@@ -0,0 +1,144 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck (
```
nfrisby, line 8: @njd42 Please confirm this logic matches what you and I chatted about a few weeks ago. Thanks!

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -2,8 +2,8 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE RankNTypes          #-}
-{-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
 -- | Tests for the chain sync client.
```
nfrisby, line 9: @amesgen Please confirm that this module is testing strictly _more_ than it was before. For that purpose, I split the changes into two parts. I'm hoping it's relatively trivial to see that the first commit is purely a refactor (... modulo some innocuous RTS thread scheduling, maybe?). And the second one is purely additive.

nfrisby: Notably: some blocks are from the future now. But not far future, since those trivialize the test (maybe we should include those even so?).

nfrisby:
FYI, the PR is draft because I've so far only propagated the ChainSync client interface change to its test. I'll tackle the rest tomorrow.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -706,104 +717,151 @@ chainSyncClient mkPipelineDecision0 tracer cfg
...
-            -- precondition: the intersection is within the last @k@ blocks of
-            -- the current chain.
-            forecast <-
+            lst <-
               maybe
                 (error $
                    "intersection not within last k blocks: " <> show mostRecentIntersection)
-                (ledgerViewForecastAt (configLedger cfg) . ledgerState)
+                ledgerState
                 <$> getPastLedger mostRecentIntersection
```
amesgen, line 841: I was first confused by the `join` as this function doesn't actually perform any monadic effects apart from the `atomically`, everything is in `STM`, we just use `join` for `WithEarlyExit`.

It could be rewritten like this
```haskell
    readLedgerState kis prj = atomically $ do
        -- We must first find the most recent intersection with the current
        -- chain. Note that this is cheap when the chain and candidate haven't
        -- changed.
        kis'@KnownIntersectionState { mostRecentIntersection } <-
          EarlyExit.earlyExit $ intersectsWithCurrentChain kis
        lst <- EarlyExit.lift $
          maybe
            (error $
               "intersection not within last k blocks: " <> show mostRecentIntersection)
            ledgerState
            <$> getPastLedger mostRecentIntersection
        case prj lst of
          Nothing         -> retry
          Just ledgerView -> pure $ Intersects kis' ledgerView
```
by making use of the fact that
```haskell
type STM (WithEarlyExit m) = WithEarlyExit (STM m)
```
Wrinkle is that `EarlyExit.earlyExit` is not exported:
https://github.com/input-output-hk/ouroboros-consensus/blob/6614998b855d19e947bd9d7a166458a2baea66d4/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/EarlyExit.hs#L60-L62
Seems like a rather natural function to me (could also be renamed to `exitOnNothing`), and of course doesn't actually add any power.

amesgen (commented):
Looks great, mostly minor comments and a few suggestions. Will review again after the chores (formatting, compilation failures, changelog entries, etc.).

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -312,99 +337,98 @@ runChainSync securityParam (ClientUpdates clientUpdates)
...
+      let loop tick = do
+              updateChainsDuringTick tick
+              when (tick == startSyncingAt) $ initiateChainSync
+              when (tick < finalTick)       $ loop (tick + 1)
+      loop (Tick 1)
+
+      -- This delay seems enough to let all threads finish their final work.
+      --
+      -- TODO what is the necessary threshold?
+      threadDelay 86400
```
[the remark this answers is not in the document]
nfrisby: Nothing concrete. I guessed at some thresholds that seemed intuitive, but nothing worked. So I bumped it to something more obviously an intentional over-approximation (seconds in a day) even from a `mainnet` perspective... as opposed to the previous very-arbitrary "2000".

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -261,7 +281,12 @@ runChainSync securityParam (ClientUpdates clientUpdates)
     -- at the final state of each candidate.
     varFinalCandidates <- uncheckedNewTVarM Map.empty
 
-    (tracer, getTrace) <- first (LogicalClock.tickTracer clock) <$> recordingTracerTVar
+    (tracer, getTrace) <- do
+          (tracer', getTrace) <- recordingTracerTVar
+          let foo ev = do
```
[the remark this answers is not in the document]
nfrisby: Actually, I explicitly chose `foo` because...... j/k

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Fragment/InFuture.hs
```diff
@@ -18,6 +18,7 @@ module Ouroboros.Consensus.Fragment.InFuture (
   , defaultClockSkew
     -- ** opaque
   , ClockSkew
+  , unClockSkew
```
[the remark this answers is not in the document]
nfrisby: I'm _pretty_ sure the idea was to prevent constructing a `ClockSkew` the `Num` instance of its payload, since that could easily lead to mistakes about which units are involved.

So I changed "opaque" to say something more along those lines.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -706,104 +717,151 @@ chainSyncClient mkPipelineDecision0 tracer cfg
...
-        -- Before obtaining a 'LedgerView', we must find the most recent
-        -- intersection with the current chain. Note that this is cheap when
-        -- the chain and candidate haven't changed.
+
+        mLedgerView <- EarlyExit.withEarlyExit $ do
+          Intersects kis2 lst        <- checkArrivalTime kis arrival
+          Intersects kis3 ledgerView <- case projectLedgerView slotNo lst of
+              Just ledgerView -> pure $ Intersects kis2 ledgerView
+              Nothing         -> readLedgerState kis2 (projectLedgerView slotNo)
+          pure $ Intersects kis3 ledgerView
```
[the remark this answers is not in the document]
nfrisby: I really need to get around to writing my `unlet` proposal one of these days...

Correct me if I'm wrong: but there's still no way to locally disable name shadowing warnings, right?

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -532,6 +568,13 @@ slotLength = slotLengthFromSec $ toEnum slotLengthInSeconds
 slotLengthInSeconds :: Int
 slotLengthInSeconds = 20
 
+-- | Multiply a diff by a scalar coefficient
+--
+-- The right conversions to use are somewhat tricky. The key fact is that
+-- 'fromInteger' interprets its argument as seconds.
+scaleNominalDelay :: Integral a => NominalDiffTime -> a -> NominalDiffTime
+scaleNominalDelay dur i = dur * (fromInteger . toInteger $ i)
```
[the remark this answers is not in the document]
nfrisby: Done, thanks.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -427,6 +456,13 @@ runChainSync securityParam (ClientUpdates clientUpdates)
   where
     k = maxRollbacks securityParam
 
+    toSkewedOnset :: SlotNo -> RelativeTime
+    toSkewedOnset slot = RelativeTime $
+        scaleNominalDelay
+          (getSlotLength slotLength)
+          (fromEnum $ unSlotNo slot)
```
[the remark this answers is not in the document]
nfrisby: Done, thanks.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -566,17 +637,26 @@ instance Arbitrary ChainSyncClientSetup where
           ]
     invalidBlocks <- InvalidBlocks <$> (genSchedule =<< shuffle trapBlocks)
 
+    clientSlowBy <- do
+       numSlots  <- choose (0, 5)
+       reduction <- choose (0, 3)   -- so it's sometimes not aligned
+
+       return $ clockSkewInSeconds $
+         max 0 $ toEnum slotLengthInSeconds * numSlots - reduction
```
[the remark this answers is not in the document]
nfrisby: Ah ha! I was definitely thinking in terms of `Int`s when I wrote those `chooses`. I've simplified. Thanks.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -611,6 +691,9 @@ instance Arbitrary ChainSyncClientSetup where
             , lastTick (getServerUpdates serverUpdates)  - 1
             ]
     , startTick' <- [1..min startTick maxStartTick]
+    ] <>
+    [ cscs { clientSlowBy = clockSkewInSeconds 0 }
+    | clientSlowBy /= clockSkewInSeconds 0
     ]
 
 instance Show ChainSyncClientSetup where
```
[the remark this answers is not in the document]
nfrisby: Done, thanks. I also changed it to a proper `Show` instance (ie copy-pasteable) since that's useful for QuickCheck counterexamples. I relegated this `condense` stuff to a separate call to `counterexample`.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
nfrisby: Done, thanks!

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -108,6 +113,7 @@ prop_chainSync ChainSyncClientSetup {
   , serverUpdates
   , startTick
   , invalidBlocks
+  , clientSlowBy
   } =
     counterexample
```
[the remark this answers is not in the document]
nfrisby: I've pushed up a commits that do both of these things. Thanks.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: I think I prefer the way it is, with the `join`. I pushed up a commit that reveals an intermediate type signature---is that commit enough to make the type shenanigans easier to follow?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -199,7 +201,8 @@ data Handlers m addr blk = Handlers {
 
 mkHandlers
   :: forall m blk addrNTN addrNTC.
-     ( IOLike m
+     ( HasHardForkHistory blk
```
nfrisby, line 204: Every block should be able to instantiate this class, so I don't think this addition is a detriment.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -129,6 +130,8 @@ data NodeKernelArgs m addrNTN addrNTC blk = NodeKernelArgs {
       tracers                 :: Tracers m (ConnectionId addrNTN) addrNTC blk
     , registry                :: ResourceRegistry m
     , cfg                     :: TopLevelConfig blk
+    , tolerableClockSkew      :: ClockSkew
```
nfrisby, line 133: @amesgen I'd appreciate your thoughts on whether these two fields are the right ones to add.

- It's kind of weird to have both `SystemTime` and `BlockchainTime`.

- Maybe we want to cordon these off into a `chainSyncArgs` hierarchical field?

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -103,7 +133,16 @@ tests = testGroup "ChainSyncClient"
...
-prop_chainSync ChainSyncClientSetup {..} =
+prop_chainSync testSetup@ChainSyncClientSetup {
+    securityParam
+  , clientUpdates
+  , serverUpdates
+  , startTick
+  , invalidBlocks
+  , clientSlowBy
+  } =
+    tabulate "TickArrivalTimeStats" [show (tickArrivalTimeStats traceEvents)] $
```
nfrisby, line 144: Upon inspection, I was surpsised that ~70% of the cases had `SomeNotEarly_SomeEarly Zero Zero`, which implies the test never involved downloading a header!

My initial guess is that the client and server chain are diverging before they ever reach `startTick`.

I suspect we want to tune the generator to visit that less often. But that can be a separate Issue.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
amesgen: a

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: I still feel that
```haskell
    -- Used in 'rollForward': block until the the ledger state at the
    -- intersection with the local selection returns 'Just'.
    --
    -- Exits early if the intersection no longer exists.
    readLedgerState :: KnownIntersectionState blk
                    -> (LedgerState blk -> Maybe a)
                    -> EarlyExit.WithEarlyExit m (Intersects blk a)
    readLedgerState kis prj = atomically $ do
        -- We must first find the most recent intersection with the current
        -- chain. Note that this is cheap when the chain and candidate haven't
        -- changed.
        EarlyExit.lift (intersectsWithCurrentChain kis) >>= \case
          Nothing   -> EarlyExit.exitEarly
          Just kis' -> do
            let KnownIntersectionState { mostRecentIntersection } = kis'
            lst <- EarlyExit.lift $
              maybe
                (error $
                   "intersection not within last k blocks: " <> show mostRecentIntersection)
                ledgerState
                <$> getPastLedger mostRecentIntersection
            case prj lst of
              Nothing         -> retry
              Just ledgerView -> pure $ Intersects kis' ledgerView
```
is strictly more direct, but this is not a major point (especially with the intermediate type signature, it should be hard to be confused also with the current approach, despite being a bit verbose).

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -312,99 +392,135 @@ runChainSync securityParam (ClientUpdates clientUpdates)
...
+        checkTipTime = do
+            (now, candidates) <- atomically $ do
+              now        <- readTVar varLocalClock
+              candidates <- readTVar varCandidates >>= traverse readTVar
+              pure (now, candidates)
+            forM_ candidates $ \candidate -> case AF.headSlot candidate of
+              Origin  -> pure ()
+              At slot ->
+                when (now < toSkewedOnset slot) $ do
+                  error $ "future tip! " <> show (slot, now, candidate)
```
amesgen, line 482: Hmm, this check does not seem to fail when eg commenting out the `threadDelay` in `InFutureCheck`. I think this is due to using `readTVar varLocalClock`, which is completely unaffected by whether the ChainSync client calls `threadDelay` or not.

---

 - I find it plausible that this check is actually detecting all violations of the "Chronos principle", but eg only because `slotNo == blockNo` in this test, so it is eg not possible that during a logical tick, an client with an incorrect delay logic could add a header `H` with a very high slot number to the fragment, only to roll back to say `H'` with a sufficiently small one (which would be missed by this check). Maybe mention this, or say that this check isn't even intended to *fully* check the Chronos principle, but rather just a sanity check?
   
   Alternatives to the approach here are ofc also possible (probably a bit less lightweight), eg a new trace event that is emitted on header arrival, or some watcher of the candidate fragment that wakes up on every change to it (relying on io-sim that we don't miss any updates).
 - Due to the use of `error`, the QuickCheck counterexample information is missing on failure. This might be fine if we are confident that this check failing will always be trivial to fix without any counterexample context.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -638,6 +836,65 @@ ppBlocks :: Point TestBlock -> [TestBlock] -> String
...
+      let onset = RelativeTime $
+            scaleNominalDelay
+              (getSlotLength slotLength)
+              (unSlotNo (blockSlot hdr))
+          thisSlot = Set.singleton n
+      in
+      if now < onset
+      then OnlyNotEarly_SomeEarly { onlyNotEarlyTATS = Set.empty, someEarlyTATS = thisSlot }
+      else OnlyNotEarly_SomeEarly { onlyNotEarlyTATS = thisSlot, someEarlyTATS = Set.empty }
+    _ -> mempty
```
amesgen, line 900: This `now` here comes from `readTVar varLocalClock` which is only updated at the start of logical ticks, but it can be the case that during a logical tick in the ChainSync client, before the arrival of a header `H`, another header `H'` arrived from the future (and hence caused a call of `threadDelay`), such that `H` is *not* from the future when it arrives in the ChainSync client even though this logic would categorize it as such.

This probably doesn't make a huge difference, but it is sufficiently tricky that it maybe should at least be called out. Another option would be to use `getCurrentTime` in `pairWithNow`.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Good point, I created #529 for this.

> My initial guess is that the client and server chain are diverging before they ever reach `startTick`.

Yeah, that seems be the case, when always setting `startTick` to `1`, the `Zero Zero` case shrinks to ~9% of all cases.

amesgen (changes requested):

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Hmm, yeah, I like that better.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: Thanks for the note. I've made some changes and it does not fail if I inject that bug.

Primary changes:

- I replaced `varLocalClock` with merely the io-sim wall clock. This seems less error prone but does require a bit more ceremony in the setup.
- I replaced `toSkewedOnset` with `toOnset` in that conditional.
- The invariant violation is no longer handled via an exception; so the `counterexample`s et all work.
- Found a cheeky bug in the infrastructure: if the local chain gets ahead of the clock (due to clock skew), and then the peer rolls back to that point, the invariant can be violated, since the intersection is now ahead of the wall clock. Hence `mkClockUpdates` now distinguishes the greatest-slots so that the clock updating logic can handle them separately: exclude skew when updating for new max client slot but include it when updating for a new max server slot.

I'm OK with rollbacks being missed, since a separate QC counterexample wouldn't hide the violation in that way. I'll add a comment in the code.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: Hmm. I think you're saying `H` and `H'` are both from the same logical tick, right?

In that case, it's fine. The semantics of `someEarlyTATs` is a count of how many logical ticks include at least one future header. It's not counting how many headers themselves.

Is that right?

(I also don't see how `getCurrentTime` would help.)

nfrisby:
Just FYI, I just now rebased onto `origin/main`---there were no conflicts.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -638,6 +836,65 @@ ppBlocks :: Point TestBlock -> [TestBlock] -> String
...
+-------------------------------------------------------------------------------}
+
+data TickArrivalTimeStats a = OnlyNotEarly_SomeEarly {
+    onlyNotEarlyTATS   :: !a
+    -- ^ Logical ticks in which some headers are arriving but none are from the future.
+  , someEarlyTATS :: !a
+    -- ^ Logical ticks in which some headers are arriving from the future
+  }
+  deriving (Functor)
+  deriving (Generic)
```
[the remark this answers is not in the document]
nfrisby: I used the the `cardano-base` repo's `base-deriving-via` package, which is basically `Generically`.

nfrisby:
CI is green 👍

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
amesgen: Addressed :+1:

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -312,99 +412,164 @@ runChainSync securityParam (ClientUpdates clientUpdates)
...
+    -- If the candidate's tip's slot's onset is ahead of the local wall-clock
+    -- (which is skewed by 'clientSlowBy'), then the ChainSync client
+    -- mishandled a block from the future.
+    let checkTipTime :: m ()
+        checkTipTime = do
+            (now, candidates) <- do
+              now        <- systemTimeCurrent systemTime
+              candidates <-
+                atomically $ readTVar varCandidates >>= traverse readTVar
+              pure (now, candidates)
```
amesgen, line 503:
```suggestion
            now        <- systemTimeCurrent clientSystemTime
            candidates <-
              atomically $ readTVar varCandidates >>= traverse readTVar
```

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -638,6 +935,68 @@ ppBlocks :: Point TestBlock -> [TestBlock] -> String
 ppBlocks a bs = ppPoint a <> " ] " <> intercalate " :> " (map ppBlock bs)
 
 ppTraceEvent :: TraceEvent -> String
-ppTraceEvent (Tick n, ev) = show n <> " | " <> case ev of
+ppTraceEvent (Tick n, m, ev) = show (n, m) <> " | " <> case ev of
```
amesgen, line 938: For terse output:
```suggestion
ppTraceEvent (Tick n, RelativeTime m, ev) = show (n, m) <> " | " <> case ev of
```

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -312,99 +412,164 @@ runChainSync securityParam (ClientUpdates clientUpdates)
...
   where
     k = maxRollbacks securityParam
 
-    slotLength :: SlotLength
-    slotLength = slotLengthFromSec 20
+    toOnset :: SlotNo -> RelativeTime
+    toOnset slot = RelativeTime $
+        scaleNominalDelay
+          (getSlotLength slotLength)
+          (unSlotNo slot)
```
amesgen, line 563: What about making this a top-level declaration (instead of `scaleNominalDelay), such that it can also be used in `tickArrivalTimes`?

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Looks good, nice that this also revealed the fact that we should not set the clock in such a way that we consider our own adopted blocks to be from the (near) future (I guess one *could* frame this as modeling clock rewind, but I don't think that is too useful).

> I'm OK with rollbacks being missed, since a separate QC counterexample wouldn't hide the violation in that way.

It felt *slightly* weird to implement a non-exhaustive check when it wouldn't be that hard to implement, but I agree it is fine, especially as it seems quite hard to implement the Chronos property incorrectly in such a way that this code even make a difference.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: > Hmm. I think you're saying `H` and `H'` are both from the same logical tick, right?
> 
> In that case, it's fine. The semantics of `someEarlyTATs` is a count of how many logical ticks include at least one future header. It's not counting how many headers themselves.
> 
> Is that right?

Ah right, that is a very good point, it is irrelevant whether just `H` or both `H` and `H` are considered to be from the future if they are from the same tick :+1:

Also, my comment doesn't apply anymore anyway as `now` is now backed by `getCurrentTime` via `SystemTime`, so it incorporates the `threadDelay`s in the ChainSync client.

amesgen (approved):
LGTM (after squashing), only very minor comments left

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -237,18 +301,42 @@ data ChainSyncOutcome = ChainSyncOutcome {
...
     -> InvalidBlocks
     -> Tick  -- ^ Start chain syncing at this time
     -> m ChainSyncOutcome
-runChainSync securityParam (ClientUpdates clientUpdates)
+runChainSync skew securityParam (ClientUpdates clientUpdates)
     (ServerUpdates serverUpdates) (InvalidBlocks invalidBlocks)
     startSyncingAt = withRegistry $ \registry -> do
 
-    clock <- LogicalClock.new registry numTicks
+    systemTime <- do
```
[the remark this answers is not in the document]
nfrisby: > `clientSystemTime :: SystemTime m <-`

I generally avoid pattern guards with type variables, since they have subtle/sometimes surprising rules for binding variables. That's my only reason.

Good rename 👍

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: Thanks: it had been a common `atomically` before a refactor 👍

nfrisby:
FYI: I based a second time to sign the commits.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/InFutureCheck.hs
nfrisby: @njd42 It's being merged now, but I'd still appreciate your thoughts on this module. Thanks.

njd42 (commented):
It all looks good to me - I've added some comments as to why I believe that you have already effectively dealt with the leap second issue (as best as can be done)

PR #528, opened 2023-11-29
Relax text bound

michaelpj:
Forbidding text-1 causes problems in some cross builds.

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -146,9 +146,11 @@ library
     , sop-core                      ^>=0.5
     , sop-extras                    ^>=0.1
     , strict-sop-core               ^>=0.1
-    , text                          ^>=2.0
     , these                         ^>=1.2
     , vector-map
+
+  -- GHC 8.10.7 on aarch64-darwin cannot use text-2
+  build-depends:   text >=1.2.5.0 && <2.2
 
 library unstable-byronspec
   import:          common-lib
--- a/ouroboros-consensus-diffusion/ouroboros-consensus-diffusion.cabal
+++ b/ouroboros-consensus-diffusion/ouroboros-consensus-diffusion.cabal
@@ -84,10 +84,12 @@ library
     , random
     , serialise                    ^>=0.2
     , si-timers                    ^>=1.3
-    , text                         ^>=2.0
     , time
     , transformers
     , typed-protocols
+
+  -- GHC 8.10.7 on aarch64-darwin cannot use text-2
+  build-depends:      text >=1.2.5.0 && <2.2
 
 library unstable-diffusion-testlib
   import:          common-lib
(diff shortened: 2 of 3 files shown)


angerman (approved):
🥳

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -146,7 +146,10 @@ library
     , sop-core                      ^>=0.5
     , sop-extras                    ^>=0.1
     , strict-sop-core               ^>=0.1
-    , text                          ^>=2.0
+    -- text-2 requires some C++ parts, and won't build in
+    -- some cross situations. So we continue to allow 
+    -- text-1 for now
+    , text                          >= 1 && < 2.2
```
jasagredo, line 152: I wonder if we could gate this behind a conditional to only enable it in cross builds. I don't know what the `os` flag says when cross compiling.

jasagredo (approved):

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
michaelpj: I'm unsure if that buys you anything. In both cases you need to retain compatibility across the text versions, this way is if anything easier to test because you don't have to trigger the cross conditional.

jasagredo: It is indeed easier to test, but are we ever going to test on Linux for earlier versions? I don't think I have seen such a case yet

jasagredo: Unless something goes wrong on the compilation pipeline but in that case you can do `cabal build --allow-older --constraint "text ==<problematic-version>`

jasagredo: Actually the `os` flag says:

> The argument is tested against System.Info.os on the target system. 

So I think `os(windows)` will evaluate to true when cross-compiling for windows.

michaelpj:
Argh. You use `cabal-fmt` which has the downside of moving comments outside of `build-depends` fields. So the comment is going to have to be above, sorry.

jasagredo:
I think you can do:
```
library
  build-depends: ....

  -- Your comment
  if os(windows)
    build-depends: text >=1
  else
    build-depends: text ^>=2.0
```

And we don't usually merge main into our branches, we instead rebase our branches onto main

michaelpj:
I can do a separate stanza if you prefer, sure.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
michaelpj: Sure, but what does restricting it get you? The code has to work for the wider text range regardless! If you had a situation where, say, you had some code that was only conditionally included when not building on windows that required a newer text then I would see the point, but here I'm not sure that I do.

jasagredo (approved):

PR #535, opened 2023-11-29
Refactor BlockFetch to use the same abstraction as ChainSync

tek:


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/BlockFetch.hs
```diff
@@ -104,12 +99,12 @@ runBlockFetchClient
   => peer
   -> FetchClientRegistry peer (Header TestBlock) TestBlock m
   -> ControlMessageSTM m
-  -> m (AnchoredFragment TestBlock)
+  -> BlockFetchServer TestBlock (Point TestBlock) m ()
   -> m ()
-runBlockFetchClient peerId fetchClientRegistry controlMsgSTM getCurrentServerChain =
+runBlockFetchClient peerId fetchClientRegistry controlMsgSTM server =
```
Niols, line 104: Why is it called `runBlockFetchClient` when it runs both the client and the server? Isn't the equivalent for chain sync called `runChainSyncConnectionThread` or something?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -50,20 +55,19 @@ handlerFindIntersection ::
...
       fragment = fromJust $ BT.findFragment tipPoint blockTree
   case intersectWith fragment clientPoints of
     Nothing ->
-      pure (IntersectNotFound tip', [])
+      pure (Just (IntersectNotFound tip'), [])
     Just intersection -> do
       writeTVar currentIntersection intersection
-      pure (IntersectFound intersection tip', [])
-  where
+      pure (Just (IntersectFound intersection tip'), [])
```
Niols, line 70: What's the reason to return a `Maybe FindIntersect` if we always return a `Just`? Or did you mean to return `Nothing` when `BT.findFragment` returns `Nothing`?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -85,26 +89,24 @@ handlerRequestNext currentIntersection blockTree points =
...
   where
-    headerIsOrigin = pure Nothing
-
-    withHeader :: Point TestBlock -> Header TestBlock -> WriterT [String] (STM m) (Maybe RequestNext)
+    withHeader :: Point TestBlock -> WithOrigin (Header TestBlock) -> WriterT [String] (STM m) (Maybe RequestNext)
     withHeader intersection h =
       maybe noPathError (analysePath hp) (BT.findPath intersection hp blockTree)
       where
-        hp = AF.castPoint $ blockPoint h
+        hp = AF.castPoint $ withOrigin GenesisPoint blockPoint h
```
Niols, line 98: I feel `withOrigin GenesisPoint blockPoint` should be a named helper near `blockPoint`. Maybe `blockPoint'` or something?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -144,3 +146,49 @@ handlerRequestNext currentIntersection blockTree points =
...
+    AdvertisedPoints {} -> pure Nothing
+  where
+
+    withHpFragment (hpChain, bp) =
+      -- First, check whether the block point is on the same chain, and before or equal to, the header point.
+      case AF.rollback (blockPoint bp) hpChain of
+        Nothing ->
+          -- REVIEW: Should this be fatal?
+          error "Block point isn't on the same chain as header point"
+          -- pure (Just NoBlocks)
```
Niols, line 172: I think it should and we should document somewhere that this is an invariant of point schedules.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -144,3 +146,49 @@ handlerRequestNext currentIntersection blockTree points =
...
+        Just bpChain ->
+          -- Next, check whether the requested range is contained in the fragment before the block point.
+          -- REVIEW: this is Nothing if only _part_ of the range is on the current chain.
+          -- Is it correct that we don't want to send anything in this case?
```
Niols, line 176: So right now we either send all the required blocks or nothing? I suppose this makes sense between honest nodes, though, because an honest client will only ask for blocks that you do indeed have and you will always serve all of them. Calling @amesgen or @nfrisby though; they should know better.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -144,3 +146,49 @@ handlerRequestNext currentIntersection blockTree points =
...
+              -- need to give the client a chance to adapt the range before trying again with the next tick's
+              -- chain.
+              -- Otherwise, we simply have to wait for BP to advance sufficiently, and we block without sending
+              -- a message, to simulate a slow response.
+              case AF.withinFragmentBounds to hpChain of
+                False -> pure (Just NoBlocks)
+                True  -> pure Nothing
+            Just slice -> do
+              trace ("Sending slice " ++ terseFrag slice)
+              pure (Just (StartBatch (AF.toOldestFirst slice)))
```
Niols, line 192: Considering how tiny and rather trivial this is, I would maybe rather write it before the `Nothing` case. Like “if we have all the blocks, we serve them; this is now out of the way; the other case is more interesting and requires more thinking.”

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Resources.hs
```diff
@@ -63,25 +66,43 @@ data SharedResources m =
...
-
     -- | The current known intersection with the chain of the client.
     csrCurrentIntersection :: StrictTVar m (Point TestBlock),
 
     -- | The final server passed to typed-protocols.
-    csrServer :: ChainSyncServer (Header TestBlock) (Point TestBlock) (Tip TestBlock) m ()
+    csrServer :: ChainSyncServer (Header TestBlock) (Point TestBlock) (Tip TestBlock) m (),
+
+    -- | This action blocks while this peer is inactive in the point schedule.
+    csrTickStarted :: STM m ()
```
Niols, line 76: Although it's accurate, it's funny to say “this action blocks” when it is called `*started` :p

Niols (commented):
Very neat improvement in my opinion! Only a partial review at this point; will come back to it tomorrow.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
tek: hah, I just added that 😅 though `blockPoint` is in `ouroboros-network`, so I put it in this file

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
amesgen: I think this was clarified on [Slack](https://moduscreate.slack.com/archives/C05CEUFFH7Z/p1701272848904619). Answer: yes, it is correct to send `MsgNoBlocks` unless all blocks are available.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
tek: because it's the unified API of `runHandler`

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
tek: clarified on slack, updated the comment

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Resources.hs
tek: yes, I am very open for better naming and/or phrasing 😅

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
tek: this was a misconception; it has been removed.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
```diff
@@ -144,3 +152,49 @@ handlerRequestNext currentIntersection blockTree points =
...
+        -- This should be a precondition violation, but because it would make the schedule generator more complex, we
+        -- simply serve the missing blocks anyway here.
+        --
+        -- Otherwise, we simply have to wait for BP to advance sufficiently, and we block without sending
+        -- a message, to simulate a slow response.
+        case not (AF.withinFragmentBounds to hpChain) && AF.withinFragmentBounds (toPoint bp) hpChain of
+          True ->
+            case AF.sliceRange (fragmentUpTo "requested point" to) from to of
+              Just slice -> pure (Just (StartBatch (AF.toOldestFirst slice)))
+              Nothing    -> error "BlockFetch: Client requested "
```
facundominguez, line 188: Here I'd expect:

if the request is not in the chain of the HP serve the blocks (or send NoBlock)
else wait for the BP to be updated (this is the meaning of yielding Nothing IIUC)

I don't understand the relevance of checking if the BP is in the chain of HP.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/BlockFetch.hs
tek: there's also `startBlockFetchConnectionThread` in `Run.hs`. I've created a corresponding file for `ChainSync`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Handlers.hs
tek: oh I thought that was what you suggested – if BP is on HP's chain, but `to` isn't, we definitely have switched both to a fork without letting BP catch up.

Unless BP is so far back that it's before the fork intersection…what do you think?

tek: FWIW if we want to precisely detect it we could additionally check whether BP is on the chain that leads up to `to`

facundominguez:
Cherry-picked the commit to genesis/point-scheduler-squashed

Issue #101, opened 2023-05-22
Decide on implementation of caught-up-ness information needed by Network


amesgen:
There are two general ways to implement the functionality described in #91:

 - Implement a minimal version of the interface, likely based on some recency condition of the immutable/volatile tip.
 - Decide that the functionality nicely fits into the Genesis State Machine, and we should implement the full/non-trivial parts of the GSM already today.

For context, the Genesis State Machine (GSM) is a part of the current Genesis design that determines when to "disable" certain properties/functionalities of the Consensus layer, as part of "switching" from Genesis to Praos. However, it does not depend on other Genesis components and could be implemented today.

The goal of this ticket is to decide on the implementation route we want to take.

---

Some notes from the 1:1 with @nfrisby today:

 - In general, it seems feasible to use the GSM for that purpose; the core question is, in both cases: "Can we safely resume operation in Praos mode, or do we have to switch to a (in some way more expensive) Syncing mode?"
    - In particular, the transition rule "All peers claim no more headers" for "Syncing -> CaughtUp" also works without Genesis, as we can/have to assume that we are only connected to trusted peers (probably just one) during Syncing.
    - We should avoid transient intermediary states, such as entering the CaughtUp state on startup just to immediately leave it due to our tip being too old.
 - The current interface in https://github.com/input-output-hk/ouroboros-network/pull/4555 does not allow Consensus to "notify" the Networking layer that it should switch back to Syncing after we were caught up, as Network decides when to call the function, implemented by Consensus, that returns the ledger peers to use.
   
   But that might actually be fine, if the transition "CaughtUp -> Syncing" always involves losing all of our peers (in which case Network has to call the function). That should e.g. true when "closing the laptop lid", but also when falling behind due to CPU starvation, as our peers will disconnect from us when the fork gets too deep. We should check this in more detail.

nfrisby:
@amesgen and I discussed this during our call today. Here's the whole argument.

To be explicit: we're proposing implementing this simplified GSM, using Option 2 and Option B below.

-----

The broad question:

- Until Genesis is released, someone (eg the CF, IOG relays, etc) will run a (mostly-)centralized pool of trusted nodes.

- After Issue 91, the node will, at least on start-up and effectively continuously decide whether _all_ of the current upstream peers must be from that trusted pool. If _not_, the upstream peers can include untrusted nodes.

[...]

PR #808, opened 2023-12-02
Add the pre-Genesis State Machine for bootstrap peers

nfrisby:
The Genesis SoW will deliver a proper GSM. In the interim, this PR defines a GSM that is suitable for use with the bootstrap-peers intermediate solution (see https://github.com/input-output-hk/ouroboros-network/pull/4555).

The relevant high-level specification is rendered here https://ouroboros-consensus.cardano.intersectmbo.org/docs/for-developers/BootstrapPeersIER

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -31,3 +31,23 @@ tests: true
 benchmarks: true
 
 import: ./asserts.cabal
+
+source-repository-package
+    type: git
+    location: https://github.com/input-output-hk/ouroboros-network/
+    tag: 8ab444850b9d654110cd751a17c21db814bba593
+    subdir: monoidal-synchronisation
+            network-mux
+            ouroboros-network
+            ouroboros-network-api
+            ouroboros-network-framework
+            ouroboros-network-protocols
+            ouroboros-network-testing
+    --sha256: 1vw8r8csa6lq4330bwgddxa4nl2m1pw1ps5y6l7lw2zw3xdlr0hf
+
+source-repository-package
+    type: git
+    location: https://github.com/input-output-hk/io-sim
+    tag: 85d633d6f9c76ffb678b36396c8fd39b2ff7219e
+    subdir: io-sim
+    --sha256: 10f54xg88wq2w7ylg3n33f1yiswd6w4dam7fvl12raiqj0rk21qm
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -21,6 +21,7 @@ module Ouroboros.Consensus.Node (
   , StdRunNodeArgs (..)
   , stdBfcSaltIO
   , stdChainSyncTimeout
+  , stdGsmAntiThunderingHerdIO
   , stdKeepAliveRngIO
   , stdLowLevelRunNodeArgsIO
   , stdMkChainDbHasFS
@@ -66,6 +67,7 @@ import           Data.Hashable (Hashable)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe, isNothing)
+import           Data.Time (NominalDiffTime)
 import           Data.Typeable (Typeable)
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime hiding (getSystemStart)
@@ -82,6 +84,8 @@ import           Ouroboros.Consensus.Node.DbLock
 import           Ouroboros.Consensus.Node.DbMarker
 import           Ouroboros.Consensus.Node.ErrorPolicy
 import           Ouroboros.Consensus.Node.ExitPolicy
+import           Ouroboros.Consensus.Node.GSM (GsmNodeKernelArgs (..))
+import qualified Ouroboros.Consensus.Node.GSM as GSM
 import           Ouroboros.Consensus.Node.InitStorage
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Node.ProtocolInfo
@@ -222,6 +226,9 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
       -- | Ie 'bfcSalt'
     , llrnBfcSalt :: Int
 
+      -- | Ie 'gsmAntiThunderingHerd'
+    , llrnGsmAntiThunderingHerd :: StdGen
+
       -- | Ie 'keepAliveRng'
     , llrnKeepAliveRng :: StdGen
 
@@ -254,6 +261,10 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
 
       -- | node-to-client protocol versions to run.
     , llrnNodeToClientVersions :: Map NodeToClientVersion (BlockNodeToClientVersion blk)
+
+      -- | If the volatile tip is older than this, then the node will exit the
+      -- @CaughtUp@ state.
+    , llrnMaxCaughtUpAge :: NominalDiffTime
 
       -- | Maximum clock skew
     , llrnMaxClockSkew :: ClockSkew
@@ -365,8 +376,15 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                   , ChainDB.cdbVolatileDbValidation  = ValidateAll
                   }
 
-        chainDB <- openChainDB registry inFuture cfg initLedger
-                  llrnChainDbArgsDefaults customiseChainDbArgs'
+        let finalChainDbArgs =
+              mkFinalChainDbArgs
+                registry
+                inFuture
+                cfg
+                initLedger
+                llrnChainDbArgsDefaults
+                customiseChainDbArgs'
+        chainDB <- ChainDB.openDB finalChainDbArgs
 
         continueWithCleanChainDB chainDB $ do
           btime <-
@@ -385,17 +403,29 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
               , hfbtMaxClockRewind = secondsToNominalDiffTime 20
               }
 
-          nodeKernelArgs <-
-              fmap (nodeKernelArgsEnforceInvariants . llrnCustomiseNodeKernelArgs) $
-              mkNodeKernelArgs
-                registry
-                llrnBfcSalt
-                llrnKeepAliveRng
-                cfg
-                rnTraceConsensus
-                btime
-                (InFutureCheck.realHeaderInFutureCheck llrnMaxClockSkew systemTime)
-                chainDB
+          nodeKernelArgs <- do
+              durationUntilTooOld <- GSM.realDurationUntilTooOld
+                (configLedger cfg)
+                (ledgerState <$> ChainDB.getCurrentLedger chainDB)
+                llrnMaxCaughtUpAge
+                systemTime
+              let gsmMarkerFileView =
+                    case ChainDB.cdbHasFSGsmDB finalChainDbArgs of
+                        SomeHasFS x -> GSM.realMarkerFileView chainDB x
+              fmap (nodeKernelArgsEnforceInvariants . llrnCustomiseNodeKernelArgs)
+                $ mkNodeKernelArgs
+                    registry
+                    llrnBfcSalt
+                    llrnGsmAntiThunderingHerd
+                    llrnKeepAliveRng
+                    cfg
+                    rnTraceConsensus
+                    btime
+                    (InFutureCheck.realHeaderInFutureCheck llrnMaxClockSkew systemTime)
+                    chainDB
+                    llrnMaxCaughtUpAge
+                    (Just durationUntilTooOld)
+                    gsmMarkerFileView
           nodeKernel <- initNodeKernel nodeKernelArgs
           rnNodeKernelHook registry nodeKernel
 
@@ -602,13 +632,25 @@ openChainDB ::
       -- ^ Customise the 'ChainDbArgs'
   -> m (ChainDB m blk)
 openChainDB registry inFuture cfg initLedger defArgs customiseArgs =
-    ChainDB.openDB args
-  where
-    args :: ChainDbArgs Identity m blk
-    args = customiseArgs $
-             mkChainDbArgs registry inFuture cfg initLedger
-             (nodeImmutableDbChunkInfo (configStorage cfg))
-             defArgs
+    ChainDB.openDB
+  $ mkFinalChainDbArgs registry inFuture cfg initLedger defArgs customiseArgs
+
+mkFinalChainDbArgs
+  :: forall m blk. (RunNode blk, IOLike m)
+  => ResourceRegistry m
+  -> CheckInFuture m blk
+  -> TopLevelConfig blk
+  -> ExtLedgerState blk
+     -- ^ Initial ledger
+  -> ChainDbArgs Defaults m blk
+  -> (ChainDbArgs Identity m blk -> ChainDbArgs Identity m blk)
+      -- ^ Customise the 'ChainDbArgs'
+  -> ChainDbArgs Identity m blk
+mkFinalChainDbArgs registry inFuture cfg initLedger defArgs customiseArgs =
+    customiseArgs $
+      mkChainDbArgs registry inFuture cfg initLedger
+      (nodeImmutableDbChunkInfo (configStorage cfg))
+      defArgs
 
 mkChainDbArgs ::
      forall m blk. (RunNode blk, IOLike m)
@@ -642,21 +684,29 @@ mkNodeKernelArgs ::
   => ResourceRegistry m
   -> Int
   -> StdGen
+  -> StdGen
   -> TopLevelConfig blk
   -> Tracers m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
   -> BlockchainTime m
   -> InFutureCheck.SomeHeaderInFutureCheck m blk
   -> ChainDB m blk
+  -> NominalDiffTime
+  -> Maybe (GSM.WrapDurationUntilTooOld m blk)
+  -> GSM.MarkerFileView m
   -> m (NodeKernelArgs m addrNTN (ConnectionId addrNTC) blk)
 mkNodeKernelArgs
   registry
   bfcSalt
+  gsmAntiThunderingHerd
   keepAliveRng
   cfg
   tracers
   btime
   chainSyncFutureCheck
   chainDB
+  maxCaughtUpAge
+  gsmDurationUntilTooOld
+  gsmMarkerFileView
   = do
     return NodeKernelArgs
       { tracers
@@ -671,6 +721,12 @@ mkNodeKernelArgs
       , miniProtocolParameters  = defaultMiniProtocolParameters
       , blockFetchConfiguration = defaultBlockFetchConfiguration
       , keepAliveRng
+      , gsmArgs = GsmNodeKernelArgs {
+          gsmAntiThunderingHerd
+        , gsmDurationUntilTooOld
+        , gsmMarkerFileView
+        , gsmMinCaughtUpDuration = maxCaughtUpAge
+        }
       }
   where
     defaultBlockFetchConfiguration :: BlockFetchConfiguration
@@ -721,6 +777,9 @@ stdMkChainDbHasFS rootPath (ChainDB.RelativeMountPoint relPath) =
 
 stdBfcSaltIO :: IO Int
 stdBfcSaltIO = randomIO
+
+stdGsmAntiThunderingHerdIO :: IO StdGen
+stdGsmAntiThunderingHerdIO = newStdGen
 
 stdKeepAliveRngIO :: IO StdGen
 stdKeepAliveRngIO = newStdGen
@@ -839,12 +898,14 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo
                                     , rnPeerSharing
                                     }
                          StdRunNodeArgs{..} = do
-    llrnBfcSalt      <- stdBfcSaltIO
-    llrnKeepAliveRng <- stdKeepAliveRngIO
+    llrnBfcSalt               <- stdBfcSaltIO
+    llrnGsmAntiThunderingHerd <- stdGsmAntiThunderingHerdIO
+    llrnKeepAliveRng          <- stdKeepAliveRngIO
     pure LowLevelRunNodeArgs
       { llrnBfcSalt
       , llrnChainSyncTimeout = fromMaybe stdChainSyncTimeout srnChainSyncTimeout
       , llrnCustomiseHardForkBlockchainTimeArgs = id
+      , llrnGsmAntiThunderingHerd
       , llrnKeepAliveRng
       , llrnChainDbArgsDefaults =
           updateChainDbDefaults $ ChainDB.defaultArgs mkHasFS
@@ -881,6 +942,7 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo
             (supportedNodeToClientVersions (Proxy @blk))
       , llrnWithCheckedDB =
           stdWithCheckedDB (Proxy @blk) srnDatabasePath networkMagic
+      , llrnMaxCaughtUpAge = secondsToNominalDiffTime $ 20 * 60   -- 20 min
       , llrnMaxClockSkew =
           InFuture.defaultClockSkew
       }
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -120,6 +120,7 @@ import           Ouroboros.Network.NodeToNode (DiffusionMode (..),
                      ExceptionInHandler (..), MiniProtocolParameters,
                      NodeToNodeVersionData (..), RemoteAddress, Socket,
                      blockFetchPipeliningMax, defaultMiniProtocolParameters)
+import           Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers)
 import           Ouroboros.Network.PeerSelection.LedgerPeers
                      (LedgerPeersConsensusInterface (..))
 import           Ouroboros.Network.PeerSelection.PeerMetric (PeerMetrics,
@@ -189,6 +190,8 @@ data RunNodeArgs m addrNTN addrNTC blk (p2p :: Diffusion.P2P) = RunNodeArgs {
 
       -- | Network PeerSharing miniprotocol willingness flag
     , rnPeerSharing :: PeerSharing
+
+    , rnGetUseBootstrapPeers :: STM m UseBootstrapPeers
     }
 
 -- | Arguments that usually only tests /directly/ specify.
@@ -426,6 +429,7 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                     llrnMaxCaughtUpAge
                     (Just durationUntilTooOld)
                     gsmMarkerFileView
+                    rnGetUseBootstrapPeers
           nodeKernel <- initNodeKernel nodeKernelArgs
           rnNodeKernelHook registry nodeKernel
 
@@ -693,6 +697,7 @@ mkNodeKernelArgs ::
   -> NominalDiffTime
   -> Maybe (GSM.WrapDurationUntilTooOld m blk)
   -> GSM.MarkerFileView m
+  -> STM m UseBootstrapPeers
   -> m (NodeKernelArgs m addrNTN (ConnectionId addrNTC) blk)
 mkNodeKernelArgs
   registry
@@ -707,6 +712,7 @@ mkNodeKernelArgs
   maxCaughtUpAge
   gsmDurationUntilTooOld
   gsmMarkerFileView
+  getUseBootstrapPeers
   = do
     return NodeKernelArgs
       { tracers
@@ -727,6 +733,7 @@ mkNodeKernelArgs
         , gsmMarkerFileView
         , gsmMinCaughtUpDuration = maxCaughtUpAge
         }
+      , getUseBootstrapPeers
       }
   where
     defaultBlockFetchConfiguration :: BlockFetchConfiguration
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -550,8 +550,11 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                 | (version, blockVersion) <- Map.toList llrnNodeToClientVersions
                 ],
             Diffusion.daLedgerPeersCtx =
-              LedgerPeersConsensusInterface
-                (getPeersFromCurrentLedgerAfterSlot kernel)
+              LedgerPeersConsensusInterface {
+                  lpGetLatestSlot = getImmTipSlot kernel,
+                  lpGetLedgerPeers = fromMaybe [] <$> getPeersFromCurrentLedger kernel (const True),
+                  lpGetLedgerStateJudgement = getLedgerStateJudgement kernel
+                }
           }
 
         localRethrowPolicy :: RethrowPolicy
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
@@ -1,0 +1,494 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ViewPatterns        #-}
+
+-- | The Genesis State Machine decides whether the node is caught-up or not.
+module Ouroboros.Consensus.Node.GSM (
+    CandidateVersusSelection (..)
+  , DurationFromNow (..)
+  , GsmEntryPoints (..)
+  , GsmNodeKernelArgs (..)
+  , GsmView (..)
+  , MarkerFileView (..)
+  , WrapDurationUntilTooOld (..)
+    -- * Auxiliaries
+  , TraceGsmEvent (..)
+  , initializationLedgerJudgement
+    -- * Constructors
+  , realDurationUntilTooOld
+  , realGsmEntryPoints
+  , realMarkerFileView
+  ) where
+
+import qualified Cardano.Slotting.Slot as Slot
+import qualified Control.Concurrent.Class.MonadSTM.TVar as LazySTM
+import           Control.Monad (forever, join, unless)
+import           Control.Monad.Class.MonadSTM (MonadSTM, STM, atomically, check)
+import           Control.Monad.Class.MonadThrow (MonadThrow)
+import           Control.Monad.Class.MonadTimer (threadDelay)
+import qualified Control.Monad.Class.MonadTimer.SI as SI
+import           Control.Tracer (Tracer, traceWith)
+import           Data.Functor ((<&>))
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+import           Data.Time (NominalDiffTime)
+import qualified Ouroboros.Consensus.BlockchainTime.WallClock.Types as Clock
+import qualified Ouroboros.Consensus.HardFork.Abstract as HardFork
+import qualified Ouroboros.Consensus.HardFork.History as HardFork
+import qualified Ouroboros.Consensus.HardFork.History.Qry as Qry
+import qualified Ouroboros.Consensus.Ledger.Basics as L
+import           Ouroboros.Consensus.Storage.ChainDB.API (ChainDB)
+import           Ouroboros.Consensus.Util.NormalForm.StrictTVar (StrictTVar)
+import qualified Ouroboros.Consensus.Util.NormalForm.StrictTVar as StrictSTM
+import           Ouroboros.Network.PeerSelection.LedgerPeers.Type
+                     (LedgerStateJudgement (..))
+import           System.FS.API (HasFS, createDirectoryIfMissing, doesFileExist,
+                     removeFile, withFile)
+import           System.FS.API.Types (AllowExisting (..), FsPath, OpenMode (..),
+                     mkFsPath)
+import           System.Random (StdGen, uniformR)
+
+{-------------------------------------------------------------------------------
+  Interface
+-------------------------------------------------------------------------------}
+
+data DurationFromNow =
+    After !NominalDiffTime
+    -- ^ INVARIANT positive
+  |
+    Already
+    -- ^ This value represents all non-positive durations, ie events from the
+    -- past
+  deriving (Eq, Show)
+
+data CandidateVersusSelection =
+    CandidateDoesNotIntersect
+    -- ^ The GSM assumes that this is ephemeral
+    --
+    -- For example, the ChainSync client will either disconnect from the peer
+    -- or update the candidate to one that is not stale. It's also technically
+    -- possible that the selection is stale, which the ChainDB would also
+    -- resolve as soon as possible.
+  |
+    WhetherCandidateIsBetter !Bool
+    -- ^ Whether the candidate is better than the selection
+  deriving (Eq, Show)
+
+data GsmView m upstreamPeer selection candidate = GsmView {
+    antiThunderingHerd        :: Maybe StdGen
+    -- ^ An initial seed used to randomly increase 'minCaughtUpDuration' by up
+    -- to 15% every transition from OnlyBootstrap to CaughtUp, in order to
+    -- avoid a thundering herd phenemenon.
+    --
+    -- 'Nothing' should only be used for testing.
+  ,
+    candidateOverSelection    ::
+        selection -> candidate -> CandidateVersusSelection
+  ,
+    durationUntilTooOld       :: Maybe (selection -> m DurationFromNow)
+    -- ^ How long from now until the selection will be so old that the node
+    -- should exit the @CaughtUp@ state
+    --
+    -- 'Nothing' means the selection can never become too old.
+  ,
+    equivalent                :: selection -> selection -> Bool
+    -- ^ Whether the two selections are equivalent for the purpose of the
+    -- Genesis State Machine
+  ,
+    getChainSyncCandidates    ::
+        STM m (Map.Map upstreamPeer (StrictTVar m candidate))
+    -- ^ The latest candidates from the upstream ChainSync peers
+  ,
+    getChainSyncIdlers        :: STM m (Set.Set upstreamPeer)
+    -- ^ The ChainSync peers whose latest message claimed that they have no
+    -- subsequent headers
+  ,
+    getCurrentSelection       :: STM m selection
+    -- ^ The node's current selection
+  ,
+    minCaughtUpDuration       :: NominalDiffTime
+    -- ^ How long the node must stay in CaughtUp after transitioning to it from
+    -- OnlyBootstrap, regardless of the selection's age. This prevents the
+    -- whole network from thrashing between CaughtUp and OnlyBootstrap if
+    -- there's an outage in block production.
+    --
+    -- See also 'antiThunderingHerd'.
+  ,
+    setCaughtUpPersistentMark :: Bool -> m ()
+    -- ^ EG touch/delete the marker file on disk
+  ,
+    writeLedgerStateJudgement :: LedgerStateJudgement -> m ()
+    -- ^ EG update the TVar that the Diffusion Layer monitors
+  }
+
+-- | The two proper GSM states for boot strap peers
+--
+-- See the @BootstrapPeersIER.md@ document for their specification.
+--
+-- See 'initializationLedgerJudgement' for the @Initializing@ pseudo-state.
+data GsmEntryPoints m = GsmEntryPoints {
+    enterCaughtUp      :: forall neverTerminates. m neverTerminates
+    -- ^ ASSUMPTION the marker file is present on disk, a la
+    -- @'setCaughtUpPersistentMark' True@
+    --
+    -- Thus this can be invoked at node start up after determining the marker
+    -- file is present (and the tip is still not stale)
+  ,
+    enterOnlyBootstrap :: forall neverTerminates. m neverTerminates
+    -- ^ ASSUMPTION the marker file is absent on disk, a la
+    -- @'setCaughtUpPersistentMark' False@
+    --
+    -- Thus this can be invoked at node start up after determining the marker
+    -- file is absent.
+  }
+
+-----
+
+-- | Determine the initial 'LedgerStateJudgment'
+--
+-- Also initializes the persistent marker file.
+initializationLedgerJudgement ::
+     ( L.GetTip (L.LedgerState blk)
+     , Monad m
+     )
+  => m (L.LedgerState blk)
+  -> Maybe (WrapDurationUntilTooOld m blk)
+     -- ^ 'Nothing' if @blk@ has no age limit
+  -> MarkerFileView m
+  -> m LedgerStateJudgement
+initializationLedgerJudgement
+    getCurrentLedger
+    mbDurationUntilTooOld
+    markerFileView
+  = do
+    wasCaughtUp <- hasMarkerFile markerFileView
+    if not wasCaughtUp then pure TooOld else do
+        case mbDurationUntilTooOld of
+            Nothing -> return YoungEnough
+            Just wd -> do
+                sno <- L.getTipSlot <$> getCurrentLedger
+                getDurationUntilTooOld wd sno >>= \case
+                    After{}     -> return YoungEnough
+                    Already     -> do
+                        removeMarkerFile markerFileView
+                        return TooOld
+
+{-------------------------------------------------------------------------------
+  A real implementation
+-------------------------------------------------------------------------------}
+
+-- | The actual GSM logic for boot strap peers
+--
+-- See the @BootstrapPeersIER.md@ document for the specification of this logic.
+realGsmEntryPoints :: forall m upstreamPeer selection tracedSelection candidate.
+     ( SI.MonadDelay m
+     , SI.MonadTimer m
+     , Eq upstreamPeer
+     )
+  => (selection -> tracedSelection, Tracer m (TraceGsmEvent tracedSelection))
+  -> GsmView m upstreamPeer selection candidate
+  -> GsmEntryPoints m
+realGsmEntryPoints tracerArgs gsmView = GsmEntryPoints {
+    enterCaughtUp
+  ,
+    enterOnlyBootstrap
+  }
+  where
+    (cnvSelection, tracer) = tracerArgs
+
+    GsmView {
+        antiThunderingHerd
+      ,
+        candidateOverSelection
+      ,
+        durationUntilTooOld
+      ,
+        equivalent
+      ,
+        getChainSyncCandidates
+      ,
+        getChainSyncIdlers
+      ,
+        getCurrentSelection
+      ,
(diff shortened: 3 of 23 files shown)


nfrisby:
I squashed and rebased to `origin/main`. I'll consider later whether to split the commit up again.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -155,19 +188,59 @@ initNodeKernel
...
+                    computeDurationUntilTooOld cfg wd lst
+              , GSM.equivalent                = (==) `on` (AF.headPoint . fst)
+              , GSM.getChainSyncCandidates    = readTVar varCandidates
+              , GSM.getChainSyncIdlers        = readTVar varIdlers
+              , GSM.getCurrentSelection       = do
+                  headers        <- ChainDB.getCurrentChain  chainDB
+                  extLedgerState <- ChainDB.getCurrentLedger chainDB
+                  return (headers, ledgerState extLedgerState)
+              , GSM.setCaughtUpPersistentMark = \upd ->
+                  (if upd then GSM.touchMarkerFile else GSM.removeMarkerFile)
```
nfrisby, line 232: @bolt12 is seeing an error in his initial integration.

```
{
"app": [],
"at": "2024-01-11T07:27:57.71Z",
"data": {
"error": "ExceptionInLinkedThread \"ThreadId 46\" (FsError {fsErrorType = FsResourceDoesNotExist, fsErrorPath = /var/lib/cardano-node/db-testnet/gsm/CaughtUpMarker, fsErrorString = \"No such file or directory\", fsErrorNo = Just (Errno 2), fsErrorStack = CallStack(from HasCallStack):\n  prettyCallStack, called at src/System/FS/API/Types.hs:294:23 in fs-api-0.2.0.1-7CfJAP5WJBILIrvJjDwNip:System.FS.API.Types\n  ioToFsError, called at src/System/FS/IO.hs:90:41 in fs-api-0.2.0.1-7CfJAP5WJBILIrvJjDwNip:System.FS.IO\n  handleError,called at src/System/FS/IO.hs:86:23 in fs-api-0.2.0.1-7CfJAP5WJBILIrvJjDwNip:System.FS.IO\n  rethrowFsError, called at src/System/FS/IO.hs:68:36 in fs-api-0.2.0.1-7CfJAP5WJBILIrvJjDwNip:System.FS.IO\n  removeFile, called at src/System/FS/API.hs:130:5 in fs-api-0.2.0.1-7CfJAP5WJBILIrvJjDwNip:System.FS.API, fsLimitation = False})",
"kind": "ConsensusStartupException"
},
"env": "8.7.3:00000",
"host": "server-e",
"loc": null,
"msg": "",
"ns": [
"cardano.node"
],
"pid": "1925269",
"sev": "Critical",
"thread": "7"
}```

I'm this `removeMarkerFile` is involved somehow. (Or maybe a different one.)

File: ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs [hunk 2]
```diff
@@ -0,0 +1,327 @@
...
+    enterCaughtUp = do
+        atomically $ LazySTM.writeTVar varLedgerStateJudgement YoungEnough
+        setCaughtUpPersistentMark True
+
+        blockWhileCaughtUp
+        enterOnlyBootstrap
+
+    enterOnlyBootstrap :: forall neverTerminates. m neverTerminates
+    enterOnlyBootstrap = do
+        setCaughtUpPersistentMark False
```
nfrisby, line 159: This line should instead be in `enterCaughtUp`, just before the call to `enterOnlyBootstrap`. Locally, this is indistinguishable. However, the node start up calls `enterOnlyBootstrap` exactly when the file is not present (ie when it's _already_ `False`).

cc @bolt12

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Ah, I think this is the explanation: https://github.com/IntersectMBO/ouroboros-consensus/pull/808/files#r1448966065

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
nfrisby: I pushed up what I think is a fix for this.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,330 @@
...
+  -> WrapDurationUntilTooOld m blk
+realDurationUntilTooOld maxCaughtUpAge systemTime =
+    DurationUntilTooOld $ \hfSummary woSlot -> do
+        now <- Clock.systemTimeCurrent systemTime
+        pure $ case woSlot of
+            Slot.Origin  -> toDur now $ Clock.RelativeTime 0
+            Slot.At slot ->
+                let qry = Qry.slotToWallclock slot
+                      -- TODO cache this in the KnownIntersectionState? Or even
+                      -- in the LedgerDB?
```
amesgen, line 265: We now have this comment in several places (here, it should probably not refer to `KnownIntersectionState`, but the basic point about caching still stands), it might make sense to think about factoring this out at some point.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -776,6 +805,7 @@ findIntersectionTop cfgEnv dynEnv intEnv =
                 ourFrag
               , ourHeaderStateHistory
               } = uis
+        let theyAreIdling = intersection == getTipPoint (unTheir theirTip)
```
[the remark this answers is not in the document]
nfrisby: Thanks.

It has two occurrences, one of which is a `NamedFieldPun`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -367,6 +381,10 @@ data KnownIntersectionState blk = KnownIntersectionState {
     -- headers in 'theirFrag', including the anchor.
     --
     -- See the \"Candidate fragment size\" note above.
+  ,
+    theyAreIdling           :: !Bool
+    -- ^ Whether the remote peer's latest message indicates it has subsequent
+    -- headers to send.
```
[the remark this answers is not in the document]
nfrisby: Done, thanks.

For record types where the constructor is almost certainly never used without the selectors, I think `Bool` makes for much more readable code.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1046,15 +1105,19 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
...
-        Stateful $ \kis ->
-            traceException
-          $ let KnownIntersectionState {
+        Stateful $ \kis -> traceException $ do
+            let KnownIntersectionState {
                     mostRecentIntersection
                   , ourFrag
                   , theirFrag
                   , theirHeaderStateHistory
+                  , theyAreIdling
```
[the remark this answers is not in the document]
nfrisby: I removed the `theyWere`... that whole binding is just to avoid the 80 char width limit :'(

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1258,14 +1333,35 @@ checkTime cfgEnv intEnv =
...
+            nonBlockingCall =
+                readLedgerStateHelper
+                    kis
+                    prj
+                    (return $ castM $ do
+                        stopIdling
+                        blockingCall (\x -> x{ theyAreIdling = False })
+                    )
+
+        castM $ if theyAreIdling kis then nonBlockingCall else blockingCall id
```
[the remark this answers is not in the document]
nfrisby: After trying to re-convince myself, I've decided that the approach in this PR is a dead-end.

I've opened https://github.com/IntersectMBO/ouroboros-network/pull/4791 instead. I'm about to push a commit that uses that upstream PR instead of all the `theyAreIdling` logic in this PR. Unfortunately, GHC hangs when I compile that commit locally. Here's hoping the various CI servers handle that gracefully 🤞 (the memory is not growing unbounded, at least).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
nfrisby: > It might also make sense to compare this to a different approach to implement this, namely based on when the server sends `MsgAwaitReply`. Did you already consider this and possible associated (dis)advantages?

I just now pushed that up, since it's far simpler. It relies an upstream PR, though, since the current `ouroboros-network` code does not let pipelined `MsgRequestNext` handlers observe the intermediate `MsgAwaitReply`.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,330 @@
...
+
+    blockUntilCaughtUp :: m ()
+    blockUntilCaughtUp = do
+        varsCandidate <- atomically $ do
+            -- STAGE 1: all ChainSync clients report no subsequent headers
+            idlers        <- getChainSyncIdlers
+            varsCandidate <- getChainSyncCandidates
+            check $
+                               0  < Map.size    varsCandidate
+              && Set.size idlers == Map.size    varsCandidate
```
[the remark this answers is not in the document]
nfrisby: Yep

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1459,6 +1556,9 @@ cross c1 c2 = do
 castM :: Monad m => m (WithEarlyExit m x) -> WithEarlyExit m x
 castM = join . EarlyExit.lift
 
+idlingNow :: StandardHash blk => Point blk -> Their (Tip blk) -> Bool
+idlingNow hdrPoint theirTip = hdrPoint == getTipPoint (unTheir theirTip)
```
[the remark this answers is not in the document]
nfrisby: I'm hoping the `MsgAwaitReply` path works out (see my comment in the other GitHub Thread). But if not, I like your second "Possible change", relying on the previous tip.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -161,14 +163,21 @@ bracketChainSyncClient ::
  -> StrictTVar m (Map peer (StrictTVar m (AnchoredFragment (Header blk))))
     -- ^ The candidate chains, we need the whole map because we
     -- (de)register nodes (@peer@).
+ -> StrictTVar m (Set peer)
+    -- ^ The upstream peers that have not claimed any headers subsequent to
+    -- their candidate, aka those that are /idle/.
```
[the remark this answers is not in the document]
nfrisby: My intuition is to avoid "unnecessary" `TVar`s.

-----

Are you sure we'll need to add more? I'm not recalling why.

> Given that we might want to add even more information here in the future (at least peers that are stuck due to the forecasting horizon?), should this be bundled in some type?

coot:
We're merging IntersectMBO/ouroboros-network#4555.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
nfrisby: In this instance, I was able to use `runWithCachedSummary`.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,330 @@
...
+    blockWhileCaughtUp =
+        atomically getCurrentSelection >>= blockWhileCaughtUpHelper
+
+    blockWhileCaughtUpHelper :: selection -> m ()
+    blockWhileCaughtUpHelper selection = do
+        let computeDuration = case durationUntilTooOld of
+                Nothing -> pure Nothing
+                Just g  -> Just <$> g selection
+        computeDuration >>= \case
+            Nothing          -> atomically retry   -- never unblock
```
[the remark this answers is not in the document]
nfrisby: Thanks, I've written that in for now.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,330 @@
...
+
+    blockWhileCaughtUp :: m ()
+    blockWhileCaughtUp =
+        atomically getCurrentSelection >>= blockWhileCaughtUpHelper
+
+    blockWhileCaughtUpHelper :: selection -> m ()
+    blockWhileCaughtUpHelper selection = do
+        let computeDuration = case durationUntilTooOld of
+                Nothing -> pure Nothing
+                Just g  -> Just <$> g selection
```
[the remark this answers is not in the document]
nfrisby: I used `mapM` and a type signature. Thanks.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,428 @@
...
+
+    enterOnlyBootstrap' :: StdGen -> forall neverTerminates. m neverTerminates
+    enterOnlyBootstrap' g = do
+        atomically $ StrictSTM.writeTVar varLedgerStateJudgement TooOld
+        blockUntilCaughtUp
+        atomically $ StrictSTM.writeTVar varLedgerStateJudgement YoungEnough
+
+        setCaughtUpPersistentMark True
+
+        -- When transitioning from OnlyBootstrap to CaughtUp, the node will
```
nfrisby, line 235: I suddenly recalled discussing this logic with PNSol, but it wasn't in the IER.

What do you remember @amesgen? (I'll ask them either way.)

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -50,7 +50,7 @@ import           Ouroboros.Network.Protocol.LocalStateQuery.Examples
                      (localStateQueryClient)
```
nfrisby, line 50: This PR currently has some unrelated "integration" changes, since it depends on an upstream commit of `ouroboros-network`.

Once `main` is updated, this PR can be collapsed.

amesgen: This has already been updated on `main`, you can rebase on top of it to get eg this change (the PR branch is currently 61 commits behind `main`).

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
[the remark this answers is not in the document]
amesgen: FTR: I don't see this in the latest push.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
amesgen: I also remember that we talked about this in the context of avoiding "flickering" state transitions in case the tip of the honest chain is not very recent, which would induce undue load on the bootstrap peers.

We also explicitly listed the following advantage of the current transition rule for OnlyBootstrap->CaughtUp (see https://github.com/IntersectMBO/ouroboros-consensus/issues/101#issuecomment-1607767722):

> PRO: The trusted remote peer can also tell its downstream peer a "harmless lie" to make them conclude that they are caught up slightly prematurely. This might be especially useful during a disaster scenario (to relieve the trusted notes of load).

This of course only makes sense if the downstream peer actually stays in CaughtUp for at least some time and doesn't immediately switch back.

So I think it makes a lot of sense to have this "flickering-avoidance" mechanism.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
amesgen: > > Given that we might want to add even more information here in the future (at least peers that are stuck due to the forecasting horizon?), should this be bundled in some type?
> 
> Are you sure we'll need to add more? I'm not recalling why.

I think I thought about the LoP (as we have to pause it while being waiting for the forecasting horizon to advance), but I don't think we have to record any information in `TVar`s for that :+1:

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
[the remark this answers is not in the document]
nfrisby: Forgot to push.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
nfrisby: rebased --- I should pull that remaining commit into its own PR

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,330 @@
...
+                 == candidateOverSelection selection candidate
+            check $ all ok candidates
+
+            pure varsCandidate
+
+        -- STAGE 3: the previous check wasn't so slow that the peer set changed
+        --
+        -- If the node is actually done syncing, then this should rarely fail.
+        --
+        -- TODO is this even unnecessary?
```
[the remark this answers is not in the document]
nfrisby: I updated this.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/IOSimQSM/Test/StateMachine/Sequential.hs
```diff
@@ -0,0 +1,193 @@
...
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+
+-----------------------------------------------------------------------------
+-- |
+--
+--
+-- /___This is a superficial variation of the quickcheck-state-machine___/
```
nfrisby, line 11: @coot @jasagredo Have either of you considered this particular obstacle before? Any thoughts to share?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
nfrisby: Yep. I just now pushed up the first draft of the GSM test.

That file has my WIP notes at the bottom of it about what _else_ to test.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PolyKinds #-}
```
nfrisby, line 6: Deriving `Generic1` failed without `PolyKinds`, which is annoying 😞

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+  deriving anyclass (TD.ToExpr)
+
+candidateOverSelection ::
+     Selection
+  -> Candidate
+  -> GSM.CandidateVersusSelection
+candidateOverSelection (Selection b _s) (Candidate b') =
+    -- TODO this ignores CandidateDoesNotIntersect, which seems harmless, but
+    -- I'm not quite sure
+    GSM.WhetherCandidateIsBetter (b < b')
```
amesgen, line 512: I agree that it seems harmless because a candidate branching off before the immutable tip should only ever be temporary until we disconnect. And by Praos, the best chain in the system does *not* fork off before the immutable tip, so it is not feasible for an adversary to delay us from concluding we are caught up this way.

Another way to think about it: The current GSM logic as well as the model here do not make a difference between `CandidateDoesNotIntersect` and `WhetherCandidateIsBetter True`, so sometimes returning `WhetherCandidateIsBetter True` already gives full code coverage. (More directly, one could also take in something of type `ok` instead of the current type of `candidateOverSelection`.)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+
+    notThrashing timestamp =
+        SI.addTime thrashLimit timestamp < clk   -- NB 'boringDur' prevents ==
+
+    -- The /last/ time the node instantaneously visited OnlyBootstrap during
+    -- the 'TimePasses' command.
+    timestamp' timestamp =
+        foldl max fortiethBirthday
+      $ filter (< clk)   -- NB 'boringDur' prevents ==
+      $ iterate (SI.addTime thrashLimit) timestamp
```
amesgen, line 600: I *think* I understand what the purpose of this is, but I was surprised that replacing this definition by
```haskell
    timestamp' timestamp = fortiethBirthday
```
also results in the tests passing just fine.

amesgen (commented):
I think the BSM/GSM q-s-m test very nicely shows that the implementation adheres to the IER :+1:

I also agree that the listed TODO items (as well as more labeling) are good further steps for testing, but not necessarily *all* of them have to happen in this PR.

Most comments are very minor

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -662,3 +663,17 @@ getPeersFromCurrentLedgerAfterSlot kernel slotNo =
...
+     , UpdateLedger blk
+     )
+  => NodeKernel m addrNTN addrNTC blk
+  -> STM m SlotNo
+getImmTipSlotNo kernel = do
+    immutableLedger <-
+      ledgerState <$> ChainDB.getImmutableLedger (getChainDB kernel)
+    case ledgerTipSlot immutableLedger of
+      Origin -> retry
+      NotOrigin tip -> return tip
```
bolt12, line 755: If the node is starting up without any DB, a call to this is going to block until the node starts catching up. Me and @johnalotoski found an issue where the node without local roots nor bootstrap peers, _only_ public roots, would not make progress. I think this happens because when the node is trying to fetch public roots it is going to call `lpGetLatestSlot` in order to decide if it should fetch peers from the ledger or from public configuration. Since this `lpGetLatestSlot` call blocks, the node doesn't make any progress.

So I ask: Is the best way to solve this to return 0 in the `Origin` case instead of retrying?

nfrisby: This function is passed directly as `lpGetLatestSlot` in the interface.

So I think the question is ultimately: should `lpGetLatestSlot` be blocking?

If it should not block, then the type of `lpGetLatestSlot` should be `STM (WithOrigin SlotNo)`. In that case, `getImmTipSlotNo` would end with `pure $ ledgerTipSlot immutableLedger`.

nfrisby: `WithOrigin` is defined in `cardano-base`: https://github.com/IntersectMBO/cardano-base/blob/31185617fa029db5332892203c37cb01ae248c65/cardano-slotting/src/Cardano/Slotting/Slot.hs#L50-L51

bolt12: Could we skip one step and provide an abstraction to `SlotNo` instead? We would have to map a value from `Origin` anyway. Is this not desired? Previously I believe you just returned a `Maybe` value

nfrisby: We agreed on Slack that it should return `WithOrigin SlotNo`, and something closer to the "business logic" can squash that to `SlotNo 0`, with a comment explaining why that is harmless in this case.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
nfrisby: Thank you very much for poking around these parts. Along with your

> However, the tests often just hang indefinitely (for example, for --quickcheck-replay=249156). Is there a simple explanation for this?

comment, I ended up finding a nest of problems around this definition! (The planned task of labeling QC seeds would also have revealed this, FTR.)

- I hard coded a 10^-9 for pico, instead of 10^-12 🤦 
- This `filter` should be a `takeWhile`.

The second one explains the hang. The first one explains why `boringDur` had rendered this dead-code.

There were more problems. I _think_ I've resolved them (at the cost of further complexity in the `precondition` and `generator` functions, sadly), but I'm still seeing failures. After a few hours, I think those are due to an `io-sim` bug 😢. I've pinged on the Slack and will escalate to an Issue once someone confirms (or else explains my misunderstanding in such a way that I can fix my code).

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+thrashLimit = 10   -- seconds
+
+-- | Checks that a 'TimePasses' command does not end exactly when a timeout
+-- could fire
+--
+-- This insulates the test from race conditions that are innocuous in the real
+-- world.
+boringDur :: Model r -> Int -> Bool
+boringDur model dur =
+    boringSelection && boringState
```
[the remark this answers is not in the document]
nfrisby: For the hanging; see https://github.com/IntersectMBO/ouroboros-consensus/pull/808#discussion_r1488641194

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+    secondsToPicoseconds x = x * (1000 * 1000 * 1000)
+
+-----
+
+-- | This is called before reading the implementation's state in order to
+-- ensure that all pending STM transactions have commited before this read.
+--
+-- I'm unsure how many are actually necessary, but ten is both small and also
+-- seems likely to suffice.
+yield10 :: MonadFork m => m ()
```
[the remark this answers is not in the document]
nfrisby: Yes, unfortunately. See the overt Note comment in my latest push.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+yield10 =
+    do yield5; yield5
+  where
+    yield5 = do yield; yield; yield; yield; yield
```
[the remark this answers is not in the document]
nfrisby: Combination of: I never remember the name `replicateM_` and was also overthinking it.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+    ReadMarker ->
+        QSM.Top
+    StartIdling peer -> QSM.Boolean $
+        (peer `Map.member` cands)
+     &&
+        (peer `Set.notMember` idlers)
+    TimePasses dur -> QSM.Boolean $
+        (0 < dur)
+     &&
+        (boringDur model dur)
```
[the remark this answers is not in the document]
nfrisby: I was avoiding that, since the annotations are unused by this test.

However, I realized that they _are_ used when you start manually manipulating repros. So now I've embraced the mini DSL.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+  |
+    ReadJudgment
+  |
+    ReadMarker
+  |
+    StartIdling UpstreamPeer
+    -- ^ INVARIANT existing peer, not idling
+  |
+    TimePasses Int
+    -- ^ tenths of a second
```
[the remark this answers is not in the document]
nfrisby: - I want a bare `Int` so that the counterexamples are more legible and ergonamic.
- I want to use 1 second slot duration, but I want the time passes to be more granular than that.
- But not _too_ granular, for the sake of legibility again.

That's the motivation for the current use of bare "tenths".

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,456 @@
...
+    -- whole network from thrashing between CaughtUp and OnlyBootstrap if
+    -- there's an outage in block production.
+    --
+    -- See also 'antiThunderingHerd'.
+  ,
+    setCaughtUpPersistentMark :: Bool -> m ()
+    -- ^ EG touch/delete the marker file on disk
+  ,
+    varLedgerStateJudgement   :: StrictTVar m LedgerStateJudgement
+    -- ^ EG update the TVar that the Diffusion Layer monitors
```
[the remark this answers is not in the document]
nfrisby: Done. And in fact I actually needed to do that during a debugging session; thanks for suggesting it.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+  deriving stock    (Eq, Ord, Generic, Show)
+  deriving newtype  (Num)
+  deriving anyclass (TD.ToExpr)
+
+data UpstreamPeer = Amara | Bao | Cait | Dhani | Eric
+  deriving stock    (Bounded, Enum, Eq, Ord, Generic, Show)
+  deriving anyclass (TD.ToExpr)
+
+-- | The cumulative growth relative to whatever length the initial selection
+-- was and the slot relative to the start of the test (which is assume to be
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+  ,
+    QSM.transition = transition
+  }
+
+prop_sequential ::
+     LedgerStateJudgement
+  -> QC.Property
+prop_sequential j0 =
+    QSM.forAllCommands
+        (sm (undefined :: Vars IO) j0)   -- NB the specific IO type is unused here
```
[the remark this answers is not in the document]
nfrisby: I made a few, with helpful names. Thanks.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+              ,
+                GSM.varLedgerStateJudgement = varJudgment
+              }
+            gsmEntryPoint = case j0 of
+                TooOld      -> GSM.enterOnlyBootstrap gsm
+                YoungEnough -> GSM.enterCaughtUp      gsm
+
+        (hist, _model, res) <- id
+          $ withAsync gsmEntryPoint
+          $ \_async -> runCommands' (pure sm') cmds
```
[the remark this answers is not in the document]
nfrisby: Thanks for the note. (Again, I actually needed this during a debugging session.)

I went with `Async.poll`, because I don't want to abort the main thread, since it's the one that ends by nicely rendering the repro commands etc.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+semantics vars = \case
+    Disconnect peer -> do
+        atomically $ do
+            modifyTVar varCandidates $ Map.delete peer
+            modifyTVar varIdlers     $ Set.delete peer
+        pure Unit
+    ExtendSelection sdel -> do
+        atomically $ do
+            Selection b s <- readTVar varSelection
+            writeTVar varSelection $! Selection (b + 1) (s + sdel)
```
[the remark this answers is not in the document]
nfrisby: Agreed, but I'm leaving it just to avoid distracting future maintainers.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+-- | TODO restarts (or maybe just killing the GSM thread)
+type Command :: (Type -> Type) -> Type
+data Command r =
+    Disconnect UpstreamPeer
+    -- ^ INVARIANT must be an existing peer
+    --
+    -- Mocks the necessary ChainSync client behavior.
+  |
+    ExtendSelection S
+    -- ^ INVARIANT 'selectionIsBehind'
```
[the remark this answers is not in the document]
nfrisby: Yes, for realism. Also for simplicity.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,698 @@
...
+precondition :: Model Symbolic -> Command Symbolic -> QSM.Logic
+precondition model = \case
+    ExtendSelection _sdel -> QSM.Boolean $
+        selectionIsBehind model
+    Disconnect peer -> QSM.Boolean $
+        peer `Map.member` cands
+    ModifyCandidate peer _bdel -> QSM.Boolean $
+        peer `Map.member` cands
+    NewCandidate peer _bdel -> QSM.Boolean $
+        (not $ peer `Map.member` cands)
```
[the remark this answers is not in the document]
nfrisby: Done, thanks

cabal.project
```diff
@@ -31,3 +31,23 @@ tests: true
 benchmarks: true
 
 import: ./asserts.cabal
+
```
nfrisby, line 34: These need to be removed before this PR is merged. So we're block on `io-sim` and `ouroboros-network` CHaP releases and integration into Consensus.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/IOSimQSM/Test/StateMachine/Sequential.hs
nfrisby: The consensus from the Slack discussion was that `quickcheck-dynamic` is easier to use with `IOSim` since it's pure to begin with. But no one yet has much experience doing so.

The changes in this file compared to the upstream QSM are trivial, so I want to merge this as-is for now. A tech debt issue can reformulate the GSM test away from QSM into `quickcheck-dynamic`, and then delete this module.

(Or we can pushing this generalization upstream, though in the maintainer's shoes I would be hesitant to take on the necessary `io-classes` dependency.)

cabal.project
nfrisby: The `ouroboros-network` release also needs to change the type of `lpGetLatestSlot` to `STM m (WithOrigin SlotNo)`. (I can't find a branch there with a commit making such a change. This orphan does, but it's not on any branch right now: https://github.com/IntersectMBO/ouroboros-network/commit/10eeb922569fa243ad062cac35b93e458f3145dd )

nfrisby: Otherwise, good news: the GSM test is has passed 2.4million tests.

amesgen: > So we're block on `io-sim` and `ouroboros-network` CHaP releases and integration into Consensus.

I think it would even be fine to merge with an `io-sim` s-r-p as the new release won't change the API and is only relevant for tests. IIRC we did something similar with quickcheck-state-machine in the past. (also very minor: `io-sim` is actually released to Hackage, not to CHaP)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM/Model.hs
```diff
@@ -0,0 +1,652 @@
...
+    arbitrary = elements [TooOld, YoungEnough]
+    shrink    = \case
+        TooOld      -> [YoungEnough]
+        YoungEnough -> []
+
+instance QC.Arbitrary MarkerState where
+    arbitrary = elements [Absent, Present]
+    shrink    = \case
+        Absent  -> [Present]
+        Present -> []
```
amesgen, line 563: These could be moved to `Test.Util.Orphans.{ToExpr,Arbitrary}` (or maybe even upstreamed for `LedgerStateJudgement` @bolt12)?

nfrisby: To clarify: `MarkerState ` is not an orphan instance.

But the three `LedgerStateJudgement` instances are. And the `SI.Time` as well.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,410 @@
...
+If we don't yield, one write happens during the ReadMarker and the other
+happens /between/ 'ReadMarker' and 'ReadJudgment'.
+
+It seems most intuitive for the updates to happen "as part of" the
+'ExtendSelection', so I'm favoring yielding after.
+
+And since we're yielding after the command, we should also yield before the
+first command, for consistency.
+
+-}
```
amesgen, line 351: I tried to convert this to a regression test:
```haskell
prop_yield :: QC.Property
prop_yield =
   prop_sequential1 YoungEnough $ QSM.Commands
     [ QSM.Command (NewCandidate Amara (B 1)) Unit []
     , QSM.Command (StartIdling Amara) Unit []
     , QSM.Command (TimePasses 61) Unit []
     , QSM.Command (ExtendSelection (S (-4))) Unit []
     , QSM.Command ReadMarker (ReadThisMarker Absent) []
     ]
```
First, I had to apply this diff to get a useful counterexample output:
```diff
diff --git a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
index 4d34fb8ed..066b2171c 100644
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
@@ -195,7 +195,7 @@ prop_sequential1 j0 cmds = runSimQC $ do
     let executionSM = sm (Just maxBound) vars j0
 
         -- NB the specific IO type is unused here
-        prettySM = sm undefined undefined undefined
+        prettySM = sm undefined undefined j0
 
     let gsm = GSM.realGsmEntryPoints GSM.GsmView {
             GSM.antiThunderingHerd = Nothing
```
After that, the test failed due to a precondition violation (non-boring selection). However, that precondition is actually semantically irrelevant (https://github.com/IntersectMBO/ouroboros-consensus/pull/808/#discussion_r1487577686), so I disabled it.

Then, the tests pass both with the current definition of `yieldSeveralTimes` as well as when replacing it with `pure ()`.

Initially, I thought this note was trying to justify why `yieldSeveralTimes` is necessary to get the tests to pass. However, I now think it is "just" used to get nicer counterexample output. Can you confirm and maybe mention this explicitly?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM/Model.hs
bolt12: `LedgerStateJudgement` already has an arbitrary instance but it is given by `ArbitraryLedgerStateJudgement`, however I am not sure if you can access it

bolt12: It should exist in `ouroboros-network:sim-tests-lib`

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
nfrisby: > Can you confirm and maybe mention this explicitly?

I believe the `yield` should be necessary. It seems so, for me --- which should be true for everyone, since `IOSim` is deterministic.

> I tried to convert this to a regression test:

Great point! I've done so now.

> ... non-boring selection ...

We chatted on the call about this; there was a misunderstanding. _Morever_, though, `boringDur` was too conservative --- I've fixed that, and now this repro is accepted.

> Then, the tests pass both with the current definition of yieldSeveralTimes as well as when replacing it with pure ().

I don't see that---all tests fail if I change that definition. Try again with my latest pushed commit?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -0,0 +1,410 @@
...
+
+----- trivial event accumulator, useful for debugging test failures
```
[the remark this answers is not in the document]
nfrisby: I tried this again, after our discussion. It still feels too much like ceremony. The root cause is that `recordTracerTVar` doesn't take the monotonic time. This gives me two options:

- Keep all the declarations I currently have, but change the implementation to use an underlying `Tracer`. This is a bit "overly complicated" since `recordTracerTVar` returns the read action separately from the tracer. It's two closures that share a hidden variable instead of the simple transparent first-order newtype in my current approach.
- Drop my definitions and replace them with `recordTracerTVar` and a similar combinator that adds in the montonic time. But at this point, `Vars` becomes a misnomer, since the `Tracer` is doing more than "just" wrapping a var: it's also sampling the time.

So---since I'm not otherwise using `Tracer`s---either options seems excessively disruptive.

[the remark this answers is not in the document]
nfrisby: Also, the "few lines" I reuse are lost to imports and wrappers lol

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
```diff
@@ -0,0 +1,465 @@
...
+        -- Randomly add up to 5min.
+        --
+        -- Under the ideal circumstances, nodes have perfectly synchronized
+        -- clocks. However, if there's a block production outage, that means
+        -- /all/ nodes will switch back to the bootstrap peers
+        -- /simultaneously/, incurring a thundering herd of requests on that
+        -- relatively small population. This random change will spread that
+        -- load out.
+        --
+        -- TODO should the Diffusion Layer do this? IE the node /promptly/
```
nfrisby, line 259: A question for the Networking team. @bolt12 @coot @njd42

In this commit, the GSM will (excluding some exceptional scenarios) transition from CaughtUp (aka `YoungEnough`) back to OnlyBootstrap (aka `TooOld`) once its selection exceeds some age that is randomly chosen between 20min and 25min.

- The randomness is to avoid thundering herds "attacking" the bootstrap peers whenever `mainnet` block production does lapse for >20min.

- The IER/our discussions indicate that the GSM should remain in CaughtUp until the selection is at least 20min old --- we don't want the network to "panic" too frequently, and 20min is what PNSol recommended as "robust enough".

But it occurs to me now: when the selection more than 20min old, we should not simply stay in CaughtUp. Any time beyond that is extra exposure while vulnerable (eg perhaps we're eclipsed). So, how about the following subtly different behavior instead?

- After 20min, the GSM _promptly_ transitions from CaughtUp to OnlyBootstrap.
- As per usual, this promptly causes the Diffusion layer to disconnect from all current peers. (Aside: it could keep any that happen to be bootstrap peers.)
- But then, instead of immediately connecting to bootstrap peers, the Diffusion layer intentionally leaves the valency at 0 for a random duration between 0 and 5min.

The key benefit is that: the node is safe while the GSM is in OnlyBootstrap and the valency is 0.

The key cost is that: when there actually is a very long block production outage, then the randomness would manifest as an expectation of 11% of (honest) nodes intentionally having 0 valency at any given time. E(Uniform(0,5)) = 2.5min, and 2.5min / (20min+2.5min) is 10/9 ~ 11%.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
amesgen: > > Then, the tests pass both with the current definition of yieldSeveralTimes as well as when replacing it with pure ().
> 
> I don't see that---all tests fail if I change that definition. Try again with my latest pushed commit?

Ah yeah, I commented out the `ExtendSelection` command for debugging and then forgot to add it back when testing with/without `yieldSeveralTimes` :facepalm: Now get the failures as expected :+1:

---

Minor: The regression test could use `withMaxSuccesses 1`.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
bolt12: You mention the thundering herd problem for a particular edge case where there's a global outage and want a random delay to not over stress the bootstrap peers, and if I understood you correctly you think it doesn't make sense to switch to `TooOld` later due to the possibility of being eclipsed. I say it the eclipse part doesn't matter too much since the outage is global, right ?

But considering your proposal, who's to say that 0-5 min is enough for the other nodes to connect and become caught up? Should this delay be decided based on a statistical likelihood of the time it would take for a node to catch up after a 20+ min outage?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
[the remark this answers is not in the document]
amesgen: > since I'm not otherwise using `Tracer`s

This made me think: Wouldn't it be natural to have a tracer in the GSM logic when it does a state transition? This sounds very useful for debugging a full node. Then, the `EvGsmWrite` events could come from that tracer instead of a watcher.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
nfrisby: > I say it the eclipse part doesn't matter too much since the outage is global, right ?

The mechanism exists because of the global possibility, but the mechanism needs to behave well even if it's not global, since the individual node unfortunately cannot distinguish the two scenarios.

nfrisby: > Should this delay be decided based on a statistical likelihood of the time it would take for a node to catch up after a 20+ min outage?

Yes. I was roughly ballparking that around 5min: 20 min should include around 60 active slots, and I anticipate a node should sync ~60 blocks from the bootstrap peers in much less than 5 minutes. (Of course, these numbers might change if we change the Cardano block dynamics, etc.)

nfrisby: For the main question, though: what do you think about moving the "anti-thundering herd" delay into the Diffusion Layer, instead of baking it into the Consensus Layer's state machine?

bolt12: from what I could understand the "anti-thundering herd" solution seems a lot simpler if implemented on consensus side. On networking side it would require significantly more work, so with this being said I'd say I think it should leave on consensus. I agree that ideally it should be implemented on the networking side but for a temporary feature I believe we should keep it as simple as possible.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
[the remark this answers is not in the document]
nfrisby: I added the tracer and removed the watcher 👍

nfrisby:
This is ready for review, except I need to rename the DONOTMERGE commit, since we opted to go with the s-r-ps for now.

cabal.project
```diff
@@ -31,3 +31,23 @@ tests: true
...
+            ouroboros-network-protocols
+            ouroboros-network-testing
+    --sha256: 1vw8r8csa6lq4330bwgddxa4nl2m1pw1ps5y6l7lw2zw3xdlr0hf
+
+source-repository-package
+    type: git
+    location: https://github.com/input-output-hk/io-sim
+    tag: 85d633d6f9c76ffb678b36396c8fd39b2ff7219e
+    subdir: io-sim
+    --sha256: 10f54xg88wq2w7ylg3n33f1yiswd6w4dam7fvl12raiqj0rk21qm
```
amesgen, line 53: [`io-sim-1.4.1.0`](https://hackage.haskell.org/package/io-sim-1.4.1.0) with the fix is already on Hackage, so it could already be removed in this PR, but it is also no big deal to do it in a later PR before the release.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -668,6 +720,13 @@ mkNodeKernelArgs
       , miniProtocolParameters  = defaultMiniProtocolParameters
       , blockFetchConfiguration = defaultBlockFetchConfiguration
       , keepAliveRng
+      , gsmArgs = GsmNodeKernelArgs {
+          gsmAntiThunderingHerd  = snd (split keepAliveRng)
```
amesgen, line 724: Maybe `keepAliveRng` should be renamed to sound more "generic" and then splitted for different domain specific purposes?

amesgen (approved):
LGTM, also confirming that https://github.com/IntersectMBO/ouroboros-consensus/pull/808#issuecomment-1954913114 looks good :+1:

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
nfrisby: I pushed the split all the way up to a new `llrnGsmAntiThunderingHerdRng` field 👍

cabal.project
nfrisby: Thanks. Leaving the s-r-p, since it's not the only one 👍

PR #811, opened 2023-12-04
Implement QSM's `CanDiff` via `tree-diff`

jasagredo:
# Description

Implement QSM's CanDiff via tree-diff

### DIFF
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/ToExpr.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/ToExpr.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | This module implements QSM's @CanDiff@ typeclass using @tree-diff@'s
+-- @ToExpr@.
+module Test.Util.ToExpr () where
+
+import           Data.TreeDiff as T
+import qualified Test.StateMachine as QSM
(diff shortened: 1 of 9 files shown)


ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/ToExpr.hs
```diff
@@ -0,0 +1,36 @@
...
+-- | This module implements QSM's @CanDiff@ typeclass using @tree-diff@'s
+-- @ToExpr@.
+module Test.Util.ToExpr () where
+
+import           Data.TreeDiff as T
+import qualified Test.StateMachine as QSM
+import           Test.StateMachine.Diffing (CanDiff (..))
+import qualified Test.StateMachine.Types.References as QSM
+
+instance ToExpr x => CanDiff x where
```
dnadales, line 16: Would it make sense to add more comments on this typeclass, and perhaps explain how it is related to `ToExpr`?

jasagredo: Well, this is the implementation. The comments live in https://github.com/stevana/quickcheck-state-machine/blob/00db0b4ebcf4faef434b98b0caaeeba0f17002b9/src/Test/StateMachine/Diffing.hs

jasagredo: I think the module header should point the reader in the right direction to visit that other file
```
-- | This module implements QSM's @CanDiff@ typeclass using @tree-diff@'s
-- @ToExpr@.
```

dnadales: Good point. I missed that.

jorisdral (approved):

PR #812, opened 2023-12-04
Add a pretty-printer for the current state of the node under test

tek:
image


Niols:
What is the meaning of the block with green colour on yellow background? Is it just to indicate that it's on both honest and adversary 1's fragments?

tek:
yeah. not sure if that's better, but we can tweak that as we please

tek:
great questions, but I think we shouldn't bikeshed in this PR. this is not a feature, so we are free to break and improve it without restrictions. but it would be nice to get it merged so we can use it for debugging.

> What happens when there are more than two peers on the same block?

they are sorted by significance (fork > selection > candidate) for the line, and the colors are taken for only two of them, the rest are ignored.

> Is the first coloured block the anchor or the first block of the fragment?

the anchor is not included iirc.

> It would be nice to have consistent colouring.

yeah I was being frivolous with the 256 colors, this will look different in every terminal. go ahead and tweak it as you wish, please.

for the other two, sounds good. but needs work.

Niols (approved):
lítur vel út fyrir mér

PR #814, opened 2023-12-05
Make to build with ghc-9.8

erikd:
# Description

Update codebase to GHC 9.8.1


jasagredo (approved):

jasagredo:
Notice that GHC 9.8.1 on Windows is not able to derive `Data` therefore it will fail to build `Only` and anything that depends on it. https://gitlab.haskell.org/ghc/ghc/-/issues/24392

PR #823, opened 2023-12-06
Bundle arguments to ChainSync.Client and internally refactor the core definition

nfrisby:
While working on PR 808, I feel like the ChainSync client module needs some refactoring.

- It takes way too many arguments, so I've introduced data types that bundle those.
- It has way too many declarations in its `where`-clause, and with the new argument bundles, it's easier than ever before to float those out to their own top-level definitions.
- Moreover, some lines of logic in the message handlers can also be floated out as their own separate logic. This makes the definitions of the higher-level steps easier to read---eg, each one fits on a single "page" now.
- Finally: the whole module's styling is not internally consistent.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -55,7 +55,7 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Ledger.SupportsMempool
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.MiniProtocol.BlockFetch.Server
-import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client as CsClient
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Server
 import           Ouroboros.Consensus.Node.ExitPolicy
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
@@ -69,7 +69,6 @@ import           Ouroboros.Consensus.Util (ShowProxy)
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Orphans ()
 import           Ouroboros.Consensus.Util.ResourceRegistry
-import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import           Ouroboros.Network.Block (Serialised (..), decodePoint,
                      decodeTip, encodePoint, encodeTip)
 import           Ouroboros.Network.BlockFetch
@@ -85,8 +84,7 @@ import           Ouroboros.Network.Mux
 import           Ouroboros.Network.NodeToNode
 import           Ouroboros.Network.NodeToNode.Version (isPipeliningEnabled)
 import           Ouroboros.Network.PeerSelection.PeerMetric.Type
-                     (FetchedMetricsTracer, HeaderMetricsTracer,
-                     ReportPeerMetrics (..))
+                     (FetchedMetricsTracer, ReportPeerMetrics (..))
 import qualified Ouroboros.Network.PeerSelection.PeerSharing as PSTypes
 import           Ouroboros.Network.PeerSharing (PeerSharingController,
                      bracketPeerSharingClient, peerSharingClient,
@@ -131,15 +129,14 @@ data Handlers m addr blk = Handlers {
       hChainSyncClient
         :: ConnectionId addr
         -> IsBigLedgerPeer
-        -> NodeToNodeVersion
-        -> ControlMessageSTM m
-        -> HeaderMetricsTracer m
-        -> StrictTVar m (AnchoredFragment (Header blk))
-        -> ChainSyncClientPipelined (Header blk) (Point blk) (Tip blk) m ChainSyncClientResult
-        -- TODO: we should consider either bundling these context parameters
-        -- into a record, or extending the protocol handler representation
-        -- to support bracket-style initialisation so that we could have the
-        -- closure include these and not need to be explicit about them here.
+        -> CsClient.DynamicEnv m blk
+        -> ChainSyncClientPipelined (Header blk) (Point blk) (Tip blk) m
+             CsClient.ChainSyncClientResult
+        -- TODO: we should reconsider bundling these context parameters into a
+        -- record, perhaps instead extending the protocol handler
+        -- representation to support bracket-style initialisation so that we
+        -- could have the closure include these and not need to be explicit
+        -- about them here.
 
     , hChainSyncServer
         :: ConnectionId addr
@@ -217,15 +214,20 @@ mkHandlers
       NodeKernel {getChainDB, getMempool, getTopLevelConfig, getTracers = tracers}
       computePeers =
     Handlers {
-        hChainSyncClient = \peer _isBigLedgerpeer ->
-          chainSyncClient
-            (pipelineDecisionLowHighMark
-              (chainSyncPipeliningLowMark  miniProtocolParameters)
-              (chainSyncPipeliningHighMark miniProtocolParameters))
-            (contramap (TraceLabelPeer peer) (Node.chainSyncClientTracer tracers))
-            getTopLevelConfig
-            chainSyncFutureCheck
-            (defaultChainDbView getChainDB)
+        hChainSyncClient = \peer _isBigLedgerpeer dynEnv ->
+          CsClient.chainSyncClient
+            CsClient.ConfigEnv {
+                CsClient.cfg                     = getTopLevelConfig
+              , CsClient.someHeaderInFutureCheck = chainSyncFutureCheck
+              , CsClient.chainDbView             =
+                  CsClient.defaultChainDbView getChainDB
+              , CsClient.mkPipelineDecision0     = pipelineDecisionLowHighMark
+                  (chainSyncPipeliningLowMark  miniProtocolParameters)
+                  (chainSyncPipeliningHighMark miniProtocolParameters)
+              , CsClient.tracer                  =
+                  contramap (TraceLabelPeer peer) (Node.chainSyncClientTracer tracers)
+              }
+            dynEnv
       , hChainSyncServer = \peer _version ->
           chainSyncHeadersServer
             (contramap (TraceLabelPeer peer) (Node.chainSyncServerHeaderTracer tracers))
@@ -562,9 +564,9 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout ReportPe
       -- can be used to fetch blocks for that chain.
       bracketSyncWithFetchClient
         (getFetchClientRegistry kernel) them $
-        bracketChainSyncClient
+        CsClient.bracketChainSyncClient
             (contramap (TraceLabelPeer them) (Node.chainSyncClientTracer (getTracers kernel)))
-            (defaultChainDbView (getChainDB kernel))
+            (CsClient.defaultChainDbView (getChainDB kernel))
             (getNodeCandidates kernel)
             them
             version $ \varCandidate -> do
@@ -577,9 +579,15 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout ReportPe
                   (timeLimitsChainSync chainSyncTimeout)
                   channel
                   $ chainSyncClientPeerPipelined
-                  $ hChainSyncClient them isBigLedgerPeer version controlMessageSTM
-                      (TraceLabelPeer them `contramap` reportHeader)
-                      varCandidate
+                  $ hChainSyncClient
+                      them
+                      isBigLedgerPeer
+                      CsClient.DynamicEnv {
+                          CsClient.version
+                        , CsClient.controlMessageSTM
+                        , CsClient.headerMetricsTracer = TraceLabelPeer them `contramap` reportHeader
+                        , CsClient.varCandidate
+                        }
               return (ChainSyncInitiatorResult r, trailing)
 
     aChainSyncServer
--- a/ouroboros-consensus/bench/ChainSync-client-bench/Bench/Consensus/ChainSyncClient/Driver.hs
+++ b/ouroboros-consensus/bench/ChainSync-client-bench/Bench/Consensus/ChainSyncClient/Driver.hs
@@ -1,0 +1,143 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE LambdaCase   #-}
+
+module Bench.Consensus.ChainSyncClient.Driver (mainWith) where
+
+import           Control.Monad (when)
+import qualified Data.Array.IO as UA
+import           Data.Int (Int64)
+import           Data.Time.Clock (diffTimeToPicoseconds)
+import qualified Data.Time.Clock.System as SysTime
+import qualified Data.Time.Clock.TAI as TAI
+import           Data.Word (Word32)
+import           System.Environment (getArgs)
+import           Text.Read (readMaybe)
+
+-- | The argument to the function under test
+--
+-- This is comparable to the "size" argument in QuickCheck.
+newtype X = X Int64
+
+{-# INLINE mainWith #-}
+mainWith :: (Int64 -> IO ()) -> IO ()
+mainWith fut = do
+    xx <- getArgs >>= \case
+        [loStr, hiStr]
+          | Just lo <- readMaybe loStr
+          , Just hi <- readMaybe hiStr
+          -> pure (X lo, X hi)
+        _ -> fail "Pass min and max index as arguments $1 and $2"
+
+    let zz@(!lo, !hi) = mkMeasurementIndexInterval xx
+
+    -- all the extraneous allocation happens up front
+    --
+    -- TODO except the getSystemTime FFI overhead. I haven't find a
+    -- platform-agnostic package for reading the clock into a pre-allocated
+    -- buffer (eg via Storable)
+    varStarts <- newTimeArray_ zz
+    varStops  <- newTimeArray_ zz
+
+    let go !z = do
+            recordTimeArray varStarts z
+            let X i = fst $ coords z in fut i
+            recordTimeArray varStops  z
+            when (z < hi) $ go (incrMeasurementIndex z)
+    go lo
+
+    -- and all the rendering overhead happens after the fact
+    render varStarts varStops zz
+
+-----
+
+-- | 0-based index into the samples for a specific 'X' value
+newtype RepetitionIndex = RepetitionIndex Int64
+
+-- | 0-based index into the cross product @'X' * 'RepetitionIndex'@
+newtype MeasurementIndex = MeasurementIndex Int64
+  deriving (Eq, Ord, UA.Ix)
+
+-- | Reduces variance by the sqrt of this, if the distribution is normal
+samplesPerX :: Int64
+samplesPerX = 1000
+
+-- | The interval of 'MeasurentIndexes' necessary to represent 'samplesPerX'
+-- samples of each 'X' in the given interval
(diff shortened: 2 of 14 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -429,38 +460,47 @@ chainSyncClient
...
-                controlMessageSTM
-                headerMetricsTracer
-                varCandidate = ChainSyncClientPipelined $
-    continueWithState () $ initialise
+chainSyncClient
+    cfgEnv@(ConfigEnv {
+        someHeaderInFutureCheck =
+          InFutureCheck.SomeHeaderInFutureCheck 
+            headerInFutureCheck@(InFutureCheck.HeaderInFutureCheck {
+                InFutureCheck.proxyArrival = Proxy :: Proxy arrival
```
nfrisby: Still binding the type variable, as in the previous commit.

nfrisby:
@dnadales @RenateEilers The last two commits have very ugly diffs.

- The last commit has hundreds of lines of layout changes, due to it being a restyling commit.
- The second-to-last commit moved a lot of code around, since it's floating out existing definitions and folding chunks to create some new definitions.

For the last commit, I recommend using a smart differ, eg

```
$ x=<the restyling commit hash>
$ >before.hs git show ${x}^:ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
$ >after.hs git show $x:ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
$ nix-shell -p difftastic --run 'difft before.hs after.hs'
```

For the second-to-last commit, I recommend having the two files open side-by-side in your editor, so you can manually cross-reference the new definitions versus their previous location in the preceding commit as you come across each of them in the commit's diff.

```
$ x=<the rearranging commit hash>
$ >before.hs git show ${x}^:ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
$ >after.hs git show $x:ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
$ emacs before.hs after.hs &
$ git show $x
```

If that's untenable, then I could split that commit into smaller pieces, one per new definition. But I'd rather avoid that work, if the editor-based method works for you. Please let me know 👍

nfrisby:
I'm leaving it in Draft until we assess the performance concerns that @jasagredo raised about thunks/laziness/etc.

nfrisby:
I pushed up commits that add a small "benchmark" exe that dumps a lot of measurements. I eyeballed them to convince myself that the time behavior hasn't changed in this PR.

Maybe that benchmark should be split into a different PR before merging this one.

I'm still working on convincing myself about space leaks.

nfrisby:
No indication of a space leak in the tip of this PR; 10x 50x and 100x number of messages didn't change the max residency of the benchmark.

```
$ git diff --stat; git diff --stat --cached; git lg -1; cabal build ChainSync-client-bench; for i in 10 50 100; do echo $i; time cabal run ChainSync-client-bench -- $i $i +RTS -s 2>&1 1>/dev/null | grep -e
 'maximum residency' ; done
8cc6d46ee - (HEAD -> nfrisby/ChainSync-refactor, origin/nfrisby/ChainSync-refactor) Update changelogs for ChainSync.Client refactor (2 days ago) <Nicolas Frisby>
Up to date
10
         118,736 bytes maximum residency (10 sample(s))

real	0m9.557s
user	0m9.440s
sys	0m0.267s
50
         118,848 bytes maximum residency (43 sample(s))

real	0m41.005s
user	0m40.943s
sys	0m0.061s
100
         118,768 bytes maximum residency (51 sample(s))

real	1m23.457s
user	1m23.254s
sys	0m0.152s
```

nfrisby:
I'd like to use `NoThunks` to test this, but it's currently blocked on https://github.com/input-output-hk/cardano-base/pull/449 and so on. (I could do without it, but it'd require a good deal of yak-shaving.)

ouroboros-consensus/bench/ChainSync-client-bench/Bench/Consensus/ChainSyncClient/Driver.hs
```diff
@@ -0,0 +1,128 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Bench.Consensus.ChainSyncClient.Driver (mainWith) where
+
+import           Control.Monad (when)
+import qualified Data.Array.IO as UA
```
RenateEilers, line 6: Just curious, why 'UA'?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -577,9 +579,15 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout ReportPe
...
                   channel
                   $ chainSyncClientPeerPipelined
-                  $ hChainSyncClient them isBigLedgerPeer version controlMessageSTM
-                      (TraceLabelPeer them `contramap` reportHeader)
-                      varCandidate
+                  $ hChainSyncClient
+                      them
+                      isBigLedgerPeer
+                      CsClient.DynamicEnv {
+                          CsClient.version
```
jasagredo, line 586: I find it confusing that a qualified field inherits the unqualified value of the same unqualified name.

ouroboros-consensus/bench/ChainSync-client-bench/Bench/Consensus/ChainSyncClient/Driver.hs
jasagredo, line 6: Why `UA`?

ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
```diff
@@ -0,0 +1,302 @@
...
+            Resting rp -> do
+                let (ipoint, blks) = TB.updateToNextNumeral rp
+                writeIORef varState $ Switched blks
+                atomically $ writeTVar varTip $ tipFromHeader $ NE.last blks
+                pure $ RollBack ipoint
+
+    pure ChainDB.Follower {
+        ChainDB.followerClose = pure ()
+      ,
+        ChainDB.followerForward = \case
```
jasagredo, line 287:
```suggestion
        ChainDB.followerClose = pure ()
      
      , ChainDB.followerForward = \case
```

I would prefer commas before names, even if we leave an empty line in between. But this is just an opinion.

ouroboros-consensus/changelog.d/20231206_171848_nick.frisby_ChainSync_refactor.md
```diff
@@ -0,0 +1,26 @@
...
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Bundled up the arguments to the ChainSync client into new record types.
```
jasagredo, line 22: I would mention the type names for quick grepping if needed.

ouroboros-consensus/changelog.d/20231206_171848_nick.frisby_ChainSync_refactor.md
```diff
@@ -0,0 +1,26 @@
...
+
+- Also introduced a SomeHeaderInFutureCheck that binds the existential type
+  variables separately from the actual payload of the HeaderInFutureCheck
+  record type.
```
jasagredo, line 26:
```suggestion
- Also introduced a `SomeHeaderInFutureCheck` that binds the existential type
  variables separately from the actual payload of the `HeaderInFutureCheck`
  record type.
```

Let's take advantage of this being markdown

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -664,3 +664,36 @@ benchmark mempool-bench
...
+    -Wall -Wcompat -Wincomplete-uni-patterns
+    -Wincomplete-record-updates -Wpartial-fields -Widentities
+    -Wredundant-constraints -Wmissing-export-lists -Wunused-packages
+    -Wno-unticked-promoted-constructors -rtsopts -with-rtsopts=-A32m
+
+  -- We use this option to avoid skewed results due to changes in cache-line
+  -- alignment. See
+  -- https://github.com/Bodigrim/tasty-bench#comparison-against-baseline
+  if impl(ghc >=8.6)
+    ghc-options: -fproc-alignment=64
```
jasagredo, line 699: I think it is worth grouping this in a common stanza for benchmarks.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -406,205 +439,427 @@ checkKnownIntersectionInvariants cfg KnownIntersectionState
...
+-- | Arguments determined by configuration
+--
+-- These are available before the diffusion layer is online.
+data ConfigEnv m blk = ConfigEnv {
+    mkPipelineDecision0     :: MkPipelineDecision
+    -- ^ The pipelining decider to use after 'MsgFoundIntersect' arrives
+  , tracer                  :: Tracer m (TraceChainSyncClientEvent blk)
+  , cfg                     :: TopLevelConfig blk
+  , someHeaderInFutureCheck :: InFutureCheck.SomeHeaderInFutureCheck m blk
+  , chainDbView             :: ChainDbView m blk
```
jasagredo, line 475: Despite appreciating the attempt to reduce renaming, once we are in the journey of making a breaking change I would rather prefix the fields with `ce` at least.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -406,205 +439,427 @@ checkKnownIntersectionInvariants cfg KnownIntersectionState
...
+
+-- | Arguments determined dynamically
+--
+-- These are available only after the diffusion layer is online and/or on per
+-- client basis.
+data DynamicEnv m blk = DynamicEnv {
+    version             :: NodeToNodeVersion
+  , controlMessageSTM   :: ControlMessageSTM m
+  , headerMetricsTracer :: HeaderMetricsTracer m
+  , varCandidate        :: StrictTVar m (AnchoredFragment (Header blk))
```
jasagredo, line 486: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -21,16 +22,25 @@
 -- an unexplained thunk in 'KnownIntersectionState' and thus a space leak. See
 -- #1356.
 
-module Ouroboros.Consensus.MiniProtocol.ChainSync.Client (
-    ChainDbView (..)
+-- | This module is intended for qualified import.
```
jasagredo, line 25: I would use this opportunity to link to the relevant external documentation here, like `typed-protocols` and the mini-protocols document as well as the ouroboros-network implementation of the ChainSync miniprotocol.

jasagredo (commented):
I can suggest adding this initial `.hlint.yaml` file:

```yaml
- modules:
  - { name: Ouroboros.Consensus.MiniProtocol.ChainSync.Client
    , importStyle: qualified
    , message: "Exports fields with very common names"
    }

- ignore: {}

- warn: {name: Ouroboros.Consensus.MiniProtocol.ChainSync.Client should be imported qualified}
```

Which will only trigger the import warning. It already fails for:
```
ouroboros-consensus\test\consensus-test\Test\Consensus\MiniProtocol\ChainSync\Client.hs
ouroboros-consensus-diffusion\src\ouroboros-consensus-diffusion\Ouroboros\Consensus\Node\ErrorPolicy.hs
ouroboros-consensus-diffusion\src\ouroboros-consensus-diffusion\Ouroboros\Consensus\Node\RethrowPolicy.hs
ouroboros-consensus-diffusion\src\ouroboros-consensus-diffusion\Ouroboros\Consensus\Node\Tracers.hs
ouroboros-consensus-diffusion\test\consensus-test\Test\Consensus\PeerSimulator\Run.hs
ouroboros-consensus-diffusion\test\consensus-test\Test\Consensus\PeerSimulator\Trace.hs
```

I left some comments but I'm approving because I think they are not severe.

jasagredo (approved):
Approving because what I said above is not severe.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/InFutureCheck.hs
```diff
@@ -136,7 +136,7 @@ realHeaderInFutureCheck skew systemTime =
         unless tooEarly $ do
             now <- systemTimeCurrent systemTime
             let ageNow         = now `diffRelTime` onset
-                syntheticDelay = negate ageNow
+                syntheticDelay = max 0 $ negate ageNow
```
RenateEilers, line 139: Why was `max 0` added? Can `onset` occur after `now`?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
```diff
@@ -823,3 +825,66 @@ instance Serialise ptype => HasBinaryBlockInfo (TestBlockWith ptype) where
       }
 
 instance (Serialise ptype, PayloadSemantics ptype) => SerialiseDiskConstraints (TestBlockWith ptype)
+
+-----
+
+-- | Given a point to a chain of length L, generates a 'SwitchFork' that
+-- switches to the "next" block of length L, where "next" is determined by
```
RenateEilers, line 832: Should this be `chain of lenght L`?

ouroboros-consensus/bench/ChainSync-client-bench/Bench/Consensus/ChainSyncClient/Driver.hs
RenateEilers: I had the same question, and saw later on that it's because of the unboxed arrays in the module (I think). Do we have any guidelines when it comes to naming imports?

jasagredo: We don't really have many conventions. The only ones I can think of from the top of my head are:
- `Ledger.Core` is usually `Core` or `LC`
- `Consensus.Util.IOLike` as `IOLike`
- `Data.X.Strict` as `X` (with `X` being a container say `Map` or `Set`).
- `CBOR....` as `CBOR`
- `...AnchoredFragment` as `AF`
- `...AnchoredSeq` as `AS`
- `Data.ByteString` as `BS`
- `Data.ByteString.Lazy` as `BSL`
- `Data.Text` as `Text` or `T`
- `Test.StateMachine.*` as `QSM`
But we are not very consistent with this I think.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: In the call, we agreed to keep unprefixed names. But we're going to add a hlint rule that required qualified import and limits the names to `CSC` or `CSClient` or `CsClient`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/InFutureCheck.hs
nfrisby: Thanks for asking! It's trickier than I thought: see https://github.com/input-output-hk/io-sim/issues/129

ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
nfrisby: Done

ouroboros-consensus/ouroboros-consensus.cabal
nfrisby: Done

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
nfrisby: That's a reasonable expectation! But because a block chain has the Merkle tree property, a chain and the block at its tip are in many ways indistinguishable. In particular, the length of a block is exactly the length of _the_ chain that ends in that block.

So you'll see this language elsewhere; sometimes "chain" or "block" feels more appropriate depending on context.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: I finally found the HLint manual: https://github.com/ndmitchell/hlint/blob/master/README.md#customizing-the-hints

FYI: we want `explicitOrQualified` instead of just `qualified`.

nfrisby: However, it seems to not work :/ I'll open an Issue to investigate.

nfrisby: Doh! Our `.gitignore` explicitly forbids `.hlint.yaml`. I don't want to shave this yak in this PR.

nfrisby: I opened https://github.com/input-output-hk/ouroboros-consensus/issues/833

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
nfrisby: Yeahs, it's a weird corner case of `-XNamedFieldPuns`.

ouroboros-consensus/changelog.d/20231206_171848_nick.frisby_ChainSync_refactor.md
nfrisby: Done

nfrisby:
Last step before merging will be me locally patching things to build with `NoThunks` using Joris' most recent CHaP release.

nfrisby:
OK, thanks to Joris's help, I have run the benchmark with `NoThunks`-checking enabled and no thunks were found 🖐️ 

I pushed my commit to `nfrisby/ChainSync-refactor-NoThunks`.

Using that commit, I:

- built the benchmark _without_ the flags
- added the flags to my `cabal.project`
- rebuilt the benchmark, confirming that the relevant packages were rebuilt
- ran the benchmark (with arguments 1 1, since it's very slow now)

I saw no `NoThunks` messages, merely the normal benchmark output.

nfrisby:
The last few force pushes were just be supplicating to the .cabal ahd .hs style checkers.

nfrisby:
And this last force-push was because I forgot again about signing my commits.

PR #824, opened 2023-12-08
Updates to alternative LedgerDB API

jorisdral:
# Description

Follow-up to #510

This PR includes changes to the alternative LedgerDB API that are required for integration, which is happening in #815. These changes include:

* LedgerDB API changes:
   * The LedgerDB can now be closed
  * The LedgerDB API now exposes `getToplevelConfig`
  * Forker acquisition is refactored because we now have three variants instead of two, and we can't reuse `StaticEither`
  * Remove the STM function argument from forker acquisition
  * Forker acquisition now requires passing in a resource registry
  * Moved `QueryBatchSize` to the API level, because it's not implementation specific
  * Forkers now have a `forkerRangeReadTablesDefault` function that uses a `QueryBatchSize` that is configured for the whole LedgerDB.
  * `AnnLedgerError` now holds a `Forker`. 
   * Added a few helper functions
* V1 implementation:
  * Export more definitions from the `Snapshots` module
  * Refactored resource acquisition for backing store handles and db changelogs
  * Forkers are no longer implemented as separate handles, but stored directly in the `LedgerDBEnv`. This ensures better concurrency behaviour with respect to closing the LedgerDB and closing forkers. Note: forkers are now implemented similar to ChainDB followers.
  * Expose `LedgerDbSerialiseConstraints` 
  * Small renamings/refactorings

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
@@ -3,17 +3,17 @@
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE DeriveAnyClass           #-}
 {-# LANGUAGE DeriveGeneric            #-}
-{-# LANGUAGE DerivingStrategies       #-}
+{-# LANGUAGE DerivingVia              #-}
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE FlexibleInstances        #-}
 {-# LANGUAGE FunctionalDependencies   #-}
 {-# LANGUAGE GADTs                    #-}
 {-# LANGUAGE LambdaCase               #-}
+{-# LANGUAGE NumericUnderscores       #-}
 {-# LANGUAGE QuantifiedConstraints    #-}
 {-# LANGUAGE RankNTypes               #-}
 {-# LANGUAGE ScopedTypeVariables      #-}
 {-# LANGUAGE StandaloneKindSignatures #-}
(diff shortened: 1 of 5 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -536,21 +619,21 @@ newtype ValidLedgerState l = ValidLedgerState { getValidLedgerState :: l }
...
 --     2. If we are applying rather than reapplying, we might have ledger errors.
-type Ap :: (Type -> Type) -> LedgerStateKind -> Type -> Constraint -> Type
-data Ap m l blk c where
-  ReapplyVal ::           blk -> Ap m l blk ()
-  ApplyVal   ::           blk -> Ap m l blk ( ThrowsLedgerError m l blk )
-  ReapplyRef :: RealPoint blk -> Ap m l blk ( ResolvesBlocks    m   blk )
-  ApplyRef   :: RealPoint blk -> Ap m l blk ( ResolvesBlocks    m   blk
-                                            , ThrowsLedgerError m l blk )
+type Ap :: (Type -> Type) -> (Type -> Type) -> LedgerStateKind -> Type -> Constraint -> Type
+data Ap bm m l blk c where
```
jasagredo, line 623: `bm` stands for? `base monad`?

jorisdral: Yes -- the `MonadBase` class uses `b` as a convention, but then it's less clear in my opinion that it is a monad

PR #817, opened 2023-12-05
[WIP] Add basic support for `PointSchedule` shrinking

Niols:
Obsolete. Extract information and close. Refer to https://github.com/IntersectMBO/ouroboros-consensus/pull/856 for new work on shrinking.

PR #826, opened 2023-12-11
[WIP] Add support for “smart” `PointSchedule` shrinking

Niols:
Obsolete. Extract information and close. Refer to https://github.com/IntersectMBO/ouroboros-consensus/pull/856 for new work on shrinking.

-----

This PR builds on top of #817, which adds “basic” shrinking. In this context, we call _basic_ shrinking the default shrinking that comes with QuickCheck, where shrinkers have type `Input -> [Input]` for some `Input` type. This makes sense in many contexts, but we might want more in ours, and hence we introduce _smart_ shrinking. (I am very unattached to the terms _basic_ and _smart_ so feel free to find better ones.)

_Smart_ shrinking is a generalised form of shrinking where the shrinkers get access to more information than just the input. This makes sense when properties that we want to check require a lot of computation and we would want to use the information gained during this computation. We introduce the pattern “gen-run-check” which fits quite well our use case:

- `gen :: Gen Input` is a standard QuickCheck generator of `Input`s. In our case, those are pairs `(GenesisTest, PointSchedule)`.
- `run :: Input -> Output` contains the expensive part of the computation; it does a lot of things to `Input` until it gathered enough information to produce an `Output`. In our case, this is a `StateView` and `run` is basically our `PeerSimulator`.
- `check :: Output -> Bool` then only has to check a few things on this `Output` to decide whether the test passes.

In this context, we want our shrinkers to not have the usual type `Input -> [Input]` but the richer type `Input -> Output -> [Input]`. In this PR, we therefore:

- Add a QuickCheck helper `forAllGenRunShrinkCheck` providing exactly this functionality. There would be other approaches or more general versions of this helper; see [this private thread on Tweag's Slack](https://moduscreate.slack.com/archives/C02EU5SFBTN/p1701885748346819) for some examples.
- Split the previous `runTest` into two `runGenesisTest` and `runGenesisTest'` functions giving us access to more information.
- Add two Genesis testing-specific helpers `forAllGenesisTest` and `forAllGenesisTest'` that provide a QuickCheck's `forAll`-like interface specific to tests based on generating `(GenesisTest, PointSchedule)`, then running the `PeerSimulator`, then shrinking.
- Rewrite the long range attack and the rollback tests using this new helper.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrink.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrink.hs
@@ -1,6 +1,10 @@
 {-# LANGUAGE NamedFieldPuns #-}
 
-module Test.Consensus.PointSchedule.Shrink (shrinkPointSchedule) where
+module Test.Consensus.PointSchedule.Shrink (
+    shrinkAlertPointSchedule
+  , smartShrinkAlertPointSchedule
+  , smartShrinkPointSchedule
+  ) where
 
 import           Data.Function ((&))
 import           Data.List.NonEmpty (toList)
@@ -8,6 +12,7 @@ import           Data.Maybe (mapMaybe)
 import           Ouroboros.Network.AnchoredFragment (headPoint)
 import           Ouroboros.Network.Block (Point)
 import           Test.Consensus.BlockTree (BlockTree (btTrunk))
+import           Test.Consensus.PeerSimulator.StateView (StateView)
 import           Test.Consensus.PointSchedule
 import           Test.QuickCheck (shrinkList)
 import           Test.Util.TestBlock (TestBlock)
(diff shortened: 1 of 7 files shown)


tek (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrink.hs
```diff
@@ -35,10 +40,26 @@ isReasonablePointSchedule blockTree PointSchedule{ticks} =
...
     & filter (isReasonablePointSchedule blockTree)
+
+-- | Shrink a 'PointSchedule' in a smart way: this function assumes nothing on
+-- the given 'PointSchedule' but it also gets a 'StateView' which allows it to
+-- inspect the run that this 'PointSchedule' triggered. This can help it take
+-- smart decisions of shrunk 'PointSchedule's which we know have big chances to
+-- trigger the same error.
+smartShrinkPointSchedule :: BlockTree TestBlock -> PointSchedule -> StateView -> [PointSchedule]
+smartShrinkPointSchedule _bt _ps _sv =
+  []
```
facundominguez, line 59: It would be nice if these functions were documented to explain the actual shrinking strategies that they implement. Saying that the shrinking is smart or generic doesn't offer much insight currently.

facundominguez (commented):
> smart shrinking

Other names:
* output-aware shrinking
* I/O shrinking

Currently a PointSchedule is offered to the shrinker. I think it would be easier to shrink the input if the separate peer schedules were offered as `[(DiffTime, PointSchedule)]` for the honest peer and `[[(DiffTime, PointSchedule)]]` for the adversaries.

Otherwise, the PR looks good to me.

Niols:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/881 apart from the notion of honest peer alertness. The commits are too old to be reused easily anyways.

PR #827, opened 2023-12-11
Add a test for leashing attacks

facundominguez:
Test with
```
cabal run ouroboros-consensus-diffusion:consensus-test -- -p '/uniform/'
```

There are two new tests "leashing attack" and "time limited leashing attack".

* "leashing attack" randomly drops schedule points from adversarial peers.
* "time limited leashing attack" stops running the point schedule after an amount of time that is computed from the amount of blocks to sync up and the timeouts allowed by Limit of Patience.

This should be enough to detect the two conditions we identified for leashing attacks:

> A leashing attack would be a point schedule meeting either of two conditions:
> 1) it causes the node under test to stop making progress (i.e. the immutable tip doesn't get close to the last genesis window of the honest chain), or
> 2) it causes the node under test to still make progress but it is too slow (in some sense of slow)

The leashing attack tests are expected to fail because we don't use the genesis implementation. We try to run them 1000 times and terminate as soon as a failure is encountered. There might be cases when finding a failing test takes longer. I haven't encountered yet that running the tests 1000 times doesn't produce a failure.


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -134,3 +146,47 @@ prop_uniformSchedule = do
     schedulerConfig = (noTimeoutsSchedulerConfig scheduleConfig) {scEnableGdd = True}
 
     scheduleConfig = defaultPointScheduleConfig
+
+-- | Test that the leashing attacks do not delay the immutable tip
+--
+-- This test is expected to fail because we don't test a genesis implementation
+-- yet.
+prop_leashingAttackNoProgress :: QC.Gen QC.Property
+prop_leashingAttackNoProgress = QC.expectFailure <$> do
```
facundominguez, line 155: `QC.expectFailure` causes the trace of the first failing test to be printed to stdout. The trace of running the scheduler is very long. I'd like to make it shorter, but I don't know what flag to use for that. Please, let me know if you have any advise.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -134,3 +148,114 @@ prop_uniformSchedule = do
...
+      { scEnableGdd = True
+      , scTrace = False
+      }
+
+    scheduleConfig = defaultPointScheduleConfig
+
+    -- | A schedule which doesn't run past the last event of the honest peer
+    genTimeLimitedSchedule :: GenesisTest -> QC.Gen PointSchedule
+    genTimeLimitedSchedule genesisTest =
+      unsafeMapSuchThatJust $ do
```
tek, line 248: this has the effect that the leader schedule will not be regenerated when the point schedule produces `Nothing`. unsure whether that's relevant for this case or undesirable

facundominguez: When does the point schedule produce Nothing?

tek: FTR we decided to replace `Maybe` with `error` since it will never happen.

tek (approved):

PR #836, opened 2023-12-14
Switch the order in which ensureLowerDensity... and unfillRaces are called

facundominguez:
This change simplifies the code because we no longer need to be aware of which the first active slot is after the intersection in the alternative schema.


nfrisby (approved):
LGTM thanks 👍

P.S. - Once your other goal of ensuring both schemas have at least k+1 active slots after the intersection, I suspect `ensureLowerDensityThanHonestSchema` will no longer be necessary.

facundominguez:
Merged.

PR #842, opened 2023-12-15
Introduce a QuickCheck-like Genesis-specific `forAllGenesisTest` helper

Niols:
In particular, this paves the way to adding shrinking in an easy way.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -10,6 +10,7 @@ module Test.Consensus.Genesis.Setup (
     module Test.Consensus.Genesis.Setup.GenChains
   , exceptionCounterexample
   , runGenesisTest
+  , runGenesisTest'
   ) where
 
 import           Control.Exception (AsyncException (ThreadKilled))
@@ -32,16 +33,20 @@ import           Test.Util.Orphans.IOLike ()
 import           Test.Util.TersePrinting (terseFragment)
 import           Test.Util.Tracer (recordingTracerTVar)
 
+-- | See 'runGenesisTest'.
+data RunGenesisTestResult = RunGenesisTestResult {
+  rgtrTrace :: String,
+  rgtrStateView :: StateView
+  }
+
 -- | Runs the given 'GenesisTest' and 'PointSchedule' and evaluates the given
 -- property on the final 'StateView'.
 runGenesisTest ::
-  Testable prop =>
   SchedulerConfig ->
   GenesisTest ->
   PointSchedule ->
-  (StateView -> prop) ->
-  Property
-runGenesisTest schedulerConfig genesisTest schedule makeProperty =
+  RunGenesisTestResult
+runGenesisTest schedulerConfig genesisTest schedule =
   runSimOrThrow $ do
     (recordingTracer, getTrace) <- recordingTracerTVar
     let tracer = if scDebug schedulerConfig then debugTracer else recordingTracer
@@ -57,14 +62,30 @@ runGenesisTest schedulerConfig genesisTest schedule makeProperty =
       "    mustReply = " ++ show (mustReplyTimeout scChainSyncTimeouts)
       ] ++ prettyGenesisTest genesisTest
 
-    finalStateView <- runPointSchedule schedulerConfig genesisTest schedule tracer
-    traceWith tracer (condense finalStateView)
-    trace <- unlines <$> getTrace
+    rgtrStateView <- runPointSchedule schedulerConfig genesisTest schedule tracer
+    traceWith tracer (condense rgtrStateView)
+    rgtrTrace <- unlines <$> getTrace
 
-    pure $ counterexample trace $ makeProperty finalStateView
+    pure $ RunGenesisTestResult {rgtrTrace, rgtrStateView}
   where
     SchedulerConfig {scChainSyncTimeouts} = schedulerConfig
     GenesisTest {gtBlockTree} = genesisTest
+
+-- | Variant of 'runGenesisTest' that also takes a property on the final
+-- 'StateView' and returns a QuickCheck property. The trace is printed in case
+-- of counter-example.
+runGenesisTest' ::
+  Testable prop =>
+  SchedulerConfig ->
+  GenesisTest ->
+  PointSchedule ->
+  (StateView -> prop) ->
+  Property
+runGenesisTest' schedulerConfig genesisTest schedule makeProperty =
+    counterexample rgtrTrace $ makeProperty rgtrStateView
+  where
+    RunGenesisTestResult{rgtrTrace, rgtrStateView} =
+      runGenesisTest schedulerConfig genesisTest schedule
 
 -- | Print counterexamples if the test result contains exceptions.
 exceptionCounterexample :: Testable a => (StateView -> [PeerId] -> a) -> StateView -> Property
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -91,7 +91,7 @@ forAllGenesisTest ::
   Testable prop =>
   Gen (GenesisTest, PointSchedule) ->
   SchedulerConfig ->
-  (StateView -> prop) ->
+  (GenesisTest -> PointSchedule -> StateView -> prop) ->
   Property
 forAllGenesisTest generator schedulerConfig mkProperty =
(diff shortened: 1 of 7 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -550,3 +551,17 @@ stToGen ::
...
+lastHonestBlockPoint :: PointSchedule -> WithOrigin TestBlock
+lastHonestBlockPoint = go Nothing . toList . ticks
+  where
+    go :: Maybe (WithOrigin TestBlock) -> [Tick] -> WithOrigin TestBlock
+    go Nothing [] = error "lastHonestBlockPoint"
+    go (Just bp) [] = bp
+    go _ (Tick{active=Peer HonestPeer (NodeOnline points)} : ticks) =
+      let AdvertisedPoints{block = BlockPoint bp} = points
+       in go (Just bp) ticks
+    go bp (_ : ticks) = go bp ticks
```
facundominguez, line 567: An attempt without explicit recursion:
```Haskell
lastHonestBlockPoint s =
   let bps = mapMaybe honestBlockPoint $ toList (ticks s)
    in case reverse bps of
         [] -> error "lastHonestBlockPoint"
         bp:_ -> bp
  where
    honestBlockPoint Tick{active=Peer HonestPeer (NodeOnline points)} =
        let AdvertisedPoints{block = BlockPoint bp} = points
         in Just bp
    honestBlockPoint _ = Nothing
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -207,10 +207,10 @@ prop_leashingAttackStalling = QC.expectFailure <$> do
 prop_leashingAttackTimeLimited :: QC.Gen QC.Property
 prop_leashingAttackTimeLimited = QC.expectFailure <$> do
   genesisTest <- genChains (QC.choose (1, 4))
-  schedulePoints <- genTimeLimitedSchedule genesisTest
+  schedule <- fromSchedulePoints <$> genTimeLimitedSchedule genesisTest
   pure $
-    runGenesisTest' schedulerConfig genesisTest (fromSchedulePoints schedulePoints) $
-    makeProperty genesisTest schedulePoints
+    runGenesisTest' schedulerConfig genesisTest schedule $
+    theProperty genesisTest schedule
```
facundominguez: I think I need an account of why it is better to have properties using `PointSchedule`. Maybe in the form of a comment about `forAllGenesisTest`.

facundominguez (commented):
Looks good. I think I need a refresher of why we need the flattened `PointSchedule` as interface to `forAllGenesisTest` instead of the more structured `Peer PeerSchedule`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
facundominguez: After reviewing #856, I suppose this depends on the discussion there.

Niols:
Re `PointSchedule` vs. `Peer PeerSchedule`. I think this is an important discussion and probably we should conclude something on that before moving forward with this PR or https://github.com/IntersectMBO/ouroboros-consensus/pull/856.

I don't have a particular opinion on which point schedule is “better”, be it `PointSchedule`, `Peer PeerSchedule` or the `FullTablePointSchedule` of #856. They are all more or less equivalent and it is easy to convert from one to the other. For now, however, it seems to me like `PointSchedule` is the de facto “universal” one: it is the one used in the peer simulator and it is also the one produced by generators. `Peer PeerSchedule` seems to be more of a convenient way of generating point schedules; we often convert from that to `PointSchedule` but not the other way around (and I don't think we even have a function for that, although it would be easy to write?) The name `PointSchedule` itself, matching the test plan, and the location in the modules (in `Test.Consensus.PointSchedule` and not a submodule of that), suggests this universality.

Given that, it seemed natural to me that a generic helper around the peer simulator would also use this kind of point schedule, and therefore consume a generator of `PointSchedule`, then run the peer simulator, and then check a property on the `PointSchedule` itself. This also seemed important when we start shrinking because we will then yield new values of type `PointSchedule`, so a property on `Peer PeerSchedule` would need to convert them back or something. In the specific case of this PR, I don't think the property is easier to write with `Peer PeerSchedule` than with `PointSchedule`, but I assume it might be different for more complex properties on the point schedule.

Again, I don't care which point schedule gets to be _the_ one, but I do think it would be good to keep one to communicate between the components, even if we use others for specific tasks (such as generating). If we decided that it wasn't a big deal to have several kinds of point schedules, I don't see a problem in having two variants of `forAllGenesisTest`, one being the current one using `PointSchedule` everywhere, and one relying on `Peer PeerSchedule` in the generator, shrinker and property and converting to `PointSchedule` only to run the peer simulator. I will happily add such a function to my current PR and rewrite the uniform tests accordingly!

facundominguez:
The structure in `Peers PeerSchedule` has been convenient to write the Genesis properties. Also, it is easy to flatten to PointSchedule for pretty printing or for execution.

> it is also the one produced by generators

Maybe you mean the generators for rollback and timeout tests? 

The multiple representations for point schedules are more code to maintain and understand, and I'm not sure they are a net gain (maybe they are).

I'm less clear about the comparison for shrinking. But I'd defer this analysis until I understand better the need to introduce the full table point schedule.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
```diff
@@ -56,24 +59,57 @@ runTest schedulerConfig genesisTest schedule makeProperty = do
...
-      (ChainSyncException peer e) | Just ThreadKilled <- fromException e -> Right peer
-      exc -> Left exc
+    RunGenesisTestResult{rgtrTrace, rgtrStateView} =
+      runGenesisTest schedulerConfig genesisTest schedule
+
+type ForAllGenesisTest schedule prop =
+  Gen (GenesisTest, schedule) ->
+  SchedulerConfig ->
+  (GenesisTest -> schedule -> StateView -> prop) ->
+  Property
```
Niols, line 91: I would like to add the `Testable prop =>` in there and not require `prop` as an argument of `ForAllGenesisTest`, so as to simplify the type signature of the following functions. I'm sure there is a language extension for this?

Niols:
@facundominguez https://github.com/IntersectMBO/ouroboros-consensus/pull/842/commits/eaf7a15230b83b747a86839d5eb053b5d3af635e generalises `forAllGenesisTest` into two flavours, one working exclusively on `PointSchedule` and one working on `Peers PeerSchedule`. https://github.com/IntersectMBO/ouroboros-consensus/pull/842/commits/28569d4f6b896e27f80bcdb8bee4ea0ff6f7e8b7 then partially reverts https://github.com/IntersectMBO/ouroboros-consensus/commit/cd82d765018cdb07de88cd76271d5afe6b62d6f2 to bring back the property written exclusively in term of `Peers PeerSchedule`. I think it looks rather clean; what do you think?

I also think it would be best to try and rely on the least amount of point schedule types possible, or at least to limit their scope. It looks like new point schedules are only written in term of `Peers PeerSchedule` so maybe this should become the main point schedule type, with the current `PointSchedule` being only used in the peer simulator (although we could even imagine rewriting this). Still, with the latest changes, I think we can decide that this is out of scope of this PR and get it merged?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: With the last changes, this function has been made obsolete. I also ended up needing a more generic version in another piece of work which I think should be preferred to this `lastHonestBlockPoint`. See https://github.com/IntersectMBO/ouroboros-consensus/commit/61b84ef942d5357216ce828a0d078b4868ffffd1.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
facundominguez: Maybe inlining the type synonym at the few places where it is used would offer the less friction.

facundominguez (approved):
LGTM

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
Niols: You mean 072962b58458436a0b6662a1c04131ad227ed578?

PR #840, opened 2023-12-15
First tests for the genesis implementation

facundominguez:
This PR contributes three tests that rehearse the genesis implementation in different ways.

Please, note that this PR targets the branch of #536. 

All the tests check that the immutable tip of the node under test ends up in the honest chain, at a block that is recent enough from the tip of the honest chain. Recent enough is defined as within s+d+1 slots of the tip of the honest chain, where s is the genesis window and d is the maximum Praos delay.

The tests differ in the inputs that they rehearse:
* "serve adversarial branches": tests schedules that provide as much time as necessary for the node under test to reach the expected immutable tip.
* "stalling leashing attack": tests schedules where adversarial peers sometimes omit sending random blocks or headers, which would cause the node under test to delay 
* "timed leashing attack": tests schedules where the schedules are executed up to a given time bound by which the node under test is expected to reach a recent immutable tip.

All tests are expected to fail at the moment since they depend on having a working genesis implementation.

Contents of the PR:
* The new genesis tests in commits https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/6ab827720f1aeba1a7915a3bd4580658c7a06828 and https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/01641f77c04807866d7bfea556a13f3db1b0065c
* Improvements to the peer simulator and the schedule generator in commits https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/065614cbebbdfa9bed158d590c56e0902b8eee58, https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/343695cde9eb7f7119042ce0dbf05a97400e8ec5, https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/e7ef3833c299b91415a4694fc61e7334ef3b036a, and https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/3714105301d6b485149f4325689041d7672b79fc
* Improvements to the schema generator in https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/441092ab7ebdfe58a76ab014e71612d18aa585e7 and https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/692c099597e5dd0ea320210cbf09824b163f7147
* Improvements to peer simulator tests in https://github.com/IntersectMBO/ouroboros-consensus/pull/840/commits/e3c0f5e7060752f6147a95896486b57e3475d510

PR #843, opened 2023-12-15
Express tests in terms of schedules and not the block tree

facundominguez:
This PR allows to prune a suffix of the schedule of the honest peer and still get a valid input.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
@@ -24,12 +24,12 @@ import           Ouroboros.Network.Block (blockNo, unBlockNo)
 import           Test.Consensus.BlockTree (BlockTree (..))
 import           Test.Consensus.Genesis.Setup
 import           Test.Consensus.Genesis.Setup.Classifiers
-import           Test.Consensus.PeerSimulator.Run (noTimeoutsSchedulerConfig,
-                     scTraceState)
+import           Test.Consensus.PeerSimulator.Run (SchedulerConfig (scTrace),
+                     noTimeoutsSchedulerConfig, scTraceState)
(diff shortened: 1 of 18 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -105,23 +108,27 @@ makeProperty genesisTest advCount StateView {svSelectedChain} killed =
       [] -> "No peers were killed"
       peers -> "Some peers were killed: " ++ intercalate ", " (condense <$> peers)
 
-    honestTipSlot = AF.headSlot (btTrunk gtBlockTree)
+    honestTipSlot = At $ tbSlot $ snd $ last $ mapMaybe fromBlockPoint $ value $ honest schedule
```
Niols, line 111: This is a very elaborated line; makes me question whether we have the right representations or helpers. Maybe we'll see things emerge from such uses!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -55,11 +56,11 @@ tests =
 
 makeProperty ::
   GenesisTest ->
-  Int ->
+  Peers PeerSchedule ->
   StateView ->
   [PeerId] ->
   Property
```
Niols, line 62: Having the property rely on the `Peers PeerSchedule` instead of a `PointSchedule` will be incompatible with https://github.com/IntersectMBO/ouroboros-consensus/pull/842. I'm sure we can find a way to make https://github.com/IntersectMBO/ouroboros-consensus/pull/842 allow that though.

Niols (approved):
LGTM. I'll try to give the same treatment to the long range attack test!

facundominguez:
Squashed in #840

PR #848, opened 2023-12-18
Ensure lower density in adversarial schemas

facundominguez:
This commit ensures lower density in the first stability window after the intersection, and in any other subwindow that extends the first race window after the intersection.

Ensuring lower density of the subwindows is necessary to avoid chains like
```
k: 3
s: 9
H: 0111100
A: 1110011
```
where the honest chain wins the race to the k+1st block, might win the density comparison if the chains are extended to include a full stability window after the intersection, but loses the density comparison if the chains aren't extended.                                                                                                                                                                

For the sake of shrinking test inputs, we also prevent the above scenario to occur in any intersection, not just the intersections near the end of the chains.

### DIFF
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -678,11 +678,11 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
         -- intersection.
         stablePrefixWindowsContaining w =
           let
-            stableSlotsAfterIntersection =
-              min s (C.getCount (C.lengthMV mv) - C.getCount (C.windowStart w))
-            n = stableSlotsAfterIntersection - C.getCount (C.windowSize w)
-          in
-            take n $ drop 1 $ iterate increaseSizeW w
+            start = C.windowStart w
+            size  = C.getCount (C.windowSize w)
+            end   = s `min` C.getCount (C.lengthMV mv)
+           in
+            [ C.UnsafeContains start (C.Count size') | size' <- [ size+1 .. end ] ]
 
         -- Updates mv to ensure the density of the adversarial schema is lower
         -- than the density of the honest schema in @increaseSizeW w@.
@@ -707,10 +707,6 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
                 pure ac'
 
           return (hc', ac'')
-
-    -- | Increase the size of a window by one slot
-    increaseSizeW (C.UnsafeContains start size) =
-        C.UnsafeContains start (size C.+ 1)
 
     -- | Ensure the density of the adversarial schema is less than the density
     -- of the honest schema in the given window.
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -287,7 +287,7 @@ checkAdversarialChain recipe adv = do
           -- window after the intersection
           hwSum = Vector.toList $ Vector.drop (C.getCount $ C.windowSize w0') $ Vector.take (s + 1) hSum
           awSum = Vector.toList $ Vector.drop (C.getCount $ C.windowSize w0') $ Vector.take (s + 1) aSum
-        case filter (\(x, y) -> x < y) (zip hwSum awSum) of
+        case filter (\(x, y) -> x <= y) (zip hwSum awSum) of
           []         -> pure ()
           ((x, y):_) -> Exn.throwError $ BadDensity x y
 
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -265,7 +265,7 @@ checkAdversarialChain recipe adv = do
           winHAfterIntersection =
               C.UnsafeContains
                   (fromJust $ C.toWindow winH $ C.windowStart winA)
-                  (C.Count $ C.getCount (C.windowSize winA))
+                  (C.windowSize winA)
           -- honest schema after the intersection
           vHAfterIntersection = C.sliceV winHAfterIntersection vH
           iterH :: RI.Race adv
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -28,7 +28,7 @@ module Test.Ouroboros.Consensus.ChainGenerator.Adversarial (
   ) where
 
 import           Control.Applicative ((<|>))
-import           Control.Monad (void, when)
+import           Control.Monad (foldM, void, when)
 import qualified Control.Monad.Except as Exn
 import           Control.Monad.ST (ST)
 import           Data.Maybe (fromJust)
@@ -654,36 +654,50 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
           -- A window after the intersection as short as the shortest of the
           -- stability window or the first race to the k+1st block.
           w0' = C.UnsafeContains (C.windowStart w0) (min (C.Count s) (C.windowSize w0))
-          hCount = BV.countActivesInV S.notInverted (C.sliceV w0' vA)
+          hCount = C.toVar $ BV.countActivesInV S.notInverted (C.sliceV w0' vA)
 
         aCount <- ensureLowerDensityInWindow w0' g mv hCount
 
-        go w0' (C.toVar hCount) aCount
+        void $ foldM
+          updateDensityOfMv
+          (hCount, aCount)
+          (stablePrefixWindowsContaining w0')
+
       where
-        -- Ensure low densities in all windows containing @w@ until the first
-        -- stability window after the intersection.
+        -- Yields all windows of the adversarial schema with proper prefix @w@
+        -- that are prefixes of the first stability window after the
+        -- intersection.
+        stablePrefixWindowsContaining w =
+          let
+            stableSlotsAfterIntersection =
+              min s (C.getCount (C.lengthMV mv) - C.getCount (C.windowStart w))
+            n = stableSlotsAfterIntersection - C.getCount (C.windowSize w)
+          in
+            take n $ drop 1 $ iterate increaseSizeW w
+
+        -- Updates mv to ensure the density of the adversarial schema is lower
+        -- than the density of the honest schema in @increaseSizeW w@.
         --
-        -- @hc@ is the number of active slots in the honest schema in @w@
+        -- @hc@ is the number of active slots in the honest schema in @w@ minus
+        -- the last slot
+        --
         -- @ac@ is the number of active slots in the adversarial schema in @w@
-        go w hc ac =
-          when (C.windowSize w C.+ 1 <= C.Count s
-                && C.windowLast w C.+ 1 < C.toIndex (C.toVar $ C.lengthMV mv)) $ do
-              let w' = increaseSizeW w
+        -- minus the last slot
+        updateDensityOfMv (hc, ac) w = do
+          sA <- BV.testMV S.notInverted mv (C.windowLast w)
 
-              sA <- BV.testMV S.notInverted mv (C.windowLast w')
+          let
+            ac' = if sA then ac C.+ 1 else ac
+            sH = BV.testV S.notInverted vA (C.windowLast w)
+            hc' = if sH then hc C.+ 1 else hc
 
-              let
-                ac' = if sA then ac C.+ 1 else ac
-                sH = BV.testV S.notInverted vA (C.windowLast w')
-                hc' = if sH then hc C.+ 1 else hc
+          ac'' <-
+              if ac' >= hc' then
+                ensureLowerDensityInWindow w g mv hc'
+              else
+                pure ac'
 
-              ac'' <-
-                  if ac' >= hc' then
-                    ensureLowerDensityInWindow w' g mv hc'
-                  else
-                    pure ac'
-
-              go w' hc' ac''
+          return (hc', ac'')
 
     -- | Increase the size of a window by one slot
     increaseSizeW (C.UnsafeContains start size) =
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
@@ -277,7 +277,7 @@ checkAdversarialChain recipe adv = do
         C.SomeWindow pw0 w0 <- let RI.Race x = iterH in pure x
 
         let
-          w0' = C.UnsafeContains (C.windowStart w0) (min (C.Count s) (C.windowSize w0))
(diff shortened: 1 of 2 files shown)


Niols:
Note: We had a doubt today as to whether this PR had been integrated into the December delivery (#840) and the answer is that it has not.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -600,28 +627,84 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
...
+        aCount <- ensureLowerDensityInWindow w0' g mv hCount
+
+        go w0' (C.toVar hCount) aCount
+      where
+        -- Ensure low densities in all windows containing @w@ until the first
+        -- stability window after the intersection.
+        --
+        -- @hc@ is the number of active slots in the honest schema in @w@
+        -- @ac@ is the number of active slots in the adversarial schema in @w@
+        go w hc ac =
```
nfrisby, line 668: I think the `go` loop is fine, but it might be nicer to further separate the generation of the prefixes into something like `prefixesFromTo`, which could be used like this:

```
let
    prefix = the race window
    stab   = the stability window (or the end of the mv)
(\snoc -> foldM snoc (hCount, aCount) (prefixesFromTo prefix stab)) $ \(hc, ac) -> do
    ...
    pure (hc', ac')
```

I'm guessing `prefixesFromTo` could ultimately be implemented a list comprehension.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
+    -- | Check that the density of the adversarial schema is less than the
+    -- density of the honest schema in the first stability window after the
+    -- intersection and in any subwindow that contains the first race to the
+    -- k+1st block.
+    --
+    -- See description of @ensureLowerDensityInWindows@
+    checkDensity = do
+        let
+          -- window of the honest schema after the intersection
+          winHAfterIntersection =
```
nfrisby, line 265: If I understand & recall correctly, I think this could be written `winHAfterIntersection = C.joinWin winH carWin`, if the argument to `checkAdversarialChain` were a _checked_ recipe. (Or perhaps `checkAdversarialChain` could simply start by checking the given recipe?)

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
+          vvH = C.getVector vHAfterIntersection
+          vvA = C.getVector vA
+          -- cumulative sums of active slots per slot after the intersection
+          hSum = Vector.scanl (+) 0 (Vector.map (\x -> if S.test S.notInverted x then 1 else 0) vvH)
+          aSum = Vector.scanl (+) 0 (Vector.map (\x -> if S.test S.notInverted x then 1 else 0) vvA)
+          -- cumulative sums of active slots per slot until the first stability
+          -- window after the intersection
+          hwSum = Vector.toList $ Vector.drop (C.getCount $ C.windowSize w0') $ Vector.take (s + 1) hSum
+          awSum = Vector.toList $ Vector.drop (C.getCount $ C.windowSize w0') $ Vector.take (s + 1) aSum
+        case filter (\(x, y) -> x < y) (zip hwSum awSum) of
```
nfrisby, line 290: Maybe add an index to `BadDensity` that indicates which prefix was the problematic one? And then use `zip [0..] $ zip hwSum awSum` eg.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
-                aScgWin = C.UnsafeContains (C.Count 0) (C.Count s)
-                -- number of active slots in the first stability window after
-                -- the intersection in the honest schema
-                hCount = C.getCount $ BV.countActivesInV S.notInverted (C.sliceV hScgWin vH)
-                aCount = C.getCount $ BV.countActivesInV S.notInverted (C.sliceV aScgWin vA)
-            when (aCount >= hCount) $ Exn.throwError $ BadDensity hCount aCount
+
+    -- | Check that the density of the adversarial schema is less than the
+    -- density of the honest schema in the first stability window after the
+    -- intersection and in any subwindow that contains the first race to the
```
nfrisby, line 258: What do you think about using the term "prefix (of the stability window)" instead of "subwindow"?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
+          iterH =
+              maybe (RI.initConservative vHAfterIntersection) id
+            $ RI.init (Kcp k) vHAfterIntersection
+
+        -- first race window after the intersection
+        C.SomeWindow _ w0 <- let RI.Race x = iterH in pure x
+
+        let
+          w0' = C.UnsafeContains (C.windowStart w0) (min (C.Count s) (C.windowSize w0))
+          vvH = C.getVector vHAfterIntersection
```
nfrisby, line 281: My initial instinct is that functions on `Vector`s should be defined in the `Test.Ouroboros.Consensus.ChainGenerator.BitVector` module, instead of via `C.getVector`.

Although there are several windows in play here, so it might not be trivial to choose a helpful type for such a thing. But at least what you have here could be transplanted there quite directly, and given strongly-typed domains and perhaps loosely-typed codomains.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
+          -- window of the honest schema after the intersection
+          winHAfterIntersection =
+              C.UnsafeContains
+                  (fromJust $ C.toWindow winH $ C.windowStart winA)
+                  (C.Count $ C.getCount (C.windowSize winA))
+          -- honest schema after the intersection
+          vHAfterIntersection = C.sliceV winHAfterIntersection vH
+          iterH :: RI.Race adv
+          iterH =
+              maybe (RI.initConservative vHAfterIntersection) id
```
nfrisby, line 273: We definitely want to include the "conservative" windows, ie the ones that assume the honest schema after the end of time is full of 1s?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -249,20 +252,44 @@ checkAdversarialChain recipe adv = do
...
+            $ RI.init (Kcp k) vHAfterIntersection
+
+        -- first race window after the intersection
+        C.SomeWindow _ w0 <- let RI.Race x = iterH in pure x
+
+        let
+          w0' = C.UnsafeContains (C.windowStart w0) (min (C.Count s) (C.windowSize w0))
```
nfrisby, line 280: This occurs below, too. Seems like a reasonable definition to factor out into the `.Counting` module. Something like `truncateWin (C.UnsafeContains start len) x = C.UnsafeContains start (min len x)`?

nfrisby (commented):
The logic looks convincing to me 👍

There are several windows in play here 😵 :

- the suffix of the honest schema's window that starts after the intersection
- the adversarial schema's window, which starts after the intersection (a la `C.lengthMV mv`)
    - `carWin` relates this to the previous bullet
- the first stability window after the intersection (a la `C.Count s`)
- the first Praos race window
- every extension of the first race window that fits inside the first stability window

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Adopted in 41db9fe.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -484,7 +511,7 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
             maybe (RI.initConservative vA) id
           $ RI.init kcp vA
 
-    ensureLowerDensityThanHonestSchema g mv
+    ensureLowerDensityInWindows iterH g mv
 
     unfillRaces kPlus1st (C.Count 0) UnknownYS iterH g mv
```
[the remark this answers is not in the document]
facundominguez: Adopted.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: If the purpose of the index is to save the trouble of counting active slots when looking at a printing of the schemas, I don't think it will be effective at that.

I added the prefix window itself in e16a34e.

facundominguez (commented):
I'll address the remaining comments in a bit.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Currently
```haskell
winHAfterIntersection :: C.Contains SlotE hon ...
```
because the code is using it to slice a `hon` vector.

On the other hand
```haskell
C.joinWin winH carWin :: C.Contains SlotE base adv
```
which doesn't fit.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -600,28 +634,98 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
...
+        void $ foldM
+          updateDensityOfMv
+          (hCount, aCount)
+          (stablePrefixWindowsContaining w0')
+
+      where
+        -- Yields all windows of the adversarial schema with proper prefix @w@
+        -- that are prefixes of the first stability window after the
+        -- intersection.
+        stablePrefixWindowsContaining w =
```
nfrisby, line 677: Is this equivalent? Seems more direct in my opinion, if so.

```
stablePrefixWindowsContaining w =
    let start = C.windowStart w
        size  = C.getCount (C.windowSize w)
        end   = s `min` C.getCount (C.lengthMV mv)
    in
    [ C.UnsafeContains start (C.Count size') | size' <- [ size+1 .. end ] ]
```

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
nfrisby: Very nice, yeah 👍

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
nfrisby: Hmm, I see, thanks.

Actually, is it just `winHAfterIntersection = carWin`?

```
winA :: C.Contains SlotE base adv
winH :: C.Contains SlotE base hon
carWin :: C.Contains SlotE hon adv

winA == C.joinWin winH carWin
```

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -484,8 +515,11 @@ uniformAdversarialChain mbAsc recipe g0 = wrap $ C.createV $ do
             maybe (RI.initConservative vA) id
           $ RI.init kcp vA
 
-    ensureLowerDensityThanHonestSchema g mv
+    ensureLowerDensityInWindows iterH g mv
```
nfrisby, line 518: Now that your other function declares `iterH = ... vH`, I'm wondering if the race iterator here (declared as `iterH` just a few lines up) should be named `iterA` instead, since it's iterating over the race windows based on the adversarial schema. What do you think?

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
nfrisby: Yeah, interpreting this as a postcondition of `uniformAdversarialChain` does suffice to justify it.

---

> Check that the density of the adversarial schema is less than the
> density of the honest schema in the first stability window after the
> intersection and in any prefix that contains the first race to the
> k+1st block.

The use of `initConservative` here means we're doing this check even if there aren't `k+1` _actual_ active slots after the intersection in the schema. (And, indirectly, it also means that `uniformAdversarialChain` is.)

I suppose that's fine, merely hard to think about (as considering slots past the end of the schemas is always confusing :/). Perhaps it's somewhat "wasteful" when `RI.init` returns `Nothing`, since I _think_ in that case this check simplifies down to requiring that the honest schema has at least as many active slots after the intersection as the adversarial schema does. But "wasteful" is lower priority than "clear", and using the same predicate regardless of the ctor returned by `RI.init` is probably clearer.

Maybe it's worth a comment, if you agree with my conjecture about the simpler interpretation?

nfrisby (approved):
Approved.

I either Resolved or replied to every thread 👍. But none of the unResolved are crucial.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Looks good to me. Adopted in 4798d0c.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: > in that case this check simplifies down to requiring that the honest schema has at least as many active slots after the intersection as the adversarial schema does.

I think in that case the check simplifies to ensuring that the adversarial chain has at most k blocks after the intersection.

I'm considering what it would take to relax the generation and the check.

facundominguez: hm, actually, if we merge #888, this is not an issue we need to fix as chains will always have @k+1@ blocks after the intersection.

I'll add a comment on this aspect in #888 instead.

facundominguez:
Merged to `genesis/milestone8` as 6898f0a.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Addressed in #888 in 55f7e56 by removing the calls to `initConservative`.

PR #849, opened 2023-12-18
Add hlint to CI and an initial .hlint.yaml file

jasagredo:
# Description
Add HLint to CI, and add an initial .hlint.yaml file.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,10 @@
+- modules:
+  - { name: Ouroboros.Consensus.MiniProtocol.ChainSync.Client
+    , importStyle: explicitOrQualified
+    , as: [CSC, CsClient, CSClient]
+    , message: "Exports fields with very common names"
+    }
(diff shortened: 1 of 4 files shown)


.hlint.yaml
```diff
@@ -0,0 +1,10 @@
+- modules:
+  - { name: Ouroboros.Consensus.MiniProtocol.ChainSync.Client
+    , importStyle: explicitOrQualified
+    , as: [CSC, CsClient, CSClient]
+    , message: "Exports fields with very common names"
+    }
+
+- ignore: {}
+
+- warn: {name: Ouroboros.Consensus.MiniProtocol.ChainSync.Client should be imported qualified or with an explicit import list}
```
jasagredo, line 10: It is a pain in the neck that we cannot use globs, I would have written `* should be imported *` and then we would get all the warnings of this kind always, but it can't be done

PR #858, opened 2023-12-29
Implement LDB V2

jasagredo:
# Description

this PR might resolve.


jasagredo:
This organization looks quite nice I think

mods

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -0,0 +1,479 @@
...
+
+{-------------------------------------------------------------------------------
+  LedgerTablesHandles
+-------------------------------------------------------------------------------}
+
+data LedgerTablesHandle m l = IOLike m => LedgerTablesHandle {
+    close       :: !(m ())
+  , duplicate   :: !(m (LedgerTablesHandle m l))
+  , read        :: !(LedgerTables l KeysMK -> m (LedgerTables l ValuesMK))
+  , readAll     :: !(m (LedgerTables l ValuesMK))
```
jorisdral, line 77: Range lookups for LSM-tree style handles would require a range of keys as an argument (see [the docs](https://input-output-hk.github.io/lsm-tree/lsm-tree/Database-LSMTree-Normal.html#t:Range)). Reading all keys at once would be possible, but that means we would be reading the full UTxO into memory which would greatly increase memory use, so probably we should use a range of keys here too

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -0,0 +1,479 @@
...
+  , write       :: !(LedgerTables l DiffMK -> m ())
+  , writeToDisk :: !(SomeHasFS m -> FsPath -> m ())
```
jorisdral, line 79: Make sure that this is compatible with LSM-tree style snapshots (see [the docs](https://input-output-hk.github.io/lsm-tree/lsm-tree/Database-LSMTree-Normal.html#v:snapshot)). LSM-tree style snapshots don't allow snapshotting into arbitrary directories. The snapshot directory is fixed, but names for the snapshots can be picked.

So, maybe it should be something like `String -> m ()`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,447 @@
...
+  => ForkerEnv m l blk
+  -> RangeQuery l
+  -> m (LedgerTables l ValuesMK)
+implForkerRangeReadTables env rq0 = do
+    ldb <- readTVarIO $ foeLedgerSeq env
+    case rqPrev rq0 of
+      Nothing -> readAll (tables $ currentHandle ldb)
+      Just (LedgerTables (KeysMK ks)) -> do
+        LedgerTables (ValuesMK m) <- readAll (tables $ currentHandle ldb)
+        pure $ LedgerTables $ ValuesMK (maybe m ( snd . flip Map.split m) (Set.lookupMax ks))
```
jorisdral, line 301: `rqCount` is unused

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -0,0 +1,479 @@
...
+  , tablesSize  :: !(m Int)
```
jorisdral, line 80: This should return `Maybe Int`, because LSM-trees can not compute this (cheaply)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,447 @@
...
+  => LedgerDBHandle m l blk
+  -> ForkerKey
+  -> m ()
+implForkerClose (LDBHandle varState) forkerKey =
+    atomically $ readTVar varState >>= \case
+      LedgerDBClosed       -> pure ()
+      LedgerDBOpen ldbEnv -> do
+        modifyTVar
+            (ldbForkers ldbEnv)
+            (Map.updateWithKey (\_ _ -> Nothing) forkerKey)
```
jorisdral, line 279: This should call `close` on handles that aren't part of the current ledger seq

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,447 @@
...
+
+-- | Close all open block and header 'Follower's.
+closeAllForkers ::
+     MonadSTM m
+  => LedgerDBEnv m l blk
+  -> m ()
+closeAllForkers ldbEnv =
+    atomically $ writeTVar forkersVar Map.empty
+  where
+    forkersVar = ldbForkers ldbEnv
```
jorisdral, line 447: This should call `close` on handles that aren't part of the current ledger seq

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -0,0 +1,367 @@
...
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+-- | Common logic and types for LedgerDB Snapshots.
+
+module Ouroboros.Consensus.Storage.LedgerDB.Snapshots (
+    -- * Disk policy
+    DiskPolicy (..)
+  , defaultDiskPolicy
```
jorisdral, line 15: IMO, this should be part of the `API` module, and the rest is impl-speicifc

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB.hs
```diff
@@ -1,12 +1,94 @@
...
+{-# LANGUAGE GADTs            #-}
+{-# LANGUAGE RankNTypes       #-}
+{-# LANGUAGE TypeOperators    #-}
+
 module Ouroboros.Consensus.Storage.LedgerDB (
     -- * API
     module Ouroboros.Consensus.Storage.LedgerDB.API
-  , module Ouroboros.Consensus.Storage.LedgerDB.API.DiskPolicy
+  , module Ouroboros.Consensus.Storage.LedgerDB.Config
+  , module Ouroboros.Consensus.Storage.LedgerDB.Snapshots
```
jorisdral, line 10: I think we should only export `DiskPolicy` here as part of the API

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -565,12 +562,47 @@ data AnnLedgerError m l blk = AnnLedgerError {
...
+  doResolveBlock :: ResolveBlock m blk
+
+instance Monad m => ResolvesBlocks (ReaderT (ResolveBlock m blk) m) blk where
+  doResolveBlock r = ReaderT $ \f -> f r
+
+-- Quite a specific instance so we can satisfy the fundep
+instance Monad m
+      => ResolvesBlocks (ExceptT e (ReaderT (ResolveBlock m blk) m)) blk where
+  doResolveBlock = lift . doResolveBlock
+
```
jorisdral, line 597: Technically not part of the API, but part of initialisation (IMO that's impl-specific)

jorisdral (commented):
Some early comments

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: Should we keep an Int next to the tables handle to provide this value?

jorisdral: That might work -- you mean as in Consensus will keep track of the value instead of the LSM-tree?

jasagredo: Yup

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: Defered to next PR

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
jasagredo: Defered for next PR

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,462 @@
...
+     , LedgerSupportsProtocol blk
+     , NoThunks (l EmptyMK)
+     , GetTip l
+     )
+  => LedgerDBHandle impl m l blk
+  -> LedgerDBEnv impl m l blk
+  -> LedgerSeq m l
+  -> m (Forker m l blk)
+newForker h ldbEnv lseq = do
+  lseqVar <- newTVarIO lseq
```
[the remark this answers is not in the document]
jasagredo: Reworked now!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,462 @@
...
+  -> ForkerKey
+  -> m ()
+implForkerClose (LDBHandle varState) forkerKey = do
+    menv <- atomically $ readTVar varState >>= \case
+      LedgerDBClosed       -> pure Nothing
+      LedgerDBOpen ldbEnv -> do
+        stateTVar
+            (ldbForkers ldbEnv)
+            (Map.updateLookupWithKey (\_ _ -> Nothing) forkerKey)
+    whenJust menv (\e -> traceWith (foeTracer e) ForkerClose)
```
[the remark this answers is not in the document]
jasagredo: Reworked now!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Common.hs
```diff
@@ -0,0 +1,462 @@
...
+
+-- | Close all open block and header 'Follower's.
+closeAllForkers ::
+     MonadSTM m
+  => LedgerDBEnv impl m l blk
+  -> m ()
+closeAllForkers ldbEnv =
+    atomically $ writeTVar forkersVar Map.empty
+  where
+    forkersVar = ldbForkers ldbEnv
```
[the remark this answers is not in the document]
jasagredo: Reworked now!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
```diff
@@ -0,0 +1,190 @@
...
+        hs <- readTVarIO ioref
+        guardClosed hs (\st -> pure $ ltliftA2 rawRestrictValues st keys)
+    , readRange = undefined -- TODO (js) -- \(f, t) -> do
+        -- hs <- readTVarIO ioref
+        -- guardClosed hs (\(LedgerTables (ValuesMK m)) ->
+        --                   pure . LedgerTables . ValuesMK . fst . Map.split t . snd . Map.split f $ m)
+    , write = \diffs -> do
+        atomically
+        $ modifyTVar ioref
+        (`guardClosed` (\st -> LedgerTablesHandleOpen (ltliftA2 rawApplyDiffs st diffs)))
```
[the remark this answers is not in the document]
jasagredo: Not working, let's proceed

PR #859, opened 2024-01-03
Bump flake inputs and tools

amesgen:
- Update flake inputs
 - Update HLS to 2.5
 - Use upstream scriv as https://github.com/nedbat/scriv/pull/91 was merged and released in 1.4.0 (latest is 1.5.1 already).

### DIFF
--- a/docs/website/contents/for-developers/ReleaseProcess.md
+++ b/docs/website/contents/for-developers/ReleaseProcess.md
@@ -231,12 +231,12 @@ A - ...
 # Installing `scriv`
 
 To manage the workflow described above, we will use the `scriv` tool slightly
-modified to support cabal files.. If you use `nix` then you will find `scriv` in
+modified to support cabal files. If you use `nix` then you will find `scriv` in
 the Nix shell. Otherwise, the way to install it from source is:
 
-1. Clone [our fork of scriv](https://github.com/IntersectMBO/scriv/) and `cd`
-   into it.
-2. Run `pip install -e $(pwd)`
+```
+pip install scriv
+```
 
 If you encounter an error mentioning:
 `pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: ...` we
@@ -248,7 +248,7 @@ installation complains with the error `error: externally-managed-environment`,
 pass in the flag `--break-system-packages`:
 
 ```
-pip install --break-system-packages -e $(pwd)
+pip install --break-system-packages scriv
 ```
 
 # Adding a changelog fragment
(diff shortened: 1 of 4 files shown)


docs/website/contents/for-developers/ReleaseProcess.md
amesgen, line 1: Can a non-Nix user confirm that installing scriv like this works and yields a version >=1.4.0?

flake.lock
```diff
@@ -359,16 +360,16 @@
     "hls-2.4": {
```
RenateEilers, line 360: I have no idea what I'm talking about, but just in case: should the version here match the update in `shell.nix`?

amesgen: This file is the [Nix flake lock file](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake#lock-files), which contains the exact pins (eg git commit hashes + further integrity hashes) of everything external used by our flake here, and also all inputs used transitively by flakes we use etc. In this case, `hls-2.4` is *not* a direct input of our flake, but rather a transitive input via [haskell.nix](https://github.com/input-output-hk/haskell.nix).[^1]

We don't use this input ourselves currently (we use HLS 2.5 from Hackage)[^2], so there is no need to adapt anything to stay in sync with our declared version 2.5.0.0 in `shell.nix`. The flake lock file is usually updated automatically by commands like `nix flake update` which I ran for this PR, which updated haskell.nix and hence also this transitive flake input, which is why it shows up in the diff.

Hope that makes a bit of sense :smiley: 

[^1]: You can explore the flake graph interactively with eg [nix-melt](https://github.com/nix-community/nix-melt).
[^2]: Motivation why haskell.nix has this input: In the past, HLS didn't compile if you didn't use its own [`cabal.project`](https://github.com/haskell/haskell-language-server/blob/master/cabal.project) (which you don't get if you compile it from Hackage as we do ATM), so haskell.nix pinned HLS from GitHub which contains the `cabal.project`.

RenateEilers: Thank you for explaining!

docs/website/contents/for-developers/ReleaseProcess.md
fraser-iohk: yup, `pip3 install scriv` on mac installs `scriv-1.5.1` 👍

jasagredo: `pip3 install scriv --break-system-packages` in Ubuntu also installs `scriv-1.5.1`

jasagredo (approved):

PR #863, opened 2024-01-04
Homogenize Shelley versions in Cardano NTC versions

### LINKED ISSUE #862: Homogenize Shelley versions in Cardano NTC versions
Currently `CardanoNodeToClientVersion12` uses `ShelleyNodeToClientVersion8` for Conway but `ShelleyNodeToClientVersion7` for all previous Shelley-based eras.

This issue is to change it so that all Shelley-based eras advance to `Shelley8` in `Cardano12`. (No previous Cardano NTC version uses heterogeneous Shelley NTC versions.

(We think this will be a 5 line PR.)

The intent is to avoiding complicating the downstream versioning logic in eg `cardano-cli` (https://github.com/IntersectMBO/cardano-api/pull/403/)

Related:

- https://github.com/IntersectMBO/ouroboros-consensus/pull/191#discussion_r1252282714 - Esgen raised this idea in the original PR. Aniket's response can be resolved by using an additional mediating per-era type-class in the Consensus layer (which we already do for some other queries).

amesgen:
Closes #862 

Note that `CardanoNodeToClientVersion12` is experimental, so this change does not impact backwards-compatibility.


carbolymer (approved):

coot:
I like this changes :).

PR #867, opened 2024-01-08
A script for sanity checking build plans for releases

jorisdral:
# Description

Add a small script for sanity checking build plans

### DIFF
--- a/scripts/release/cabal-plan-diff.sh
+++ b/scripts/release/cabal-plan-diff.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+echo "=== cabal-plan-diff.sh ===
+
+In our CI and in most development setups, we build our projects with
+--enable-tests and --enable-benchmarks, but this can cause problems when
+uploading a package to CHaP (or Hackage)
+
+When a package is built from CHaP, only the dependencies on public libs are
+resolved. With --enable-tests and --enable-benchmarks, the dependencies for
+test-suites and benchmark suites are checked as well.
+
+So, enabling tests might lead to a build plan that compiles successfully.
+However, omitting these deps might lead to a different build plan that does not
+compile, for example because newer versions of packages are used.
+
+This script show the diff between two build plans for our published packages: (i)
+the local dev setup, and (ii) a fresh configuration that disables test suites and
+benchmark suitesthe, and uses the most recent index-states for CHaP and Hackage.
+
+After the script is run, check the diff to see if the build plan changes little.
+In particular, it is preferable to have no changes in major version for
+packages. It is fine if dependencies disappear.
+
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+      with:
+        path: ${{ steps.setup-haskell.outputs.cabal-store }}
+        key: cabal-release-${{ runner.os }}-${{ matrix.ghc }}-${{ env.MONTHNUM }}-${{ hashFiles('dependencies.txt') }}
+        restore-keys: |
+          cabal-release-${{ runner.os }}-${{ matrix.ghc }}-${{ env.MONTHNUM }}-${{ hashFiles('dependencies.txt') }}
+          cabal-release-${{ runner.os }}-${{ matrix.ghc }}-${{ env.MONTHNUM }}-
+
+    - name: Build dependencies
+      id: build-dependencies
+      run: cabal build --only-dependencies all -j
```
jasagredo, line 100: Do we really need to build the project? can't a cabal-plan diff be enough?

jasagredo: i.e. I think the libraries being `buildable` is much weaker than ensuring there are minimal changes wrt dependencies. In general I think we assume always that dependencies won't change a lot when testing and when not testing. But with this workflow they can change entirely as long as the project keeps building for the libs only.

jorisdral: Good point -- that means it can also be a regular sanity check for PRs. I'll do that instead

scripts/release/cabal-plan-diff.sh
```diff
@@ -0,0 +1,77 @@
...
+benchmarks: False
+
+packages:
+  ouroboros-consensus
+  ouroboros-consensus-cardano
+  ouroboros-consensus-protocol
+  ouroboros-consensus-diffusion
+  sop-extras
+  strict-sop-core
+" > ./tmp/cabal.project
```
jasagredo, line 68: Can't you instead do `cabal build all --dry-run --minimize-conflict-set --index-state=hackage.haskell.org,HEAD --index-state=cardano-haskell-packages,HEAD --disable-benchmarks --disable-tests`?

jorisdral: I chose to create a new file because it's future-proof. In the future, options could be added to the `cabal.project` file (like `s-r-p`s, `allow-newer`s,  other compiler options), in which case we would have to update the script's `cabal build all` command with additional arguments that override all the options in the `cabal.project` file

jorisdral: So all in all, it's to ensure that no local dev configuration is influencing the build plan

jorisdral:
Subsumed by #869

PR #869, opened 2024-01-09
Improve plan diff script and make it a GHA job

jasagredo:
# Description

The check builds both with and without test-suites+benchmarks, and shows a summary of the differences. It will fail if the "fresh" plan fails to build. It emits warnings on major bumps.

### DIFF
--- a/scripts/release/cabal-plan-diff.sh
+++ b/scripts/release/cabal-plan-diff.sh
@@ -4,7 +4,7 @@ echo "=== cabal-plan-diff.sh ===
 
 In our CI and in most development setups, we build our projects with
 --enable-tests and --enable-benchmarks, but this can cause problems when
-uploading a package to CHaP (or Hackage)
+uploading a package to CHaP (or Hackage).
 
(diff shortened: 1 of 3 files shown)


scripts/release/cabal-plan-diff.sh
```diff
@@ -0,0 +1,151 @@
...
+  sop-extras
+  strict-sop-core
+
+-- Remove once QSM 0.9 is released, see
+-- https://github.com/stevana/quickcheck-state-machine/pull/31
+source-repository-package
+  type: git
+  location: https://github.com/stevana/quickcheck-state-machine
+  tag: 8e2bfb2214c7af1dae728180f2092effea6fbeb7
+  --sha256: sha256-SRcuyKs0Sz+GBsCKE16LSnH14p3q8ZLxkcVyfErvfPQ=
```
jorisdral, line 76: The s-r-p shouldn't be included in the "fresh" `cabal.project` file, I think

jasagredo: Done

jorisdral (approved):

Issue #125, opened 2023-05-31
Run a sanity check on node startup and shutdown


dnadales:
We have had failed tests and benchmarks that resulted from wrong node configurations. This required several teams to investigate the source of the problem. We should perform a sanity checklist on startup and shutdown that would issue warnings when the configuration looks suspicious. Examples of said check include:

- `k` is the same for all HFC eras.
- Certain protocol versions (e.g. the HF triggers) are in ascending order.
- Warning if certain (derived) parameters seem "unreasonable", e.g. the ledger snapshot interval.

fraser-iohk:
I'm working on the assumption that this ticket can for all practical purposes subsumes #75 entirely, but please let me know if you don't think this is the case (@dnadales)

does this have any particular reason for being both a startup *and* a shutdown check? javier and I thought that running checks on startup would be sufficient, but did you have anything specific in mind that necessitates being checked at shutdown?

dnadales:
> does this have any particular reason for being both a startup and a shutdown check?

None that I can remember. So running these tests at startup should be fine. Feel free to edit the title and description!

nfrisby:
I edited the Issue description to include `k < maxBound`. EG their node would (slowly?) run out of memory due to the LedgerDB being effectively unbounded.

fraser-iohk:
The first change to `cardano-node` has the new `SanityCheckIssue` exceptions simply being traced in the standard way, but in future it should terminate the node if a fatal `SanityCheckIssue` is detected. Whether or not an issue is fatal will be determined by a new function inside `Ouroboros.Consensus.Block.SupportsSanityCheck`.

PR #874, opened 2024-01-11
Sanity checking for node configuration on startup 

fraser-iohk:
- This PR adds a a new mechanism for sanity checking node configuration on startup
- These checks are intended to operate as a warning to users that they may have misconfigured their node, and are designed to be easily bypassed by a user who is _intentionally_ configuring their node in an unusual but possibly acceptable way
- For safety reasons, the first change to `cardano-node` to support this will only log the exception in the standard way, though in future it should cause the node to terminate. We plan to add a way for the `cardano-node` tracer implementation to distinguish between sanity check issues which are fatal and those which are only warnings. 
- The first of these checks is simply to ensure that K, the security parameter, is consistent between all known eras. If eras' security parameters are different, a `SanityCheckIssue` exception will be traced in the new `sanityCheckIssueTracer`.

There is an [associated branch in the cardano-node repository](https://github.com/IntersectMBO/cardano-node/tree/fraser-iohk/consensus-startup-sanity-checks) with changes required to support this additional `Tracer`.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20240214_182440_fraser.murray_startup_sanity_checks.md
+++ b/ouroboros-consensus-cardano/changelog.d/20240214_182440_fraser.murray_startup_sanity_checks.md
@@ -1,0 +1,27 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+<!--
+
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category
+
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/SupportsSanityCheck.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/SupportsSanityCheck.hs
@@ -1,0 +1,81 @@
+{-# LANGUAGE NamedFieldPuns #-}
+module Test.Consensus.Cardano.SupportsSanityCheck (tests) where
+
+import           Ouroboros.Consensus.Cardano.Block
+import           Ouroboros.Consensus.Config
+import           Ouroboros.Consensus.HardFork.Combinator.Basics
+import           Ouroboros.Consensus.Node.ProtocolInfo
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+import           Test.Consensus.Cardano.ProtocolInfo
+import qualified Test.QuickCheck as QC
+import qualified Test.QuickCheck.Gen as Gen
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+import qualified Test.ThreadNet.Infra.Shelley as Shelley
+import           Test.Util.SanityCheck
+
+tests :: TestTree
+tests = testGroup "SupportsSanityCheck"
+  [ testProperty "cardano block top level config passes a sanity check" prop_cardanoBlockSanityChecks
+  , testProperty "intentionally-misconfigured top level config fails a sanity check" prop_intentionallyBrokenConfigDoesNotSanityCheck
+  ]
+
+prop_cardanoBlockSanityChecks :: QC.Property
+prop_cardanoBlockSanityChecks =
+  forAllBlind genSimpleTestProtocolInfo (prop_sanityChecks . pInfoConfig)
+
+prop_intentionallyBrokenConfigDoesNotSanityCheck :: QC.Property
+prop_intentionallyBrokenConfigDoesNotSanityCheck =
+  forAllBlind genSimpleTestProtocolInfo $ \pinfo ->
+    let saneTopLevelConfig =
+          pInfoConfig pinfo
+        brokenConfig = breakTopLevelConfig saneTopLevelConfig
+    in expectFailure $ prop_sanityChecks brokenConfig
+
+breakTopLevelConfig :: TopLevelConfig (CardanoBlock StandardCrypto) -> TopLevelConfig (CardanoBlock StandardCrypto)
+breakTopLevelConfig tlc =
+  let TopLevelConfig{topLevelConfigProtocol} = tlc
+      HardForkConsensusConfig{hardForkConsensusConfigK} = topLevelConfigProtocol
+      SecurityParam k = hardForkConsensusConfigK
+  in tlc
+    { topLevelConfigProtocol = topLevelConfigProtocol
+      { hardForkConsensusConfigK = SecurityParam (succ k)
+      }
+    }
+
+genSimpleTestProtocolInfo :: Gen (ProtocolInfo (CardanoBlock StandardCrypto))
+genSimpleTestProtocolInfo = do
+  setup <- arbitrary
+  pure $
+    mkSimpleTestProtocolInfo
+      (decentralizationParam setup)
+      (securityParam setup)
+      (byronSlotLength setup)
+      (shelleySlotLength setup)
+      (hardForkSpec setup)
+
+data SimpleTestProtocolInfoSetup = SimpleTestProtocolInfoSetup
+  { decentralizationParam :: Shelley.DecentralizationParam
+  , securityParam         :: SecurityParam
+  , byronSlotLength       :: ByronSlotLengthInSeconds
+  , shelleySlotLength     :: ShelleySlotLengthInSeconds
+  , hardForkSpec          :: HardForkSpec
+  }
+
+instance Arbitrary SimpleTestProtocolInfoSetup where
+  arbitrary = do
+    SimpleTestProtocolInfoSetup
+      <$> arbitrary
+      <*> genSecurityParam
+      <*> genByronSlotLength
+      <*> genShelleySlotLength
+      <*> genHardForkSpec
+    where
+      genSecurityParam =
+        SecurityParam <$> Gen.choose (8, 12)
+      genByronSlotLength =
+        ByronSlotLengthInSeconds <$> Gen.choose (1, 4)
+      genShelleySlotLength =
+        ShelleySlotLengthInSeconds <$> Gen.choose (1, 4)
+      genHardForkSpec =
+        hardForkInto <$> Gen.chooseEnum (Byron, Conway)
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -60,7 +60,7 @@ import qualified Codec.CBOR.Encoding as CBOR
 import           Codec.Serialise (DeserialiseFailure)
 import qualified Control.Concurrent.Class.MonadSTM.Strict as StrictSTM
 import           Control.DeepSeq (NFData)
-import           Control.Monad (when)
+import           Control.Monad (forM_, when)
 import           Control.Monad.Class.MonadTime.SI (MonadTime)
 import           Control.Monad.Class.MonadTimer.SI (MonadTimer)
 import           Control.Tracer (Tracer, contramap, traceWith)
@@ -421,6 +421,9 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                 -- When the last shutdown was not clean, validate the complete
                 -- ChainDB to detect and recover from any disk corruption.
               = ChainDB.ensureValidateAll
+
+        forM_ (sanityCheckConfig cfg) $ \issue ->
+          traceWith (consensusSanityCheckTracer rnTraceConsensus) issue
 
         (chainDB, finalArgs) <- openChainDB
                      registry
--- a/ouroboros-consensus/changelog.d/20240207_130158_fraser.murray_startup_sanity_checks.md
+++ b/ouroboros-consensus/changelog.d/20240207_130158_fraser.murray_startup_sanity_checks.md
@@ -1,0 +1,22 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Add BlockSupportsSanityCheck to check for common configuration issues which may manifest themselves in unusual but not necessarily immediately obvious ways. For now it only checks that `k` is the same across all eras.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
@@ -1,0 +1,82 @@
+{-# LANGUAGE LambdaCase #-}
+-- | This module adds support for sanity checking consensus configuration
+--   on node startup. These checks should primarily look for unusual
+--   configuration choices that may point to an accidentally-misconfigured node
+--   and quietly cause problems, rather than incoherent configurations that will
+--   result in fatal errors at a later point.
+--
+--   While in most situations they can be handled as fatal issues, there are
+--   situations when intentionally configuring a node "weirdly" can be useful,
+--   and so the user should be able to opt out of the sanity checks at their
+--   own peril.
+module Ouroboros.Consensus.Block.SupportsSanityCheck (
+    BlockSupportsSanityCheck (..)
+  , SanityCheckIssue (..)
+  , checkSecurityParamConsistency
+  , sanityCheckConfig
+  ) where
+
+import           Control.Exception
+import           Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import           Data.Maybe (catMaybes)
+import           Ouroboros.Consensus.Config (TopLevelConfig)
+import           Ouroboros.Consensus.Config.SecurityParam
+
+-- | An issue found in the 'TopLevelConfig' for a block. See 'displayException'
+--   for human-readable descriptions of each of these cases, especially when
+--   presenting these to users.
+data SanityCheckIssue
+  -- | Configuration contains multiple security parameters. This may cause
+  --   strange behaviour around era boundaries.
+  = InconsistentSecurityParam (NonEmpty SecurityParam)
+  deriving (Show, Eq)
+
+instance Exception SanityCheckIssue where
+  displayException = \case
+    InconsistentSecurityParam ks -> mconcat
+      [ "InconsistentSecurityParam: "
+      , "SecurityParams (K) were found to be inconsistent between constituent "
+      , "eras of a HardForkBlock: "
+      , show (NonEmpty.toList ks)
+      ]
+
+-- | 'BlockSupportsSanityCheck' provides evidence that a block can be sanity
+--   checked for common issues on node startup. 'sanityCheckConfig', which runs
+--   performs each check and returns a list with each 'SanityCheckIssue' found,
+--   should be preferred over using these methods directly.
+class BlockSupportsSanityCheck blk where
+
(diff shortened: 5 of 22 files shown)


nfrisby:
> is a Tracer the correct way to report these errors / warnings? 

My immediate thought is that the kinds of errors we're concerned with here deserve exceptions.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
```diff
@@ -12,6 +12,7 @@
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
```
nfrisby, line 15: This seems unnecessary?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
```diff
@@ -112,6 +115,13 @@ instance CanHardFork xs => ConsensusProtocol (HardForkProtocol xs) where
   -- Security parameter must be equal across /all/ eras
   protocolSecurityParam = hardForkConsensusConfigK
 
+  protocolSecurityParamConsistencyCheck HardForkConsensusConfig {..} = do
+    let allSecurityParams = hardForkConsensusConfigK :|
+          perEraConsensusConfigSecurityParams hardForkConsensusConfigPerEra
+    case nub allSecurityParams of
```
nfrisby, line 121: `nub` is quadratic; use eg `Data.Set` instead

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
```diff
@@ -112,6 +115,13 @@ instance CanHardFork xs => ConsensusProtocol (HardForkProtocol xs) where
   -- Security parameter must be equal across /all/ eras
   protocolSecurityParam = hardForkConsensusConfigK
 
+  protocolSecurityParamConsistencyCheck HardForkConsensusConfig {..} = do
+    let allSecurityParams = hardForkConsensusConfigK :|
```
nfrisby, line 119: Minor: `NonEmpty` is somewhat overkill; `1 < Data.Set.size (Data.Set.fromList (...))` would suffice, I think. (and avoid `nub`)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -170,6 +171,9 @@ class ( Show (ChainDepState   p)
   -- | We require that protocols support a @k@ security parameter
   protocolSecurityParam :: ConsensusConfig p -> SecurityParam
```
nfrisby, line 172: Hmm. This is an awkward juxtaposition :/
```
  protocolSecurityParam                 :: ConsensusConfig p -> SecurityParam
  protocolSecurityParamConsistencyCheck :: ConsensusConfig p -> Maybe (NonEmpty SecurityParam)
```

I'm not sure what to do about it. My initial idea is to make the check specific to the hardfork combinator somehow? 🤔 Since that's where block types collide. (Ie don't put it in our of our fully abstract classes like `ConsensusProtocol` here.

Or perhaps just changing the return type to `Maybe StartupWarning` would suffiice to avoid the awkwardness? Still kind of feels like it belongs in a different class

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
nfrisby: Err, just `or (hardFOrkConsensusConfigK /=) $ perEraConsensusConfigSecurityParams hardForkConsensusConfigPerEra` actually.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
fraser-iohk: removed, don't need a `nub` effect at all

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
fraser-iohk: removed this after alex suggested an entirely separate `BlockSupportsSanityCheck` class

fraser-iohk:
> My immediate thought is that the kinds of errors we're concerned with here deserve exceptions.

when I discussed this initially with damien and javier, we thought that tracers would be better, since the "errors" we want to raise here shouldn't necessarily be fatal, but probably should at least be communicated to the user as a warning. we thought that implementing them with a `Tracer` meant that we'd have the flexibility to change it easily and have different issues with varying severity. I don't think exceptions are a bad idea, but it feels a little more awkward to have someone (probably a developer) add `catch`es than to add a filter to a `Tracer`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,89 @@
...
+-- | Run all supported sanity checks on a given "TopLevelConfig".
+sanityCheckConfig
+  :: BlockSupportsSanityCheck blk
+  => TopLevelConfig blk
+  -> [SanityCheckIssue]
+sanityCheckConfig cfg =
+  catMaybes [checkSecurityParamConsistency cfg]
+
+class ProtocolConfigHasSecurityParam p where
+  protocolConfigSecurityParam :: ConsensusConfig p -> SecurityParam
```
amesgen, line 89: Is there a particular reason to define this class instead of using the existing `protocolSecurityParam` from `ConsensusProtocol`?

fraser-iohk: no reason except recycling the PartialConfig-related version that this was using previously -- using `protocolSecurityParam` now, thanks for the suggestion 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
fraser-iohk: collecting the security params allows us (or node, whoever) to spit out the actual values of k in error messages, which is nice (and in theory allows HFC-style multi-type blocks with more than one level of nesting, which I don't think is actually useful)

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,83 @@
...
+prop_intentionallyBrokenConfigDoesNotSanityCheck :: QC.Property
+prop_intentionallyBrokenConfigDoesNotSanityCheck =
+  forAllBlind genSimpleTestProtocolInfo $ \pinfo ->
+    let saneTopLevelConfig =
+          pInfoConfig pinfo
+        brokenConfig = breakTopLevelConfig saneTopLevelConfig
+    in expectFailure $ prop_sanityChecks brokenConfig
+
+breakTopLevelConfig :: TopLevelConfig (CardanoBlock StandardCrypto) -> TopLevelConfig (CardanoBlock StandardCrypto)
+breakTopLevelConfig tlc = do
```
nfrisby, line 38: Minor style: we generally avoid `do` without a monad/applicative type.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+issueDescription :: SanityCheckIssue -> String
+issueDescription issue = case issue of
+  InconsistentSecurityParam ks ->
+    mconcat
+      [ "SecurityParams (K) were found to be inconsistent between constituent "
+      , "eras of a HardForkBlock: "
+      , show (NonEmpty.toList ks)
+      ]
+
+-- | "BlockSupportsSanityCheck" provides evidence that a block can be sanity
```
nfrisby, line 46: Minor: multiple Haddock comments in this module use double-quotes instead of single-quotes, and so they won't become links.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+--   checked for common issues on node startup. "sanityCheckConfig", which runs
+--   performs each check and returns a list with each "SanityCheckIssue" found,
+--   should be preferred over using these methods directly.
+class BlockSupportsSanityCheck blk where
+
+  -- | Generate a "NonEmpty" list of security parameters for a given block type.
+  --   For individual eras' block types, this is simply a singleton list
+  --   containing the chosen "SecurityParam", but combined block types will
```
nfrisby, line 54: Please explicitly list the HardForkCombinator as an example of a "combined block type".

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+  --   return all of their constituent eras' configurations' security
+  --   parameters.
+  configAllSecurityParams
+    :: TopLevelConfig blk
+    -> NonEmpty SecurityParam
+
+-- | Check a "TopLevelConfig" for any inconsistency in constituent choices for
+--   "SecurityParam" (colloquially 'k'). For a block type to be considered
```
nfrisby, line 62: Haddock markup: `'k'` -> `@k@`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,84 @@
...
+allSame :: Eq a => NonEmpty a -> Bool
+allSame (x :| xs) = all (x ==) xs
+
+-- | Run all supported sanity checks on a given "TopLevelConfig".
+sanityCheckConfig
+  :: BlockSupportsSanityCheck blk
+  => TopLevelConfig blk
+  -> [SanityCheckIssue]
+sanityCheckConfig cfg =
+  catMaybes [checkSecurityParamConsistency cfg]
```
nfrisby, line 84: Immediate reaction was slightly confused about the singleton list, but this absolutely makes sense given that we're planning to add more checks.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
```diff
@@ -59,6 +59,7 @@ class ( LedgerSupportsProtocol blk
       , HasTxId (GenTx blk)
       , QueryLedger blk
       , HasPartialConsensusConfig (BlockProtocol blk)
+      -- , PartialConsensusConfigHasSecurityParam (BlockProtocol blk)
```
nfrisby, line 62: Dead code?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node/SanityCheck.hs
```diff
@@ -0,0 +1,38 @@
...
+import           Ouroboros.Consensus.HardFork.Combinator.AcrossEras
+import           Data.SOP.Constraint
+import           Data.SOP.BasicFunctors
+import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.HardFork.Combinator.PartialConfig
+import           Data.SOP.Strict
+import           Ouroboros.Consensus.HardFork.History.EpochInfo
+
+instance CanHardFork xs => BlockSupportsSanityCheck (HardForkBlock xs) where
+  configAllSecurityParams tlc = do
```
nfrisby, line 24: Minor: `do` seems unnecessary/distracting here

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/SanityCheck.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE RankNTypes #-}
```
nfrisby, line 1: Minor: None of these LANGUAGE pragmas are necessary, right?

nfrisby (dismissed):
I'm "Requesting changes" only because of the dangling dead code comment.

Otherwise, looks very good. Great comments. Idiomatic design. 👍 

Regarding tracers, versus exceptions: tracers does seem reasonable, given that these checks are not necessarily fatal.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -387,6 +388,10 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                 initLedger
                 llrnChainDbArgsDefaults
                 customiseChainDbArgs'
+
+        forM_ (sanityCheckConfig cfg) $ \issue ->
+          traceWith (consensusSanityCheckTracer rnTraceConsensus) issue
```
amesgen, line 393: Hmm, I see that these are currently "just" traced. I agree with https://github.com/IntersectMBO/ouroboros-consensus/pull/874/#issuecomment-1887436828 that the check for security parameter consistency should be an exception, not just a warning. In contrast, eg

> Warning if certain (derived) parameters seem "unreasonable", e.g. the ledger snapshot interval.

from #125 seem like a candidate for a warning.

Did you intend this logic to be part of the implementation of `rnTraceConsensus`? Or would it make sense to let `sanityCheckConfig` return a list of either warnings or exceptions, and trace or throw them here, respectively?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,83 @@
...
+
+instance Exception SanityCheckIssue
+
+issueDescription :: SanityCheckIssue -> String
+issueDescription (InconsistentSecurityParam ks) =
+  mconcat
+    [ "SecurityParams (K) were found to be inconsistent between constituent "
+    , "eras of a HardForkBlock: "
+    , show (NonEmpty.toList ks)
+    ]
```
amesgen, line 43: Maybe this could be the implementation of `displayException` in addition to/instead of a separate function?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Node.hs
```diff
@@ -55,6 +55,10 @@ instance NodeInitStorage (SimpleBlock SimpleMockCrypto ext) where
 instance BlockSupportsMetrics (SimpleBlock c ext) where
   isSelfIssued = isSelfIssuedConstUnknown
 
+instance ConsensusProtocol (BlockProtocol (SimpleBlock c ext)) => BlockSupportsSanityCheck (SimpleBlock c ext) where
+  configAllSecurityParams =
+    pure . protocolSecurityParam . topLevelConfigProtocol
```
amesgen, line 60: Golf:
```suggestion
    pure . configSecurityParam
```
(this should work for all single-era blocks)

ouroboros-consensus-cardano/changelog.d/20240214_182440_fraser.murray_startup_sanity_checks.md
```diff
@@ -0,0 +1,24 @@
...
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Add BlockSupportsSanityCheck instances for all Cardano (and related tests') block types
```
jasagredo, line 16: We don't mention tests in changelogs

ouroboros-consensus/changelog.d/20240207_130158_fraser.murray_startup_sanity_checks.md
```diff
@@ -0,0 +1,22 @@
...
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Add BlockSupportsSanityCheck to check for common configuration issues which may manifest themselves in unusual but not necessarily immediately obvious ways
```
jasagredo, line 15:
```suggestion
- Add BlockSupportsSanityCheck to check for common configuration issues which may manifest themselves in unusual but not necessarily immediately obvious ways. For now it checks that `k` is the same across all eras.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsSanityCheck.hs
```diff
@@ -0,0 +1,82 @@
...
+  ) where
+
+import           Control.Exception
+import           Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import           Data.Maybe (catMaybes)
+import           Ouroboros.Consensus.Config (TopLevelConfig)
+import           Ouroboros.Consensus.Config.SecurityParam
+
+-- | An issue found in the 'TopLevelConfig' for a block. See 'issueDescription'
```
jasagredo, line 26: `issueDescription` does not exist?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node/SanityCheck.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Ouroboros.Consensus.HardFork.Combinator.Node.SanityCheck
+  () where
```
jasagredo, line 7: This indentation? did stylish-haskell do this?

fraser-iohk: haven't run stylish on this branch at all, apparently 😵

fraser-iohk:
appropriate changes to `cardano-node` are available on the `fraser-iohk/consensus-startup-sanity-checks` branch: https://github.com/IntersectMBO/cardano-node/tree/fraser-iohk/consensus-startup-sanity-checks

fraser-iohk:
> Wasn't the idea (also see [#874 (comment)](https://github.com/IntersectMBO/ouroboros-consensus/pull/874#discussion_r1560717398)) to throw an exception for `InconsistentSecurityParam` (as part of the tracer, based on a "fatality" classification function in Consensus)?

yep, that's the goal for the future, but for the first iteration (when stuff might go wrong) I've just made it trace the exception since I don't want to potentially break `node` use cases without warning the user for at least one version first

amesgen (approved):
> > Wasn't the idea (also see [#874 (comment)](https://github.com/IntersectMBO/ouroboros-consensus/pull/874#discussion_r1560717398)) to throw an exception for `InconsistentSecurityParam` (as part of the tracer, based on a "fatality" classification function in Consensus)?
>
> yep, that's the goal for the future, but for the first iteration (when stuff might go wrong) I've just made it trace the exception since I don't want to potentially break `node` use cases without warning the user for at least one version first

Makes sense, can you record that somewhere, maybe in #125?

---

Apart from that, LGTM after formatting, squasing and updating the PR description :+1:

PR #875, opened 2024-01-11
Extract peer-related code to own module

Niols:
This is a cherry pick from https://github.com/IntersectMBO/ouroboros-consensus/pull/856. I believe those commits would be useful in general. They extract code related to the `Peers` type from `Test.Consensus.PointSchedule` to `Test.Consensus.PointSchedule.Peers`, making the code clearer. It also adds a bunch of new helpers for peers.


facundominguez (approved):
LGTM

PR #877, opened 2024-01-11
Add explicit shrinking to `forAllGenesisTest`

Niols:
This is a cherry pick from https://github.com/IntersectMBO/ouroboros-consensus/pull/856. I believe those commits would be useful in general. They introduce explicit shrinking in `forAllGenesisTest`. For now, this is mandatory because we have several flavours of `forAllGenesisTest` for different types of point schedules. If we ever agree on just one type, we could imagine shrinking becoming the default in `forAllGenesisTest` so that tests would not have to specify it.


ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/QuickCheck.hs
```diff
@@ -72,6 +73,32 @@ checkShrinker p =
...
+  Testable prop =>
+  Gen input ->
+  (input -> output) ->
+  (input -> output -> [input]) ->
+  (input -> output -> prop) ->
+  Property
+forAllGenRunShrinkCheck gen run shrink_ check =
+  forAllBlind gen $ \input ->
+    shrinking
+      (run' <.> uncurry shrink_)
```
facundominguez, line 95: I still find `map run' . uncurry shrink_` or `(\(i, o) -> map run' (shrink i o))` easier to grok.

facundominguez (approved):
LGTM

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/QuickCheck.hs
Niols: Sorry, I extracted the commits from the other PR but forgot to also apply this comment. No strong opinion on what's better; I took your first suggestion in 27f63dec1ee3f4fb14e7682dea46b65cd365b2f5.

PR #880, opened 2024-01-12
Include point schedule in `GenesisTest`

Niols:
Not super sure about this one.

I feel it makes sense to rework this type because I keep manipulating pairs of `GenesisTest` and point schedule (sometimes `PointSchedule` and sometimes `Peers PeerSchedule`). Considering that a point schedule does not make sense in itself but only within a context, which is what `GenesisTest` provides, I think it makes sense to have a type contain all the information. We could also decide that types such as `PointSchedule` must be self-contained, but I am not entirely convinced by this solution either. What do you guys think?

Also, this might be the right moment to rename this type or to move it. For now, it lives in `Test.Consensus.PointSchedule` and is called `GenesisTest`. My poor naming skills would make me want to rename it into something like `RichPointSchedule`, or just `Rich` so as to write `Rich PointSchedule` or `Rich (Peers PeerSchedule)` but I'm not convinced :p


Niols:
Rebased on `genesis/milestone-7-9-10` now.

tek:
could be called `PeerSimulation` or something else that ties it to the simulator

PR #885, opened 2024-01-16
Update QSM, remove Sing and Dict

jasagredo:


cabal.project
```diff
@@ -14,7 +14,7 @@ repository cardano-haskell-packages
 -- update either of these.
 index-state:
   -- Bump this if you need newer packages from Hackage
-  , hackage.haskell.org 2023-12-14T23:57:38Z
+  , hackage.haskell.org 2024-01-16T12:10:07Z
```
jorisdral, line 17: `hackageNix` should be updated too

scripts/release/cabal-plan-diff.sh
```diff
@@ -72,82 +72,81 @@ cabal build all --dry-run --minimize-conflict-set --project-file='./tmp/cabal.pr
...
-echo -e "- PATCH BUMPS"
-printf '%s\n' "${patch[@]}" | sed 's/^\(.\+\)/   - \1/g'
-
-if [[ -n $GH && ${#major[@]} ]]; then
-  str=$(printf '%s\n' "${major[@]}")
-  echo "::warning title=Major bumps detected::$str"
-fi
 
-rm -r ./tmp
+  rm -r ./tmp
```
jorisdral, line 151: This cleanup should be outside of the `if-else` statements

ouroboros-consensus-cardano/test/byron-test/Test/Consensus/Byron/Serialisation.hs
```diff
@@ -46,7 +47,7 @@ tests = testGroup "Byron"
         ]
     ]
   where
-    dictNestedHdr :: forall a. NestedCtxt_ ByronBlock Header a -> Dict (Eq a, Show a)
+    dictNestedHdr :: forall a. NestedCtxt_ ByronBlock Header a -> Dict (And Eq Show) a
```
amesgen, line 50: Minor: Thinking out loud: Do we actually want to use the `Data.SOP.Dict` everywhere? I think it is more natural to still use a "normal" `Dict` (eg from [`constraints`](https://hackage.haskell.org/package/constraints-0.14/docs/Data-Constraint.html#t:Dict)) in places like this, and only use the `sop-core` `Dict` when we have to partially apply it.

Also, this approach won't work *directly* (without sth like `Id` just below), eg for `Dict (F a)` where `F` is a type family.

Or instead of using a different `Dict`, just use `Dict'` defined like this in terms of the `sop-core` `Dict`:
```haskell
type Dict' c = Dict Id c

type Id :: Constraint -> Constraint
class    (c) => Id c
instance (c) => Id c
```

cabal.project
jasagredo: Indeed but as Alex mentioned this has to wait 1 day

scripts/release/cabal-plan-diff.sh
jasagredo: Good point

ouroboros-consensus-cardano/test/byron-test/Test/Consensus/Byron/Serialisation.hs
jasagredo: Downgrading it to `constaints` Dicts is also fine for me. I just didn't want just another Dict type around

amesgen: Another idea: to avoid the perception of different kinds of `Dict`s, we could also enforce that eg the `sop-core` `Dict` always has to be imported qualified (via hlint?).

scripts/release/cabal-plan-diff.sh
jasagredo: Implemented with `trap`

cabal.project
jasagredo: Done

ouroboros-consensus-cardano/test/byron-test/Test/Consensus/Byron/Serialisation.hs
jasagredo: >Also, this approach won't work directly (without sth like Id just below), eg for Dict (F a) where F is a type family.

I think we have a fairly heavy usage of type families in Consensus that would justify keeping the SOP Dict. Or maybe I'm misunderstanding you

amesgen: I mean that you can't rewrite `Data.Constraint.Dict (F a)` to `Data.SOP.Dict` without sth like `Id` AFAIK.

amesgen: Resolved to use `constraint` `Dict` by default, and only `sop-core` `Dict` when necessary and with a qualified import.

jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -125,7 +125,7 @@ instance Inject SerialisedHeader where
 instance Inject WrapHeaderHash where
   inject _ (idx :: Index xs x) =
     case dictIndexAll (Proxy @SingleEraBlock) idx of
-      Dict ->
+      Dict.Dict ->
```
amesgen, line 128: Any particular reason to qualify `Dict` here? In other places, you imported it unqualified (but still explicitly).

amesgen (approved):
Thanks, LGTM, I restarted the failed Windows jobs and they now succeed :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
jasagredo: No reason honestly, let me fix that

PR #888, opened 2024-01-17
Ensure chain schemas have k+1 active slots after intersections

facundominguez:
Rebase of #477 on top of `genesis/milestone8`.

This PR causes chain schemas to have more than $k$ blocks after intersections.

Having more than k blocks after intersections makes the inputs interesting for the long range attack, and for testing that rollbacks of more than $k$ can't happen in caught up nodes.

This guarantee comes from the generation algorithm, which ensures that every $s$ window has exactly $k$ blocks in the honest chain.

For alternative chains, we ensure that the chains have more than s+k+d+1 slots, and we choose the first active slot before this many slots from the end. Then the acceleration bound is before the last k blocks of the alternative chain, which we always set to active.

Race windows do not push the acceleration bound to the last k blocks because the intersections with the honest chain are always behind the last block preceding the last s+k+d+1 slots.

### DIFF
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
@@ -113,26 +113,34 @@ data NoSuchHonestChainSchema =
 -- rollbacks.
 --
 -- The minimum length is calculated to allow k+1 active slots in every schema,
--- and then allowing the intersection to be the genesis block.
+-- and then allowing the intersection to be the genesis block. When the schemas
+-- are longer than the minimum, chosing a later intersection is possible.
 --
--- To ensure the honest schema has at least k+1 active slots, we need a length
--- of 2s - (k - 1). We know that because 2s slots have at a minimum 2k blocks
--- because of the chain growth assumption. But that is ensuring k-1 more blocks
--- than we actually need. Thus it's safe to remove the k-1 slots.
+-- We divide the length of the schemas in the following three segments:
+--
+-- 2s - (k - 1) | d | k
+--
+-- To ensure the honest schema has at least k+1 active slots, we need at least
+-- the length of the first segment: 2s - (k - 1). We know that because 2s slots
+-- have at a minimum 2k blocks by the chain growth assumption. But that would be
+-- ensuring k-1 more blocks than we want. Thus we subtract the redundant k-1
+-- slots.
 --
 -- To ensure the alternative schema can have k+1 active slots, we reserve
--- k unstable slots at the end of the schema, and we make sure to activate one
--- slot a stability window earlier.
+-- k unstable slots at the end of the schema (this is the last of our segments),
+-- and we make sure to activate one more earlier slot.
 --
--- To reserve k unstable slots, there needs to be a gap of d slots between
--- the first unstable slot and the k+1st active slot in the honest schema. Thus
--- we need to extend the previous length by d+k, to obtain
--- 2s - (k - 1) + d + k = 2s + d + 1.
+-- To reserve k unstable slots, we need to meet two conditions. One condition is
+-- that there needs to be a gap of d slots between the first unstable slot and
+-- the k+1st active slot in the honest schema. This is the middle segment in the
+-- schema. Therefore, to obtain the total length, we need to extend the length
+-- of the first segment by d+k, to obtain 2s - (k - 1) + d + k = 2s + d + 1.
 --
--- To ensure we can activate a slot at least a stability window earlier, we
--- chose k greater or equal to two, which we already needed to ensure that the
--- alternative schema can have at least one active slot, yet lose density and
--- race comparisons.
+-- The second condition for unstable slots requires that the first alternative
+-- active slot after the intersection is s slots before the first unstable slot.
+-- To ensure that we can activate a slot at this point or earlier, we chose k>1,
+-- which we already needed to ensure that the alternative schema can have at
+-- least one active slot, yet lose density and race comparisons.
 
 genHonestRecipe :: QC.Gen HonestRecipe
 genHonestRecipe = sized1 $ \sz -> do


ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -105,10 +105,14 @@ data NoSuchHonestChainSchema =
...
 genHonestRecipe = sized1 $ \sz -> do
     (Kcp k, Scg s, Delta d) <- genKSD
-    -- 2s slots has at least 2k blocks. But that is ensuring k-1 more blocks
+    -- 2s slots have at least 2k blocks. But that is ensuring k-1 more blocks
     -- than we actually need. Thus it's safe to remove the k-1 last slots.
-    -- Therefore we set for a length of at least 2s - (k - 1).
-    l <- (+ (2*s - (k - 1))) <$> QC.choose (0, 5 * sz)
+    -- Therefore we first set for a length of at least 2s - (k - 1).
+    --
+    -- If we then add k+d slots (to get a total size of 2s+d+1), we get enough
```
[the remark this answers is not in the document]
nfrisby: I'm not remembering why adding `k+d` slots ensures room for `k+1` adversarial active slots after the intersection.

What I do recall is that the basic idea is to ensure there are enough slots after the stability window anchored in the first adversarial active slot. I _think_ "enough" means `k`, since the first adversarial active slot that anchors the delimiting window counts as the `+1`.

The following is I've come up with (after writing PR https://github.com/IntersectMBO/ouroboros-consensus/pull/897 in preparation)---though I'm not constraining my reasoning here with the assumption that `uniformAdversarialChain` can't change.

I'm hoping I'm missing something, and the reasoning you've already done is simpler :/ At least my counts are similar to the one you have.

----

Let sai abbreviate "slot after the intersection".
Let asai abbreviate "active sai".

`uniformAdversarialChain` requires that the honest schema permits an intersection with both `Kcp+1` honest asais and also enough sais to also allow at least `Kcp+1` alternative asais.
Specifically, there must be enough sais so that at least `Kcp` slots are _unstable_, ie both more than `Del` after the `Kcp+1`st honest asai and also more than `Scg` after the first alternative asai.

Supose `1<Kcp`.
The `Kcp+1`st honest asai can be no later than the `2Scg-(Kcp-1)`th sai.
The unstable slots could begin as late as `Del+1` after that.
The first alternative asai is unconstrained when `1<Kcp`, and thus merely needs to be no later than the `2Scg-(Kcp-1)+Del+1`st sai to ensure there are at least `Kcp` unstable slots before the end of the schema.
Thus `2Scg+Del+2` sais would suffice to ensure there are at least `Kcp` unstable slots before the end of the schema.

Suppose `Kcp=0`.
Note that there does not need to be an unstable slots, merely at least one alternative asai.
The `Kcp+1`st honest asai is the first honest sai, and so can be no later than the `Scg`th sai, since Genesis requires the first `Scg` sais contains _strictly_ more honest active than alternative active.
(This requires that the honest schema has two active slots in the same window of `Scg+1` slots, one of which may be the genesis block.)
Thus it's possible that the first alternative asai can be no sooner than `Scg+Del+1`.
Thus `Scg+Del+1` sais would suffice to ensure there are at least `Kcp=0` unstable slots before the end of the schema (actually: that there is at least one alternative asai).

Suppose `Kcp=1`.
(Again, the honest schema must have two active slots in the same window of `Scg+1` slots, one of which may be the genesis block.)
This is similar to the `1<Kcp` case, except there might only be one honest active in the first `Scg` sais, and so the first alternative active sai might need to be after the `Scg`th sai, meaning the unstable slots couldn't begin before the `2Scg+2`nd sai.
Also, the `Kcp+1=2`nd honest asai might be as late as the `2Scg`th sai, so the unstable slots might not begin until the `2Scg+Del+1`st sai.
Thus `2Scg + max (Del+2) 3` sais would suffice to ensure there are at least `Kcp=1` unstable slot before the end of the schema.

Since the genesis block can always be chosen as the intersection (as long as the first honest active slot is at most the `Scg`th slot), the above sai counts are also the minimum bound for `Len` in order to ensure at least `Kcp+1` asais for both honest and alternative.

nfrisby (commented):
I'm having trouble convincing myself that the

> If we then add k+d slots (to get a total size of 2s+d+1), we get enough room for alternative chains to have k+1 blocks when they branch before the last s+d+1 slots.

part is correct.

Are we assuming somewhere that `1<k`? Otherwise, I think the `k=1` case might not necessarily work. Also, maybe there's an off by one, a la `2s+d+2`?

Please see my big GitHub comment.

facundominguez (commented):
I don't think my changes had great comments. This conversation is going to improve that a lot.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: Heavily edited my previous comment.

facundominguez:
I made an attempt at explaining the minimum schema length in a code note. Let me know if it helps.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: > At the moment Kcp is always chosen as 2 or more.

If we're relying on that, we should enforce it in `checkHonestRecipe`.

nfrisby: How about the following?

```
{- |

Note [Minimum schema length]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Consider this diagram of a necessary prefix of the schemas' slots (the schemas
may have a suffix of more slots, not shown).

@
    GA................B......C.......D
@

* @G@ is the genesis point.

* @A@ is the first slot in the schema.

* @B@ is the @2s - (k - 1)@th slot in the schema.

* @C = B + d@.

* @D = C + k@.

Let sai abbreviate /slot after the intersection/. Let asai abbreviate /active
sai/.

Suppose @G@ is the intersection, and consider these facts.

* The hypothetical closed interval @(A, 2s)@ (which might exceed @D@) would
  contain at least @2k@ honest asais. Thus the interval @(A, B)@ contains at
  least @2k-k+1=k+1@ honest asais; so @k+1@ honest asais are ensured.

* All slots @>C@ can be unstable, since both requirements are satisfiable.
  First, @C@ is at least @d@ after the @k+1@st honest asai (which is @<=B@), so
  @C+1@ can be the @k+1@st adversarial asai, if necessary. Second, the
  generators require @1<k@ so that @A@ can be adversarial active if necessary,
  which ensures that the slot @C+1@ is more than @s@ after the first
  adversarial asai.

* Since they are unstable, the @k@ slots in the half-open interval @[C, D)@ can
  be adversarial active. Together with @A@ if necessary, this ensures at least
  @k+1@ adversarial asais.

* Therefore, @2s + 2k + d - 1@ ensures @G@ is a valid intersection.

-}
```

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -576,15 +786,42 @@ withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
 --
 -- The count will be strictly smaller than the number of active slots in the given 'ChainSchema'.
```
nfrisby, line 787: I think you can remove this line now

> T`he count will be strictly smaller than the number of active slots in the given 'ChainSchema'.`

since we're only ensuring `k+1` after the intersection, instead of only ensuring `1`.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -576,15 +786,42 @@ withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
...
+--
+-- PRECONDITION: @schemaSize schedH >= 2*s + d + 1@
+genPrefixBlockCount :: R.RandomGen g => HonestRecipe -> g -> ChainSchema base hon -> C.Var hon 'ActiveSlotE
+genPrefixBlockCount (HonestRecipe (Kcp k) (Scg s) (Delta d) _len) g schedH
+    | C.lengthV v < C.Count (2*s + d + 1) =
+        error "size of schema is smaller than 2*s + d + 1"
+    | numBlocks == C.Count 0 =
+          -- Pick genesis if there are no blocks
+          0
+    | otherwise =
```
nfrisby, line 802: How about the following?

```
    | otherwise =
        -- @uniformIndex n@ yields a value in @[0..n-1]@, we add 1 to the
        -- argument to account for the possibility of intersecting at the
        -- genesis block
        C.toVar $ R.runSTGen_ g $ C.uniformIndex (phase2 C.+ 1)
  where
    ChainSchema winH vH = schedH

    -- Ultimately, the intersection must have >=k+1 honest asais and >=k+1
    -- adversarial asais.
    --
    -- Our crude approach is as follows; refer to the Note [Minimum schema
    -- length] in "Test.Ouroboros.Consensus.ChainGenerator.Honest".
    --
    -- Let phase1 be the set of honest active slots with at least s+d+k slots
    -- after them. Since we already require 1<k, the first sai can be active.
    -- But those first s sais must also contain the k+1st honest asai, while
    -- only k are guaranteed. Therefore we also drop the youngest member of
    -- phase1, which then serves as the missing +1 for the remaining members.

    phase1 =
        BV.countActivesInV S.notInverted
      $ C.sliceV
          (C.truncateWin winH $ C.windowSize winH C.- (s + k + d))
          vH

    phase2 = phase1 C.- 1
```

It's awkward that we then add `1` back to `phase2` :D, but the comment already explains that.

facundominguez: Adopted with edits in 0a406d4.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
facundominguez: Removed in 6da11a8

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -105,10 +105,14 @@ data NoSuchHonestChainSchema =
 genHonestRecipe :: QC.Gen HonestRecipe
 genHonestRecipe = sized1 $ \sz -> do
     (Kcp k, Scg s, Delta d) <- genKSD
-    -- 2s slots has at least 2k blocks. But that is ensuring k-1 more blocks
+    -- 2s slots have at least 2k blocks. But that is ensuring k-1 more blocks
     -- than we actually need. Thus it's safe to remove the k-1 last slots.
-    -- Therefore we set for a length of at least 2s - (k - 1).
-    l <- (+ (2*s - (k - 1))) <$> QC.choose (0, 5 * sz)
+    -- Therefore we first set for a length of at least 2s - (k - 1).
```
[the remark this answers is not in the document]
facundominguez: Done in 0579750.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Adversarial.hs
```diff
@@ -742,17 +782,42 @@ withinYS (Delta d) !mbYS !(RI.Race (C.SomeWindow Proxy win)) = case mbYS of
...
-        C.toVar $ R.runSTGen_ g $ C.uniformIndex numChoices
+        -- @uniformIndex n@ yields a value in @[0..n-1]@, we add 1 to the
+        -- argument to account for the possibility of intersecting at the
+        -- genesis block
+        C.toVar $ R.runSTGen_ g $ C.uniformIndex (possibleBlocks C.+ 1)
   where
     ChainSchema _slots v = schedH
 
-    numChoices = actives C.- k
+    -- 'H.uniformTheHonestChain' ensures there is at least one block in the
```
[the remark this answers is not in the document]
facundominguez: Resolved in 0a406d4.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
facundominguez: > we should enforce it in checkHonestRecipe

Done in 9b0a6ee.

> How about the following?

This would be a bit harder to integrate. But I adopted a simpler diagram in 43ac3c9f and made a pass to clarify and relate the explanations.

The main difficulty I find with just switching to your version verbatim is that it starts with a few definitions and arguments without conveying the purpose of them other than the title.

Since I don't have much hints on what makes it better than the current explanation, I'm also limited to iterate effectively on the text. Happy to discuss it further, though, if you think it is worth the effort.

facundominguez:
Merged as 42bc271 in `genesis/milestone-7-9-10`.

PR #891, opened 2024-01-18
Discussion PR on chain order transitivity

amesgen:
[Rendered](https://github.com/IntersectMBO/ouroboros-consensus/blob/amesgen/order-discussion/docs/ChainOrderTransitivity.md)

### DIFF
--- a/docs/ChainOrderTransitivity.md
+++ b/docs/ChainOrderTransitivity.md
@@ -4,7 +4,7 @@
 
 The abstract Consensus layer currently implements the chain order like this:
 From every header, one can extract a `SelectView`, which has an `Ord` instance.
-A chain `C` is _preferrable_ to a chain `C'` if `selectView(H) > selectView(C')` where `C` and `C'` are the tip headers of `C` and `C'`, respectively.
+A chain `C` is _preferable_ to a chain `C'` if `selectView(C) > selectView(C')` where `C` and `C'` are the tip headers of `C` and `C'`, respectively.
 
 ## Examples for intransitive orders
 
--- a/docs/ChainOrderTransitivity.md
+++ b/docs/ChainOrderTransitivity.md
@@ -42,6 +42,8 @@ So we have `A < B < C < A < B < C < ...`.
 
 ### "Duncan's rule" is non-transitive
 
+For context, see [this comment](https://github.com/IntersectMBO/ouroboros-network/issues/2913#issuecomment-816953407) and [ouroboros-consensus#524](https://github.com/IntersectMBO/ouroboros-consensus/issues/524).
+
 Consider the following alternative rule, for simplicity without the issuer/opcert rule as it is not relevant for the point:
 
  1. Chain length, the longer chain is preferred.
@@ -145,7 +147,7 @@ Remarks:
 
 ### Regarding Duncan's rule
 
-Proposed chain order with similar properties as Duncan's rule.
+We propose the following chain order with similar properties as Duncan's rule:
 
  1. Chain length, preferring longer chains.
  2. Slot number, preferring earlier slots.
@@ -154,12 +156,16 @@ Proposed chain order with similar properties as Duncan's rule.
 
 As this is a lexicographic combination of total orders, it is a total order, so in particular transitive.
 
- - SPOs can't influence the slots they are elected, just their VRF tiebreakers, so it is still random.
- - Arrival times don't matter, so we keep the VRF benefit of discouraging centralization.
- - If a node forges a block `B`, and another block forges another block several (eg `Δ = 5` slots) later, but does not extend `B`, then we will now prefer `B`.
-   This was one of the motivations of Duncan's tiebreaker, compared to the status quo.
+A few remarks about this rule:
 
-TODO elaborate on properties a bit more
+ - SPOs can't influence in which slots they are elected, just as they can't influence their VRF tiebreakers.
+
+   For example, if node `X` forges block `B` in slot `s` and node `X'` forges `B'` in slot `s+1` with `blockNo(B) = blockNo(B')`, then with both Duncan's rule and the status quo, the VRF tiebreaker would decide probabilistically who wins the "battle" and will hence likely be extended by future blocks.
+   With the rule proposed here, however, `B` will always win against `B'`. Note that any node will on average end up in the role of `X` just as often as in that of `X'`.
+ - The proposed rule has the potential advantage of favoring blocks that (due to their earlier slot) had more time to diffuse through the network and hence reach the next block issuer.
+ - Arrival times don't matter, so we keep the benefit of discouraging centralization that using VRFs as the sole tiebreaker (apart from opcert numbers) had.
+ - If a node forges a block `B`, and another block forges another block several (eg `Δ = 5` slots) later, but does not extend `B` as it should have, then we will now prefer `B`.
+   This was one of the motivations of Duncan's tiebreaker, compared to the status quo.
 
 [^vrf-tpraos-vs-praos]: TPraos used the leader VRF, while Praos uses the VRF prior to range extension, see [ouroboros-network#4051](https://github.com/IntersectMBO/ouroboros-network/issues/4051), but this shouldn't matter for this discussion.
 
--- a/docs/ChainOrderTransitivity.md
+++ b/docs/ChainOrderTransitivity.md
@@ -113,12 +113,12 @@ The motivation for the opcert number tiebreaker is described in the ["Design Spe
 We describe our understanding of this use case for the opcert number tiebreaker:
 
 Consider an attacker who corrupts an SPO's node with a specific issuer identity, including the hot key, but not the cold key.
-They can now forge blocks on behalf of this issuer identity, deciding which transactions are included in the block, potentially spoiling the reputation of the SPO in the process.
+They can now mint blocks on behalf of this issuer identity, deciding which transactions are included in the block, potentially spoiling the reputation of the SPO in the process.
 (Rewards still go to the account of the issuer identity, which the attacker can not access.)
 The attacked SPO can now issue a new hot key with a higher opcert number, and start a new node with it installed.
-Now, in every slot where this issuer identity is elected, both the attacker and the attacked SPO can forge a block.
+Now, in every slot where this issuer identity is elected, both the attacker and the attacked SPO can mint a block.
 
- - If the SPO forges a strictly longer or shorter chain, the opcert tiebreaker does not matter, citing from the aforementioned specification:
+ - If the SPO mints a strictly longer or shorter chain, the opcert tiebreaker does not matter, citing from the aforementioned specification:
    > Note that nodes take the precedence amongst operational key certificates into account only after comparing the length of the chains.
    > When the node is already up to date and receives two conflicting blocks that add to its current chain, the length will of course always be the same.
    > But this rule is important: if we did not compare the lengths of the chains before giving preference to the block with the newer operational certificate, it would be possible to force a node to do a rollback of arbitrary length, by sending it a block from a past slot, signed using a newer certificate than the block that the node already has in its chain for that slot.
@@ -130,7 +130,7 @@ An important observation is that the VRF is the same for both blocks, as the VRF
 This means _swapping_ the order of the opcert number and the VRF tiebreaker also satisfies the requirements of the opcert number mechanism as just described (again, this _was_ the original implementation before [ouroboros-network#2348](https://github.com/IntersectMBO/ouroboros-network/pull/2348)).
 
 Note that there is a difference: With the current chain order, a block `B` with a higher opcert number will be preferred over an attacker's block `B'` with the same block number even if it is in a different slot, as opposed to a random tiebreak using the VRF when using the proposed order with these two tiebreakers swapped to restore transitivity.
-However, this difference is irrelevant since the scenario implies that one party can just forge in the larger of the two slots on top of the other block, superseding the tiebreaker due to having a longer chain.
+However, this difference is irrelevant since the scenario implies that one party can just mint in the larger of the two slots on top of the other block, superseding the tiebreaker due to having a longer chain.
 
 Remarks:
  - Given that the VRF also has all the nice properties (in particular [collision resistance][vrf-rfc-collision]) of a cryptographic hash, we could also elide the check that the issuers are the same.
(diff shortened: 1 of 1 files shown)


docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
+# Transitivity of the chain order
+
+## Context
+
+The abstract Consensus layer currently implements the chain order like this:
+From every header, one can extract a `SelectView`, which has an `Ord` instance.
+A chain `C` is _preferable_ to a chain `C'` if `selectView(C) > selectView(C')` where `C` and `C'` are the tip headers of `C` and `C'`, respectively.
```
nfrisby, line 7: Reading this again, I'm guessing you wanted the middle 4 Cs to be Hs?

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Examples for intransitive orders
```
nfrisby, line 9: It seems "intransitive" sometimes means something more specific. So I've been striving to write "not transitive", which is awkward :(

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Current chain order is not transitive
+
+For all Shelley-based eras, we currently use a lexicographic combination of the following comparisons in descending order:
```
nfrisby, line 13: Part of me doubts whether it's fair to call it lexicographic with the conditional in there.

It's seems fairer to say "we either use lexicographic (length, issue number, VRF) or lexicographic (length, VRF)", eg. I dunno, maybe I'm splitting hairs.

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+ 1. Chain length, preferring longer chains.
+ 2. If the issuer identity is the same, compare by the issue/opcert number, preferring higher values, otherwise, no preference.
+ 3. VRF tiebreaker[^vrf-tpraos-vs-praos], preferring lower values.
+
+This is not a total order: It is is reflexive, antisymmetric and total, but not transitive.
```
nfrisby, line 19: is is

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+Consider the following three `SelectView`s:
+
+|              | A | B | C |
+| ------------ | - | - | - |
+| Chain length | l | l | l |
+| Issuer       | X | Y | X |
+| Opcert no    | 2 | o | 1 |
+| VRF          | 3 | 2 | 1 |
+
+Lower-case letters stand for arbitrary values (but if they occur repeatedly, they designate the same value).
```
nfrisby, line 30: Did you mean for X and Y to be lower case?

docs/ChainOrderTransitivity.md
nfrisby: Odd to say it's not a "total order" and then call it "total". Maybe writing "the relation is reflexive..." instead of "it is reflexive..." would help.

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+## History
+
+The relevant changes to the chain order occurred in these PRs in chronological order:
+
+ - [ouroboros-network#2108](https://github.com/IntersectMBO/ouroboros-network/pull/2108) added the opcert number comparison step as the first tiebreaker (before that, only chain length was used).
+ - [ouroboros-network#2195](https://github.com/IntersectMBO/ouroboros-network/pull/2195) added the VRF tiebreaker, but lexicographically *before* the opcert number tiebreaker, in contrast to the status quo.
+   This means that the order still was total at this point, also see below.
+ - [ouroboros-network#2348](https://github.com/IntersectMBO/ouroboros-network/pull/2348) mostly did a change unrelated to the purpose of this document, but crucially, it swapped the order in which the VRF and opcert number comparison are done, introducing the current non-transitivity.
+
+## Impact of non-transitivity
```
nfrisby, line 70: I'd say, if possible say "not transitive". But if that's too awkward, use exactly one of "intransitive" and "non-transitive". Personally I think "non-transitive" is preferable to "intransitive".

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+Fundamentally, the Praos allows ties to be broken arbitrarily.
```
nfrisby, line 72: the Praos +protocol+

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+      According to some preliminary property tests, however, this is not the case for our particular non-transitive order, but it is certainly a weird state of affairs.
+      For example, do we have to worry that GHC changes its implementation of `sortBy` in a way that interacts non-trivially with our non-transitive order?
+    - Also, the reliance on `sortBy` is not crucial, we see several ways how to change our code such that it would do something well-defined even for non-transitive tiebreakers.
+ - Reasoning-related.
+    - Having a non-transitive order could turn out to be very confusing for reasoning about anything that is related to the chain order.
+      This is because transitivity is so ingrained when thinking about orders that it is hard to not implicitly assume.
+    - Certain "obvious" properties suddenly fail to hold, for example:
+
+      > Repeatedly observing the selection of a node over a period of time yields a sequence that is strictly improving for every two adjacent selections.
```
nfrisby, line 92: I'd say "adjacent observations" instead of "adjacent selections", to emphasize the subsequence aspect of it?

nfrisby: I think you'll want to use at least one of the words "subsequence" or "sample" in order to emphasize that the observations do not necessarily include every block the node selected during that time.

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+
+ - If the SPO forges a strictly longer or shorter chain, the opcert tiebreaker does not matter, citing from the aforementioned specification:
+   > Note that nodes take the precedence amongst operational key certificates into account only after comparing the length of the chains.
+   > When the node is already up to date and receives two conflicting blocks that add to its current chain, the length will of course always be the same.
+   > But this rule is important: if we did not compare the lengths of the chains before giving preference to the block with the newer operational certificate, it would be possible to force a node to do a rollback of arbitrary length, by sending it a block from a past slot, signed using a newer certificate than the block that the node already has in its chain for that slot.
+   > This would open up an attack where a stake pool operator could force nodes to do arbitrary rollbacks.
+ - If the SPO forges on top of a chain that is just as long as the one the attacker forged on, the opcert number tiebreaker gives precedence to the SPO as they have the higher opcert number.
```
nfrisby, line 119: "If the SPO and attacker instead mint chains of the same length," seems more direct

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+   Note that this is the usual case when two nodes are elected in the same slot.
+
+An important observation is that the VRF is the same for both blocks, as the VRF does not depend on the hot key.
+This means _swapping_ the order of the opcert number and the VRF tiebreaker also satisfies the requirements of the opcert number mechanism as just described (again, this _was_ the original implementation before [ouroboros-network#2348](https://github.com/IntersectMBO/ouroboros-network/pull/2348)).
+
+Note that there is a difference: With the current chain order, a block `B` with a higher opcert number will be preferred over an attacker's block `B'` with the same block number even if it is in a different slot, as opposed to a random tiebreak using the VRF when using the proposed order with these two tiebreakers swapped to restore transitivity.
+However, this difference is irrelevant as one party can just forge in the larger of the two slots on top of the other block, superseding the tiebreaker due to having a longer chain.
```
nfrisby, line 126: "is irrelevant +since the scenario implies that+ one party can just..."

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+   > This would open up an attack where a stake pool operator could force nodes to do arbitrary rollbacks.
+ - If the SPO forges on top of a chain that is just as long as the one the attacker forged on, the opcert number tiebreaker gives precedence to the SPO as they have the higher opcert number.
+   Note that this is the usual case when two nodes are elected in the same slot.
+
+An important observation is that the VRF is the same for both blocks, as the VRF does not depend on the hot key.
+This means _swapping_ the order of the opcert number and the VRF tiebreaker also satisfies the requirements of the opcert number mechanism as just described (again, this _was_ the original implementation before [ouroboros-network#2348](https://github.com/IntersectMBO/ouroboros-network/pull/2348)).
+
+Note that there is a difference: With the current chain order, a block `B` with a higher opcert number will be preferred over an attacker's block `B'` with the same block number even if it is in a different slot, as opposed to a random tiebreak using the VRF when using the proposed order with these two tiebreakers swapped to restore transitivity.
```
nfrisby, line 125: "even if it is in a different slot +(despite being the same length)+"

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,153 @@
...
+However, this difference is irrelevant as one party can just forge in the larger of the two slots on top of the other block, superseding the tiebreaker due to having a longer chain.
+
+Remarks:
+ - Given that the tiebreaker also has all the nice properties of a cryptographic hash, we could also elide the check that the issuers are the same.
```
nfrisby, line 129: "Give that the ~tiebreaker~ +VRF+"

docs/ChainOrderTransitivity.md
amesgen: Yeah, apparently "intransitive" sometimes means "antitransitive": https://en.wikipedia.org/wiki/Intransitivity#Antitransitivity

I think I will use "non-transitive" everywhere, in particular as it has a noun ("non-transitivity") compared to "not transitive" :+1:

docs/ChainOrderTransitivity.md
amesgen: https://github.com/IntersectMBO/ouroboros-consensus/pull/891#discussion_r1458775301

docs/ChainOrderTransitivity.md
amesgen: I will rephrase and only allude to lexicographic combination :+1:

docs/ChainOrderTransitivity.md
amesgen: Rephrased and added an example :+1:

docs/ChainOrderTransitivity.md
amesgen: Wouldn't this lead to stating the same thing (same block number/length almost immediately next to each other)?

> with the same block number even if it is in a different slot despite being the same length

nfrisby: Oh dear. I had missed the "with the same block number" when I was reading.

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,175 @@
...
+
+   For example, if node `X` mints block `B` in slot `s` and node `X'` mints `B'` in slot `s+1` with `blockNo(B) = blockNo(B')`, then with both Duncan's rule and the status quo, the VRF tiebreaker would decide probabilistically who wins the "battle" and will hence likely be extended by future blocks.
+   With the rule proposed here, however, `B` will always win against `B'`. Note that any node will on average end up in the role of `X` just as often as in that of `X'`.
+ - The proposed rule has the potential advantage of favoring blocks that (due to their earlier slot) had more time to diffuse through the network and hence reach the next block issuer.
+ - Arrival times don't matter (as long as all blocks arrive at the next block issuer before they mint), so we keep the benefit of discouraging centralization that using VRFs as the sole tiebreaker (apart from opcert numbers) had.
+ - If a node mints a block `B`, and another block mints another block several (eg `Δ = 5` slots) later, but does not extend `B` as it should have (empirically, this seems to be due to underresourced nodes/relays), then we will now prefer `B`.
```
nfrisby, line 167: "and another ~block~ node mints another block"

nfrisby (approved):
Approved. Nice!

nfrisby:
I'm currently thinking that this document cannot be significantly improved until we discuss it with other architects/IOG Researchers and decide on a path forward. Specifically, the `SlotNo` tiebreaker.

Regarding the `SlotNo` tiebreaker, I think it'd be useful to cite the Tse paper and [straightpool's Issue](https://github.com/IntersectMBO/ouroboros-network/issues/2913).

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,179 @@
+# Transitivity of the chain order
+
+Remember that a relation `~` is _transitive_ if `a ~ b` and `b ~ c` implies `a ~ c` for all `a,b,c`.
+It is _non-transitive_ if this is not true, ie if there exist `a,b,c` such that `a ~ b` and `b ~ c`, but not `a ~ c`.
+
+## Context
+
+The abstract Consensus layer currently implements the chain order like this:
+From every header, one can extract a `SelectView`, which has an `Ord` instance.
+A chain `C` is _preferable_ to a chain `C'` if `C'` is empty or if `selectView(H) > selectView(H')` where `H` and `H'` are the tip headers of `C` and `C'`, respectively.
```
jorisdral, line 10: Are we talking here about full chains? Or chain fragments?

docs/ChainOrderTransitivity.md
```diff
@@ -0,0 +1,179 @@
...
+Consider the following three `SelectView`s:
+
+|              | A | B | C |
+| ------------ | - | - | - |
+| Chain length | l | l | l |
+| Issuer       | x | y | x |
+| Opcert no    | 2 | o | 1 |
+| VRF          | 3 | 2 | 1 |
+
+Lower-case letters stand for arbitrary values (but if they occur repeatedly, they designate the same value).
```
jorisdral, line 33: "... and distinct variables designate non-equal values"

amesgen: Thanks, fixed:

> Lower-case letters stand for arbitrary values (two letters designate the same value if and only if they are same letter).

amesgen:
The approach in #1063 allows to keep a total (in particular transitive) chain order for eg sorting while still allowing the flexibility to implement the restricted VRF tiebreaker ("Duncan's rule"), which resolves the issue.

PR #896, opened 2024-01-19
Enter Limit on Patience

Niols:
Notes:

- ~This PR targets `main` because of https://github.com/IntersectMBO/ouroboros-consensus/pull/823 that recently reworked the ChainSync client fully. It would not have made sense to target `genesis/milestone8` to then have a big conflict. This does mean that we will have to be mindful when creating `genesis/milestone7` (yay numbering) in February. Or maybe we should rebase `genesis/milestone8` on a more recent version of `main`?~

- The commits can be read individually.

- ~For now, the bucket is unconditional: it starts when the ChainSync client starts and never ever stops. It would make sense to introduce a mechanism to disable it, which could be useful for tests that do not want to bother with LoP and for when we're caught-up. The `LeakyBucket` module already paves the way to add interactions with the bucket (stop, resume, etc.) by providing a `Handler` type that can for now only refill things.~

- ~We have discussed in sync today where I should hook into, so I'm fairly confident about that. I am not confident at all about my approach to configuration, however.~

- There will be the need for discussions on how the bucket should behave when going from Syncing to CaughtUp back to Syncing. Should we keep the previous level? Should we make it full again? Should we slowly fill it back in?

- There will be the need for discussions on how to choose the capacity and the rate to make it interesting in our Genesis tests.

- The LeakyBucket has for now been written with the idea that it should be leaking and do something when empty. We could modify it every so slightly to also handle the case where the bucket fills up and prevents actions from happening when full. This would allow reuse it later (and, who knows, maybe we want to slowly fill up the bucket when caught up).

- ~The LeakyBucket comes with a fairly extensive test suite, I'd say. There is however no LoP-specific tests of the ChainSync client at this point. I have had to disable it because it was killing peers in the other tests (a good sign imo) and I have tested a previous version based on `genesis/milestone8` and it seemed to behave fine.~

- ~In our Genesis tests, it seems the `EmptyBucket` exception gets wrapped in a `ExceptionInLinkedThread` and I don't know where it comes from but it bugs me. We were rather careful to unwrap such things in the bucket and there is even a test for that so I don't think that it comes from this; probably from something in our peer simulator?~

### DIFF
--- a/ouroboros-consensus-diffusion/ouroboros-consensus-diffusion.cabal
+++ b/ouroboros-consensus-diffusion/ouroboros-consensus-diffusion.cabal
@@ -220,6 +220,7 @@ test-suite consensus-test
     Test.Consensus.Genesis.Tests
     Test.Consensus.Genesis.Tests.LoE
     Test.Consensus.Genesis.Tests.LongRangeAttack
+    Test.Consensus.Genesis.Tests.LoP
     Test.Consensus.Genesis.Tests.Uniform
     Test.Consensus.GSM
     Test.Consensus.GSM.Model
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -540,10 +540,11 @@ mkApps ::
   -> (NodeToNodeVersion -> Codecs blk addrNTN e m bCS bCS bBF bBF bTX bKA bPS)
   -> ByteLimits bCS bBF bTX bKA
   -> m ChainSyncTimeout
+  -> CsClient.ChainSyncLoPBucketConfig
   -> ReportPeerMetrics m (ConnectionId addrNTN)
   -> Handlers m addrNTN blk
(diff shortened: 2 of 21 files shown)


nfrisby:
> 10 and 20 on actual instances.

These are the numbers we guessed at on the call, but they are wrong. The correct numbers are 200 to 300.

https://github.com/IntersectMBO/ouroboros-network/blob/9896eda3600a0ea79025f8b5fa90fedca8fc4b68/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs#L197-L209

https://github.com/IntersectMBO/ouroboros-network/blob/9896eda3600a0ea79025f8b5fa90fedca8fc4b68/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs#L219-L225

Niols:
Ah, I also have 10--20 in mind, but that came from consensus tests (eg. [here](https://github.com/IntersectMBO/ouroboros-consensus/blob/cb4141a67d686e95d441ad17e758fb63480a0111/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs#L415), and also in benchmarks). 200--300 is an impressive range. For now, I will put this PR on hold until we decide on the solution we believe in most.

Niols:
Now builds on top of https://github.com/IntersectMBO/ouroboros-consensus/pull/944

PR #904, opened 2024-01-24
Implement `Electric` and enclose relevant code

jasagredo:
# Description

`Electric` code is one of which only one copy should be running at any given moment. This captures the assumption that there is only one chain selection and there is only one process copying stuff to the imm db.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
@@ -39,10 +39,10 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.Background (
 
 import           Control.Exception (assert)
 import           Control.Monad (forM_, forever, void)
+import           Control.Monad.Trans.Class (lift)
 import           Control.Tracer
 import           Data.Foldable (toList)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (fromMaybe)
 import           Data.Sequence.Strict (StrictSeq (..))
 import qualified Data.Sequence.Strict as Seq
 import           Data.Time.Clock
@@ -68,7 +68,7 @@ import           Ouroboros.Consensus.Storage.ChainDB.Impl.Types
 import qualified Ouroboros.Consensus.Storage.ImmutableDB as ImmutableDB
 import           Ouroboros.Consensus.Storage.LedgerDB (TimeSinceLast (..))
 import qualified Ouroboros.Consensus.Storage.VolatileDB as VolatileDB
-import           Ouroboros.Consensus.Util ((.:))
+import           Ouroboros.Consensus.Util
 import           Ouroboros.Consensus.Util.Condense
 import           Ouroboros.Consensus.Util.Enclose (Enclosing' (..))
 import           Ouroboros.Consensus.Util.IOLike
@@ -93,12 +93,12 @@ launchBgTasks
   -> m ()
 launchBgTasks cdb@CDB{..} replayed = do
     !addBlockThread <- launch "ChainDB.addBlockRunner" $
-      addBlockRunner cdb
+      addBlockRunner cdbChainSelFuse cdb
     gcSchedule <- newGcSchedule
     !gcThread <- launch "ChainDB.gcScheduleRunner" $
       gcScheduleRunner gcSchedule $ garbageCollect cdb
     !copyAndSnapshotThread <- launch "ChainDB.copyAndSnapshotRunner" $
-      copyAndSnapshotRunner cdb gcSchedule replayed
+      copyAndSnapshotRunner cdb gcSchedule replayed cdbCopyFuse
     atomically $ writeTVar cdbKillBgThreads $
       sequence_ [addBlockThread, gcThread, copyAndSnapshotThread]
   where
@@ -120,14 +120,6 @@ launchBgTasks cdb@CDB{..} replayed = do
 --
 -- The 'SlotNo' of the tip of the ImmutableDB after copying the blocks is
 -- returned. This can be used for a garbage collection on the VolatileDB.
---
--- NOTE: this function would not be safe when called multiple times
--- concurrently. To enforce thread-safety, a lock is obtained at the start of
--- this function and released at the end. So in practice, this function can be
--- called multiple times concurrently, but the calls will be serialised.
---
--- NOTE: this function /can/ run concurrently with all other functions, just
--- not with itself.
 copyToImmutableDB ::
      forall m blk.
      ( IOLike m
@@ -137,8 +129,8 @@ copyToImmutableDB ::
      , HasCallStack
      )
   => ChainDbEnv m blk
-  -> m (WithOrigin SlotNo)
-copyToImmutableDB CDB{..} = withCopyLock $ do
+  -> Electric m (WithOrigin SlotNo)
+copyToImmutableDB CDB{..} = electric $ do
     toCopy <- atomically $ do
       curChain <- readTVar cdbChain
       let nbToCopy = max 0 (AF.length curChain - fromIntegral k)
@@ -196,15 +188,6 @@ copyToImmutableDB CDB{..} = withCopyLock $ do
         -- happen if the precondition was satisfied.
         _ -> error "header to remove not on the current chain"
 
-    withCopyLock :: forall a. HasCallStack => m a -> m a
-    withCopyLock = bracket_
-      (fmap mustBeUnlocked $ tryTakeMVar cdbCopyLock)
-      (putMVar cdbCopyLock ())
-
-    mustBeUnlocked :: forall b. HasCallStack => Maybe b -> b
-    mustBeUnlocked = fromMaybe
-                   $ error "copyToImmutableDB running concurrently with itself"
-
 -- | Copy blocks from the VolatileDB to ImmutableDB and take snapshots of the
 -- LgrDB
 --
@@ -243,8 +226,9 @@ copyAndSnapshotRunner
   => ChainDbEnv m blk
   -> GcSchedule m
   -> Word64 -- ^ Number of immutable blocks replayed on ledger DB startup
+  -> Fuse m
   -> m Void
-copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
+copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed fuse =
     if onDiskShouldTakeSnapshot NoSnapshotTakenYet replayed then do
       updateLedgerSnapshots cdb
       now <- getMonotonicTime
@@ -267,7 +251,7 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
       --
       -- This is a synchronous operation: when it returns, the blocks have been
       -- copied to disk (though not flushed, necessarily).
-      copyToImmutableDB cdb >>= scheduleGC'
+      withFuse fuse (copyToImmutableDB cdb) >>= scheduleGC'
 
       now <- getMonotonicTime
       let distance' = distance + numToWrite
@@ -527,21 +511,24 @@ addBlockRunner
      , HasHardForkHistory blk
      , HasCallStack
      )
-  => ChainDbEnv m blk
+  => Fuse m
+  -> ChainDbEnv m blk
   -> m Void
-addBlockRunner cdb@CDB{..} = forever $ do
+addBlockRunner fuse cdb@CDB{..} = forever $ do
     let trace = traceWith cdbTracer . TraceAddBlockEvent
     trace $ PoppedBlockFromQueue RisingEdge
     -- if the `addBlockSync` does not complete because it was killed by an async
     -- exception (or it errored), notify the blocked thread
-    bracketOnError (getBlockToAdd cdbBlocksToAdd)
-                   (\blkToAdd -> atomically $ do
-                     _ <- tryPutTMVar (varBlockWrittenToDisk blkToAdd)
-                                     False
-                     _ <- tryPutTMVar (varBlockProcessed blkToAdd)
-                                     (FailedToAddBlock "Failed to add block synchronously")
-                     closeBlocksToAdd cdbBlocksToAdd)
-                   (\blkToAdd -> do
-                     trace $ PoppedBlockFromQueue $ FallingEdgeWith $
-                             blockRealPoint $ blockToAdd blkToAdd
-                     addBlockSync cdb blkToAdd)
+    withFuse fuse $
+      bracketOnError
+        (lift $ getBlockToAdd cdbBlocksToAdd)
+        (\blkToAdd -> lift $ atomically $ do
+            _ <- tryPutTMVar (varBlockWrittenToDisk blkToAdd)
+                   False
+            _ <- tryPutTMVar (varBlockProcessed blkToAdd)
+                   (FailedToAddBlock "Failed to add block synchronously")
+            closeBlocksToAdd cdbBlocksToAdd)
+        (\blkToAdd -> do
+            lift $ trace $ PoppedBlockFromQueue $ FallingEdgeWith $
+              blockRealPoint $ blockToAdd blkToAdd
+            addBlockSync cdb blkToAdd)
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
@@ -3,11 +3,8 @@
 {-# LANGUAGE LambdaCase           #-}
 {-# LANGUAGE MultiWayIf           #-}
 {-# LANGUAGE NamedFieldPuns       #-}
-{-# LANGUAGE PatternSynonyms      #-}
 {-# LANGUAGE RecordWildCards      #-}
 {-# LANGUAGE ScopedTypeVariables  #-}
-{-# LANGUAGE StandaloneDeriving   #-}
-{-# LANGUAGE TypeApplications     #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 -- | Operations involving chain selection: the initial chain selection and
@@ -78,7 +75,7 @@ import           Ouroboros.Consensus.Storage.ImmutableDB (ImmutableDB)
 import qualified Ouroboros.Consensus.Storage.ImmutableDB as ImmutableDB
 import           Ouroboros.Consensus.Storage.VolatileDB (VolatileDB)
 import qualified Ouroboros.Consensus.Storage.VolatileDB as VolatileDB
-import           Ouroboros.Consensus.Util (eitherToMaybe, whenJust)
+import           Ouroboros.Consensus.Util
 import           Ouroboros.Consensus.Util.AnchoredFragment
 import           Ouroboros.Consensus.Util.Enclose (encloseWith)
 import           Ouroboros.Consensus.Util.IOLike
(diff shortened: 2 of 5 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -265,9 +263,9 @@ addBlockSync
      )
   => ChainDbEnv m blk
   -> BlockToAdd m blk
-  -> m ()
+  -> HapaxProcess m ()
 addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } = do
```
jorisdral, line 267: Couldn't we lift the whole `do` block?

```suggestion
addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } = lift $ do
```

Idem in other places

jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -390,20 +389,20 @@ chainSelectionForFutureBlocks
      , HasHardForkHistory blk
      , HasCallStack
      )
-  => ChainDbEnv m blk -> BlockCache blk -> m (Point blk)
+  => ChainDbEnv m blk -> BlockCache blk -> HapaxProcess m (Point blk)
 chainSelectionForFutureBlocks cdb@CDB{..} blockCache = do
```
jorisdral, line 393:
```suggestion
chainSelectionForFutureBlocks cdb@CDB{..} blockCache = lift $ do
```

jasagredo: Can't, there is a `chainSelectionForBlock` in the middle which doesn't have to be lifted.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,48 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+newtype WithIntactFuse m a = WithIntactFuse (m a)
+  deriving newtype (Functor, Applicative, Monad, MonadThrow, MonadCatch)
+
+instance MonadTrans WithIntactFuse where
+  lift = WithIntactFuse
+
+-- | Declare that an action is not safe to be run concurrently, i.e. two
+-- concurrent accesses should throw an exception.
+blowFuse = WithIntactFuse
+blowFuse :: m a -> WithIntactFuse m a
```
jasagredo, line 408: This is in the wrong order!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,48 @@ fib n = round $ phi ** fromIntegral n / sq5
...
 eitherToMaybe (Left _)  = Nothing
 eitherToMaybe (Right x) = Just x
+
+{-------------------------------------------------------------------------------
+  Fuse code, i.e. regions of code that will throw an exception if accessed
+  concurrently.
+-------------------------------------------------------------------------------}
+
+-- | A WithIntactFuse expression is one of which a single instance can be
+-- evaluated at any given moment in the execution of the program.
```
amesgen, line 398:
```suggestion
-- executed at any given moment in the lifetime of the program.
```
"Evaluation" has an association with pure code for me, but the fuse stuff is specifically about side effects in `IO`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,48 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+
+-- | Run the 'WithIntactFuse' action throwing an exception if it is already
+-- running in a separate thread.
+blowAndResetFuse ::
+     (MonadThrow m, MonadMVar m)
+  => Fuse m
+  -> WithIntactFuse m a
+  -> m a
+blowAndResetFuse (Fuse name m) (WithIntactFuse io) = do
+    () <- onlyOne =<< tryTakeMVar m
```
amesgen, line 424:
```suggestion
    onlyOne =<< tryTakeMVar m
```
(if this statement would return something other than `()`, we would get a `-Wunused-do-bind` warning)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,48 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+    a <- io
+    onlyOne . bool Nothing (Just ()) =<< tryPutMVar m ()
+    pure a
+  where
+    onlyOne = maybe (throwIO $ FuseCodeAccessedConcurrently name) pure
```
amesgen, line 429: As this is currently written without `bracket`, `tryPutMVar` does not happen when `io` throws. In fact, *any* subsequent `blowAndResetFuse` with the same `Fuse` will then later fail. I think this is fine for our use case, but it might be at least worth to document this behavior as it is a bit subtle.

Basically: A `Fuse` (as of the state of the PR right now) is not only about making sure that a piece of code is not run concurrently, but also about making sure that it will always fail if it failed once.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -265,8 +262,9 @@ addBlockSync
      )
   => ChainDbEnv m blk
   -> BlockToAdd m blk
+  -> Fuse m
   -> m ()
-addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } = do
+addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } hl = do
```
amesgen, line 267: What is the rationale here for now having one single big `blowAndResetFuse`?

jasagredo: Im fine either way. Joris seemed to prefer fewer lifts.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
jasagredo: Good point!

jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
jorisdral: I'm confused -- aren't there multiple `blowAndResetFuse`s here? @amesgen  did you mean why it is /not/ a big `blowAndResetFuse`?

amesgen: (comment now edited, thanks for pointing this out)

jasagredo: Agreed on having only one big `blockAndResetFuse`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -286,22 +283,22 @@ addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } = do
...
-        deliverWrittenToDisk False
+        lift $ traceWith addBlockTracer $ IgnoreInvalidBlock (blockRealPoint b) reason
+        lift $ deliverWrittenToDisk False
 
         -- We wouldn't know the block is invalid if its prefix was invalid,
         -- hence 'InvalidBlockPunishment.BlockItself'.
-        InvalidBlockPunishment.enact
+        lift $ InvalidBlockPunishment.enact
           blockPunish
           InvalidBlockPunishment.BlockItself
```
amesgen, line 303: Could also move the `lift` before the `do` here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
amesgen: FTR:

> In the status quo, with multiple `blowAndResetFuse`, I immediately start thinking about scenarios like: "Could two invocations of `addBlockSync` run in parallel and be interleaved in such a way that the fuse is not blown"

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
jasagredo: Ah indeed

jasagredo: Ah no we can't! there is a `chainSelectionForFutureBlocks` below!

amesgen: Ah sorry, I was tricked by exactly these two lines being hidden in the GH UI :laughing:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,51 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+-- not be resetted, and any subsequent attempts to use it will fail.
+blowAndResetFuse ::
+     (MonadThrow m, MonadMVar m)
+  => Fuse m
+  -> WithIntactFuse m a
+  -> m a
+blowAndResetFuse (Fuse name m) (WithIntactFuse io) = do
+    onlyOne =<< tryTakeMVar m
+    a <- io
+    onlyOne . bool Nothing (Just ()) =<< tryPutMVar m ()
```
nfrisby, line 429: Minor: I think `bool` is poor for legibility: it's not obvious which argument is which.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,51 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+data Fuse m = Fuse String (MVar m ())
+
+newFuse :: MonadMVar m => String -> m (Fuse m)
+newFuse name = Fuse name <$> newMVar ()
+
+-- | Run the 'WithIntactFuse' action throwing an exception if it is already
+-- running in a separate thread.
+--
+-- Note that if the 'WithIntactFuse' action throws an exception, the fuse will
+-- not be resetted, and any subsequent attempts to use it will fail.
```
nfrisby, line 420: Minor: `resetted` -> `reset`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,51 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+blowFuse :: m a -> WithIntactFuse m a
+blowFuse = WithIntactFuse
+
+-- | Lock required to run a 'WithIntactFuse'
+data Fuse m = Fuse String (MVar m ())
+
+newFuse :: MonadMVar m => String -> m (Fuse m)
+newFuse name = Fuse name <$> newMVar ()
+
+-- | Run the 'WithIntactFuse' action throwing an exception if it is already
```
nfrisby, line 416: Minor: `action throwing` -> `action, but instead throw`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +388,51 @@ fib n = round $ phi ** fromIntegral n / sq5
 eitherToMaybe :: Either a b -> Maybe b
 eitherToMaybe (Left _)  = Nothing
 eitherToMaybe (Right x) = Just x
+
+{-------------------------------------------------------------------------------
+  Fuse code, i.e. regions of code that will throw an exception if accessed
+  concurrently.
+-------------------------------------------------------------------------------}
+
+-- | A WithIntactFuse expression is one of which a single instance can be
```
nfrisby, line 397: I was going to suggest:

> All of these `lift`s are annoying, would a simple `withFuse :: Fuse m -> m a -> m a` combinator suffice?

Technically, it would, so `Fuse m -> ...` could replace `WithIntactFuse (...)`. However, I think without the newtype wrapper it'd be far too easy to make the mistake in which recursive calls also try to claim the fuse :/

It might be worth adding a comment explaining that this "obvious" tempting simplification was intentionally dismissed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
nfrisby: Also, I think we should throw a distinct exception when the write fails. Because that means some other code somehow entered when it shouldn't have, which would be debugged differently?

nfrisby: Also, in my opinion, it'd be easier to grok the implementation if you inlined `onlyOne` and simplified down to something like:

```
tryTakeMVar m >>= \case
  Nothing -> throwIO $ ...
  Just () -> pure ()

...
  
tryPutMVar m () >>= \b -> unless b $ throwIO $ ...
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -63,10 +66,21 @@ module Ouroboros.Consensus.Util (
     -- * Miscellaneous
   , eitherToMaybe
   , fib
+    -- * Fuse code
+  , Fuse
+  , FuseCodeAccessedConcurrently (..)
+  , WithIntactFuse
```
nfrisby, line 72: This comment is about naming, so take a deep breath :D

I propose these names:

```
-- | A simple semaphore, though instead of blocking a fatal exception is thrown.
newFuse :: String -> m (Fuse m)

-- | An action that cannot be ran without drawing current through a 'Fuse'.
Electric :: (Type -> Type) -> Type -> Type

-- | Too much electrical lode was put on the 'Fuse', see 'drawCurrent'.
FuseBlownException :: Type

-- | See 'Electric'
electric :: m a -> Electric m a

-- | Put full load on the 'Fuse' while the 'Electric' is running.
--
-- Thus any two 'withFuse' calls with the same 'Fuse' will throw one fatal exception.
--
-- NOTE The metaphor is: when I run at most one waffle iron concurrently, my kitchen's fuse doesn't blow. But it blows if I run more than one waffle iron concurrently.
--
-- WARNING If the given action throws its own exception, then it will never stop putting load on the 'Fuse'.
withFuse :: Fuse m -> Electric m a -> m a
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -228,12 +228,6 @@ data ChainDbEnv m blk = CDB
     -- not when hashes are garbage-collected from the map.
   , cdbNextIteratorKey :: !(StrictTVar m IteratorKey)
   , cdbNextFollowerKey :: !(StrictTVar m FollowerKey)
-  , cdbCopyLock        :: !(StrictMVar m ())
```
nfrisby, line 231: I think you do want to keep the fuses in this data structure. Otherwise, someone might call `launchBgTasks` more than once on the same `ChainDB`, defeating the whole purpose of the fuse, right?

And, in general, it'd be fine for a program to have two `ChainDB`s open, as long as they were separate `ChainDB`s. And so this data structure is exactly where the fuses should live.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -527,21 +513,24 @@ addBlockRunner
      , HasHardForkHistory blk
      , HasCallStack
      )
-  => ChainDbEnv m blk
+  => Fuse m
+  -> ChainDbEnv m blk
   -> m Void
-addBlockRunner cdb@CDB{..} = forever $ do
+addBlockRunner hl cdb@CDB{..} = forever $ do
```
nfrisby, line 519: Why `hl`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -243,8 +228,9 @@ copyAndSnapshotRunner
   => ChainDbEnv m blk
   -> GcSchedule m
   -> Word64 -- ^ Number of immutable blocks replayed on ledger DB startup
+  -> Fuse m
   -> m Void
-copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
+copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed hl =
```
nfrisby, line 233: Why `hl`?

nfrisby (changes requested):
I Requested Changes for two reasons:

- The fuses should still (like `cdbCopyLock`) live in the ChainDB, instead of being allocated once per `launchBgTasks` call.

- I think the metaphor of a "blown fuse" is being used incorrectly. Therefore one of my comments suggests different names. 🚲 🏠 🎨

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -374,3 +394,59 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+  -> m a
+withFuse (Fuse name m) (Electric io) = do
+  tryTakeMVar m >>= \case
+    Nothing -> throwIO $ FuseBlownException name
+    Just () -> pure ()
+  a <- io
+  tryPutMVar m () >>= \b -> unless b $ throwIO $ FuseBlownException name
+  pure a
+
+-- | Too much electrical lode was put on the 'Fuse', see 'withFuse'.
```
nfrisby, line 449: `lode` -> `load`.

nfrisby (approved):
Approved! 👏 nice.

One minor typo, see my new GitHub comment below.

Issue #328, opened 2023-08-28
Long-term plan for having more than 24 HFC eras


nfrisby:
This is just stub --- need to double-check if there is already a better Issue for this.

https://github.com/input-output-hk/ouroboros-consensus/blob/4c0a421d187478f9f740d27521b682988d843596/sop-extras/src/Data/SOP/Index.hs#L173-L182

Issue #325, opened 2022-10-26
Stop checking maxMajorPV in the header envelopeCheck


nfrisby:
(This tech debt issue is only on the now unused Jira, so I'm creating this Issue instead.)

See https://github.com/input-output-hk/ouroboros-network/pull/3891#discussion_r919388102 for more context.

Issue #324, opened 2023-08-27
Simplify the `protVer` arguments


nfrisby:
Without loss of generality, focus on Allegra. Start from the binding of `protVerAllegra` in `protocolInfoCardano`. EG at https://github.com/input-output-hk/ouroboros-consensus/blob/0079ee65f6c49d1c05a8d30ffa492b20c46b12b8/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L658

There, `protVerAllegra`:

- (A) is passed to `mkShelleyBlockConfig`
- (B) influences `maxMajorProtVer`

Also, `maxMajorProtVer`:

- (C) is passed to `tpraosParams` and used to construct `praosParams`, to eventually be an upper bound in the `ProtocolHeaderSupportsEnvelope` `envelopeChecks` method (to throw the "obsolete node" errors).
- (D) is passed to `mkPartialLedgerConfigShelley` which eventually uses it to constructor `SL.Globals`

```
mkShelleyBlockConfig ::
     ShelleyBasedEra era
  => SL.ProtVer
  -> SL.ShelleyGenesis (EraCrypto era)
  -> [SL.VKey 'SL.BlockIssuer (EraCrypto era)]
  -> BlockConfig (ShelleyBlock proto era)
mkShelleyBlockConfig protVer genesis blockIssuerVKeys = ShelleyConfig {
      shelleyProtocolVersion  = protVer
    , shelleySystemStart      = SystemStart  $ SL.sgSystemStart  genesis
    , shelleyNetworkMagic     = NetworkMagic $ SL.sgNetworkMagic genesis
    , shelleyBlockIssuerVKeys = Map.fromList
        [ (SL.hashKey k, k)
        | k <- blockIssuerVKeys
        ]
    }
```

```
data instance BlockConfig (ShelleyBlock proto era) = ShelleyConfig {
      -- | The highest protocol version this node supports. It will be stored
      -- the headers of produced blocks.
      shelleyProtocolVersion  :: !SL.ProtVer
```

```
-- | Praos parameters that are node independent
data PraosParams = PraosParams
    ...
    -- | All blocks invalid after this protocol version, see
    -- 'Globals.maxMajorPV'.
    praosMaxMajorPV        :: !MaxMajorProtVer,
```

```
data Globals = Globals
  ...
  , maxMajorPV :: !Version
  -- ^ All blocks invalid after this protocol version
```

[...]

PR #913, opened 2024-01-24
consensus: remove the EraNodeToNodeVersion sum type

### LINKED ISSUE #873: Remove `HardForkEncoderDisabledEra` from NTN
This Issue description is a stub. (See GitHub Comments below for elaboration.)

The footnote in the description of Issue https://github.com/IntersectMBO/ouroboros-network/issues/4770 suggests that `HardForkEncoderDisabledEra` is unnecessary. Moreover, it might be counterproductive: it seems more useful for the party who needs to update their code (to support the new era) to be the one whose codec fails, resulting in a log message indicating as much.

It's interesting that the opposite is true for NTC queries: we prefer the failure&log message on the encoder side, since they're encoding a value that the negotiated version disallows (ie they're buggy). Perhaps it's "always the client side", instead of "always the encoder"? Or maybe it's just that NTN payloads (headers, blocks, transactions) are outside the scope of the two connected parties (since they alone can't control the whole network's evolution) whereas the NTC payloads (ie queries and their responses) are fully determined by just these two parties.

Edit: I had initially thought `{Byron,Shelley,Cardano}NodeToNodeVersion*` were also unnecessary. However: the codec for headers (at least on-the-wire) is not solely determined by the protocol version. See eg `ByronNodeToNodeVersion1` vs `ByronNodeToNodeVersion2`. Indeed, even blocks and transactions have a degree of freedom between the over-the-wire codec and the storage codec (although one of them has to be chosen as canonical for the hashing).

nfrisby:
Closes #873. The NTN versions do not need to relate negotiated versions to eras, since every payload in those messages will self-identify the era. The guarding of "experimental" era support should be handled by the (independently motivated) max-major prot-ver check.


nfrisby:
This PR is overriding this approach sketched in April 2023: https://github.com/IntersectMBO/ouroboros-consensus/issues/324#issuecomment-1695928343

> The MaxMajorPV mechanism introduced here https://github.com/IntersectMBO/cardano-ledger/commit/441e23bc7541ab0a535e9d24cda1d836c04b64ed is older than the HFC. At a high-level, the HFC renders MaxMajorPV unnecessary: if the node doesn't yet know about the Nth ledger era, then it won't be able to even parse a header in that era. However, because of "intra-era hard forks", it may be that MaxMajorPV is necessary even in the presence of the HFC.
> 
> It seems possible that we could ban intra-era hard forks in the future. Or maybe even implement them as duplicates in the HFC type-level list (but that would be inconvenient in terms of performance, and Issue https://github.com/IntersectMBO/ouroboros-consensus/issues/328, eg). Either would let us remove the MaxMajorPV mechanism, and instead rely only on the HFC for this.

As of this PR, the `MaxMajorPV` mechanism is the _only_ way to prevent a node from using "experimental" eras. That seems fine to me today, if only because I think "banning intra-era hard forks" seems impractical/naive.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
```diff
@@ -371,6 +363,15 @@ data HardForkEncoderException where
   HardForkEncoderFutureEra :: SingleEraInfo blk -> HardForkEncoderException
 
   -- | HFC enabled, but we saw a value from a disabled era
+  --
+  -- This is only thrown by the Node-to-Client encoder. Two nodes' negotiated
```
amesgen, line 367: It is also thrown in the decoder: https://github.com/IntersectMBO/ouroboros-consensus/blob/cdf512c29091d5b66642b025085100a967b6fa3b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs#L108

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
```diff
@@ -371,6 +363,15 @@ data HardForkEncoderException where
...
+  -- version does not constraint how the /distributed/ chain will evolve, so
+  -- the Node-to-Node communication does not need this. The
+  -- 'Ouroboros.Consensus.Protocol.Praos.Common.MaxMajorProtVer' check will
+  -- enforce it appropriately, and incur explanatory log messages on the node
+  -- that needs to be updated in order to handle the latest hard fork.
```
amesgen, line 372: The Haddock link doesn't work here as `MaxMajorProtVer` is only defined in a downstream module (it is rendered as `MaxMajorProtVer` with a non-existing link target). Maybe rephrase to sth like

> If it is desirable to effectively disable certain (eg experimental) eras based on a runtime flag, this needs to be a block-specific mechanism like the @MaxMajorProtVer@-based check in Cardano.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
```diff
@@ -23,6 +23,14 @@ import           Ouroboros.Network.NodeToNode.Version
...
 
 -- | Protocol versioning
+--
+-- IMPORTANT Note that this is entirely independent of the
+-- 'Ouroboros.Consensus.Shelley.Node.TPraos.shelleyProtVer' field et al.
+--
+-- Its primary purpose is to control the details of on-the-wire codecs. And
+-- additionally which queries are allowed, in the case of
+-- 'BlockNodeToClienVersion' (this use is already handled by
+-- 'Ouroboros.Consensus.Shelley.Node.TPraos.shelleyProtVer' in the NTN case).
```
amesgen, line 33: Similar point about non-functional Haddock link (maybe say just `shelleyProtVer`?)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -46,6 +46,45 @@ type instance ShelleyProtocolHeader (Praos c) = Header c
...
+    -- ^ This is a subtle case.
+    --
+    -- This node is explicitly rejecting the header, but the header isn't
+    -- necessarily _directly_ at fault.
+    --
+    -- This rejection specifically happens when the ledger state being used to
+    -- validate this header contains a protocol major version (the first
+    -- 'Version') that exceeds the maximum major protocol version allowed for
+    -- this era this specific node's configuration (the second 'Version'). The
+    -- only thing the header did "wrong" was extend such a ledger state.
```
amesgen, line 58: Maybe say "*ticked* ledger state" as in the unticked one, the major protocol version was not yet incremented.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -46,6 +46,45 @@ type instance ShelleyProtocolHeader (Praos c) = Header c
...
+    -- TODO (cont'd) It's not even obviously that specific ledger
+    -- state's/block's fault, since the protocol version is the consequence of
+    -- on-chain governance. Is it the voters' fault? Is the fault of the first
+    -- block that was after the voting deadline? So "extending the ledger state
+    -- that resulting from the block after the epoch that extended the ancestor
+    -- block that was after the voting deadline that ..." is merely one step
+    -- more removed. And this 'envelopeChecks' approach does avoid the surprise
+    -- (since the rejection doesn't even depend on the block's non-header
+    -- content either) where the header could be validated but its underlying
+    -- block could not.
```
amesgen, line 87: We have an issue for potentially moving this check to check to a better place, we could link to it here: #325

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
```diff
@@ -174,25 +174,57 @@ validateGenesis = first errsToString . SL.validateGenesis
...
+      -- Second, and more importantly, it's passed to the protocol logic. In
+      -- particular, the node's envelope check will begin rejecting all blocks
+      -- (actually, their headers) if the chain moves to a greater protocol
+      -- version. This should never happen in a node that is using up-to-date
+      -- software and config files. Note that the missing software update is
+      -- not necessarily a 'HardForkBlock' era transition: it might be an
+      -- /intra-era hard fork/ (ie conditionals in the ledger rules).
+      --
+      -- Third, it's passed to the ledger rules. TODO Confirm that that is
+      -- vestigial.
```
amesgen, line 195: Could link to https://github.com/IntersectMBO/cardano-ledger/issues/3682

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -257,101 +258,46 @@ pattern CardanoNodeToNodeVersion1 :: BlockNodeToNodeVersion (CardanoBlock c)
...
-      (  EraNodeToNodeEnabled ByronNodeToNodeVersion2
-      :* EraNodeToNodeEnabled ShelleyNodeToNodeVersion1
-      :* EraNodeToNodeEnabled ShelleyNodeToNodeVersion1
-      :* EraNodeToNodeDisabled
-      :* EraNodeToNodeDisabled
-      :* EraNodeToNodeDisabled
-      :* EraNodeToNodeDisabled
-      :* Nil
-      )
+{-# DEPRECATED CardanoNodeToNodeVersion3 "Use the equivalent CardanoNodeToNodeVersion2 instead" #-}
```
amesgen, line 279: There is no occurrence of `CardanoNodeToNode` in cardano-api, cardano-cli or cardano-node, to I think it would also be fine to skip the deprecation cycle here for simplicity, especially as it is mentioned in the changelog.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -46,6 +46,45 @@ type instance ShelleyProtocolHeader (Praos c) = Header c
...
+    --
+    -- This rejection specifically happens when the ledger state being used to
+    -- validate this header contains a protocol major version (the first
+    -- 'Version') that exceeds the maximum major protocol version allowed for
+    -- this era this specific node's configuration (the second 'Version'). The
+    -- only thing the header did "wrong" was extend such a ledger state.
+    --
+    -- Note that the ChainSync client ensures that ledger state is one of the
+    -- latest k+1 ledger states on the node's current chain (modulo STM
+    -- scheduling).
```
amesgen, line 62: Might be slightly confusing as the latest `k+1` ledger states as stored in the LedgerDB are unticked, so they do *not* have a protocol major version that is larger than the max major prot ver.

amesgen (approved):
Looks great, only some comments about the documentation

In addition to #873, this PR also partially addresses #587.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
nfrisby: > downstream module

Did you mean _downstream package_? I've never realized the package hierarchy mattered (😞), but I'm pretty certain the module hierarchy does not.

nfrisby: I'm tempted to leave it as it is. The idea being that even the broken link contains all the info the user would need to find it 🤷 

Moreover, if the Haddock solution improves, then suddenly this markup would suddenly do exactly what we want!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
nfrisby: Same reply: https://github.com/IntersectMBO/ouroboros-consensus/pull/913#discussion_r1466560864

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
nfrisby: Done

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
nfrisby: Done, thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
amesgen: > Did you mean *downstream package*? I've never realized the package hierarchy mattered (😞), but I'm pretty certain the module hierarchy does not.

Yes, I meant downstream *package*.

---

Seems fine to leave this as is, was just briefly confused when I locally built the Haddock and the link went nowhere, but everything will be clear as soon as someone looks at the source :+1:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
nfrisby: Done, thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
nfrisby: Done, thanks!

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: Done, thanks! Yeah, seems fair.

PR #926, opened 2024-01-26
Mempool: trace reason for transaction removal

### LINKED ISSUE #886: Enrich mempool tracers with reason for transaction removal
Currently, when a transaction was at first valid and added to the mempool, but is later removed upon revalidation after the most recent ledger state changed, we do not log the reason for why this happened in `TraceMempoolRemoveTxs`, in contrast to `TraceMempoolRejectedTx`.

https://github.com/IntersectMBO/ouroboros-consensus/blob/399f3dbe8c11ecb057f1cdb737092cea5e4b85c6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs#L528-L534

Several people expressed the desire to be able to see the corresponding reason in the node logs. The goal of this ticket is to enrich the mempool tracers appropriately (eg refining the first argument to `[(GenTx blk, ApplyTxErr blk)]`) and cooperate with the release engineer/the node team to ensure that the tracing system actually uses this new information (potentially guarded behind a severity flag to avoid excessive verbosity).

amesgen:
Closes #886

Also adapts the corresponding mempool test to check that the attached `ApplyTxErr`s are actually those we expect.

### DIFF
--- a/ouroboros-consensus/changelog.d/20240126_122429_alexander.esgen_mempool_remove_trace_reason.md
+++ b/ouroboros-consensus/changelog.d/20240126_122429_alexander.esgen_mempool_remove_trace_reason.md
@@ -1,0 +1,4 @@
+### Breaking
+
+- Mempool: add reason for transaction removal to `TraceMempoolRemoveTxs`. This
+  can be used in the node to enrich the trace output.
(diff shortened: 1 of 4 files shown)


jasagredo (approved):
Nice one

Issue #390, opened 2023-10-02
Add a ShelleyBlock query for which upcoming slots the node will lead


nfrisby:
The Node Team, under our recommendation, has so far used a `Debug*` query in order to build a CLI query that computes the node's upcoming leadership schedule. This is a non-trivial calculation with various pitfalls, and so it should instead be implemented as a proper `ShelleyBlock` query in the `ouroboros-consensus` code base.

The design of such a query is unfortunately related to Issue #389. In Slack discussion with architecture stakeholders, we have already decided that the ideal behavior would be for the node to only predict slots for which the answer could not change under ordinary circumstances (ie a rollback of at most `k` blocks). Thus...

This Issue is to design and implement a new `ShelleyBlock` query that returns the list of slot numbers with their corresponding UTC times such that BOTH this node's configured leadership credentials will be elected in those slots AND ALSO the slots' UTC times and those election predicates will be correct for any extension of the node's current immutable tip---ie the query's result must not be subject to a rollback of at most `k` blocks.

This is the first query with that constraint, other than the HFC queries which already involve a non-trivial mechanism to ensure it.

nfrisby:
Esgen and I just now realized this, in general, may need to be a "cross-era query", which will need to be added to `CanHardFork` etc. Therefore this query should be a hard fork query, not a `ShelleyBlock` query---otherwise the `ShelleyBlock` query would have to somehow know how to compute the leader schedule of the next era, which violates the HFC prinicpal that era's are not responsible for other eras.

amesgen:
Also see #600 for some additional remarks

PR #927, opened 2024-01-26
 Praos: use larger nonce stabilization window starting in Conway 

### LINKED ISSUE #297: Switch Epoch structure from 4-3-3 to 3-3-4
This is important for Genesis work. Ideally it should be done before Conway, however if it can be implemented in isolation within a single rule, then an intra-era hardfork could be possible as well.

Timeline for this is sometime next year (unclear when exactly).

Any questions about this should be directed to Christian Badertscher and Duncan

amesgen:
Closes #297, closes https://github.com/IntersectMBO/cardano-ledger/issues/2478, closes https://github.com/IntersectMBO/cardano-ledger/issues/1914

See erratum 17.3 (amended in https://github.com/IntersectMBO/cardano-ledger/pull/4015) in the Shelley ledger specs for context.

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
@@ -730,7 +730,13 @@ protocolInfoCardano paramsCardano
         praosMaxMajorPV = maxMajorProtVer,
         praosMaxLovelaceSupply = SL.sgMaxLovelaceSupply genesisShelley,
         praosNetworkId = SL.sgNetworkId genesisShelley,
-        praosSystemStart = SystemStart $ SL.sgSystemStart genesisShelley
+        praosSystemStart = SystemStart $ SL.sgSystemStart genesisShelley,
+        praosRandomnessStabilisationWindow =
+          -- This value is used for all Praos eras /except/ Babbage, see
+          -- 'partialConsensusConfigBabbage'.
+          SL.computeRandomnessStabilisationWindow
+            (SL.sgSecurityParam genesisShelley)
+            (SL.mkActiveSlotCoeff $ SL.sgActiveSlotsCoeff genesisShelley)
       }
 
     PraosParams { praosSlotsPerKESPeriod, praosMaxKESEvo } = praosParams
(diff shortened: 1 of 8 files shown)


amesgen:
Note: We should also inform other teams about this change; in particular such that `cardano-cli` commands reporting the leader schedule can be adapted (as they have to implement the logic themselves until something like #390) lands.

jasagredo (approved):
Looks good. The glossary also is understandable.

amesgen:
We should not merge this for the Consensus release that goes into node 8.9 as that release it supposed to be rather minimal (and this PR requires all existing Conway chains to be truncated, in particular Sanchonet).

ouroboros-consensus-protocol/changelog.d/20240206_174209_alexander.esgen_change_epoch_structure.md
```diff
@@ -0,0 +1,4 @@
+### Breaking
+
+- Add `praosRandomnessStabilisationWindow` to `PraosParams`, allowing to
+  configure where the epoch nonce is snapshotted.
```
nfrisby, line 4: where -> when.... or even better, "in which slot in the era"

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -730,7 +730,11 @@ protocolInfoCardano paramsCardano
         praosMaxMajorPV = maxMajorProtVer,
         praosMaxLovelaceSupply = SL.sgMaxLovelaceSupply genesisShelley,
         praosNetworkId = SL.sgNetworkId genesisShelley,
-        praosSystemStart = SystemStart $ SL.sgSystemStart genesisShelley
+        praosSystemStart = SystemStart $ SL.sgSystemStart genesisShelley,
+        praosRandomnessStabilisationWindow =
+          SL.computeRandomnessStabilisationWindow
```
nfrisby, line 735: Maybe add a comment here, at least saying `see 'partialConsensusConfigBabbage'` or something like that

nfrisby (approved):
LGTM

amesgen:
Merging now as the Consensus packages for Node 8.9 were released in #969. (Context: https://github.com/IntersectMBO/ouroboros-consensus/pull/927#issuecomment-1930377325)

PR #930, opened 2024-01-29
Update CODEOWNERS

dnadales:
Now, the Consensus maintainers are the only codeowners.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # Note: later rules override earlier rules.
 
 # Default
-* @IntersectMBO/ouroboros-consensus @IntersectMBO/consensus-superowners
+* @IntersectMBO/ouroboros-consensus-maintainers @IntersectMBO/consensus-superowners


amesgen (approved):

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 # Note: later rules override earlier rules.
 
 # Default
-* @IntersectMBO/ouroboros-consensus @IntersectMBO/consensus-superowners
+* @IntersectMBO/ouroboros-consensus-maintainers @IntersectMBO/consensus-superowners
```
amesgen, line 4: Could even remove `@IntersectMBO/consensus-superowners` as it doesn't seem to exist:

>  Unknown owner on line 4: make sure the team @IntersectMBO/consensus-superowners exists, is publicly visible, and has write access to the repository

amesgen: Thanks to @jasagredo, I learned that this group might be revived in the future, so let's leave it there for now.

PR #931, opened 2024-01-30
Add GHC 9.8.1 to Nix CI

amesgen:
- Add GHC 9.8.1 to the Nix Hydra CI as a follow-up to #814 
 - Build HLS from GitHub instead from Hackage. HLS only builds for GHC 9.8 with a few `allow-newer` ATM, so it is convenient to reuse the upstream `cabal.project` of HLS for this (which is not distributed on Hackage). It also sets an `index-state`, so we don't need to set our `tool-index-state` for it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,13 +57,15 @@
             (import ./nix/pdfs.nix)
           ];
         };
-        hydraJobs = import ./nix/ci.nix pkgs;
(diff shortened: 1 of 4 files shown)


jasagredo:
This is not building `haskell98` in CI? https://ci.iog.io/build/2645543#tabs-constituents

PR #932, opened 2024-01-30
add an LoE updater for the Genesis Density Disconnect method

tek:


ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
```diff
@@ -652,7 +656,8 @@ singleNodeTestConfig :: TopLevelConfig TestBlock
 singleNodeTestConfig = singleNodeTestConfigWithK (SecurityParam 4)
 
 singleNodeTestConfigWithK :: SecurityParam -> TopLevelConfig TestBlock
-singleNodeTestConfigWithK = singleNodeTestConfigWith TestBlockCodecConfig TestBlockStorageConfig
+singleNodeTestConfigWithK k =
+  singleNodeTestConfigWith TestBlockCodecConfig TestBlockStorageConfig k (GenesisWindow (2 * maxRollbacks k))
```
facundominguez, line 660: Why `2 * maxRollbacks k`?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block/BFT.hs
```diff
@@ -106,6 +106,7 @@ instance ( SimpleCrypto c
          ) => LedgerSupportsProtocol (SimpleBftBlock c c') where
   protocolLedgerView   _ _ = ()
   ledgerViewForecastAt _   = trivialForecast
+  computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
facundominguez, line 109: Explain what needs to be done instead?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block/Praos.hs
```diff
@@ -118,6 +118,7 @@ instance ( SimpleCrypto c
          ) => LedgerSupportsProtocol (SimplePraosBlock c c') where
   protocolLedgerView   _ _  = ()
   ledgerViewForecastAt _ st = constantForecastOf () (getTipSlot st)
+  computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
facundominguez, line 121: Explain what needs to be done instead?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block/PraosRule.hs
```diff
@@ -97,7 +97,7 @@ instance
   ) => LedgerSupportsProtocol (SimplePraosRuleBlock c) where
   protocolLedgerView   _ _ = ()
   ledgerViewForecastAt _   = trivialForecast
-
+  computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
facundominguez, line 100: Explain what needs to be done?

ouroboros-consensus/src/unstable-tutorials/Ouroboros/Consensus/Tutorial/Simple.lhs
```diff
@@ -669,6 +669,7 @@ the block to index both the ledger and the protocol.
 > instance LedgerSupportsProtocol BlockC where
 >   protocolLedgerView _lcfg  _tl = ()
 >   ledgerViewForecastAt _lccf = trivialForecast
+>   computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
facundominguez, line 672: Explain what needs to be done?

ouroboros-consensus/src/unstable-tutorials/Ouroboros/Consensus/Tutorial/WithEpoch.lhs
```diff
@@ -632,6 +632,8 @@ ledger view: (1) the slot (`for` in the code below) is in the current epoch and
 >     -- after the following epoch:
 >     maxFor :: SlotNo
 >     maxFor = nextEpochStartSlot at + SlotNo slotsInEpoch
+>
+>   computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
facundominguez, line 636: Explain what needs to be done?

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/Infra.hs
```diff
@@ -111,6 +113,7 @@ genEraParams = do
     eraEpochSize  <- EpochSize         <$> choose (1, 10)
     eraSlotLength <- slotLengthFromSec <$> choose (1, 5)
     eraSafeZone   <- genSafeZone
+    eraGenesisWin <- GenesisWindow     <$> choose (1, 10)
```
facundominguez, line 116: Does it make sense to choose a genesis window without considering values chosen for other parameters?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -254,7 +256,7 @@ genBlockFromLedgerState = pure . genBlock . lastAppliedPoint . ledgerState
 extLedgerDbConfig :: SecurityParam -> LedgerDbCfg (ExtLedgerState TestBlock)
 extLedgerDbConfig secParam = LedgerDbCfg {
       ledgerDbCfgSecParam = secParam
-    , ledgerDbCfg         = ExtLedgerCfg $ singleNodeTestConfigWith TestBlockCodecConfig TestBlockStorageConfig secParam
+    , ledgerDbCfg         = ExtLedgerCfg $ singleNodeTestConfigWith TestBlockCodecConfig TestBlockStorageConfig secParam (GenesisWindow (2 * maxRollbacks secParam))
```
facundominguez, line 259: Why `2 * maxRollbacks secParam`?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/TestBlock.hs
```diff
@@ -675,6 +679,7 @@ mkTestConfig k ChunkSize { chunkCanContainEBB, numRegularBlocks } =
         eraEpochSize  = EpochSize numRegularBlocks
       , eraSlotLength = slotLength
       , eraSafeZone   = HardFork.StandardSafeZone (maxRollbacks k * 2)
+      , eraGenesisWin = GenesisWindow (maxRollbacks k * 2)
```
facundominguez, line 682: Why `maxRollbacks k * 2`?

facundominguez (commented):
Made a first pass. Many of my comments are repetitive, but I think repetition is good to surface all the places that might need an action. There's no need to discuss them all separately.

I think the requests for documentation that would help me the most making progress are in the `Genesis.Governor` module, and a note explaining the interaction of the GDD governor with ChainSync.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -90,4 +105,153 @@ updateLoEFragStall k getCandidates =
...
+              endOfGenesisWindow `min` claimedTip
+            where
+              claimedTip = succWithOrigin $ getTipSlotNo theirTip
+          hasBlockAfterGenesisWindow =
+            -- Note that if
+            -- > NotOrigin s = AF.headSlot candidateSuffix
+            -- this check is equivalent to
+            -- > s >= endOfGenesisWindow
+              succWithOrigin (AF.headSlot candidateSuffix)
+            > endOfGenesisWindow
```
amesgen, line 147: For blocks with a finite forecasting range (like `CardanoBlock`, but unlike `TestBlock` ATM), it can be the case that a peer already sent us a header beyond the Genesis window (equal to the forecasting range), but we don't know about this here as we only update the candidate fragment after we validated the header.

Therefore, we have to enrich the ChainSync client to also indicate whether it received, but could not yet process a header beyond the forecast horizon (ie whether it is "LedgerView-starved").

(cc @nbacquey @Niols as we just talked about this)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -90,4 +105,153 @@ updateLoEFragStall k getCandidates =
...
-      setLoEFrag (sharedCandidatePrefix k curChain candidates)
+      setLoEFrag (fst (sharedCandidatePrefix k curChain candidates))
+
+showPeers ::
+     Show peer
+  => Map peer String
+  -> String
+showPeers = intercalate ", " . fmap (\ (peer, v) -> show peer ++ " -> " ++ v) . Map.toList
+
+densityDisconnect ::
```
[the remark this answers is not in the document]
facundominguez: Here's an attempt to explain it:
```
-- | @densityDisconnect cfg immutableLedgerSt candidateSuffixes theirTips loeFrag@
-- yields the list of peers which are known to lose the density comparison with any
-- other peer, when looking at the genesis window after @loeFrag@.
--
-- The peers are taken from the keys of @candidateSuffixes@.
--
-- @candidateSuffixes@ tells for every peer what is the fragment that the peer
-- proposes to use after the immutable tip.
--
-- @theirTips@ tells for every peer what is the last header that the peer is
-- claiming to have.
--
-- TODO: @loeFrag@, what is this?
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -46,16 +61,16 @@ sharedCandidatePrefix ::
   SecurityParam ->
   AnchoredFragment (Header blk) ->
   Map peer (AnchoredFragment (Header blk)) ->
-  AnchoredFragment (Header blk)
+  (AnchoredFragment (Header blk), Map peer (AnchoredFragment (Header blk)))
 sharedCandidatePrefix (SecurityParam k) curChain candidates =
-  trunc
+  (trunc, suffixes)
```
facundominguez, line 66: Document what `suffixes` is. Also, `trunc` is not just a prefix of `curChain` as the documentation claims. Depending on the value of `excess` some elements might be skipped.

facundominguez: Maybe the documentation of `sharedCandidatePrefix` could say:
```
-- | Compute the fragment loeFrag between the inferred immutable tip and the
-- earliest intersection between @curChain@ and any of the @candidates@.
--
-- The inferred tip might be the anchor of @curChain@, if there are less than
-- k headers before the first intersection, otherwise it will be the kth ancestor
-- of the first intersection.
--
-- The function also yields the suffixes of the intersection of loeFrag with
-- every candidate fragment.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -90,4 +105,153 @@ updateLoEFragStall k getCandidates =
...
+      -- Obtaining these before calling setLoEFrag appeared to result in a wrong state,
+      -- but maybe that was a fluke.
+      theirTips <-
+        fmap (Map.mapMaybe id) $ flip Map.traverseWithKey candidateSuffixes $ \peer _ ->
+          cschTheirTip (handles Map.! peer)
+      pure (candidateSuffixes, handles, theirTips, loeFrag)
+
+    let
+      (losingPeers, boundsDesc) =
+        densityDisconnect cfg immutableLedgerSt candidateSuffixes theirTips loeFrag
```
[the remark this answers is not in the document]
facundominguez: The LoE fragment stays outdated until the next block is added to the chain db. It is still to be justified that this doesn't cause any sort of deadlock because the LoE isn't sufficiently advanced.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
```diff
@@ -0,0 +1,280 @@
...
+movePeer :: EvolvingPeer -> EvolvingPeer
+movePeer = \case
+  peer@EvolvingPeer {candidate, suffix = h : t} -> peer {candidate = candidate AF.:> h, suffix = t}
+  peer -> peer
+
+-- | Repeatedly run the GDD, each time updating a random peer to advance by one block.
+-- The selection is set to the first k blocks of the first fork.
+-- The tips are the last blocks of each full branch.
+-- The returned 'Bool' indicates whether the honest peer won and no monotonicity violations were detected.
+evolveBranches ::
```
tek, line 231: Is this roughly what you had in mind, @amesgen?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -93,4 +114,147 @@ updateLoEFragStall k getCandidates =
...
+     , Ord peer
+     , Show peer
+     , LedgerSupportsProtocol blk
+     )
+  => TopLevelConfig blk
+  -> Tracer m String
+  -> STM m (Map peer (AnchoredFragment (Header blk)))
+  -> STM m (Map peer (ChainSyncClientHandle m blk))
+  -> UpdateLoEFrag m blk
+updateLoEFragGenesis cfg tracer getCandidates getHandles =
```
facundominguez, line 225: At the moment, the LoE fragment is updated every time a block is added to the chain db. 

An open question is how is LoE updated if all the blocks from all the peers have been received already.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
amesgen: Yeah, exactly!

FTR:
> One idea for a GDD test: Test that the GDD logic is "consistent under chain extension":
> For a given set of candidate chains (plus "claims more headers" information), the GDD logic will for every peer either decide to disconnect or to keep them for now ("inconclusive"). Now consider a specific peer. As more and more information about its candidate chain gets available for the GDD logic (ie we receive more extending headers), the GDD might eventually change its judgement from "inconclusive" to "disconnect", but never the other way around. A property test could make sure that the "disconnect"->"inconclusive" judgement state transition is never occurring.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -260,17 +264,18 @@ runPointSchedule ::
...
   m StateView
-runPointSchedule schedulerConfig GenesisTest {gtSecurityParam = k, gtBlockTree} pointSchedule tracer0 =
+runPointSchedule schedulerConfig GenesisTest {gtSecurityParam = k, gtBlockTree, gtGenesisWindow} pointSchedule tracer0 =
   withRegistry $ \registry -> do
     stateViewTracers <- defaultStateViewTracers
     resources <- makePeerSimulatorResources tracer gtBlockTree (pointSchedulePeers pointSchedule)
     let getCandidates = traverse readTVar =<< readTVar (psrCandidates resources)
-        updateLoEFrag = updateLoEFragStall k getCandidates
+        getHandles = readTVar (psrHandles resources)
+        updateLoEFrag = updateLoEFragGenesis config tracer getCandidates getHandles
```
facundominguez, line 273: The documentation of `scEnableLoE` needs to be updated to reflect the fact that we no longer call `updateLoEFragStall`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -510,6 +502,14 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
 
     processLoE curChain (LgrDB.ledgerDbCurrent ledgerDB) (writeTVar cdbLoEFrag) cdbLoE
 
+    loeFrag <- atomically (cross curChain <$> readTVar cdbLoEFrag) <&> \case
+        Just (_, frag) -> frag
+        -- We don't crash if the LoE fragment doesn't intersect with the selection
+        -- because we update the selection _after_ updating the LoE fragment, which
+        -- means it could move to another fork or beyond the end of the LF, depending
+        -- on the implementation of @updateLoEFrag@.
+        Nothing        -> AF.Empty (AF.anchor curChain)
```
facundominguez, line 511: It looks to me like `computeLoEMaxExtra` below should produce `loeLimit + however many blocks are needed to reach the tip of loeFrag`. However, it is computed instead as `loeLimit - budgetAlreadyUsed`.

amesgen: Are you just suggesting a refactoring, or a logic change? In case of the latter: Can you give a concrete example of a situation you think this rule would handle better/correctly in contrast to the existing one?

facundominguez: Suppose `loeFrag`'s tip is the block following `hdr` (the block being added). The current implementation will return `k` as maxExtra, when LoE would allow `k+1`.

amesgen: Right, the current `computeLoEMaxExtra` is too conservative when the `loeFrag` is an extension of the chain we are switching to. We should change that :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -49,16 +70,16 @@ sharedCandidatePrefix ::
...
-  AnchoredFragment (Header blk)
+  (AnchoredFragment (Header blk), Map peer (AnchoredFragment (Header blk)))
 sharedCandidatePrefix (SecurityParam k) curChain candidates =
-  trunc
+  (trunc, suffixes)
   where
     trunc | excess > 0 = snd (AF.splitAt excess shared)
           | otherwise = shared
 
     excess = AF.length shared - fromIntegral k
```
facundominguez, line 80: Splitting the result with `excess` here is causing the loe fragment to not intersect with the current selection.

What is the purpose of trimming the loe fragment like this?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -93,4 +114,147 @@ updateLoEFragStall k getCandidates =
...
+        densityDisconnect sgen (configSecurityParam cfg) candidateSuffixes theirTips loeFrag
+
+    trace ("Density bounds: " ++ boundsDesc)
+    trace ("New candidate tips: " ++ showPeers (showTip <$> Map.map AF.headPoint candidateSuffixes))
+    trace ("Losing peers: " ++ show losingPeers)
+    trace ("Setting loeFrag: " ++ show (AF.headAnchor loeFrag))
+
+    for_ losingPeers $ \peer -> do
+      trace ("Killing peer: " ++ show peer)
+      cschKill (handles Map.! peer)
```
facundominguez, line 255: This kills the ChainSync client and allows the peer to continue sending blocks via BlockFetch. Is not killing `BlockFetch` an omission?

amesgen: The diffusion layer should entirely disconnect from a peer when an exception is thrown in any miniprotocol for that peer, so it should be enough to kill one miniprotocol. Are you observing something to the contrary?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
facundominguez, line 511: `pointHash tipPoint == headerPrevHash hdr` in the checks below is not a sufficient condition to extend the selection. In addition, it should be checked that the new header is a descendent of the tip of the LoE fragment, or it is contained in the LoE fragment. 

Otherwise, if the block belongs to an adversary that has been disconnected, the node will continue to extend the selection on a chain which is served by the adversary.

This raises the question that LoE should determine on which chain the selection is supposed to be, unlike the current code that resets LoE if it is on a different chain.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Yes. In the test "serve adversarial branches" the disconnected adversary still sends blocks.

amesgen: Ah, I guess we don't invoke the actual network diffusion layer, so this is probably something that should be fixed in the Genesis peer simulator infrastructure.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
amesgen: > `pointHash tipPoint == headerPrevHash hdr` in the checks below is not a sufficient condition to extend the selection. In addition, it should be checked that the new header is a descendent of the tip of the LoE fragment, or it is contained in the LoE fragment.
> 
> Otherwise, if the block belongs to an adversary that has been disconnected, the node will continue to extend the selection on a chain which is served by the adversary.

It should be unproblematic/even expected if we are selecting some adversarial blocks after the adversary has been disconnected (eg because the adversary sent them very quickly, and they were still in the ChainSel queue after we disconnected from the adversary, waiting to be processed) as long as we don't select more than `k` blocks past the LoE. `computeLoEMaxExtra` will return `Nothing` if that is not the case, and hence we won't adopt the block.

facundominguez: I think you are correct in essence. I have made `computeLoEMaxExtra` more precise in 716d975.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Edited in e455562.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: I removed the `excess` computation in e455562. I believe the old documentation of the function was trying to justify it, but it was unintelligible to me, and the old behavior wasn't working well with chain selection.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
facundominguez: `computeLoEMaxExtra` has been revised in 716d975.

facundominguez:
There is an interesting failure with
```
cabal run ouroboros-consensus-diffusion:consensus-test -- -p "/serve adversarial branches/"  --quickcheck-tests 1000 --quickcheck-replay=775401 --quickcheck-max-size 3
```
Here's the block tree:
```
       GenesisTest:                                                                                                                                                                                  
          gtHonestAsc: 0.765625                                                                                                                                                                       
          gtSecurityParam: 3                                                                                                                                                                          
          gtGenesisWindow: 5                                                                                                                                                                          
          gtDelay: 0                                                                                                                                                                                  
          gtBlockTree:                                                                                                                                                                                
            slots:    0  1  2  3  4  5  6  7  8  9 10 11 12 13                                                                                                                                        
            trunk:  --1-----2--3--4--5--6--7--8--9----10-11-12                                                                                                                                        
                      `-----2--3-----4-----5-----6--7--8--9-10                                                                                                                                        
                                  `--5-----------------6--7--8
```
The immutable tip stays at slot 0, and `chainSelectionForBlock` is only called once according to tracing (only one occurrence of `ChainSelectionLoEDebug`) despite of the block points advancing multiple times.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -735,12 +745,20 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
+    computeLoEMaxExtra loeLimit loeFrag newBlockFrag =
+        -- Both fragments are on the same chain
+        if loeSuffixLength == 0 || rollback == 0 then
+          if rollback > loeLimit + loeSuffixLength
+            then Nothing
+            else Just $ loeLimit + loeSuffixLength - rollback
+        else
+          if rollback > loeLimit
+            then Nothing
+            else Just $ loeLimit - rollback
```
amesgen, line 762: Currently, we disable the LoE by setting `loeLimit = maxBound`, so we should either refactor that or make sure we don't add to `loeLimit` without prior checks.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: Is what I just pushed coming close to what we need?

amesgen: As suggested by Nick today, it seems much more direct/easier to justify to just store the last `SlotNo` that was received in some `TVar` (either the `SlotNo` by the next header, or the rollback point), and use that here to judge whether it is beyond the Genesis window (instead of the current `AF.headSlot candidateSuffix`).

facundominguez:
Superseded by #984.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: This was addressed in #981.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
facundominguez: The type of loeLimit was made isomorphic to Maybe in be94a46.

facundominguez:
> There is an interesting failure with

We are figuring this in the context of #984. Known causes for this are that
1. the BlockFetch server can never reply a batch request to the client if the server has switched chains.
2. the bfcMaxConcurrencyBulkSync limit in the BlockFetch configuration was set to 1, meaning that while one BlockFetch server was pausing a download waiting on the BP, no other BlockFetch server would get requests to download blocks.
3. all blocks might have been added to the chain DB, yet the selection needs to be updated still after changes to the candidate fragments (done in #972).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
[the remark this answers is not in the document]
facundominguez: Added this description in 7f5ab3f5f.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -90,4 +105,153 @@ updateLoEFragStall k getCandidates =
...
+
+    frag1: A - B - C - D - ...            <- exact
+            \
+    frag2:     E           claimed tip: E <- exact
+
+-}
+
+-- | Run the Genesis disconnection logic once.
+--
+--   * Maintain and update the LoE.
```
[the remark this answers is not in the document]
facundominguez: Addressed in 6974049a.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -152,6 +155,13 @@ newtype Our a = Our { unOur :: a }
   deriving stock   (Eq)
   deriving newtype (Show, NoThunks)
 
+data ChainSyncClientHandle m blk = ChainSyncClientHandle {
+    cschKill     :: !(m ())
+  , cschTheirTip :: !(STM m (Maybe (Tip blk)))
+  }
```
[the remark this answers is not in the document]
facundominguez: Addressed in 1af06312.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -162,32 +172,46 @@ bracketChainSyncClient ::
...
     -- ^ The candidate chains, we need the whole map because we
     -- (de)register nodes (@peer@).
+ -> StrictTVar m (Map peer (ChainSyncClientHandle m blk))
  -> peer
  -> NodeToNodeVersion
- -> (StrictTVar m (AnchoredFragment (Header blk)) -> m a)
+ -> (    StrictTVar m (AnchoredFragment (Header blk))
+      -> (Their (Tip blk) -> STM m ())
+      -> m a
+    )
```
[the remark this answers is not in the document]
facundominguez: Done in 6d85c8cd.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Done in #984.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
facundominguez: Done in 838a160f4.

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -288,6 +296,7 @@ byronEraParamsNeverHardForks genesis = HardFork.EraParams {
       eraEpochSize  = fromByronEpochSlots $ Gen.configEpochSlots genesis
     , eraSlotLength = fromByronSlotLength $ genesisSlotLength genesis
     , eraSafeZone   = HardFork.UnsafeIndefiniteSafeZone
+    , eraGenesisWin = GenesisWindow (2 * Gen.unBlockCount (Gen.configK genesis))
```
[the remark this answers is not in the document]
facundominguez: Moved conversation to #984

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/Generators.hs
```diff
@@ -617,6 +619,7 @@ instance Arbitrary History.EraParams where
       <$> (EpochSize <$> arbitrary)
       <*> arbitrary
       <*> arbitrary
+      <*> (GenesisWindow <$> arbitrary)
```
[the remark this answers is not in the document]
facundominguez: Moved conversation to #984.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -150,6 +152,7 @@ prop_simple_hfc_convergence testSetup@TestSetup{..} =
         <*> testSetupSlotLength
         <*> AB (History.StandardSafeZone (safeFromTipA k))
                (safeZoneB k)
+        <*> pure (GenesisWindow ((maxRollbacks k) * 2))
```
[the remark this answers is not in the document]
facundominguez: Moved conversation to #984

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -243,6 +243,7 @@ instance BlockSupportsProtocol BlockA where
 instance LedgerSupportsProtocol BlockA where
   protocolLedgerView   _ _  = ()
   ledgerViewForecastAt _    = trivialForecast
+  computeGenesisWindow _ _  = GenesisWindow 42 -- TODO
```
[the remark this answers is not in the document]
facundominguez: This was addressed in #963

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/B.hs
```diff
@@ -203,6 +203,7 @@ instance BlockSupportsProtocol BlockB where
 instance LedgerSupportsProtocol BlockB where
   protocolLedgerView   _ _ = ()
   ledgerViewForecastAt _   = trivialForecast
+  computeGenesisWindow _ _ = GenesisWindow 42 -- TODO
```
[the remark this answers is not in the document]
facundominguez: Moved conversation to #963

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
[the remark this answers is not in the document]
facundominguez: Moved conversation to #984

PR #934, opened 2024-01-31
Factor out application of transition configuration to the Ledger repository

### LINKED ISSUE #899: Factor out application of transition config to the Ledger repo
Currently, the Consensus layer injects transition configuration data, such as initial funds, into the initial ledger state. This breaks abstraction and requires the involvement of Consensus every time Ledger-related data in the transition configuration needs to be injected. By abstracting away this functionality and moving the concrete implementation to Ledger, we simplify the process for both Ledger and Consensus. Note that this injection only happens during tests and benchmarks. 

We could do this in a two-stage approach:

1. Define a type class for injecting transition configuration data into the initial ledger state. Then, move the era-specific injection logic from Consensus to [`cardano-ledger`](https://github.com/input-output-hk/cardano-ledger).
2. Use the type class above to define [registerAny](https://github.com/IntersectMBO/ouroboros-consensus/blob/fe245ac1d8dbfb563ede2fdb6585055e12ce9738/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L1033) in terms of SOP combinators.

Since the Ledger engineers have more intimate knowledge of how the injection logic is supposed to work in each era, by requiring that the Ledger defines these injection instances, we prevent errors originating from mistakenly copying the injection function from one era to define the injection function of the next.

dnadales:
Closes #899 

Needs https://github.com/IntersectMBO/cardano-ledger/pull/4023/

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE TypeApplications         #-}
 {-# LANGUAGE TypeFamilies             #-}
 {-# LANGUAGE TypeOperators            #-}
+{-# LANGUAGE UndecidableSuperClasses  #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -- Disable completeness checks on GHC versions pre-9.6, where this can be
@@ -969,19 +970,20 @@ protocolInfoCardano paramsCardano
 
         registerAny :: NP (LedgerState -.-> LedgerState) (CardanoShelleyEras c)
         registerAny =
-             injectIntoTestState transitionConfigShelley
-          :* injectIntoTestState transitionConfigAllegra
-          :* injectIntoTestState transitionConfigMary
-          :* injectIntoTestState transitionConfigAlonzo
-          :* injectIntoTestState transitionConfigBabbage
-          :* injectIntoTestState transitionConfigConway
-          :* Nil
+            hcmap (Proxy @IsShelleyBlock) injectIntoTestState $
+                WrapTransitionConfig transitionConfigShelley
+             :* WrapTransitionConfig transitionConfigAllegra
+             :* WrapTransitionConfig transitionConfigMary
+             :* WrapTransitionConfig transitionConfigAlonzo
+             :* WrapTransitionConfig transitionConfigBabbage
+             :* WrapTransitionConfig transitionConfigConway
+             :* Nil
 
         injectIntoTestState ::
              L.EraTransition era
-          => L.TransitionConfig era
+          => WrapTransitionConfig (ShelleyBlock proto era)
           -> (LedgerState -.-> LedgerState) (ShelleyBlock proto era)
-        injectIntoTestState cfg = fn $ \st -> st {
+        injectIntoTestState (WrapTransitionConfig cfg) = fn $ \st -> st {
             Shelley.shelleyLedgerState = L.injectIntoTestState cfg (Shelley.shelleyLedgerState st)
           }
 
@@ -1110,3 +1112,16 @@ mkPartialLedgerConfigShelley transitionConfig maxMajorProtVer shelleyTriggerHard
               maxMajorProtVer
         , shelleyTriggerHardFork = shelleyTriggerHardFork
         }
+
+class
+  ( ShelleyBasedEra (ShelleyBlockLedgerEra blk)
+  , blk ~ ShelleyBlock (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
+  ) => IsShelleyBlock blk
+instance ShelleyBasedEra era => IsShelleyBlock (ShelleyBlock proto era)
+
+type family ShelleyBlockLedgerEra blk where
+  ShelleyBlockLedgerEra (ShelleyBlock proto era) = era
+
+-- | We need this wrapper to partially apply a 'TransitionConfig' in an NP.
+newtype WrapTransitionConfig blk =
+    WrapTransitionConfig (L.TransitionConfig (ShelleyBlockLedgerEra blk))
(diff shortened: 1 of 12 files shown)


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -1137,3 +1112,16 @@ mkPartialLedgerConfigShelley transitionConfig maxMajorProtVer shelleyTriggerHard
...
+  , blk ~ ShelleyBlock (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
+  ) => IsShelleyBlock blk
+instance ShelleyBasedEra era => IsShelleyBlock (ShelleyBlock proto era)
+
+type family ShelleyBlockLedgerEra blk where
+  ShelleyBlockLedgerEra (ShelleyBlock proto era) = era
+
+-- | We need this wrapper to partially apply a 'TransitionConfig' in an NP.
+newtype WrapTransitionConfig blk =
+    WrapTransitionConfig (L.TransitionConfig (ShelleyBlockLedgerEra blk))
```
jasagredo, line 1127: Shouldn't this go into the NewtypeWrappers file?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -1137,3 +1112,16 @@ mkPartialLedgerConfigShelley transitionConfig maxMajorProtVer shelleyTriggerHard
               maxMajorProtVer
         , shelleyTriggerHardFork = shelleyTriggerHardFork
         }
+
+class
+  ( ShelleyBasedEra (ShelleyBlockLedgerEra blk)
+  , blk ~ ShelleyBlock (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
+  ) => IsShelleyBlock blk
+instance ShelleyBasedEra era => IsShelleyBlock (ShelleyBlock proto era)
```
jasagredo, line 1120: Are we using this anywhere? It doesn't seem so.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -1137,3 +1112,16 @@ mkPartialLedgerConfigShelley transitionConfig maxMajorProtVer shelleyTriggerHard
...
+
+type family ShelleyBlockLedgerEra blk where
+  ShelleyBlockLedgerEra (ShelleyBlock proto era) = era
```
jasagredo, line 1123: I think this should live in the shelley subdir somewhere, perhaps in `Shelley/Block.hs`

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: It is used in `registerAny` above.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: Could even be used to share code with
https://github.com/IntersectMBO/ouroboros-consensus/blob/af4b27ed334a9a08de1be3c78e8a76f3acbc37b9/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs#L203-L207

dnadales: I'd argue for leaving it here for [the same reasons](https://github.com/IntersectMBO/ouroboros-consensus/pull/934#discussion_r1508821949).

dnadales: For some reason, I missed your last comment, @amesgen. Could I use `ShelleyBlockEra` instead?

dnadales: Also consider replacing `IsShelleyBlock` with `IsShelleyBlock` in https://github.com/IntersectMBO/ouroboros-consensus/blob/af4b27ed334a9a08de1be3c78e8a76f3acbc37b9/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs#L203-L207

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen: > so we can define this there.

You mean "can't", right? :smile:

dnadales: Yes :D

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
dnadales: I didn't realize the snippet you sent was in `DBAnalyser`. I could introduce a new module and reuse definitions from there :+1:

dnadales:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/1032

PR #935, opened 2024-01-31
Remove the type `PointSchedule`

Niols:
Builds on https://github.com/IntersectMBO/ouroboros-consensus/pull/880 that included the schedules in `GenesisTest` instead of carrying pairs `(GenesisTest, schedule)` everywhere.

This PR removes the `PointSchedule` type. It came from the observation that this type was nowadays only used as an intermediary between the `SchedulePoint`-based schedules and the peer simulator, but that it was actually only used in few places. I think it is nicer to reduce the number of notions of schedules that we have in our code -- it would make it easier to onboard on and maintain. I would however want to check first with you @tek @facundominguez whether you can think of reasons not to get rid of this type; in particular if there are pieces of work that are not on `genesis/milestone8` and that crucially use it.

This PR contains an additional commit, ac3aad8d9d5f77a2d1ba7a5ea5cf60cf18055805, that uses the type `Time` to make more explicit where we use `DiffTime` as an actual time duration and where we use it as an absolute time (in an arbitrary clock, but nonetheless). I personally think it drastically improves the reading of the functions in `SinglePeer` (and, actually, this commit started with me just wondering whether `DiffTime`s were meant as relative or absolute in some place in this PR) but we can also drop it.

I had other clean up in mind and most importantly to drop the `NodeState` type and some other related things, but I will not pursue them in this PR and I would rather have your opinion on this part of the work first.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
@@ -1,16 +1,13 @@
 module Test.Consensus.PeerSimulator.Tests.Timeouts (tests) where
 
 import           Data.Functor (($>))
-import           Data.List.NonEmpty (NonEmpty ((:|)))
 import           Data.Maybe (fromJust)
-import           Ouroboros.Consensus.Block (getHeader)
 import           Ouroboros.Consensus.Util.Condense
-import           Ouroboros.Consensus.Util.IOLike (DiffTime, fromException)
+import           Ouroboros.Consensus.Util.IOLike (DiffTime, Time (Time),
+                     fromException)
 import qualified Ouroboros.Network.AnchoredFragment as AF
-import           Ouroboros.Network.Block (tipFromHeader)
 import           Ouroboros.Network.Driver.Limits
                      (ProtocolLimitFailure (ExceededTimeLimit))
-import           Ouroboros.Network.Point (WithOrigin (At))
 import           Ouroboros.Network.Protocol.ChainSync.Codec (mustReplyTimeout)
 import           Test.Consensus.BlockTree (btTrunk)
 import           Test.Consensus.Genesis.Setup
@@ -18,7 +15,8 @@ import           Test.Consensus.PeerSimulator.Run (SchedulerConfig (..),
                      defaultSchedulerConfig)
 import           Test.Consensus.PeerSimulator.StateView
 import           Test.Consensus.PointSchedule
-import           Test.Consensus.PointSchedule.Peers (Peer (..), PeerId (..))
+import           Test.Consensus.PointSchedule.Peers (Peers, peersOnlyHonest)
+import           Test.Consensus.PointSchedule.SinglePeer (SchedulePoint (..))
 import           Test.QuickCheck
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
@@ -33,19 +31,16 @@ prop_timeouts = do
   genesisTest <- genChains (pure 0)
 
   -- Use higher tick duration to avoid the test taking really long
-  let scSchedule' = PointScheduleConfig {pscTickDuration = 1}
-
-      schedulerConfig = defaultSchedulerConfig scSchedule' (gtHonestAsc genesisTest)
+  let schedulerConfig = defaultSchedulerConfig defaultPointScheduleConfig (gtHonestAsc genesisTest)
 
       schedule =
         dullSchedule
-          scSchedule'
           (fromJust $ mustReplyTimeout (scChainSyncTimeouts schedulerConfig))
           (btTrunk $ gtBlockTree genesisTest)
 
       genesisTest' = genesisTest $> schedule
 
-  -- NOTE: Because the scheduler configuration depends on the generated
+  -- FIXME: Because the scheduler configuration depends on the generated
   -- 'GenesisTest' itself, we cannot rely on helpers such as
   -- 'forAllGenesisTest'.
   pure $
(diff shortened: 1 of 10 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer/Indices.hs
```diff
@@ -146,50 +147,59 @@ tipPointSchedule g slotLength msgDelayInterval slots = do
...
+      let (pointSeq, newBranch) = span (\(a, _) -> addTime (fst msgDelayInterval) a <= lastTime) xss
           nseq = length pointSeq
           -- The first point in xss that is not in the past
           firstLater = case newBranch of
             -- If there is no later point, pick an arbitrary later time interval
             -- to sample from
-            []           -> lastTime + toDiffTime (toEnum nseq)
-            ((a, _) : _) -> a + fst msgDelayInterval
-      times <- replicateM nseq (uniformRMDiffTime (lastTime, firstLater) g)
+            []           -> addTime (let Time lt = lastTime in lt) (slotTime (toEnum nseq)) -- REVIEW: why are we summing two absolute times?
```
facundominguez, line 184: The let statement in the first argument throws me off. I think it would work better to call a function that unwraps the `DiffTime` value or introduce the let before the call to addTime.

facundominguez, line 184: Don't forget to remove the REVIEW remark.

It is probably clear now that the second operand is not absolute.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer/Indices.hs
```diff
@@ -290,20 +300,20 @@ headerPointSchedule g msgDelayInterval xs =
...
+       go (max tNext0 (addTime t tTip)) nextHp0 []
       where
-        go :: DiffTime -> Int -> [(DiffTime, Int)] -> m ((DiffTime, Int), [(DiffTime, Int)])
+        go :: Time -> Int -> [(Time, Int)] -> m ((Time, Int), [(Time, Int)])
         go tNext nextHp acc = do
           if maybe False (tNext >) mtMax || nextHp > tp then
             pure ((tNext, nextHp), reverse acc)
           else do
-            t <- (+tNext) <$> uniformRMDiffTime msgDelayInterval g
+            t <- (flip addTime tNext) <$> uniformRMDiffTime msgDelayInterval g
```
facundominguez, line 316: I think the most idiomatic form is
```
(`addTime` tNext) <$> ...
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Tests.hs
```diff
@@ -121,7 +120,7 @@ instance QC.Arbitrary HeaderPointScheduleInput where
     branchTips <- genTipPoints
     let branchCount = length branchTips
         tpCount = sum $ map length branchTips
-    ts <- scanl1 (+) . sort <$> replicateM tpCount (chooseDiffTime (7, 12))
+    ts <- map Time <$> scanl1 (+) . sort <$> replicateM tpCount (chooseDiffTime (7, 12))
```
facundominguez, line 123: If you use `Data.Coerce.coerce` here, you can do the conversion without computing a new list. e.g.
```
Data.Coerce.coerce <$> scanl1 (+) . sort <$> replicateM tpCount (chooseDiffTime (7, 12))
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -220,35 +210,6 @@ instance Condense Tick where
     where
       showDT t = printf "%.6f" (realToFrac t :: Double)
 
-tickDefault :: PointScheduleConfig -> Word -> Peer NodeState -> Tick
```
facundominguez, line 216: Is the datatype `Tick` unused now?

Is the datatype `PointScheduleConfig` unused now?

facundominguez (approved):
LGTM

The separate commits do help the review. I left some comments and questions.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer/Indices.hs
Niols: Indeed, I asked you but apparently didn't commit the change afterwards! Fixed in e51e9c724aa5eedde950de4f1eabba4bfd8ae921.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/SinglePeer/Indices.hs
Niols: Thanks. Fixed in 86ddbd9b42bc3d22c2597d44855cf70411173d82.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: You might very well be right. Let me see if I can get rid of them.

Niols: You were absolutely right. Done in 6b5de2661ba6d2a9774ee575f4f85d0d7385e90f.

Niols:
Rebased on `niols/rework-genesistest`, itself rebased on `genesis/milestone-7-9-10`.

facundominguez (commented):
`banalStates` is dead code now.

The haddock of the module `Test.Consensus.PointSchedule` needs to be updated now that `PointSchedule` has been removed.

Niols:
I still have to investigate why CI is upset. I think I messed up something on the timeout test and it's now being killed by CI for some reason, maybe because of too many recursions? Anyways, I will try to see what happens. Then, I will proceed to try and get rid of `NodeState`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
```diff
@@ -61,9 +62,11 @@ prop_timeouts = do
...
-      let tickDuration = 1 -- 1s
+      let tickDuration = secondsToDiffTime 1 -- 1s
           maximumNumberOfTicks = round $ timeout / tickDuration
        in peersOnlyHonest $
             (Time 0, ScheduleTipPoint tipBlock)
               : (Time 0, ScheduleHeaderPoint tipBlock)
-              : zip (map (Time . (* tickDuration)) [0..]) (replicate maximumNumberOfTicks (ScheduleBlockPoint tipBlock))
+              : zip
+                  (map (Time . (* tickDuration) . secondsToDiffTime) [0..])
+                  (replicate (1 + maximumNumberOfTicks) (ScheduleBlockPoint tipBlock))
```
facundominguez, line 72: Instead of many ticks, maybe the schedule could have just one `ScheduleBlockPoint` at time `maximumNumberOfTicks*tickDuration`.

facundominguez (approved):

facundominguez:
After enabling "serve adversarial branches" which was marked as an expected failure, the test started to fail. It is a different failure than in December though, looks like when there is only one honest peer the immutable tip doesn't move.
```
ouroboros-consensus     
  Genesis tests                                                                                    
    uniform                                                                                        
      serve adversarial branches: FAIL (0.03s)                                                     
        *** Failed! Falsified (after 1 test and 1 shrink):                                         
        G | 1-0 2-2 3-3 4-5 5-6                                                                    
        SchedulerConfig:                                                                           
          ChainSyncTimeouts:                                                                       
            canAwait = Nothing                                                                     
            intersect = Nothing                                                                    
            mustReply = Nothing               
        GenesisTest:                      
          gtHonestAsc: 0.1015625
          gtSecurityParam: 2                                                                                                                                                                          
          gtGenesisWindow: 3                                                                       
          gtDelay: 0             
          gtBlockTree:                                                                                                                                                                                
            slots:    0  1  2  3  4  5  6                                                                                                                                                             
            trunk:  --1-----2--3-----4--5 
        
        SelectedChain: G | 1-0 2-2
        ChainSyncExceptions:
        
        
        No peers were killed
        Age of the immutable tip: 6
        6 > 4
        Use --quickcheck-replay=71561 --quickcheck-max-size=1 to reproduce.
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Timeouts.hs
Niols: Yes, that would work too, and maybe make the code simpler.

PR #937, opened 2024-02-02
simplify the LSQ server test

nfrisby:
For example, I removed the replication of the Immutable/VolatileTip acquire messages, since the test is so simple that the number of those is currently irrelevant---it's misleading to replicate them, as if that was already expected to matter.

### DIFF
--- a/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
+++ b/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
@@ -78,25 +78,29 @@ tests = testGroup "LocalStateQueryServer"
 --   chain, a state and send the 'QueryLedgerTip'. Collect these results.
 -- * Check that when acquiring failed, it rightfully failed. Otherwise, check
 --   whether the returned tip matches the block.
-prop_localStateQueryServer
-  :: SecurityParam
+prop_localStateQueryServer ::
+     SecurityParam
   -> BlockTree
   -> Permutation
-  -> Positive (Small Int)
   -> Property
-prop_localStateQueryServer k bt p (Positive (Small n)) = checkOutcome k chain actualOutcome
+prop_localStateQueryServer k bt p = checkOutcome k chain actualOutcome
   where
     chain :: Chain TestBlock
     chain = treePreferredChain bt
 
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -144,7 +148,7 @@ checkOutcome k chain = conjoin . map (uncurry checkResult)
         | otherwise
         -> tabulate "Acquired" ["AcquireFailurePointNotOnChain"] $ property True
       Left AcquireFailurePointTooOld
-        | pointSlot pt >= immutableSlot
+        | pointSlot pt >= immutableSlot   -- TODO what if the immtip is a multi-leader slot?
```
amesgen, line 151: Then we ask for a point with the same slot as the immutable tip, but a different hash, we get a `AcquireFailurePointNotOnChain`, right?
https://github.com/IntersectMBO/ouroboros-consensus/blob/ef8186daf6bd70fc82d7df5817cf033f44a0ee6f/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalStateQuery/Server.hs#L47-L51

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -78,25 +78,29 @@ tests = testGroup "LocalStateQueryServer"
...
   where
     chain :: Chain TestBlock
     chain = treePreferredChain bt
 
-    points :: [Target (Point TestBlock)]
-    points = permute p $
-         replicate n VolatileTip
-      ++ (SpecificPoint . blockPoint <$> treeToBlocks bt)
+    -- A random sequence of targets: one for each block in the tree and also a
+    -- random number of immtip/voltip queries
```
amesgen, line 92: > A random sequence of targets: one for each block in the tree and also **a random number of immtip/voltip queries**

Isn't the change that is explained in the PR description to exactly *remove* the fact that there is a random number of voltip queries, instead just using one?

amesgen (commented):
Looks good!

(Before merging, the "WIP" from the commit message should ofc be removed.)

PR #942, opened 2024-02-06
use the updated LoE fragment right away

tek:


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -510,6 +502,14 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
 
     processLoE curChain (LgrDB.ledgerDbCurrent ledgerDB) (writeTVar cdbLoEFrag) cdbLoE
 
+    loeFrag <- atomically (cross curChain <$> readTVar cdbLoEFrag) <&> \case
```
tek, line 505: @amesgen Facundo and I noticed this today, it seems like a straightforward mistake I made, do you concur?

Also now it looks like `prcoessLoE` could just return the fragment without storing it in the state...

amesgen: > Also now it looks like `prcoessLoE` could just return the fragment without storing it in the state...

Yes, though this of course only works in case we are sure we don't want to update the `loeFrag` asynchronously.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -510,6 +502,14 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
+        Just (_, frag) -> frag
+        -- We don't crash if the LoE fragment doesn't intersect with the selection
+        -- because we update the selection _after_ updating the LoE fragment, which
+        -- means it could move to another fork or beyond the end of the LF, depending
+        -- on the implementation of @updateLoEFrag@.
```
amesgen, line 510: Just noticed that this should refer to `processLoE` instead of `updateLoEFrag`, but it just as well applies to the comment before this PR:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
amesgen, line 505: `cdbLoEFrag` is now no longer read in the same big STM transaction as everything above. I think this should be fine; in the worst case, we fall back to the immutable tip.

Minor: You could use `readTVarIO` here.

amesgen: > Facundo and I noticed this today, it seems like a straightforward mistake I made, do you concur?

Indeed, I don't see any reason to not read the LoE fragment *after* it had a chance to be updated in `processLoE` (I think we discussed this on Monday), as this way, we have the chance to potentially select more blocks in case the LoE advanced (eg due to a disconnect by the GDD logic); but it is not immediately clear to me that the previous state had any immediate problems beyond the general concerns about whether this synchronous approach always works (ie is triggered "sufficiently often" by incoming blocks), cf. https://github.com/IntersectMBO/ouroboros-consensus/pull/932#discussion_r1480634036 or https://github.com/IntersectMBO/ouroboros-consensus/pull/878#discussion_r1463161186.

tek:
Subsumed this into https://github.com/IntersectMBO/ouroboros-consensus/pull/932, where we implemented the change to use the fragment returned from the updater, not maintain a TVar

PR #944, opened 2024-02-08
Enter finite forecast for test blocks

Niols:
A few notes/questions:

- Do we want the possibility to disable the forecast range?
- Do we want to test with forecast range != genesis window? (If so, with which bounds?)

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -117,6 +117,7 @@ genChains genNumForks = do
   pure $ GenesisTest {
     gtSecurityParam = SecurityParam (fromIntegral kcp),
     gtGenesisWindow = GenesisWindow (fromIntegral scg),
+    gtForecastRange = ForecastRange (fromIntegral scg), -- REVIEW: Do we want to generate those randomly?
     gtDelay = delta,
     gtSlotLength,
     gtChainSyncTimeouts = chainSyncTimeouts gtSlotLength asc,
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -22,6 +22,7 @@
 module Test.Consensus.PointSchedule (
     AdvertisedPoints (..)
   , BlockPoint (..)
+  , ForecastRange (..)
   , GenesisTest (..)
   , GenesisWindow (..)
   , HeaderPoint (..)
@@ -322,10 +323,14 @@ uniformPoints BlockTree {btTrunk, btBranches} g = do
 newtype GenesisWindow = GenesisWindow { unGenesisWindow :: Word64 }
   deriving (Show)
 
+newtype ForecastRange = ForecastRange { unForecastRange :: Word64 }
+  deriving (Show)
+
 -- | All the data used by point schedule tests.
 data GenesisTest schedule = GenesisTest {
   gtSecurityParam     :: SecurityParam,
   gtGenesisWindow     :: GenesisWindow,
+  gtForecastRange     :: ForecastRange, -- REVIEW: Do we want to allow infinite forecast ranges?
   gtDelay             :: Delta,
   gtBlockTree         :: BlockTree TestBlock,
   gtChainSyncTimeouts :: ChainSyncTimeout,
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -343,6 +343,7 @@ prettyGenesisTest genesisTest =
   [ "GenesisTest:"
   , "  gtSecurityParam: " ++ show (maxRollbacks gtSecurityParam)
   , "  gtGenesisWindow: " ++ show (unGenesisWindow gtGenesisWindow)
+  , "  gtForecastRange: " ++ show (unForecastRange gtForecastRange)
   , "  gtDelay: " ++ show delta
   , "  gtSlotLength: " ++ show gtSlotLength
   , "  gtChainSyncTimeouts: "
@@ -356,6 +357,7 @@ prettyGenesisTest genesisTest =
     GenesisTest {
         gtSecurityParam
       , gtGenesisWindow
+      , gtForecastRange
       , gtDelay = Delta delta
       , gtBlockTree
       , gtChainSyncTimeouts = ChainSyncTimeout{canAwaitTimeout, intersectTimeout, mustReplyTimeout}
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
@@ -63,7 +63,7 @@ constantForecastInRange range' a at = Forecast {
           then return a
           else error "constantForecastOf: precondition violated"
     forecastForWithRange (Just range) = \for ->
-        let outsideForecastMaxFor = fromWithOrigin (-1) at + range + 1
+        let outsideForecastMaxFor = succWithOrigin at + range
          in if for >= outsideForecastMaxFor
               then throwError $ OutsideForecastRange {
                   outsideForecastAt = at
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
@@ -1,8 +1,11 @@
-{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveFunctor    #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns   #-}
 
 module Ouroboros.Consensus.Forecast (
     Forecast (..)
   , OutsideForecastRange (..)
+  , constantForecastInRange
   , constantForecastOf
   , mapForecast
   , trivialForecast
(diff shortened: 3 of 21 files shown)


Niols:
Pinging you too, @amesgen, as Torsten and Facundo are away.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
```diff
@@ -543,7 +545,19 @@ testInitExtLedgerWithState st = ExtLedgerState {
     , headerState = genesisHeaderState ()
     }
 
-type instance LedgerCfg (LedgerState (TestBlockWith ptype)) = HardFork.EraParams
+data TestBlockLedgerConfig = TestBlockLedgerConfig {
+  hardForkParams :: HardFork.EraParams,
+  -- | `Nothing` means an infinite forecast range.
+  -- Instead of SlotNo, it should be something like "SlotRange"
+  forecastRange  :: Maybe SlotNo
```
amesgen, line 552: We usually try to avoid giving record fields "generic" names as they could otherwise "pollute" the namespace (GHC 8.10 doesn't support eg `NoFieldSelectors` yet).
```suggestion
  testBlockEraParams :: HardFork.EraParams,
  -- | `Nothing` means an infinite forecast range.
  -- Instead of SlotNo, it should be something like "SlotRange"
  testBlockForecastRange  :: Maybe SlotNo
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
```diff
@@ -43,13 +46,31 @@ trivialForecast x = constantForecastOf () (getTipSlot x)
...
-                        else error "constantForecastOf: precondition violated"
+    , forecastFor = forecastForWithRange range'
     }
+  where
+    forecastForWithRange Nothing = \for ->
+        if NotOrigin for >= at
+          then return a
+          else error "constantForecastOf: precondition violated"
+    forecastForWithRange (Just range) = \for ->
+        let outsideForecastMaxFor = fromWithOrigin (-1) at + range + 1
```
amesgen, line 66: `SlotNo` is unsigned, so the `-1` is wrapping around to `18446744073709551615`.
```suggestion
        let outsideForecastMaxFor = succWithOrigin at + range
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
```diff
@@ -116,6 +116,7 @@ genChains genNumForks = do
     gtHonestAsc = asc,
     gtSecurityParam = SecurityParam (fromIntegral kcp),
     gtGenesisWindow = GenesisWindow (fromIntegral scg),
+    gtForecastRange = ForecastRange (fromIntegral scg), -- REVIEW: Do we want to generate those randomly?
```
amesgen, line 119: I think we always want to set it to something equal to or greater than the Genesis window for tests that are expected to succeed. It might be interesting to set it to smaller values to ensure we have good test coverage as there should be cases where this causes a failure (liveness).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
Niols: Is it true that `succWithOrigin Origin = 0` and `succWithOrigin (At (SlotNo 0)) = 1`?

Niols: Somehow, though, changing from `fromWithOrigin 0 at` to `fromWithOrigin (-1) at` fixed a bug. Peculiar.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
Niols: Is it better to prefix with `test` or `tblc`?

Niols: Done with `tblc` in 90df5b3ad291695e0d4684f136a834df8f9e0008.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Forecast.hs
Niols: > Is it true that `succWithOrigin Origin = 0` and `succWithOrigin (At (SlotNo 0)) = 1`?

It is. Fixed in 443aed5b705a5673c9f74a9b4a70df8fb3591b4c.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/InMemory.hs
```diff
@@ -275,7 +275,7 @@ csBlockConfig = csBlockConfig' . csSecParam
 csBlockConfig' :: SecurityParam -> LedgerDbCfg (LedgerState TestBlock)
 csBlockConfig' secParam = LedgerDbCfg {
       ledgerDbCfgSecParam = secParam
-    , ledgerDbCfg         = HardFork.defaultEraParams secParam slotLength
+    , ledgerDbCfg         = testBlockLedgerConfigFrom $ HardFork.defaultEraParams secParam slotLength
```
amesgen, line 278: Style nitpick: line length

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1204,7 +1204,12 @@ checkTime cfgEnv intEnv =
         Intersects kis2 lst        <- checkArrivalTime kis arrival
         Intersects kis3 ledgerView <- case projectLedgerView slotNo lst of
             Just ledgerView -> pure $ Intersects kis2 ledgerView
-            Nothing         -> readLedgerState kis2 (projectLedgerView slotNo)
+            Nothing         -> do
+              let forecastAtSlot = forecastAt (ledgerViewForecastAt (configLedger cfg) lst)
+              EarlyExit.lift $ traceWith (tracer cfgEnv) $ TraceWaitingBeyondForecastHorizon forecastAtSlot slotNo
+              res <- readLedgerState kis2 (projectLedgerView slotNo)
```
amesgen, line 1210: Hmm, we are now calling `ledgerViewForecastAt` again just for tracing :thinking: With a bit of refactoring, it should be possible to reuse the result from `projectLedgerView`. We could also keep it like this for now, and create an action item to change this later (or even just decide to remove these trace events if they are only useful for local debugging).

Also: style nitpick: line length

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -464,21 +465,26 @@ uniformPoints BlockTree {btTrunk, btBranches} g = do
...
 
+newtype ForecastRange = ForecastRange { unForecastRange :: Word64 }
+  deriving (Show)
+
 -- | All the data used by point schedule tests.
 data GenesisTest = GenesisTest {
   gtHonestAsc     :: Asc,
   gtSecurityParam :: SecurityParam,
   gtGenesisWindow :: GenesisWindow,
+  gtForecastRange :: ForecastRange, -- REVIEW: Do we want to allow infinite forecast ranges?
```
amesgen, line 476: I think it is fine to test with finite ones, as that is also the only thing we will use in production. We can still test with very large forecast ranges in any case.

amesgen (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
Niols: Related: https://github.com/IntersectMBO/ouroboros-consensus/commit/4466b724dd324e11e9eafc970d010a3b942655f3.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/InMemory.hs
Niols: Is there a size limit used at IOG? Which is it?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: To be fair, I am not sure how necessary it is to trace where the forecast is taken from, now that we know that forecasting seems to be working fine, it was mostly to debug forecasting itself. I will try to think of a solution to either get rid of this completely or at least avoid computing things twice.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: Yes, `ForecastRange maxBound` is virtually infinite.

amesgen: We might have to be a bit careful with `maxBound` in particular to avoid overflows though.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/InMemory.hs
Niols: I'm not sure what the proper style would be, I think maybe:
```haskell
    , ledgerDbCfg         =   testBlockLedgerConfigFrom
                            $ HardFork.defaultEraParams secParam slotLength
```
or
```haskell
    , ledgerDbCfg         =
          testBlockLedgerConfigFrom
        $ HardFork.defaultEraParams secParam slotLength
```
(By the way, this anti-indentation of `$`, I really cannot. It's not a matter of style, I just find it genuinely much harder to read and regularly rewrite expressions just to read the code before reverting my change.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: I removed it altogether and fixed the style in f7ed01f4c386413e546dfce66123b5738269c21b.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: True. Maybe `maxBound / 2` which is also virtually infinite?

PR #946, opened 2024-02-09
Make db-analyser the same ghc-options as in the node

dnadales:
'db-analyser' is often used as a benchmarking tool. Thus, by using the same GHC flags as the node, we are more likely to get performance observations that correspond to those we get from a running node.

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -542,7 +542,7 @@ library unstable-cardano-tools
     , transformers-except
 
 executable db-analyser
-  import:         common-exe
+  import:         common-lib
   hs-source-dirs: app
   main-is:        db-analyser.hs
   build-depends:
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -542,7 +542,6 @@ library unstable-cardano-tools
     , transformers-except
 
 executable db-analyser
-  import:         common-exe
```
amesgen, line 545: We probably still want to import at least `common-lib` to get eg various warning flags:
https://github.com/IntersectMBO/ouroboros-consensus/blob/49181c9fea384b4fbb144f1ec86dabdf72b99e84/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal#L35-L44

jasagredo (changes requested):
This still has a merge commit instead of a rebase: e182ffd011353a5b4436c68fa81d25adef85b5a5

PR #948, opened 2024-02-12
Mempool API: generalize types of `addTxs` and `addLocalTxs`

amesgen:
This allows us to get rid of an `error` call.

Inspired by https://oleg.fi/gists/posts/2023-10-12-use-traversals-for-batch-operations.html

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxSubmission/Server.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxSubmission/Server.hs
@@ -10,6 +10,7 @@ module Ouroboros.Consensus.MiniProtocol.LocalTxSubmission.Server (
   ) where
 
 import           Control.Tracer
+import           Data.SOP.BasicFunctors
 import           Ouroboros.Consensus.Ledger.SupportsMempool
 import           Ouroboros.Consensus.Mempool.API
 import           Ouroboros.Consensus.Util.IOLike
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxSubmission/Server.hs
```diff
@@ -30,13 +31,10 @@ localTxSubmissionServer tracer mempool =
...
       recvMsgSubmitTx = \tx -> do
         traceWith tracer $ TraceReceivedTx tx
-        res <- addLocalTxs mempool [tx]
-        case res of
-          [addTxRes] -> case addTxRes of
-            MempoolTxAdded _tx             -> return (SubmitSuccess, server)
-            MempoolTxRejected _tx addTxErr -> return (SubmitFail addTxErr, server)
-          -- The output list of addTxs has the same length as the input list.
-          _                 -> error "addTxs: unexpected result"
+        I addTxRes <- addLocalTxs mempool (I tx)
```
amesgen, line 34: If we wouldn't have to support GHC 8.10, I think [`Solo`](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Tuple.html#t:Solo) would be more idiomatic here.

jasagredo: I would note this down on a comment here. For future action.

jasagredo (approved):
Looks good. I see no downsides at all.

jorisdral (approved):

PR #949, opened 2024-02-12
fix recently-broken handling of WhetherToIntervene in applyAlonzoBasedTx caught by node integration tests

fraser-iohk:
# Description

fixes a bug introduced in #865 that caused transactions with invalid scripts to be rejected, even in cases (i.e. node-to-client) where the user should be able to submit transactions with invalid scripts if they really want to


amesgen (approved):
Can you write a shorter PR/commit message, as mentioned in our [git process](https://ouroboros-consensus.cardano.intersectmbo.org/docs/for-developers/GitProcess#guiding-principles)?

PR #953, opened 2024-02-15
Move chain sync timeouts from `SchedulerConfig` to `GenesisTest`

Niols:
The idea is that the value of the timeouts depends on the chain, asc, slot length, etc., so it makes more sense to have them defined in `GenesisTest` imo. This will also avoid having us move `Asc` around only to build a `SchedulerConfig`. We keep a boolean flag `scEnableChainSyncTimeouts` in `SchedulerConfig` describing whether we should follow those timeouts when running the scheduler.

In `SchedulerConfig`:

- Remove `scChainSyncTimeouts`; replace it by a boolean flag `scEnableChainSyncTimeouts`.
- Remove `*SchedulerConfig` and replace by `defaultSchedulerConfig`.
- Remove `scSlotLength`

In `GenesisTest`:

- Add `gtSlotLength`; fill it in `GenChains`.
- Add `gtChainSyncTimeouts`; fill it in `GenChains`.
- Remove `gtHonestAsc`.


tek (approved):
looks sensible, I like it!

PR #954, opened 2024-02-15
Fix build with GHC 9.4

arrowd:
# Description

Adjusted the preprocessor directive to include GHC 9.4 and fixed a typo.


amesgen (approved):
Thanks! We don't test with GHC 9.4 on CI (only with 9.6 and, for legacy reasons, 8.10), so we can't guarantee that 9.4 will continue to work, but simple changes like this that are increasing compatibility sound fine :+1:

Can you push again (without any changes) such that the changelog check is re-triggered?

PR #955, opened 2024-02-16
New diffusion pipelining criterion

### LINKED ISSUE #938: Make diffusion pipelining criterion independent of the chain order
We soon want to adopt a non-transitive (see #891) tiebreaker in #524, which is incompatible with the current criterion for when to set the tentative header for diffusion pipelining.
https://github.com/IntersectMBO/ouroboros-consensus/blob/ef8186daf6bd70fc82d7df5817cf033f44a0ee6f/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/TentativeState.hs#L23-L31
The issue is that if a sequence of `SelectView`s is strictly improving for every adjacent pair, the same might not be true for subsequences in the presence of non-transitivity. In the case of diffusion pipelining, our downstream peers might only observe a subsequence of all tentative headers and hence disconnect from us due to this property not being fulfilled.

Instead, the idea is to only pipeline at most one block per issuer (identified by hot key and opcert number) and per block number. The goal of this issue is to design and implement this new mechanism by introducing an appropriate cache.

amesgen:
Closes #938


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
```diff
@@ -0,0 +1,230 @@
...
+-- > deriving via DisableDiffusionPipelining MyBlock
+-- >   instance BlockSupportsProtocol blk
+-- >   => BlockSupportsDiffusionPipelining MyBlock
+--
+-- This requires that the 'SelectView' is totally ordered, in particular that
+-- the order is transitive.
+--
+-- For example, if @'SelectView' ~ 'BlockNo'@, this means that a header can be
+-- pipelined if it has a larger block number than the last tentative trap
+-- header. So if someone diffused a trap header for a particular block height,
```
nfrisby, line 193: "So if ..." it'd be helpful to elaborate to something like "This would violate "Usefulness despite adversarial activity", because ..."

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node/DiffusionPipelining.hs
```diff
@@ -0,0 +1,98 @@
...
+--  - @'updateTentativeHeaderState' bcfg hdr st@ works like this:
+--
+--      - If @hdr@ and @st@ are in the same era, we call the corresponding
+--        era-specific 'updateTentativeHeaderState'.
+--
+--      - If @hdr@ is in a later era than @st@, we replace @st@ with the
+--        'initialTentativeHeaderState' for the era of @hdr@, and then proceed
+--        like in the previous step.
+--
+--      - If @hdr@ is in an earlier era than @st@, we return 'Nothing' to
```
nfrisby, line 42: Maybe call out that this has some undesirable consequences, but they don't seem worth the extra complexity?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/BlockFetch/ClientInterface.hs
```diff
@@ -203,13 +202,14 @@ mkBlockFetchConsensusInterface
          WhetherReceivingTentativeBlocks
       -> STM m (Point blk -> blk -> m ())
     mkAddFetchedBlock enabledPipelining = do
-      unlessImproved <- InvalidBlockPunishment.mkUnlessImproved (Proxy @blk)
+      unlessImproved <- InvalidBlockPunishment.mkForDiffusionPipelining
```
nfrisby, line 205: Is `unlessImproved` still a good variable name?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/BlockFetch/ClientInterface.hs
```diff
@@ -203,13 +202,14 @@ mkBlockFetchConsensusInterface
...
       pure $ mkAddFetchedBlock_ unlessImproved enabledPipelining
 
     -- Waits until the block has been written to disk, but not until chain
     -- selection has processed the block.
     mkAddFetchedBlock_ ::
-         (   SelectView (BlockProtocol blk)
+         (   BlockConfig blk
+          -> Header blk
```
nfrisby, line 212: Is `unlessImproved` still a good variable name (for the formal)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API/Types/InvalidBlockPunishment.hs
```diff
@@ -63,27 +62,23 @@ data PeerSentAnInvalidBlockException = PeerSentAnInvalidBlockException
...
-      io <- readTVar var <&> \case
-        NoLastInvalidTentative   -> pure ()
-        LastInvalidTentative old -> unless (new > old) $ do
-          enact punish invalidity
-      writeTVar var $ LastInvalidTentative new
-      pure io
+mkForDiffusionPipelining = do
+    var <- newTVar (initialTentativeHeaderState (Proxy @blk))
+    pure $ \cfg new punish -> InvalidBlockPunishment $ \invalidity -> join $ atomically $ do
+      readTVar var <&> updateTentativeHeaderState cfg new >>= \case
```
nfrisby, line 79: I don't think this line is easy to read, since the operator fixities are not extremely common.

I'd suggest binding a `mbSt'` var, for example, to make it more legible.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -925,15 +932,15 @@ chainSelection chainSelEnv chainDiffs =
             -- When the body of the tentative block turns out to be invalid, we
             -- have a valid *empty* prefix, as the tentative header fits on top
             -- of the current chain.
-            assert (isSNothing mTentativeHeader) $ do
+            assert (isNothing mTentativeHeader) $ do
```
nfrisby, line 935: ... Should we be using `StrictMaybe` in the new code?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -1268,23 +1275,23 @@ type ChainAndLedger blk = ValidatedFragment (Header blk) (LedgerDB' blk)
...
-  -> TentativeState blk
+  -> TentativeHeaderState blk
   -> ChainDiff (Header blk)
-  -> StrictMaybe (Header blk)
-isPipelineable bcfg tentativeState ChainDiff {..}
-  | -- we apply exactly one header
-    AF.Empty _ :> hdr <- getSuffix
-  , preferToLastInvalidTentative bcfg tentativeState hdr
+  -> Maybe (Header blk, TentativeHeaderState blk)
+isPipelineable bcfg st ChainDiff {..} = do
```
nfrisby, line 1288: very minor: I like the explicit guard syntax better. In particular, `MonadFail` always smells bad to me---especially with a `pure` at the head 🤷

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
```diff
@@ -1,16 +1,117 @@
...
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving anyclass (NoThunks)
+
+data ShelleyTentativeHeaderState proto =
+    -- | Legacy state, can be removed once mainnet is in Conway.
+    LegacyShelleyTentativeHeaderState !(SelectViewTentativeState proto)
+  | ShelleyTentativeHeaderState
+      -- | The block number of the last trap tentative header.
+      !(WithOrigin BlockNo)
+      -- | The set of all hot identies of those who issued trap tentative
```
nfrisby, line 54: I suggest clarifying that the scope of "all ... who ..." is dependent on the context. In ChainSel, it's all the blocks this node has discovered are invalid. In BlockFetch clients, it's what the upstream peer has sent. Right?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
```diff
@@ -1,16 +1,117 @@
...
+--
+--  - For eras before Conway, this uses the logic from
+--    'SelectViewDiffusionPipelining' for backwards-compatibility.
+--
+--  - For all eras since Conway, this uses a new scheme: A header can be
+--    pipelined iff no trap header with the same block number and by the same
+--    issuer was pipelined before. See 'HotIdentity' for what exactly we use for
+--    the issuer identity.
+--
+-- Once mainnet transitioned to Conway, we can remove the pre-Conway logic here.
```
nfrisby, line 70: transitioned -> has transitioned (or transitions)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
```diff
@@ -1,16 +1,117 @@
...
+
+      ShelleyTentativeHeaderState lastBlockNo badIdentities ->
+        case compare (NotOrigin (blockNo hdr)) lastBlockNo of
+          LT -> Nothing
+          EQ -> do
+            -- TODO Usually (especially during syncing), @badIdentities@ will be
+            -- empty. However, @hdrIdentity@ will still be forced by
+            -- 'Set.notMember'. According to the comment on 'isSelfIssued' for
+            -- @BlockSupportsMetrics ShelleyBlock@, we could save 850ns for
+            -- hasing the cold key in this case.
```
nfrisby, line 100: hasing  -> hashing

nfrisby: One microsecond seems inconsequential?

nfrisby (commented):
Beautiful code, commits and comments. Thank you for the pleasant reviewing experience.

HDR:
Haha, i got an email because i got "mentioned" in this PR
image

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
amesgen: Expanded on this :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
```diff
@@ -0,0 +1,230 @@
...
+--  - The ChainSel logic. We make sure that we only set the tentative header if
+--    this criterion is fulfilled.
+--
+--  - The BlockFetch clients, in combination with the ChainSel validation logic.
+--    For every upstream BlockFetch peer, we make sure that the invalid blocks
+--    this peer sent adhere to the pipelining criterion.
+--
+-- Concretely, the abstract Consensus layer maintains 'TentativeHeaderState's
+-- (one in ChainSel, and one for each (BlockFetch) upstream peer). Suppose that
+-- @hdr@ is a header whose block has not yet been validated, or that is a trap
```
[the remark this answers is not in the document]
nfrisby: Yep, "or is already" was clearer for me 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
nfrisby: > in contrast to before

👍

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
amesgen:
```haskell
            -- One could even replace @badIdentities@ by a list of the unhashed
            -- cold keys when its cardinality is small, for reasons similar to
            -- the ones in the comment mentioned above.
```
For ease of reference:
https://github.com/IntersectMBO/ouroboros-consensus/blob/af4b27ed334a9a08de1be3c78e8a76f3acbc37b9/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node.hs#L75-L76

Any intuition about whether this is worth to pursue/assess? Currently, I think I will just mention this in a comment as a potential opportunity.

amesgen: I think this is inconsequential, resolving.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/DiffusionPipelining.hs
```diff
@@ -0,0 +1,101 @@
...
+    where
+      gen :: Gen (NP ([] :.: WrapTentativeHeaderView) xs)
+      gen = hctraverse' (Proxy @GenTentativeHeaderViews)
+        (\p -> Comp . fmap WrapTentativeHeaderView <$> genTentativeHeaderViews p)
+        (hpure Proxy)
+
+instance GenTentativeHeaderViews ByronBlock where
+  genTentativeHeaderViews _ = do
+      bnos :: [BlockNo] <- nubOrd <$> orderedList
+      pure [PBftSelectView bno IsNotEBB | bno <- bnos]
```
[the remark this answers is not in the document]
amesgen: Yes, changed :+1: (it doesn't really matter of course as we will never (even if someone runs a Byron-only chain using our code) produce EBBs)

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/DiffusionPipelining.hs
```diff
@@ -0,0 +1,45 @@
...
+prop_diffusionPipeliningSubsequenceConsistency ::
+     forall blk.
+     BlockSupportsDiffusionPipelining blk
+  => Proxy blk
+  -> [TentativeHeaderView blk]
+     -- ^ Have to satisfy the pipelining criterion.
+  -> Property
+prop_diffusionPipeliningSubsequenceConsistency _ thvs =
+    assert (isRight $ satisfyPipeliningCriterion thvs) $
+    property $ do
```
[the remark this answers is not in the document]
amesgen: Yeah, `Testable (Gen a)` is maybe too "implicit"/"overloaded", changed to `forAllShrink` + `shrinkList (const [])` :+1:

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Consensus/DiffusionPipelining.hs
```diff
@@ -0,0 +1,45 @@
...
+-- 'BlockSupportsDiffusionPipelining'.
+prop_diffusionPipeliningSubsequenceConsistency ::
+     forall blk.
+     BlockSupportsDiffusionPipelining blk
+  => Proxy blk
+  -> [TentativeHeaderView blk]
+     -- ^ Have to satisfy the pipelining criterion.
+  -> Property
+prop_diffusionPipeliningSubsequenceConsistency _ thvs =
+    assert (isRight $ satisfyPipeliningCriterion thvs) $
```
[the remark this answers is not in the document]
amesgen: The input list should satisfy the pipelining criterion as a precondition, as the property doesn't make sense otherwise (ie, it is fine if the property fails when `isLeft $ satisfyPipeliningCriterion thvs`). So the assertion is intended to signal the difference between testing "A and B" and "A implies B" (in both cases, you only have to test something when A is true, but A being false only invalidates the first statement, not the second one). But that is minor, discarding such cases or making it part of the property also seem fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -281,7 +280,7 @@ data ChainDbEnv m blk = CDB
 -- | We include @blk@ in 'showTypeOf' because it helps resolving type families
 -- (but avoid including @m@ because we cannot impose @Typeable m@ as a
 -- constraint and still have it work with the simulator)
-instance (IOLike m, LedgerSupportsProtocol blk)
+instance (IOLike m, LedgerSupportsProtocol blk, BlockSupportsDiffusionPipelining blk)
```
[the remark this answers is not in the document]
amesgen: Yeah, I think they are just used as they conveniently imply some of the `NoThunks` instances. We could change this though to be more finegrained and explicitly list the relevant `NoThunks` instances requires.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
```diff
@@ -0,0 +1,273 @@
...
+--
+-- == High-level context
+--
+-- With DPvDV, a node is, under certain conditions, already announcing a new
+-- block to its downstream peers /before/ it has fully validated the block body.
+-- Concretely, the node maintains a /tentative header/ that, if present, extends
+-- the current selection, and is announced via the ChainSync servers to
+-- downstream peers.
+--
+-- Ideally, the body turns out to be valid, in which case the tentative header
```
nfrisby, line 48: > the tentative header is cleared

I stumbled when reading this sentence. What exactly does "cleared" mean? I think there are two possible meanings, which both seem relevant: "cleared" as in "survived scrutiny" (like an IRL background check) versus "cleared" as in "reset" (like a bit in a byte). Maybe "set to `Nothing`" would be less ambiguous.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
```diff
@@ -0,0 +1,273 @@
...
+-- (one in ChainSel, and one for each (BlockFetch) upstream peer). Suppose that
+-- @hdr@ either might turn out or is already known to be a trap header. Then
+--
+-- @'applyTentativeHeaderView' ('Proxy' \@blk) thv st@
+--
+-- (where @thv = 'tentativeHeaderView' bcfg hdr@) will return
+--
+--  - 'Nothing' if @hdr@ does not satisfy the pipelining criterion.
+--
+--      - In ChainSel, this means that @hdr@ should not be pipelined, as it might
```
nfrisby, line 86: > as it might turn out to be a trap header

I suggest either removing that clause, or changing it to

> as it would violate the criterion if it turns out to be a trap header

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
```diff
@@ -0,0 +1,273 @@
...
+  -- | See 'TentativeHeaderView'.
+  tentativeHeaderView ::
+       BlockConfig blk
+    -> Header blk
+    -> TentativeHeaderView blk
+
+  -- | Apply a 'TentativeHeaderView' to the 'TentativeHeaderState'. This returns
+  -- @'Just' st@ to indicate that the underlying header can be pipelined, and
+  -- that the 'TentativeHeaderState' must be updated to @st@ if the header
+  -- turns/turned out to be a trap header (the corresponding block body is
```
nfrisby, line 167:
```suggestion
  -- turns/turned out to be a trap header (ie the corresponding block body is
```

nfrisby (approved):
Looks good!


Every Comment I'm making here is just a minor suggestion for improving the clarity of the prose in the comments.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/DiffusionPipelining.hs
```diff
@@ -0,0 +1,104 @@
...
+  genTentativeHeaderViews _
+    | isBeforeConway (Proxy @era)
+    -- TODO These tests will reveal that the current chain order is
+    -- non-transitive. We could fix that easily, but it seems like a
+    -- distraction, and this case can be removed once we remove support for the
+    -- legacy pipelining scheme.
+    = pure []
+    | otherwise = do
+        bnos          <- nubOrd <$> orderedList
+        issuerHashes  <- nubOrd <$> replicateM numIssuers arbitrary
```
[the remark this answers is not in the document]
amesgen: As the tests happen to use Blake2B 256, this would be world breaking news :smile:

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/DiffusionPipelining.hs
```diff
@@ -0,0 +1,104 @@
...
+instance ShelleyCompatible proto era => GenTentativeHeaderViews (ShelleyBlock proto era) where
+  genTentativeHeaderViews _
+    | isBeforeConway (Proxy @era)
+    -- TODO These tests will reveal that the current chain order is
+    -- non-transitive. We could fix that easily, but it seems like a
+    -- distraction, and this case can be removed once we remove support for the
+    -- legacy pipelining scheme.
+    = pure []
+    | otherwise = do
+        bnos          <- nubOrd <$> orderedList
```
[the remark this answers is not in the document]
amesgen: I think so, with the default QuickCheck size of 100, the list will have length 50 on average, which seems fine. We technically don't need to generate lists of length <2, and long lists mostly correspond to more test cases, but I don't think this requires changes here. Thanks for raising this!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsDiffusionPipelining.hs
amesgen: Changed to "set to `Nothing`" :+1:

PR #957, opened 2024-02-16
Ensure UTF-8 support of all terminal handles

Niols:
This pull request introduces the use of the `with-utf8` library and its `withStdTerminalHandles` helper that ensures support of UTF-8 characters for all terminal handles. This is espcially useful for tests that want to output Unicode characters but do not in fear of having the Hydra Windows workers get an error of the form:

```
consensus-test.exe:<stdout>:commitBuffer:invalidargument(cannotencodecharacter'\8773')
```

There is also a helper `withUtf8` that could be used instead but it changes the behaviour of all the handles and it was not a risk I wanted to take, considering how much file and network manipulations Cardano libraries involve. It would also be possible to use the same `withStdTerminalHandles` but more locally, for instance only in test suites that do require this.

Other solutions for the same problem hinted above that I can think about are:

- to make Hydra Windows workers (and possibly Darwin?) issue eg `chcp 65001` to change the code page to unicode.
- to give up on Unicode in any outputs.

See https://serokell.io/blog/haskell-with-utf8 for an introduction of the `with-utf8` library.

Thank you, @amesgen, for the pointers.

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Main.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Main.hs
@@ -17,7 +17,6 @@ import           Test.Util.TestEnv (defaultMainWithTestEnv,
 main :: IO ()
 main = do
   hSetBuffering stdout LineBuffering
-  hSetEncoding stdout utf8
   defaultMainWithTestEnv defaultTestEnvConfig tests
 
 tests :: TestTree
--- a/ouroboros-consensus-cardano/test/cardano-test/Main.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Main.hs
@@ -1,7 +1,6 @@
 module Main (main) where
 
-import           System.IO (BufferMode (LineBuffering), hSetBuffering,
-                     hSetEncoding, stdout, utf8)
+import           System.IO (BufferMode (LineBuffering), hSetBuffering, stdout)
 import qualified Test.Consensus.Cardano.ByronCompatibility
 import qualified Test.Consensus.Cardano.Golden
 import qualified Test.Consensus.Cardano.MiniProtocol.LocalTxSubmission.Server
--- a/scripts/release/create-release.hs
+++ b/scripts/release/create-release.hs
@@ -10,6 +10,7 @@
     prettyprinter,
     text,
     turtle ^>=1.6.0,
+    with-utf8,
 -}
 {-# OPTIONS_GHC -Wall -Wextra #-}
 {-# LANGUAGE BlockArguments      #-}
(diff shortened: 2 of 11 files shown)


ouroboros-consensus-cardano/test/cardano-test/Main.hs
```diff
@@ -16,7 +17,7 @@ import           Test.Util.TestEnv (defaultMainWithTestEnv,
                      defaultTestEnvConfig)
 
 main :: IO ()
-main = do
+main = withStdTerminalHandles $ do
   hSetBuffering stdout LineBuffering
   hSetEncoding stdout utf8
```
amesgen, line 22: I think you can now remove the
```haskell
hSetEncoding stdout utf8
```
line

ouroboros-consensus-cardano/test/cardano-test/Main.hs
```diff
@@ -16,7 +17,7 @@ import           Test.Util.TestEnv (defaultMainWithTestEnv,
...
   defaultMainWithTestEnv defaultTestEnvConfig tests
```
amesgen, line 23: You can minimize the diff a bit by adapting `defaultMainWithTestEnv` to call `withStdTerminalHandles`, instead of having to adapt evey single `test-suite`'s `main`.

scripts/release/create-release.hs
```diff
@@ -30,12 +30,13 @@ import           Data.Semigroup (Max (..))
...
 import qualified Data.Text.IO as Text
 import           Data.Version
+import           Main.Utf8 (withStdTerminalHandles)
 import           Prettyprinter
 import           System.FilePath
 import           Turtle hiding (d, fp, l, o)
 
 main :: IO ()
-main = sh do
+main = withStdTerminalHandles $ sh do
```
amesgen, line 39: You also have to adapt the `build-depends` here at the top of the file.

amesgen (approved):
Thanks a lot! A few minor comments.

ouroboros-consensus-cardano/test/cardano-test/Main.hs
Niols: I'm not super convinced myself, I think it is nice to make it explicit in the main file rather than pushing it in a helper, but that is not a strong opinion. Done in 8a5bb42ffe236582fba64d84f4ee66c0d2640875.

scripts/release/create-release.hs
Niols: Done in 2fd5bfa4d25f7e73391396be69aef48569f61bf9.

ouroboros-consensus-cardano/test/cardano-test/Main.hs
Niols: Done in bff359cbb6b53ee9ff774f9ac040fa455c86998d.

PR #963, opened 2024-02-21
Add a HF `Qry` for the Genesis window, remove `computeGenesisWindow`

amesgen:
This is both simpler and more general.

The first and third commit are purely mechanical.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
@@ -1,9 +1,11 @@
-{-# LANGUAGE FlexibleContexts   #-}
-{-# LANGUAGE GADTs              #-}
-{-# LANGUAGE PatternSynonyms    #-}
-{-# LANGUAGE RankNTypes         #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TypeFamilies       #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TypeFamilies               #-}
 
 module Ouroboros.Consensus.Block.Abstract (
     -- * Protocol
@@ -30,6 +32,8 @@ module Ouroboros.Consensus.Block.Abstract (
   , encodeRawHash
     -- * Utilities for working with WithOrigin
   , succWithOrigin
+    -- * Ouroboros Genesis window
+  , GenesisWindow (..)
     -- * Re-export basic definitions from @ouroboros-network@
   , ChainHash (..)
   , HasHeader (..)
@@ -71,7 +75,8 @@ import           Data.ByteString.Short (ShortByteString)
 import qualified Data.ByteString.Short as Short
 import           Data.Kind (Type)
 import           Data.Maybe (isJust)
-import           Data.Word (Word32)
+import           Data.Word (Word32, Word64)
+import           NoThunks.Class (NoThunks)
 import           Ouroboros.Consensus.Block.EBB
 import           Ouroboros.Network.Block (ChainHash (..), HasHeader (..),
                      HeaderFields (..), HeaderHash, Point, StandardHash,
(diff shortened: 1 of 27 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
```diff
@@ -337,6 +340,12 @@ evalExprInEra EraSummary{..} = \(ClosedExpr e) -> go e
         guardEnd $ \end -> e < boundEpoch end
         return eraEpochSize
 
+    go (EGenesisWindow expr) = do
+        s <- go expr
+        guard    $ s >= boundSlot eraStart
```
tek, line 345: do these just match the slot to the proper era or is it possible that we get `Nothing`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -239,16 +240,35 @@ updateLoEFragGenesis cfg tracer getCandidates getHandles =
...
-      (losingPeers, bounds) =
-        densityDisconnect sgen (configSecurityParam cfg) candidateSuffixes theirTips latestSlots loeFrag
-      loeHead = AF.headAnchor loeFrag
+    let msgen :: Maybe GenesisWindow
+        msgen = eitherToMaybe $ runQuery qry summary
+          where
+            -- Usually, the initial ledger state (with tip slot 'Origin') is in
+            -- the first non-empty era, see 'injectInitialExtLedgerState' and
+            -- its call to 'State.extendToSlot' with @'SlotNo' 0@.
+            slot    = fromWithOrigin 0 $ AF.headSlot loeFrag
```
Niols, line 249: I don't understand the comment. Is it explaining why it's fine to not make the distinction between `Origin` and `NotOrigin (SlotNo 0)`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
amesgen: This function, `evalExprInEra` runs the given `Expr` in the context of one specific `EraSummary`. In `runQuery`, we then run it against all eras we know something about (see `Summary`, which contains an `EraSummary` for a (non-empty) prefix of all eras), and returns the first one where `evalExprInEra` succeeded:
https://github.com/IntersectMBO/ouroboros-consensus/blob/ae6b5a37a334a6b1e9c02452de5aa4899f00256e/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs#L394-L396

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -241,3 +246,12 @@ pattern NotOrigin t = Cardano.At t
 -- 'SlotNo' and 'BlockNo'.
 succWithOrigin :: (Bounded t, Enum t) => WithOrigin t -> t
 succWithOrigin = withOrigin minBound succ
+
+{-------------------------------------------------------------------------------
+  Ouroboros Genesis window
+-------------------------------------------------------------------------------}
+
+-- TOOD docs/better place?
```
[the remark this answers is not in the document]
amesgen: 8dd539346ce4291b7661486daf574ea1423410bd

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
amesgen: > Is it explaining why it's fine to not make the distinction between `Origin` and `NotOrigin (SlotNo 0)`?

Now that I think about it again, an alternative would be to use `succWithOrigin` here and argue that we should use the Genesis window associated with the first slot *after* the intersection. This only makes a difference with the current behavior if we the intersection is in the last slot of an era. In that case, all slots in the Genesis window are in the successor era, so `succWithOrigin` actually sounds nicer here. Will change it to that :+1:

amesgen: 6059c08a1d89e6c35ce0144f52f5dea2c6b423f7

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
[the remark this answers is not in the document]
tek: ugh that happens to me _all_ the time

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
tek: thanks

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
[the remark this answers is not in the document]
Niols: Me too :p

PR #965, opened 2024-02-22
consensus-diffusion: add lower bound on io-sim

nfrisby:
The GSM QSM test will fail if the io-sim dep doesn't include a recent bug fix.


nfrisby:
Nevermind. The upstream dependencies are more intertwined than I anticipated, so this PR would much more work than I expected.

PR #967, opened 2024-02-22
Various cleanup of the schedules

Niols:
This PR brings various improvements to the schedules. Most importantly:

- Get rid of the old type `NodeState = NodeOffline | NodeOnline AdvertisedPoints` because we never actually used `NodeOffline` and there was no observable difference between a tick with `NodeOffline` and no tick. The old `AdvertisedPoints` takes its place and gets named `NodeState`. We also remove the types `TipPoint`, `HeaderPoint` and `BlockPoint`.

- Make the `SchedulePoint`s carry `WithOrigin TestBlock`s and not just `TestBlock`s. Although not used by the current generators, it does make sense to add this and causes only a minimal diff. In some places, we actually would need this (cc @nbacquey).

- Generalise quite a few types and functions to any `blk` and not just `TestBlock`. This makes more explicit the places where we actually use features of `TestBlock`s. This will also allow reuse of schedule generators with other blocks if anyone ever wishes to do this. @amesgen> At this point, the peer simulator is still stuck with `TestBlock`s, but this is only for printing purposes. We could consider going for tracing in a similar way as other parts of the code with a type `TracePeerSimulatorEvent blk`. Do you think that this is valuable or will nobody ever want to use the peer simulator with other block types?

The commits can be read in order.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -9,9 +9,9 @@
 
 -- | Data types and generators for point schedules.
 --
--- Each generator takes a set of 'AnchoredFragment's corresponding to the tested peers'
--- chains, and converts them to a point schedule consisting of a sequence of states
--- ('AdvertisedPoints'), each of which is associated with a single peer.
+-- Each generator takes a set of 'AnchoredFragment's corresponding to the tested
+-- peers' chains, and converts them to a point schedule consisting of a sequence
+-- of 'NodeState's, each of which is associated with a single peer.
 --
 -- When a schedule is executed in a test, each tick is processed in order.
 -- The peer associated with the current tick is considered "active", which means that
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -20,8 +20,7 @@
 -- and once it fulfills the state's criteria, it yields control back to the scheduler,
 -- who then activates the next tick's peer.
 module Test.Consensus.PointSchedule (
-    BlockPoint (..)
-  , ForecastRange (..)
+    ForecastRange (..)
   , GenesisTest (..)
   , GenesisWindow (..)
   , HeaderPoint (..)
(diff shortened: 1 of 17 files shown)


amesgen:
> * At this point, the peer simulator is still stuck with `TestBlock`s, but this is only for printing purposes. We could consider going for tracing in a similar way as other parts of the code with a type `TracePeerSimulatorEvent blk`. Do you think that this is valuable or will nobody ever want to use the peer simulator with other block types?

I haven't looked at the details, but if the only step to make it block-generic is just to pass in some block-specific tracer, I think that would be neat to showcase we are not making the tests too `TestBlock`-specific, even if we don't use it for now.

Niols:
Alright, thanks! This is a change that involves a lot of lines in many places so I will keep it for a subsequent PR.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -20,18 +20,18 @@
 -- and once it fulfills the state's criteria, it yields control back to the scheduler,
```
nbacquey, line 21: There still is a dangling `AdvertisedPoints` in the comment above

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -80,127 +79,57 @@ import           Test.Ouroboros.Consensus.ChainGenerator.Params (Delta (Delta))
...
-
-genesisAdvertisedPoints :: AdvertisedPoints
-genesisAdvertisedPoints =
-  AdvertisedPoints {
-    tip = TipPoint TipGenesis,
-    header = HeaderPoint Origin,
-    block = BlockPoint Origin
+      (padListWith PadLeft $ map (terseWithOrigin terseBlock . nsTip) points)
+      (padListWith PadLeft $ map (terseWithOrigin terseBlock . nsHeader) points)
+      (padListWith PadLeft $ map (terseWithOrigin terseBlock . nsBlock) points)
```
nbacquey, line 118: The lists of formerly tip/header/block-points used to be left-aligned (i.e. padded with `PadRight`).
I'm not particularly opinionated on the question, but do you think right-alignment looks better?

nbacquey (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: Found only one. Done in 6a7a36fb1a94e90a9134172b9955531d7b291fc9.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
Niols: It was a mistake from my side, I do want those things left-aligned. Done in cbe13492d4c44ede88075990eeb35480f6966de4.

PR #969, opened 2024-02-23
Release ouroboros-consensus packages

bolt12:
Release Consensus packages for Node 8.9.

Technically, ouroboros-consensus-cardano could be a minor version bump, but in recent releases, it seems we usually released major versions for new node releases, and it is certainly not incorrect.


amesgen (approved):
CHaP CI passes: https://github.com/IntersectMBO/cardano-haskell-packages/pull/678

PR #970, opened 2024-02-26
Add a `db-analyser` analysis to get the UTxO size per slot

dnadales:
This patch adds a new analysis to `db-analyser`: `--get-block-application-metrics N`. Currently, for each block every `N`, we compute the block slot, block number, UTxO size, and UTxO map size. The analysis can be extended to other metrics extracted from the previous ledger state, current block, and ledger state resulting from applying said block.

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
@@ -23,8 +23,10 @@ module Cardano.Tools.DBAnalyser.Block.Cardano (
   , CardanoBlockArgs
   ) where
 
+import qualified Cardano.Chain.Block as Byron.Block
 import qualified Cardano.Chain.Genesis as Byron.Genesis
 import qualified Cardano.Chain.Update as Byron.Update
+import qualified Cardano.Chain.UTxO as Byron.UTxO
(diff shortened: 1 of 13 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
```diff
@@ -269,6 +282,77 @@ instance (HasAnnTip (CardanoBlock StandardCrypto), GetPrevHash (CardanoBlock Sta
...
+  -> IO Builder
+dispatch cardanoSt fByron fShelley =
+ case cardanoSt of
+      Cardano.Block.LedgerStateByron   st -> fByron   st
+      Cardano.Block.LedgerStateShelley st -> fShelley st
+      Cardano.Block.LedgerStateAllegra st -> fShelley st
+      Cardano.Block.LedgerStateMary    st -> fShelley st
+      Cardano.Block.LedgerStateAlonzo  st -> fShelley st
+      Cardano.Block.LedgerStateBabbage st -> fShelley st
+      Cardano.Block.LedgerStateConway  st -> fShelley st
```
jasagredo, line 323: This could be probably just an NP?
```
fn fByron :* fn fShelley :* .... :* Nil
```

jasagredo (commented):
Rest looks good I think

jasagredo (approved):

dnadales:
Using the Cardano chain synced up to slot 109510116, block 9600000, yields these results:

image

PR #990, opened 2024-03-11
Introduce a time-aware tracer

Niols:
...instead of relying on new ticks to print the time. The upside is that we now print the time of events that occur mid-tick (and not at the beginning of ticks), which can be useful when debugging eg. timeout-based tests. The downside is that ticks that happen at the same absolute time are not separated anymore (although maybe that's for the best?) and we rely on IOSim a bit more because we assume that time moves in chunks.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
@@ -119,10 +119,11 @@ tracerTestBlock tracer0 = do
       tracer = Tracer $ \msg -> do
         time <- getMonotonicTime
         tickTime <- readTVarIO tickTimeVar
-        traceWith tracer0 $
-          if time /= tickTime
-            then concat $ intersperse "\n" $ map ((prettyTime time ++ " ") ++) $ lines msg
-            else msg
+        let timeHeader = prettyTime time ++ " "
+            prefix = if time /= tickTime
+              then timeHeader
+              else replicate (length timeHeader) ' '
+        traceWith tracer0 $ concat $ intersperse "\n" $ map (prefix ++) $ lines msg
   pure $ Tracer $ traceEventTestBlockWith setTickTime tracer0 tracer
 
 traceEventTestBlockWith ::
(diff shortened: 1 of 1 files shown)


facundominguez:
My intuition, so far, is that time is only advanced between ticks. What is the purpose of printing the time for every trace message?

Niols:
We don't print the time at each message, we only print a horizontal bar and
the time when the time changes (which makes sense in IOSim but not in IO).
The thing is:

- Time may remain unchanged between two ticks (in case of zero-duration
ticks).

- Time may change within a tick (in case of timeouts or LoP or other
time-based mechanisms).

I found it frustrating to not have the time of events that happen not at
the beginning of ticks and wanted them to stand out a bit more. I feel this
is achieved. However, I am not particularly convinced by the fact that the
traces would become awful in IO or by the fact that ticks aren't
necessarily separated anymore.

I think something that could make sense would be to always separate ticks
with horizontal lines and to prefix the messages by their time if they
don't happen at the beginning of a tick? This would end up prefixing all
the messages in IO and only some in IOSim, but it would achieve my initial
goal without being as disruptive as the current work.

In general, I think we're hitting something that's been puzzling me a bit
for the past weeks: we basically have two semantics of ticks:

- The intended semantics is that they are blocks of time long enough to
ensure that the system stabilises with the new state before the end of it.

- The IOSim semantics is that the system stabilises instantaneously and
that we then wait for the duration of the tick.

These are close but behave a bit differently with zero-duration ticks or
with time-based events unrelated to the schedules. I think we rely more on
the IOSim semantics in many places and some tests might therefore not even
pass in IO.

facundominguez:
> - Time may remain unchanged between two ticks (in case of zero-duration ticks).
> 
> - Time may change within a tick (in case of timeouts or LoP or other time-based mechanisms).

These looks to me like good reasons.

> some tests might therefore not even pass in IO.

Our simulated peers wouldn't work in IO because of how running the schedule dispatches the state updates. I think it would be fine to continue to disregard IO.

> I think something that could make sense would be to always separate ticks with horizontal lines

When reading traces, it has been helpful to jump between the point schedule and the traces specific to a tick. It would be worth keeping this navigation easy to do. Perhaps the horizontal lines with separators work for that sake.

Also maybe share here an example of a trace?

facundominguez (approved):
LGTM!

tek:
pretty cool

Niols:
Hum, I enjoy the alignment but we now have a prefix of 38 characters + QuickCheck prefix. Most of the time, 10 of these are useless (because most of the time things happen at the beginning of a tick) and often some others are (like there is a loooot of padding after “ChainDB”). I think that's why we had initially given up on the alignment and systematically prefixing with time. I'm also not convinced by dropping the “Scheduled”. It is obvious for us, but I wouldn't want to confuse a random person trying to debug why a test broke in the future. Still, that's all a matter of taste, so let's keep it that way.

PR #991, opened 2024-03-11
Use cardano-git-rev from cardano-base repo

erikd:
# Description
The `cardano-git-rev` package has been moved from the `cardano-node` repo  to `cardano-base`. `gitRev` has also changed from being a function to a TH splice (needed for it to work correctly).

Also note that:

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -499,7 +499,7 @@ library unstable-cardano-tools
     , cardano-crypto
     , cardano-crypto-class
     , cardano-crypto-wrapper
-    , cardano-git-rev
+    , cardano-git-rev                ^>=0.2.1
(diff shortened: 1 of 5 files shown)


dnadales (approved):
LGTM. Thanks!

erikd:
Got some strangeness that I need to figure out before this can be merged.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -499,7 +499,7 @@ library unstable-cardano-tools
     , cardano-crypto
     , cardano-crypto-class
     , cardano-crypto-wrapper
-    , cardano-git-rev
```
amesgen, line 502: Can you also add a `<0.3` upper bound?

erikd: Done.

erikd:
TH strangeness fixed.

PR #992, opened 2024-03-12
Store the fork history for HP in the node state for BlockFetch

tek:
PoC for storing all chains visited by HP in the node state to allow BlockFetch to send blocks from before fork switches.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LoE.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LoE.hs
@@ -83,11 +83,13 @@ prop_adversaryHitsTimeouts timeoutsEnabled =
 
     delaySchedule :: HasHeader blk => BlockTree blk -> Peers (PeerSchedule blk)
     delaySchedule tree =
-      let trunkTip = case btTrunk tree of
+      let trunk = btTrunk tree
+          trunkTip = case trunk of
             (AF.Empty _)       -> error "tree must have at least one block"
             (_ AF.:> tipBlock) -> tipBlock
           branch = getOnlyBranch tree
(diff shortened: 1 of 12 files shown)


facundominguez (commented):
The PR seems to be delivering what's intended. I have questions about two aspects.

1. Using AnchoredFragments requires some modifications to tests and generators, when I was expecting the changes would be local to the logic of the BlockFetch server. Is this not possible?

2. Using AnchoredFragments involves some manipulations to update them. The motivation for AnchoredFragments, however, would be to have an implementation that can work with other blocks than TestBlock. Since we aren't planning to use other types of blocks for testing. Can we stick with a list of header points instead and use `isAncestorOf`?

tek:
>     1. Using AnchoredFragments requires some modifications to tests and generators, when I was expecting the changes would be local to the logic of the BlockFetch server. Is this not possible?

That was my initial plan, but Alex suggested in the daily to store them in the node state.
Additionally, discarding the fragments when generating peer schedules would mean that we'd have to look up every HP in the block tree later, which seemed too expensive to me. Otherwise I could have done the same in `peerStates` as well.

>     2. Using AnchoredFragments involves some manipulations to update them. The motivation for AnchoredFragments, however, would be to have an implementation that can work with other blocks than TestBlock. Since we aren't planning to use other types of blocks for testing. Can we stick with a list of header points instead and use `isAncestorOf`?

Since we already implemented an alternative solution with `isAncestorOf`, I guess that's what were doing?

tek:
Closing in favor of doing a simple ancestor check with `TestHash`.

For the record, this approach would have the following potential benefits:

* BlockFetch clients cannot request ranges that hadn't been advertised before. Unclear if we have a use for that.
* Other handlers can use the fragments to slightly improve performance.
* Not depending on `TestBlock`.

PR #993, opened 2024-03-12
Fix mempool and backing storage benchmarks

RenateEilers:
# Description

Fix the compilation errors in the mempool and backing storage benchmarks.


jasagredo (commented):
Looks good, we should find out why the mempool benchmark doesn't terminate before merging this

PR #996, opened 2024-03-13
Bump `NodeToNodeVersion` part of `latestReleasedNodeVersion`  to `NodeToNodeV_13`

amesgen:
This bumps the `NodeToNodeVersion` part of `latestReleasedNodeVersion` to `NodeToNodeV_13` from `NodeToNodeV_11`.

 - `NodeToNodeV_12` does nothing, see https://github.com/IntersectMBO/ouroboros-network/pull/4825
 - `NodeToNodeV_13`, which enables the latest version of PeerSharing.

PR #940, opened 2024-02-05
Integrate #4795

bolt12:
# Description

This PR integrates changes in [ouroboros-network #4795](IntersectMBO/ouroboros-network#4795). Which basically removes the `computePeers` callback from diffusion functions and provides a cleaner interface for consensus.

Note: This PR probably depends on #808 to be merged since bootstrap peers work has already been merged in ouroboros-network.

PR #997, opened 2024-03-13
Backports for Consensus release for Node 8.9.1

amesgen:
This backports:

 - CHaP revision for `cardano-git-rev`
 - #996
 - #940


amesgen:
(Some Hydra jobs show here as cancelled even though they completed successfully after a restart)

PR #984, opened 2024-03-05
[WIP] Implement the GDD governor

facundominguez:
The GDD governor disconnects nodes when there is disagreement about which chain to follow.

Rebase of #932.

PR #1004, opened 2024-03-18
Run GDD asynchronously

tek:


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -272,7 +274,9 @@ runPointSchedule schedulerConfig genesisTest tracer0 =
           = pure nullTracer
     stateTracer <- mkStateTracer
     BlockFetch.startBlockFetchLogic registry chainDb fetchClientRegistry getCandidates
-    void $ forkLinkedThread registry "ChainSel trigger" (reprocessLoEBlocksOnCandidateChange chainDb getCandidates)
+    -- void $ forkLinkedThread registry "ChainSel trigger" (reprocessLoEBlocksOnCandidateChange chainDb getCandidates)
```
facundominguez, line 277: Remove the comment. (?)
Is `reprocessLoEBlocksOnCandidateChange` dead code now?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -127,6 +128,30 @@ reprocessLoEBlocksOnCandidateChange chainDb getCandidates =
...
+    let newUpdateLoEFrag = UpdateLoEFrag $ \_ _ -> readTVarIO varLoEFrag
+    pure (newUpdateLoEFrag, spin varLoEFrag mempty)
+  where
+    spin varLoEFrag oldTrigger chainDb getTrigger = do
+      (newTrigger, curChain, curLedger) <- atomically $ do
+        (_, newTrigger) <- blockUntilChanged id oldTrigger getTrigger
+        curChain <- ChainDB.getCurrentChain chainDb
+        curLedger <- ChainDB.getCurrentLedger chainDb
+        pure (newTrigger, curChain, curLedger)
+      loeFrag <- updateLoEFrag loeUpdater curChain curLedger
```
facundominguez, line 150: If `updateLoEFrag` is updated here, shouldn't it be removed from ChainSel?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -272,7 +274,9 @@ runPointSchedule schedulerConfig genesisTest tracer0 =
...
+    void $ forkLinkedThread registry "LoE updater background" $
+      runLoEUpdaterBackground chainDb getLatestSlots
```
facundominguez, line 279: Should the LoE fragment continue to be updated when `getCandidates` changes? I don't see now where this happens.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -127,6 +128,30 @@ reprocessLoEBlocksOnCandidateChange chainDb getCandidates =
...
+      atomically $ writeTVar varLoEFrag loeFrag
+      reprocessLoEAsync chainDb
```
facundominguez, line 152: `reprocessLoEAsync` is a bit confusing now. The LoE fragment has been already updated. Maybe this is meant to update the chain selection instead (?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -127,6 +128,30 @@ reprocessLoEBlocksOnCandidateChange chainDb getCandidates =
         reprocessLoEAsync chainDb
         spin newTips
 
+runLoEUpdaterOnChange ::
+  (Monoid a, Eq a, IOLike m, LedgerSupportsProtocol blk) =>
+  UpdateLoEFrag m blk ->
+  m ( UpdateLoEFrag m blk
+    , ChainDB m blk ->
+      STM m a -> -- Re-run the LoE updater whenever this changes
+      m Void
```
facundominguez, line 137: Document what the returned computation does.

Also, I don't understand what the return `UpdateLoEFrag m blk` is for.

facundominguez (commented):
I left some comments. I think the commit message should mention that one purpose of this PR is to have GDD disconnect nodes when there are changes to the last set slot.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -246,7 +246,9 @@ runPointSchedule schedulerConfig genesisTest tracer0 =
     resources <- makePeerSimulatorResources tracer gtBlockTree (getPeerIds gtSchedule)
     let getCandidates = traverse readTVar =<< readTVar (psrCandidates resources)
         getHandles = readTVar (psrHandles resources)
-        updateLoEFrag = updateLoEFragGenesis config (mkGDDTracerTestBlock tracer) getCandidates getHandles
+        getLatestSlots = traverse CSClient.cschLatestSlot =<< getHandles
```
amesgen, line 249: FTR: I think we additionally want to wake up the GDD when any of the CS clients (no longer) sends `MsgAwaitReply`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
amesgen: Yes, when `getCandidates` changes, `getLatestSlots` also changes, which is used as the trigger above.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
amesgen: This can probably be refactored if we remove `UpdateLoEFrag` as brought up by @tek (also see [Slack](https://moduscreate.slack.com/archives/C05CEUFFH7Z/p1710783577790569)).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
amesgen: Yeah, `reprocessLoEAsync` is maybe a somewhat misleading name, maybe `triggerChainSelection` is more intuitive?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
facundominguez: The latest slot is set in a separate transaction from the candidate fragment and "their" tip. It is possible that the GDD runs before the candidate is updated. Either everything needs to be done on a single transaction now, or this PR needs to take other measures to update after candidates change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: It's reprocessing blocks postponed by the LoE, which I perceive as the overarching mechanism of which the LoE fragment is one part...so it should probably be `reprocessLoEBlocksAsync`?

`triggerChainSelection` seems a bit too general, but maybe it actually is, since we're processing all blocks that can possibly be eligible for extending the chain?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
tek: I deleted the function

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
tek: Also it looks like the last slot isn't updated at all in `nextStep`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: I untangled it a bit

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
tek: alright I'll look into it

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: I like `triggerChainSelection`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Resources.hs
```diff
@@ -120,7 +122,9 @@ data PeerSimulatorResources m blk =
 
     -- | Handlers to interact with the ChainSync client of each peer.
     -- See 'ChainSyncClientHandle' for more details.
-    psrHandles :: StrictTVar m (Map PeerId (ChainSyncClientHandle m TestBlock))
+    psrHandles :: StrictTVar m (Map PeerId (ChainSyncClientHandle m TestBlock)),
+
+    psrIdling :: StrictTVar m (Set PeerId)
```
facundominguez, line 127: Documentation please :)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Resources.hs
```diff
@@ -240,4 +244,5 @@ makePeerSimulatorResources tracer blockTree peers = do
     pure (peerId, peerResources)
   psrCandidates <- uncheckedNewTVarM mempty
   psrHandles <- uncheckedNewTVarM mempty
-  pure PeerSimulatorResources {psrCandidates, psrPeers = Map.fromList $ toList resources, psrHandles}
+  psrIdling <- uncheckedNewTVarM $ Set.empty
```
facundominguez, line 247:
```suggestion
  psrIdling <- uncheckedNewTVarM Set.empty
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -110,22 +135,32 @@ updateLoEFragStall getCandidates =
...
-  Ord peer =>
-  MonadSTM m =>
-  GetHeader blk =>
+-- | Run the GDD governor in a background thread, writing the LoE fragment
+-- computed by @loEUpdater@ to @varLoEFrag@ whenever the STM action
+-- @getTrigger@ changes.
+--
+-- After writing the fragment, send a message to ChainSel to reprocess all
+-- blocks that had previously been postponed by the LoE.
+runGddAsync ::
```
facundominguez, line 144: `runGddAsync` doesn't run GDD in a background thread.

```suggestion
-- | A never ending computation that runs the GDD governor whenever
-- the STM action @getTrigger@ changes, writing the LoE fragment
-- computed by @loEUpdater@ to @varLoEFrag@, and then triggering
-- ChainSel to reprocess all blocks that had previously been
-- postponed by the LoE.
runGddAsync ::
```

facundominguez, line 144: Maybe `runGddAsync` should be called `runGDD` without the `Async` suffix.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -256,23 +270,28 @@ runPointSchedule schedulerConfig genesisTest tracer0 =
...
           = pure nullTracer
+
+        gdd = updateLoEFragGenesis config (mkGDDTracerTestBlock tracer) getCandidates getHandles
+
     stateTracer <- mkStateTracer
     BlockFetch.startBlockFetchLogic registry chainDb fetchClientRegistry getCandidates
-    void $ forkLinkedThread registry "ChainSel trigger" (reprocessLoEBlocksOnCandidateChange chainDb getCandidates)
+    for_ loEVar $ \ var ->
+        void $ forkLinkedThread registry "LoE updater background" $
+          runGddAsync gdd var chainDb ((,) <$> getLatestSlots <*> readTVar (psrIdling resources))
```
facundominguez, line 294: The GDD governor is not using `psrIdling` to compute densities. Is this going to be the subject of another PR?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
```diff
@@ -159,10 +159,10 @@ runChainSyncClient
   StateViewTracers {svtPeerSimulatorResultsTracer}
   varCandidates
   varHandles
+  varIdling
   channel = do
     -- We don't need this shared Set yet. If we need it at some point,
     -- it ought to be passed to `runChainSyncClient`.
```
facundominguez, line 165:
```suggestion
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -162,14 +165,30 @@ runGddAsync loEUpdater varLoEFrag chainDb getTrigger =
...
+  -- | The candidate fragment is empty and the TVar containing the latest
+  -- slot ChainSync has seen does not contain an entry for the current peer.
+  NoLatestSlot
+  |
+  -- | The candidate fragment head is the latest known header.
+  LatestSlotCandidate SlotNo
+  |
+  -- | ChainSync has seen a header after the candidate fragment head, most
+  -- likely because it is beyond the forecast horizon.
+  LatestSlotForecast SlotNo
```
facundominguez, line 181: How about we change `LatestSlotForecast` to `LatestSlotAfterSGen`? I don't think we care about the forecast horizon as much as we care of the header being beyond the genesis window.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -197,52 +216,83 @@ densityDisconnect ::
...
+          caughtUp = Set.member peer caughtUpPeers
+
+          latestSlot = case (AF.headSlot candidateSuffix, latestSlots Map.!? peer) of
+            (Origin, Nothing) -> NoLatestSlot
+            (Origin, Just latest)
+              | latest > 0 -> LatestSlotForecast latest
+              | otherwise -> NoLatestSlot
+            (NotOrigin cand, Nothing) -> LatestSlotCandidate cand
+            (NotOrigin cand, Just latest)
+              | latest > cand -> LatestSlotForecast latest
```
facundominguez, line 243:
```
| latest > sgen + AF.anchorSlot candidateSuffix -> LatestSlotAfterSGen latest
```

facundominguez (commented):
It is looking good. Also looks to me like we need to check explicitly if the last header is after the genesis window. See comments.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -197,52 +216,83 @@ densityDisconnect ::
...
+          lowerBound = fromIntegral $ AF.length fragment
+
+          upperBound = lowerBound + potentialSlots
+
+          -- The number of blocks we know of on the candidate chain, not limited to the Genesis window.
+          totalBlockCount = fromIntegral (AF.length candidateSuffix)
+
+          -- Does the peer have more than k known blocks in _total_ after the intersection?
+          -- If not, it is not qualified to compete by density (yet).
+          offersMoreThanK = case latestSlot of
```
facundominguez, line 281: I'm not sure this plays well with lightweight checkpointing, where the honest chain might have lower densities. Why are peers required to offer more than k blocks?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: the point of `Forecast` is that the block is not on the candidate fragment, so we have to count differently. Even with `LatestSlotCandidate` the last block can be after `sgen`

tek: This is only relevant for `offersMoreThanK`, where a candidate fragment with `k` blocks can be considered to meet the condition because the last slot implies that there is a header outside of the candidate fragment.

The constructor is mostly for readability I guess? Though this conversation suggests that it's not intuitive enough

tek: maybe it should be

```haskell
| LatestSlotBeforeSgen
| LatestSlotAfterSgen { forecast :: Bool }
```

Would you find that more accessible?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -162,52 +216,83 @@ densityDisconnect ::
...
+densityDisconnect (GenesisWindow sgen) (SecurityParam k) candidateSuffixes caughtUpPeers latestSlots loeFrag =
   (losingPeers, densityBounds)
   where
     densityBounds = Map.fromList $ do
       (peer, fragment) <- Map.toList competingFrags
       -- Skip peers that haven't advertised their tip yet.
       -- They should be disconnected by timeouts instead.
-      theirTip <- toList (theirTips Map.!? peer)
+      -- TODO We probably need an alternative approach for this now
+      -- that we don't use the tip anymore
```
amesgen, line 231: If we adopt https://github.com/IntersectMBO/ouroboros-consensus/pull/984#discussion_r1521388834, we could check whether the `lastSlot` for the peer is `Origin`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: > This is only relevant for `offersMoreThanK`

The meaning of `LatestSlotForecast` affects the computation of the upper bound as well.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: That still wouldn't tell us if the peer is done at `Origin` or just uninitialized...

tek: hmm are you thinking of:
* no entry in `Map` -> uninitialized
* `Origin` -> tip was advertised and it is `Origin`
?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: > Would you find that more accessible?

I still don't understand why we would care of the last header being after the forecast horizon. See my comment in `offersMoreThanK`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
amesgen: FTR: `Maybe (WithOrigin SlotNo)` would be expressive enough to capture the three possibilities:

 - `Nothing`: Not initialized (`MsgIntersectFound` not yet received by the server). We should ignore such a peer in this GDD round.
 - `Just s`: Initialized, we can do density disconnections involving this peer.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
amesgen: FTR: When almost caught-up, there might be light disagreements (eg forks of depth 1) even between honest peers due to normal Praos dynamics. Without this criterion, we might disconnect from a peer because it offers just one block after the intersection, compared to two for another peer.

Lightweight checkpointing should be orthogonal, in case of a disaster scenario, we just have to add sufficient checkpoints such that any possible alternative chain is invalid.

facundominguez: In that case, we should restrict the computation of `offersMoreThanK` to the actual headers that have been validated, and not increase it speculatively depending on headers that are after the forecast horizon and therefore weren't checked for checkpoints.


Also, the forecast horizon isn't relevant to the scenario where `offersMoreThanK` is useful.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -162,52 +216,83 @@ densityDisconnect ::
...
+            -- If the last slot is not on the candidate chain, we know that there is at least one more block, so we use
+            -- '>=' here.
+            LatestSlotForecast _  -> totalBlockCount >= k
+
+          -- For tracing: Is there a block after the end of the Genesis window?
+          -- Note that 'endOfGenesisWindow' is exclusive, so we use '>='.
+          hasBlockAfter = case latestSlot of
+            NoLatestSlot             -> False
+            LatestSlotCandidate slot -> slot >= endOfGenesisWindow
+            LatestSlotForecast slot  -> slot >= endOfGenesisWindow
```
facundominguez, line 293: `endOfGenesisWindow` points at the first slot after the genesis window. I feel like this would deserve a renaming to `firstSlotAfterGenesisWindow`. (`endOfGenesisWindow` means to me the last slot of the genesis window, i.e. `intersection_slot + sgen`).

facundominguez:
Rebased `genesis/loe-gdd-2`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -162,52 +214,83 @@ densityDisconnect ::
...
+              -- The peer either has not advertised its tip yet or simply has no blocks whatsoever and won't progress
+              -- either.
+              -- In the latter case, it should be killed by the LoP.
+              -- REVIEW: What do we want to do in tests here? Both cases are possible, and we cannot distinguish them
+              -- without the tip (even with the tip we have been relying on the invariant that each peer advertises the
+              -- tip before any of them sends headers).
+              -- NoLatestSlot -> 0
+              NoLatestSlot -> SlotNo sgen
+              -- 'endOfGenesisWindow' is exclusive, so we have to add 1 to the non-exclusive last slot
+              LatestSlotCandidate slot -> intervalLength endOfGenesisWindow (slot + 1)
```
facundominguez, line 264: I think we want instead `intervalLength (slot + 1) endOfGenesisWindow`.

tek: damn.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Renamed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Both Forecast and Candidate cases now compute the same `totalBlockCount>k`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: I eliminated the distinction between Candidate and Forecast as it wasn't needed for `offersMoreThanK`. This is the current form:
```haskell
data LatestSlot = NoLatestSlot | LatestSlot SlotNo
```
In the case of the computation of the upper bound, it isn't dangerous to use the slots of not-yet-validated headers after the genesis window, as it allows to prune adversarial nodes early.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Renamed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
tek: > The meaning of LatestSlotForecast affects the computation of the upper bound as well.

btw that's just a simplification, because we know that the difference will always be `<= 0`

PR #980, opened 2024-03-04
[WIP] Genesis milestone 11 - GDD Governor

Niols:
WIP PR to get CI on the feature branch.

PR #975, opened 2024-02-29
GSM: split OnlyBootstrap into PreSyncing and Syncing

amesgen:
This PR splits the previous OnlyBootstrap state of the GSM into separate PreSyncing and Syncing states, making it ready to be used for Genesis, while allowing to still use the GSM in conjunction with bootstrap peers.

Compared to the [Bootstrap Peers IER](https://ouroboros-consensus.cardano.intersectmbo.org/docs/for-developers/BootstrapPeersIER), there is one more state and two more transition rules, both based on the Honest Availability Assumption (HHA), which the Network layer will try to establish and inform us about, likely by ensuring a minimum amount of (big) ledger peers, see `isHaaSatisfied` in `GsmView`.

The new state transition diagram looks like this:

```mermaid
graph
    PreSyncing[PreSyncing]
    Syncing[Syncing]
    CaughtUp[CaughtUp]

    PreSyncing -- "Honest Availability Assumption\nis satisfied" --> Syncing
    Syncing -- "Honest Availability Assumption\nis no longer satisfied" --> PreSyncing
    Syncing -- "no peers claim to have\nsubsequent headers,\nand its selection is ≥\nthe best header" --> CaughtUp
    CaughtUp -- "vol tip became older than X" --> PreSyncing

    StartUp[[Node start-up]]
    StartUp -- "node was most recently in CaughtUp\nand vol tip is younger than X" --> CaughtUp
    StartUp -- "otherwise" --> PreSyncing
```

PR #1007, opened 2024-03-19
Make the LoP follow GSM states

Niols:
This PR builds on top of #980 but includes the commits of #975. You might want to read commit-per-commit or you'll also see the changes of the latter.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -42,7 +42,7 @@ import qualified Codec.CBOR.Decoding as CBOR
 import           Codec.CBOR.Encoding (Encoding)
 import qualified Codec.CBOR.Encoding as CBOR
 import           Codec.CBOR.Read (DeserialiseFailure)
-import           Control.Concurrent.Class.MonadSTM.Strict.TVar as TVar.Unchecked
+import qualified Control.Concurrent.Class.MonadSTM.Strict.TVar as TVar.Unchecked
 import           Control.Monad.Class.MonadTime.SI (MonadTime)
 import           Control.Monad.Class.MonadTimer.SI (MonadTimer)
 import           Control.Tracer
@@ -571,6 +571,7 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout lopBucke
             (contramap (TraceLabelPeer them) (Node.chainSyncClientTracer (getTracers kernel)))
             (CsClient.defaultChainDbView (getChainDB kernel))
             (getChainSyncHandles kernel)
+            (getGsmState       kernel)
             them
             version
             lopBucketConfig
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -80,6 +80,8 @@ import           Ouroboros.Consensus.Util.AnchoredFragment
                      (preferAnchoredCandidate)
 import           Ouroboros.Consensus.Util.EarlyExit
 import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Consensus.Util.LeakyBucket
+                     (atomicallyWithMonotonicTime)
 import           Ouroboros.Consensus.Util.Orphans ()
 import           Ouroboros.Consensus.Util.ResourceRegistry
 import           Ouroboros.Consensus.Util.STM
@@ -128,9 +130,10 @@ data NodeKernel m addrNTN addrNTC blk = NodeKernel {
       --
     , getFetchMode            :: STM m FetchMode
 
-      -- | The ledger judgement, used by diffusion.
+      -- | The GSM state, used by diffusion. A ledger judgement can be derived
+      -- from it with 'GSM.gsmStateToLedgerJudgement'.
       --
-    , getLedgerStateJudgement :: STM m LedgerStateJudgement
+    , getGsmState             :: STM m GSM.GsmState
 
       -- | The kill handle and exposed state for each ChainSync client.
     , getChainSyncHandles     :: StrictTVar m (Map (ConnectionId addrNTN) (ChainSyncClientHandle m blk))
@@ -201,7 +204,7 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
           , mempool
           , peerSharingRegistry
           , varChainSyncHandles
-          , varLedgerJudgement
+          , varGsmState
           } = st
 
     do  let GsmNodeKernelArgs {..} = gsmArgs
@@ -237,13 +240,16 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
               , GSM.setCaughtUpPersistentMark = \upd ->
                   (if upd then GSM.touchMarkerFile else GSM.removeMarkerFile)
                     gsmMarkerFileView
-              , GSM.writeGsmState = \x -> atomically $ do
-                  writeTVar varLedgerJudgement $ GSM.gsmStateToLedgerJudgement x
+              , GSM.writeGsmState = \gsmState ->
+                  atomicallyWithMonotonicTime $ \time -> do
+                    writeTVar varGsmState gsmState
+                    handles <- readTVar varChainSyncHandles
+                    traverse_ (($ time) . ($ gsmState) . cschGsmCallback) handles
               , -- In the context of bootstrap peers, it is fine to always
                 -- return 'True' as all peers are trusted during syncing.
                 GSM.isHaaSatisfied            = pure True
               }
-        judgment <- readTVarIO varLedgerJudgement
+        judgment <- GSM.gsmStateToLedgerJudgement <$> readTVarIO varGsmState
         void $ forkLinkedThread registry "NodeKernel.GSM" $ case judgment of
           TooOld      -> GSM.enterPreSyncing gsm
           YoungEnough -> GSM.enterCaughtUp   gsm
@@ -272,7 +278,7 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
       , getTopLevelConfig       = cfg
       , getFetchClientRegistry  = fetchClientRegistry
       , getFetchMode            = readFetchMode blockFetchInterface
-      , getLedgerStateJudgement = readTVar varLedgerJudgement
+      , getGsmState             = readTVar varGsmState
       , getChainSyncHandles     = varChainSyncHandles
       , getPeerSharingRegistry  = peerSharingRegistry
       , getTracers              = tracers
@@ -305,9 +311,9 @@ data InternalState m addrNTN addrNTC blk = IS {
     , blockFetchInterface :: BlockFetchConsensusInterface (ConnectionId addrNTN) (Header blk) blk m
     , fetchClientRegistry :: FetchClientRegistry (ConnectionId addrNTN) (Header blk) blk m
     , varChainSyncHandles :: StrictTVar m (Map (ConnectionId addrNTN) (ChainSyncClientHandle m blk))
+    , varGsmState         :: StrictTVar m GSM.GsmState
     , mempool             :: Mempool m blk
     , peerSharingRegistry :: PeerSharingRegistry addrNTN m
-    , varLedgerJudgement  :: StrictTVar m LedgerStateJudgement
     }
 
 initInternalState ::
@@ -324,9 +330,9 @@ initInternalState NodeKernelArgs { tracers, chainDB, registry, cfg
                                  , mempoolCapacityOverride
                                  , gsmArgs, getUseBootstrapPeers
                                  } = do
-    varLedgerJudgement <- do
+    varGsmState <- do
       let GsmNodeKernelArgs {..} = gsmArgs
-      j <- GSM.initializationLedgerJudgement
+      j <- GSM.initializationGsmState
         (atomically $ ledgerState <$> ChainDB.getCurrentLedger chainDB)
         gsmDurationUntilTooOld
         gsmMarkerFileView
@@ -350,7 +356,7 @@ initInternalState NodeKernelArgs { tracers, chainDB, registry, cfg
           btime
           (ChainDB.getCurrentChain chainDB)
           getUseBootstrapPeers
-          (readTVar varLedgerJudgement)
+          (GSM.gsmStateToLedgerJudgement <$> readTVar varGsmState)
         blockFetchInterface :: BlockFetchConsensusInterface (ConnectionId addrNTN) (Header blk) blk m
         blockFetchInterface = BlockFetchClientInterface.mkBlockFetchConsensusInterface
           (configBlock cfg)
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -74,6 +74,7 @@ module Ouroboros.Consensus.MiniProtocol.ChainSync.Client (
   ) where
 
 import           Control.Monad (join, void)
+import           Control.Monad.Class.MonadTimer (MonadTimer)
 import           Control.Monad.Except (runExcept, throwError)
 import           Control.Tracer
 import           Data.Kind (Type)
@@ -102,6 +103,7 @@ import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.Jumping as Jumping
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.State
+import           Ouroboros.Consensus.Node.GsmState (GsmState (..))
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.Storage.ChainDB (ChainDB,
@@ -113,6 +115,8 @@ import           Ouroboros.Consensus.Util.Assert (assertWithMsg)
 import           Ouroboros.Consensus.Util.EarlyExit (WithEarlyExit, exitEarly)
 import qualified Ouroboros.Consensus.Util.EarlyExit as EarlyExit
 import           Ouroboros.Consensus.Util.IOLike hiding (handle)
+import           Ouroboros.Consensus.Util.LeakyBucket
+                     (atomicallyWithMonotonicTime)
 import qualified Ouroboros.Consensus.Util.LeakyBucket as LeakyBucket
 import           Ouroboros.Consensus.Util.STM (Fingerprint, Watcher (..),
                      WithFingerprint (..), withWatcher)
@@ -313,16 +317,20 @@ deriving anyclass instance (
   NoThunks (Header blk)
   ) => NoThunks (ChainSyncStateView m blk)
 
-bracketChainSyncClient :: forall m peer blk a.
+bracketChainSyncClient ::
+  forall m peer blk a.
     ( IOLike m
     , Ord peer
     , LedgerSupportsProtocol blk
+    , MonadTimer m
     )
  => Tracer m (TraceChainSyncClientEvent blk)
  -> ChainDbView m blk
  -> StrictTVar m (Map peer (ChainSyncClientHandle m blk))
     -- ^ The kill handle and states for each peer, we need the whole map because we
     -- (de)register nodes (@peer@).
+ -> STM m GsmState
+    -- ^ A function giving the current GSM state; only used at startup.
  -> peer
  -> NodeToNodeVersion
  -> ChainSyncLoPBucketConfig
@@ -333,19 +341,21 @@ bracketChainSyncClient
     tracer
     ChainDbView { getIsInvalidBlock }
     varHandles
+    getGsmState
     peer
     version
     csBucketConfig
     csjConfig
     body
-  = mkChainSyncClientHandleState >>= \csHandleState ->
-    withCSJCallbacks csHandleState csjConfig $ \csjCallbacks ->
+  =
+    LeakyBucket.execAgainstBucket'
+  $ \lopBucket ->
+    mkChainSyncClientHandleState >>= \csHandleState ->
+    withCSJCallbacks lopBucket csHandleState csjConfig $ \csjCallbacks ->
         withWatcher
             "ChainSync.Client.rejectInvalidBlocks"
             (invalidBlockWatcher csHandleState)
-      $ LeakyBucket.execAgainstBucket lopBucketConfig
-      $ \lopBucket ->
-            body ChainSyncStateView {
+      $ body ChainSyncStateView {
               csvSetCandidate =
               modifyTVar csHandleState . \ c s -> s {csCandidate = c}
             , csvSetLatestSlot =
@@ -355,9 +365,9 @@ bracketChainSyncClient
               , idlingStop = atomically $ modifyTVar csHandleState $ \ s -> s {csIdling = False}
               }
             , csvLoPBucket = LoPBucket {
-                lbPause = LeakyBucket.setPaused lopBucket True
-              , lbResume = LeakyBucket.setPaused lopBucket False
-              , lbGrantToken = void $ LeakyBucket.fill lopBucket 1
+                lbPause = LeakyBucket.setPaused' lopBucket True
+              , lbResume = LeakyBucket.setPaused' lopBucket False
+              , lbGrantToken = void $ LeakyBucket.fill' lopBucket 1
               }
             , csvJumping = csjCallbacks
             }
@@ -370,34 +380,43 @@ bracketChainSyncClient
         }
 
     withCSJCallbacks ::
+      LeakyBucket.Handlers m ->
       StrictTVar m (ChainSyncState blk) ->
       CSJConfig ->
       (Jumping.Jumping m blk -> m a) ->
       m a
-    withCSJCallbacks cschState CSJDisabled f = do
+    withCSJCallbacks lopBucket cschState CSJDisabled f = do
       tid <- myThreadId
       cschJumpInfo <- newTVarIO Nothing
       cschJumping <- newTVarIO (Disengaged DisengagedDone)
       let handle = ChainSyncClientHandle {
               cschGDDKill = throwTo tid DensityTooLow
+            , cschGsmCallback = updateLopBucketConfig lopBucket
             , cschState
             , cschJumping
             , cschJumpInfo
             }
-          insertHandle = atomically $ modifyTVar varHandles $ Map.insert peer handle
+          insertHandle = atomicallyWithMonotonicTime $ \time -> do
+            initialGsmState <- getGsmState
+            updateLopBucketConfig lopBucket initialGsmState time
+            modifyTVar varHandles $ Map.insert peer handle
           deleteHandle = atomically $ modifyTVar varHandles $ Map.delete peer
       bracket_ insertHandle deleteHandle $ f Jumping.noJumping
 
-    withCSJCallbacks csHandleState (CSJEnabled csjEnabledConfig) f =
-      bracket (acquireContext csHandleState csjEnabledConfig) releaseContext $ \peerContext ->
+    withCSJCallbacks lopBucket csHandleState (CSJEnabled csjEnabledConfig) f =
+      bracket (acquireContext lopBucket csHandleState csjEnabledConfig) releaseContext $ \peerContext ->
         f $ Jumping.mkJumping peerContext
-    acquireContext cschState (CSJEnabledConfig jumpSize) = do
+
+    acquireContext lopBucket cschState (CSJEnabledConfig jumpSize) = do
         tid <- myThreadId
-        atomically $ do
+        atomicallyWithMonotonicTime $ \time -> do
+          initialGsmState <- getGsmState
+          updateLopBucketConfig lopBucket initialGsmState time
           cschJumpInfo <- newTVar Nothing
           context <- Jumping.makeContext varHandles jumpSize
           Jumping.registerClient context peer cschState $ \cschJumping -> ChainSyncClientHandle
             { cschGDDKill = throwTo tid DensityTooLow
+            , cschGsmCallback = updateLopBucketConfig lopBucket
             , cschState
             , cschJumping
             , cschJumpInfo
@@ -409,19 +428,33 @@ bracketChainSyncClient
         invalidBlockRejector
             tracer version getIsInvalidBlock (csCandidate <$> readTVar varState)
 
+    -- | Update the configuration of the bucket to match the given GSM state.
+    -- NOTE: The new level is currently the maximal capacity of the bucket;
+    -- maybe we want to change that later.
+    updateLopBucketConfig :: LeakyBucket.Handlers m -> GsmState -> Time -> STM m ()
+    updateLopBucketConfig lopBucket gsmState =
+      LeakyBucket.updateConfig lopBucket $ \_ ->
+        let config = lopBucketConfig gsmState in
+          (LeakyBucket.capacity config, config)
+
     -- | Wrapper around 'LeakyBucket.execAgainstBucket' that handles the
     -- disabled bucket by running the given action with dummy handlers.
-    lopBucketConfig :: LeakyBucket.Config m
-    lopBucketConfig =
-      case csBucketConfig of
-        ChainSyncLoPBucketDisabled -> LeakyBucket.dummyConfig
-        ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig {csbcCapacity, csbcRate} ->
+    lopBucketConfig :: GsmState -> LeakyBucket.Config m
+    lopBucketConfig gsmState =
+      case (gsmState, csBucketConfig) of
+        (Syncing, ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig {csbcCapacity, csbcRate}) ->
             LeakyBucket.Config
               { capacity = fromInteger $ csbcCapacity,
                 rate = csbcRate,
                 onEmpty = throwIO EmptyBucket,
                 fillOnOverflow = True
               }
+        -- NOTE: If we decide to slow the bucket down when “almost caught-up”,
+        -- we should add a state to the GSM and corresponding configuration
+        -- fields and a bucket config here.
+        (_, ChainSyncLoPBucketDisabled) -> LeakyBucket.dummyConfig
+        (PreSyncing, ChainSyncLoPBucketEnabled _) -> LeakyBucket.dummyConfig
+        (CaughtUp, ChainSyncLoPBucketEnabled _) -> LeakyBucket.dummyConfig
 
 -- Our task: after connecting to an upstream node, try to maintain an
 -- up-to-date header-only fragment representing their chain. We maintain
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
@@ -30,18 +30,30 @@ module Ouroboros.Consensus.Util.LeakyBucket (
     Config (..)
   , Handlers (..)
   , State (..)
+  , atomicallyWithMonotonicTime
   , diffTimeToSecondsRational
   , dummyConfig
   , evalAgainstBucket
   , execAgainstBucket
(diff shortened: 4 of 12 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -244,17 +252,28 @@ bracketChainSyncClient
...
               }
+        -- NOTE: If we decide to slow the bucket down when “almost caught-up”,
+        -- we should add a state to the GSM and corresponding configuration
+        -- fields and a bucket config here.
+        _ -> LeakyBucket.dummyConfig
+
+    -- | Follow changes in the GSM state by updating the configuration of the
+    -- LoP bucket.
+    lopBucketGsmUpdater :: GsmState -> (LeakyBucket.Config m -> m ()) -> m ()
+    lopBucketGsmUpdater prevGsmState setConfig = do
```
amesgen, line 273: Hmm, we might be able to avoid blocking on GSM state changes (although they shouldn't be too frequent) by passing in the `LeakyBucket` from outside, and also updating the bucket config in the callback invoked by the GSM logic.

amesgen:
We also need to (maybe in another PR) disable the LoE and the GDD based on the GSM.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: That's exactly the kind of comments I was looking for. I was wondering which approach to take. It seemed to me that passing the bucket to the GSM was the more natural one but broke the scoping of the LoP being local to the ChainSync client. I suppose we could also have something like the varCandidates but for the buckets, where each peer registers their bucket in a TVar and the GSM just updates them all at once.

Niols:
@amesgen The last commit breaks the LoP and I have yet to figure out why. I suppose the function `updateConfig` of the leaky bucket is broken in a way. But anyway, do you like this other approach better on principle?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -195,6 +197,12 @@ bracketChainSyncClient ::
     -- It's more important that the client is removed from the set promptly
     -- than it is for the client to be added promptly, because of how this is
     -- used by the GSM to determine that the node is done syncing.
+ -> (STM m GsmState,
+     StrictTVar m (Map peer (GsmState -> m ())))
```
amesgen, line 201: Yeah, this looks good; with #1019, it seems natural to store these callbacks (or maybe (equivalently) the LoP buckets directly) in `ChainSyncClientHandle`/`ChainSyncState` as introduced there.

Niols:
Rebased on top of #1033; will now proceed to integrate into @tek's `ChainSyncState`s.

Niols:
Done in 5807d4cb27d75ab5d0327812c8ff5551b2c899ae.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Done in 5807d4cb27d75ab5d0327812c8ff5551b2c899ae.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -42,7 +42,8 @@ import qualified Codec.CBOR.Decoding as CBOR
 import           Codec.CBOR.Encoding (Encoding)
 import qualified Codec.CBOR.Encoding as CBOR
 import           Codec.CBOR.Read (DeserialiseFailure)
-import           Control.Concurrent.Class.MonadSTM.Strict.TVar as TVar.Unchecked
+import           Control.Concurrent.Class.MonadSTM.Strict.TVar as TVar.Unchecked hiding
+                     (modifyTVar)
```
amesgen, line 46: I think this should have been a `qualified` import in the first place, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -329,6 +335,8 @@ bracketChainSyncClient ::
  -> StrictTVar m (Map peer (ChainSyncClientHandle m blk))
     -- ^ The kill handle and states for each peer, we need the whole map because we
     -- (de)register nodes (@peer@).
+ -> STM m GsmState
+    -- ^ A function giving the current GSM state; only used at startup.
```
amesgen, line 339: If it is only used once, shouldn't we take a `GsmState` instead of a `STM m GsmState`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -348,8 +357,13 @@ bracketChainSyncClient
         withWatcher
             "ChainSync.Client.rejectInvalidBlocks"
             (invalidBlockWatcher varState)
-      $ LeakyBucket.execAgainstBucket lopBucketConfig
-      $ \lopBucket ->
+      $ LeakyBucket.execAgainstBucket'
+      $ \lopBucket -> do
+            -- NOTE: Updating the bucket configuration and registering the
+            -- callback should be in the same STM transaction.
```
amesgen, line 363: Updating the bucket configuration is not an `STM` action, so how can it be done "in the same STM transaction" as something else? Is this about changing `updateLopBucketConfig` such that it *is* in STM?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -376,6 +390,7 @@ bracketChainSyncClient
         atomically $ do
           modifyTVar varHandles $ Map.insert peer ChainSyncClientHandle {
               cschGDDKill = throwTo tid DensityTooLow
+            , cschGsmCallback = const (pure ()) -- will be updated later
```
amesgen, line 393: Hmm, couldn't we directly initialize it here?

amesgen (commented):
Nice!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: We could, but if we do I'm afraid that the `GsmState` may change between the time we call `bracketChainSyncClient` and the time we register the callback, and then we would miss a `GsmState` change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: That is what I mean, indeed. I would like `updateLoPBucketConfig` to be in `STM` and that would allow the code here to be just one STM transaction which I find more robust.

Niols: That is what I mean, indeed. I would like `updateLoPBucketConfig` to be in `STM` and that would allow the code here to be just one STM transaction which I find more robust.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: The callback works on the LoP bucket which is not available at that point of the execution; but I agree that it is not super clean.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
Niols: Indeed. Fixed in 2d434fed7d5c31f63abe896224a63974782f22ea. It's a bit puzzling to me why there is both STM's TVars and io-classes's TVars in the same project, but that's out of the scope of this PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: I tried to improve the comment in 40a04720cf92814c7bd081c9e7f0f0b6050bcef8.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -348,8 +357,14 @@ bracketChainSyncClient
...
-      $ LeakyBucket.execAgainstBucket lopBucketConfig
-      $ \lopBucket ->
+      $ LeakyBucket.execAgainstBucket'
+      $ \lopBucket -> do
+            -- NOTE: If `updateLopBucketConfig` was in STM, then all the
+            -- following could be in the same STM transaction which would make
+            -- it significantly more robust.
+            initialGsmState <- atomically getGsmState
+            updateLopBucketConfig lopBucket initialGsmState
+            atomically $ modifyTVar varHandles $ Map.update (\csch -> Just $! csch{cschGsmCallback = updateLopBucketConfig lopBucket}) peer
```
amesgen, line 367: I think it would be worth to get this right; races in this area seem like a horrifying surprise in the future. AFAICT, the non-STM stuff is: 

 1. `getMonotonicTime` at the very beginning.

    This can probably be solved by sth like `m (STM m ())` as we talked about previously (the `m` takes the time and returns an `STM` transaction that captured it). This relies on the fact that the transaction will complete promptly, but that is not new, the current implementation of `snapshotFill` already relies on that.
 2. Potentially stopping/starting the background thread.

    Maybe this could be done by an action *returned* by the `STM` transaction?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Eg like this:
```diff
diff --git a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
index 8a0c1d57c0..63f334a4f0 100644
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -352,19 +352,13 @@ bracketChainSyncClient
     csBucketConfig
     body
   =
-    bracket acquireHandle releaseHandle
+    LeakyBucket.execAgainstBucket'
+  $ \lopBucket ->
+    bracket (acquireHandle lopBucket) releaseHandle
   $ \varState ->
         withWatcher
             "ChainSync.Client.rejectInvalidBlocks"
-            (invalidBlockWatcher varState)
-      $ LeakyBucket.execAgainstBucket'
-      $ \lopBucket -> do
-            -- NOTE: If `updateLopBucketConfig` was in STM, then all the
-            -- following could be in the same STM transaction which would make
-            -- it significantly more robust.
-            initialGsmState <- atomically getGsmState
-            updateLopBucketConfig lopBucket initialGsmState
-            atomically $ modifyTVar varHandles $ Map.update (\csch -> Just $! csch{cschGsmCallback = updateLopBucketConfig lopBucket}) peer
+            (invalidBlockWatcher varState) $ do
             body ChainSyncStateView {
               csvSetCandidate =
               modifyTVar varState . \ c s -> s {csCandidate = c}
@@ -381,17 +375,22 @@ bracketChainSyncClient
               }
             }
   where
-    acquireHandle = do
+    acquireHandle lopBucket = do
         cschState <- newTVarIO $ ChainSyncState {
             csCandidate = AF.Empty AF.AnchorGenesis
           , csLatestSlot = Nothing
           , csIdling = False
           }
         tid <- myThreadId
+        -- NOTE: If `updateLopBucketConfig` was in STM, then all the
+        -- following could be in the same STM transaction which would make
+        -- it significantly more robust.
+        initialGsmState <- atomically getGsmState
+        updateLopBucketConfig lopBucket initialGsmState
         atomically $ do
           modifyTVar varHandles $ Map.insert peer ChainSyncClientHandle {
               cschGDDKill = throwTo tid DensityTooLow
-            , cschGsmCallback = const (pure ()) -- will be updated later
+            , cschGsmCallback = updateLopBucketConfig lopBucket
             , cschState
             }
         pure cschState
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -386,19 +402,31 @@ bracketChainSyncClient
...
     -- | Wrapper around 'LeakyBucket.execAgainstBucket' that handles the
     -- disabled bucket by running the given action with dummy handlers.
-    lopBucketConfig :: MonadThrow m => LeakyBucket.Config m
-    lopBucketConfig =
-      case csBucketConfig of
-        ChainSyncLoPBucketDisabled -> LeakyBucket.dummyConfig
-        ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig {csbcCapacity, csbcRate} ->
+    lopBucketConfig :: GsmState -> LeakyBucket.Config m
+    lopBucketConfig gsmState =
+      case (gsmState, csBucketConfig) of
```
amesgen, line 418: Can we exhaustively match on these to make sure we don't forget do update this when there are eg new `GSMState`s?

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -497,6 +498,7 @@ runChainSync skew securityParam (ClientUpdates clientUpdates)
                  chainSyncTracer
                  chainDbView
                  varHandles
+                 (pure Syncing)
```
amesgen, line 501: Maybe leave a comment this this only affects the LoP bucket, which is disabled here, so the choice is irrelevant?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Right, it might make races a bit less likely, but the current code also doesn't guarantee race freedom as discussed in another thread.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -226,13 +227,16 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
               , GSM.setCaughtUpPersistentMark = \upd ->
                   (if upd then GSM.touchMarkerFile else GSM.removeMarkerFile)
                     gsmMarkerFileView
-              , GSM.writeGsmState = \x -> atomically $ do
-                  writeTVar varLedgerJudgement $ GSM.gsmStateToLedgerJudgement x
+              , GSM.writeGsmState = \gsmState -> do
+                  handles <- atomically $
+                    writeTVar varGsmState gsmState >>
+                    readTVar varChainSyncHandles
+                  traverse_ (($ gsmState) . cschGsmCallback) handles
```
amesgen, line 234: Consider the scenario where we disconnect from a peer in-between fetching the handles  and executing the GSM callbacks. Wouldn't that mean that we might start the bucket background thread again? Doesn't seem super severe, but still a bit weird.

(Also, we don't seem to have any tests for `updateConfig`, and the implementation doesn't seem that thread-safe/async-exception-safe at a cursory glance. Maybe a `runSimStrictShutdown` test could have caught this?)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
amesgen: Yeah, that's weird, note that it is "normal" TVars from io-classes, and TVars with NoThunks checks from strict-checked-vars.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Done in e038cc87378b1afa94fa3d3cd49b0f50eb4cb1cd.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Done in 3890e9bf6b0df74d3a0faff17f7beb703b16a799.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Yes, I was wondering about that one. I was just not sure what that would mean for exceptions thrown before `bracket` has fully “installed” itself. But probably that just will not happen because the bucket will not run out that fast and there shouldn't be other exceptions that are not bugs.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Wouldn't the best solution here to give up on pure STM endpoints (because that will just not happen and I'm not sure how to make sure we won't have races with the thread starting and stopping, for instance). We could use another kind of semaphore, though, like `putMVar` before all this and `takeMVar` after all this?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
Niols: > Consider the scenario where we disconnect from a peer in-between fetching the handles and executing the GSM callbacks. Wouldn't that mean that we might start the bucket background thread again? Doesn't seem super severe, but still a bit weird.

I agree both that it wouldn't be the worst but that it's weird. Yet another place where I'd want everything to happen in STM or in something thread-safe.

> (Also, we don't seem to have any tests for updateConfig, and the implementation doesn't seem that thread-safe/async-exception-safe at a cursory glance. Maybe a runSimStrictShutdown test could have caught this?)

Oh, I was sure there was such a test. Let me fix that.

Niols: You're right; definitely not thread-safe, and the tests in f4539402d418906b1c65e4c6c81257592be6be3e fail precisely because of races.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Yeah, using non-STM synchronization also sounds fine, anything that makes sure that there are no races if the GSM state changes at almost the same time we are registering a new ChainSync peer.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: The bucket won't actually leak until we potentially change the config in the `bracket` allocate part, as you use `execAgainstBucket'` which uses `dummyConfig`, right?

Niols:
So, for the time, I am still very disappointed that I can't read the time in an STM transaction, but maybe we can go the `m (STM m ())` route and that's fine. For the starting and stopping threads, we could instead always have a thread running (maybe started on demand the first time) that reads a TMVar to know whether it should actually do something or not. That would be a way to “pause” the thread when need be, and that could all be done from STM.

Alternatively, we could go for a lock mechanism that would be acquired and released in STM but where whatever happens in between could run in IOLike.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: That's a fair point; now that we're using `execAgainstBucket'` it must be safe (and it probably was almost always safe before anyways).

amesgen:
> Which one do you prefer, @amesgen? Which one should I pursue?

I don't have a strong a priori opinion, might start with the `STM` approach and if that seems to brittle/invasive switch to using an explicit lock/`MVar`.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
Niols: Fixed in https://github.com/IntersectMBO/ouroboros-consensus/pull/1007/commits/dc6ca25347f2984df61e495056e3983764aa1b23 à la https://github.com/IntersectMBO/ouroboros-consensus/pull/1007#issuecomment-2070080993.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Fix in https://github.com/IntersectMBO/ouroboros-consensus/pull/1007/commits/dc6ca25347f2984df61e495056e3983764aa1b23 à la https://github.com/IntersectMBO/ouroboros-consensus/pull/1007#issuecomment-2070080993.

Niols:
I am now working to get the tests to pass. I think I made good progress: things used to crash immediately and then now need hundreds of test cases before finding things. Still, not fully there yet. Hopefully it will be fine tomorrow!

Niols:
As of the last commit, the implementation of the leaky bucket seems robust enough to pass the tests correctly. This involved quite a few changes, but I think the quality improved drastically with them. @amesgen WDYT of the current state of things?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
Niols: Done in 18140d8b298d6219be52512ad1efe3752d9dc102.

amesgen:
FTR: This needs to be rebased

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -242,24 +347,24 @@ snapshot bucket = fst <$> snapshotFill bucket 0
 -- need to measure the time, and @io-classes@'s STM does not allow running IO in
 -- an STM.
 snapshotFill ::
-  (MonadSTM m, MonadMonotonicTime m) =>
+  ( MonadSTM m
```
amesgen, line 350: The comment of this function should now be mostly addressed, right?

Niols: Indeed! Removed in 3f590a07ed5d4222147f1a2c4f3cf7fc6309aae8.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -191,42 +271,67 @@ init ::
...
+          -- We have run the action on empty, there is nothing left to do,
+          -- unless someone changes the configuration.
+          void $ atomically $ blockUntilChanged configGeneration oldConfigGeneration $ readTVar bucket
+          go
+        else do
+          -- Wait for the bucket to empty, or for the thread to be stopped or
+          -- restarted.
+          race_
+            (threadDelay timeToWait)
+            (void $ atomically $ blockUntilChanged id (Just oldRunThread) $ tryReadTMVar runThreadVar)
```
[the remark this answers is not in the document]
Niols: Good point! Feels to me like we should go for [`timeout`](https://hackage.haskell.org/package/base-4.18.1.0/docs/System-Timeout.html#v:timeout) instead, though? (In `io-classes`'s `MonadTimer`.) Which would give:
```haskell
void $ timeout timeToWait $
  blockUntilChanged ...
```

amesgen:
> > FTR: This needs to be rebased
> 
> Indeed; I will take care of this now. Must I understand from the absence of comments from you that you like this approach? Or have you just not reviewed the whole thing yet?

I ran out of time to fully finalize a review, but it definitely looks good :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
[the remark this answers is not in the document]
amesgen: That would likely also have slightly more overhead as it also involves a `forkIO` for sending an async exception to the main thread on timeout.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -126,62 +191,77 @@ runAgainstBucket ::
...
+    killThread = uninterruptibleCancel
 
-    stopThread :: StrictTVar m (Maybe (m ())) -> m ()
-    stopThread killThreadVar =
-      readTVarIO killThreadVar >>= \case
-        Just killThread' -> killThread'
-        Nothing -> pure ()
+    -- Start the thread (that is, write to its 'runThreadVar') if it is useful.
+    -- Takes a potential old value of the 'runThreadVar' as first argument,
+    -- which will be increased to help differentiate between restarts.
```
amesgen, line 226: From my understanding, the background thread will now always be running, right? So "restart" might be a bit misleading here. Maybe there could be a high-level comment somewhere explaining the lifecycle and state flow of the background thread?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -126,62 +191,77 @@ runAgainstBucket ::
...
-                setPaused = setPaused bucket,
-                updateConfig = updateConfig killThreadVar bucket tid
-              }
-        state <- snapshot bucket
-        pure (state, result)
-    )
-    (stopThread killThreadVar)
+  bracket
+    (spawnThreadAndMaybeStart runThreadVar bucket)
+    killThread
```
amesgen, line 200: Couldn't we now use `withAsync` or even `withRegistry` here instead of the "low-level" operations `async` and `uninterruptibleCancel`?

amesgen (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: Absolutely! Done in 09665f2de91dc3e8ad6e7439b0263fa0e7c455ae.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -126,62 +191,77 @@ runAgainstBucket ::
...
+      State
+        { level = oldLevel,
+          paused,
+          configGeneration = oldConfigGeneration,
+          config = oldConfig
+        } <- snapshot bucket time
+      let (newLevel, newConfig) = f (oldLevel, oldConfig)
+          Config {capacity = newCapacity} = newConfig
+          -- Ensure that 0 <= newLevel <= newCapacity
+          newLevel' = max 0 (min newLevel newCapacity)
```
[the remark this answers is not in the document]
Niols: Nice; thanks! Done in d1055c8bf8d1a8d425fd08d882bc69a77aebd31f.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: This is correct, the thread is now always running. I'm not super satisfies with this solution but I can't think of any solution that fits all our criteria. It is also a thread that mostly does nothing, so I don't think it's such a big deal.

I'm fine getting rid of “start”/“restart” as vocabulary for writing to the TMVar and rewriting to the TMVar, but I don't have any other ideas right now :p Where would you want the comment in question? More details in the description of `leak`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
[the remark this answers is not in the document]
Niols: Good point. `io-classes` doesn't seem to give me the instance for `Alternative (STM m)` though; did I miss something or is it not there?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
amesgen: > This is correct, the thread is now always running. I'm not super satisfies with this solution but I can't think of any solution that fits all our criteria. It is also a thread that mostly does nothing, so I don't think it's such a big deal.

I think it is fine, it shouldn't be *that* expensive and it indeed mostly sleeps to occasionally wake up and to trivial arithmetic :+1:

> I'm fine getting rid of “start”/“restart” as vocabulary for writing to the TMVar and rewriting to the TMVar, but I don't have any other ideas right now :p Where would you want the comment in question? More details in the description of `leak`?

It could be a standalone top-level comment, like a [GHC-style note](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/coding-style#21-notes-the-basic-idea).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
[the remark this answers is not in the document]
amesgen: Oh right, surprising that it doesn't provide it (might create a PR adding it). You can `orElse`, that's what `<|>` is implemented with.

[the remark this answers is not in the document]
Niols: Hum, this requires propagating a `MonadTimer m` constraint everywhere including in the ChainSync client and everything that uses it. Apparently that is not in IOLike?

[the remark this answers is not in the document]
amesgen: I think it would be fine to add it to `IOLike`, it already includes `MonadDelay` which contains `threadDelay`, and one can implement `registerDelay` and `timeout` already using that, just less efficiently.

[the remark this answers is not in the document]
amesgen: (If that requires too much yak shaving, feel free to keep eg the `race` version + a TODO in a comment.)

amesgen (approved):
Looks great, not sure though whether we want to merge this into the milestone branch right now, but I also don't see a big reason not to, as it shouldn't be hard to rebase on top of this.

> You might want to have another look at the changes to `bracketChainSyncClient`; they are now very different.

FTR: I was first confused as I thought you changed sth there in the most recent force pushes today; but I think you referred to what you changed last week, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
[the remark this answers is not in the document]
Niols: It was not that bad, actually. Done in 2b0fdd3215e08df4a68e75b4aad5a4aac640dbcd. Does this look good to you?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: Tried in 1d6b842f0dc982e50e318a2cbfef74af7435a2aa. WDYT?

Niols:
> > You might want to have another look at the changes to `bracketChainSyncClient`; they are now very different.
> 
> FTR: I was first confused as I thought you changed sth there in the most recent force pushes today; but I think you referred to what you changed last week, right?

Sorry, I just meant that `bracketChainSyncClient` had changed quite a lot on milestone 12 and that therefore this last rebase was more of a re-implementation. I thought you maybe wanted to check how it was done after the rebase.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
amesgen: Looks great, thanks :+1:

Niols:
I'm not exactly sure why the CI is red; the actions are describes as cancelled? There does not seem to be failed tests, though.

amesgen:
> I'm not exactly sure why the CI is red; the actions are describes as cancelled? There does not seem to be failed tests, though.

I think some tests now have a deadlock; I can locally reproduce this for eg `wait almost too long`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -367,7 +368,8 @@ leak runThreadVar actionThreadId bucket = go
           go
         else do
           -- Wait for the bucket to empty, or for the thread to be stopped or
-          -- restarted.
+          -- restarted. Beware not to call 'registerDelay' with argument 0, that
+          -- is ensure that @timeToWait >= microsecondsPerSecond@.
           varTimeout <- registerDelay $ diffTimeToMicrosecondsAsInt timeToWait
```
amesgen, line 373: Hmm, why `registerDelay` (before 5abc38f56150d3a5cce7b98b30c650b44c74600b) ever be called with 0 as its argument? Wouldn't we be in the first case of the `if` above (as its condition contains `|| timeToWait == 0`)? Was it a rounding thing (in which case we should probably check `diffTimeToMicrosecondsAsInt timeToWait == 0`)?

Niols: `DiffTime` has picosecond precision while `registerDelay` only has microsecond precision, so even when `timeToWait /= 0` there islots of room for `diffTimeToMicrosecondsAsInt timeToWait == 0`. I suppose we could check the latter indeed.

Niols:
Rebased on top of more recent commits and added 1970291542a1f4fa3ec4c4bf2604dcc3bccf4b45.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: >  which case we should probably check `diffTimeToMicrosecondsAsInt timeToWait == 0`

1970291542a1f4fa3ec4c4bf2604dcc3bccf4b45

amesgen (approved):

PR #1010, opened 2024-03-20
Introduce classifiers for Genesis test results

nbacquey:
Similarly to Genesis test input classifiers, this PR introduces classifiers for Genesis test _results_.


facundominguez (approved):
LGTM

amesgen (approved):
Looks good! One thought: Do we already fail or at least log a warning if an adversary is killed by an exception we didn't expect?

PR #1012, opened 2024-03-21
Generate k+1 headers in each Genesis window

nbacquey:
This PR modifies the test tree generation process to ensure that each Genesis window contains at least `k+1` blocks (instead of `k` previously)


facundominguez:
Notice the redundancy with #1013 which additionally adjusts the tests to demand at least k+1 blocks, simplifies the computation of the minimum length of schemas (which changes from `2*s+d+1` to `s+d+k`), adjusts the minimum allowed value of k (which can be set to 1 now), and fixes the delta mutation in tests.

facundominguez:
Closing in favor of #1013.

PR #1013, opened 2024-03-21
Ensure that any stability windows of the honest schema has k+1 active slots

facundominguez:
Now that GDD expects every stability window to have k+1 blocks, implement the Extended Praos Chain Growth assumption. The computation of the minimum schema length has been simplified accordingly from `2*s+d+1` to `s+d+k+1`. Now setting `k=1` is allowed.

---

A bug was found in delta mutation of mutation tests of generation of the adversarial schemas. The comments in the code have the details, but the gist of it is that the increment for the delta mutation depends on what the honest schema is.


facundominguez:
@nfrisby, please review these changes at your convenience. I'm merging for the sake of simplifying the integration of ongoing work, and I'll revert it if I made a catastrophic mistake.

PR #1015, opened 2024-03-22
Implement the GDD governor

facundominguez:
This PR builds upon #979 to provide the GDD governor, the component responsible for identifying and disconnecting peers offering sparser chains than the best. This has the effect of unblocking the Limit on Eagerness, since removing disagreeing peers allows the current selection to advance.

The GDD governor runs in a background thread and evaluates candidate chains whenever they change, or whenever a peer claims to have no more headers, or whenever a peer starts sending headers beyond the forecast horizon.

Whenever GDD disconnects peers, the chain selection is updated.

The implementation of the GDD governor is in commit 3fc6d28f00b0a1b0d3187cb7916f1e192dd79fcb. A range of supporting changes is provided as well.

-------------
Peer simulator improvements in f02d8df9e7b96646a239485590b89cc344f27197 include:
* Our simulator of peers now kills all miniprotocol threads whenever one of the miniprotocols is terminated. The GDD governor might terminate a ChainSync client, and this termination should also cause the termination of the BlockFetch client.
* Our tests now also check that no threads are left lingering when the peer simulator is terminated.
* Traces now print a timestamp when events occur between the start of consecutive ticks. This need arose when Limit on Patience started disconnecting peers because of exhausted buckets, and these events could happen at any moment between ticks. See [this comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/990#issuecomment-2007619414) of #990 for an example trace.

-------------
Post-GDD improvements in ea12e12eeb73b5feff55c354db764672813bbd18 include:
* Started tracing the information about BlockFetch timeouts and BlockFetch logic decisions. This information became relevant when we had to analyze why blocks wouldn't be downloaded from the honest peer, or why a quiet adversarial peer wasn't being disconnected.
* Have the Limit on Eagerness recalculated when the change selection changes. Limit on Eagerness does not depend on the chain selection, but this mechanism has the desired side effect of retriggering chain selection in succession until no more blocks are added to it. This change is necessary after chain selection stopped being idempotent in 7ace045be64710e15b6d29cc10df3d81b4863adc (later on this same PR). This change is included here to demonstrate that the GDD governor prevents the node from being leashed, but otherwise we intend to recover idempotency of chain selection in a later PR.

-------------
Tests for the GDD in b8af4b40f715e00c5f01d16bf57e5e3e9fcf9f4a include:
* Added a new test to check that GDD actually disconnects nodes in small examples. This test is subsumed by the already existing test "serve adversarial branches", but it tries only small peer schedules which are easier to approach on test failures.
* Enabled one of the leashing attack tests. This test was waiting for the GDD governor to be implemented. It generates an honest peer schedule for every peer, and then drops from the adversary schedules some ticks chosen at random. This has the effect of having the adversaries introduce delays to send headers or blocks that shouldn't stop the node from reaching the honest tip by the end of the test.

There is one other leashing attack test called "time limited leashing attack", which it is enabled in [a follow up PR](https://github.com/IntersectMBO/ouroboros-consensus/pull/1035).

-------------
The classifiers added in cfefa6cc3fc6a1e2078b1fba0776e2c106e4fe2e keep track of whether peers are being disconnected by the GDD governor, and the LoP, and timeouts. It also keep tracks of adversaries who did a rollback of their chain. The commit also implements a check ensuring no peer is killed because of an unforeseen exception.

-------------
Chain generators implement the extended praos chain growth assumption (EPCG) in d8c6546e145c2e88db17d2286639c5dd0c4d6560. This causes the honest chain to have at least `k+1` blocks in every stability window. The former Praos Chain Growth assumption required at least `k`, and in practice chains with less than `k+1` blocks in a stability window are still very rare. The EPCG assumption makes a difference to testing the GDD governor.

We cannot let the almost caught up GDD governor disconnect honest peers which disagree on a few final blocks. Therefore, we ask the governor to require a disagreement of more than `k` blocks before deciding a disconnection. But asking for more than `k` blocks without the EPCG assumption means that the GDD governor might not be able to make progress if the `k+1`st block is after the first stability window after the intersection and after the forecast horizon. In this case, the candidate fragments wouldn't be extended with the `k+1`st block, and the GDD governor wouldn't realize that there is a discrepancy of more than `k` blocks.

-------------
In c8468c28f49aea116116f5676ff931ce3e39eb3c we don't allow the chain selection to fork more than k blocks from the honest chain. The honest chain is considered to be the one containing the first intersection at which candidate fragments fork into different chains.

It used to be the case that we allowed the selection to choose up to k blocks after the intersection. But it turned out that if the selection tip is still X blocks behind the intersection, allowing chain selection to pick `X+k` blocks, there is no guarantee that the selection will only add blocks from the candidate fragments. It could select more than `k` blocks of an older fork in the chain DB, and this is catastrophic because now the immutable tip would leave the honest chain.

-------------
bf3f09a3a40deb5ffbbd168e576ba23e7af5cc0c refactors and properly bundles the various state that a ChainSync client maintains.


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
```diff
@@ -278,3 +298,80 @@ prop_densityDisconnectMonotonic =
     gen = do
       gt <- genChains (QC.choose (1, 4))
       evolveBranches (initCandidates gt)
+
```
dnadales, line 317: Minor: unnecessary extra newline?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM.hs
```diff
@@ -81,34 +80,32 @@ semantics vars cmd = pre $ case cmd of
...
+            v <- newTVar $! PeerState (Candidate (b + bdel)) (Idling False)
+            modifyTVar varStates $ Map.insert peer v
         pure Unit
     ReadGsmState -> do
         fmap ReadThisGsmState $ atomically $ readTVar varGsmState
     ReadMarker -> do
         fmap ReadThisMarker $ atomically $ readTVar varMarker
-    StartIdling peer -> do
-        atomically $ modifyTVar varIdlers $ Set.insert peer
+    StartIdling peer -> atomically $ do
```
dnadales, line 98: Would it make sense to guard against the case in which `peer` is not in the map?

ouroboros-consensus-cardano/changelog.d/20240328_165135_alexander.esgen_milestone_11_gdd_governor_squashed.md
```diff
@@ -0,0 +1,7 @@
+### Non-Breaking
+
+- Set Genesis window for Byron and Shelley.
```
dnadales, line 3: We should either use past tense here, or present in the other fragments :)

amesgen: I think "Set ..." is conveniently both present and past tense :sweat_smile:

PR #1023, opened 2024-03-26
Replace `uncheckedNewTVarM` in leaky bucket by `newTVarIO`

Niols:


Niols:
See instead https://github.com/IntersectMBO/ouroboros-consensus/pull/1024 that targets https://github.com/IntersectMBO/ouroboros-consensus/pull/979.

PR #1024, opened 2024-03-26
Replace `uncheckedNewTVarM` in leaky bucket by `newTVarIO`

Niols:


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -37,14 +41,16 @@ module Ouroboros.Consensus.Util.LeakyBucket (
...
-                     (MonadAsync (Async, async, uninterruptibleCancel),
+                     (MonadAsync (async, uninterruptibleCancel),
                      MonadCatch (handle), MonadDelay (threadDelay),
                      MonadFork (throwTo), MonadMask, MonadMonotonicTime,
                      MonadSTM, MonadThread (ThreadId, myThreadId),
                      MonadThrow (finally), SomeException, StrictTVar, Time,
-                     atomically, diffTime, getMonotonicTime, readTVar,
-                     readTVarIO, uncheckedNewTVarM, writeTVar)
+                     atomically, diffTime, getMonotonicTime, newTVarIO,
+                     readTVar, readTVarIO, writeTVar)
```
amesgen, line 53: We usually import this without an explicit import list, feel free to also do that if you are annoyed by the churn.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -123,43 +132,43 @@ evalAgainstBucket config action = fst <$> runAgainstBucket config action
...
   (Handlers m -> m a) ->
   m (State (Config m), a)
 runAgainstBucket config action = do
   bucket <- init config
   tid <- myThreadId
-  thread <- uncheckedNewTVarM Nothing
-  startThread thread bucket tid
+  killThreadVar <- newTVarIO Nothing
+  startThread killThreadVar bucket tid
   finally
```
amesgen, line 144: Not really related to this PR, and very minor, but this isn't async exception safe if the exception is thrown just after `startThread killThreadVar bucket tid`, but before the `finally` block.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
```diff
@@ -104,7 +113,7 @@ data Handlers m = Handlers
 -- first case, return the value returned by the action. In the second case,
 -- return @Nothing@.
 execAgainstBucket ::
-  (MonadDelay m, MonadAsync m, MonadFork m, MonadMask m) =>
+  (MonadDelay m, MonadAsync m, MonadFork m, MonadMask m, NoThunks (m ())) =>
```
amesgen, line 116: I think this is fine (also in the other signatures), but `IOLike m` would also work (in particular, it automatically it gives the `NoThunks` instance).

An intermediary approach would be to define
```haskell
type LeakyBucketConstraints m = (MonadDelay m, MonadAsync m, MonadFork m, MonadMask m, NoThunks (m ()))
```
and reuse that.

amesgen (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: 7664b8b7bbd2557e1b563853e77a83278ce7306a

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: Indeed, and it used to be that I had to for variable scoping reasons but that's not the case anymore, so a4387acf7e1e029f97afdbc826d2db9ee7b5d9a3.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/LeakyBucket.hs
Niols: Does that mean you would want me to do those changes? I quite like listing the fine-grained constraints as it gives an idea of what kind of things happen in the function very quickly, but I don't mind switching to `IOLike m` or using a type equality as suggested.

amesgen: No, just an option, the status quo also seems fine :+1:

Niols:
Merged as 9247b5359ed1d56834dd983a41ceccf72f642a65.

Issue #1016, opened 2024-03-22
Fix snapshot complications in UTXO-HD


jasagredo:
Hmmm, the thing is that snapshots will be one of:
```
dbDir/ledger/{snap1, snap2}/{tables,state}
```
```
dbDir/ledger/{snap1, snap2}/state
ssdDir/ledger/{snap1, snap2}/tables
```
```
dbDir/ledger/{snap1, snap2}/tables
ssdDir/ledger/{snap1, snap2}/state
```
```
ssdDir/ledger/{snap1, snap2}/{tables,state}
```

So they might be kind of "duplicated" (more like split in 2)

_Originally posted by @jasagredo in https://github.com/IntersectMBO/ouroboros-consensus/pull/1005#discussion_r1533499706_

PR #1026, opened 2024-03-26
Fix SnapshotPolicy

### LINKED ISSUE #1018: UTXO-HD is missing number of snapshots
It seems this option was lost in the numerous UTXO-HD refactorings. the fact is that `srnSnapshotInterval` still exists and `srnDiskPolicyArgs` doesn't.

RenateEilers:
# Description

Fixes  #1018 and #1016.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Impl/Snapshots.hs
```diff
@@ -186,14 +186,23 @@ trimSnapshots ::
...
+    diskSnapshotsPerFS <-  mapM getdiskSnapshotsPerFS fss
+    deletedSnapshots <- forM diskSnapshotsPerFS $ \(fs, ss) -> do
+      mapM
+        (\s -> do
+          deleteSnapshot fs s
+          traceWith tracer $ DeletedSnapshot s
+          pure s
+        )
+        ss
+    pure $ concat deletedSnapshots
```
RenateEilers, line 205: This will return copies of the same snapshots. Do we want to de-dup?

jasagredo (approved):

PR #1027, opened 2024-03-27
Introduce exception `DensityTooLow`

nbacquey:
This PR introduces a new exception `DensityTooLow`, which the GDD throws to ChainSync clients when their chain has been proven to be adversarial.

It also makes tests and labels aware of that exception, so that counting and labeling GDD disconnections can happen.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
@@ -4,7 +4,7 @@
 module Test.Consensus.Genesis.Tests.DensityDisconnect (tests) where
 
 import           Cardano.Slotting.Slot (WithOrigin (..), unSlotNo)
-import           Control.Exception (SomeException (..), fromException)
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
```diff
@@ -320,12 +322,11 @@ prop_densityDisconnectTriggersChainSel =
...
           exnCorrect = case exceptionsByComponent ChainSyncClient stateView of
-            -- FIXME: what exceptions do we see when the GDD disconnects the adversary?
             [exn] ->
               case fromException exn of
-                Just (SomeException _) -> True
-                _                      -> True
-            _ -> True
+                Just (DensityTooLow) -> True
+                _                    -> False
+            _ -> False
```
amesgen, line 329: Code golf with `ViewPatterns`:
```suggestion
          exnCorrect = case exceptionsByComponent ChainSyncClient stateView of
            [fromException -> Just DensityTooLow] -> True
            _                                     -> False
```

amesgen (approved):

PR #1035, opened 2024-04-02
Fix test: leashing attack with limited time

nbacquey:
- [x] Sometimes BlockFetch client doesn't get the block from the adversary and gets stalled
- [x] Compute time bounds according to real bucket parameters
- [x] Label empty schedules & trivial schedules
- [ ] ~Complete empty schedules with a dummy initial adversary point, to start the LoP~
- [ ] ~Enable timeouts only if a reasonable portion of tests are nonempty/nontrivial~
- [x] Extend the schedule so that adversary disconnection can happen


facundominguez:
I squashed and merged this as 281c36b.

PR #1038, opened 2024-04-04
Enter ChainSync jumping - Governor-less approach

Niols:
- [x] Enable CSJ in existing tests.
- [ ] Add a changelog entry, prepare milestone PR
- [x] Review and add tracing information to allow debugging of tests.
- [ ] Write a test that checks that CSJ syncs up by downloading from at most one peer in the happy path.
- [ ] Write a test that checks that CSJ syncs up by downloading from at most one honest peer at a time in any path, when CSJ is finished.
- [x] Explain CSJ implementation in a code note
- [x] Fix bucket leaking resuming in ChainSync. Noted by @amesgen.
- [x] Detect jumper rollbacks when they become objectors.
- [X] Disengage dynamos that rollback earlier than their last requested jump, and objectors that rollback to earlier than their rejected jump. Disengaging means continuing to download headers from the peer, but never asking it to jump, or to be dynamos or objectors.
- [x] Disengage from CSJ all peers that send MsgAwaitReply. Makes simpler to reason that GDD doesn't block.
- [x] Recover back the logic that enabled only one objector at a time. This logic can work again after [this discussion in slack](https://moduscreate.slack.com/archives/C05CEUFFH7Z/p1713524908238559) and [these changes to GDD](https://github.com/IntersectMBO/ouroboros-consensus/pull/1068).

---
Optional:
- [ ] Disable historical rollbacks of dynamos and historical rollbacks of jumpers. Jumper rollbacks might be detected when they send as objectors a header that agrees with the dynamo. This task is optional because peers making historical rollbacks don't induce asymmetric load in the network if they are kept connected. Jumpers might also show rollbacks by accepting jumps that they previously rejected, but these don't seem to affect syncing.

---
- [x] Make CSJ possible enable/disable and enable it by default.
- [x] Make the jump size configurable
- [x] Pause LoP for jumpers. Perhaps when jgNextInstruction returns JumpTo.
it varies with the eras. We should do the same as the GDD governor to get it.
- [x] Don't read more tvars than necessary in getDynamo
- [x] Turn off CSJ when all peers are idle (therefore no dynamo can be elected).
- [x] Get rid of buggy `fromJust` calls
- [x] Ask jumpers to jump to the tip of the dynamo's candidate fragment when the dynamo is freshly elected. Otherwise, if the new dynamo is blocked, no jumps might occur.
- [x] Update GDD to disconnect nodes with empty genesis windows

This one was sidestepped in a9f4c97 by eliminating the dependency on the genesis window:
- [ ] Find the genesis window for csj needs. At the moment it is given in `bracketChainSyncClient` as a configuration parameter, but this does not make much sense as it varies with the eras. We should do the same as the GDD governor to get it.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
@@ -12,7 +12,7 @@ module Test.Consensus.PeerSimulator.Run (
   , runPointSchedule
   ) where
 
-import           Control.Monad (foldM)
+import           Control.Monad (foldM, forM)
 import           Control.Monad.Class.MonadTime (MonadTime)
 import           Control.Monad.Class.MonadTimer.SI (MonadTimer)
 import           Control.Tracer (Tracer (..), nullTracer, traceWith)
@@ -26,7 +26,8 @@ import           Ouroboros.Consensus.Genesis.Governor (runGdd,
                      updateLoEFragGenesis)
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
                      (LedgerSupportsProtocol)
-import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client (ChainDbView,
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
+                     (CSJConfig (..), CSJEnabledConfig (..), ChainDbView,
                      ChainSyncClientHandle, ChainSyncLoPBucketConfig (..),
                      ChainSyncLoPBucketEnabledConfig (..), ChainSyncState (..),
                      viewChainSyncState)
@@ -56,7 +57,7 @@ import           Test.Consensus.PeerSimulator.StateView
 import           Test.Consensus.PeerSimulator.Trace
 import qualified Test.Consensus.PointSchedule as PointSchedule
 import           Test.Consensus.PointSchedule (BlockFetchTimeout,
-                     GenesisTest (GenesisTest), GenesisTestFull,
+                     CSJParams (..), GenesisTest (GenesisTest), GenesisTestFull,
                      LoPBucketParams (..), NodeState, PeersSchedule,
                      peersStatesRelative)
 import           Test.Consensus.PointSchedule.Peers (Peer (..), PeerId,
@@ -93,13 +94,17 @@ data SchedulerConfig =
     -- governor (GDD).
     , scEnableLoE                :: Bool
 
-    -- | Whether to enable to LoP. The parameters of the LoP come from
+    -- | Whether to enable the LoP. The parameters of the LoP come from
     -- 'GenesisTest'.
     , scEnableLoP                :: Bool
 
     -- | Enable node downtime if this is 'Just', using the value as minimum tick
     -- duration to trigger it.
     , scDowntime                 :: Maybe DiffTime
+
+    -- | Whether to enable ChainSync Jumping. The parameters come from
+    -- 'GenesisTest'.
+    , scEnableCSJ                :: Bool
   }
 
 -- | Default scheduler config
@@ -113,7 +118,8 @@ defaultSchedulerConfig =
     scTraceState = False,
     scEnableLoE = False,
     scEnableLoP = False,
-    scDowntime = Nothing
+    scDowntime = Nothing,
+    scEnableCSJ = False
   }
 
 -- | Enable debug tracing during a scheduler test.
@@ -139,6 +145,7 @@ startChainSyncConnectionThread ::
   ChainSyncResources m blk ->
   ChainSyncTimeout ->
   ChainSyncLoPBucketConfig ->
+  CSJConfig ->
   StateViewTracers blk m ->
   StrictTVar m (Map PeerId (ChainSyncClientHandle m blk)) ->
   m (Thread m (), Thread m ())
@@ -152,6 +159,7 @@ startChainSyncConnectionThread
   ChainSyncResources {csrServer}
   chainSyncTimeouts_
   chainSyncLoPBucketConfig
+  csjConfig
   tracers
   varHandles
   = do
@@ -159,7 +167,7 @@ startChainSyncConnectionThread
     clientThread <-
       forkLinkedThread registry ("ChainSyncClient" <> condense srPeerId) $
         bracketSyncWithFetchClient fetchClientRegistry srPeerId $
-          ChainSync.runChainSyncClient tracer cfg chainDbView srPeerId chainSyncTimeouts_ chainSyncLoPBucketConfig tracers varHandles clientChannel
+          ChainSync.runChainSyncClient tracer cfg chainDbView srPeerId chainSyncTimeouts_ chainSyncLoPBucketConfig csjConfig tracers varHandles clientChannel
     serverThread <-
       forkLinkedThread registry ("ChainSyncServer" <> condense srPeerId) $
         ChainSync.runChainSyncServer tracer srPeerId tracers csrServer serverChannel
@@ -238,10 +246,20 @@ dispatchTick tracer varHandles peers lifecycle node (number, (duration, Peer pid
     traceNewTick :: m ()
     traceNewTick = do
       currentChain <- atomically $ ChainDB.getCurrentChain (lnChainDb node)
-      csState <- atomically $ do
+      (csState, jumpingStates) <- atomically $ do
          m <- readTVar varHandles
-         traverse (readTVar . CSClient.cschState) (m Map.!? pid)
-      traceWith tracer $ TraceNewTick number duration (Peer pid state) currentChain (CSClient.csCandidate <$> csState)
+         csState <- traverse (readTVar . CSClient.cschState) (m Map.!? pid)
+         jumpingStates <- forM (Map.toList m) $ \(peer, h) -> do
+           st <- readTVar (CSClient.cschJumping h)
+           pure (peer, st)
+         pure (csState, jumpingStates)
+      traceWith tracer $ TraceNewTick
+        number
+        duration
+        (Peer pid state)
+        currentChain
+        (CSClient.csCandidate <$> csState)
+        jumpingStates
 
 -- | Iterate over a 'PointSchedule', sending each tick to the associated peer in turn,
 -- giving each peer a chunk of computation time, sequentially, until it satisfies the
@@ -329,6 +347,7 @@ startNode schedulerConfig genesisTest interval = do
           prChainSync
           chainSyncTimeouts_
           chainSyncLoPBucketConfig
+          csjConfig
           lnStateViewTracers
           handles
         BlockFetch.startKeepAliveThread peerRegistry fetchClientRegistry pid
@@ -373,6 +392,7 @@ startNode schedulerConfig genesisTest interval = do
         gtChainSyncTimeouts
       , gtBlockFetchTimeouts
       , gtLoPBucketParams = LoPBucketParams { lbpCapacity, lbpRate }
+      , gtCSJParams = CSJParams { csjpJumpSize }
       } = genesisTest
 
     chainSyncTimeouts_ =
@@ -384,6 +404,11 @@ startNode schedulerConfig genesisTest interval = do
       if scEnableLoP schedulerConfig
         then ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig { csbcCapacity = lbpCapacity, csbcRate = lbpRate }
         else ChainSyncLoPBucketDisabled
+
+    csjConfig =
+      if scEnableCSJ schedulerConfig
+        then CSJEnabled CSJEnabledConfig { csjcJumpSize = csjpJumpSize }
+        else CSJDisabled
 
     blockFetchTimeouts_ =
       if scEnableBlockFetchTimeouts schedulerConfig
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -22,6 +22,7 @@
 -- who then activates the next tick's peer.
 module Test.Consensus.PointSchedule (
     BlockFetchTimeout (..)
+  , CSJParams (..)
   , ForecastRange (..)
   , GenesisTest (..)
   , GenesisTestFull
@@ -444,6 +445,11 @@ data LoPBucketParams = LoPBucketParams {
   lbpRate     :: Rational
   }
 
+data CSJParams = CSJParams {
+    csjpJumpSize :: SlotNo
+  }
+  deriving Show
+
 -- | Similar to 'ChainSyncTimeout' for BlockFetch. Only the states in which the
 -- server has agency are specified. REVIEW: Should it be upstreamed to
 -- ouroboros-network-protocols?
@@ -462,6 +468,7 @@ data GenesisTest blk schedule = GenesisTest
     gtChainSyncTimeouts  :: ChainSyncTimeout,
     gtBlockFetchTimeouts :: BlockFetchTimeout,
     gtLoPBucketParams    :: LoPBucketParams,
+    gtCSJParams          :: CSJParams,
     gtSlotLength         :: SlotLength,
     gtSchedule           :: schedule
   }
@@ -482,6 +489,7 @@ prettyGenesisTest prettySchedule genesisTest =
   , "  gtForecastRange: " ++ show (unForecastRange gtForecastRange)
   , "  gtDelay: " ++ show delta
   , "  gtSlotLength: " ++ show gtSlotLength
+  , "  gtCSJParams: " ++ show gtCSJParams
   , "  gtChainSyncTimeouts: "
   , "    canAwait = " ++ show canAwaitTimeout
   , "    intersect = " ++ show intersectTimeout
@@ -510,6 +518,7 @@ prettyGenesisTest prettySchedule genesisTest =
       , gtBlockFetchTimeouts = BlockFetchTimeout{busyTimeout, streamingTimeout}
       , gtLoPBucketParams = LoPBucketParams{lbpCapacity, lbpRate}
       , gtSlotLength
+      , gtCSJParams
       , gtSchedule
       } = genesisTest
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -54,15 +54,19 @@ module Ouroboros.Consensus.MiniProtocol.ChainSync.Client (
   , Consensus
   , Our (..)
   , Their (..)
-    -- * Trace events
+    -- * Genesis configuration
+  , CSJConfig (..)
+  , CSJEnabledConfig (..)
   , ChainSyncLoPBucketConfig (..)
   , ChainSyncLoPBucketEnabledConfig (..)
+    -- * Trace events
   , InvalidBlockReason
   , TraceChainSyncClientEvent (..)
     -- * State shared with other components
   , ChainSyncClientHandle (..)
   , ChainSyncState (..)
   , ChainSyncStateView (..)
+  , Jumping.noJumping
   , chainSyncStateFor
   , noIdling
   , noLoPBucket
@@ -75,6 +79,7 @@ import           Control.Tracer
 import           Data.Kind (Type)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import           Data.Maybe (fromMaybe)
 import           Data.Proxy
 import           Data.Typeable
 import           Data.Word (Word64)
@@ -95,6 +100,7 @@ import           Ouroboros.Consensus.Ledger.Basics (LedgerState)
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.Jumping as Jumping
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.State
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Protocol.Abstract
@@ -106,7 +112,7 @@ import           Ouroboros.Consensus.Util.AnchoredFragment (cross)
 import           Ouroboros.Consensus.Util.Assert (assertWithMsg)
 import           Ouroboros.Consensus.Util.EarlyExit (WithEarlyExit, exitEarly)
 import qualified Ouroboros.Consensus.Util.EarlyExit as EarlyExit
-import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Consensus.Util.IOLike hiding (handle)
 import qualified Ouroboros.Consensus.Util.LeakyBucket as LeakyBucket
 import           Ouroboros.Consensus.Util.STM (Fingerprint, Watcher (..),
                      WithFingerprint (..), withWatcher)
@@ -163,6 +169,36 @@ data ChainSyncLoPBucketConfig
   |
     -- | Enable the leaky bucket.
     ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig
+
+-- | Configuration of ChainSync Jumping
+data CSJConfig
+  =
+    -- Disable ChainSync Jumping. All clients will fully synchronize with
+    -- the chain of its peer.
+    CSJDisabled
+  |
+    -- | Enable ChainSync Jumping
+    CSJEnabled CSJEnabledConfig
+
+data CSJEnabledConfig = CSJEnabledConfig {
+  -- | The _ideal_ size for ChainSync jumps. Note that the algorithm
+  -- is best-effort: there might not be exactly `csjcJumpSize` slots between two
+  -- jumps, depending on the chain.
+  --
+  -- There can be a few less slots between jumps if there is not a block exactly
+  -- at the boundary. Jumps are often made when a block is announced after the
+  -- jump boundary.
+  --
+  -- There can be even less slots if a dynamo is elected and it requires an
+  -- initial jump regardless of how far we are from the next jump boundary.
+  --
+  -- csjcJumpSize must be greater than 0 and smaller or equal to the genesis
+  -- window size. The larger the jump size, the less jumps are made and peers
+  -- are less involved in the syncing. A jump size as large as the genesis
+  -- window has a higher change that dishonest peers can delay syncing by a
+  -- small margin (around 2 minutes per dishonest peer with mainnet parameters).
+  csjcJumpSize       :: SlotNo
+}
 
 defaultChainDbView ::
      (IOLike m, LedgerSupportsProtocol blk)
@@ -265,6 +301,9 @@ data ChainSyncStateView m blk = ChainSyncStateView {
 
     -- | Control the 'LeakyBucket' for the LoP.
   , csvLoPBucket     :: !(LoPBucket m)
+
+    -- | Jumping-related API.
+  , csvJumping       :: !(Jumping.Jumping m blk)
   }
   deriving stock (Generic)
 
@@ -273,7 +312,8 @@ deriving anyclass instance (
   HasHeader blk,
   NoThunks (Header blk)
   ) => NoThunks (ChainSyncStateView m blk)
-bracketChainSyncClient ::
+
+bracketChainSyncClient :: forall m peer blk a.
     ( IOLike m
     , Ord peer
     , LedgerSupportsProtocol blk
@@ -286,6 +326,7 @@ bracketChainSyncClient ::
  -> peer
  -> NodeToNodeVersion
  -> ChainSyncLoPBucketConfig
+ -> CSJConfig
  -> (ChainSyncStateView m blk -> m a)
  -> m a
 bracketChainSyncClient
@@ -295,46 +336,74 @@ bracketChainSyncClient
     peer
     version
     csBucketConfig
+    csjConfig
     body
-  =
-    bracket acquireHandle releaseHandle
-  $ \varState ->
+  = mkChainSyncClientHandleState >>= \csHandleState ->
+    withCSJCallbacks csHandleState csjConfig $ \csjCallbacks ->
         withWatcher
             "ChainSync.Client.rejectInvalidBlocks"
-            (invalidBlockWatcher varState)
+            (invalidBlockWatcher csHandleState)
       $ LeakyBucket.execAgainstBucket lopBucketConfig
       $ \lopBucket ->
             body ChainSyncStateView {
               csvSetCandidate =
-              modifyTVar varState . \ c s -> s {csCandidate = c}
+              modifyTVar csHandleState . \ c s -> s {csCandidate = c}
             , csvSetLatestSlot =
-              modifyTVar varState . \ ls s -> s {csLatestSlot = Just $! ls}
+              modifyTVar csHandleState . \ ls s -> s {csLatestSlot = Just $! ls}
             , csvIdling = Idling {
-                idlingStart = atomically $ modifyTVar varState $ \ s -> s {csIdling = True}
-              , idlingStop = atomically $ modifyTVar varState $ \ s -> s {csIdling = False}
+                idlingStart = atomically $ modifyTVar csHandleState $ \ s -> s {csIdling = True}
+              , idlingStop = atomically $ modifyTVar csHandleState $ \ s -> s {csIdling = False}
               }
             , csvLoPBucket = LoPBucket {
                 lbPause = LeakyBucket.setPaused lopBucket True
               , lbResume = LeakyBucket.setPaused lopBucket False
               , lbGrantToken = void $ LeakyBucket.fill lopBucket 1
               }
+            , csvJumping = csjCallbacks
             }
   where
-    acquireHandle = do
-        cschState <- newTVarIO $ ChainSyncState {
-            csCandidate = AF.Empty AF.AnchorGenesis
-          , csLatestSlot = Nothing
-          , csIdling = False
-          }
-        tid <- myThreadId
-        atomically $ do
-          modifyTVar varHandles $ Map.insert peer ChainSyncClientHandle {
+    mkChainSyncClientHandleState =
+      newTVarIO ChainSyncState {
+          csCandidate = AF.Empty AF.AnchorGenesis
+        , csLatestSlot = Nothing
+        , csIdling = False
+        }
+
+    withCSJCallbacks ::
+      StrictTVar m (ChainSyncState blk) ->
+      CSJConfig ->
+      (Jumping.Jumping m blk -> m a) ->
+      m a
+    withCSJCallbacks cschState CSJDisabled f = do
+      tid <- myThreadId
+      cschJumpInfo <- newTVarIO Nothing
+      cschJumping <- newTVarIO Disengaged
+      let handle = ChainSyncClientHandle {
               cschGDDKill = throwTo tid DensityTooLow
             , cschState
+            , cschJumping
+            , cschJumpInfo
             }
-        pure cschState
+          insertHandle = atomically $ modifyTVar varHandles $ Map.insert peer handle
+          deleteHandle = atomically $ modifyTVar varHandles $ Map.delete peer
+      bracket_ insertHandle deleteHandle $ f Jumping.noJumping
 
-    releaseHandle _ = atomically $ modifyTVar varHandles $ Map.delete peer
+    withCSJCallbacks csHandleState (CSJEnabled csjEnabledConfig) f =
+      bracket (acquireContext csHandleState csjEnabledConfig) releaseContext $ \peerContext ->
+        f $ Jumping.mkJumping peerContext
+    acquireContext cschState (CSJEnabledConfig jumpSize) = do
+        tid <- myThreadId
+        atomically $ do
+          cschJumpInfo <- newTVar Nothing
+          context <- Jumping.makeContext varHandles jumpSize
+          Jumping.registerClient context peer cschState $ \cschJumping -> ChainSyncClientHandle
+            { cschGDDKill = throwTo tid DensityTooLow
+            , cschState
+            , cschJumping
+            , cschJumpInfo
+            }
+
+    releaseContext = atomically . Jumping.unregisterClient
 
     invalidBlockWatcher varState =
         invalidBlockRejector
@@ -342,7 +411,7 @@ bracketChainSyncClient
 
     -- | Wrapper around 'LeakyBucket.execAgainstBucket' that handles the
     -- disabled bucket by running the given action with dummy handlers.
-    lopBucketConfig :: MonadThrow m => LeakyBucket.Config m
+    lopBucketConfig :: LeakyBucket.Config m
     lopBucketConfig =
       case csBucketConfig of
         ChainSyncLoPBucketDisabled -> LeakyBucket.dummyConfig
@@ -668,6 +737,7 @@ data DynamicEnv m blk = DynamicEnv {
   , setLatestSlot       :: WithOrigin SlotNo -> STM m ()
   , idling              :: Idling m
   , loPBucket           :: LoPBucket m
+  , jumping             :: Jumping.Jumping m blk
   }
 
 -- | General values collectively needed by the top-level entry points
@@ -761,8 +831,7 @@ chainSyncClient cfgEnv dynEnv =
       } = chainDbView
 
     DynamicEnv {
-        idling,
-        loPBucket
+        idling
       } = dynEnv
 
     mkIntEnv ::
@@ -808,7 +877,7 @@ chainSyncClient cfgEnv dynEnv =
                   recvMsgRollForward  = \_hdr _tip -> go n' s
                 , recvMsgRollBackward = \_pt  _tip -> go n' s
                 }
-      in Stateful $ \s -> do (idlingStop idling >> lbResume loPBucket); go n0 s
+      in Stateful $ \s -> idlingStop idling >> go n0 s
 
     terminate ::
         ChainSyncClientResult
@@ -898,6 +967,7 @@ findIntersectionTop cfgEnv dynEnv intEnv =
 
     DynamicEnv {
         setCandidate
+      , jumping
       } = dynEnv
 
     InternalEnv {
@@ -999,9 +1069,6 @@ findIntersectionTop cfgEnv dynEnv intEnv =
                         disconnect
                       $ InvalidIntersection
                             intersection (ourTipFromChain ourFrag) theirTip
-            atomically $ do
-              setCandidate theirFrag
-              setLatestSlot dynEnv (AF.headSlot theirFrag)
             let kis =
                    assertKnownIntersectionInvariants (configConsensus cfg)
                  $ KnownIntersectionState {
@@ -1011,6 +1078,10 @@ findIntersectionTop cfgEnv dynEnv intEnv =
                      , theirHeaderStateHistory
                      , kBestBlockNo            = uBestBlockNo
                      }
+            atomically $ do
+              updateJumpInfoSTM jumping kis
+              setCandidate theirFrag
+              setLatestSlot dynEnv (AF.headSlot theirFrag)
             continueWithState kis $
                 knownIntersectionStateTop cfgEnv dynEnv intEnv theirTip
 
@@ -1048,6 +1119,7 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
       , idling
       , loPBucket
       , setCandidate
+      , jumping
       } = dynEnv
 
     InternalEnv {
@@ -1071,7 +1143,7 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
     -- intersection before calling 'getLedgerView'.
     --
     -- This is also the place where we checked whether we're asked to terminate
-    -- by the mux layer.
+    -- by the mux layer or to wait and perform a CSJ jump.
     nextStep ::
         MkPipelineDecision
      -> Nat n
@@ -1079,11 +1151,37 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
      -> Stateful m blk
             (KnownIntersectionState blk)
             (ClientPipelinedStIdle n)
-    nextStep mkPipelineDecision n theirTip = Stateful $ \kis -> do
+    nextStep mkPipelineDecision n theirTip = Stateful $ \kis ->
         atomically controlMessageSTM >>= \case
             -- We have been asked to terminate the client
             Terminate -> terminateAfterDrain n $ AskedToTerminate
             _continue -> do
+                -- Wait until next jumping instruction, which can be either to
+                -- jump or to run normal ChainSync.
+                -- Pause LoP while waiting, we'll resume it if we get `RunNormally`
+                traceWith tracer TraceJumpingWaitingForNextInstruction
+                lbPause loPBucket
+                instruction <- Jumping.jgNextInstruction jumping
+                traceWith tracer $ TraceJumpingInstructionIs instruction
+                case instruction of
+                    Jumping.JumpTo jumpInfo ->
+                      continueWithState kis
+                        $ drainThePipe n
+                        $ offerJump mkPipelineDecision jumpInfo
+                    Jumping.RunNormally -> do
+                      lbResume loPBucket
+                      continueWithState kis
+                        $ nextStep' mkPipelineDecision n theirTip
+
+    nextStep' ::
+        MkPipelineDecision
+     -> Nat n
+     -> Their (Tip blk)
+     -> Stateful m blk
+            (KnownIntersectionState blk)
+            (ClientPipelinedStIdle n)
+    nextStep' mkPipelineDecision n theirTip =
+        Stateful $ \kis ->
                 atomically (intersectsWithCurrentChain kis) >>= \case
                     -- Our chain (tip) didn't change or if it did, it still
                     -- intersects with the candidate fragment, so we can
@@ -1093,6 +1191,7 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
                                 theirFrag
                               } = kis'
                         atomically $ do
+                          updateJumpInfoSTM jumping kis'
                           setCandidate theirFrag
                           setLatestSlot dynEnv (AF.headSlot theirFrag)
                         return $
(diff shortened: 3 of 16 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    Jumper nextJumpVar goodPoint state ->
+        case (state, jumpResult) of
+          (FoundIntersection, _) ->
+            -- Ignore jump results when we already found the intersection.
+            pure ()
+          (Happy, AcceptedJump goodPoint') -> do
+            -- The jump was accepted; we bump the peer's candidate fragment to
+            -- the dynamo's candidate fragment up to the accepted point.
+            dynamoFragment <- fromJust <$> getDynamoFragment governor
+            let slicedDynamoFragment = fst $ fromJust $ splitAfterPoint dynamoFragment goodPoint'
```
facundominguez, line 150: Is `fromJust` going to provide the source location of the error? This was not the case with `head` in all compiler versions. Maybe `error` should be called explicitly here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    Dynamo _ -> pure ()
+    Objector _ -> pure ()
+    Jumper nextJumpVar goodPoint state ->
+        case (state, jumpResult) of
+          (FoundIntersection, _) ->
+            -- Ignore jump results when we already found the intersection.
+            pure ()
+          (Happy, AcceptedJump goodPoint') -> do
+            -- The jump was accepted; we bump the peer's candidate fragment to
+            -- the dynamo's candidate fragment up to the accepted point.
```
facundominguez, line 148: What is the meaning of "bumping a candidate fragment" and why should it be bumped?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    lookForIntersection nextJumpVar goodPoint badPoint = do
+      (dynamoFragment0 :: AnchoredFragment (Header blk)) <- fromJust <$> getDynamoFragment governor -- full fragment
+      let (dynamoFragment1 :: AnchoredFragment (Header blk)) = snd $ fromJust $ splitAfterPoint dynamoFragment0 goodPoint -- fragment from goodPoint (excl)
+      let (dynamoFragment2 :: AnchoredFragment (Header blk)) = fst $ fromJust $ splitBeforePoint dynamoFragment1 badPoint -- fragment from goodPoint to badPoint (excl)
+      case (toOldestFirst dynamoFragment2 :: [Header blk]) of
+        [] -> do
+          modifyTVar clientStates . Map.insert peer $ Jumper nextJumpVar goodPoint FoundIntersection
+          electNewObjector governor
+        points -> do
+          let (middlePoint :: Header blk) = points !! (length points `div` 2)
```
facundominguez, line 176: Can't this be accomplished better with the `AnchoredFragment` representation?

facundominguez (commented):
Did a first pass.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -297,20 +253,44 @@ noLoPBucket =
     , lbGrantToken = pure ()
     }
 
+data JumpingGovernor m blk = JumpingGovernor {
+    jgNextInstruction   :: !(m (JumpingGovernor.Instruction blk))
+  , jgProcessJumpResult :: !(JumpingGovernor.JumpResult blk -> m ())
+  }
+  deriving stock (Generic)
```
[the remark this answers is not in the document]
Niols: I never quite know how to document those things, because they are only the hooks. So if you want to fully change their semantics, you're free to do so. In practice, there will just be two types of hooks, though, either _the_ ChainSync jumping ones or no-op versions for when CSJ is unwanted/unneeded. I gave it a try in 56b09b9e13af4534272a17ef73155a5509213b17; is that enough?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    -- the 'Jumper' constructor is concerned.
+    Happy
+  | -- | The jumper disagrees with the dynamo and we are searching where exactly
+    -- that happens. All we know is a point where the jumper agrees with the
+    -- dynamo (in the 'Jumper' constructor) and a point where the jumper
+    -- disagrees with the dynamo, carried by this constructor.
+    LookingForIntersection !(Point blk)
+  | -- | The jumper disagrees with the dynamo and we have found where exactly.
+    -- The last point where the jumper agrees with the dynamo is stored in the
+    -- ClientState.
```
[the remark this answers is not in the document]
Niols: Added in 6fec532248ecbeaaffdda58b057d4a21e72c0290.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    Dynamo lastJumpSlot -> maybeJump lastJumpSlot >> pure RunNormally
+    Objector _ -> pure RunNormally
+    Jumper nextJumpVar _ _ ->
+      readTVar nextJumpVar >>= \case
+        Nothing -> retry
+        Just fragment -> do
+          writeTVar nextJumpVar Nothing
+          pure $ JumpTo fragment
+  where
+    maybeJump lastJumpSlot = do
```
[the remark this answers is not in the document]
Niols: Done in 835123f7ea5a6249bf763273c02957cc59ef33e5.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+      dynamoFragment <- fromJust <$> getDynamoFragment governor
+      when (succWithOrigin (headSlot dynamoFragment) >= succWithOrigin lastJumpSlot + jumpSize) $ do
+        clientStates' <- readTVar clientStates
+        forM_ (Map.elems clientStates') $ \case
+          Jumper nextJumpVar _ Happy -> writeTVar nextJumpVar $ Just $! castPoint (headPoint dynamoFragment)
+          _ -> pure ()
```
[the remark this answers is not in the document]
Niols: Done in 835123f7ea5a6249bf763273c02957cc59ef33e5.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
facundominguez: I would like to have the story of how these hooks are supposed to be used by the ChainSync client. Otherwise, how do we know that they aren't omitted in needed places?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    Ord peer,
+    HasHeader blk
+  ) =>
+  Handle m peer blk ->
+  STM m ()
+demoteObjector Handle {clientStates} = do
+  clientStates' <- Map.toList <$> readTVar clientStates
+  forM_ clientStates' $ \case
+    (peer, Objector intersection) -> modifyTVar clientStates . Map.insert peer =<< newJumper intersection FoundIntersection
+    _ -> pure ()
```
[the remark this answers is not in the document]
Niols: Done in 0fac6002356499935433fe94cac589c0627cec70.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+    HasHeader blk
+  ) =>
+  Point blk ->
+  JumperState blk ->
+  STM m (ClientState m blk)
+newJumper intersection state = do
+  nextJumpVar <- newTVar Nothing
+  pure $ Jumper nextJumpVar intersection state
+
+registerClient ::
```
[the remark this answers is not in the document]
Niols: Done in d32a96214bff05d5b9782ed0e92b1638f0b04577.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+unregisterClient governor@Handle {clientStates} peer = do
+  clientState <- (Map.! peer) <$> readTVar clientStates
+  modifyTVar clientStates $ Map.delete peer
+  case clientState of
+    Jumper _ _ _ -> pure ()
+    Objector _   -> electNewObjector governor
+    Dynamo _     -> electNewDynamo governor
+
+-- | REVIEW: This must exist somewhere else.
+pointSlot :: Point blk -> WithOrigin SlotNo
```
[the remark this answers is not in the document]
Niols: Thanks! Wonder how I missed it... Fixed in 24f2f89571ae112d59f7c5f35b7eba6d7ed29535.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+  -- Get everybody; choose a random new dynamo and put everyone else back to
+  -- being jumpers. NOTE: This might stop an objector from syncing and send it
+  -- back to being a jumper only for it to start again later, but nevermind.
+  (Map.keys <$> readTVar clientStates)
+    >>= \case
+      [] -> pure ()
+      dynamo : jumpers -> do
+        modifyTVar clientStates $ Map.insert dynamo $ Dynamo Origin
+        forM_ jumpers $ \jumper ->
+          modifyTVar clientStates . Map.insert jumper =<< newJumper GenesisPoint Happy
```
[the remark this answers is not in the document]
Niols: Indeed! I replaced “random” by “unspecified” in c297207b78df00f0627dbc2db4ad4b79ef373835.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+      -- | The last slot at which we triggered jumps for the jumpers.
+      !(WithOrigin SlotNo)
+  | Objector
+      -- | The last known point where the objector agrees with the dynamo.
+      !(Point blk)
+  | Jumper
+      -- | The next jump to be executed.
+      !(StrictTVar m (Maybe (Point blk)))
+      -- | The result of the last jump.
+      !(Point blk)
```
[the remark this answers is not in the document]
Niols: I don't think it depends on `JumperState`; it is always the youngest point that we know the jumper and dynamo agree on. I tried making that more precise in b8d8d04df2113392973cea796e154feaa1c24e29.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+          let (middlePoint :: Header blk) = points !! (length points `div` 2)
+          writeTVar nextJumpVar $ Just $! (castPoint (blockPoint middlePoint) :: Point blk)
+          modifyTVar clientStates . Map.insert peer $ Jumper nextJumpVar goodPoint $ LookingForIntersection badPoint
+
+getDynamo ::
+  (MonadSTM m) =>
+  Handle m peer blk ->
+  STM m (Maybe peer)
+getDynamo Handle {clientStates} =
+  (fst <$>) . List.find (\(_, clientState) -> case clientState of Dynamo _ -> True; _ -> False) . Map.toList <$> readTVar clientStates
```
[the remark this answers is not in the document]
Niols: That is what I did initially, mostly because that was how it was done in the happy path implementation by @amesgen. I do find it more satisfactory to see a map of all the peers, each having a state, rather than having a record with a dynamo field, an objector field, and then a map of jumper peers, but it isn't based on much. I'm happy changing this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.State
+                     (ChainSyncClientHandle (..), ChainSyncState (..))
+import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Network.AnchoredFragment (AnchoredFragment,
+                     blockPoint, headPoint, headSlot, splitAfterPoint,
+                     splitBeforePoint, toOldestFirst)
+
+data Handle m peer blk = Handle
+  { clientHandles :: !(StrictTVar m (Map peer (ChainSyncClientHandle m blk))),
+    clientStates  :: !(StrictTVar m (Map peer (ClientState m blk)))
```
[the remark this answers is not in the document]
Niols: This is an artifact of me trying to get things done before cleaning up. I have since then pushed the CSJ-specific state into `ChainSyncHandle`; cf 4be366d53504d63bd174d2c86d9db3f639b42c8f.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
Niols: Since we know that the jumper accepts the jump, that means we know their whole candidate fragment is the same as the dynamo's at least up to the jump point. We can therefore set the candidate TVar to that fragment immediately. This will then impact the LoE, GDD, and whatever else reads the ChainSync clients' candidate fragments. That's what I called “bumping” the fragment in my head, but it's just the word that popped in and I'm happy with any other wording!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
Niols: Possibly! And I'd find that more satisfactory tbh; I was mostly trying to get things done to have a draft. I will look into cleaning this up.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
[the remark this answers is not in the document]
facundominguez: Ok, my bad.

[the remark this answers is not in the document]
Niols: Not really; this whole thing was (still is) extremely poorly documented!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/JumpingGovernor.hs
```diff
@@ -0,0 +1,337 @@
...
+            -- intersection.
+            lookForIntersection nextJumpVar goodPoint badPoint
+          (LookingForIntersection badPoint, AcceptedJump goodPoint') ->
+            -- The peer that was looking for an intersection accepts the given
+            -- point, helping us narrow things down.
+            lookForIntersection nextJumpVar goodPoint' badPoint
+          (LookingForIntersection _, RejectedJump badPoint') ->
+            -- The peer that was looking for an intersection rejects the given
+            -- point, helping us narrow things down.
+            lookForIntersection nextJumpVar goodPoint badPoint'
```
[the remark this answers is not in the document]
facundominguez: Jumpers `LookingForIntersection` are only asked to jump to better approximations. This could be said explicitly in the comment here.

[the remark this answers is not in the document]
facundominguez: I suggest a fix in another conversation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+    >>= \case
+      [] -> pure ()
+      dynamo : jumpers -> do
+        setJumpingState handlesVar dynamo $ Dynamo Origin
+        forM_ jumpers $ \jumper ->
+          setJumpingState handlesVar jumper =<< newJumper GenesisPoint Happy
+
+-- | Size of jumps, in number of slots. FIXME: Make this configurable.
+jumpSize :: SlotNo
+jumpSize = 2
```
facundominguez, line 294: Why is the jump so small?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+            case cschJumping handle of
+              Jumper _ intersection FoundIntersection -> Just (peer, intersection)
+              _ -> Nothing
+        )
+      . Map.toList
+      <$> readTVar handlesVar
+  case sortedObjectorCandidates of
+    [] -> pure ()
+    (objector, intersection) : _ ->
+      setJumpingState handlesVar objector $ Objector intersection
```
facundominguez, line 271: What is the effect of ignoring the jumpers that haven't found the intersection yet?

Is LoP leaking paused for them? If the bucket is leaking, they must accept jumps quickly for the candidate fragment to be updated.

Are they considered as idling? If so, they also must accept jumps quickly so GDD does not disconnect them when they disagree by more than k slots when the dynamo or the objector move ahead.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
Niols: Just because I wanted to see it in peer simulator tests. This will actually need to come from a configuration option somewhere that will also allow disabling CSJ altogether.

In general, I don't think the actual value of jump sizes for main net is set. @nfrisby and @amesgen mentioned half the size of a Genesis window, but I don't think that's set in stone. Some of the considerations I remember are:

- The tinier the jumps, the heavier the work on the network (and the honest nodes in particular).

- The longer the jumps, the more imprecise the information is, which has several impacts:

  - In the current implementation, this means that more time is spent with back-and-forth `MsgFindIntersection` to pinpoint the exact intersection. (There is definitely work to do on making this process faster but it will probably always remain logarithmic in the size of the jumps.)

  - When an objector boots up, they might be `jumpSize` slots behind the dynamo, so they might take time to catch up (on purpose or just because that's many headers), and that's time spent not moving forward with the sync. This should happen very rarely in practice because that then eliminates an adversary from the list of peers.

  - If `jumpSize` is close to the size of the forecast window, the delay of response for the jump requests will start implying time spent with the dynamo beyond the forecast horizon, which is time spent not exchanging headers and potentially where the TCP window shrinks.

I'm sure there are many other considerations I don't remember but that's it off the top of my head.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
Niols: > What is the effect of ignoring the jumpers that haven't found the intersection yet? I'm guessing they will hold back LoE until something disconnects them.

Indeed, we know that they are unhappy with the dynamo's fragment so there is an intersection somewhere and the LoE should prevent us from moving forward. I think we have to ignore them because we don't know what to do with them yet: there is no point asking them to jump to the dynamo's fragment because we already know they disagree but we also don't have enough information yet to decide if they are the objector whose intersection with the dynamo is the oldest.

The key point here is that we need to make sure that the objector whose intersection with the dynamo is the oldest is running ChainSync; otherwise, the whole system might block because the LoE cannot move forward (because of them) but then the GDD might not have enough information to kill the other peers. In the current implementation with at most one objector, that means that we need to make sure that's the one that's running. We could imagine other designs (eg. where all unhappy jumpers start ChainSync, or maybe we classify them by answers to the jump requests, etc.).

Note that for now we try to find the exact intersection while we in fact only would need to pinpoint it down enough to decide that they are the oldest!

> Is LoP leaking paused for them? If the bucket is leaking, they must accept jumps quickly for the candidate fragment to be updated. But they never earn tokens, because they don't request headers.
> 
> Are they considered as idling? If so, they also must accept jumps quickly so GDD does not disconnect them when they disagree by more than k slots when the dynamo or the objector move ahead.

The ChainSync clients are indeed idling when waiting for the next instruction from ChainSync jumping.

> If no Genesis mechanism is active to disconnect them, they can delay each jump request by as much as the intersect timeout.

Absolutely, so 10s per jump request, which is a lot. Finding the intersection as it is done currently can require quite a few back-and-forth with `MsgFindIntersect` and they can all take a while. Again, the idea is that then we'll kill the peer that disagrees so it's not too bad to have paid even a few minutes to decide; the rest of the sync should be fast. However, it would be easy for an adversary to try and leash the sync by serving the honest chain but waiting 9.9s to answer each jump request. I think this pushes for a value of `jumpSize` small enough that the dynamo still has at least 10s worth of work to do.

facundominguez: If all jumpers are idling, they will be disconnected by GDD when the dynamo or the objector move more than k blocks ahead.

Are there any guarantees that jumpers corresponding to honest peers will have a fair chance to find intersections before being disconnected?

Also, when intersections are finally found, I'm not seeing any code to update the candidate fragments of those jumpers.

facundominguez: @amesgen suggested considering jumpers as not idling and using short timeouts for MsgFindIntersect.

Perhaps ensuring safety and liveness is possible, but we need extra provisions,
* Set the candidate fragments when intersections are found, so the LoE can move forward promptly on prefixes where there is agreement.
* Hold objectors accountable for their earlier objections. If they object when jumping, but yield the honest chain when acting as objectors, GDD and LoP won't be able to disconnect anyone.
* Select a new objector when the current objector requires headers after the forecast horizon. This can happen when the intersection of the objector is much later than the LoE tip.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/State.hs
```diff
@@ -0,0 +1,123 @@
...
+    Dynamo
+      -- | The last slot at which we triggered jumps for the jumpers.
+      !(WithOrigin SlotNo)
+  | -- | The objector, of which there is at most one, also runs normal
+    -- ChainSync. It is a former jumper that disagreed with the dynamo. When
+    -- that happened, we spun it up to let normal ChainSync and Genesis decide
+    -- which one to disconnect from.
+    Objector
+      -- | The last known point where the objector agrees with the dynamo.
+      !(Point blk)
```
facundominguez, line 88: At the moment I'm not seeing where this point is being used.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+          setJumpingState handlesVar peer $ Jumper nextJumpVar goodPoint (LookingForIntersection badPoint)
+
+    -- | Select a slice of an anchored fragment between two points, inclusive.
+    -- Both points must exist on the chain, in order, or the result is Nothing.
+    -- REVIEW: How does this function behave if @fromExcl == toExcl@? How should
+    -- it behave? It doesn't matter much because we never have this in our code,
+    -- but still.
+    sliceRangeExcl fragment fromExcl toExcl = do
+      fragmentFrom <- snd <$> splitAfterPoint fragment fromExcl
+      fst <$> splitBeforePoint fragmentFrom toExcl
```
facundominguez, line 157:
```suggestion
    -- | Select a slice of an anchored fragment between two points. The slice
    -- includes neither of the points.
    --
    -- Both points must exist on the anchored fragment, in order,
    -- and `fromExcl /= toExcl`, otherwise the result is `Nothing`.
    sliceRangeExcl fragment fromExcl toExcl = do
      fragmentFrom <- snd <$> splitAfterPoint fragment fromExcl
      fst <$> splitBeforePoint fragmentFrom toExcl
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+            -- point, helping us narrow things down.
+            lookForIntersection nextJumpVar goodPoint' badPoint
+          (LookingForIntersection _, RejectedJump badPoint') ->
+            -- The peer that was looking for an intersection rejects the given
+            -- point, helping us narrow things down.
+            lookForIntersection nextJumpVar goodPoint badPoint'
+  where
+    lookForIntersection nextJumpVar goodPoint badPoint = do
+      (dynamoFragment0 :: AnchoredFragment (Header blk)) <- fromJust <$> getDynamoFragment handlesVar -- full fragment
+      let dynamoFragment = fromJust $ sliceRangeExcl dynamoFragment0 goodPoint badPoint
```
facundominguez, line 140: This `fromJust` call is going to fail if `goodPoint` is before the anchor of the dynamo fragment. e.g. if `goodPoint` is `GenesisPoint` and the fragment is not anchored at genesis.

Maybe this works better:
```
let dynamoAnchor = anchorPoint (anchor dynamoFragment0)
    dynamoFragment = fromJust $ do
      maxPoint <- maybeMax goodPoint dynamoAnchor
      sliceRangeExcl dynamoFragment0 maxPoint badPoint
```
where `dynamoAnchor` must be an ancestor or descendant of `goodPoint`, or otherwise it is fine to disconnect the jumper.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+processJumpResult handlesVar peer jumpResult =
+  (cschJumping . (Map.! peer) <$> readTVar handlesVar) >>= \case
+    Dynamo _ -> pure ()
+    Objector _ -> pure ()
+    Jumper nextJumpVar goodPoint state ->
+        case (state, jumpResult) of
+          (FoundIntersection, _) ->
+            -- Ignore jump results when we already found the intersection.
+            pure ()
+          (Happy, AcceptedJump goodPoint') -> do
```
facundominguez, line 117: It is looking to me like the candidate fragment of the jumper should be updated whenever the jump is accepted, no matter what the state of the Jumper is.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,294 @@
...
+            -- The jump was accepted; we bump the peer's candidate fragment to
+            -- the dynamo's candidate fragment up to the accepted point.
+            dynamoFragment <- fromJust <$> getDynamoFragment handlesVar
+            let slicedDynamoFragment = fst $ fromJust $ splitAfterPoint dynamoFragment goodPoint'
```
facundominguez, line 121: This `fromJust` call might fail if the dynamo changes right before performing the jump, or if the dynamo does a rollback.

A possible fix is to store the dynamoFragment together with the jump offer, so the candidate fragment can be updated afterwards regardless of what happens with the dynamo.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: > Hold objectors accountable for their earlier objections.

After more discussion I offer the following fix.

If the first header of the objector after the intersection agrees with the dynamo and the slot is inside the last genesis window, move the objector to a new state `NoCSJer` to which no jumps are offered, and from which headers are downloaded via ChainSync as if CSJ was disabled.

If the slot is before the last genesis window, disconnect the peer, as the rollback is too deep.

CSJ should be turned off when no jumpers are left.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: Relatedly, the `fromJust` call might also fail if the dynamo makes a rollback.

If `badPoint` is not in the fragment, I think the last jump should be retried.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -128,6 +128,10 @@ nextInstruction context =
         handles <- readTVar (handlesVar context)
         forM_ (Map.elems handles) $ \ChainSyncClientHandle{cschJumping = cschJumping'} ->
           readTVar cschJumping' >>= \case
+            -- REVIEW: We are now proposing a jump to 'headPoint', which is the
+            -- first block _after_ (including) @lastJumpSlot + jumpSize@. We
+            -- might want to propose the jump to the last block _before_ that
+            -- point.
```
facundominguez, line 134: Why so?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -136,6 +146,102 @@ nextInstruction context =
...
         writeTVar (cschJumping (handle context)) $ Dynamo (headSlot dynamoFragment)
 
+-- | Given a 'SlotNo', find the block in the dynamo's candidate fragment that is
+-- either right before if there is one or right after otherwise. Returns
+-- 'Nothing' if there is no block in the fragment.
+--
+-- NOTE: The candidate fragment is contained within a genesis window of the
+-- common intersection. Therefore, we now that the selected block, if it exists,
+-- will also be in that genesis window, making it validable and therefore
+-- eligible for jumping.
```
facundominguez, line 156: Looks to me like the candidate fragment can extend beyond the genesis window.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -136,6 +146,102 @@ nextInstruction context =
             _ -> pure ()
         writeTVar (cschJumping (handle context)) $ Dynamo (headSlot dynamoFragment)
 
+-- | Given a 'SlotNo', find the block in the dynamo's candidate fragment that is
+-- either right before if there is one or right after otherwise. Returns
+-- 'Nothing' if there is no block in the fragment.
+--
+-- NOTE: The candidate fragment is contained within a genesis window of the
+-- common intersection. Therefore, we now that the selected block, if it exists,
```
facundominguez, line 154:
```suggestion
-- common intersection. Therefore, we know that the selected block, if it exists,
```

facundominguez:
I simplified a bit the logic of `jgTriggerJumping`. Instead of triggering jumps, I'm requesting a new dynamo election. This change is going to be followed by another commit that causes dynamo election to produce a client which is not idling or blocked on the forecast horizon (same as the algorithm description says).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: Following the CSJ description, this measure is not necessary. Proscribing historical rollbacks might be helpful, but it is an optimization, if we elect new objectors for as long as GDD is inconclusive.

facundominguez:
Here's my last attempt at defining the algorithm.

----------
Elect only dynamos that are not idle.

When a dynamo sends `MsgAwaitReplay` demote it to jumper and elect a new one. CSJ should be turned off when no dynamo is eligible.

If the dynamo receives a header after the first genesis window after the last jump that it requested, ask to jump to the tip of the candidate fragment. If this tip is at the last requested jump already, disconnect the dynamo and elect a new one.

If a jump is rejected, promote the jumper to objector. No need to find the exact intersection.

If the dynamo is disconnected, elect a new dynamo, all objectors become jumpers. If the objectors are disconnected, go on as normal.

We need to disable deep rollbacks, or otherwise objectors might agree with the dynamo after rejecting jumps. To detect rollbacks, we need to remember the rejected jump, and compare it with the actual header when it is sent.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1140,16 +1140,19 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
...
+                -- Pause LoP while waiting, we'll resume it if we get `RunNormally`
                 traceWith tracer TraceJumpingWaitingForNextInstruction
+                lbPause loPBucket
                 instruction <- jgNextInstruction jumping
                 traceWith tracer $ TraceJumpingInstructionIs instruction
                 case instruction of
                     Jumping.JumpTo point ->
-                          continueWithState kis
+                      continueWithState kis
                         $ drainThePipe n
```
facundominguez, line 1151: `drainThePipe` is restarting the leaking. Something should be done about it.

facundominguez: Resolved by 3c81a1b

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,489 @@
...
+    Objector _ -> pure ()
+    Jumper nextJumpVar goodPoint jumperState ->
+        case jumpResult of
+          AcceptedJump fragment -> do
+            -- The jump was accepted; we set the jumper's candidate fragment to
+            -- the dynamo's candidate fragment up to the accepted point.
+            --
+            -- The candidate fragments of jumpers don't grow otherwise, as only the
+            -- objector and the dynamo request further headers.
+            modifyTVar (cschState (handle context)) $ \csState -> csState {csCandidate = fragment}
```
facundominguez, line 268: Is there any restriction to set the candidate fragment here? e.g. does it need to be anchored at the immutable tip or does it need to contain the immutable tip?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+{-# LANGUAGE RankNTypes           #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE StandaloneDeriving   #-}
+{-# LANGUAGE TupleSections        #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | ChainSync jumping (CSJ) is an optimization for the ChainSync protocol that
+-- allows nodes to sync without downloading headers from all of the honest
+-- peers. This load is undesirable as it slows down all the peers involved.
```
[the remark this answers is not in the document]
facundominguez: Happy to go with another text.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+    -- that depends on the candidate fragment (like the header state history).
+    -- Restarts are requested when a client transitions from being a jumper
+    -- to being an objector or dynamo.
+    jgNextInstruction   :: !(m (Instruction blk)),
+
+    -- | To be called whenever the peer claims to have no more headers.
+    jgOnAwaitReply      :: !(m ()),
+
+    -- | To be called whenever a header is received from the peer.
+    jgOnRollForward     :: !(Point (Header blk) -> m ()),
```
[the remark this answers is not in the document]
facundominguez:
```suggestion
    -- | To be called whenever a header is received from the peer
    -- before it is validated.
    jgOnRollForward     :: !(Point (Header blk) -> m ()),
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+                max
+                  (Just $ AF.headSlot badFragment)
+                  (csLatestSlot csState)
+            }
+    _ -> pure ()
+  where
+    -- Avoid redundant constraint "HasHeader blk" reported by some ghc's
+    _ = getHeaderFields @blk
+
+-- | This function is called when we receive a 'MsgRollForward' message.
```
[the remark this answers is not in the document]
facundominguez:
```suggestion
-- | This function is called when we receive a 'MsgRollForward' message before
-- validating it.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,656 @@
...
+        Just nextFragment -> do
+          let len = AF.length nextFragment
+          if len <= 1 then do
+            -- If the fragment only contains the bad tip, we know the
+            -- intersection is the good point.
+            writeTVar (cschJumping (handle context)) $
+              Jumper nextJumpVar goodPoint $
+                FoundIntersection $ AF.headPoint badFragment
+            demoteObjector (stripContext context)
+            electNewObjector (stripContext context)
```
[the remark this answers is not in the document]
facundominguez: Looking at the code, I think electing the same objector is as if it never had been demoted.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Run.hs
```diff
@@ -219,10 +228,20 @@ dispatchTick tracer stateTracer chainDb varHandles peers (number, (duration, Pee
...
       currentChain <- atomically $ ChainDB.getCurrentChain chainDb
-      csState <- atomically $ do
+      (csState, jumpingStates) <- atomically $ do
          m <- readTVar varHandles
-         traverse (readTVar . CSClient.cschState) (m Map.!? pid)
-      traceWith tracer $ TraceNewTick number duration (Peer pid state) currentChain (CSClient.csCandidate <$> csState)
+         csState <- traverse (readTVar . CSClient.cschState) (m Map.!? pid)
+         jumpingStates <- forM (Map.toList m) $ \(peer, h) -> do
+           st <- readTVar (CSClient.cschJumping h)
+           pure (peer, st)
```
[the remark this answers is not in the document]
facundominguez: I don't feel strong about that. You can merge that if you prefer it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -163,6 +169,23 @@ data ChainSyncLoPBucketConfig
...
+    -- the chain of its peer.
+    CSJDisabled
+  |
+    -- | Enable ChainSync Jumping
+    CSJEnabled CSJEnabledConfig
+
+data CSJEnabledConfig = CSJEnabledConfig {
+  -- | The _ideal_ size for ChainSync jumps. Note that the algorithm
+  -- is best-effort: there might not be exactly `csjcJumpSize` between two jumps,
+  -- depending on the chain.
```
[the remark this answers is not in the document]
facundominguez: There can be more slots between jumps if there is not a block exactly at the boundary.

There can be less slots, if a dynamos is elected and it requires an initial jump to a slot that is earlier than the jump size.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+-- 1. They both server the same chain
+-- 2. They both claim to have no more headers
+--
+-- To avoid (1) CSJ checks that the objector disagrees with the dynamo at the
+-- point it claimed to disagree as a jumper. If the objector agrees with the
+-- dynamo, it is disengaged. A disengaged peer is not asked to jump or act as
+-- dynamo or objector. Instead, it continues to offer headers for the rest of
+-- the syncing. When the objector is disengaged, a new objector is elected
+-- among the dissenting jumpers. If there are no dissenting jumpers left, the
+-- syncing continues with the dynamo and the remaining jumpers.
```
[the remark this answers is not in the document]
facundominguez: > Or is the idea that this situation could only happen on the historical chain with adversarial peers because it cannot happen without rollbacks?

I don't think peers can be disengaged unless they do historical rollbacks, or they do honest rollbacks when the node is almost caught up.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+    HasHeader blk,
+    HasHeader (Header blk),
+    NoThunks (Header blk)
+  ) =>
+  PeerContext m peer blk ->
+  Jumping m blk
+mkJumping peerContext = Jumping
+  { jgNextInstruction = do
+      -- Demoted objectors might need to set their candidate fragment before
+      -- blocking for the next jump.
```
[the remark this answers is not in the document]
facundominguez: Changed approaches in CSJ so this is no longer needed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/State.hs
```diff
@@ -0,0 +1,143 @@
...
+  | -- | Headers continued to be downloaded from 'Disengaged' peers. They
+    -- are not requested to jump, nor elected as dynamos or objectors.
+    Disengaged
+  | -- | The jumpers can be in arbitrary numbers. They are queried regularly to
+    -- see if they agree with the chain that the dynamo is serving; otherwise,
+    -- they become candidates to be the objector. See
+    -- 'ChainSyncJumpingJumperState' for more details.
+    Jumper
+      -- | A TVar containing the next jump to be executed.
+      !(StrictTVar m (AnchoredFragment (Header blk)))
```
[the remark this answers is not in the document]
facundominguez: Changed to `Maybe (JumpInfo blk)`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,676 @@
...
+        retry
+      else do
+        writeTVar nextJumpVar (AF.Empty AF.AnchorGenesis)
+        pure $ JumpTo fragment
+
+
+-- A demoted objector might end up with a shorter candidate fragment than
+-- the last jump. This is because the objector restarts chain sync and might
+-- not download sufficient headers before being demoted. Here we correct the
+-- candidate fragment before blocking for the next jump.
```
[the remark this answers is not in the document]
facundominguez: This was undone to avoid restarting ChainSync.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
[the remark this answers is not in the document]
Niols: There is still the change that `electNewObjector` finds another objector that forks from the same point as the current one, so then it would have been better to just keep the current one. This might in particular occur if the objector is honest, because we expect many other honest peers to object and they will all object with the same intersection with the dynamo.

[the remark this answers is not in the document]
facundominguez: I tried to address that in 7cee943.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -315,6 +316,10 @@ data LoPBucketParams = LoPBucketParams {
   lbpRate     :: Rational
   }
 
+data CSJParams = CSJParams {
+  csjpJumpSize :: SlotNo
+}
```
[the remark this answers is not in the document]
facundominguez: I explained this a bit in 0abe52640.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,661 @@
...
+        Just nextFragment -> do
+          let len = AF.length nextFragment
+          if len <= 1 then do
+            -- If the fragment only contains the bad tip, we know the
+            -- intersection is the good point.
+            writeTVar (cschJumping (handle context)) $
+              Jumper nextJumpVar goodPoint $
+                FoundIntersection $ AF.headPoint badFragment
+            findObjector (stripContext context) >>=
+              electNewObjector (stripContext context)
```
Niols, line 483: Now that I think about it some more, this sounds a bit overkill. When we figure out the intersection of a peer, there is really only two cases:

- If there is no objector, the peer becomes the objector.

- If there is an objector, then the objector's intersection with the dynamo is the oldest (possibly equal) of all the intersection of `FoundIntersection` jumpers with the dynamo, so we only need to compare this new `FoundIntersection` jumper's intersection with that of the objector and switch objector if it is strictly older.

This is based on the following invariant that all endpoints respect:

- If there exist `FoundIntersection` jumpers, then there is an objector and the intersection of the objector with the dynamo is older or equal to all the intersections of the `FoundIntersection` jumpers with the dynamo.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,661 @@
...
+  , jgProcessJumpResult = atomically . processJumpResult peerContext
+  , jgUpdateJumpInfo = updateJumpInfo peerContext
+  }
+
+-- | A context for ChainSync jumping, pointing for some data.
+--
+-- Invariants:
+--
+-- - If 'handlesVar' is not empty, then there is exactly one dynamo in it.
+-- - There is at most one objector in 'handlesVar'.
```
Niols, line 203:
```suggestion
-- | A context for ChainSync jumping, pointing for some data.
--
-- Invariants:
--
-- - If 'handlesVar' is not empty, then there is exactly one dynamo in it.
-- - There is at most one objector in 'handlesVar'.
-- - If there exist 'FoundIntersection' jumpers in 'handlesVar', then there
--   is an objector and the intersection of the objector with the dynamo is
--   older or equal to all the intersections of the `FoundIntersection` jumpers
--   with the dynamo.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -0,0 +1,665 @@
...
+-- compete with the dynamo, or are in agreement with it, or are disengaged
+-- (see next section).
+--
+-- The algorithm might still download headers redundantly from peers that do
+-- historical rollbacks. These rollbacks, however, constitute dishonest
+-- behavior, and CSJ does not concern itself with avoiding load to dishonest
+-- peers. Avoiding the load induced by dishonest peers on the syncing node would
+-- require additionally to disconnect peers that do historical rollbacks. This
+-- is not done by CSJ.
+--
```
Niols, line 43:
```suggestion
--
-- Overview of the state transitions
-- ---------------------------------
--
-- See 'ChainSyncJumpingState' for the implementation of the states.
--
--                 j       ╔════════╗
--             ╭────────── ║ Dynamo ║ ◀─────────╮
--             │           ╚════════╝           │f
--             ▼                  ▲             │
--     ┌────────────┐             │     k     ┌──────────┐
--     │ Disengaged │ ◀───────────│────────── │ Objector │
--     └────────────┘       ╭─────│────────── └──────────┘
--                          │     │             ▲    ▲ │
--                         g│     │e         b  │    │ │
--                          │     │       ╭─────╯   i│ │c
--                  ╭╌╌╌╌╌╌╌▼╌╌╌╌╌╌╌╌╌╌╌╌╌│╌╌╌╌╌╌╌╌╌╌│╌▼╌╌╌╮
--                  ┆ ╔═══════╗  a   ┌──────┐  d   ┌─────┐ |
--                  ┆ ║ Happy ║ ───▶ │ LFI* │ ───▶ │ FI* │ |
--                  ┆ ╚═══════╝ ◀─╮  └──────┘      └─────┘ |
--                  ┆ Jumper      ╰─────┴────────────╯h    |
--                  ╰╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╯
--
-- *: LookingForIntersection and FoundIntersection, abbreviated for this
--    drawing only; this abbreviation will not be used elsewhere.
--
-- 0. A new peer starts as the dynamo if there is no other peer or as a Happy
--    jumper otherwise. The dynamo periodically requests jumps from happy
--    jumpers who, in the ideal case, accept them.
--
-- 1. In the event that a jumper rejects a jump, it goes from Happy to LFI* (a).
--    From there starts a back-and-forth of intersection search messages until
--    the exact point of disagreement with the dynamo is found.
--
-- 2. Once the exact point of disagreement is found, and if there is no objector
--    yet, the jumper becomes the objector (b). If there is an objector, then we
--    compare the intersections of the objector and the jumper with the dynamo. If
--    the jumper's intersection is strictly older, then the jumper replaces the
--    dynamo (b+c). Otherwise, the jumper is marked as FI* (d).
--
-- 3. If the dynamo disconnects, one peer is elected as the new dynamo (e|f) and
--    all other peers revert to being happy jumpers (g+h).
--
-- 4. If the objector disconnects, and there are FI* jumpers, then the one with the
--    oldest intersection with the dynamo gets elected (i).
--
-- 5. If the dynamo rolls back to a point older than the last jump it requested, it
--    is disengaged (j) and a new dynamo is elected (e|f).
--
-- 6. If the objector agrees with the dynamo, it is disengaged (k). If there are FI*
--    jumpers, then one of them gets elected as the new objector (i).
--
```

Do you think something like this is useful? I might have missed transitions, and also I'm not sure if we're implementing exactly this, in particular for (2), but that's related to my previous review comments.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: Tried another fix at b6f08ca.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: Looks good to me! It should be useful to those approaching the implementation by first time.

PR #1047, opened 2024-04-08
Conway: restrict VRF tiebreaker based on slot distance

### LINKED ISSUE #524: Ignore the VRF tiebreaker when the slots are too far apart
From comment https://github.com/input-output-hk/ouroboros-network/issues/2913#issuecomment-1781407345

aka "Duncan's tiebreaker"

amesgen:
Closes #524 

This is a revived version of https://github.com/IntersectMBO/ouroboros-network/pull/3721

Based on top of #1063

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20240409_142855_alexander.esgen_duncan_tiebreaker.md
+++ b/ouroboros-consensus-cardano/changelog.d/20240409_142855_alexander.esgen_duncan_tiebreaker.md
@@ -1,0 +1,3 @@
+### Breaking
+
+- Restricted the VRF tiebreaker based on slot distance starting in Conway.
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Config.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Config.hs
@@ -4,7 +4,10 @@
 {-# LANGUAGE DerivingVia                #-}
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns             #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 module Ouroboros.Consensus.Shelley.Ledger.Config (
@@ -29,7 +32,9 @@ import           NoThunks.Class (NoThunks (..))
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Config
-import           Ouroboros.Consensus.Shelley.Eras (EraCrypto)
+import           Ouroboros.Consensus.Protocol.Praos.Common
+                     (VRFTiebreakerFlavor (..))
+import           Ouroboros.Consensus.Shelley.Eras (EraCrypto, isBeforeConway)
 import           Ouroboros.Consensus.Shelley.Ledger.Block
 import           Ouroboros.Network.Magic (NetworkMagic (..))
 
@@ -40,14 +45,15 @@ import           Ouroboros.Network.Magic (NetworkMagic (..))
 data instance BlockConfig (ShelleyBlock proto era) = ShelleyConfig {
       -- | The highest protocol version this node supports. It will be stored
       -- the headers of produced blocks.
-      shelleyProtocolVersion  :: !SL.ProtVer
-    , shelleySystemStart      :: !SystemStart
-    , shelleyNetworkMagic     :: !NetworkMagic
+      shelleyProtocolVersion     :: !SL.ProtVer
+    , shelleySystemStart         :: !SystemStart
+    , shelleyNetworkMagic        :: !NetworkMagic
       -- | For nodes that can produce blocks, this should be set to the
       -- verification key(s) corresponding to the node's signing key(s). For non
       -- block producing nodes, this can be set to the empty map.
-    , shelleyBlockIssuerVKeys :: !(Map (SL.KeyHash 'SL.BlockIssuer (EraCrypto era))
-                                       (SL.VKey 'SL.BlockIssuer (EraCrypto era)))
+    , shelleyBlockIssuerVKeys    :: !(Map (SL.KeyHash 'SL.BlockIssuer (EraCrypto era))
+                                          (SL.VKey 'SL.BlockIssuer (EraCrypto era)))
+    , shelleyVRFTiebreakerFlavor :: !VRFTiebreakerFlavor
     }
   deriving stock (Generic)
 
@@ -55,20 +61,29 @@ deriving instance ShelleyBasedEra era => Show     (BlockConfig (ShelleyBlock pro
 deriving instance ShelleyBasedEra era => NoThunks (BlockConfig (ShelleyBlock proto era))
 
 mkShelleyBlockConfig ::
-     ShelleyBasedEra era
+     forall proto era. ShelleyBasedEra era
   => SL.ProtVer
   -> SL.ShelleyGenesis (EraCrypto era)
   -> [SL.VKey 'SL.BlockIssuer (EraCrypto era)]
   -> BlockConfig (ShelleyBlock proto era)
 mkShelleyBlockConfig protVer genesis blockIssuerVKeys = ShelleyConfig {
-      shelleyProtocolVersion  = protVer
-    , shelleySystemStart      = SystemStart  $ SL.sgSystemStart  genesis
-    , shelleyNetworkMagic     = NetworkMagic $ SL.sgNetworkMagic genesis
-    , shelleyBlockIssuerVKeys = Map.fromList
+      shelleyProtocolVersion     = protVer
+    , shelleySystemStart         = SystemStart  $ SL.sgSystemStart  genesis
+    , shelleyNetworkMagic        = NetworkMagic $ SL.sgNetworkMagic genesis
+    , shelleyBlockIssuerVKeys    = Map.fromList
         [ (SL.hashKey k, k)
         | k <- blockIssuerVKeys
         ]
+    , shelleyVRFTiebreakerFlavor
     }
+  where
+    shelleyVRFTiebreakerFlavor
+      | isBeforeConway (Proxy @era)
+      = UnrestrictedVRFTiebreaker
+      | otherwise
+      -- See 'RestrictedVRFTiebreaker' for context. 5 slots is the "usual" value
+      -- we consider when talking about the maximum propagation delay.
+      = RestrictedVRFTiebreaker 5
 
 {-------------------------------------------------------------------------------
   Codec config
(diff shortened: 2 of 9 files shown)


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+  | -- | Only compare the VRF tiebreakers when the slot numbers differ by at
+    -- most the given number of slots.
+    --
+    -- The main motivation is as follows:
+    --
+    -- When two blocks A and B with the same block number differ in their slot
+    -- number by more than Δ (the maximum message delay from Praos), say
+    -- @slot(A) + Δ < slot(B)@, the issuer of B should have been able to mint a
+    -- block with a block number higher than A (eg by minting on top of A).
+    -- Therefore, we do not want to allow B to win against A by having a better
```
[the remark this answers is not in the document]
nfrisby: It occurs to me that today's excessively granular tiebreakers mean that short forks resolve themselves ASAP on mainnet. As of this change, the network instead won't resolve until someone creates a sole longest chain again (ie the actual Praos paper). I wonder if the community needs to be "warned" about short forks becoming "relatively long lasting" after this change?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+  deriving anyclass (NoThunks)
+
+-- Used to implement the 'Ord' and 'ChainOrder' instances for Praos.
+comparePraos ::
+  Crypto c =>
+  VRFTiebreakerFlavor ->
+  PraosChainSelectView c ->
+  PraosChainSelectView c ->
+  Ordering
+comparePraos tiebreakerFlavor =
```
nfrisby, line 88: How about this instead?

```
  mconcat
    [ compare `on` csvChainLength,
      when' ((==) `on` csvIssuer) (compare `on` csvIssueNo),
      when' vrfArmed  (compare `on` Down . csvTieBreakVRF)
    ]  where
    -- When the given condition is satisfied, use the given comparison
    -- function, otherwise, no preference.
    when' b comp v1 v2 = if b v1 v2 then comp v1 v2 else EQ

    vrfArmed v1 v2 = case tiebreakerFlavor of
        UnrestrictedVRFTiebreaker       -> True
        RestrictedVRFTiebreaker maxDist ->
            slotDist (csvSlotNo v1) (csvSlotNo v2) <= maxDist

    slotDist :: SlotNo -> SlotNo -> SlotNo
    slotDist s t
      -- slot numbers are unsigned, so have to take care with subtraction
      | s >= t    = s - t
      | otherwise = t - s
```

nfrisby: And, more of a personal preference, since we're in the neighborhood :)

```
  (compare `on` csvChainLength)
    <> when' ((==) `on` csvIssuer) (compare `on` csvIssueNo)
    <> when' vrfArmed  (compare `on` Down . csvTieBreakVRF)
```

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+--
+-- 1. The abstract Consensus layer requires that we first compare based on chain
+--    length (see __Chain extension precedence__ in 'ChainOrder').
+--
+-- 2. Consider the scenario where the hot key of a block issuer was compromised,
+--    and the attacker is now minting blocks using that identity. The actual
+--    block issuer can use their cold key to issue a new hot key with a higher
+--    opcert issue number and set up a new pool. Due to this tiebreaker rule,
+--    the blocks minted by that pool will take precedence (allowing the actual
+--    block issuer to decide on the block contents), and they will end up on the
```
nfrisby, line 197: not just the contents, but the predecessor as well

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+--    honest chain quickly, which means that the adversary can't extend any
+--    chain containing such a block as it would violate the monotonicity
+--    requirement on opcert issue numbers.
+--
+--     See "3.7 Block Validity and Operational Key Certificates" in "Design
+--     Specification for Delegation and Incentives in Cardano" by Kant et al for
+--     more context.
+--
+-- 3. The main motivation to do VRF comparisons the usually final tiebreaker is
```
nfrisby, line 206: > The main motivation to do VRF comparisons the usually final tiebreaker ...

Something is non-grammatical in there; not sure what you wanted. Perhaps:

> The main motivation to do VRF comparisons, which is usually the final tiebreaker, ...

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+--    to avoid the "Frankfurt problem":
+--
+--     With only the first two rules for the chain order, almost all blocks with
+--     equal block number are equally preferrable. Consider two block issuers
+--     minting blocks in very nearby slots. As we never change our selection
+--     from one chain to an equally preferrable one, the first block to arrive
+--     at another pool is the one to be adopted, and will be extended the next
+--     time the pool is elected if no blocks with a higher block number arrive
+--     in the meantime. This incentivizes block producers to concentrate
```
nfrisby, line 215: > This incentivizes ...

Perhaps we should say "We observed that this effectively incentivizes ...", since it was never established whether a fully rational actor would actually be incentivized :)

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
...
+--     geographically (historically, in Frankfurt) in order to minimize their
+--     diffusion times. This works against the goal of geographic
+--     decentralisation.
+--
+--     With the VRF tiebreaker, also a block with a somewhat lower propagation
```
nfrisby, line 220: > Also, with the VRF tiebreaker, a block...

ouroboros-consensus-cardano/changelog.d/20240409_142855_alexander.esgen_duncan_tiebreaker.md
```diff
@@ -0,0 +1,3 @@
+### Breaking
```
amesgen, line 1: tmp

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Config.hs
```diff
@@ -40,35 +45,48 @@ import           Ouroboros.Network.Magic (NetworkMagic (..))
...
     }
+  where
+    shelleyVRFTiebreakerFlavor
+      | isBeforeConway (Proxy @era)
+      = UnrestrictedVRFTiebreaker
+      | otherwise
+      -- See 'RestrictedVRFTiebreaker' for context. 5 slots is the "usual" value
+      -- we consider when talking about the maximum propagation delay.
+      --
+      -- TODO derive/clamp this value from something else, eg active slot
```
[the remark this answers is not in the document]
amesgen: Yes, removed the TODO :+1:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/DiffusionPipelining.hs
```diff
@@ -20,7 +20,6 @@ module Ouroboros.Consensus.Shelley.Node.DiffusionPipelining (
   , isBeforeConway
```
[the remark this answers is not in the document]
amesgen: Right, removed it :+1:

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -54,36 +55,178 @@ data PraosChainSelectView c = PraosChainSelectView
   }
   deriving (Show, Eq, Generic, NoThunks)
 
+-- | When to compare the VRF tiebreakers.
+data VRFTiebreakerFlavor =
+    -- | Always compare the VRF tiebreakers. This is the behavior of all eras
+    -- before Conway. Once mainnet has transitioned to Conway, we can remove
+    -- this option.
```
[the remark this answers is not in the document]
amesgen: Yes, we could, though the "5 slots" would then be "more hardcoded" than ATM.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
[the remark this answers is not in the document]
amesgen: > There's also a scenario in which case A was the poorly managed pool. Perhaps B didn't extend A because A was very late sending their block. The tiebreaker still works in that case, since B would (likely) have arrived to the network before A.

Good point, will expand this to make it clear that both A and B could be the culprit here.

> It occurs to me that today's excessively granular tiebreakers mean that short forks resolve themselves ASAP on mainnet. As of this change, the network instead won't resolve until someone creates a sole longest chain again (ie the actual Praos paper). I wonder if the community needs to be "warned" about short forks becoming "relatively long lasting" after this change?

Yeah, that is an advantage of the current "deterministic"[^det]/"consistent"[^con] tiebreaker. I think it makes sense to mention that in https://github.com/IntersectMBO/ouroboros-network/issues/2913 when the HF approaches :+1:

[^det]: That's how it is called eg in [CAD-2545](https://input-output.atlassian.net/browse/CAD-2545).
[^con]: That's how it is called in [this paper](https://eprint.iacr.org/2020/041.pdf) that the researchers referred us to when we talked about this; the high-level summary back then was that you get *slightly* better settlement bounds with a consistent tiebreaker. The paper only seems to prove a similar asymptotic bound under weaker conditions (Theorem 2 vs Theorem 1); but this is fully consistent that.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
amesgen: Yeah, moving this function around seems like a good opportunity to condense the logic :+1:

PR #1056, opened 2024-04-11
Have GDD disconnect peers with disagreeing chains of equal density

facundominguez:
Before this change, the GDD would tolerate different chains having the same density. This is problematic to ChainSync jumping if the only active peers (dynamo and objector) disagree with chains of the same density, and GDD cannot disconnect any of them.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
@@ -269,8 +269,11 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
       guard $ AF.lastPoint frag0 /= AF.lastPoint frag1
       -- peer1 offers more than k blocks
       guard offersMoreThanK
-      -- peer1 definitely has higher density than peer0
-      guard $ lb1 > ub0
+      -- peer1 has the same or better density than peer0
+      --
+      -- This matters to ChainSync jumping, where dynamo and objector
+      -- could offer chains of equal density.
+      guard $ lb1 >= ub0
       pure peer0
 
     loeIntersectionSlot = AF.headSlot loeFrag
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
@@ -260,10 +260,19 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
       pure (peer, DensityBounds {fragment, offersMoreThanK, lowerBound, upperBound, hasBlockAfter, latestSlot, idling})
 
     losingPeers = nubOrd $ do
-      (peer0 , DensityBounds {fragment = frag0, upperBound = ub0}) <-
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -260,10 +260,19 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
...
       (_peer1, DensityBounds {fragment = frag1, offersMoreThanK, lowerBound = lb1 }) <-
         Map.toList densityBounds
+      -- Don't disconnect peer0 if it sent no headers after the intersection yet
+      -- and it is not idling.
+      --
+      -- See Note [Chain disagreement]
+      guard $ idling0 || not (AF.null frag0) || hasBlockAfter0
       -- ensure that the two peer fragments don't share any
       -- headers after the LoE
       guard $ AF.lastPoint frag0 /= AF.lastPoint frag1
```
amesgen, line 278: Ah right, when we disconnect from someone due to `lb1 == ub0`, we need to be sure that they actually committed to fragments such that their intersection is the LoE tip, which is *not* guaranteed by the existing `AF.lastPoint frag0 /= AF.lastPoint frag1`.

Maybe we could extract this logic (these two `guard`s) to a separate function `definitelyIntersectAtLoE` and test it, as the logic isn't immediately obvious?

amesgen: Hmm, what happens when both the dynamo and the objector serve a header after the forecast horizon? Ideally, the GDD should disconnect from both of them, but I don't think it does currently, as `guard $ AF.lastPoint frag0 /= AF.lastPoint frag1` will prevent that. This seems to be an exception to the rule "the fragments have to disagree after the intersection".

Maybe that case is handled in an ad-hoc fashion in the CSJ logic though; but it seems good to be robust here.

facundominguez: > Hmm, what happens when both the dynamo and the objector serve a header after the forecast horizon?

I think this is a good point that deserves another PR.

facundominguez: > Maybe we could extract this logic (these two guards) to a separate function definitelyIntersectAtLoE and test it, as the logic isn't immediately obvious?

I'd be happy to review some proposal to do that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -269,8 +269,11 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
...
       -- peer1 offers more than k blocks
       guard offersMoreThanK
-      -- peer1 definitely has higher density than peer0
-      guard $ lb1 > ub0
+      -- peer1 has the same or better density than peer0
+      --
+      -- This matters to ChainSync jumping, where dynamo and objector
+      -- could offer chains of equal density.
+      guard $ lb1 >= ub0
```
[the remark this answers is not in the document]
facundominguez: Adopted in 70391ae52.

facundominguez:
Merged manually as 27a11610d.

PR #1061, opened 2024-04-12
Fix prop_densityDisconnectMonotonic

tek:
The test logic failed to ever update the `killed` flag that indicated
that a peer was selected for disconnection before, essentially making
the test tautological.

This revealed the need for another fix – the test must be terminated
when the LOE fragment advances beyond the honest chain (because a branch
forks off the last block).

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
@@ -1,29 +1,37 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE ViewPatterns          #-}
 module Test.Consensus.Genesis.Tests.DensityDisconnect (tests) where
 
-import           Cardano.Slotting.Slot (WithOrigin (..), unSlotNo)
+import           Cardano.Slotting.Slot (SlotNo (unSlotNo), WithOrigin (..))
 import           Control.Exception (fromException)
 import           Control.Monad.Class.MonadTime.SI (Time (..))
-import           Data.Bifunctor (second)
-import           Data.Foldable (minimumBy, toList)
+import           Data.Bifunctor
+import           Data.Foldable (maximumBy, minimumBy, toList)
 import           Data.Function (on)
 import           Data.Functor (($>), (<&>))
 import           Data.List (intercalate)
+import           Data.List.NonEmpty (nonEmpty)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import           Data.Maybe (fromMaybe)
 import           Data.Semigroup (Endo (..))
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/DensityDisconnect.hs
```diff
@@ -278,7 +295,12 @@ evolveBranches EvolvingPeers {k, sgen, peers = initialPeers} =
               }
           (loeFrag, suffixes) = sharedCandidatePrefix curChain candidates
           disconnect = fst (densityDisconnect sgen k states suffixes loeFrag)
-      either (pure . second (result loeFrag)) step (updatePeers sgen target disconnect next)
+          updated
+            | AF.anchorToSlotNo (AF.anchor loeFrag) > AF.headSlot (candidate (value (honest ps)))
+            = Left (Finished, ps)
```
facundominguez, line 300: Why must the test be terminated in this case?

I would expect this to not be possible unless the honest chain is disconnected.

tek: because the adversary forks off the last block of the honest chain (maybe this only happens here because the honest candidate wasn't completed by the random fragment-extender?)

facundominguez: > because the adversary forks off the last block of the honest chain

I follow until there. I think you mean that the adversary forks off the last block of the candidate fragment of the honest peer. How does that become problematic then?

tek: now that you say it like that, I realize that there should still be an intersection...

I only saw that my in my test, the LoE frag was anchored in the adversarial fragment, and assumed that the reason was that there was no actual "fork" (i.e. the honest chain had no extra blocks).
but `sharedCandidatePrefix` should still use the last honest block as the anchor, I guess.

facundominguez:
I rebased the base branch after merging #1056 there.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
```diff
@@ -471,6 +453,33 @@ terseGDDEvent = \case
...
+      ["      Density bounds:"] ++
+      prettyDensityBounds bounds ++
+      ["      New candidate tips:"] ++
+      showPeers (tersePoint . castPoint <$> Map.map AF.headPoint candidateSuffixes) ++
+      [
+        "      Losing peers: " ++ show losingPeers,
+      "      Setting loeFrag: " ++ terseAnchor (AF.castAnchor loeHead)
+      ]
+  where
+
```
facundominguez, line 482: I'm not getting what is the difference in this diff, and I'm seeing that some conflict is going to appear here with csj related changes.

facundominguez: I think I get it after some more staring.

tek: just that `prettyDensityBounds` was pulled out

facundominguez (approved):

PR #1046, opened 2024-04-08
Refactor chain ordering away from `Ord`

amesgen:
Closes #939

This PR is intended to be reviewed commit-by-commit.

This PR undoes some of the changes of https://github.com/IntersectMBO/ouroboros-network/pull/2743.

WIP: still needs more docs

PR #1063, opened 2024-04-17
Allow to customize `preferCandidate`

### LINKED ISSUE #939: Remove chain order transitivity requirement from the abstract Consensus layer
- [x] #938
 - [ ] Don't use/define `Ord SelectView`. In particular, in ChainSel, sorting based on `SelectView` requires thought[^1]:
       https://github.com/IntersectMBO/ouroboros-consensus/blob/ef8186daf6bd70fc82d7df5817cf033f44a0ee6f/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs#L905-L907

[^1]: Note that this only happens when blocks arrive out-of-order, so during syncing, where tiebreakers are largely irrelevant anyway.

amesgen:
Closes #939

The core idea is to keep a total order (via `Ord`) for `SelectView`, but allowing to customize the logic of `preferCandidate` (previously, it was always defined in terms of `Ord`, i.e. `preferCandidate ours cand = cand > ours`), such that it becomes possible to eg implement the restricted VRF tiebreaker (#524) this way, which is fundamentally intransitive.

This is in contrast to #1046 (an alternative to this PR), where `preferCandidate` is not customizable, and we instead live with the fact the ordering might not be transitive, which techically would require replacing our calls to `sortBy` with eg a topological sorting algorithm, or just assume/test that `sortBy` still works "well enough" with non-transitive orders.

Also see #1047 as a follow-up PR that leverages the new flexibility offered by this PR.

This PR undoes some of the changes of https://github.com/IntersectMBO/ouroboros-network/pull/2743 and https://github.com/IntersectMBO/ouroboros-network/pull/2732 which were back then merged to increase simplicity/ease of exposition in the report, but it now turns out that the flexibility they removed is actually useful for us today.

### DIFF
--- a/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
+++ b/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
@@ -2,8 +2,11 @@
 {-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DerivingVia                #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE UndecidableInstances       #-}
 
 -- | Various things common to iterations of the Praos protocol.
 module Ouroboros.Consensus.Protocol.Praos.Common (
@@ -41,15 +44,7 @@ newtype MaxMajorProtVer = MaxMajorProtVer
   deriving (Eq, Show, Generic)
   deriving newtype NoThunks
 
--- | View of the ledger tip for chain selection.
---
--- We order between chains as follows:
---
--- 1. By chain length, with longer chains always preferred.
--- 2. If the tip of each chain was issued by the same agent, then we prefer
---    the chain whose tip has the highest ocert issue number.
--- 3. By a VRF value from the chain tip, with lower values preferred. See
---    @pTieBreakVRFValue@ for which one is used.
+-- | View of the tip of a header fragment for chain selection.
 data PraosChainSelectView c = PraosChainSelectView
   { csvChainLength :: BlockNo,
     csvSlotNo      :: SlotNo,
@@ -59,6 +54,13 @@ data PraosChainSelectView c = PraosChainSelectView
   }
   deriving (Show, Eq, Generic, NoThunks)
 
+-- | We order between chains as follows:
+--
+-- 1. By chain length, with longer chains always preferred.
+-- 2. If the tip of each chain was issued by the same agent, then we prefer
+--    the chain whose tip has the highest ocert issue number.
+-- 3. By a VRF value from the chain tip, with lower values preferred. See
+--    @pTieBreakVRFValue@ for which one is used.
 instance Crypto c => Ord (PraosChainSelectView c) where
   compare =
     mconcat
@@ -79,6 +81,9 @@ instance Crypto c => Ord (PraosChainSelectView c) where
             comp v1 v2
         | otherwise =
             EQ
+
+deriving via SimpleChainOrder (PraosChainSelectView c)
+  instance Crypto c => ChainOrder (PraosChainSelectView c)
 
 data PraosCanBeLeader c = PraosCanBeLeader
   { -- | Certificate delegating rights from the stake pool cold key (or
--- a/ouroboros-consensus-protocol/test/protocol-test/Test/Consensus/Protocol/Praos/SelectView.hs
+++ b/ouroboros-consensus-protocol/test/protocol-test/Test/Consensus/Protocol/Praos/SelectView.hs
@@ -1,0 +1,91 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Test.Consensus.Protocol.Praos.SelectView (tests) where
+
+import qualified Cardano.Crypto.Hash as Crypto
+import qualified Cardano.Crypto.Util as Crypto
+import           Cardano.Crypto.VRF (OutputVRF, mkTestOutputVRF)
+import           Cardano.Ledger.Crypto (Crypto (..), StandardCrypto)
+import qualified Cardano.Ledger.Keys as SL
+import           Codec.Serialise (encode)
+import           Control.Monad
+import           Data.Containers.ListUtils (nubOrdOn)
+import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Protocol.Praos.Common
+import           Test.Cardano.Ledger.Binary.Arbitrary ()
+import           Test.Ouroboros.Consensus.Protocol
+import           Test.QuickCheck.Gen (Gen (..))
+import           Test.QuickCheck.Random (mkQCGen)
+import           Test.Tasty
+import           Test.Tasty.QuickCheck hiding (elements)
+import           Test.Util.QuickCheck
+import           Test.Util.TestEnv
+
+tests :: TestTree
+tests = testGroup "PraosChainSelectView"
+    [   adjustQuickCheckTests (* 50)
+      -- Use a small max size by default in order to have a decent chance to
+      -- trigger the actual tiebreaker cases.
+      $ adjustQuickCheckMaxSize (`div` 10)
+      $ tests_chainOrder (Proxy @(PraosChainSelectView StandardCrypto))
+    ]
+
+instance Crypto c => Arbitrary (PraosChainSelectView c) where
+  arbitrary = do
+      size           <- fromIntegral <$> getSize
+      csvChainLength <- BlockNo <$> choose (1, size)
+      csvSlotNo      <- SlotNo  <$> choose (1, size)
+      csvIssuer      <- elements knownIssuers
+      csvIssueNo     <- genIssueNo
+      pure PraosChainSelectView {
+          csvChainLength
+        , csvSlotNo
+        , csvIssuer
+        , csvIssueNo
+        , csvTieBreakVRF = mkVRFFor csvIssuer csvSlotNo
+        }
+   where
+     -- We want to draw from the same small set of issuer identities in order to
+     -- have a chance to explore cases where the issuers of two 'SelectView's
+     -- are identical.
+     knownIssuers :: [SL.VKey SL.BlockIssuer c]
+     knownIssuers =
+           nubOrdOn SL.hashKey
+         $ unGen (replicateM numIssuers (SL.VKey <$> arbitrary)) randomSeed 100
+       where
+         randomSeed = mkQCGen 4 -- chosen by fair dice roll
+         numIssuers = 10
+
+     -- TODO Actually randomize this once the issue number tiebreaker has been
+     -- fixed to be transitive. See the document in
+     -- https://github.com/IntersectMBO/ouroboros-consensus/pull/891 for
+     -- details.
+     --
+     -- TL;DR: In an edge case, the issue number tiebreaker prevents the
+     -- chain order from being transitive. This could be fixed relatively
+     -- easily, namely by swapping the issue number tiebreaker and the VRF
+     -- tiebreaker. However, this is technically not backwards-compatible,
+     -- impacting the current pre-Conway diffusion pipelining scheme.
+     --
+     -- See https://github.com/IntersectMBO/ouroboros-consensus/issues/1075.
+     genIssueNo = pure 1
+
+     -- The header VRF is a deterministic function of the issuer VRF key, the
+     -- slot and the epoch nonce. Additionally, for any particular chain, the
+     -- slot determines the epoch nonce.
+     mkVRFFor :: SL.VKey SL.BlockIssuer c -> SlotNo -> OutputVRF (VRF c)
+     mkVRFFor issuer slot =
+           mkTestOutputVRF
+         $ Crypto.bytesToNatural
+         $ Crypto.hashToBytes
+         $ Crypto.xor (Crypto.castHash issuerHash)
+         $ Crypto.hashWithSerialiser encode slot
+       where
+         SL.KeyHash issuerHash = SL.hashKey issuer
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsProtocol.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsProtocol.hs
@@ -34,3 +34,13 @@ class ( GetHeader blk
                      => BlockConfig blk
                      -> Header blk -> SelectView (BlockProtocol blk)
   selectView _ = blockNo
+
+  projectChainOrderConfig ::
+       BlockConfig blk
+    -> ChainOrderConfig (SelectView (BlockProtocol blk))
+
+  default projectChainOrderConfig ::
+       ChainOrderConfig (SelectView (BlockProtocol blk)) ~ ()
+    => BlockConfig blk
+    -> ChainOrderConfig (SelectView (BlockProtocol blk))
+  projectChainOrderConfig _ = ()
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
@@ -21,6 +21,7 @@
 module Ouroboros.Consensus.HardFork.Combinator.AcrossEras (
     -- * Value for /each/ era
     PerEraBlockConfig (..)
+  , PerEraChainOrderConfig (..)
   , PerEraCodecConfig (..)
   , PerEraConsensusConfig (..)
   , PerEraLedgerConfig (..)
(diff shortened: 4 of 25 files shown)


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -59,6 +54,13 @@ data PraosChainSelectView c = PraosChainSelectView
...
   deriving (Show, Eq, Generic, NoThunks)
 
+-- | We order between chains as follows:
+--
+-- 1. By chain length, with longer chains always preferred.
+-- 2. If the tip of each chain was issued by the same agent, then we prefer
+--    the chain whose tip has the highest ocert issue number.
+-- 3. By a VRF value from the chain tip, with lower values preferred. See
+--    @pTieBreakVRFValue@ for which one is used.
 instance Crypto c => Ord (PraosChainSelectView c) where
```
jorisdral, line 64: Is it useful to test that it is total, i.e. by using [quickcheck-classes](https://hackage.haskell.org/package/quickcheck-classes)? I'm not sure how useful this is, and how easy it is to generate an arbitrary `PraosChainSelectView`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -170,13 +176,59 @@ class ( Show (ChainDepState   p)
...
-                -> SelectView p  -- ^ Tip of our chain
-                -> SelectView p  -- ^ Tip of the candidate
-                -> Bool
-preferCandidate _ ours cand = cand > ours
+-- See 'preferCandidate' for the primary documentation.
+--
+-- Additionally, we require a total order on this type, such that eg different
+-- candidate chains can be sorted for prioritization.
+class Ord a => ChainOrder a where
+  type ChainOrderConfig a :: Type
```
jorisdral, line 187: Out of interest: would use the `ChainOrderConfig` anywhere? Currently, none of the protocols rely on a config, do they?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
amesgen: I will definitely do that, that's how I found out that the chain order is not transitive in the first place :sweat_smile:

amesgen: Now implemented :+1:

ouroboros-consensus-protocol/test/protocol-test/Test/Consensus/Protocol/Praos/SelectView.hs
```diff
@@ -0,0 +1,88 @@
...
+       where
+         randomSeed = mkQCGen 4 -- chosen by fair dice roll
+
+         -- TODO Increase this to a value like 10 once the issue number
+         -- tiebreaker has been fixed to be transitive. See the document in
+         -- https://github.com/IntersectMBO/ouroboros-consensus/pull/891 for
+         -- details.
+         --
+         -- TL;DR: In an edge case, the issue number tiebreaker prevents the
+         -- chain order to be transitive. This could be fixed relatively easily,
```
nfrisby, line 71: to be -> from being

ouroboros-consensus-protocol/test/protocol-test/Test/Consensus/Protocol/Praos/SelectView.hs
```diff
@@ -0,0 +1,103 @@
...
+         -- namely by swapping the issue number tiebreaker and the VRF
+         -- tiebreaker. However, this is technically not backwards-compatible,
+         -- impacting the current pre-Conway diffusion pipelining scheme.
```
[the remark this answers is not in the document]
nfrisby: Perhaps in this PR we just use a constant issue number? And relax that in the PR that fixes the issue numbers?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsProtocol.hs
```diff
@@ -34,3 +34,13 @@ class ( GetHeader blk
                      => BlockConfig blk
                      -> Header blk -> SelectView (BlockProtocol blk)
   selectView _ = blockNo
+
+  projectChainOrderConfig ::
+       BlockConfig blk
+    -> ChainOrderConfig (SelectView (BlockProtocol blk))
+
+  default projectChainOrderConfig ::
```
nfrisby, line 42: Bernie meme: I am once again expressing my distrust of type class `default`s. :D

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
```diff
@@ -97,11 +98,12 @@ import           Ouroboros.Consensus.Util.Condense (Condense (..))
...
-newtype PerEraLedgerConfig    xs = PerEraLedgerConfig    { getPerEraLedgerConfig    :: NP WrapPartialLedgerConfig    xs }
-newtype PerEraStorageConfig   xs = PerEraStorageConfig   { getPerEraStorageConfig   :: NP StorageConfig              xs }
+newtype PerEraBlockConfig      xs = PerEraBlockConfig      { getPerEraBlockConfig      :: NP BlockConfig                xs }
+newtype PerEraChainOrderConfig xs = PerEraChainOrderConfig { getPerEraChainOrderConfig :: NP WrapChainOrderConfig       xs }
+newtype PerEraCodecConfig      xs = PerEraCodecConfig      { getPerEraCodecConfig      :: NP CodecConfig                xs }
+newtype PerEraConsensusConfig  xs = PerEraConsensusConfig  { getPerEraConsensusConfig  :: NP WrapPartialConsensusConfig xs }
+newtype PerEraLedgerConfig     xs = PerEraLedgerConfig     { getPerEraLedgerConfig     :: NP WrapPartialLedgerConfig    xs }
+newtype PerEraStorageConfig    xs = PerEraStorageConfig    { getPerEraStorageConfig    :: NP StorageConfig              xs }
 
 newtype PerEraProtocolParams  xs = PerEraProtocolParams  { getPerEraProtocolParams  :: NP ProtocolParams             xs }
```
nfrisby, line 108: I don't know why this one is separated from the group. But it was aligned with the group you edited.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -50,55 +55,97 @@ data AcrossEraSelection :: Type -> Type -> Type where
...
+  => AcrossEraMode cfg a
+  -> NP cfg                          xs
+  -> Tails AcrossEraSelection        xs
   -> WithBlockNo (NS WrapSelectView) xs
   -> WithBlockNo (NS WrapSelectView) xs
-  -> Ordering
-acrossEraSelection = \ffs l r ->
-    goLeft ffs (distribBlockNo l, distribBlockNo r)
+  -> a
+acrossEraSelection (mode :: AcrossEraMode cfg a) = \cfg ffs l r ->
```
nfrisby, line 111: I think our style prefers using a `forall` in the signature instead of binding type variables in patterns.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -50,55 +55,97 @@ data AcrossEraSelection :: Type -> Type -> Type where
...
+data FlipArgs = KeepArgs | FlipArgs
+
 acrossEras ::
-     forall blk blk'. SingleEraBlock blk
-  => WithBlockNo WrapSelectView blk
+     forall blk blk' cfg a. SingleEraBlock blk
+  => FlipArgs
+  -> AcrossEraMode cfg a
+  -> cfg                        blk'
+     -- ^ The configuration corresponding to the later block/era.
```
nfrisby, line 83: Could you add a comment explaining why we favor the later era's config?

The only case we have is Babbage - Conway, right? Does it matter in that case, or is it just arbitrary? (My immediate intuition is that we can use the new one because we prefer it and since a new block is involved there's no risk of a breaking change of behavior.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -50,55 +55,97 @@ data AcrossEraSelection :: Type -> Type -> Type where
...
   -> WithBlockNo (NS WrapSelectView) xs
   -> WithBlockNo (NS WrapSelectView) xs
-  -> Ordering
-acrossEraSelection = \ffs l r ->
-    goLeft ffs (distribBlockNo l, distribBlockNo r)
+  -> a
+acrossEraSelection (mode :: AcrossEraMode cfg a) = \cfg ffs l r ->
+    goLeft cfg ffs (distribBlockNo l, distribBlockNo r)
   where
     goLeft ::
```
nfrisby, line 114: I know it's strictly beyond the scope, but I find the `goLeft` and `goRight` to be confusing, since `goRight` is can be applied to the right one :(

I think eg `goBoth` and `goOne` would be better?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -66,12 +74,10 @@ class ( Show (ChainDepState   p)
   -- two things independent of a choice of consensus protocol: we never switch
   -- to chains that fork off more than @k@ blocks ago, and we never adopt an
   -- invalid chain. The actual comparison of chains however depends on the chain
-  -- selection protocol. We define chain selection (which is itself a partial
-  -- order) in terms of a totally ordered /select view/ on the headers at the
-  -- tips of those chains: chain A is strictly preferred over chain B whenever
-  -- A's select view is greater than B's select view. When the select view on A
-  -- and B is the same, the chains are considered to be incomparable (neither
-  -- chain is preferred over the other).
+  -- selection protocol. We define chain selection in terms of a /select view/
```
nfrisby, line 77: It occurs to me: does the tutorial Galois wrote need to be updated in this PR?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -170,13 +176,66 @@ class ( Show (ChainDepState   p)
...
--- protocols /always/ sticks with the current chain.
-preferCandidate :: ConsensusProtocol p
-                => proxy      p
-                -> SelectView p  -- ^ Tip of our chain
-                -> SelectView p  -- ^ Tip of the candidate
-                -> Bool
-preferCandidate _ ours cand = cand > ours
+-- See 'preferCandidate' for the primary documentation.
+--
+-- Additionally, we require a total order on this type, such that eg different
```
nfrisby, line 184: I think we need to reword and/or elaborate this comment.

- My basic concern: it says eg that the total order is used to prioritize candidates in the BlockFetch decision logic. That sure sounds like we're not restricting the VRF tiebreaker in the BlockFetch decision logic.
- I guess because we're not? ... Are there in-flight limits etc that could mean this unrestricted VRF determines which blocks are even fetched?
- Also I think it'd be helpful if this comment explained that the thing these situations have in common is that these decisions are not made in a context where "block arrival order" is not inherently available. (Initial selection and out-of-order blocks would require us to record timestamps in the VolDB. In BlockFetch decision logic, I suppose header arrival order sort of does apply (if you're the only better header, you'll be fetched), unless _multiple_ tied headers arrived between the two BlockFetch invocations----again unless we recorded timestamps.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -170,13 +176,66 @@ class ( Show (ChainDepState   p)
...
+-- candidate chains that are preferred over our current selection can be sorted
+-- for prioritization. For example, this is used in ChainSel during initial
+-- chain selection or when blocks arrive out of order (not the case when the
+-- node is caught up), or in the BlockFetch decision logic.
+class Ord a => ChainOrder a where
```
nfrisby, line 189: Maybe write `sv` instead of `a`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -135,7 +136,9 @@ data PBftSelectView = PBftSelectView {
       pbftSelectViewBlockNo :: BlockNo
     , pbftSelectViewIsEBB   :: IsEBB
     }
-  deriving (Show, Eq, Generic, NoThunks)
+  deriving stock (Show, Eq, Generic)
```
nfrisby, line 139: very minor: alphabetize?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -170,13 +176,66 @@ class ( Show (ChainDepState   p)
...
+
+-- | A @DerivingVia@ helper to implement 'preferCandidate' in terms of the 'Ord'
+-- instance.
+newtype SimpleChainOrder a = SimpleChainOrder a
+  deriving newtype (Eq, Ord)
+
+instance Ord a => ChainOrder (SimpleChainOrder a) where
+  type ChainOrderConfig (SimpleChainOrder a) = ()
+
+  preferCandidate _cfg ours cand = cand > ours
```
nfrisby, line 239: `ours < cand`, in order to match your nice Haddock?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
amesgen: > * My basic concern: it says eg that the total order is used to prioritize candidates in the BlockFetch decision logic. That sure sounds like we're not restricting the VRF tiebreaker in the BlockFetch decision logic.
> * I guess because we're not? ... Are there in-flight limits etc that could mean this unrestricted VRF determines which blocks are even fetched?

Yes, this could be the case, it is the same as today. Note that we still only download blocks corresponding to chains that are better than the current selection according to `preferCandidate`, so we won't download a chain that is better due to `Ord`, but not due to `preferCandidate`.

> Also I think it'd be helpful if this comment explained that the thing these situations have in common is that these decisions are not made in a context where "block arrival order" is not inherently available. (Initial selection and out-of-order blocks would require us to record timestamps in the VolDB. In BlockFetch decision logic, I suppose header arrival order sort of does apply (if you're the only better header, you'll be fetched), unless _multiple_ tied headers arrived between the two BlockFetch invocations----again unless we recorded timestamps.)

I like the absence of arrival times as a connecting reason for the applications of the `Ord` instance. I am a *bit* hesitant as I am not yet completely sure how the logic would/should look like even if we recorded arrival times there. Eg write `a ⊏ b` for `preferCandidate cfg a b`, and suppose that the BlockFetch decision logic just became aware of the new headers `a,b,c`, which were received in this order, each extending the current selection, with `sv(a) ⊏ sv(b)`, `sv(b) ⊏ sv(c)` and `sv(c) = sv(a)` (`sv` = `selectView`).

 - Downloading `block(a)` first seems weird as `b` is strictly preferrable.
 - Downloading `block(b)` first seems weird as `c` is strictly preferrable.
 - Downloading `block(c)` first seems weird as `a` is just as preferrable as `c`, but it arrived earlier.

Maybe I could just mention "absence of arrival times" as an "intuition", but stress that there are actually no plans to implement arrival times for these use cases?

---

Also FTR, for the particular case of BlockFetch, [this paper ("the Tse paper")](https://eprint.iacr.org/2021/1545.pdf) that we might want prioritize downloading blocks with higher slot numbers, not block numbers. The core issue remains though for initial chain selection and out-of-order arrival.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
jorisdral: 'twas my doing IIRC, and indeed it's because it's not a `Config` so it didn't seem to fit in the group

ouroboros-consensus-protocol/test/protocol-test/Test/Consensus/Protocol/Praos/SelectView.hs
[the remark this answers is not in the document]
nfrisby: 👍 

No, I'm not worried about pipelining.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
nfrisby: > Note that we still only download blocks corresponding to chains that are better than the current selection according to preferCandidate, so we won't download a chain that is better due to Ord, but not due to preferCandidate.

Good point 👍. So the "risk" is the same as in ChainSel.

> I am a bit hesitant as I am not yet completely sure how the logic would/should look like even if we recorded arrival times there

Great point! And thanks for the concrete example. I do think I see a clear path, though: treat it the exact same as ChainSel. In particular: fetch whichever of `a`, `b`, and `c` are better than _our current selection_ (regardless of how they compare to each other[^1]). Fetch them in their arrival order, since in-flight limits might not allow us to fetch them all immediately.

That resolves it, yeah?

[^1]: Although maybe I'm assuming they're all the same length? Don't download _shorter_ ones (yet). But I think that is veering towards a topo sort.

amesgen: So you are suggesting to completely ignore tiebreakers to prioritize between different chains that are preferred over our selection? I.e. sorting on `(BlockNo, Down ArrivalTime)`? That makes sense (the Tse rule would be another option); I now changed this paragraph by adding a sentence in the end:

```haskell
-- Additionally, we require a total order on this type, such that eg different
-- candidate chains that are preferred over our current selection can be sorted
-- for prioritization. For example, this is used in ChainSel during initial
-- chain selection or when blocks arrive out of order (not the case when the
-- node is caught up), or in the BlockFetch decision logic. Future work could
-- include also recording\/storing arrival information and use that instead
-- of\/in addition to the 'Ord' instance.
```

nfrisby: Yeah, I think that captures it well.

nfrisby (approved):
Great!

Issue #167, opened 2023-06-20
More exhaustive golden tests for Node-to-Client queries


amesgen:
Currently, we have golden tests for some, but not all queries (both the encoded query and result messages):
https://github.com/input-output-hk/ouroboros-consensus/blob/15d2e93a44cf9697d99634695e67c685540410da/ouroboros-consensus-cardano/src/shelley-testlib/Test/Consensus/Shelley/Examples.hs#L199-L217

The goal of this ticket is to extend this:

 - Add golden tests for all queries, which also includes adding necessary examples for ledger types to Ledger.
 - Make sure that the existing examples are sufficiently interesting; e.g. the example `PoolDistr` used for `GetStakeDistribution` has only one entry, which means that we couldn't notice that a recent Ledger update changed the encoding for maps with more than 23 elements.

PR #1067, opened 2024-04-18
Support for responding to a LocalStateQuery demand for a snapshot of big ledger peers

crocodile-dentist:
This change equips LocalStateQuery mini protocol app to respond to a request for a snapshot, taken from the current tip, of big ledger peers from a client, eg. cardano-cli. A new query tag ```GetBigLedgerPeerSnapshot``` is added to ```BlockQuery```. This data can be loaded later by a node, and it may be leveraged by the network layer to facilitate syncing up a node from a blank or stale state. Diffusion layer is expected to make heavy use of these relays when bootstrapping from scratch in Genesis consensus mode.


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -283,6 +286,14 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
     => Set (SL.Credential 'SL.Staking (EraCrypto era))
     -> BlockQuery (ShelleyBlock proto era) (VoteDelegatees (EraCrypto era))
 
+  -- | Obtain a snapshot of big ledger peers. CLI can serialize these,
+  -- and if made available to the node by topology configuration,
+  -- the diffusion layer can use these peers when syncing up from scratch
+  -- or stale ledger state - especially useful for Genesis mode
+  GetPeerSnapshot
```
amesgen, line 293: In case you do not already have golden tests upstream: You could add an example query + result here:
https://github.com/IntersectMBO/ouroboros-consensus/blob/568897ddfab3de0932ec16a477df46d8bb41ac56/ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs#L106-L123
in order to be sure that the serialization format is not accidentally changed in the future. (also see https://github.com/IntersectMBO/ouroboros-consensus/issues/167)

amesgen (commented):
Looks good, only minor comments.

FTR: This builds on top of https://github.com/IntersectMBO/ouroboros-network/pull/4850

ouroboros-consensus-diffusion/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ NOTE: version jumps from `0.13.0.0` to `0.15.0.0` because `0.14.0.0` was created
 - Updated `ouroboros-consensus-diffusion` to use `ouroboros-network-0.14.0.0`.
   `LowLevelRounNodeArgs` and `NodeKernel` records hold
   `PublicPeerSelectionState` variable.
+- Added `GetPeerSnapshot` block query
+
+### Non-Breaking
```
amesgen, line 17: Ah sorry for not saying this earlier: Can you revert the changes to this file, and instead add a [changelog fragment](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/docs/website/contents/for-developers/ReleaseProcess.md#adding-a-changelog-fragment)?

crocodile-dentist: Sure, no problem

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -283,6 +289,13 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
...
```
nfrisby, line 296: Perhaps `GetBigLedgerPeerSnapshot` or something that is similarly more specific.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -434,6 +447,11 @@ instance (ShelleyCompatible proto era, ProtoCrypto proto ~ crypto)
           SL.queryCommitteeMembersState coldCreds hotCreds statuses st
         GetFilteredVoteDelegatees stakeCreds ->
           getFilteredVoteDelegatees st stakeCreds
+        GetPeerSnapshot ->
```
nfrisby, line 450: This seems relatively inexpensive, but do you know roughly ballpark?

crocodile-dentist: I had not analyzed this - as far as calculating the big ledger peers from the whole, it is just some folds and sorting, with strictness applied where appropriate, so it should be well behaved for reasonable input. I'm not sure about the workload of fetching raw peers from the ledger. Should I take a closer overall look?

nfrisby: > I'm not sure about the workload of fetching raw peers from the ledger. Should I take a closer overall look?

Perhaps other members of the Networking Team (eg Marcin? Armando?) have already worried about that being "too expensive", since "getting ledger peers" is pretty primary to the Diffusion Layer?

---

We don't need to pin it down exactly. But it'd be regrettable if some SPO's system was performing poorly because they didn't know that they shouldn't be sending this query to often. So we should be able to document the "rough cost" or some such if that sort of thing is a risk.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -37,6 +37,9 @@ data ShelleyNodeToClientVersion =
 
     -- | New queries introduced: GetConstitutionHash, GetFilteredVoteDelegatees
   | ShelleyNodeToClientVersion8
+
+    -- | New queries introduced: GetPeerSnapshot
+  | ShelleyNodeToClientVersion9
```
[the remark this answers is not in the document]
crocodile-dentist: Right, I've changed it back to use version 8.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
crocodile-dentist: Done.

crocodile-dentist:
I've looked into the performance impact on cpu and memory use of this query on a fully synced node without any peers connected, and results are as follows. Measurements are taken every second for a duration of 5 minutes on an i7@4.8 GHz with 32 GB of RAM.

5 minute baseline plot when idling:
idle

5 minutes of snapshot requests, where to node is queried in random intervals every 0 to 1 seconds:
snapshot

There isn't any discernable impact on memory use - no spikes or growth could be determined, and any memory use triggered by the query is neglibigle. Some minor CPU use in excess of the idling load is visible on the second plot, but nothing noteworthy can be concluded from it as it appears to be within reasonable limits. 

The query itself executes in around 20ms on the node:
50-100 us to acquire ledger state in an atomic operation 
<3 ms to calculate stake pool relays by getPeers (time complexity~ O(n log n) - almost, there is a square term due to removal of duplicate relays for each stake pool, but I assume that the number of relays for each pool is small and can be treated as a constant to simplify things)
<15 ms to calculate which are the big ledger peers (time complexity is O(n log n) with most likely bigger constant terms as there are multiple passes over the stake pools, eg. one pass to calculate the total stake and another to re-compute the relative stake of each pool, and another to sort things)

The resulting snapshot file size is ~50kb.

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -19,9 +19,12 @@ module Test.Consensus.Shelley.Examples (
   , examplesShelley
   ) where
 
+import           Data.List.NonEmpty
```
amesgen, line 22: Minor: Unqualified import of `Data.List.NonEmpty` causes a lot of conflicts with the prelude (
```suggestion
import qualified Data.List.NonEmpty as NE
```

jasagredo:
For the `fs-sim` error I see in ci, update to the CHaP index-state at `2024-06-06T15:28:08Z`

crocodile-dentist:
> For the `fs-sim` error I see in ci, update to the CHaP index-state at `2024-06-06T15:28:08Z`

Yes, that's what I figured

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -119,6 +125,13 @@ fromShelleyLedgerExamples ShelleyLedgerExamples {
         , ("NonMyopicMemberRewards", SomeResult (GetNonMyopicMemberRewards Set.empty)
                                      (NonMyopicMemberRewards $ sreNonMyopicRewards))
         , ("GenesisConfig",          SomeResult GetGenesisConfig (compactGenesis sreShelleyGenesis))
+        , ("GetBigLedgerPeerSnapshot",
+           SomeResult GetBigLedgerPeerSnapshot
+                      (LedgerPeerSnapshot (At slotNo, [(AccPoolStake 0.9
```
amesgen, line 130: Instead of `At`, we use [`NotOrigin`](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus/Ouroboros-Consensus-Block-Abstract.html#t:WithOrigin):
```suggestion
                      (LedgerPeerSnapshot (NotOrigin slotNo, [(AccPoolStake 0.9
```
Then, the `Cardano.Slotting.Slot` import and the new `cardano-slotting` dependency can be removed.

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -19,9 +19,12 @@ module Test.Consensus.Shelley.Examples (
   , examplesShelley
   ) where
 
+import qualified Data.List.NonEmpty as NE
+
```
amesgen, line 23: Blank line

amesgen (approved):
Approving the code that is stacked on top of #1223 :+1: (last two commits)

Looks great, only two trivial comments

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
amesgen: The costs described in https://github.com/IntersectMBO/ouroboros-consensus/pull/1067#issuecomment-2141971151 seem completely fine and in line (or below) what other queries require.

amesgen:
Ah, and the new golden files have to be checked in (that's why CI is red ATM). You can generate them locally via `cabal build all` and then `cabal test all --test-options='-p Golden'`.

PR #1068, opened 2024-04-18
Disconnect peers with 0 density

facundominguez:
This affects ChainSync jumping, where genesis windows with no headers prevent jumps from happening.


facundominguez:
Please feel free to contribute different code layouts. The current presentation is minimizing the diff.

facundominguez:
The motivation to not mix idling with the upper bound computation, is that I want to know if all headers have been sent when `lb==ub`, which is true in particular when the peer sends a header in the last slot of the genesis window as well as after it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -268,7 +268,7 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
       -- If the density is 0, the peer should be disconnected. This affects
       -- ChainSync jumping, where genesis windows with no headers prevent jumps
       -- from happening.
-      if ub0 == 0 && not idling0 then pure peer0 else do
+      if ub0 == 0 && hasBlockAfter0 then pure peer0 else do
```
amesgen: From the commit message:

> This is not the same as not idling, since we want to disconnect when the peer is idling and has a header after sgen (if this is ever possible).

To double-check: In case we can forecast further than the Genesis window (which can happen ATM as the forecasting range starts at the intersection between the candidate fragment and the current selection), the peer might already have a block after the Genesis window *and* be idle at the same time. In that case, we want to disconnect when the density is exactly zero :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -242,7 +242,12 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
           -- that number of remaining slots.
           potentialSlots =
             if idling || hasBlockAfter then 0
-            else sgen - totalBlockCount
+            else unknownTrailingSlots
+
+          -- Number of trailing slots in the genesis window that could have
+          -- headers which haven't been sent yet
+          unknownTrailingSlots = unSlotNo $
+            firstSlotAfterGenesisWindow - succWithOrigin (AF.headSlot fragment)
```
amesgen, line 242: Maybe deserves a comment that this shall only be used when `not hasBlockAfter` is satisfied, as it will otherwise underflow.

Could use [`countSlots`](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus/Ouroboros-Consensus-HardFork-History-Util.html#v:countSlots) here, also has an assertion that this doesn't underflow.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -227,21 +227,13 @@ densityDisconnect (GenesisWindow sgen) (SecurityParam k) states candidateSuffixe
...
-          --
-          -- If the peer has more headers that it hasn't sent yet, each slot
-          -- between the latest header we know of and the end of the Genesis
-          -- window could contain a block, so the upper bound for the total
-          -- number of blocks in the window is the sum of the known blocks and
-          -- that number of remaining slots.
           potentialSlots =
-            if idling || hasBlockAfter then 0
+            if hasBlockAfter then 0
             else unknownTrailingSlots
```
amesgen, line 237: What is the motivation for this change?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: That's my thinking. This code is further refined in a later commit.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: I tried to explain it in [this comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/1068#issuecomment-2066650093).

In essence, it gives GDD the capacity to query if all headers have been received in the genesis window after the intersection. Before this change, we could only ask if a header has been received after the genesis window.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Since `fragment` is clipped to the genesis window, it is not possible for this to underflow.

facundominguez:
```suggestion
          unknownTrailingSlots = unSlotNo $
            -- Cannot underflow as @fragment@ is clipped to the genesis window
            firstSlotAfterGenesisWindow - succWithOrigin (AF.headSlot fragment)
```

amesgen: Right, thanks, this is using `fragment`, not `candidateSuffix` (the latter is *not* clipped). I think these could be renamed for more clarity.

amesgen:
> Please feel free to contribute different code layouts. The current presentation is minimizing the diff.

Thanks a lot for doing that, appreciated. I also agree that once we are happy with the logic, it makes sense to rewrite/-structure it, but that has no immediate urgency.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: > Eventually, the LoP wil guarantee that they will either send us a header beyond the Genesis window or send MsgAwaitReply.

I think so.

> such that changes to these notions are reflected in updated documentation.

The upper bound is only used in `densityDisconnect` at the moment.

I guess whatever documentation exists will need to be updated to explain the nuances here indeed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Renamed `fragment` to `clippedFragment` in 04841a9.

facundominguez:
Merged as d994eb34a.

PR #1069, opened 2024-04-18
Integrate `cardano-ledger` for `node-8.11.0`

Lucsanszky:
Related to https://github.com/IntersectMBO/cardano-node/issues/5790


Lucsanszky:
Obviously, this is far from done but I opted to open a draft early so everyone can track the progress of the integration.

One thing that is worth mentioning right away: [This](https://github.com/IntersectMBO/cardano-ledger/pull/4252) removes the `Crypto c` constraint from `exampleConwayGenesis` and uses `StandardCrypto` instead. I made some changes to the affected test to make it compile but I haven't ran it yet. It is now using `StandardCrypto` instead of `MockCryptoCompatByron`. Is this okay or should we figure out something else?

cc: @dnadales

amesgen:
@Lucsanszky  FYI: You can cherry-pick the first commit from #1031 when/if you are integrating https://github.com/IntersectMBO/ouroboros-network/pull/4846.

ouroboros-consensus-cardano/changelog.d/20240430_233346_Lucsanszky_integration_8_11.md
```diff
@@ -0,0 +1,20 @@
...
+  - `cardano-ledger-conway`: `^>=1.14`
+  - `cardano-ledger-core`: `^>=1.12`
+  - `cardano-ledger-mary`: `^>=1.6`
+  - `cardano-ledger-shelley`: `^>=1.11`
+
+### Breaking
+
+- Add `c ~ StandardCrypto` constraint to:
+  - `mkSimpleTestProtocolInfo`
+  - `mkTestProtocolInfo`
```
jasagredo, line 20: Unless I'm mistaken, these are changes in the test libraries. We don't record changes to those in the changelog.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -353,30 +327,13 @@ instance SupportsTwoPhaseValidation (BabbageEra c) where
...
-                            )
-                        )
-                    )
-                )
-            )
-        ) -> True
-    SL.ConwayUtxowFailure
-      ( Babbage.UtxoFailure
-          ( Babbage.AlonzoInBabbageUtxoPredFailure
-              ( Alonzo.UtxosFailure
```
jasagredo, line 373: Were these unified on `Conway.UtxoFailure`?

jasagredo: I think this is answered by this ticket, right? https://github.com/IntersectMBO/cardano-ledger/issues/4170

amesgen: Yeah, there is also this in the commit message
```
  * Update `SupportsTwoPhaseValidation` instance for `Conway`:
    `ConwayUtxowFailure` was fixed to get rid of the possibility
    of injecting `UtxoFailure` in two separate ways
    (as seen in the code removed by this commit)
```

ouroboros-consensus-cardano/changelog.d/20240430_233346_Lucsanszky_integration_8_11.md
Lucsanszky: Yes, you are right. Those are only related to test libraries, so if you don't track those then I can remove the `Breaking` part of the fragment.

jasagredo: Yes, please do so

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
```diff
@@ -8,166 +8,55 @@
...
     where
-      genesis :: SL.ShelleyGenesis (EraCrypto era)
-      genesis = shelleyLedgerGenesis (configLedger tlc)
 
-      updatesBefore, updatesAfter :: [ProtocolUpdate era]
-      updatesBefore = protocolUpdates genesis before
-      updatesAfter  = protocolUpdates genesis after
+      updatesBefore, updatesAfter :: ShelleyLedgerUpdate era
+      updatesBefore = pparamsUpdate before
+      updatesAfter  = pparamsUpdate after
```
amesgen, line 52: FTR: With the new code, we will log one `LedgerUpdate` in every epoch (when we first tick across the Shelley voting deadline, ie 6k/f before the end of the epoch), even if nothing changed. This is in contrast to the previous state of the code. This seems fine to me, at least for now.

amesgen (approved):

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Cardano.hs
```diff
@@ -73,7 +73,7 @@ import           Test.Util.TestEnv
 
 -- | Use 'MockCryptoCompatByron' so that bootstrap addresses and
 -- bootstrap witnesses are supported.
-type Crypto = MockCryptoCompatByron
+type Crypto = StandardCrypto
```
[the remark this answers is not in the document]
amesgen: The only difference between `MockCryptoCompatByron` and `StandardCrypto` is in the VRF and KES:

 - `MockKES` doesn't do any evolution.
 - `MockKES` and `MockVRF` are faster (see below).
 - The types involved in `MockKES` and `MockVRF` are potentially simpler to construct (though we mostly use `Arbitrary` instances that are defined for any `Crypto`).
 - `MockKES` and `MockVRF` yield more compact output in QuickCheck counterexamples.

A small benchmark with `hyperfine` yields that the Cardano ThreadNet is now ~50% slower: 

```bash
hyperfine -L crypto MockCryptoCompatByron,StandardCrypto './cardano-test-{crypto} -p "Cardano ThreadNet" --quickcheck-tests=2 --quickcheck-replay=0'
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `MockCryptoCompatByron` | 6.319 ± 1.329 | 5.350 | 9.020 | 1.00 |
| `StandardCrypto` | 9.547 ± 2.617 | 7.489 | 15.095 | 1.51 ± 0.52 |

Overall CI time of the entire `cardano-test` suite didn't regress by the same factor as it does lots of other things (see [Hydra](https://ci.iog.io/job/IntersectMBO-ouroboros-consensus/main/x86_64-linux.native.haskell96.checks.ouroboros-consensus-cardano.cardano-test/all)), it seems to be longer by 20s from 5min20s to 5min40s.

---

Also, I don't think that the usage of `StandardCrypto` in ledger is absolutely necessary; it could be generalized to any `Crypto c` where just the hashes/signatures are constrained, but the VRF/KES is left unconstrained.

Given that larger changes in the crypto parameterization are planned[^1] I am inclined to leave the code as is for now.

[^1]: See https://github.com/IntersectMBO/cardano-ledger/issues/4223, which would make all ledger definitions crypto-monomorphic, but removing KES/VRF from `Crypto` to sth like `HeaderCrypto`, which would solve the issue at hand (as `MockCryptoCompatByron` and `StandardCrypto` only differ in header crypto).

PR #1073, opened 2024-04-23
Factor out class `ShelleyBasedBlock` and type family `ShelleyBlockLedgerEra`

dnadales:
Pending change from https://github.com/IntersectMBO/ouroboros-consensus/pull/934#discussion_r1508794799

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilyDependencies     #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
 {-# LANGUAGE UndecidableSuperClasses    #-}
 module Ouroboros.Consensus.Shelley.Ledger.Block (
     GetHeader (..)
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
```diff
@@ -138,6 +140,18 @@ mkShelleyBlock raw = ShelleyBlock {
...
     }
 
+class
+  ( ShelleyBasedEra (ShelleyBlockLedgerEra blk)
+  , blk ~ ShelleyBlock (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
+  ) => ShelleyBasedBlock blk
+
+instance ( proto ~ BlockProtocol (ShelleyBlock proto era)
+         , ShelleyBasedEra era
+         ) => ShelleyBasedBlock (ShelleyBlock proto era)
```
amesgen, line 150: We could give this class "more power", ie make its superclass constraints stronger, eg to `ShelleyCompatible` instead of just `ShelleyBasedEra`:
```suggestion
class
  ( ShelleyCompatible (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
  , blk ~ ShelleyBlock (BlockProtocol blk) (ShelleyBlockLedgerEra blk)
  ) => ShelleyBasedBlock blk

instance ( proto ~ BlockProtocol (ShelleyBlock proto era)
         , ShelleyCompatible proto era
         ) => ShelleyBasedBlock (ShelleyBlock proto era)
```
This might come in handy in a future use case where we want to call a function in an SOP combinators that requires `ShelleyCompatible` and not just `ShelleyBasedEra`.

amesgen (approved):

jorisdral (approved):

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
jasagredo: I also proposed the same. I do wonder if this is not equivalent to requiring it in the constructors? as in:
```
data ShelleyBlock proto era where
  ShelleyBlock :: ShelleyCompatible proto era => 
    { ... } -> ShelleyBlock proto era
```

leaving that aside, it might be intereseting to provide instances for `Header (ShelleyBlock ...)` and `LedgerState (ShelleyBlock ...)`? I feel like this might simplify some aux classes we might have around.

jorisdral: I think one should prefer function constraints, not constraints on constructors. They are not entirely equivalent

Most of the places where I use constraints inside constructors is when dealing with existentials or `Dict`s

PR #1076, opened 2024-04-25
Fix honest shrinking by keeping the last point of adversary schedules

nbacquey:


facundominguez:
This fixes the shrinking for the test failure I encountered. But I could not advice if it affects other things without further familiarizing with the code. I'm tagging @Niols for review.

facundominguez:
Merged and squashed this to 38b866e.

PR #1078, opened 2024-04-25
Utilize new PParams prediction functionality for HFC

lehins:
# Description

this PR might resolve.

Also note that:


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
```diff
@@ -8,166 +8,52 @@
...
+      updatesAfter  = pparamsUpdate after
+
+pparamsUpdate ::
+       forall era proto. ShelleyBasedEra era
+    => LedgerState (ShelleyBlock proto era)
+    -> ShelleyLedgerUpdate era
+pparamsUpdate st =
+    let nes = shelleyLedgerState st
+     in ShelleyUpdatedPParams
+          (nes ^. SL.newEpochStateGovStateL . SL.futurePParamsGovStateG)
```
amesgen, line 58: If want to further minimize the API surface: It would suffice to just store the protocol version here.

amesgen (commented):
Great to see the hard fork detection code being moved to Ledger, looks very clean! :tada:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -137,9 +137,8 @@ shelleyTransition ::
 shelleyTransition ShelleyPartialLedgerConfig{..}
                   transitionMajorVersionRaw
                   state =
-      takeAny
-    . mapMaybe isTransition
-    . Shelley.Inspect.protocolUpdates genesis
+      isTransition
+    . Shelley.Inspect.pparamsUpdate
```
amesgen, line 141: `pparamsUpdate` calls `futurePParamsGovStateG`, which according to the documentation in https://github.com/IntersectMBO/cardano-ledger/pull/4253 has the following behavior:

> This getter is only reliable and efficient 2 stability before the end of the epoch. Depending on the era, if called earlier in the epoch, it will either produce unreliable results or getting those results will be somewhat costly.

`shelleyTransition` will be invoked on every forecast/tick (via `singleEraTransition`), so it will be called in particular in the first 4k/f of an epoch. If that is undesirable, it would be fine for us if `futurePParamsGovStateG` would just always immediately return `Nothing` in that case.

lehins: I think to properly solve this is to run
```haskell
  guard $ shelleyAfterVoting >= fromIntegral k
```
before we look into `pparamsUpdate`, right?

lehins: >  it would be fine for us if futurePParamsGovStateG would just always immediately return Nothing in that case.

This is quite a bit harder to do on the ledger side, unfortunately. I was hoping it is possible to not force this calculation before it is really needed

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
lehins: EpochNo is used in `shelleyTransition`, but my worry was that `guard $ updatesBefore /= updatesAfter` is affected by removal of this epochNo, so if they are guaranteed to be the same, then we can remove it from `ShelleyLedgerUpdate`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
lehins: Giving it a bit more though, I think I know a nice way to solve this in ledger. Thank you for the feedback

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
amesgen: Right, I was imprecise; I think it makes sense to keep `EpochNo` here; but the `Maybe (Core.PParams era)` could be changed to `Maybe ProtVer`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -153,11 +153,13 @@ shelleyTransition ShelleyPartialLedgerConfig{..}
 
     isTransition :: ShelleyLedgerUpdate era -> Maybe EpochNo
     isTransition (ShelleyUpdatedPParams maybePParams newPParamsEpochNo) = do
-         pp <- maybePParams
+         -- We need to run this check first in order to ensure that PParams value is not
+         -- forced unnecessarily. See documentation of `futurePParamsGovStateG` in ledger.
+         guard $ shelleyAfterVoting >= fromIntegral k
+         SL.SJust pp <- Just maybePParams
```
amesgen, line 159: FTR: We just discussed in a call that this is no longer necessary, `nextEpochUpdatedPParams` is always cheap (by returning `SNothing` when the proposals for the next epoch are not yet definitive)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs
lehins: So this suggestion would be fine if `inspectLedger` was only used for predicting protocol version. I am just really not sure that this is indeed the case. I would suspect that updates to other protocol parameters could be useful for inspection as well?

In other words if I thin this to only `ProtVer` then `inspecLedger` will return `Nothing` on any other parameter update, while previously it would return a `Just` value, due to some parameter change and this `guard $ updatesBefore /= updatesAfter`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
lehins: New and improved approach has been implemented in https://github.com/IntersectMBO/cardano-ledger/pull/4253

With this new approach we no longer need to worry about any of the conditions that we previously had to worry about in here:
* quorum of same votes on proposals have been reached
* are we in the correct part of the epoch (two stability windows before the end of the epoch or not)

This is all now taken care of by the ledger with the help of `nextEpochUpdatedPParams` that is cheap and safe to be called through any point of the epoch.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
lehins: Yep. This check is now removed.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -130,46 +130,20 @@ type ProtocolShelley = HardForkProtocol '[ ShelleyBlock (TPraos StandardCrypto)
...
-         Shelley.Inspect.UpdateState{..}    = protocolUpdateState
-
-    -- In principle there could be multiple proposals that all change the
-    -- major protocol version. In practice this can't happen because each
-    -- delegate can only vote for one proposal, but the types don't guarantee
-    -- this. We don't need to worry about this, and just pick any of them.
-    takeAny :: [a] -> Maybe a
-    takeAny = listToMaybe
+         guard $ SL.pvMajor protVer == transitionMajorVersion
+         return newPParamsEpochNo
```
amesgen, line 146: `guard $ shelleyAfterVoting >= fromIntegral k` now seems to be missing

lehins: I guess I misunderstood you. I thought you said it is no longer unnecessary.

amesgen: > I thought you said it is no longer unnecessary.

You probably mean "necessary"?

---

> I guess I misunderstood you.

Yeah, I could have made that more clear; when I commented in [this thread](https://github.com/IntersectMBO/ouroboros-consensus/pull/1078#discussion_r1587876479) on this code
```haskell
         -- We need to run this check first in order to ensure that PParams value is not
         -- forced unnecessarily. See documentation of `futurePParamsGovStateG` in ledger.
         guard $ shelleyAfterVoting >= fromIntegral k
```
I was just commenting on the fact that the comment does no longer apply, ie it is not relevant to run this check first to avoid forcing the PParams. The check is still necessary, but it is irrelevant if it runs before/after the `SL.pvMajor protVer == transitionMajorVersion` check. Does that make sense?

lehins: > You probably mean "necessary"?

yes, that is what I meant :grin: 

From the name `shelleyAfterVoting` it sounded to me like a mark that indicates the two stability windows point before the end of the epoch, cause that is the only voting I am aware of in Shelley. That's why I inferred you comment as this guard now being redundant. Which led me to believe that it could now be redundant.
However, after reading the haddock a couple of times I realize that it means a number of blocks that was created after that point of two stability windows before the end of the epoch. In order to ensure that last block that potentially contained some votes can't be rolled back.

Which totally makes sense to me why we don't want to remove that guard. Brought it back.

amesgen: Perfect, thanks :+1:

PR #1033, opened 2024-04-02
Genesis milestone 12 - Chain Sync Jumping

nbacquey:
The main contribution of this PR is the implementation of ChainSync Jumping (0e265c4), a mechanism to avoid overloading the honest peers in the network when one or more peers connect to them for syncing. The details of ChainSync Jumping are discussed in the [Jumping module](https://github.com/IntersectMBO/ouroboros-consensus/blob/78e16d0/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs).

A range of supporting commits are provided in addition.

## More tests

* The time limited leashing attack test is configured and enabled in 6c25310. This test is aimed at the GDD governor implemented in #1015.
* A test to show that a node can resume syncing after being disconnected for a while has been implemented in 5d3c200.
* There is a test checking that the GDD governor doesn't regret disconnecting nodes as more headers are known to it (dbbc3ab ). This was included in the previous milestone but we discovered later it wasn't checking as much as we intended.
* In da7a6db  and 3220e7d there is a test for ChainSync Jumping that ensures that a syncing node downloads headers from at most one peer when there is no disagreement between the peers.
* There are various fixes to tests and documentation in d6f6ddf, b0ed1c8, d70e604 , cee466f, and dbbc3ab.

## GDD refinement

The ability of the GDD governor to disconnect nodes has increased in dcb20f5. With respect to #1015, the GDD governor can now disconnect peers in the following scenarios:
1. when a peer has a chain that can't be denser than another chain served by another peer (before, we would disconnect the peer only if there was a denser chain); or
2. when a peer has a chain of density 0 and it claims to have more headers after the genesis window; or
3. when a peer sent a header for the last slot of the genesis window or later, and loses the density comparison to another peer (before, we would disconnect the peer only after it loses the density comparison to a peer that sent more than k headers in the genesis window).

These changes help ChainSync Jumping make progress while downloading headers from only two disagreeing peers (even if both are adversarial).

## Other changes

* A version of shrinking of honest schedules of messages has been implemented in 9794292.
* 9572256 contains a fix to the implementation of Limit on Patience.
* 915238e  implements a recording tracer that doesn't affect scheduling in IOSim.

PR #1080, opened 2024-04-26
ChainSync client: remove redundant intersection check

amesgen:
This step has caused confusions in the past.

See the commit description of the single commit.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -1099,7 +1099,6 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
     InternalEnv {
         drainThePipe
       , headerInFutureCheck
-      , intersectsWithCurrentChain
       , terminateAfterDrain
       , traceException
       } = intEnv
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1108,13 +1107,7 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
         recordHeaderArrival
       } = headerInFutureCheck
 
-    -- Request the next message (roll forward or backward), unless our chain
```
nfrisby, line 1111: What do you think about adding a comment on the MsgRollForward handler explaining why it's sufficient for it to be the only place that checks for the intersection?

amesgen: Added a comment to `rollForward` :+1:

facundominguez:
Not sure if it is relevant, but in #1033 (3164620) I found a trace were the removed code sets the candidate fragment to `G` from `G | 1-0`. [This](https://github.com/IntersectMBO/ouroboros-consensus/actions/runs/8853931627/job/24315856092?pr=1033) this is the failure. And I'm attaching the [trace](https://github.com/IntersectMBO/ouroboros-consensus/files/15136755/out.txt) where you can see that at tick 44, adversary 2 gets its candidate fragment set to G.

I'm not sure it is relevant because it could be a CSJ artifact.

amesgen:
> the removed code sets the candidate fragment to `G` from `G | 1-0`.

I don't think this can be the case; the code that is removed in this PR never changes the tip of the candidate; it only ever causes us to disconnect or to change the *anchor* of the candidate fragment.

facundominguez:
> I found a trace were the removed code sets the candidate fragment to `G` from `G | 1-0`

Today, with additional tracing, looks like this is not the case. In all likelihood, it is a false alarm.

nfrisby (approved):
Thanks for the simplification!

PR #1081, opened 2024-04-27
More CSJ fixes

facundominguez:


Niols (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -391,7 +391,8 @@ onRollForward context point = whenEnabled context () $
     Disengaged -> pure ()
     Jumper{} -> pure ()
     Dynamo lastJumpSlot
-      | pointSlot point >= ((jumpSize context +) <$> lastJumpSlot) -> do
+      | let jumpBoundaryPlus1 = jumpSize context + succWithOrigin lastJumpSlot
+      , succWithOrigin (pointSlot point) >= jumpBoundaryPlus1 -> do
```
Niols, line 395: This fixes the weird interaction of `+`, `<$>` and `WithOrigin`; nice!

However, when the new point is exactly jumpSize away from the previous jump point (ie. in the equality case here), we trigger a jump to the tip of the candidate fragment, even though the new point would be a “better” jump point.

For instance:
```
jumpSize = 3

gtBlockTree:                             
  slots:    0  1  2  3  4  5  6  7  8  9 
  trunk:  ──1─────2──3─────4──5─────6──7 
``` 

We could hope to jump from G to 2-2, then 4-5, then 6-8. Instead, 2-2 will be the first point to pass this condition (see below) and we will trigger a jump to 0-1. Then, 3-3 will be the next point passing the condition again, triggering us to jump to 2-2. etc.

```haskell
let jumpBoundaryPlus1 = jumpSize context + succWithOrigin lastJumpSlot -- = 3 + succWithOrigin G == 3
succWithOrigin (pointSlot point) >= jumpBoundaryPlus1 -- succWithOrigin 2 >= 3 == True
```

This is probably not a very big deal on big jumpSizes (eg. half main net's genesis window) but it is still counter-intuitive. Should we put greater than instead? Do you see any shortcomings to this solution?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1279,6 +1296,35 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
...
+                  -- rewindable because it was rewindable when the good point
+                  -- was discovered.
+                  fromMaybe (error "offerJump: cannot rewind history for a good point") mRewoundHistory
+                intersection =
+                  case AF.intersect ourFrag fragment of
+                    Just (po, _, _, _) -> castPoint $ AF.headPoint po
+                    -- ourFrag should intersect with the good fragment, or otherwise
+                    -- it would not intersect with it at the time the good
+                    -- fragment was discovered.
+                    Nothing -> error "offerJump: the fragment should have a valid intersection with the current selection"
```
facundominguez, line 1319: @amesgen, @niols, attention here.

I believe the error calls in this function are unreachable, but this is not trivial to see. Perhaps the comments are easier to grasp when assuming the dynamo doesn't change. But when the dynamo changes, the last good point is assumed to be the anchor of the dynamo's candidate fragment (i.e. it hasn't actually been sent before via `MsgFindIntersect`).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
facundominguez: > Do you see any shortcomings to this solution?

No shortcomings. We could try that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1215,29 +1219,42 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
 
     offerJump ::
         MkPipelineDecision
-     -> JumpInfo blk
+     -> Either (AnchoredFragment (Header blk)) (JumpInfo blk)
```
amesgen, line 1222: What about extracting this as a separate type, eg
```haskell
data Instruction blk
  = RunNormally
  | JumpInstruction !(JumpInstruction blk)

data JumpInstruction blk =
  | JumpTo !(JumpInfo blk)
  | JumpToGoodPoint !(AF.AnchoredFragment (Header blk))
```
and use `JumpInstruction` here?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: This seems fine to me when the dynamo does not change (couldn't we "freeze" it as soon as someone objected?), as we are in this scenario:

 - The dynamo offers a jump to all jumpers.
 - Someone objects (suppose just one jumper); we find the intersection with the dynamo.
 - Hence, when spinning up that objector, we want to set its `KnownIntersectionState` such that its `theirFrag objector` ends the computed intersection. So `theirFrag objector` will be a prefix of `theirFrag dynamo`, and as `theirFrag dynamo` intersects with `ourFrag dynamo`, so does `theirFrag objector`.

Minor: This code could use `AF.intersectionPoint` instead of `AF.intersect`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
facundominguez: Adopted.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
facundominguez: Adopted `intersectionPoint`. Iterated once more on the implementation to use a kis from the dynamo instead of working with the kis from the objector to set the intersection of the ChainSync server.

facundominguez:
I merged and squashed these commits in milestone-12.

PR #1083, opened 2024-05-03
Added `GetAccountState` query

Soupstraw:
# Description

This PR adds the `GetAccountState` query.

resolves IntersectMBO/cardano-ledger#4293

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
@@ -45,6 +45,7 @@ import qualified Cardano.Ledger.EpochBoundary as SL
 import           Cardano.Ledger.Keys (KeyHash, KeyRole (..))
 import qualified Cardano.Ledger.Shelley.API as SL
 import qualified Cardano.Ledger.Shelley.Core as LC
+import           Cardano.Ledger.Shelley.LedgerState (AccountState)
 import qualified Cardano.Ledger.Shelley.LedgerState as SL (RewardAccounts,
                      newEpochStateGovStateL)
 import qualified Cardano.Ledger.Shelley.PParams as SL (emptyPPPUpdates)
@@ -283,6 +284,9 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
     => Set (SL.Credential 'SL.Staking (EraCrypto era))
     -> BlockQuery (ShelleyBlock proto era) (VoteDelegatees (EraCrypto era))
 
+  GetAccountState
+    :: BlockQuery (ShelleyBlock proto era) AccountState
+
   -- WARNING: please add new queries to the end of the list and stick to this
   -- order in all other pattern matches on queries. This helps in particular
   -- with the en/decoders, as we want the CBOR tags to be ordered.
@@ -434,6 +438,8 @@ instance (ShelleyCompatible proto era, ProtoCrypto proto ~ crypto)
           SL.queryCommitteeMembersState coldCreds hotCreds statuses st
         GetFilteredVoteDelegatees stakeCreds ->
           getFilteredVoteDelegatees st stakeCreds
+        GetAccountState ->
+          SL.queryAccountState st
     where
       lcfg    = configLedger $ getExtLedgerCfg cfg
       globals = shelleyLedgerGlobals lcfg
@@ -585,6 +591,8 @@ instance SameDepIndex (BlockQuery (ShelleyBlock proto era)) where
     | otherwise
     = Nothing
   sameDepIndex GetFilteredVoteDelegatees {} _ = Nothing
+  sameDepIndex GetAccountState {} GetAccountState {} = Just Refl
+  sameDepIndex GetAccountState {} _ = Nothing
 
 deriving instance Eq   (BlockQuery (ShelleyBlock proto era) result)
 deriving instance Show (BlockQuery (ShelleyBlock proto era) result)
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -774,6 +790,8 @@ encodeShelleyQuery query = case query of
       CBOR.encodeListLen 4 <> CBOR.encodeWord8 27 <> toCBOR coldCreds <> toCBOR hotCreds <> LC.toEraCBOR @era statuses
     GetFilteredVoteDelegatees stakeCreds ->
       CBOR.encodeListLen 2 <> CBOR.encodeWord8 28 <> LC.toEraCBOR @era stakeCreds
+    GetAccountState nes ->
+      CBOR.encodeListLen 2 <> CBOR.encodeWord8 29 <> LC.toEraCBOR @era nes
 
 decodeShelleyQuery ::
```
jasagredo, line 796: I can't comment on unchanged lines, but you are missing a decoder here.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -297,6 +298,15 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
   -- longer used (because mainnet has hard-forked to a newer version), it can be
   -- removed.
 
+  GetAccountState 
+    :: ( CG.ConwayEraGov era
+       , LC.EraTxOut era
+       , Show (StashedAVVMAddresses era)
+       , Eq (StashedAVVMAddresses era)
+       )
+    => SL.NewEpochState era
```
jasagredo, line 307: Does this query expect a full NewEpochState? This seems wrong. There is an `st` parameter when answering the queries that is the ledgerstate at the requested point (tip or immutable tip) so this state shall not be an argument to the query. Otherwise we would be sending whole ledgerstates between clients and nodes.

jasagredo (changes requested):
(actually my review above should have been a "Request changes" review)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
Soupstraw: Thanks!

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -297,6 +298,14 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
   -- longer used (because mainnet has hard-forked to a newer version), it can be
   -- removed.
 
+  GetAccountState 
+    :: ( CG.ConwayEraGov era
+       , LC.EraTxOut era
+       , Show (StashedAVVMAddresses era)
+       , Eq (StashedAVVMAddresses era)
```
jasagredo, line 305: Do we need these constraints?

Soupstraw: All of them became redundant when I got rid of `NewEpochState`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -297,6 +298,9 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
   -- longer used (because mainnet has hard-forked to a newer version), it can be
   -- removed.
 
+  GetAccountState
```
jasagredo, line 301:
```suggestion
  GetAccountState
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -87,6 +87,7 @@ import           Ouroboros.Consensus.Shelley.Protocol.Abstract (ProtoCrypto)
 import           Ouroboros.Consensus.Util (ShowProxy (..))
 import           Ouroboros.Network.Block (Serialised (..), decodePoint,
                      encodePoint, mkSerialised)
+import Cardano.Ledger.Shelley.LedgerState (AccountState)
```
amesgen, line 90: You can run `./scripts/ci/run-stylish.sh` to fix the formatting CI failure.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -297,6 +298,9 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
   -- longer used (because mainnet has hard-forked to a newer version), it can be
   -- removed.
 
+  GetAccountState 
+    :: BlockQuery (ShelleyBlock proto era) AccountState
```
jasagredo, line 302: Don't we need `ConwayEraGov`?

amesgen, line 302: Can you move this above the comment, next to the other queries?

jasagredo: I checked the source in `cardano-ledger`. It seems we don't need it.

amesgen: Ah, but if we don't need it, then the `requireCG` below should be removed, right?

PR #1085, opened 2024-05-03
Recover chainsel idempotency

facundominguez:
This PR forces chain selection to pick selections that follow the LoE fragment. That is, either they extend the LoE fragment, or they are extended by the LoE fragment.

Additionally, we no longer trigger the GDD when the selection changes. This was used in the past as a convoluted mechanism to trigger chain selection repeteadly until no more changes were done to the selection.

Also we relax the amount of blocks that can be selected again. We can select up to `k` blocks after the youngest intersection of all candidate fragments. Before we would select only up to `k` blocks from the intersection of the LoE fragment and the current selection.

Lastly, we avoid triggering chain selection on every downloaded header. We do this by retriggering chain selection only when GDD changes the tip of the LoE fragment.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
@@ -39,7 +39,7 @@ module Ouroboros.Consensus.Genesis.Governor (
   , updateLoEFragUnconditional
   ) where
 
-import           Control.Monad (guard)
+import           Control.Monad (guard, when)
 import           Control.Tracer (Tracer, traceWith)
 import           Data.Containers.ListUtils (nubOrd)
 import           Data.Foldable (for_)
@@ -168,8 +168,12 @@ runGdd loEUpdater varLoEFrag chainDb getTrigger =
         curLedger <- ChainDB.getCurrentLedger chainDb
         pure (newTrigger, curChain, curLedger)
       loeFrag <- updateLoEFrag loEUpdater curChain curLedger
-      atomically $ writeTVar varLoEFrag loeFrag
-      triggerChainSelectionAsync chainDb
+      oldLoEFrag <- atomically $
+        readTVar varLoEFrag <* writeTVar varLoEFrag loeFrag
+      -- The chain selection only depends on the LoE tip, so there
+      -- is no point in retriggering it if the LoE tip hasn't changed.
+      when (AF.headHash oldLoEFrag /= AF.headHash loeFrag) $
+        triggerChainSelectionAsync chainDb
       spin newTrigger
 
 data DensityBounds blk =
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
```diff
@@ -168,8 +168,12 @@ runGdd loEUpdater varLoEFrag chainDb getTrigger =
         curLedger <- ChainDB.getCurrentLedger chainDb
         pure (newTrigger, curChain, curLedger)
       loeFrag <- updateLoEFrag loEUpdater curChain curLedger
-      atomically $ writeTVar varLoEFrag loeFrag
-      triggerChainSelectionAsync chainDb
+      oldLoEFrag <- atomically $
+        readTVar varLoEFrag <* writeTVar varLoEFrag loeFrag
```
amesgen, line 172:
```suggestion
      oldLoEFrag <- atomically $ swapTVar varLoEFrag loeFrag
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -704,6 +707,17 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+        -- Either frag extends loe or loe extends frag
+        followsLoEFrag :: LoE (AnchoredFragment (Header blk))
+                      -> AnchoredFragment (Header blk)
+                      -> Bool
+        followsLoEFrag LoEDisabled _ = True
+        followsLoEFrag (LoEEnabled loe) frag =
+          case AF.intersect frag loe of
+            Just (_, _, AF.Empty{}, _) -> True
+            Just (_, _, _, AF.Empty{}) -> True
+            _                          -> False
```
amesgen, line 719: This has complexity $O(\log(n_2)n_1)$ where $n_1,n_2$ are the lengths of `frag` and `loe`, resp. I think implementing this as
```suggestion
             AF.withinFragmentBounds (AF.headPoint loe) frag
          || AF.withinFragmentBounds (AF.headPoint frag) loe
```
should also work and have complexity $O(\log(\max(n_1,n_2))$.

Also: maybe extract this (or rather the core logic) to `Ouroboros.Consensus.Util.AnchoredFragment` as sth like `onSameChain` with preconditions and even tests?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -672,6 +674,7 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
 
         let chainDiffs = NE.nonEmpty
               $ map Diff.extend
+              $ filter (followsLoEFrag loeFrag)
```
amesgen, line 677: Don't we have to do sth similar in `switchToAFork`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs
facundominguez: Adopted. Thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
facundominguez: I think you are right. I amended that.

facundominguez:
Merged and squashed in 45953b5.

PR #1087, opened 2024-05-06
Accept arbitrary number of honest peers in `Peers`

Niols:
And follow the change everywhere. For now, “following the change” involved adding a lot of `Map.! 1` to get the honest peer. 

Follow-up work:

- [ ] Clean up `Map.! 1` and replace by properties that support several honest peers.
- [ ] Rewrite the CSJ test to rely on several honest peers.
- [ ] Unify shrinkers now that we have more semantics of what peers are.
- [ ] Generalise schedule generators to generate potentially several honest peers.


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
Niols, line 1: I don't think `GenChains` and the `GenesisTest` type is the right place to put this as it has nothing to do with the chains or the world in general. However, we have more prioritary things for this milestone so let's keep it at that.

Niols:
@facundominguez or @amesgen> Want to take a look at this? Anything wrong? And do you feel it checks the two stretch goals:
- Write a test that checks that CSJ syncs up by downloading from at most one honest peer at a time in any path, when CSJ is finished.
- Testing CSJ with multiple honest nodes the non-hacky way.

facundominguez (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -460,6 +460,8 @@ data GenesisTest blk schedule = GenesisTest
     gtLoPBucketParams    :: LoPBucketParams,
     gtCSJParams          :: CSJParams,
     gtSlotLength         :: SlotLength,
+    -- | The number of extra honest peers we want in the test.
+    -- It is stored here for convenience, and because it may affect schedule and block tree generation.
```
facundominguez, line 464:
```suggestion
    -- It is stored here for convenience, and because it may affect schedule and block tree generation.
    --
    -- The will be one adversarial peer per alternative branch, and
    -- `1 + gtExtraHonestPeers` honest peers.
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/CSJ.hs
```diff
@@ -22,55 +20,67 @@ import           Test.Consensus.PeerSimulator.Run (SchedulerConfig (..),
...
-        numOthers <- choose (1, 3)
-        otherHonests <- if synchronized
-          then pure $ replicate numOthers honest
-          else replicateM numOthers (genHonestSchedule gt)
-        pure $ gt $> mkPeers honest otherHonests
+    ( if synchronized
+        then genChainsWithExtraHonestPeers (choose (2, 4)) (if happy then pure 0 else choose (2, 4))
+          `enrichedWith` genUniformSchedulePoints
+        else genChains (if happy then pure 0 else choose (2, 4))
+          `enrichedWith` genDuplicatedHonestSchedule
```
facundominguez, line 83: It looks to me like the if condition is inverted here. Synchronized peers should use the same honest schedule according to the documentation.

PR #1095, opened 2024-05-09
Ensure that transaction tokens are unique in the mempool benchmarks 

jorisdral:


jorisdral:
We might want to switch to `criterion` instead of `tasty-bench`. `criterion` has `perRunEnv`, which is arguably more suited than `withResource`, and then mental/manual subtraction of measurements won't be required. There is the question whether `criterion` is compatible with [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark). I found [this PR](https://github.com/benchmark-action/github-action-benchmark/pull/130), but it hasn't been updated in a while, but it leads me to believe that the CSV/JSON output of `criterion` can be mapped to the format that `github-action-benchmark` requires. It would mean we'd have to write the format conversion stuff that we currently have for `tasty-bench` for `criterion` as well

jasagredo (approved):

Issue #887, opened 2024-01-16
Extend preflight guide


amesgen:
#879 added a preflight guide, with two initial topics. This epic collects further potential steps in that guide.

 1. Use basic `cardano-cli` commands, and do sth simple with the result (this maybe should be split across several steps)
     - get current tip
     - use `cardano-cli ping` to see/explain network versions and handshaking
     - get protocol parameters, explain values relevant for Consensus, in particular the on-chain protocol version (intra-era HFs, obsolete node check)
     - get stake distribution, explain decentralization parameter $k$ for why so many pools have roughly the same stake
     - get protocol state, explain nonces
 2. Explore some Cardano-related resources/website
     - Explain all(?) rows/columns of https://github.com/cardano-foundation/CIPs/blob/master/CIP-0059/feature-table.md
     - Explore https://pooltool.io/, especially https://pooltool.io/realtime (slot battles, height battles)
     - Explore some Cardano explorer page, and look at properties of blocks relevant for Consensus
 4. First exploration of leadership probability stuff
     - Introduce $\varphi(\alpha) = 1 - {(1-f)}^\alpha$ and basic properties
     - Some calculations around multi leader slots
     - Some applications of (negative) binomial distributions to get the probability of a specific number of blocks in a range of slots etc.
     - Concrete calculations with the stake distribution acquired via `cardano-cli` above

PR #1096, opened 2024-05-10
PreflightGuide: new section on the leader schedule

amesgen:
Partially addresses #887


amesgen:
[Rendered](https://github.com/IntersectMBO/ouroboros-consensus/blob/amesgen/leader-schedule-stats-docs/docs/website/contents/for-developers/PreflightGuide.md#dynamics-of-the-leader-schedule)

docs/website/contents/for-developers/PreflightGuide.md
```diff
@@ -128,6 +128,124 @@ Use a ledger snapshot of your local node (make sure to add the `_db-analyser` su
...
+> which yields 0.0185%, or the probability that it takes more than one minute until a slot is active
+> ```python
+> stats.geom(1/20).sf(60)
+> ```
+> which yields 4.6%.
+> Calculate some other probabilities that seem interesting to you, maybe inspired by some of the empirical data gathered via db-analyser.
+
+### Multi leader slots
+
+The probability that a pool with relative stake $\sigma$ is elected in any specific slot is given by $ \phi_f(\sigma) = \phi(\sigma) = 1-{(1-f)}^\sigma $.
```
jasagredo, line 188: This formula is not rendered
```suggestion
The probability that a pool with relative stake $\sigma$ is elected in any specific slot is given by $\phi_f(\sigma) = \phi(\sigma) = 1-{(1-f)}^\sigma$.
```

docs/website/contents/for-developers/PreflightGuide.md
```diff
@@ -128,6 +128,124 @@ Use a ledger snapshot of your local node (make sure to add the `_db-analyser` su
...
+The discussion above completely ignored *grinding*.
+In the context of the leader schedule, this refers to an adversarial influence on the nonce that is used as a seed for all of the VRF computations for slot election within a specific epoch.
+The exact mechanism is out-of-scope for this section, but it should be said that such an attack requires both a considerable amount of stake and computational resources, and it allows an adversary to choose the best out of multiple epoch nonces.
+An extremely conservative upper bound for how many nonces an attacker can choose from is $10^{20}$ (based on the computational resources required to even compute all of these nonces).
+The mainnet parameters of Cardano (in particular the security parameter) were chosen with resistance against grinding in mind.
+
+Concretely, an attacker with $\alpha$ stake that can choose out of $N$ epoch nonces can ensure that it leads any specific slot with probability $1-{(1-\phi(\alpha))}^N$, which is a huge advantage compared to $\phi(\alpha)$.
+For example, an attacker with $N=10^{20}$ with even a very small amount of stake has a chance of close to 100% to cause themselves to be elected in any specific slot.
```
jasagredo, line 236: I understand this as "An attacker can choose (with close to 100% probability) a slot in which they want to be elected" but this doesn't mean that they can choose with 100% probability to be elected in say 5 specific slots in the epoch right? The 100% applies if a single slot is targeted no?

jasagredo (commented):
Awesome write-up, 10/10 :ok_hand:

docs/website/contents/for-developers/PreflightGuide.md
amesgen: It probably also still applies to 5 slots, but not to large numbers of slots, the next paragraph is also supposed to point this out (slightly modified in the latest commit)

> Usually, an attacker wants to maximize their total number of active slots within a larger number of slots, like in an entire epoch. For this, the impact of grinding is smaller (but still large), as every epoch nonce affects the entire leader schedule of the epoch, and the attacker can not choose different epoch nonces for different slots.

Does that make sense, or do we want to make this more explicit?

jasagredo: I think just modifying this word would be enough:
```suggestion
For example, an attacker with $N=10^{20}$ with even a very small amount of stake has a chance of close to 100% to cause themselves to be elected in a specific slot.
```

```suggestion
For example, an attacker with $N=10^{20}$ with even a very small amount of stake has a chance of close to 100% to cause themselves to be elected in any single specific slot.
```

jasagredo: But also feel free to ignore this comment :smile:

jasagredo (approved):

PR #1099, opened 2024-05-13
Refactor tests to sanitize shrinking usage

nbacquey:


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LoE.hs
```diff
@@ -46,6 +46,9 @@ tests =
 -- NOTE: Same as 'LoP.prop_delayAttack' with timeouts instead of LoP.
 prop_adversaryHitsTimeouts :: Bool -> Property
 prop_adversaryHitsTimeouts timeoutsEnabled =
+  -- Here we can't shrink because we exploit the properties of the point schedule to wait
+  -- at the end of the test for the adversaries to get disconnected, by adding an extra point.
+  -- If this point gets removed by the shrinker, we lose that property and the test becomes useless.
```
facundominguez, line 51: I thought the shrinking was modified to preserve the duration of the schedule. Am I missing something here?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking.hs
```diff
@@ -84,15 +84,25 @@ shrinkAdversarialPeers shrink Peers {honestPeers, adversarialPeers} =
 -- We check that this operation does not changes the final state of the honest peer.
 -- NOTE: This operation makes the honest peer to end its schedule sooner, which *may*
 -- trigger disconnections when the timeout for MsgAwaitReply is reached. In those cases,
--- it is probably more pertinent to disable this timeout in tests than disabling shrinking.
+-- it is probably more pertinent to disable this timeout in tests than to disable shrinking.
```
facundominguez, line 87: Probably this documentation should say if ticks are removed from the end, and if so explain why the schedule isn't invalidated. After all, the honest peer may then not send headers or blocks that it advertised to have.

facundominguez (approved):
LGTM. Just one comment.

PR #1103, opened 2024-05-15
Re-remove ShowProxy SlotNo

jasagredo:
Reverses https://github.com/IntersectMBO/ouroboros-consensus/pull/1100


amesgen (approved):
If we need to do another release for Node 8.11, we can do it from a release branch :+1:

PR #1112, opened 2024-05-21
Disable BlockFetch timeouts in stalling leashing attack

nbacquey:
It makes no sense disabling ChainSync timeouts and keeping BlockFetch ones. Also, it caused some aberrant behaviour in some instances of the test.


facundominguez:
Leashing tests are supposed to run with timeouts enabled. The adversarial schedules are perturbed so that block fetch timeouts might be necessary if adversaries don't send their blocks.

The idle timeout and the mustReply timeouts are disabled because they are not interesting (they don't prevent any leashing).

PR #1116, opened 2024-05-22
Use simple diffs except in `DiffSeq`

jasagredo:
Closes #1114

PR #1117, opened 2024-05-22
Tests for `Diff`

jorisdral:
# Description

this PR might resolve.

Also note that:


jasagredo:
Closed in favor of #1116

PR #1120, opened 2024-05-28
Refactor `PointSchedule` to add an optional extra delay at the end

nbacquey:
TODO:
- [x] Consider introducing a new type for the extra delay in `peersStates` instead of duplicating the last `Peer (NodeState blk)`

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
@@ -64,6 +64,8 @@ data TraceSchedulerEvent blk
     TraceBeginningOfTime
   | -- | Right after running the last tick of the schedule.
     TraceEndOfTime
+  | -- | An extra optional delay to keep the simulation running
+    TraceExtraDelay DiffTime
   | -- | When beginning a new tick. Contains the tick number (counting from
     -- @0@), the duration of the tick, the states, the current chain, the
     -- candidate fragment, and the jumping states.
@@ -195,6 +197,13 @@ traceSchedulerEventTestBlockWith setTickTime tracer0 _tracer = \case
       traceLinesWith tracer0
         [ "╶──────────────────────────────────────────────────────────────────────────────╴",
           "Finished running point schedule"
+        ]
+    TraceExtraDelay delay -> do
+      time <- getMonotonicTime
+      traceLinesWith tracer0
+        [ "┌──────────────────────────────────────────────────────────────────────────────┐",
+          "└─ " ++ prettyTime time,
+          "Waiting an extra delay to keep the simulation running for: " ++ prettyTime (Time delay)
         ]
     TraceNewTick number duration (Peer pid state) currentChain mCandidateFrag jumpingStates -> do
       time <- getMonotonicTime
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
@@ -37,6 +37,7 @@ module Test.Consensus.PointSchedule (
   , ensureScheduleDuration
   , genesisNodeState
   , longRangeAttack
+  , mkPointSchedule
   , peerSchedulesBlocks
   , peerStates
   , peersStates
@@ -52,15 +53,13 @@ import           Control.Monad (replicateM)
 import           Control.Monad.Class.MonadTime.SI (Time (Time), addTime,
                      diffTime)
 import           Control.Monad.ST (ST)
-import           Data.Foldable (toList)
 import           Data.Functor (($>))
 import           Data.List (mapAccumL, partition, scanl')
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (catMaybes, fromMaybe, mapMaybe)
 import           Data.Time (DiffTime)
 import           Data.Word (Word64)
-import           Ouroboros.Consensus.Block.Abstract (WithOrigin (..),
-                     withOriginToMaybe)
+import           Ouroboros.Consensus.Block.Abstract (withOriginToMaybe)
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
                      (GenesisWindow (..))
 import           Ouroboros.Consensus.Network.NodeToNode (ChainSyncTimeout (..))
@@ -99,8 +98,9 @@ prettyPointSchedule ::
   PointSchedule blk ->
   [String]
 prettyPointSchedule peers =
-  [ "honest peers: " ++ show (Map.size (honestPeers $ unPointSchedule peers))
-  , "adversaries: " ++ show (Map.size (adversarialPeers $ unPointSchedule peers))
+  [ "honest peers: " ++ show (Map.size (honestPeers $ psSchedule peers))
+  , "adversaries: " ++ show (Map.size (adversarialPeers $ psSchedule peers))
+  , "minimal duration: " ++ show (psMinEndTime peers)
   ] ++
   zipWith3
     (\number time peerState ->
@@ -157,7 +157,8 @@ peerStates Peer {name, value = schedulePoints} =
 --
 -- The resulting schedule contains all the peers. Items are sorted by time.
 peersStates :: PointSchedule blk -> [(Time, Peer (NodeState blk))]
-peersStates peers = foldr (mergeOn fst) [] (peerStates <$> toList (peersList $ unPointSchedule peers))
+peersStates PointSchedule{psSchedule} =
+  foldr (mergeOn fst) [] (peerStates <$> peersList psSchedule)
 
 -- | Same as 'peersStates' but returns the duration of a state instead of the
 -- absolute time at which it starts holding.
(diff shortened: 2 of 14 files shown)


facundominguez:
I rebased the PR on top of the latest `genesis/milestone-13`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking/Tests.hs
```diff
@@ -77,20 +78,20 @@ doesNotChangeFinalState original shrunk =
...
+    (\sch@PointSchedule{psSchedule} ->
       conjoin $ map
       (\shrunk ->
           counterexample
           (  "Original schedule:\n"
-          ++ unlines (map ("    " ++) $ prettyPointSchedule schedule)
+          ++ unlines (map ("    " ++) $ prettyPointSchedule sch)
           ++ "\nShrunk schedule:\n"
-          ++ unlines (map ("    " ++) $ prettyPointSchedule shrunk)
+          ++ unlines (map ("    " ++) $ prettyPointSchedule $ mkPointSchedule shrunk)
```
facundominguez, line 92: Should this print a peer schedule instead of a point schedule?
I think it can be confusing to the reader if a dummy time duration is filled by `mkPointSchedule` just for the sake of printing.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking.hs
```diff
@@ -37,20 +37,30 @@ shrinkPeerSchedules ::
...
-          <&> \shrunkSchedule -> genesisTest {gtSchedule = shrunkSchedule}
+          psSchedule
+        -- No need to update the tree here, shrinking the honest peers never discards blocks
+        <&> \shrunkSchedule -> genesisTest
+          { gtSchedule = PointSchedule
+            { psSchedule = shrunkSchedule
+            , psMinDuration = simulationDuration
+            }
+          }
    in shrunkAdversarialPeers ++ shrunkHonestPeers
```
facundominguez, line 64: Is the comment of this function outdated? It is saying
`This does not affect the honest peer`, which is suggesting that there is only one honest peer, and that it isn't shrunk.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking.hs
```diff
@@ -60,10 +70,27 @@ shrinkByRemovingAdversaries ::
...
+      }
+
+duration :: PointSchedule blk -> Time
+duration PointSchedule {psSchedule, psMinDuration} = foldr
+  (\points d -> case points of
+    [] -> d
+    _  -> max d $ fst $ last points
+  )
+  psMinDuration
+  psSchedule
```
facundominguez, line 93: Perhaps this is the same as
```Haskell
maximum $ psMinDuration : [ t | sch <- psSchedule, (t, _) <- take 1 (reverse sch) ]
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking.hs
```diff
@@ -60,10 +70,27 @@ shrinkByRemovingAdversaries ::
   GenesisTestFull TestBlock ->
```
facundominguez, line 70: The comment of this function also talks of a single honest peer.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -158,7 +157,12 @@ peerStates Peer {name, value = schedulePoints} =
 --
 -- The resulting schedule contains all the peers. Items are sorted by time.
 peersStates :: PointSchedule blk -> [(Time, Peer (NodeState blk))]
-peersStates peers = foldr (mergeOn fst) [] (peerStates <$> toList (peersList $ psSchedule peers))
+peersStates PointSchedule{psSchedule, psMinDuration} =
+  case foldr (mergeOn fst) [] (peerStates <$> peersList psSchedule) of
+    [] -> []
+    peers -> case last peers of
+      (t, p) | t < psMinDuration -> peers ++ [(psMinDuration, p)]
+      _                          -> peers
```
facundominguez: I find this rendering easier to follow:
```
    peers ->
      let (t, p) = last peers
       in if t < psMinDuration then
            peers ++ [(psMinDuration, p)]
          else
            peers
```
but feel free to disregard it.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
```diff
@@ -74,6 +74,14 @@ data TraceSchedulerEvent blk
       (AnchoredFragment (Header blk))
       (Maybe (AnchoredFragment (Header blk)))
       [(PeerId, ChainSyncJumpingState m blk)]
+  | -- | When beginning a tick where the state does not change.
+    -- Contains the tick number (counting from @0@), the duration
+    -- of the tick, the candidate fragment, and the jumping states.
+    forall m. TraceEmptyTick
```
facundominguez, line 80: Other names: `TraceUneventfulTick`, `TraceBoringTick`, `TraceChangelessTick`

Is a tick a bucket to be filled with some substance? What does it mean, otherwise, that a tick is empty?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -156,17 +168,17 @@ peerStates Peer {name, value = schedulePoints} =
...
-peersStates :: PointSchedule blk -> [(Time, Peer (NodeState blk))]
+peersStates :: PointSchedule blk -> [(Time, PeerStateChange blk)]
 peersStates PointSchedule{psSchedule, psMinDuration} =
   case foldr (mergeOn fst) [] (peerStates <$> peersList psSchedule) of
-    [] -> []
+    [] -> if psMinDuration > Time 0 then [(psMinDuration, NoChange)] else []
     peers -> case last peers of
-      (t, p) | t < psMinDuration -> peers ++ [(psMinDuration, p)]
+      (t, _) | t < psMinDuration -> peers ++ [(psMinDuration, NoChange)]
       _                          -> peers
```
facundominguez, line 177: This is the only place where `NoChange` is constructed. Is it necessary to introduce a new kind of tick for this?

`runSchedule` could just call `threadDelay` once more at the end if needed, which would be a more contained change.

facundominguez (approved):
LGTM. I left some comments.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
facundominguez: hm, and then perhaps the printing of the point schedule would need an extra row to remind the reader of the minimum duration.

facundominguez:
I tested the PR with
```
git cherry-pick 63980f0df68
cabal run ouroboros-consensus-diffusion:consensus-test -- -p '/uniform/&&/stalling leashing attack/' --quickcheck-replay="(SMGen 16222173280194181131 12147767657514878291,6)"
```
This introduces a bug in change selection that used to deliver a shrunk schedule that didn't allow to disconnect adversarial peers because it was too short. Now it provides a shrink that shows that the honest peer is disconnected, which in my opinion is much better :+1:

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -555,4 +570,4 @@ ensureScheduleDuration gt ps =
            , busyTimeout bft
            , streamingTimeout bft
            ])
-    peerCount = length (peersList $ unPointSchedule ps)
+    peerCount = length (peersList psSchedule)
```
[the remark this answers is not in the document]
Niols: (We could use that time to add a documentation comment to `ensureScheduleDuration`.)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule.hs
```diff
@@ -157,7 +157,18 @@ peerStates Peer {name, value = schedulePoints} =
...
+    [] -> if psMinDuration > Time 0
+        then case peersList psSchedule of
+            []  -> []
+            p:_ -> [(psMinDuration, (const genesisNodeState) <$> p)]
+        else []
+    peers ->
+      let (t, p) = last peers
+      in if t < psMinDuration
+          then peers ++ [(psMinDuration, p)]
+          else peers
```
Niols, line 171: I thought the whole point was to get rid of these weird additional ticks, so I would have expected to get rid of this function, or at least to not have these lines 168--171. Is this really unavoidable?

Niols: I now see your comment:

> Dear reviewers, what's your opinion on that? I think the previous option was a better compromise: duplicate the last point "under the hood" so that the experiment lasts for the expected duration.

hinting at this. I do think this is a better compromise than the other one; I just wish we could do a bit better!

PR #1119, opened 2024-05-27
Add LoE to the ChainDB QSM test

Niols:
Closes #541

This PR adds a new instruction to the ChainDB q-s-m test, which updates the Limit on Eagerness (LoE) fragment and retriggers chain selection. This requires a model implementation of the LoE.

As a preparatory change, we add a way to trigger chain selection synchronously (ie such that we can wait for it to finish), which is only used in tests.

PR #1125, opened 2024-05-29
Small tweaks to the LoE

Niols:
This PR is a cherry pick of the commits of https://github.com/IntersectMBO/ouroboros-consensus/pull/1119 that make changes to the LoE:

1. We relax the bound in `computeLoEMaxExtra` in d32239b7abf275c330d240a1e3ac881e4ff99605. Previously, this function could never return `Just 0`. This made the LoE stricter than necessary. This is not a big problem but it does not match the modelled behaviour.

2. Instead of rejecting candidates that fork off the LoE fragment, we now trim them in 3957238ee3e34c86739d1f576316b70462ce6beb. This is again just relaxing the LoE a tiny bit.

3. We change the helper `followsLoEFrag` in c6b1d224fca5389cc328ca4d5ff39bc7b23d6500. It previously had a pre-condition that the anchor of the candidate had to be somewhere on the LoE fragment, but this precondition was regularly violated. The new version does not have this shortcoming.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
@@ -35,11 +35,10 @@ import           Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NE
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (isJust, isNothing)
+import           Data.Maybe (fromJust, isJust, isNothing)
 import           Data.Maybe.Strict (StrictMaybe (..), strictMaybeToMaybe)
 import           Data.Set (Set)
 import qualified Data.Set as Set
-import           Data.Word (Word64)
 import           GHC.Stack (HasCallStack)
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
@@ -176,7 +175,9 @@ initialChainSelection immutableDB volatileDB lgrDB tracer cfg varInvalid
         suffixesAfterI :: [NonEmpty (HeaderHash blk)]
         suffixesAfterI = Paths.maximalCandidates succsOf limit (AF.anchorToPoint i)
           where
-            limit = k <$ loE
+            limit = case loE of
+              LoEDisabled   -> Nothing
+              LoEEnabled () -> Just k
 
         constructChain ::
              NonEmpty (HeaderHash blk)
@@ -579,25 +580,17 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
         return tipPoint
 
       -- The block fits onto the end of our current chain
-      | pointHash tipPoint == headerPrevHash hdr
-        -- TODO could be optimized if necessary/easy enough
-      , let newBlockFrag = curChain AF.:> hdr
-      , Just maxExtra <- computeLoEMaxExtra loeFrag newBlockFrag -> do
+      | pointHash tipPoint == headerPrevHash hdr -> do
         -- ### Add to current chain
         traceWith addBlockTracer (TryAddToCurrentChain p)
-        addToCurrentChain succsOf' curChainAndLedger loeFrag maxExtra
+        addToCurrentChain succsOf' curChainAndLedger loeFrag
 
       -- The block is reachable from the current selection
       -- and it doesn't fit after the current selection
-      | Just diff <- Paths.isReachable lookupBlockInfo' curChain p
-        -- TODO could be optimized if necessary/easy enough
-      , let curChain' =
-              AF.mapAnchoredFragment (castHeaderFields . getHeaderFields) curChain
-      , Just newBlockFrag <- Diff.apply curChain' diff
-      , Just maxExtra <- computeLoEMaxExtra loeFrag newBlockFrag -> do
+      | Just diff <- Paths.isReachable lookupBlockInfo' curChain p -> do
         -- ### Switch to a fork
         traceWith addBlockTracer (TrySwitchToAFork p diff)
-        switchToAFork succsOf' lookupBlockInfo' curChainAndLedger loeFrag maxExtra diff
+        switchToAFork succsOf' lookupBlockInfo' curChainAndLedger loeFrag diff
 
       -- We cannot reach the block from the current selection
       | otherwise -> do
@@ -651,12 +644,10 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
          -- ^ The current chain and ledger
       -> LoE (AnchoredFragment (Header blk))
          -- ^ LoE fragment
-      -> LoE Word64
-         -- ^ How many extra blocks to select after @b@ at most.
       -> m (Point blk)
-    addToCurrentChain succsOf curChainAndLedger loeFrag maxExtra = do
+    addToCurrentChain succsOf curChainAndLedger loeFrag = do
         -- Extensions of @B@ that do not exceed the LoE
-        let suffixesAfterB = Paths.maximalCandidates succsOf maxExtra (realPointToPoint p)
+        let suffixesAfterB = Paths.maximalCandidates succsOf Nothing (realPointToPoint p)
 
         -- Fragments that are anchored at @curHead@, i.e. suffixes of the
         -- current chain.
@@ -675,10 +666,10 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
               return $ AF.fromOldestFirst curHead (hdr : hdrs)
 
         let chainDiffs = NE.nonEmpty
-              $ map Diff.extend
-              $ filter (followsLoEFrag loeFrag)
-              $ NE.filter (preferAnchoredCandidate (bcfg chainSelEnv) curChain)
-                candidates
+              $ filter (preferAnchoredCandidate (bcfg chainSelEnv) curChain . Diff.getSuffix)
+              $ fmap (trimToLoE loeFrag curChainAndLedger)
+              $ fmap Diff.extend
+              $ NE.toList candidates
         -- All candidates are longer than the current chain, so they will be
         -- preferred over it, /unless/ the block we just added is an EBB,
         -- which has the same 'BlockNo' as the block before it, so when
@@ -709,16 +700,45 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
         curTip      = castPoint $ AF.headPoint curChain
         curHead     = AF.headAnchor curChain
 
-    -- Either frag extends loe or loe extends frag
+    -- | Trim the given candidate fragment to respect the LoE.
     --
-    -- PRECONDITION: @AF.withinFragmentBounds (AF.anchorPoint frag) loe@
-    followsLoEFrag :: LoE (AnchoredFragment (Header blk))
-                   -> AnchoredFragment (Header blk)
-                   -> Bool
-    followsLoEFrag LoEDisabled _ = True
-    followsLoEFrag (LoEEnabled loe) frag =
-         AF.withinFragmentBounds (AF.headPoint loe) frag
-      || AF.withinFragmentBounds (AF.headPoint frag) loe
+    -- The returned fragment is such that:
+    --
+    -- - It is a prefix of the given fragment.
+    -- - If it contains the tip of the LoE fragment, then it contains at most
+    --   @k@ block after it.
+    -- - If it does not contain the tip of the LoE fragment, then it is included
+    --   in the LoE fragment.
+    --
+    -- The fragment is represented by the current chain and a diff with that
+    -- current chain. It is tempting to only consider the suffix of the diff,
+    -- but that would be incorrect, because the diff might not intersect with
+    -- the LoE fragment, because the diff suffix is anchored somewhere on the
+    -- current chain and LoE frag's tip might be older than that anchor.
+    --
+    -- PRECONDITIONS:
+    --
+    -- 1. The given 'ChainDiff' can apply on top of the given 'ChainAndLedger'.
+    -- 2. The LoE fragment intersects with the current selection.
+    trimToLoE ::
+      LoE (AnchoredFragment (Header blk)) ->
+      ChainAndLedger blk ->
+      ChainDiff (Header blk) ->
+      ChainDiff (Header blk)
+    trimToLoE LoEDisabled _ diff = diff
+    trimToLoE (LoEEnabled loe) curChain diff =
+      case Diff.apply (VF.validatedFragment curChain) diff of
+        Nothing -> error "trimToLoE: precondition 1 violated: the given 'ChainDiff' must apply on top of the given 'ChainAndLedger'"
+        Just cand ->
+          case AF.intersect cand loe of
+            Nothing -> error "trimToLoE: precondition 2 violated: the LoE fragment must intersect with the current selection"
+            Just (candPrefix, _, candSuffix, loeSuffix) ->
+              let trimmedCandSuffix = AF.takeOldest (fromIntegral k) candSuffix
+                  trimmedCand =
+                    if AF.null loeSuffix
+                      then fromJust $ AF.join candPrefix trimmedCandSuffix
+                      else candPrefix
+              in Diff.diff (VF.validatedFragment curChain) trimmedCand
 
     -- | We have found a 'ChainDiff' through the VolatileDB connecting the new
     -- block to the current chain. We'll call the intersection/anchor @x@.
@@ -734,29 +754,29 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
          -- ^ The current chain (anchored at @i@) and ledger
       -> LoE (AnchoredFragment (Header blk))
          -- ^ LoE fragment
-      -> LoE Word64
-         -- ^ How many extra blocks to select after @b@ at most.
       -> ChainDiff (HeaderFields blk)
          -- ^ Header fields for @(x,b]@
       -> m (Point blk)
-    switchToAFork succsOf lookupBlockInfo curChainAndLedger loeFrag maxExtra diff = do
+    switchToAFork succsOf lookupBlockInfo curChainAndLedger loeFrag diff = do
         -- We use a cache to avoid reading the headers from disk multiple
         -- times in case they're part of multiple forks that go through @b@.
         let initCache = Map.singleton (headerHash hdr) hdr
         chainDiffs <-
-            fmap (filter (followsLoEFrag loeFrag . Diff.getSuffix))
-
-          -- 4. Filter out candidates that are not preferred over the current
+          -- 5. Filter out candidates that are not preferred over the current
           -- chain.
           --
           -- The suffixes all fork off from the current chain within @k@
           -- blocks, so it satisfies the precondition of 'preferCandidate'.
-          . fmap
+            fmap
               ( filter
                   ( preferAnchoredCandidate (bcfg chainSelEnv) curChain
                   . Diff.getSuffix
                   )
               )
+            -- 4. Trim fragments so that they follow the LoE, that is, they
+            -- extend the LoE or are extended by the LoE. Filter them out
+            -- otherwise.
+          . fmap (fmap (trimToLoE loeFrag curChainAndLedger))
             -- 3. Translate the 'HeaderFields' to 'Header' by reading the
             -- headers from disk.
           . flip evalStateT initCache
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -811,13 +811,13 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+          if rollback < k + loeSuffixLength
+            then Just $ LoEEnabled $ k + loeSuffixLength - rollback
+            else Nothing
         else
-          if rollback > k
-            then Nothing
-            else Just $ LoEEnabled $ k - rollback
+          if rollback < k
+            then Just $ LoEEnabled $ k - rollback
+            else Nothing
```
facundominguez, line 833: This refactoring is changing the behavior of the code. In particular, it is returning `Nothing` where before it would return `Just 0`.

Niols: Damn, you're right, what the hell.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -717,6 +723,11 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
                    -> Bool
     followsLoEFrag LoEDisabled _ = True
     followsLoEFrag (LoEEnabled loe) frag =
+      -- REVIEW: This 'frag' argument is the suffix of a diff (with rollback ==
+      -- 0 in 'addToCurrentChain' and /= 0 in 'switchToAFork'), and this suffix
+      -- might not intersect with the LoE fragment, because 'frag' is anchored
+      -- somewhere on the current chain and LoE frag's tip might be older than
+      -- that.
```
facundominguez: The comments in this commit are introduced to be immediately removed in the next commit. It would be better to just drop this commit.

Niols: Indeed.

Niols:
This force push addresses your two comments (getting rid of the changes that introduce a bug and then fix it, and merging two commits). Additionally, it adds https://github.com/IntersectMBO/ouroboros-consensus/pull/1119/commits/a6816538d8f103a6977263576ca35d2702c60367 (from #1119) to what is now https://github.com/IntersectMBO/ouroboros-consensus/pull/1125/commits/bfd27d0e576de7923a754e67cbef4215da2695d5.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,26 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
 
-    -- Either frag extends loe or loe extends frag
-    --
-    -- PRECONDITION: @AF.withinFragmentBounds (AF.anchorPoint frag) loe@
+    -- Either frag extends loe or loe extends frag. The fragment is represented
+    -- by the current chain and a diff with that current chain. It is tempting
+    -- to only consider the suffix of the diff, but that would be incorrect,
+    -- because the diff might not intersect with the LoE fragment, because the
+    -- diff suffix is anchored somewhere on the current chain and LoE frag's tip
+    -- might be older than that anchor.
```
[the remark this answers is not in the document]
facundominguez: > Also, it strikes me like it may truncate the current selection too

This in turn could truncate to the same point in the current chain for any candidates forking after the intersection with the LoE fragment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,26 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
     followsLoEFrag :: LoE (AnchoredFragment (Header blk))
-                   -> AnchoredFragment (Header blk)
-                   -> Bool
-    followsLoEFrag LoEDisabled _ = True
-    followsLoEFrag (LoEEnabled loe) frag =
-         AF.withinFragmentBounds (AF.headPoint loe) frag
-      || AF.withinFragmentBounds (AF.headPoint frag) loe
+                   -> ChainAndLedger blk
+                   -> ChainDiff (Header blk)
```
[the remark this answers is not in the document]
Niols: True; it used to trim some candidates and reject others, but now it just trims everything. Removing the `Maybe` monad means we introduce the precondition that the LoE intersects with the current selection, but that is enforced in other parts of the code, so that sounds good to me. I will write a version that returns directly a `ChainDiff (Header blk)`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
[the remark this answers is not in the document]
Niols: > it may truncate the current selection

I don't think that is right. It may truncate the candidate fragment that coincides with the current selection, indeed, but the current selection is a stateful thing that is only changed if we find a _better_ candidate, so `followsLoEFrag` will not affect it.

I will propose a new documentation comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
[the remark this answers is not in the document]
Niols: 7b023f1dfed2c4f5c76dbcc252a63792817aa152

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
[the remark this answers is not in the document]
Niols: 1fea816f55571d975eda868dc2b6a9fc58264a14

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -815,6 +841,9 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
             then Nothing
             else Just $ LoEEnabled $ k + loeSuffixLength - rollback
         else
+          -- REVIEW: Since we filter out candidates that fork off the LoE later,
+          -- it does not really seem to matter what we return here and we could
+          -- maybe just return @Nothing@ or merge it in the previous case?
           if rollback > k
             then Nothing
             else Just $ LoEEnabled $ k - rollback
```
facundominguez, line 849: Returning `Nothing` causes the current selection to be left unmodified. I don't think this is what we would want.

The actual value of `k - rollback` is used to decide how long to build the candidates that don't extend the LoE. I like that `computeLoEMaxExtra` returns something meaningful in this scenario.

If less candidates or shorter candidates need to be produced, changes would need to be done elsewhere.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -815,6 +827,9 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
             then Nothing
             else Just $ LoEEnabled $ k + loeSuffixLength - rollback
         else
+          -- REVIEW: Since we filter out candidates that fork off the LoE later,
+          -- it does not really seem to matter what we return here and we could
+          -- maybe just return @Nothing@ or merge it in the previous case?
```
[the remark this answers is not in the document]
amesgen: Another thought: I think we can remove `computeLoEMaxExtra`, `followsLoEFrag` already does everything we need, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,40 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+    followsLoEFrag LoEDisabled _ diff = diff
+    followsLoEFrag (LoEEnabled loe) curChain diff =
+      let cand = fromJust $ Diff.apply (VF.validatedFragment curChain) diff
+          (candPrefix, _, candSuffix, loeSuffix) = fromJust $ AF.intersect cand loe
+          trimmedCandSuffix = AF.takeOldest (fromIntegral k) candSuffix
+          trimmedCand =
+            if AF.null loeSuffix
+              then fromJust $ AF.join candPrefix trimmedCandSuffix
+              else candPrefix
+       in Diff.diff (VF.validatedFragment curChain) trimmedCand
```
amesgen, line 745: `AF.intersect`/`Diff.diff` are linear in the lengths of one of their arguments. Maybe we can run at least a sync benchmark to see whether this has some impact?

amesgen: (FTR: We also have both short- (adapting `maximalCandidates` etc.) and longterm (chain selection datastructure) ideas for how to make this more performant if needed.)

facundominguez:
I went into the trouble of pruning the generation of candidates instead of trimming.
https://github.com/IntersectMBO/ouroboros-consensus/compare/niols/fix-loe...fd/no-trimming

I find the code easier to understand since there are no subtle interactions between generation and trimming.

Uniform tests pass except for:
```
the node is shut down and restarted after some time:       FAIL (0.18s)
    *** Failed! (after 1 test and 11 shrinks):
    Exception:
      FailureEvaluation ourFrag contains fewer than k headers and not close to genesis: 1 vs 2 with anchor At (Block {blockPointSlot = SlotNo 2, blockPointHash = (testHashFromList [0,0,0])})
        error, called at src/ouroboros-consensus/Ouroboros/Consensus/Util/Assert.hs:10:30 in ouroboros-consensus-0.17.0.0-inplace:Ouroboros.Consensus.Util.Assert
        assertWithMsg, called at src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs:743:5 in ouroboros-consensus-0.17.0.0-inplace:Ouroboros.Consensus.MiniProtocol.ChainSync.Client
        assertKnownIntersectionInvariants, called at src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs:1106:20 in ouroboros-consensus-0.17.0.0-inplace:Ouroboros.Consensus.MiniProtocol.ChainSync.Client
    Use --quickcheck-replay="(SMGen 8444413702921376177 13319386596824145657,0)" to reproduce.
    Use -p '/uniform/&&/the node is shut down and restarted after some time/' to rerun this test only.
```

For May I propose that we deliver this PR more or less as is. And then we can try to polish the code for July by tuning the generation of candidates.

I'll see if the changes here are noticeable, I'm guessing that the intersections could be noticeable in the 1 peer case of syncing, when chain selection dominates execution time.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,40 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
-                   -> AnchoredFragment (Header blk)
-                   -> Bool
-    followsLoEFrag LoEDisabled _ = True
-    followsLoEFrag (LoEEnabled loe) frag =
-         AF.withinFragmentBounds (AF.headPoint loe) frag
-      || AF.withinFragmentBounds (AF.headPoint frag) loe
+                   -> ChainAndLedger blk
+                   -> ChainDiff (Header blk)
+                   -> ChainDiff (Header blk)
+    followsLoEFrag LoEDisabled _ diff = diff
```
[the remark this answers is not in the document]
Niols: https://github.com/IntersectMBO/ouroboros-consensus/pull/1125/commits/60af22955797f92d1e69c5a02a0161951b96f430

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
Niols: I am not sure that the value within `Just` makes a real difference in this scenario, however I can believe my proposed change makes a difference in chain selection by transforming a `Just` into a nothing, when we use that to decide in which case we find ourselves.

facundominguez:
There is a small overhead to trimming. But it doesn't seem to delay syncing or not by much in any case.

```
10 peers, trimming:
trimToLoE:   0.475s  count:  2_443
real 4m5s

10 peers, no trimming:
followsLoEFrag:   0.293s  count:  2_442
real 4m16s

1 peer, trimming:
trimToLoE:  24.728s  count: 5_393
real 8m10s

1 peer, no trimming:
followsLoEFrag:   0.267s  count: 2_436
real  8m10s
```

facundominguez (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -783,16 +833,26 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+    computeLoEMaxExtra (LoEEnabled loeFrag) newBlockFrag =
+        -- Both fragments are on the same chain
+        if loeSuffixLength == 0 || rollback == 0 then
+          if rollback > k + loeSuffixLength
+            then Nothing
+            else Just $ LoEEnabled $ k + loeSuffixLength - rollback
+        else
+          -- REVIEW: Since we filter out candidates that fork off the LoE later,
+          -- it does not really seem to matter what we return here and we could
+          -- maybe just return @Nothing@ or merge it in the previous case?
```
facundominguez, line 846: I would leave the code as is. But unless you want other reviewers to give feedback here, I would remove the REVIEW note before merging.

Niols:
```suggestion
```

Niols: Agreed.

Niols:
Bundled everything into one commit and rebased on top of `genesis/milestone13`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,26 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+    -- diff suffix is anchored somewhere on the current chain and LoE frag's tip
+    -- might be older than that anchor.
     followsLoEFrag :: LoE (AnchoredFragment (Header blk))
-                   -> AnchoredFragment (Header blk)
-                   -> Bool
-    followsLoEFrag LoEDisabled _ = True
-    followsLoEFrag (LoEEnabled loe) frag =
-         AF.withinFragmentBounds (AF.headPoint loe) frag
-      || AF.withinFragmentBounds (AF.headPoint frag) loe
+                   -> ChainAndLedger blk
```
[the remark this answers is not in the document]
Niols: That is true, but that would actually be more verbose to use in `addToCurrentChain` and `switchToAFork` so I'd prefer keeping it like this. Alternatively, to make the use of arguments clear, we could have `followsLoEFrag` (now `trimToLoE`) getting only a fragment and then a variation `followsLoEFrag'` that gets a `ChainAndLedger` and a diff, turns it into a fragment, and passes it to `followsLoEFrag`. Only the latter would actually be used in code.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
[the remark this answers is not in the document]
Niols: I am wondering the same. However, `computeLoEMaxExtra` is used in `chainSelectionForBlock` in patterns of the form
```haskell
| ...
, Just maxExtra <- computeLoEMaxExtra loeFrag newBlockFrag -> do
  ...
```
so I'm wondering if getting rid of it wouldn't actually change the behaviour in unforeseen ways, and I'm unsure how. I would need to spend some more time on this, thinking about it and testing it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
Niols: Facundo seems to say it is fine, here: https://github.com/IntersectMBO/ouroboros-consensus/pull/1125#issuecomment-2139421748

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -709,16 +709,40 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = electric $ do
...
+    -- The fragment is represented by the current chain and a diff with that
+    -- current chain. It is tempting to only consider the suffix of the diff,
+    -- but that would be incorrect, because the diff might not intersect with
+    -- the LoE fragment, because the diff suffix is anchored somewhere on the
+    -- current chain and LoE frag's tip might be older than that anchor.
+    --
+    -- PRECONDITIONS:
+    --
+    -- 1. The given 'ChainDiff' can apply on top of the given 'ChainAndLedger'.
+    -- 2. The LoE fragment intersects with the current selection.
```
amesgen, line 731: Assert these with `assertWithMsg`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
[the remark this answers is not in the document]
amesgen: I am pretty confident that removing it won't impact correctness, but maybe it impacts performance? In any case, also seems fine to further clean up this later.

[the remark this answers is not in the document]
Niols: Correctness-wise, it looks alright, at least if I believe the Genesis and ChainDB QSM tests, that both pass many iterations.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
Niols: Enforced preconditions in faece5de1f796ea9dd94664eca11979003c0446f.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
amesgen: `computeLoEMaxExtra` is now gone

amesgen (approved):

PR #1127, opened 2024-05-29
Run a nightly CI job with NoThunks invariants enabled

fraser-iohk:
# Description

This change adds a new CI run that triggers every night (at midnight) and builds and runs all test suites with the `checktvarinvariants` and `checkmvarinvariants` flags enabled in the `strict-checked-vars` package. In our code, we use these invariants to detect when a `TVar` / `MVar` is populated with a value that isn't fully evaluated according to `NoThunks`, and so this nightly test run, while much slower than a normal test run, will guard against potential space leaks by detecting unmarked unevaluated thunks during execution.

## Things to check

- that all of the relevant steps are disabled when the `variant` is `no-thunks`
- that the cached builds won't be clobbered by the new `no-thunks` variant
- that the relevant cached builds *will* be used when the `variant` is `default`
- that any relevant conditions in the steps that care about the `trigger_event` are correctly configured


.github/workflows/cabal.project.no-thunks.local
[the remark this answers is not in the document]
fraser-iohk: didn't realise cabal could do this, fixed!

.github/workflows/ci.yml
```diff
@@ -23,6 +26,10 @@ jobs:
          ghc: ["8.10.7", "9.6.4", "9.8.1"]
          cabal: ["3.10.2.1"]
          os: [ubuntu-latest]
+         variant: [default, no-thunks]
+         exclude:
+          variant:
+            ${{ github.event_name == 'schedule' ? 'no-thunks' : 'default' }}
```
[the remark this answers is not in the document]
fraser-iohk: fixed `exclude` not containing a list 

I'm not sure I understand what you mean with the second part, though. As I understand it at the moment, it will always remove one or the other, so we're either excluding `no-thunks` in non-nightly runs, or `default` the nightly runs.

.github/workflows/ci.yml
```diff
@@ -23,6 +26,10 @@ jobs:
          ghc: ["8.10.7", "9.6.4", "9.8.1"]
          cabal: ["3.10.2.1"]
          os: [ubuntu-latest]
+         variant: [default, no-thunks]
+         exclude:
+           - variant:
+             ${{ github.event_name == 'schedule' ? 'no-thunks' : 'default' }}
```
amesgen, line 32: GHA [says](https://github.com/IntersectMBO/ouroboros-consensus/actions/runs/9289759191) that there is a syntax error here, I think line 32 needs more indentation (or it needs to be moved to line 31).

BTW: https://rhysd.github.io/actionlint/ is a nice tool for these kinds of things.

.github/workflows/ci.yml
```diff
@@ -44,7 +51,7 @@ jobs:
 
     - name: Configure Cabal
       run: |
-        cp .github/workflows/cabal.project.local .
+        cp .github/workflows/cabal.project.${{ matrix.variant }}.local ./cabal.project.local
```
jasagredo, line 54: If you move the project.local file, then the import won't work right? The import like should say `import: .github/workflows/cabal.project.default.local`, no?

.github/workflows/ci.yml
jasagredo: Maybe you cannot use variables in a matrix declaration? Even if I indent the line as expected:
Screenshot 2024-05-30 111104


But if I just write a literal there:
image

.github/workflows/ci.yml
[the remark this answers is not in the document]
fraser-iohk: ah, sorry, you're right, the conditional is flipped, I'll fix it

.github/workflows/ci.yml
fraser-iohk: yup, you're right, fixed it

.github/workflows/ci.yml
fraser-iohk: this section was weirdly indented (to 3 spaces?) before I changed it, so I think I'll just fix all the indentation while I'm here

it's the ternary operator that doesn't work, GHA wants a `cond && ifTrue && ifFalse` instead

.github/workflows/ci.yml
```diff
@@ -12,6 +12,9 @@ on:
   push:
     branches:
       - main
+  schedule:
+    # every night, at midnight
+    - cron: "0 0 * * *"
```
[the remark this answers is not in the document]
fraser-iohk: added a new `nothunks` input for `workflow_dispatch` 👍

jasagredo (approved):
LGTM now

PR #1143, opened 2024-06-13
ImmutableDB: allow to lookup blocks by slot, leverage in db-{analyser,truncater}

amesgen:
This PR adds a new function to the internal ImmutableDB API:
```haskell
-- | Get the hash of the block in the given slot. If the slot contains both
-- an EBB and a non-EBB, return the hash of the non-EBB.
getHashForSlot :: SlotNo -> m (Maybe (HeaderHash blk))
```
It is then used in two ways:

 - In db-analyser: Many analyses do not actually need a ledger state to perform their work. This PR adds support for avoiding this redundant and long work on startup. However, the existing ImmutableDB streaming API needs a point (previously read from the ledger state), not just a slot to start streaming. This is where `getHashForSlot` comes in.

 - In db-truncater: Previously, truncating after a slot took linear time (by iterating over the entire database). With `getHashForSlot`, it is easy to now change it to take constant time. Note that the behavior changes slightly, see the commit message for details.

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE BangPatterns               #-}
 {-# LANGUAGE BlockArguments             #-}
+{-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE GADTs                      #-}
 {-# LANGUAGE GeneralisedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE NamedFieldPuns             #-}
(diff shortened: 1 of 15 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -102,47 +110,80 @@ newtype NumberOfBlocks = NumberOfBlocks { unNumberOfBlocks :: Word64 }
...
-runAnalysis analysisName env@(AnalysisEnv { tracer }) = do
-    traceWith tracer (StartedEvent analysisName)
-    result <- go analysisName
-    traceWith tracer DoneEvent
-    pure result
+  => AnalysisName -> SomeAnalysis blk
+runAnalysis analysisName = case go analysisName of
+    SomeAnalysis p ana -> SomeAnalysis p $ \env@AnalysisEnv{ tracer } -> do
+      traceWith tracer (StartedEvent analysisName)
+      result <- ana env
```
dnadales, line 125: I'd rename `ana` to analysis. Same for `mkAna` :)

dnadales (approved):

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
amesgen: Done :+1:

PR #1152, opened 2024-06-19
db-analyser: print hashes for `--show-slot-block-no`

amesgen:
A concrete use case is to find the intersection between two ChainDBs (eg by diffing the db-analyser output).


ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -70,7 +68,7 @@ parseAnalysis :: Parser AnalysisName
 parseAnalysis = asum [
       flag' ShowSlotBlockNo $ mconcat [
           long "show-slot-block-no"
-        , help "Show slot and block number of all blocks"
+        , help "Show slot and block number and hash of all blocks"
```
amesgen, line 71: Happy to take suggestion for a better name than `--show-slot-block-no` :laughing: Candidates: `--show-slot-block-no-hash` and `--show-point-block-no`.

(But I also don't think it would be too bad to keep the current name.)

PR #1153, opened 2024-06-19
ci.yml: github.event.type => github.event_name

fraser-iohk:
# Description

Fixes `github.event.type` being checked when determining which variants to run during the scheduled nightly CI run, rather than `github.event_name`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         variant: [default, no-thunks]
         exclude:
           - variant:
-              ${{ (github.event_name == 'schedule' || github.event.inputs.nothunks) && 'default' || 'no-thunks' }}
+              ${{ (github.event_name == 'schedule' || inputs.nothunks) && 'default' || 'no-thunks' }}
     env:
       # Modify this value to "invalidate" the Cabal cache.
       CABAL_CACHE_VERSION: "2024-01-29"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         variant: [default, no-thunks]
         exclude:
           - variant:
-              ${{ (github.event.type == 'schedule' || github.event.inputs.nothunks) && 'default' || 'no-thunks' }}
+              ${{ (github.event_name == 'schedule' || github.event.inputs.nothunks) && 'default' || 'no-thunks' }}
```
amesgen, line 36: Maybe this should be `inputs.nothunks` instead of `github.event.inputs.nothunks`, as you want to use it as a boolean? I haven't tested whether this makes a difference:

> The workflow will also receive the inputs in the `github.event.inputs` context. The information in the `inputs` context and `github.event.inputs` context is identical except that the `inputs` context preserves Boolean values as Booleans instead of converting them to strings. The `choice` type resolves to a string and is a single selectable option.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs

amesgen (approved):

.github/workflows/ci.yml
fraser-iohk: oh interesting! I've only tested the manual trigger with the `nothunks` option checked, so I guess it doesn't let you run a default job manually at the moment 🤔 i'll fix it

PR #1157, opened 2024-06-20
The Big Weeder Cleanup

jasagredo:
# Description

A big cleanup of unused functions plus moving the resource registry and the normalform vars into separate packages


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -577,12 +575,12 @@ test-suite infra-test
   main-is: Main.hs
   other-modules:
     Ouroboros.Consensus.Util.Tests
-    Test.Ouroboros.Consensus.ChainGenerator.Tests
```
jasagredo, line 580: Oops, it is `Test.Ouroboros` not `Test.Consensus`, my bad

jasagredo:
This is missing some few things already:
- [ ] HLint: now some pragmas are redundant
- [ ] Changelog entries
- [ ] Add these packages to the badges and to the scripts

jasagredo:
I added the no changelog label because it doesn't make sense to add a changelog fragment for the new packages as they are precisely new

jorisdral:
I don't think `nf-vars` should be in its own package. If it is preferred to remove it from consensus, then I think the functionality should go into `strict-checked-vars`

dnadales (approved):
Thanks for doing this! :broom:

PR #1158, opened 2024-06-21
Forbid future index-states in CI

jasagredo:
This prevents future index-states to land in our cabal project.

In any case this will also be forbidden by cabal itself in the near future, so it is here just as a sanity check.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -116,3 +116,30 @@ jobs:
 
     - name: Diff plans
       run: GH=1 ./scripts/release/cabal-plan-diff.sh
+
+  check-index-states:
+    name: Check index-state timestamps
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


amesgen (approved):
Approving, but do we actually want this now given that the cabal-install 3.12.1.0 with your fix is imminent (https://github.com/haskell/cabal/issues/9885#issuecomment-2182427521) as you mention? The script looks quite ad-hoc.

jasagredo:
It is indeed very ad-hoc, but it happened just now:
```
❯ cabal build all
Warning: Requested index-state 2024-06-18T14:00:00Z is newer than
'cardano-haskell-packages'! Falling back to older state
(2024-06-18T12:42:43Z).
```
This was for some time a future index-state

jasagredo:
Closing as cabal-install 3.12.1.0 was released

Issue #560, opened 2022-09-15
Fix the no-thunk tests failures


yogeshsajanikar:
With strict-stm `checktvarinvariant` flag turned on, there are tests which fail due to presence of the thunk. Review and fix the tests. In case, thunks are required, add the necessary exceptions as mentioned in https://well-typed.com/blog/2020/09/nothunks/ using _AllowThunksIn_

blocks input-output-hk/ouroboros-network#3001

PR #1159, opened 2024-06-24
Fix several NoThunks issues 

fraser-iohk:
# Description

These changes fix several issues in which thunks are left unevaluated, triggering NoThunks errors in test suites when `checktvarinvariants` and `checkmvarinvariants` flags are enabled.

- One of these changes adds a `force` into a test suite, but I'm not sure how happy we are with this solution -- it feels a bit pleasant (but we don't have a strict `List` anywhere convenient 😄)

There are known NoThunks issues in our tests that are not fixed by these changes, but those issues are caused by other packages.

It should also be noted that the behaviour of the NoThunks test runs differ between different compiler versions and optimization levels, so it may be difficult to reproduce detection of some of these uncaught unevaluated thunks.

works towards fixing #560

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
@@ -887,7 +887,7 @@ data LoE a =
   -- When the selection's tip is @k@ blocks after the earliest intersection of
   -- of all candidate fragments, ChainSel will not add new blocks to the
   -- selection.
-  LoEEnabled a
+  LoEEnabled !a
   deriving (Eq, Show, Generic, NoThunks, Functor, Foldable, Traversable)
 
 type GetLoEFragment m blk = LoE (m (AnchoredFragment (Header blk)))
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
--- a/ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
+++ b/ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
@@ -98,6 +98,7 @@ import           Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe)
+import           Data.Maybe.Strict (StrictMaybe (..), strictMaybeToMaybe)
 import           Data.Proxy
 import           Data.Time.Calendar (fromGregorian)
 import           Data.Time.Clock (UTCTime (..))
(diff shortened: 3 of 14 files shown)


ouroboros-consensus/test/consensus-test/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -339,7 +340,7 @@ testOverrideDelay systemStart slotLength maxClockRewind numSlots = do
       withWatcher
         "testOverrideDelay"
         ( knownSlotWatcher btime $ \slotNo -> do
-            atomically $ modifyTVar slotsVar (slotNo :)
+            atomically $ modifyTVar slotsVar $ force . (slotNo :)
```
amesgen, line 343: The alternative would be to use `uncheckedNewTVarM` to create `slotsVar`. I think that would be more in line with what we do in tests usually.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
```diff
@@ -550,7 +550,7 @@ data TestBlockLedgerConfig = TestBlockLedgerConfig {
   tblcHardForkParams :: HardFork.EraParams,
```
amesgen, line 550: Maybe also make this field strict for consistency? (Surprising that it doesn't seem to cause `NoThunks` failures...)

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
```diff
@@ -550,7 +550,7 @@ data TestBlockLedgerConfig = TestBlockLedgerConfig {
...
   -- | `Nothing` means an infinite forecast range.
   -- Instead of SlotNo, it should be something like "SlotRange"
-  tblcForecastRange  :: Maybe SlotNo
+  tblcForecastRange  :: !(Maybe SlotNo) -- StrictMaybe?
```
amesgen, line 553: Using `StrictMaybe` sounds good, we already use it in a few places.

amesgen (approved):
LGTM, only minor comments

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings   #-}
```
jasagredo, line 5: Is this needed?

jasagredo (commented):
I always wonder whether we should derive NoThunks in a trivial way for the tests, everywhere, as in: we don't care if there are thunks in the tests, we care if they are in production code.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
amesgen: Yes, because of 20109f0869cf54e5fd63c95f0e28342d215cfc22

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -887,7 +887,7 @@ data LoE a =
   -- When the selection's tip is @k@ blocks after the earliest intersection of
   -- of all candidate fragments, ChainSel will not add new blocks to the
   -- selection.
-  LoEEnabled a
+  LoEEnabled !a
```
jasagredo, line 890: Done in https://github.com/IntersectMBO/ouroboros-consensus/pull/1113/commits/dd651b114624cbe86486805a256387f8adbdfb45

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestBlock.hs
amesgen: FTR: Another option would be to derive `NoThunks` via [`AllowThunk`](https://hackage.haskell.org/package/nothunks-0.2.1.0/docs/NoThunks-Class.html#t:AllowThunk), as per https://github.com/IntersectMBO/ouroboros-consensus/pull/1159#pullrequestreview-2156428591

Especially as we never "update" `LedgerConfig`s.

ouroboros-consensus/test/consensus-test/Test/Consensus/BlockchainTime/Simple.hs
fraser-iohk: yeah, I think I prefer this over `force`

PR #1162, opened 2024-06-25
Port formal specification to Agda

javierdiaz72:
We shall port the formal specification of the Consensus Layer to Agda.


docs/agda-spec/src/Spec/OperationalCertificate.lagda
```diff
@@ -0,0 +1,141 @@
...
+
+data _⊢_⇀⦇_,OCERT⦈_ where
+\end{code}
+\begin{code}
+  Update-OCert :
+    let (bhb , σ) = bh; open BHBody bhb
+        ⟦ vkₕ , n , c₀ , τ ⟧ᵒᶜ = oc
+        hk = hash issuerVk
+        kp = kesPeriod slot
+        t = kp -ᵏ c₀
```
WhatisRT, line 111: We have `HasAdd` and `HasSubtract` type classes, so you could just use regular `+` and `-` operations for these.

docs/agda-spec/src/Spec/Protocol.lagda
```diff
@@ -0,0 +1,172 @@
...
+  _⊢_⇀⦇_,PRTCL⦈_ : PrtclEnv → PrtclState → BHeader → PrtclState → Type
+\end{code}
+\caption{Protocol transition system types}
+\label{fig:ts-types:prtcl}
+\end{figure*}
+
+\begin{figure*}[h]
+\emph{Protocol helper functions}
+\begin{code}
+hBLeader : BHBody → ∃[ n ] n < 2 ^ 512 -- [0, 2^512)
```
WhatisRT, line 80: I think it'd make sense to define an alias for this type. It comes up a bunch and people that don't know dependent types will be confused.

docs/agda-spec/src/Spec/Protocol.lagda
```diff
@@ -0,0 +1,172 @@
...
+   if f ≡ 1ℚ then ⊤ else λ{f≢1ℚ} →
+     let
+       p     = pos certℕ ℚ./ (2 ^ 512)
+       p≢1ℚ  = ↥p<↧p⇒p≢1 {p} (n<m⇒↥[n/m]<↧[n/m] certℕprf)
+       q     = 1ℚ ℚ.- p
+       1-f≥0 = p≤1⇒1-p≥0 f≤1
+       1-f≢0 = p≢1⇒1-p≢0 f≢1ℚ
+       instance
+         q≢0ℚ  = ℚ.≢-nonZero (p≢1⇒1-p≢0 p≢1ℚ)
+         1-f>0 = ℚ.positive (≤∧≢⇒< 1-f≥0 $ ≢-sym 1-f≢0)
```
WhatisRT, line 99: As already discussed, I think you should hide this. You could also put `λ{f≢1ℚ} →` on a separate line and hide it as well then.

WhatisRT (commented):
We also recently started hiding `field` and `constructor X` in record definitions in the ledger spec as they are somewhat confusing to non-Agda people (e.g. I've gotten the question 'what does "field" mean?') and they don't really add much.

javierdiaz72:
> We also recently started hiding `field` and `constructor X` in record definitions in the ledger spec as they are somewhat confusing to non-Agda people (e.g. I've gotten the question 'what does "field" mean?') and they don't really add much.

Fixed in 72bd2dd5f9364d7e8afdbd8529480074a5c2ac80. Note that, in order to achieve the desired layout, I chose to use a combination of `AgdaAlign` and `AgdaSuppressSpace` instead of `AgdaMultiCode` since the latter removes additional line breaks that should not be removed.

docs/agda-spec/src/Spec/OperationalCertificate.lagda
javierdiaz72: Actually, I didn't use such type classes because the signatures do not match: For instance, `-ᵏ` has type `KESPeriod → KESPeriod → ℕ` instead of `KESPeriod → KESPeriod → KESPeriod` as required by `HasSubtract`.

docs/agda-spec/src/Spec/Protocol.lagda
javierdiaz72: Fixed in 8a72ee453fe1f069bd778b95895e9b0d6b9d54ef.

docs/agda-spec/src/Spec/Protocol.lagda
javierdiaz72: Fixed in 893f3d9579b75ea43259557f8eda2a9d13e27834.

javierdiaz72:
FTR: In addition to the review, the following tasks are still pending:

- [ ] Fix the Nix-related issues (@amesgen)
- [ ] Add a link to the document on the website (@IntersectMBO/ouroboros-consensus)

amesgen (approved):

PR #1168, opened 2024-06-27
Conway: also enable mempool ref script size limit

amesgen:
Also implementing https://github.com/IntersectMBO/ouroboros-consensus/pull/1166#discussion_r1655639209, but in a more fine-grained fashion

The insight here is that `txRefScriptSize` is implemented using `getBabbageTxDict` for Shelley-based blocks, see the usage sites of that function in `Ouroboros.Consensus.Mempool.Impl.Common` and `Ouroboros.Consensus.Mempool.Update`.

PR #1172, opened 2024-07-01
Conway: add ref scripts size to `TxMeasure`

amesgen:
Making sure that we never forge blocks that violate the new max ref scripts size limit per block that is enforced at the ledger level starting in Conway: https://github.com/IntersectMBO/cardano-ledger/pull/4450

Also see #1168 for restoring a similar limit for the size of the total mempool in Conway.


nfrisby (approved):
Looks good, thanks.

Issue #50, opened 2023-04-28
Integrate mempool fairness fix


dnadales:
This issue is about merging [this PR](https://github.com/input-output-hk/ouroboros-consensus/pull/49) [~this PR~](https://github.com/input-output-hk/ouroboros-network/pull/4436). Before merging this we should make sure that no transaction can block the mempool indefinitely, as [it is currently the case](https://github.com/input-output-hk/ouroboros-network/pull/4436/files#diff-533786ff2a23a680b8cda8ef205a5d415c7041f9210e8d2dfad532e6073bfaffR104).

- [x] Add a tx-size check to the mempool ([#21](https://github.com/input-output-hk/ouroboros-consensus/issues/21)).
- [x] Check what happens when the transactions used to forge a block take more space than the block transaction capacity.
- [x] Merge [#70](https://github.com/input-output-hk/io-sim/pull/70).
- [x] Release fixes in [#70](https://github.com/input-output-hk/io-sim/pull/70) to CHaP.
- [x] Update the changelog (del `tryAddTxs`, add `addTx`)
- [x] Ensure that the mempool is livelock free: no transaction should exceed the mempool capacity.
- [x] Rebase atop master with fairness tests.
- [x] Merge https://github.com/input-output-hk/ouroboros-consensus/pull/46
- [x] Make a `fs-sim` revision in CHaP that updates the cabal file according to 506526deecffe9e237a2bc2c755878441fad51c4. (https://github.com/input-output-hk/cardano-haskell-packages/pull/233)
- [x] Remove the source repository package that points to `fs-sim`.
- [x] Perform a final rebase.
- [x] Update the `ouroboros-network` packages bounds in our cabal files, 
- [x] Cut a release in `ouroboros-consensus`. https://github.com/input-output-hk/ouroboros-consensus/pull/54
- [x] Merge to main https://github.com/input-output-hk/ouroboros-consensus/pull/49.
      Waiting on CI.
- [x] Wait for a release of `ouroboros-network` packages that use `io-sim*` 1.1.0.0. https://github.com/input-output-hk/ouroboros-network/pull/4532. 
      Waiting on CI and then a PR in CHaP + merging it.
- [x] Remove the source repository packages that point to `ouroboros-network` Update the CHaP index to `ouroboros-network` release or later, on top of this PR https://github.com/input-output-hk/ouroboros-consensus/pull/54
- [x] Publish it to CHaP (tag the version and push new packages, PR and merge)

Issue #21, opened 2023-03-22
Check that the transactions submitted at the mempool have a valid size


dnadales:
We should change the mempool admission test from:

```haskell
mempoolsize + txsize <= mempoolcapacity
```

to:

```haskell
mempoolsize < mempoolcapacity
```

This would mean that as soon as there is a byte free in the mempool we add a transaction to it.

We should look out for invariants that assume that `mempoolsize <= mempoolcapacity` always hold. 

The expect that the ledger checks the sanity of the transactions size. We should document this expectation somewhere.

The  mempool admission check can be found [here](https://github.com/input-output-hk/ouroboros-network/pull/4436/files#diff-533786ff2a23a680b8cda8ef205a5d415c7041f9210e8d2dfad532e6073bfaffR188).

This change should take place on top of https://github.com/input-output-hk/ouroboros-network/pull/4436.

dnadales:
> We should look out for invariants that assume that `mempoolsize <= mempoolcapacity` always hold.

Upon inspection of the code, it seems that this cannot possibly be an invariant. The mempool API states that the maximum capacity can dynamically change when the ledger state is updated. 

```haskell
-- Note that the capacity of the Mempool, unless it is overridden with
-- 'MempoolCapacityBytesOverride', can dynamically change when the
-- ledger state is updated: it will be set to twice the current ledger's
-- maximum transaction capacity of a block.
```

The code can't possibly rely on `mempoolsize <= mempoolcapacity` because:

```haskell
-- When the capacity happens to shrink at some point, we /do not/ remove
-- transactions from the Mempool to satisfy this new lower limit.
```

dnadales:
Changes pushed to https://github.com/input-output-hk/ouroboros-network/pull/4436

nfrisby:
While working on PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1175, Esgen noticed that I had reverted the change suggested in this Issue's description. After some study, I think it's OK to revert that change.

Specifically: PR https://github.com/IntersectMBO/ouroboros-consensus/pull/49 made both the change suggested by this Issue's description as well as adding some centralized `MVar`s. I believe those `MVar`s supplant the description's suggested change: as soon as a thread adding a tx grabs the `MVar`, it will hold it until that tx fits in the mempool.

PR #1175, opened 2024-07-02
Consolidate TxLimits

nfrisby:
Closes https://github.com/IntersectMBO/ouroboros-consensus/issues/1177.

Beyond that Issue, this PR also changes the mempool's definition of full. The mempool will accepte transactions until the tx chosen to next enter the mempool would cause any component of the capacity vector (byte size, exunit steps, exunit mems, ref script byte size) to exceed its limit. The default capacity is chosen as twice the capacity of a block. Such a capacity can admit more txs than could fit into two back-to-back blocks, but no single component (eg ExUnits) of the mempool's contents' size could exceed what could fit in two blocks.

The mempool capacity has two consequences.

- It implements _latency hiding_: txs can cover the network a little in advance, so that two blocks on a chain whose slots are very close together could still both be full (if there are enough submitted txs).
- It bounds the potential burst of CPU load necessary to fill an empty mempool, which otherwise could be a DoS vector.

Also closes #1168 by superseding the "ad-hoc" mempool reference scripts capacity limit.

This PR also moves the logic for choosing which prefix of the mempool to put in the block out of the forging functions and into the NodeKernel forging thread --- the mempool snapshot is the perfect place to do that, since tx measurements (each using the correct ledger state) are already determined.

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
@@ -5,7 +5,6 @@
 {-# LANGUAGE GADTs                   #-}
 {-# LANGUAGE LambdaCase              #-}
 {-# LANGUAGE MultiParamTypeClasses   #-}
-{-# LANGUAGE NumericUnderscores      #-}
 {-# LANGUAGE ScopedTypeVariables     #-}
 {-# LANGUAGE TypeApplications        #-}
 {-# LANGUAGE TypeFamilies            #-}
@@ -29,7 +28,6 @@ module Ouroboros.Consensus.Shelley.Eras (
   , StandardMary
   , StandardShelley
     -- * Shelley-based era
-  , BabbageTxDict (..)
   , ConwayEraGovDict (..)
   , ShelleyBasedEra (..)
   , WrapTx (..)
@@ -47,7 +45,6 @@ import           Cardano.Ledger.Alonzo (AlonzoEra)
 import qualified Cardano.Ledger.Alonzo.Rules as Alonzo
 import qualified Cardano.Ledger.Alonzo.Translation as Alonzo
 import qualified Cardano.Ledger.Alonzo.Tx as Alonzo
-import qualified Cardano.Ledger.Api as SL
 import qualified Cardano.Ledger.Api.Era as L
 import           Cardano.Ledger.Babbage (BabbageEra)
 import qualified Cardano.Ledger.Babbage.Rules as Babbage
@@ -60,8 +57,6 @@ import qualified Cardano.Ledger.Conway.Rules as Conway
 import qualified Cardano.Ledger.Conway.Rules as SL
                      (ConwayLedgerPredFailure (..))
 import qualified Cardano.Ledger.Conway.Translation as Conway
-import qualified Cardano.Ledger.Conway.Tx as SL
-import qualified Cardano.Ledger.Conway.UTxO as SL
 import           Cardano.Ledger.Core as Core
 import           Cardano.Ledger.Crypto (StandardCrypto)
 import           Cardano.Ledger.Keys (DSignable, Hash)
@@ -73,13 +68,11 @@ import           Cardano.Ledger.Shelley.Core as Core
 import qualified Cardano.Ledger.Shelley.LedgerState as SL
 import qualified Cardano.Ledger.Shelley.Rules as SL
 import qualified Cardano.Ledger.Shelley.Transition as SL
-import qualified Cardano.Ledger.Val as SL
 import qualified Cardano.Protocol.TPraos.API as SL
 import           Control.Monad.Except
 import           Control.State.Transition (PredicateFailure)
 import           Data.Data (Proxy (Proxy))
 import           Data.List.NonEmpty (NonEmpty ((:|)))
-import           Lens.Micro ((^.))
 import           NoThunks.Class (NoThunks)
 import           Ouroboros.Consensus.Ledger.SupportsMempool
                      (WhetherToIntervene (..))
@@ -169,16 +162,6 @@ class ( Core.EraSegWits era
   -- | Whether the era has an instance of 'CG.ConwayEraGov'
   getConwayEraGovDict :: proxy era -> Maybe (ConwayEraGovDict era)
 
-  getBabbageTxDict :: proxy era -> Maybe (BabbageTxDict era)
-
-data BabbageTxDict era where
-    BabbageTxDict ::
-         SL.BabbageEraTxBody era
-      => (Integer -> Integer -> SL.ApplyTxError era)
-         -- ^ Construct an arbitrary ledger error with two integers as its
-         -- payload.
-      -> BabbageTxDict era
-
 data ConwayEraGovDict era where
     ConwayEraGovDict :: CG.ConwayEraGov era => ConwayEraGovDict era
 
@@ -189,7 +172,7 @@ isBeforeConway _ =
 -- | The default implementation of 'applyShelleyBasedTx', a thin wrapper around
 -- 'SL.applyTx'
 defaultApplyShelleyBasedTx ::
-     forall era. ShelleyBasedEra era
+     ShelleyBasedEra era
   => SL.Globals
   -> SL.LedgerEnv era
   -> SL.LedgerState era
@@ -201,55 +184,11 @@ defaultApplyShelleyBasedTx ::
        , SL.Validated (Core.Tx era)
        )
 defaultApplyShelleyBasedTx globals ledgerEnv mempoolState _wti tx = do
-    refScriptPredicate
     SL.applyTx
       globals
       ledgerEnv
       mempoolState
       tx
-  where
-    refScriptPredicate = case getBabbageTxDict (Proxy @era) of
-      Nothing -> pure ()
-      Just (BabbageTxDict mkError)
-        -- The ledger rules of Conway (and later eras) already handle ref
-        -- scripts appropriately, so we only need to perform the checks below
-        -- for Babbage.
-        | not $ isBeforeConway (Proxy @era)
-        -> pure ()
-        -- Reject it if it has more than 100 kibibytes of ref script.
-        | refScriptsSize > totalRefScriptsSizeLimit
-        -> throwError $ mkError
-            -- As we are reusing an existing error message, we add a large
-            -- number to make users running into this are productively irritated
-            -- and post this error message somewhere where they can receive
-            -- help/context.
-            (toInteger refScriptsSize           + 1_000_000_000)
-            (toInteger totalRefScriptsSizeLimit + 1_000_000_000)
-        -- Reject it if it has more than 50 kibibytes of ref script and does not
-        -- satisfy an additional fee as calculated in the table below.
-        | refScriptsSize > freeOfChargeRefScriptsBytes
-        , actualFee < expectedFee
-        -> throwError $ mkError
-            -- See above for why we add a large constant.
-            (SL.unCoin actualFee   + 100_000_000)
-            (SL.unCoin expectedFee + 100_000_000)
-        | otherwise -> pure ()
-        where
-          totalRefScriptsSizeLimit :: Int
-          totalRefScriptsSizeLimit = 100 * 1024
-
-          freeOfChargeRefScriptsBytes :: Int
-          freeOfChargeRefScriptsBytes = 50 * 1024
-
-          actualFee = tx ^. SL.bodyTxL . SL.feeTxBodyL
-          expectedFee = minFee SL.<+> refScriptsFee
-            where
-              minFee = SL.getMinFeeTx (SL.ledgerPp ledgerEnv) tx 0
-              refScriptsFee = SL.tierRefScriptFee 1.2 25600 15 refScriptsSize
-
-          refScriptsSize = SL.txNonDistinctRefScriptsSize utxo tx
-
-          utxo = SL.utxosUtxo . SL.lsUTxOState $ mempoolState
 
 defaultGetConwayEraGovDict :: proxy era -> Maybe (ConwayEraGovDict era)
 defaultGetConwayEraGovDict _ = Nothing
@@ -260,15 +199,11 @@ instance (SL.PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
 
   getConwayEraGovDict = defaultGetConwayEraGovDict
 
-  getBabbageTxDict _ = Nothing
-
 instance (SL.PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
   => ShelleyBasedEra (AllegraEra c) where
   applyShelleyBasedTx = defaultApplyShelleyBasedTx
 
   getConwayEraGovDict = defaultGetConwayEraGovDict
-
-  getBabbageTxDict _ = Nothing
 
 instance (SL.PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
   => ShelleyBasedEra (MaryEra c) where
@@ -276,40 +211,21 @@ instance (SL.PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
 
   getConwayEraGovDict = defaultGetConwayEraGovDict
 
-  getBabbageTxDict _ = Nothing
-
 instance (SL.PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
   => ShelleyBasedEra (AlonzoEra c) where
   applyShelleyBasedTx = applyAlonzoBasedTx
 
   getConwayEraGovDict = defaultGetConwayEraGovDict
 
-  getBabbageTxDict _ = Nothing
-
 instance (Praos.PraosCrypto c) => ShelleyBasedEra (BabbageEra c) where
   applyShelleyBasedTx = applyAlonzoBasedTx
 
   getConwayEraGovDict = defaultGetConwayEraGovDict
 
-  getBabbageTxDict _ = Just $ BabbageTxDict $ \a b ->
-      SL.ApplyTxError
-    $ pure
-    $ SL.UtxowFailure
-    $ Babbage.UtxoFailure
-    $ Babbage.AlonzoInBabbageUtxoPredFailure
-    $ Alonzo.MaxTxSizeUTxO a b
-
 instance (Praos.PraosCrypto c) => ShelleyBasedEra (ConwayEra c) where
   applyShelleyBasedTx = applyAlonzoBasedTx
 
   getConwayEraGovDict _ = Just ConwayEraGovDict
-
-  getBabbageTxDict _ = Just $ BabbageTxDict $ \a b ->
-      SL.ApplyTxError
-    $ pure
-    $ Conway.ConwayUtxowFailure
-    $ Conway.UtxoFailure
-    $ Conway.MaxTxSizeUTxO a b
 
 applyAlonzoBasedTx :: forall era.
   ( ShelleyBasedEra era,
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE FlexibleContexts         #-}
-{-# LANGUAGE OverloadedStrings        #-}
 {-# LANGUAGE ScopedTypeVariables      #-}
 {-# LANGUAGE TypeApplications         #-}
 {-# LANGUAGE TypeFamilies             #-}
@@ -13,14 +12,11 @@ import qualified Cardano.Ledger.Shelley.API as SL (Block (..), extractTx)
 import qualified Cardano.Ledger.Shelley.BlockChain as SL (bBodySize)
 import qualified Cardano.Protocol.TPraos.BHeader as SL
 import           Control.Exception
-import           Control.Monad.Except
-import           Data.List as List (foldl')
 import qualified Data.Sequence.Strict as Seq
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.SupportsMempool
-import           Ouroboros.Consensus.Mempool (TxLimits)
 import           Ouroboros.Consensus.Protocol.Abstract (CanBeLeader, IsLeader)
 import           Ouroboros.Consensus.Protocol.Ledger.HotKey (HotKey)
 import           Ouroboros.Consensus.Shelley.Eras (EraCrypto)
@@ -32,7 +28,6 @@ import           Ouroboros.Consensus.Shelley.Ledger.Mempool
 import           Ouroboros.Consensus.Shelley.Protocol.Abstract (ProtoCrypto,
                      ProtocolHeaderSupportsKES (configSlotsPerKESPeriod),
                      mkHeader)
-import           Ouroboros.Consensus.Util.Assert
 
 {-------------------------------------------------------------------------------
   Forging
@@ -40,14 +35,14 @@ import           Ouroboros.Consensus.Util.Assert
 
 forgeShelleyBlock ::
      forall m era proto.
-      (ShelleyCompatible proto era, TxLimits (ShelleyBlock proto era), Monad m)
+     (ShelleyCompatible proto era, Monad m)
   => HotKey (EraCrypto era) m
   -> CanBeLeader proto
   -> TopLevelConfig (ShelleyBlock proto era)
   -> BlockNo                                      -- ^ Current block number
   -> SlotNo                                       -- ^ Current slot number
   -> TickedLedgerState (ShelleyBlock proto era)   -- ^ Current ledger
-  -> [Validated (GenTx (ShelleyBlock proto era))] -- ^ Txs to add in the block
+  -> [Validated (GenTx (ShelleyBlock proto era))] -- ^ Txs to include
   -> IsLeader proto
   -> m (ShelleyBlock proto era)
 forgeShelleyBlock
@@ -64,15 +59,16 @@ forgeShelleyBlock
     let blk = mkShelleyBlock $ SL.Block hdr body
     return $
       assert (verifyBlockIntegrity (configSlotsPerKESPeriod $ configConsensus cfg) blk) $
-      assertWithMsg bodySizeEstimate blk
+      blk
   where
     protocolVersion = shelleyProtocolVersion $ configBlock cfg
 
     body =
         SL.toTxSeq @era
-      . Seq.fromList
-      . fmap extractTx
-      $ takeLargestPrefixThatFits tickedLedger txs
+      $ Seq.fromList
+      $ fmap extractTx txs
+
+    actualBodySize = SL.bBodySize protocolVersion body
 
     extractTx :: Validated (GenTx (ShelleyBlock proto era)) -> Core.Tx era
     extractTx (ShelleyValidatedTx _txid vtx) = SL.extractTx vtx
@@ -83,20 +79,3 @@ forgeShelleyBlock
       . castHash
       . getTipHash
       $ tickedLedger
-
-    bodySizeEstimate :: Either String ()
-    bodySizeEstimate
-      | actualBodySize > estimatedBodySize + fixedBlockBodyOverhead
-      = throwError $
-          "Actual block body size > Estimated block body size + fixedBlockBodyOverhead: "
-            <> show actualBodySize
-            <> " > "
-            <> show estimatedBodySize
-            <> " + "
-            <> show (fixedBlockBodyOverhead :: Int)
-      | otherwise
-      = return ()
-
-    estimatedBodySize, actualBodySize :: Int
-    estimatedBodySize = fromIntegral $ List.foldl' (+) 0 $ map (txInBlockSize . txForgetValidated) txs
-    actualBodySize    = SL.bBodySize protocolVersion body
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
@@ -149,12 +149,6 @@ instance ShelleyCompatible proto era
 
   txForgetValidated (ShelleyValidatedTx txid vtx) = ShelleyTx txid (SL.extractTx vtx)
 
-  txRefScriptSize _cfg st (ShelleyTx _ tx) = case getBabbageTxDict (Proxy @era) of
-      Nothing              -> 0
-      Just BabbageTxDict{} -> SL.txNonDistinctRefScriptsSize utxo tx
-    where
-      utxo = SL.getUTxO . tickedShelleyLedgerState $ st
-
 mkShelleyTx :: forall era proto. ShelleyBasedEra era => Tx era -> GenTx (ShelleyBlock proto era)
 mkShelleyTx tx = ShelleyTx (SL.txIdTxBody @era (tx ^. bodyTxL)) tx
 
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
@@ -7,7 +7,6 @@
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns             #-}
-{-# LANGUAGE NumericUnderscores         #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE Rank2Types                 #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
@@ -52,14 +51,12 @@ import           Cardano.Ledger.Binary (Annotator (..), DecCBOR (..),
                      ToCBOR (..), toPlainDecoder)
 import qualified Cardano.Ledger.Conway.Rules as ConwayEra
 import qualified Cardano.Ledger.Conway.Rules as SL
-import qualified Cardano.Ledger.Conway.Tx as SL (tierRefScriptFee)
 import qualified Cardano.Ledger.Conway.UTxO as SL
 import qualified Cardano.Ledger.Core as SL (txIdTxBody)
 import           Cardano.Ledger.Crypto (Crypto)
 import qualified Cardano.Ledger.SafeHash as SL
 import qualified Cardano.Ledger.Shelley.API as SL
 import qualified Cardano.Ledger.Shelley.Rules as ShelleyEra
-import qualified Cardano.Ledger.Val as SL (zero, (<+>))
 import           Control.Arrow ((+++))
 import           Control.Monad (guard)
 import           Control.Monad.Except (Except, liftEither)
@@ -302,12 +299,6 @@ validateMaybe ::
   -> Maybe a
   -> V.Validation (TxErrorSG era) a
 validateMaybe err mb = V.validate (TxErrorSG err) id mb
-
-validateGuard ::
-     SL.ApplyTxError era
-  -> Bool
-  -> V.Validation (TxErrorSG era) ()
-validateGuard err b = validateMaybe err $ guard b
 
 runValidation ::
      V.Validation (TxErrorSG era) a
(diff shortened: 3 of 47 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxMonitor/Server.hs
```diff
@@ -31,41 +36,43 @@ localTxMonitorServer mempool =
...
       , recvMsgGetSizes = do
-          let MempoolSize{msNumTxs,msNumBytes} = snapshotMempoolSize snapshot
+          let MempoolSize{msNumTxs,msSize} = snapshotMempoolSize snapshot
           let sizes = MempoolSizeAndCapacity
-                { capacityInBytes = getMempoolCapacityBytes capacity
-                , sizeInBytes     = msNumBytes
+                { capacityInBytes = getSizeInBytes $ txMeasureBytes snapshot capacity
+                , sizeInBytes     = getSizeInBytes $ txMeasureBytes snapshot msSize
                 , numberOfTxs     = msNumTxs
                 }
```
amesgen, line 71: Shall we (create an issue to) enrich the LocalTxMonitor API here to return the full-dimensional capacity (as a new request/response pair probably)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
```diff
@@ -1,28 +1,49 @@
...
+      , Typeable xs
+      , IsNonEmpty xs
+      , BoundedMeasure (HardForkTxMeasure xs)
+      , NoThunks (HardForkTxMeasure xs)
+      ) => CanHardFork xs where
+  type HardForkTxMeasure xs
+
+  hardForkMeasureTx :: SOP.NS WrapTxMeasure xs -> HardForkTxMeasure xs
+
+  hardForkTxMeasureBytes :: proxy xs -> HardForkTxMeasure xs -> SizeInBytes
```
amesgen, line 45: This looks good :+1:; `HardForkTxMeasure` basically is the "union" of the per-era `TxMeasure`s; we chould even try to make that more explicit, but probably not worth it: 
```haskell
type        TxMeasureDims :: Type -> [Type]
type family TxMeasureDims blk

type TxMeasure blk = NP I (TxMeasureDims blk)

-- for example
type instance TxMeasureDims ByronBlock = 
  '[TxSizeInBytes]
type instance TxMeasureDims (ShelleyBlock (TPraos c) (AlonzoEra c)) =  
  '[TxSizeInBytes, ExUnits' (WithTop Natural)]
type instance TxMeasureDims (ShelleyBlock (Praos c)  (ConwayEra c)) =  
  '[TxSizeInBytes, ExUnits' (WithTop Natural), RefsScriptsSizeInBytes]

type instance TxMeasureDims (HardForkBlock xs) =
  Nub (ConcatMapTxMeasureDims xs) 
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -217,9 +205,25 @@ pureTryAddTx cfg txSize wti tx is
...
+  | not (tx Measure.<= cap) = True   -- let it be found invalid
+  | otherwise               =
+    go (isTxs is) (max 0 $ isMultiplicity is - 1)
+  where
+    cap = isCapacity is
+
+    go !txseq = \case
+      0 -> msSize (TxSeq.toMempoolSize txseq) `Measure.plus` tx Measure.<= cap
+      n -> case txseq of
+        TxSeq.Empty -> go txseq 0
```
amesgen, line 224: When `txseq` is `TxSeq.Empty`: As we already established `tx Measure.<= cap` after the first guard above, we could also directly return `True` here as
```haskell
  go txseq 0
≡ msSize (TxSeq.toMempoolSize txseq) `Measure.plus` tx Measure.<= cap
≡ Measure.zero `Measure.plus`                       tx Measure.<= cap
≡                                                   tx Measure.<= cap
≡ True
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Init.hs
```diff
@@ -37,12 +38,11 @@ openMempool ::
   => ResourceRegistry m
   -> LedgerInterface m blk
   -> LedgerConfig blk
-  -> MempoolCapacityBytesOverride
+  -> TxOverrides blk
```
amesgen, line 41: Currently, the haddocks of `TxOverrides` say that it is about overriding the size of forged blocks (only enabling to make blocks *smaller* than allowed by the ledger):

https://github.com/IntersectMBO/ouroboros-consensus/blob/f2f942b4279582a4a5b296c04698d0a2e9c12c82/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Capacity.hs#L58-L76

As mentioned by you already on Slack, this might not be expressive enough (you can't increase the capacity); [here](https://input-output-rnd.slack.com/archives/CDA6LUXAQ/p1719878861071389?thread_ts=1719870923.905459&cid=CDA6LUXAQ), you asked

> Do you think "how many block's worth of tx" would be equally useful as the current "how many bytes of tx"?

which was positively affirmed. So what about instead of using `TxOverrides` here, we instead take `multiplicityOverride :: Maybe Int` here, and set `isMultiplicity = fromMaybe 2 multiplicityOverride`?

amesgen: Ah, this seems to conflict a bit with this approach to testing capacity limits:
https://github.com/IntersectMBO/ouroboros-consensus/blob/f2f942b4279582a4a5b296c04698d0a2e9c12c82/ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs#L28-L31

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxMonitor/Server.hs
nfrisby: Right, it wasn't clear to me.

@KtorZ does the use case for `LocalTxMonitor` care about the _other_ sizes of the mempool? EG total script ExUnits, etc --- beyond just the byte size?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
nfrisby: Yes; a compositional approach would technically work here I think: Every tx has a measure in every era, and they're just zeros in the other eras.

However, I think that explodes operationally, pretty quickly. We'd already have, let's see... 9 constant-zeros in the size of every Conway tx, for example (Byron, Shelley, Allegra, and Mary get one, Alonzo gets 2 and Babbage gets 3)! Plus all the pointers in the heap representation, etc. I anticipate that will become a noticeable overhead over time.

So far -- we could also require each era's measure can represent the measures of previous eras (a la the ledger's `Translate` classes). If that requires overhead, then I think it'd take a _very_ long time for it to become a problem.

For now though, I did the explicit non-compositional thing in the interface (but the translate-y thing in the implementation), just to be "safe" for now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxMonitor/Server.hs
amesgen: For context: This PR changes the mempool capacity to not only look at tx size, but also at the other measures, like script ExUnits and (starting with Babbage) reference scripts size. So a mempool might be almost full w.r.t. tx byte size, but almost empty w.r.t. reference scripts size. 

Concretely, adding a tx to the mempool will not block iff forging two blocks back-to-back now with the current mempool contents plus the new tx would result in an empty mempool.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Init.hs
nfrisby: Yeah. We're fully in sync without even talking yet :)

I'm thinking: one option to increase the block multiplicity (ie an argument to `max`) and one option to decrease the block capacity (ie an argument to `min`).

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -299,7 +299,12 @@ instance (Praos.PraosCrypto c) => ShelleyBasedEra (ConwayEra c) where
 
   getConwayEraGovDict _ = Just ConwayEraGovDict
 
-  getBabbageTxDict _ = Nothing
+  getBabbageTxDict _ = Just $ BabbageTxDict $ \a b ->
+      SL.ApplyTxError
+    $ pure
+    $ Conway.ConwayUtxowFailure
+    $ Conway.UtxoFailure
+    $ Conway.MaxTxSizeUTxO a b
```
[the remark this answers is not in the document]
nfrisby: By "this" you mean the error function, right? I've been assuming I still need the dict in order to calculate the ref script bytes measure.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxMonitor/Server.hs
KtorZ: It definitely does @nfrisby, since the main use case people have for looking at the size is knowing if they can fit one more transaction. And the transaction may overflow via any of the dimension. If we could eventually expose those other dimensions, that'd be great!

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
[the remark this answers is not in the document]
amesgen: No, you shouldn't need the `Dict` anywhere, you removed its only other use here: https://github.com/IntersectMBO/ouroboros-consensus/pull/1175/files#diff-5c74f1f9154322865ac42147eb9f16bab59458c5d53940b9624b74d62ef498d1L153-L157 (`Ouroboros.Consensus.Shelley.Ledger.Mempool.txRefScriptSize`).

Also compare #1168 (which should be superseded by this PR)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/LocalTxMonitor/Server.hs
nfrisby: Thanks for the input! I opened https://github.com/IntersectMBO/ouroboros-consensus/issues/1178

nfrisby:
Phew! I think it's compiling without warnings and I'm optimistic the tests will pass on CI.

But the commit history is a total mess, and some new comments are missing, at the very least.

intolerable:
I also have some questions about what the `TxMeasure` types should look like, especially in the context of #1178, but they're probably better suited for discussion on a call

fraser-iohk:
I also have some questions about what the `TxMeasure` types should look like, especially in the context of #1178, but they're probably better suited for discussion on a call

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Capacity.hs
```diff
@@ -1,189 +1,45 @@
...
-
--- | Apply the override
-applyOverrides ::
-     TxLimits blk
-  => TxOverrides blk
-  -> TxMeasure blk
-  -> TxMeasure blk
-applyOverrides (TxOverrides m') m = Measure.min m' m
+-- | Overriding the size of the mempool in terms of blocks
+newtype MempoolMultiplicityOverride = MempoolMultiplicityOverride Int
```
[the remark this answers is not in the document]
nfrisby: Thanks for the feedback.

Perhaps BlockDepth? BlockCount?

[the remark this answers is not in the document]
fraser-iohk: I think either of those would be an improvement

nfrisby:
I rebased

nfrisby:
The next step for this PR is to merge https://github.com/IntersectMBO/ouroboros-consensus/pull/1182. Eventually I'll rebase this one on top of that.

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
```diff
@@ -701,16 +702,15 @@ withTestMempool setup@TestSetup {..} prop =
...
         (not (isOverride testMempoolCapOverride))
-        "NoMempoolCapacityBytesOverride"
+        "no capacity override"
     $ classify (null testInitialTxs)       "empty Mempool"
     $ classify (not (null testInitialTxs)) "non-empty Mempool"
     $ runSimOrThrow setUpAndRun
   where
-    isOverride (MempoolCapacityBytesOverride _) = True
-    isOverride NoMempoolCapacityBytesOverride   = False
+    isOverride = maybe False (\_ -> True)
```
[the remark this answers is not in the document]
nfrisby: Code is gone

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -41,7 +40,10 @@ import           Ouroboros.Consensus.Util.Assert
 
 forgeShelleyBlock ::
      forall m era proto.
-      (ShelleyCompatible proto era, TxLimits (ShelleyBlock proto era), Monad m)
+      ( ShelleyCompatible proto era
+      , TxLimits (ShelleyBlock proto era)
+      , Monad m
+      )
```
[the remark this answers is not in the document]
nfrisby: Diff is gone

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Capacity.hs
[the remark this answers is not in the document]
nfrisby: For now, I'm leaving the override as a byte value. It's used to compute a "multiplicity", but only internally.

nfrisby:
I don't know if we want to merge this before the Chang fork, so this remains in Draft. I think that's the only reason, though.

ouroboros-consensus/bench/mempool-bench/Main.hs
```diff
@@ -129,18 +132,14 @@ main = withStdTerminalHandles $ do
...
+openMempoolWithCapacity :: ByteSize ->  IO (MockedMempool IO TestBlock)
+openMempoolWithCapacity capacity =
+    Mocked.openMockedMempool (Mempool.mkCapacityBytesOverride  capacity)
                              Tracer.nullTracer
-                             TestBlock.txSize
                              Mocked.MempoolAndModelParams {
                                  Mocked.immpInitialState = TestBlock.initialLedgerState
                                , Mocked.immpLedgerConfig = TestBlock.sampleLedgerConfig
                              }
-  where
```
nfrisby, line 141: I moved this whole size calculation (which caused the regression failure in this PR's CI) into the `withResource` stage --- ie "out" of "the benchmark".

What do you think @dnadales ? Was it intentionally being measured, or no?

nfrisby:
OK. I added a few commits, after chatting with Esgen.

- b90124a81 - consensus-cardano: use ConwayMeasure in Babbage
- 93829e25e - TOSQUASH fixup overflow workarounds and their explanations
- a866dbb3e - consensus: use Natural for ByteSize

The last one supersedes the middle one. Despite the last's commit message already asserting the performance penalty is acceptable, but this deserves some due diligence.

I have a relatively simple idea for that: use (a trivial new) db-analyser to emit the (new) TxMeasure for a two recent epochs from mainnet. Build a benchmark that simply adds and removes "transactions" that are nothing except one of those recorded measurements (in the same order as observed on mainnet). This can be done at the `TxSeq` level, and so won't require the menagerie of `blk` instances.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -172,23 +167,16 @@ pureTryAddTx ::
...
-  | let curSize = msNumBytes  $ isMempoolSize is
-  , curSize < getMempoolCapacityBytes (isCapacity is)
-    -- ... if the mempool has less than 2.5 mebibytes of ref scripts.
-  , let maxTotalRefScriptSize = 5 * 512 * 1024 -- 2.5 Mebibytes
-        curTotalRefScriptSize = isTotalRefScriptSize is
-  , curTotalRefScriptSize Prelude.< maxTotalRefScriptSize
+pureTryAddTx cfg wti tx is
+    -- We add the transaction if it wouldn't overrun any component of the
+    -- mempool capacity.
+  | TxSeq.toSize (isTxs is) `Measure.plus` sz Measure.<= isCapacity is
```
amesgen, line 179: The previous logic (add the transaction if there is at least one byte free left in the mempool) was added for fairness reasons: #21 (larger issue #50) (also: https://github.com/IntersectMBO/ouroboros-network/pull/4436#discussion_r1159512582)

I don't know how crucial this change was for overall mempool fairness; there was a specific scenario that prompted #50, but I don't remember the details.

nfrisby: Do you agree with my reasoning here? https://github.com/IntersectMBO/ouroboros-consensus/issues/21#issuecomment-2229479007

nfrisby:
That latest commit, "txMeasure now fails if the tx is too big", is somewhat optional, but I like that it eliminates that `error` case.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
amesgen: Yes, I can't think of any particular benefit of #21 given the use of the `MVar` :+1: Would be nice to get that confirmed eventually by Duncan/Damian (if only to understand the context back then), but this doesn't have to block this PR.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -703,8 +706,11 @@ getMempoolReader mempool = MempoolReader.TxSubmissionMempoolReader
                                       snapshotHasTx } =
       MempoolReader.MempoolSnapshot
         { mempoolTxIdsAfter = \idx ->
-            [ (txId (txForgetValidated tx), idx', getTxSize mempool (txForgetValidated tx))
-            | (tx, idx') <- snapshotTxsAfter idx
+            [ ( txId (txForgetValidated tx)
+              , idx'
+              , fromIntegral $ unByteSize byteSize   -- TODO overflow?
```
nfrisby, line 711: a TODO here

nfrisby: Bah. I just lost a really nice draft in this message. I don't have the heart to re-write it right now :'(

summary:

- Fraser raised a concern about encoding sizes in LocalTxMonitor replies
- Chatted with Esgen and distilled down a very simple concrete correctness argument
- `Natural` trivializes the correctness of all _other_ logic; it unfortunately merely defers the correctness concerns to the encoder (... which is "_merely_" for a Node-to-Client QoL feature, but still quite important)
- the mempool should (carefully!) check for overflow of the resulting mempool size (resulting in `NoSpaceLeft`) before adding a tx to the finger tree; this inductively ensures that the Word64s in the mempool have never overflowed
    - if the mempool capacity is specified as a Word64, then a (careful!) mempool capacity check would imply this
- keep the new `txMeasure` partiality; otherwise an invalid tx (eg of size `maxBound`, or even just ten times per-block limit) could force an indefinite `NoSpaceLeft`
- ExUnits are inherently "big", but even 100*1024 times the current block size limit is still less than 2^50.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -296,134 +290,349 @@ theLedgerLens f x =
...
+        err :: Integer -> Integer -> Integer -> SL.ApplyTxError era
+        err shift l r = maxTxSizeUTxO (l + shift) (r + shift)
+
+-- | We anachronistically use 'ConwayMeasure' in Babbage.
+instance ( ShelleyCompatible p (BabbageEra c)
+         ) => TxLimits (ShelleyBlock p (BabbageEra c)) where
+
+  type TxMeasure (ShelleyBlock p (BabbageEra c)) = ConwayMeasure
+  txMeasure              _cfg st tx = runValidation $ txMeasureBabbage st tx
+  blockCapacityTxMeasure _cfg       = blockCapacityConwayMeasure
```
nfrisby, line 631: This switches the 2.5mebi limit to 1mebi (... right?). Should be fine --- no legit block has needed that much.

amesgen: FTR this reduction for Babbage has already happened in #1168

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -18,12 +19,14 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -- | Shelley mempool integration
+--
+-- TODO nearly all of the logic in this module belongs in cardano-ledger, not
+-- ouroboros-consensus; ouroboros-consensus-cardano should just be "glue code".
```
amesgen, line 24: Seconding this; there even are potential ideas already to change which ref scripts to consider when calculating the size see https://github.com/IntersectMBO/cardano-ledger/pull/4550#issuecomment-2298275991 or https://github.com/IntersectMBO/cardano-ledger/pull/4550#issuecomment-2299748756.

nfrisby:
I just used `db-analyser --repro-mempool-and-forge` as a means of assessing the gross performance impact. I compared 33757cc13fb43d0f312eafdf45928766ff3a4606 (ie tip of this PR) against 468ec24015544318b0655deafb7a54cdc15f5d44 (ie the intersection with `main`).

For each commit, I executed `cabal build db-analyser` and then executed the following command three times.

```
$(cabal list-bin db-analyser) --db ../cardano-node/state-node-mainnet/db-mainnet/ --num-blocks-to-process 10000 --repro-mempool-and-forge 1 --analyse-from 100007913 cardano --config ../cardano-node/configuration/cardano/mainnet-config.json +RTS -s 2>&1
```

Once, I also executed the command with `0` blocks instead of `10000`, in order to measure the impact of loading the massive ledger state during initialization.

```
$ grep -e 'allocated in the heap' -- mempoolrepro1-log-*.txt
mempoolrepro1-log-33757cc13-1.txt: 909,220,855,280 bytes allocated in the heap
mempoolrepro1-log-33757cc13-2.txt: 909,248,846,488 bytes allocated in the heap
mempoolrepro1-log-33757cc13-3.txt: 909,266,331,424 bytes allocated in the heap
mempoolrepro1-log-468ec2401-0.txt: 130,718,106,536 bytes allocated in the heap
mempoolrepro1-log-468ec2401-1.txt: 904,240,758,944 bytes allocated in the heap
mempoolrepro1-log-468ec2401-2.txt: 904,242,716,272 bytes allocated in the heap
mempoolrepro1-log-468ec2401-3.txt: 904,240,554,720 bytes allocated in the heap
```

```
$ grep -e 'GC *time' -- mempoolrepro1-log-*.txt
mempoolrepro1-log-33757cc13-1.txt:  GC      time   58.599s  ( 50.892s elapsed)
mempoolrepro1-log-33757cc13-2.txt:  GC      time   59.633s  ( 47.981s elapsed)
mempoolrepro1-log-33757cc13-3.txt:  GC      time   59.193s  ( 47.469s elapsed)
mempoolrepro1-log-468ec2401-0.txt:  GC      time   17.906s  ( 15.275s elapsed)
mempoolrepro1-log-468ec2401-1.txt:  GC      time   59.118s  ( 48.854s elapsed)
mempoolrepro1-log-468ec2401-2.txt:  GC      time   60.646s  ( 48.179s elapsed)
mempoolrepro1-log-468ec2401-3.txt:  GC      time   59.318s  ( 47.518s elapsed)

$ grep -e 'MUT *time' -- mempoolrepro1-log-*.txt
mempoolrepro1-log-33757cc13-1.txt:  MUT     time  261.259s  (261.831s elapsed)
mempoolrepro1-log-33757cc13-2.txt:  MUT     time  262.880s  (263.619s elapsed)
mempoolrepro1-log-33757cc13-3.txt:  MUT     time  258.526s  (259.206s elapsed)
mempoolrepro1-log-468ec2401-0.txt:  MUT     time   39.908s  ( 39.972s elapsed)
mempoolrepro1-log-468ec2401-1.txt:  MUT     time  260.084s  (260.960s elapsed)
mempoolrepro1-log-468ec2401-2.txt:  MUT     time  262.230s  (262.911s elapsed)
mempoolrepro1-log-468ec2401-3.txt:  MUT     time  257.984s  (258.796s elapsed)

$ grep -e 'Total *time' -- mempoolrepro1-log-*.txt
mempoolrepro1-log-33757cc13-1.txt:  Total   time  320.294s  (312.730s elapsed)
mempoolrepro1-log-33757cc13-2.txt:  Total   time  322.887s  (311.601s elapsed)
mempoolrepro1-log-33757cc13-3.txt:  Total   time  318.094s  (306.676s elapsed)
mempoolrepro1-log-468ec2401-0.txt:  Total   time   58.161s  ( 55.250s elapsed)
mempoolrepro1-log-468ec2401-1.txt:  Total   time  319.579s  (309.820s elapsed)
mempoolrepro1-log-468ec2401-2.txt:  Total   time  322.878s  (311.100s elapsed)
mempoolrepro1-log-468ec2401-3.txt:  Total   time  317.681s  (306.320s elapsed)
```

Conclusion: the PR's diff is inconsequential with respect to the Cardano node's performance. The run time is apparently unaffected, and the allocation rate increases by less than 1% --- 1% is already very low, but note that this tool excludes everything _else_ a running node would be doing, so the same absolute extra allocation would be a much lower proportion of a real node's allocation rate.

The eventual cluster run's more careful analysis may still find regressions (contradicting this ad-hoc analysis), but this due diligence already justifies merging the PR, with optimistic that the cluster run won't find any significant performance regressions.

nfrisby:
@fraser-iohk I know you had some work related to this PR. Be advised: I'm going to rebase it today. If you're not ready to deal with a rebase, then commit 3661dcce2 should be equivalent to what I'm pushing up in the near future.

As a reminder: the PR now uses Word32 for the byte size, so the encoder challenges you were facing should have been eliminated.

nfrisby:
I haven't yet reorganized the commits into a nice sequence yet, but review can commence --- I will ensure the final diff matches that of the current tip (ie fe778a33a1fa7b1c0492efcba13800630a79ca45)

nfrisby:
I'm now attempting to split the monolithic commit. I consider it to have the following components.

- forging thread no longer splits the given list, see `snapshotTake` instead

- refactor the `TxLimits` class and `CanHardFork`

- removed the Babbage mempool ref scripts hotfix

- remove the block size override

- multi-dimensional tx sizes in the mempool itself, careful with overflow, and supplanting the `txSize` argument

I'm hoping I can tease out a separate commit for at least a few of those.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -166,3 +153,124 @@ type GenTxId blk = TxId (GenTx blk)
...
+-- | Each block has its limits of how many transactions it can hold. That limit
+-- is compared against the sum of measurements taken of each of the
+-- transactions in that block.
+--
+-- How we measure the transaction depends of the era that this transaction
+-- belongs to (more specifically it depends on the block type to which this
+-- transaction will be added). For initial eras (like Byron and initial
+-- generations of Shelley based eras) this measure was simply a byte size
+-- (block could not be bigger then given size - in bytes - specified by the
+-- ledger state). In future eras (starting with Alonzo) this measure was a bit
```
amesgen, line 170:
```suggestion
-- ledger state). In subsequent eras (starting with Alonzo) this measure was a bit
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -166,3 +153,124 @@ type GenTxId blk = TxId (GenTx blk)
...
+  -- across the network, requiring a few extra bytes compared to the actual
+  -- in-block serialisation. Another example is the transaction of the
+  -- hard-fork combinator which will include an envelope indicating its era
+  -- when sent across the network. However, when embedded in the respective
+  -- era's block, there is no need for such envelope. An example from upstream
+  -- is that the Cardano ledger's "Segregated Witness" encoding scheme
+  -- contributes to the encoding overhead.
+  --
+  -- INVARIANT Assuming no hash collisions, the size should be the same in any
+  -- state in which the transaction is valid.
```
amesgen, line 199: Add reference scripts as an example where the condition "in any state in which the transaction is valid" becomes relevant?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -166,3 +153,124 @@ type GenTxId blk = TxId (GenTx blk)
...
+  -- | What is the allowed capacity for the txs in an individual block?
+  blockCapacityTxMeasure ::
+       LedgerConfig blk
+       -- ^ at least for symmetry with 'txMeasure'
+    -> TickedLedgerState blk
+    -> TxMeasure blk
+
+-- | We intentionally do not declare a 'Num' instance! We prefer @ByteSize32@
+-- to occur explicitly in the code where possible, for
+-- legibility/perspiciousness. We also do not need nor want subtraction.
```
amesgen, line 219: I had to look this up, and apparently it is spelled
```suggestion
-- legibility/perspicuousness. We also do not need nor want subtraction.
```
:smile:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/TxSeq.hs
```diff
@@ -200,51 +222,53 @@ splitAfterTxSize (TxSeq txs) n =
...
+       -> [TxTicket sz tx]
+       -> [TxTicket sz tx]
+       -> ([TxTicket sz tx], [TxTicket sz tx])
     go accByteSize accTickets = \case
       []
         -> (reverse accTickets, [])
       t:ts
-        | let accByteSize' = accByteSize + txTicketTxSizeInBytes t
-        , accByteSize' <= n
+        | let accByteSize' = accByteSize `Measure.plus` txTicketSize t
```
amesgen, line 244: The `Byte` in `accByteSize` seems outdated. Also, `go` could be removed by using [`Measure.splitAt`](https://intersectmbo.github.io/cardano-base/measures/Data-Measure.html#v:splitAt).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -172,28 +168,84 @@ pureTryAddTx ::
...
+      -- or every component is not full) lead to some confusing behaviors
+      -- (denying some txs that would "obviously" fit and accepting some txs
+      -- that "obviously" don't, respectively).
+      --
+      -- Even with the overflow handler, it's important that 'txMeasure'
+      -- returns a well-bounded result. Otherwise, if an adversarial tx arrived
+      -- that could't even fit in an empty mempool, then that thread would
+      -- never release the 'MVar'. In particular, we tacitly assume here that a
+      -- tx that wouldn't even fit in an empty mempool would be rejected by
+      -- 'txMeasure'.
```
amesgen, line 241: Also mention this in the haddocks of `txMeasure`?

amesgen (approved):
Excellent :tada:

Only minor comments

ouroboros-consensus/bench/mempool-bench/Main.hs
amesgen: Superseded by #1227

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/TxSeq.hs
nfrisby: I fixed the name, but left the "specification" function as transparent as possible 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
```diff
@@ -105,25 +109,89 @@ instance CanHardFork xs => LedgerSupportsMempool (HardForkBlock xs) where
...
+        -> K (HardForkTxMeasure xs) blk
+      aux idx pcfg st' =
+          K
+        $ hardForkInjTxMeasure . injectNS idx . WrapTxMeasure
+        $ blockCapacityTxMeasure
+            (completeLedgerConfig' ei pcfg)
+            (unComp st')
+
+  -- TODO should the HFC try translating the tx instead of always relying on
+  -- 'hardForkInjTxMeasure'?
```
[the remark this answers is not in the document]
nfrisby: I had been thinking of `HardForTxMeasure xs = TxMeasure (Last xs)`, but trying to improve the wording of my comment revealed to me that we'd have to also translate the ledger state to that era, and doing so might _require_ those state translation functions to permit empty eras, which seems unwise to me.

So I'm removing this TODO. Thanks for considering it.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -288,101 +290,352 @@ theLedgerLens f x =
...
+
+    -- TODO recall the assertion in this module header that this logic should
+    -- be owned by Ledger, not by Consensus: Ledger prefers to delete these
+    -- checks after mainnet hard forks to Conway, rather than introduce the
+    -- necessary 'SL.PredicateFailure' constructors for Babbage (which would in
+    -- turn require a bump to "Ouroboros.Network.NodeToNode.Version" and
+    -- "Ouroboros.Network.NodeToClient.Version")
+    --
+    -- -- Babbage should have enforced a per-tx limit here, but did not.
+    -- refScriptBytes = pure $ ByteSize $ fromIntegral txsz
```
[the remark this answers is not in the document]
nfrisby: I'll add a new commit that does so.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
nfrisby: Great suggestion. I added an INVARIANT.

nfrisby:
Some collisions with `ouroboros-network` (they made `SizeInBytes` a `newtype` instead of a `type` synonym) during the rebase, but easy enough (disclaimer: I haven't ran the tests locally).

nfrisby:
The `mempool-bench` is failing.

```
Just adding
  10000 transactions
    setup mempool:             FAIL (0.25s)
      1.75 μs ± 153 ns, 6.2 KB allocated,   3 B  copied, 5.0 MB peak memory, 231% more than baseline
      Use -p '$0=="Just adding.10000 transactions.setup mempool"' to rerun this test only.
    setup mempool + benchmark: OK (0.37s)
      24.4 ms ± 741 μs,  68 MB allocated, 4.0 MB copied,  11 MB peak memory, 31% more than baseline
    test:                      OK (0.03s)
    cmds length:               OK
  20000 transactions
    setup mempool:             FAIL (0.28s)
      1.80 μs ±  99 ns, 6.4 KB allocated,   6 B  copied,  13 MB peak memory, 239% more than baseline
      Use -p '$0=="Just adding.20000 transactions.setup mempool"' to rerun this test only.
    setup mempool + benchmark: OK (0.15s)
      48.7 ms ± 3.0 ms, 141 MB allocated, 5.9 MB copied,  15 MB peak memory, 25% more than baseline
    test:                      OK (0.06s)
    cmds length:               OK
```

Joris helpfully shared the numbers from the latest baseline file:

```
Just adding
  10000 transactions
    setup mempool:             OK (0.16s)
      528  ns ±  42 ns, 2.7 KB allocated,   1 B  copied, 5.0 MB peak memory
    setup mempool + benchmark: OK (0.28s)
      18.5 ms ± 1.1 ms,  57 MB allocated, 3.4 MB copied, 9.0 MB peak memory
    test:                      OK (0.02s)
    cmds length:               OK
  20000 transactions
    setup mempool:             OK (0.19s)
      530  ns ±  46 ns, 2.7 KB allocated,   1 B  copied,  12 MB peak memory
    setup mempool + benchmark: OK (1.22s)
      38.6 ms ± 521 μs, 119 MB allocated, 7.6 MB copied,  24 MB peak memory
    test:                      OK (0.03s)
    cmds length:               OK```
    
The failing case is the setup step, which increases from 500 nanoseconds to 1750 nanoseconds.

It is very plausible that the additional calculation in `computeMempoolCapacity` explains this: it use to be `* 2` at known type `Word32` and is now a call to some small amount of arithmetic and `stimes` involving a couple dictionaries. It's harmless for the behavior of real node (setup happens once per execution, and it's still microseconds).

So I'm merging despite these failures --- they are not required "checks" and this benchmark is the _only_ reasons they fail.

PR #1179, opened 2024-07-03
Full rework of the BlockFetch logic for bulk sync mode

Niols:
Integrates a new implementation of the BulkSync mode, where blocks are downloaded from alternative peers as soon as the node has no more blocks to validate while there are longstanding requests in flight.

This PR depends on the new implementation of the BulkSync mode (https://github.com/IntersectMBO/ouroboros-network/pull/4919). `cabal.project` is made to point to a back-port of the BulkSync implementation on `ouroboros-network-0.16.1.1`.

### CSJ Changes

CSJ is involved because the new BulkSync mode requires to change the dynamo if it is also serving blocks, and it is not sending them promptly enough. The dynamo choice has an influence in the blocks that are chosen to be downloaded by BlockFetch.

For this sake, b93c379 gives the ability to order the ChainSync clients, so the dynamo role can be rotated among them whenever BlockFetch requests it.

b1c0bf8 provides the implementation of the rotation operation.

### BlockFetch tests

c4bfa37 allows to specify in tests in which order to start the peers, which has an effect on what peer is chosen as initial dynamo.

c594c09 in turn adds a new BlockFetch test to show that syncing isn't slowed down by peers that don't send blocks.

### Integration of BlockFetch changes

The collection of ChainSync client handles now needs to be passed between BlockFetch and ChainSync so dynamo rotations can be requested by BlockFetch.

The parameter `bfcMaxConcurrencyBulkSync` has been removed since blocks are not coordinated to be downloaded concurrently.

These changes are in 6926278.
 
### ChainSel changes

Now BlockFetch requires the ability to detect if ChainSel has run out of blocks to validate. This motivates 73187ba, which implements a mechanism to measure if ChainSel is waiting for more blocks (starves), and determines for how long.

The above change is not sufficient to measure starvation. The queue to send blocks for validation used to allow only for one block to sit in the queue. This would interfere with the ability to measure starvation since BlockFetch would block waiting for the queue to become empty, and the queue would quickly become empty after taking just 1 block. For download delays to be amortized, a larger queue capacity was needed. This is the reason why a fix similar to https://github.com/IntersectMBO/ouroboros-network/pull/2721 is part of 0d3fc28.

#### CSJ jump size adjustment

When syncing from mainnet, we discovered that CSJ wouldn't sync the blocks from the Byron era. This was because the jump size was set to the length of the genesis window of the Shelley era, which is much larger than Byron's. When the jump size is larger than the genesis window, the dynamo will block on the forecast horizon before offering a jump that allows the chain selection to advance. In this case, CSJ and chain selection will deadlock.

For this reason we set the default jump size to the size of Byron's genesis window in 028883a. This didn't show an impact on syncing time in our measures. Future work (as part of deploying Genesis) might involve allowing the jump size to vary between different eras.

#### GDD rate limit

GDD evaluation showed an overhead of 10% if run after every header arrives via ChainSync. Therefore, in b7fa122 we limited how often it could run, so multiple header arrivals could be handled by a single GDD evaluation.

#### Candidate fragment comparison in the ChainSync client

We stumbled upon a test case where the candidate fragments of the dynamo and an objector were no longer than the current selection (both peers were adversarial). This was problematic because BlockFetch would refuse to download blocks from these candidates, and ChainSync in turn would wait for the selection to advance in order to download more headers.

The fix in e27a73c is to have the ChainSync client disconnect a peer which is about to block on the forecast horizon if its candidate isn't better than the selection.

#### Candidate fragment truncations

At the moment, it is possible for a candidate fragment to be truncated by CSJ when a jumper jumps to a point that is not younger than the tip of its current candidate fragment. We encountered tests where the jump point could be so old that it would fall behind the immutable tip, and GDD would ignore the peer when computing the Limit on Eagerness. This in turn would cause the selection to advance into potentially adversarial chains.

The fix in dc5f6f7 is to have GDD never drop candidates. When the candidate does not intersect the current selection, the LoE is not advanced. This is a situation guaranteed to be unblocked by the ChainSync client since it will either disconnect the peer or bring the candidate to intersect with the current selection.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
@@ -18,6 +18,7 @@ module Ouroboros.Consensus.Node.Genesis (
   , mkGenesisConfig
     -- * NodeKernel helpers
   , GenesisNodeKernelArgs (..)
+  , LoEAndGDDNodeKernelArgs (..)
   , mkGenesisNodeKernelArgs
   , setGetLoEFragment
   ) where
@@ -59,7 +60,7 @@ data GenesisConfig = GenesisConfig
   { gcBlockFetchConfig         :: !GenesisBlockFetchConfiguration
   , gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
   , gcCSJConfig                :: !CSJConfig
-  , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig ())
+  , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig LoEAndGDDParams)
   , gcHistoricityCutoff        :: !(Maybe HistoricityCutoff)
   } deriving stock (Eq, Generic, Show)
 
@@ -72,6 +73,7 @@ data GenesisConfigFlags = GenesisConfigFlags
   , gcfBucketCapacity      :: Maybe Integer
   , gcfBucketRate          :: Maybe Integer
   , gcfCSJJumpSize         :: Maybe Integer
+  , gcfGDDRateLimit        :: Maybe DiffTime
   } deriving stock (Eq, Generic, Show)
 
 defaultGenesisConfigFlags :: GenesisConfigFlags
@@ -83,6 +85,7 @@ defaultGenesisConfigFlags = GenesisConfigFlags
   , gcfBucketCapacity       = Nothing
   , gcfBucketRate           = Nothing
   , gcfCSJJumpSize          = Nothing
+  , gcfGDDRateLimit         = Nothing
   }
 
 enableGenesisConfigDefault :: GenesisConfig
@@ -120,7 +123,7 @@ mkGenesisConfig (Just GenesisConfigFlags{..}) =
           }
         else CSJDisabled
     , gcLoEAndGDDConfig = if gcfEnableLoEAndGDD
-        then LoEAndGDDEnabled ()
+        then LoEAndGDDEnabled LoEAndGDDParams{lgpGDDRateLimit}
         else LoEAndGDDDisabled
     , -- Duration in seconds of one Cardano mainnet Shelley stability window
       -- (3k/f slots times one second per slot) plus one extra hour as a
@@ -147,20 +150,35 @@ mkGenesisConfig (Just GenesisConfigFlags{..}) =
     -- eras.
     defaultCSJJumpSize = 2 * 2160 -- Byron forecast range
 
+    -- Limiting the performance impact of the GDD.
+    defaultGDDRateLimit        = 1.0 -- seconds
+
     gbfcBulkSyncGracePeriod = fromInteger $ fromMaybe defaultBulkSyncGracePeriod gcfBulkSyncGracePeriod
     csbcCapacity            = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
     csbcRate                = fromInteger $ fromMaybe defaultRate gcfBucketRate
     csjcJumpSize            = fromInteger $ fromMaybe defaultCSJJumpSize gcfCSJJumpSize
+    lgpGDDRateLimit            = fromMaybe defaultGDDRateLimit gcfGDDRateLimit
+
+newtype LoEAndGDDParams = LoEAndGDDParams
+  { -- | How often to evaluate GDD. 0 means as soon as possible.
+    -- Otherwise, no faster than once every T seconds, where T is the
+    -- value of the field.
+    lgpGDDRateLimit :: DiffTime
+  } deriving stock (Eq, Generic, Show)
 
 -- | Genesis-related arguments needed by the NodeKernel initialization logic.
 data GenesisNodeKernelArgs m blk = GenesisNodeKernelArgs {
+    gnkaLoEAndGDDArgs :: !(LoEAndGDDConfig (LoEAndGDDNodeKernelArgs m blk))
+  }
+
+data LoEAndGDDNodeKernelArgs m blk = LoEAndGDDNodeKernelArgs {
     -- | A TVar containing an action that returns the 'ChainDB.GetLoEFragment'
     -- action. We use this extra indirection to update this action after we
     -- opened the ChainDB (which happens before we initialize the NodeKernel).
     -- After that, this TVar will not be modified again.
-    gnkaGetLoEFragment :: !(LoEAndGDDConfig (StrictTVar m (ChainDB.GetLoEFragment m blk)))
+    lgnkaLoEFragmentTVar :: !(StrictTVar m (ChainDB.GetLoEFragment m blk))
+  , lgnkaGDDRateLimit    :: DiffTime
   }
-
 -- | Create the initial 'GenesisNodeKernelArgs" (with a temporary
 -- 'ChainDB.GetLoEFragment' that will be replaced via 'setGetLoEFragment') and a
 -- function to update the 'ChainDbArgs' accordingly.
@@ -171,20 +189,24 @@ mkGenesisNodeKernelArgs ::
        , Complete ChainDbArgs m blk -> Complete ChainDbArgs m blk
        )
 mkGenesisNodeKernelArgs gcfg = do
-    gnkaGetLoEFragment <- for (gcLoEAndGDDConfig gcfg) $ \() ->
-        newTVarIO $ pure $
+    gnkaLoEAndGDDArgs <- for (gcLoEAndGDDConfig gcfg) $ \p -> do
+        loeFragmentTVar <- newTVarIO $ pure $
           -- Use the most conservative LoE fragment until 'setGetLoEFragment'
           -- is called.
           ChainDB.LoEEnabled $ AF.Empty AF.AnchorGenesis
-    let updateChainDbArgs = case gnkaGetLoEFragment of
+        pure LoEAndGDDNodeKernelArgs
+          { lgnkaLoEFragmentTVar = loeFragmentTVar
+          , lgnkaGDDRateLimit = lgpGDDRateLimit p
+          }
+    let updateChainDbArgs = case gnkaLoEAndGDDArgs of
           LoEAndGDDDisabled -> id
-          LoEAndGDDEnabled varGetLoEFragment -> \cfg ->
+          LoEAndGDDEnabled lgnkArgs -> \cfg ->
             cfg { ChainDB.cdbsArgs =
                   (ChainDB.cdbsArgs cfg) { ChainDB.cdbsLoE = getLoEFragment }
                 }
             where
-              getLoEFragment = join $ readTVarIO varGetLoEFragment
-    pure (GenesisNodeKernelArgs {gnkaGetLoEFragment}, updateChainDbArgs)
+              getLoEFragment = join $ readTVarIO $ lgnkaLoEFragmentTVar lgnkArgs
+    pure (GenesisNodeKernelArgs{gnkaLoEAndGDDArgs}, updateChainDbArgs)
 
 -- | Set 'gnkaGetLoEFragment' to the actual logic for determining the current
 -- LoE fragment.
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
@@ -66,26 +66,26 @@ data GenesisConfig = GenesisConfig
 
 -- | Genesis configuration flags and low-level args, as parsed from config file or CLI
 data GenesisConfigFlags = GenesisConfigFlags
-  { gcfEnableCSJ           :: Bool
-  , gcfEnableLoEAndGDD     :: Bool
-  , gcfEnableLoP           :: Bool
-  , gcfBulkSyncGracePeriod :: Maybe Integer
-  , gcfBucketCapacity      :: Maybe Integer
-  , gcfBucketRate          :: Maybe Integer
-  , gcfCSJJumpSize         :: Maybe Integer
-  , gcfGDDRateLimit        :: Maybe DiffTime
+  { gcfEnableCSJ             :: Bool
+  , gcfEnableLoEAndGDD       :: Bool
+  , gcfEnableLoP             :: Bool
+  , gcfBlockFetchGracePeriod :: Maybe Integer
+  , gcfBucketCapacity        :: Maybe Integer
+  , gcfBucketRate            :: Maybe Integer
+  , gcfCSJJumpSize           :: Maybe Integer
+  , gcfGDDRateLimit          :: Maybe DiffTime
   } deriving stock (Eq, Generic, Show)
 
 defaultGenesisConfigFlags :: GenesisConfigFlags
 defaultGenesisConfigFlags = GenesisConfigFlags
-  { gcfEnableCSJ            = True
-  , gcfEnableLoEAndGDD      = True
-  , gcfEnableLoP            = True
-  , gcfBulkSyncGracePeriod  = Nothing
-  , gcfBucketCapacity       = Nothing
-  , gcfBucketRate           = Nothing
-  , gcfCSJJumpSize          = Nothing
-  , gcfGDDRateLimit         = Nothing
+  { gcfEnableCSJ              = True
+  , gcfEnableLoEAndGDD        = True
+  , gcfEnableLoP              = True
+  , gcfBlockFetchGracePeriod  = Nothing
+  , gcfBucketCapacity         = Nothing
+  , gcfBucketRate             = Nothing
+  , gcfCSJJumpSize            = Nothing
+  , gcfGDDRateLimit           = Nothing
   }
 
 enableGenesisConfigDefault :: GenesisConfig
@@ -99,7 +99,7 @@ mkGenesisConfig :: Maybe GenesisConfigFlags -> GenesisConfig
 mkGenesisConfig Nothing = -- disable Genesis
   GenesisConfig
     { gcBlockFetchConfig = GenesisBlockFetchConfiguration
-        { gbfcBulkSyncGracePeriod = 0 -- no grace period when Genesis is disabled
+        { gbfcGracePeriod = 0 -- no grace period when Genesis is disabled
         }
     , gcChainSyncLoPBucketConfig = ChainSyncLoPBucketDisabled
     , gcCSJConfig                = CSJDisabled
@@ -109,7 +109,7 @@ mkGenesisConfig Nothing = -- disable Genesis
 mkGenesisConfig (Just GenesisConfigFlags{..}) =
   GenesisConfig
     { gcBlockFetchConfig = GenesisBlockFetchConfiguration
-        { gbfcBulkSyncGracePeriod
+        { gbfcGracePeriod
         }
     , gcChainSyncLoPBucketConfig = if gcfEnableLoP
         then ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig
@@ -134,7 +134,7 @@ mkGenesisConfig (Just GenesisConfigFlags{..}) =
     -- The minimum amount of time during which the Genesis BlockFetch logic will
     -- download blocks from a specific peer (even if it is not performing well
     -- during that period).
-    defaultBulkSyncGracePeriod = 10 -- seconds
+    defaultBlockFetchGracePeriod = 10 -- seconds
 
     -- LoP parameters. Empirically, it takes less than 1ms to validate a header,
     -- so leaking one token per 2ms is conservative. The capacity of 100_000
@@ -153,11 +153,11 @@ mkGenesisConfig (Just GenesisConfigFlags{..}) =
     -- Limiting the performance impact of the GDD.
     defaultGDDRateLimit        = 1.0 -- seconds
 
-    gbfcBulkSyncGracePeriod = fromInteger $ fromMaybe defaultBulkSyncGracePeriod gcfBulkSyncGracePeriod
-    csbcCapacity            = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
-    csbcRate                = fromInteger $ fromMaybe defaultRate gcfBucketRate
-    csjcJumpSize            = fromInteger $ fromMaybe defaultCSJJumpSize gcfCSJJumpSize
-    lgpGDDRateLimit            = fromMaybe defaultGDDRateLimit gcfGDDRateLimit
+    gbfcGracePeriod = fromInteger $ fromMaybe defaultBlockFetchGracePeriod gcfBlockFetchGracePeriod
+    csbcCapacity    = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
+    csbcRate        = fromInteger $ fromMaybe defaultRate gcfBucketRate
+    csjcJumpSize    = fromInteger $ fromMaybe defaultCSJJumpSize gcfCSJJumpSize
+    lgpGDDRateLimit = fromMaybe defaultGDDRateLimit gcfGDDRateLimit
 
 newtype LoEAndGDDParams = LoEAndGDDParams
   { -- | How often to evaluate GDD. 0 means as soon as possible.
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
@@ -140,7 +140,12 @@ mkGenesisConfig (Just GenesisConfigFlags{..}) =
     defaultCapacity = 100_000 -- number of tokens
     defaultRate     = 500 -- tokens per second leaking, 1/2ms
 
-    defaultCSJJumpSize = 3 * 2160 * 20 -- mainnet forecast range
+    -- The larger Shelley forecast range (3 * 2160 * 20) works in more recent
+    -- ranges of slots, but causes syncing to block in Byron. A future
+    -- improvement would be to make this era-dynamic, such that we can use the
+    -- larger (and hence more efficient) larger CSJ jump size in Shelley-based
+    -- eras.
+    defaultCSJJumpSize = 2 * 2160 -- Byron forecast range
 
     gbfcBulkSyncGracePeriod = fromInteger $ fromMaybe defaultBulkSyncGracePeriod gcfBulkSyncGracePeriod
     csbcCapacity            = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
@@ -1,16 +1,21 @@
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Ouroboros.Consensus.Node.Genesis (
     -- * 'GenesisConfig'
     GenesisConfig (..)
+  , GenesisConfigFlags (..)
   , LoEAndGDDConfig (..)
+  , defaultGenesisConfigFlags
   , disableGenesisConfig
   , enableGenesisConfigDefault
+  , mkGenesisConfig
     -- * NodeKernel helpers
   , GenesisNodeKernelArgs (..)
   , mkGenesisNodeKernelArgs
@@ -18,7 +23,9 @@ module Ouroboros.Consensus.Node.Genesis (
   ) where
 
 import           Control.Monad (join)
+import           Data.Maybe (fromMaybe)
 import           Data.Traversable (for)
+import           GHC.Generics (Generic)
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
                      (CSJConfig (..), CSJEnabledConfig (..),
@@ -34,47 +41,111 @@ import           Ouroboros.Consensus.Util.Args
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import qualified Ouroboros.Network.AnchoredFragment as AF
+import           Ouroboros.Network.BlockFetch
+                     (GenesisBlockFetchConfiguration (..))
 
 -- | Whether to en-/disable the Limit on Eagerness and the Genesis Density
 -- Disconnector.
 data LoEAndGDDConfig a =
     LoEAndGDDEnabled !a
   | LoEAndGDDDisabled
-  deriving stock (Show, Functor, Foldable, Traversable)
+  deriving stock (Eq, Generic, Show, Functor, Foldable, Traversable)
 
 -- | Aggregating the various configs for Genesis-related subcomponents.
-data GenesisConfig = GenesisConfig {
-    gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
+--
+-- Usually, 'enableGenesisConfigDefault' or 'disableGenesisConfig' can be used.
+-- See the haddocks of the types of the individual fields for details.
+data GenesisConfig = GenesisConfig
+  { gcBlockFetchConfig         :: !GenesisBlockFetchConfiguration
+  , gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
   , gcCSJConfig                :: !CSJConfig
   , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig ())
   , gcHistoricityCutoff        :: !(Maybe HistoricityCutoff)
+  } deriving stock (Eq, Generic, Show)
+
+-- | Genesis configuration flags and low-level args, as parsed from config file or CLI
+data GenesisConfigFlags = GenesisConfigFlags
+  { gcfEnableCSJ           :: Bool
+  , gcfEnableLoEAndGDD     :: Bool
+  , gcfEnableLoP           :: Bool
+  , gcfBulkSyncGracePeriod :: Maybe Integer
+  , gcfBucketCapacity      :: Maybe Integer
+  , gcfBucketRate          :: Maybe Integer
+  , gcfCSJJumpSize         :: Maybe Integer
+  } deriving stock (Eq, Generic, Show)
+
+defaultGenesisConfigFlags :: GenesisConfigFlags
+defaultGenesisConfigFlags = GenesisConfigFlags
+  { gcfEnableCSJ            = True
+  , gcfEnableLoEAndGDD      = True
+  , gcfEnableLoP            = True
+  , gcfBulkSyncGracePeriod  = Nothing
+  , gcfBucketCapacity       = Nothing
+  , gcfBucketRate           = Nothing
+  , gcfCSJJumpSize          = Nothing
   }
 
--- TODO justification/derivation from other parameters
 enableGenesisConfigDefault :: GenesisConfig
-enableGenesisConfigDefault = GenesisConfig {
-      gcChainSyncLoPBucketConfig = ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig {
-          csbcCapacity = 100_000 -- number of tokens
-        , csbcRate     = 500 -- tokens per second leaking, 1/2ms
-        }
-    , gcCSJConfig = CSJEnabled CSJEnabledConfig {
-          csjcJumpSize = 3 * 2160 * 20 -- mainnet forecast range
-        }
-    , gcLoEAndGDDConfig = LoEAndGDDEnabled ()
-      -- Duration in seconds of one Cardano mainnet Shelley stability window
-      -- (3k/f slots times one second per slot) plus one extra hour as a
-      -- safety margin.
-    , gcHistoricityCutoff = Just $ HistoricityCutoff $ 3 * 2160 * 20 + 3600
-    }
+enableGenesisConfigDefault = mkGenesisConfig $ Just defaultGenesisConfigFlags
 
 -- | Disable all Genesis components, yielding Praos behavior.
 disableGenesisConfig :: GenesisConfig
-disableGenesisConfig = GenesisConfig {
-      gcChainSyncLoPBucketConfig = ChainSyncLoPBucketDisabled
+disableGenesisConfig = mkGenesisConfig Nothing
+
+mkGenesisConfig :: Maybe GenesisConfigFlags -> GenesisConfig
+mkGenesisConfig Nothing = -- disable Genesis
+  GenesisConfig
+    { gcBlockFetchConfig = GenesisBlockFetchConfiguration
+        { gbfcBulkSyncGracePeriod = 0 -- no grace period when Genesis is disabled
+        }
+    , gcChainSyncLoPBucketConfig = ChainSyncLoPBucketDisabled
     , gcCSJConfig                = CSJDisabled
     , gcLoEAndGDDConfig          = LoEAndGDDDisabled
     , gcHistoricityCutoff        = Nothing
     }
+mkGenesisConfig (Just GenesisConfigFlags{..}) =
+  GenesisConfig
+    { gcBlockFetchConfig = GenesisBlockFetchConfiguration
+        { gbfcBulkSyncGracePeriod
+        }
+    , gcChainSyncLoPBucketConfig = if gcfEnableLoP
+        then ChainSyncLoPBucketEnabled ChainSyncLoPBucketEnabledConfig
+          { csbcCapacity
+          , csbcRate
+          }
+        else ChainSyncLoPBucketDisabled
+    , gcCSJConfig = if gcfEnableCSJ
+        then CSJEnabled CSJEnabledConfig
+          { csjcJumpSize
+          }
+        else CSJDisabled
+    , gcLoEAndGDDConfig = if gcfEnableLoEAndGDD
+        then LoEAndGDDEnabled ()
+        else LoEAndGDDDisabled
+    , -- Duration in seconds of one Cardano mainnet Shelley stability window
+      -- (3k/f slots times one second per slot) plus one extra hour as a
+      -- safety margin.
+      gcHistoricityCutoff = Just $ HistoricityCutoff $ 3 * 2160 * 20 + 3600
+    }
+  where
+    -- The minimum amount of time during which the Genesis BlockFetch logic will
+    -- download blocks from a specific peer (even if it is not performing well
+    -- during that period).
+    defaultBulkSyncGracePeriod = 10 -- seconds
+
+    -- LoP parameters. Empirically, it takes less than 1ms to validate a header,
+    -- so leaking one token per 2ms is conservative. The capacity of 100_000
+    -- tokens corresponds to 200s, which is definitely enough to handle long GC
+    -- pauses; we could even make this more conservative.
+    defaultCapacity = 100_000 -- number of tokens
+    defaultRate     = 500 -- tokens per second leaking, 1/2ms
+
+    defaultCSJJumpSize = 3 * 2160 * 20 -- mainnet forecast range
+
+    gbfcBulkSyncGracePeriod = fromInteger $ fromMaybe defaultBulkSyncGracePeriod gcfBulkSyncGracePeriod
+    csbcCapacity            = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
+    csbcRate                = fromInteger $ fromMaybe defaultRate gcfBucketRate
+    csjcJumpSize            = fromInteger $ fromMaybe defaultCSJJumpSize gcfCSJJumpSize
 
 -- | Genesis-related arguments needed by the NodeKernel initialization logic.
 data GenesisNodeKernelArgs m blk = GenesisNodeKernelArgs {
@@ -124,9 +195,10 @@ setGetLoEFragment readGsmState readLoEFragment varGetLoEFragment =
   where
     getLoEFragment :: ChainDB.GetLoEFragment m blk
     getLoEFragment = atomically $ readGsmState >>= \case
-        -- When the Honest Availability Assumption cannot currently be guaranteed, we should not select
-        -- any blocks that would cause our immutable tip to advance, so we
-        -- return the most conservative LoE fragment.
+        -- When the Honest Availability Assumption cannot currently be
+        -- guaranteed, we should not select any blocks that would cause our
+        -- immutable tip to advance, so we return the most conservative LoE
+        -- fragment.
         GSM.PreSyncing ->
           pure $ ChainDB.LoEEnabled $ AF.Empty AF.AnchorGenesis
         -- When we are syncing, return the current LoE fragment.
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -188,6 +188,12 @@ chainSyncTimeouts =
     -- chains are finite, and therefore an honest peer can only serve it all,
     -- then send 'MsgAwaitReply' (therefore entering 'StMustReply'), and then
     -- stall forever, and it must not be killed for it.
+    --
+    -- Note that this allows the adversaries to stall us forever in that same
+    -- situation. However, that peer is only allowed to send 'MsgAwaitReply'
+    -- when they have served their tip, which leaves them fully vulnerable to
+    -- the Genesis Density Disconnection (GDD) logic. A bug related to this
+    -- disabled timeout is in fact either a bug in the GDD or in the tests.
     mustReplyTimeout :: Maybe DiffTime
     mustReplyTimeout = Nothing
 
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
@@ -10,13 +10,12 @@ module Test.Consensus.Genesis.Setup.GenChains (
   , genChainsWithExtraHonestPeers
   ) where
 
-import           Cardano.Slotting.Time (SlotLength, getSlotLength,
-                     slotLengthFromSec)
+import           Cardano.Slotting.Time (slotLengthFromSec)
 import           Control.Monad (replicateM)
 import qualified Control.Monad.Except as Exn
 import           Data.List as List (foldl')
 import           Data.Proxy (Proxy (..))
-import           Data.Time.Clock (DiffTime, secondsToDiffTime)
+import           Data.Time.Clock (DiffTime)
 import qualified Data.Vector.Unboxed as Vector
 import           Data.Word (Word8)
 import           Ouroboros.Consensus.Block.Abstract hiding (Header)
@@ -110,7 +109,7 @@ genChains  = genChainsWithExtraHonestPeers (pure 0)
 -- However, in the future it could also be used to generate "short forks" near the tip of the trunk.
 genChainsWithExtraHonestPeers :: QC.Gen Word -> QC.Gen Word -> QC.Gen (GenesisTest TestBlock ())
 genChainsWithExtraHonestPeers genNumExtraHonest genNumForks = do
-  (asc, honestRecipe, someHonestChainSchema) <- genHonestChainSchema
+  (_, honestRecipe, someHonestChainSchema) <- genHonestChainSchema
 
   H.SomeHonestChainSchema _ _ honestChainSchema <- pure someHonestChainSchema
   let ChainSchema _ vH = honestChainSchema
@@ -128,8 +127,8 @@ genChainsWithExtraHonestPeers genNumExtraHonest genNumForks = do
     gtGenesisWindow = GenesisWindow (fromIntegral scg),
     gtForecastRange = ForecastRange (fromIntegral scg), -- REVIEW: Do we want to generate those randomly?
     gtDelay = delta,
-    gtSlotLength,
-    gtChainSyncTimeouts = chainSyncTimeouts gtSlotLength asc,
+    gtSlotLength = slotLengthFromSec 20,
+    gtChainSyncTimeouts = chainSyncTimeouts,
     gtBlockFetchTimeouts = blockFetchTimeouts,
     gtLoPBucketParams = LoPBucketParams { lbpCapacity = 50, lbpRate = 10 },
     -- These values give little enough leeway (5s) so that some adversaries get disconnected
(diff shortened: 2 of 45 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -512,51 +547,95 @@ addReprocessLoEBlocks
...
-                              (FailedToAddBlock "Queue flushed"))
-            as
-  where
-    blockAdd = \case
-      ChainSelAddBlock ab -> Just ab
-      ChainSelReprocessLoEBlocks -> Nothing
+closeChainSelQueue ChainSelQueue {varChainSelQueue} = do
+  chainSelQueue <- readTVar varChainSelQueue
+  for_ chainSelQueue $ \BlockToAdd {varBlockProcessed} ->
+    putTMVar varBlockProcessed $ FailedToAddBlock "Queue flushed"
```
amesgen, line 606: We are actually not flushing the queue here, this should probably be
```haskell
  chainSelQueue <- swapTVar varChainSelQueue mempty`
```
instead.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -512,51 +547,95 @@ addReprocessLoEBlocks
...
+      prevStarvation <- readTVarIO starvationVar
+      when (prevStarvation == ChainSelStarvationOngoing) $ do
+        tf <- getMonotonicTime
+        traceWith starvationTracer (ChainSelStarvationEnded tf $ blockRealPoint block)
+        writeTVarIO starvationVar (ChainSelStarvationEndedAt tf)
 
 -- | Flush the 'ChainSelQueue' queue and notify the waiting threads.
 --
+-- REVIEW: What about all the threads that are waiting to write in the queue and
+-- will write after the flush?!
```
amesgen, line 601: Indeed, there could be short time after the queue is flushed, but before the ChainDB is closed (and hence nothing new can be added to the ChainDB). I think this already exists on `main`.

I raised this on the IOG Slack, but I don't think we need to do anything about it in this PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -512,51 +547,95 @@ addReprocessLoEBlocks
...
+    go = do
+      (reprocessLoEBlocks, chainSelQueue) <- atomically readBoth
+      case reprocessLoEBlocks of
+        True -> do
+          writeTVarIO varChainSelReprocessLoEBlocks False
+          return ChainSelReprocessLoEBlocks
+        False ->
+          case Map.minView chainSelQueue of
+            Just (blockToAdd, chainSelQueue') -> do
+              writeTVarIO varChainSelQueue chainSelQueue'
```
amesgen, line 574: There is a race condition here when blocks are added to the `varChainSelQueue` after `readBoth` but before `writeTVarIO varChainSelQueue chainSelQueue'`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
amesgen: See https://github.com/intersectMBO/ouroboros-consensus/commit/23542f9d25b10c696bb967050eefed0e4a139abe

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
amesgen: See https://github.com/intersectMBO/ouroboros-consensus/commit/23542f9d25b10c696bb967050eefed0e4a139abe

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
```diff
@@ -32,51 +40,121 @@ import           Ouroboros.Consensus.Util.Args
...
-    gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
+data GenesisConfig = GenesisConfig
+  { gcBlockFetchConfig         :: !GenesisBlockFetchConfiguration
+  , gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
   , gcCSJConfig                :: !CSJConfig
-  , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig ())
+  , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig LoEAndGDDParams)
+  } deriving stock (Eq, Generic, Show)
+
+-- | Genesis configuration flags and low-level args, as parsed from config file or CLI
```
dnadales, line 61: It might help to add comments and examples of these flags and their effect on the node's behaviour. It will surely help whoever integrates this with the CLI/Node.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
```diff
@@ -32,51 +40,121 @@ import           Ouroboros.Consensus.Util.Args
...
+        then LoEAndGDDEnabled LoEAndGDDParams{lgpGDDRateLimit}
+        else LoEAndGDDDisabled
+    }
+  where
+    -- TODO justification/derivation from other parameters
+    defaultBlockFetchGracePeriod = 10 -- seconds
+    defaultCapacity              = 100_000 -- number of tokens
+    defaultRate                  = 500 -- tokens per second leaking, 1/2ms
+    -- 3 * 2160 * 20 works in more recent ranges of slots, but causes syncing to
+    -- block in byron.
```
dnadales, line 128:
```suggestion
    -- block in Byron.
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup/GenChains.hs
```diff
@@ -132,10 +131,11 @@ genChainsWithExtraHonestPeers genNumExtraHonest genNumForks = do
     gtSlotLength,
     gtChainSyncTimeouts = chainSyncTimeouts gtSlotLength asc,
     gtBlockFetchTimeouts = blockFetchTimeouts,
-    gtLoPBucketParams = LoPBucketParams { lbpCapacity = 100_000, lbpRate = 1_000 },
-    -- ^ REVIEW: Do we want to generate those randomly? For now, the chosen
-    -- values carry no special meaning. Someone needs to think about what values
-    -- would make for interesting tests.
+    gtLoPBucketParams = LoPBucketParams { lbpCapacity = 50, lbpRate = 10 },
+    -- ^ REVIEW: Do we want to generate those randomly?
```
dnadales, line 135: If we do not generate these randomly we should simply add a note that says that we could, but we should remove any pending `REVIEW`s from the code :pray:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -738,13 +764,63 @@ unregisterClient ::
...
+rotateDynamo ::
+  ( Ord peer,
+    LedgerSupportsProtocol blk,
+    MonadSTM m
+  ) =>
+  Tracer m (TraceEvent peer) ->
+  ChainSyncClientHandleCollection peer m blk ->
+  peer ->
+  m ()
+  -- STM m (Maybe (peer, ChainSyncClientHandle m blk))
```
dnadales, line 792: Dangling comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -334,13 +334,11 @@ chainSelSync cdb@CDB {..} (ChainSelAddBlock BlockToAdd { blockToAdd = b, .. }) =
...
     -- We follow the steps from section "## Adding a block" in ChainDB.md
-
-    -- Note: we call 'chainSelectionForFutureBlocks' in all branches instead
-    -- of once, before branching, because we want to do it /after/ writing the
-    -- block to the VolatileDB and delivering the 'varBlockWrittenToDisk'
-    -- promise, as this is the promise the BlockFetch client waits for.
-    -- Otherwise, the BlockFetch client would have to wait for
-    -- 'chainSelectionForFutureBlocks'.
+    --
+    -- Note: we call 'chainSelectionForFutureBlocks' in all branches instead of
```
dnadales, line 338: Capitalize `NOTE`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
dnadales: In that case I'd say: let's add a note and remove the `REVIEW` item.

jasagredo:
The releasing branch has been rebased on top of this one and there is now a test failure there. Is this expected? @Niols https://github.com/IntersectMBO/ouroboros-consensus/pull/1314/checks?check_run_id=34400813212

Niols:
Looking at the failure case:
```
      time limited leashing attack:                              FAIL (6.48s)
        *** Failed! Falsified (after 203 tests and 12 shrinks):
        GenesisTest:
          gtSecurityParam: 1
          gtGenesisWindow: 8
          gtForecastRange: 8
          gtDelay: 0
          gtSlotLength: SlotLength 20s
          gtCSJParams: CSJParams {csjpJumpSize = SlotNo 8}
          gtChainSyncTimeouts: 
            canAwait = Nothing
            intersect = Just 10s
            mustReply = Nothing
            idle = Nothing
          gtBlockFetchTimeouts: 
            busy = Just 60s
            streaming = Just 60s
          gtLoPBucketParams: 
            lbpCapacity = 50 tokens
            lbpRate = 10 % 1 ≅ 10.00 tokens per second
          gtBlockTree:
            G | 1-3 2-7 3-11 4-15
            G | 1-3 2-5[4] 3-12 4-15 5-16 6-18
            G | 1-3 2-8[3] 3-18
            G | 1-0[1] 2-8 3-9 4-10
            slots:    0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18
            trunk:  ───────────1───────────2───────────3───────────4
                               ╰─────2────────────────────3────────4──5─────6
                               ╰──────────────2─────────────────────────────3
                   ╰──1───────────────────────2──3──4
          gtSchedule:
            psSchedule =
               0: honest     : TP 4-15          | HP G           | BP G    @ 0.000000
               1: adversary  : TP 6-18[0,4,4x0] | HP G           | BP G    @ 0.000000
               2: adversary 2: TP 3-18[0,3,0]   | HP G           | BP G    @ 0.000000
               3: honest     : TP 4-15          | HP 3-11        | BP G    @ 0.019353
               4: honest     : TP 4-15          | HP 3-11        | BP 1-3  @ 0.032457
               5: honest     : TP 4-15          | HP 4-15        | BP 1-3  @ 0.040614
               6: adversary 2: TP 3-18[0,3,0]   | HP 2-8[0,3]    | BP G    @ 0.057771
               7: adversary  : TP 6-18[0,4,4x0] | HP 3-12[0,4,0] | BP G    @ 0.080588
               8: honest     : TP 4-15          | HP 4-15        | BP 2-7  @ 0.083783
               9: adversary 4: TP G             | HP 4-10[1,3x0] | BP G    @ 0.129071
              10: honest     : TP 4-15          | HP 4-15        | BP 4-15 @ 0.135964
            psStartOrder = []
            psMinEndTime = Time 77.1s
        SelectedChain: 1-3 | 2-7
        TipBlock: 2-7
        PeerSimulatorResults:
          - AdversarialPeer 1 (ChainSyncClient  - Interrupted) : DensityTooLow                                                                                                                    
          - AdversarialPeer 1 (ChainSyncServer  - Interrupted) : AsyncCancelled                                                                                                                   
          - AdversarialPeer 1 (BlockFetchClient - Interrupted) : AsyncCancelled                                                                                                                   
          - AdversarialPeer 1 (BlockFetchServer - Interrupted) : AsyncCancelled                                                                                                                   
          - AdversarialPeer 4 (ChainSyncClient  - Interrupted) : CandidateTooSparse Origin (Tip (SlotNo 3) (testHashFromList [0]) (BlockNo 1)) (Tip (SlotNo 0) (testHashFromList [1]) (BlockNo 1))
          - AdversarialPeer 4 (ChainSyncServer  - Interrupted) : AsyncCancelled                                                                                                                   
          - AdversarialPeer 4 (BlockFetchClient - Interrupted) : AsyncCancelled                                                                                                                   
          - AdversarialPeer 4 (BlockFetchServer - Interrupted) : AsyncCancelled                                                                                                                   
        
        
        Some peers were disconnected: adversary 4, adversary
        The immutable tip is too old: 12
        12 > 9
        Use --quickcheck-replay="(SMGen 17435934804358042701 8268706673786196345,2)" to reproduce.
        Use -p '/time limited leashing attack/' to rerun this test only.
```

The test looks reasonable, except `k = 1`? Not sure if that could lead to problems, or why such a case was generated. Somehow, despite `k = 1`, the client seems blocked on the fork between the honest chain and the chain served by `adversary 2`, unable to pass the intersection `1-3`.

The genesis window is of size 8, the honest chain has 2 blocks in there while `adversary 2` has only one. However, `adversary 2` never serves its block at `3-18`, so maybe that doesn't trigger the GDD? This should be covered by the LoP, in that case, because the adversary shouldn't be allowed to disarm it, and there is more than enough time. I would need to look at the full logs to see why that is not happening.

So either the test is flakey and we didn't catch that before and we're unlucky, or it's actually showing a subtle problem somewhere that we are not aware of, or maybe in the release branch some things changed in a way that impacts this. Will have a closer look.

Niols:
Checked out this repo on branch `neilmayhew/release-srp`; got to a commit ef51726fc33b85cbffe17058199e948b5509f208, then ran:

```
cabal run ouroboros-consensus-diffusion:consensus-test -- --quickcheck-replay="(SMGen 17435934804358042701 8268706673786196345,2)"
```
in the development environment and got:
```
Resolving dependencies...
Error: [Cabal-7107]
Could not resolve dependencies:
[__0] next goal: ouroboros-network (user goal)
[__0] rejecting: ouroboros-network-0.18.0.0 (constraint from user target requires ==0.16.1.1)
[__0] rejecting: ouroboros-network; 0.17.1.2, 0.17.1.1, 0.17.1.0, 0.17.0.0, 0.16.1.1, 0.16.1.0, 0.16.0.0, 0.15.0.0, 0.14.0.0, 0.13.1.0, 0.13.0.0, 0.12.0.0, 0.11.0.0, 0.10.2.2, 0.10.2.1, 0.10.1.0, 0.10.0.1, 0.10.0.0, 0.9.2.0, 0.9.1.0, 0.9.0.0, 0.8.2.0, 0.8.1.1, 0.8.1.0, 0.8.0.1, 0.8.0.0, 0.7.0.1, 0.7.0.0, 0.6.0.0, 0.5.0.0, 0.4.0.1, 0.4.0.0, 0.3.0.2, 0.3.0.1, 0.3.0.0, 0.2.0.0, 0.1.0.1, 0.1.0.0, 0.10.2.0 (constraint from user target requires ==0.18.0.0)
[__0] fail (backjumping, conflict set: ouroboros-network)
After searching the rest of the dependency tree exhaustively, these were the goals I've had most trouble fulfilling: ouroboros-network
```

I'm not super up-to-date on how to run this test suite, is there anything I'm missing here?

Niols:
Since the previous messages, I have been in contact with @jasagredo on the side. We have tracked down this test failure and concluded the following two things:

1. It is a bug in the peer simulator itself, and therefore a false positive — network safety is not threatened by this.
2. It is just an extremely rare occurrence, which was present already in `blockfetch/milestone-1` as of 7d82cf12fbf08d31cf29f0657ebfb226c595ecae.

For posterity, I will add some details in the following messages, but that was the gist of it. In the coming days, I will push a small fix for this peer simulator bug, probably just by re-enabling timeouts for the `canAwait` state of the ChainSync client. I will also try to measure the probability for this test failure to appear, so as to gain confidence that it is indeed fixed.

Niols:
Now, on `neilmayhew/release-srp` as of ef51726fc33b85cbffe17058199e948b5509f208, running
```
cabal run ouroboros-consensus-diffusion:consensus-test -- -p '/time limited leashing attack/' --quickcheck-replay="(SMGen 14174175684299620761 11115350297627667705,4)"
```
yields:
```
ouroboros-consensus
  Genesis tests
    uniform
      time limited leashing attack: FAIL (1.55s)
        *** Failed! Falsified (after 1 test and 21 shrinks):
        GenesisTest:
          gtSecurityParam: 2
          gtGenesisWindow: 5
          gtForecastRange: 5
          gtDelay: 1
          gtSlotLength: SlotLength 20s
          gtCSJParams: CSJParams {csjpJumpSize = SlotNo 5}
          gtChainSyncTimeouts: 
            canAwait = Nothing
            intersect = Just 10s
            mustReply = Nothing
            idle = Nothing
          gtBlockFetchTimeouts: 
            busy = Just 60s
            streaming = Just 60s
          gtLoPBucketParams: 
            lbpCapacity = 50 tokens
            lbpRate = 10 % 1 ≅ 10.00 tokens per second
          gtBlockTree:
            G | 1-0 2-1 3-4 4-5 5-6 6-7 7-9
            G | 1-0 2-2[1] 3-8 4-9
            slots:    0  1  2  3  4  5  6  7  8  9
            trunk:  ──1──2────────3──4──5──6─────7
                      ╰─────2─────────────────3──4
          gtSchedule:
            psSchedule =
              0: honest     : TP 7-9          | HP G          | BP G   @ 0.000000
              1: adversary 3: TP 4-9[0,1,2x0] | HP G          | BP G   @ 0.000000
              2: honest     : TP 7-9          | HP G          | BP 3-4 @ 0.029276
              3: honest     : TP 7-9          | HP 7-9        | BP 3-4 @ 0.040478
              4: adversary 3: TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G   @ 0.070241
              5: honest     : TP 7-9          | HP 7-9        | BP 7-9 @ 0.081851
            psStartOrder = []
            psMinEndTime = Time 61.7s
        SelectedChain: 1-0 | 2-1 3-4
        TipBlock: 3-4
        PeerSimulatorResults:
        
        
        No peers were disconnected
        The immutable tip is too old: 9
        9 > 7
        Use --quickcheck-replay="(SMGen 14174175684299620761 11115350297627667705,4)" to reproduce.
```

<details> 
  <summary>...and here are the full logs, obtained by enabling tracing.</summary>

```
GenesisTest:
    gtSecurityParam: 2
    gtGenesisWindow: 5
    gtForecastRange: 5
    gtDelay: 1
    gtSlotLength: SlotLength 20s
    gtCSJParams: CSJParams {csjpJumpSize = SlotNo 5}
    gtChainSyncTimeouts:
    canAwait = Nothing
    intersect = Just 10s
    mustReply = Nothing
    idle = Nothing
    gtBlockFetchTimeouts:
    busy = Just 60s
    streaming = Just 60s
    gtLoPBucketParams:
    lbpCapacity = 50 tokens
    lbpRate = 10 % 1 ≅ 10.00 tokens per second
    gtBlockTree:
    G | 1-0 2-1 3-4 4-5 5-6 6-7 7-9
    G | 1-0 2-2[1] 3-8 4-9
    slots:    0  1  2  3  4  5  6  7  8  9
    trunk:  ──1──2────────3──4──5──6─────7
                ╰─────2─────────────────3──4
    gtSchedule:
    psSchedule =
        0: honest     : TP 7-9          | HP G          | BP G   @ 0.000000
        1: adversary 3: TP 4-9[0,1,2x0] | HP G          | BP G   @ 0.000000
        2: honest     : TP 7-9          | HP G          | BP 3-4 @ 0.029276
        3: honest     : TP 7-9          | HP 7-9        | BP 3-4 @ 0.040478
        4: adversary 3: TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G   @ 0.070241
        5: honest     : TP 7-9          | HP 7-9        | BP 7-9 @ 0.081851
    psStartOrder = []
    psMinEndTime = Time 61.7s
00:00.000   Initiating node startup
00:00.000 BlockFetchLogic | PeersFetch []
00:00.000 GDD | Window: 0 -> 4
00:00.000       Selection: G
00:00.000       Candidates:
00:00.000       Candidate suffixes (bounds):
00:00.000       Density bounds:
00:00.000       New candidate tips:
00:00.000       Losing peers: []
00:00.000       Setting loeFrag: G
00:00.000   Node startup complete with selection G
Running point schedule ...
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.000
Tick:
    number: 0
    duration: 0s
    peer: honest
    state: TP 7-9 | HP G | BP G
    current chain: G
    candidate fragment: Nothing
    jumping states:

            ChainSyncServer honest       | handling MsgFindIntersect
            ChainSyncServer honest       |   state is TP 7-9 | HP G | BP G
            ChainSyncServer honest       | done handling MsgFindIntersect
            ChainSyncServer honest       |   intersection found: G
            ChainSyncClient honest       | Found intersection at: G
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP G | BP G
            ChainSyncServer honest       |   last intersection is G
            ChainSyncServer honest       |   intersection is exactly our header point
            ChainSyncServer honest       |   cannot serve at this point; waiting for node state and starting again
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP G | BP G
            ChainSyncServer honest       |   last intersection is G
            ChainSyncServer honest       |   intersection is exactly our header point
            ChainSyncServer honest       |   cannot serve at this point; waiting for node state and starting again
            GDD | Window: 0 -> 4
                Selection: G
                Candidates:
                    honest: G
                    adversary 3: G
                Candidate suffixes (bounds):
                    honest: G
                Density bounds:
                    honest: 0/5[ ], point: G, latest: unknown
                New candidate tips:
                    honest: G
                    adversary 3: G
                Losing peers: []
                Setting loeFrag: G
            BlockFetchLogic | PeersFetch [TraceLabelPeer (HonestPeer 1) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible)]
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.000
Tick:
    number: 1
    duration: 0.029276489858s
    peer: adversary 3
    state: TP 4-9[0,1,2x0] | HP G | BP G
    current chain: G
    candidate fragment: G
    jumping states:
    honest: Dynamo DynamoStarted G
    adversary 3: Jumper _ Happy FreshJumper Nothing

            ChainSyncServer adversary 3  | handling MsgFindIntersect
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  | done handling MsgFindIntersect
            ChainSyncServer adversary 3  |   intersection found: G
            ChainSyncClient adversary 3  | Found intersection at: G
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            GDD | Window: 0 -> 4
                Selection: G
                Candidates:
                    honest: G
                    adversary 3: G
                Candidate suffixes (bounds):
                    honest: G
                    adversary 3: G
                Density bounds:
                    honest: 0/5[ ], point: G, latest: unknown
                    adversary 3: 0/5[ ], point: G, latest: unknown
                New candidate tips:
                    honest: G
                    adversary 3: G
                Losing peers: []
                Setting loeFrag: G
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.029
Tick:
    number: 2
    duration: 0.011201112595s
    peer: honest
    state: TP 7-9 | HP G | BP 3-4
    current chain: G
    candidate fragment: G
    jumping states:
    honest: Dynamo DynamoStarted G
    adversary 3: Jumper _ Happy FreshJumper Nothing

            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP G | BP 3-4
            ChainSyncServer honest       |   last intersection is G
            ChainSyncServer honest       |   intersection is exactly our header point
            ChainSyncServer honest       |   cannot serve at this point; waiting for node state and starting again
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.040
Tick:
    number: 3
    duration: 0.029763889298s
    peer: honest
    state: TP 7-9 | HP 7-9 | BP 3-4
    current chain: G
    candidate fragment: G
    jumping states:
    honest: Dynamo DynamoStarted G
    adversary 3: Jumper _ Happy FreshJumper Nothing

            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is G
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: G | 1-0 2-1 3-4 4-5 5-6 6-7 7-9
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncServer honest       |   gotta serve 1-0
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is 1-0
            ChainSyncClient honest       | Downloaded header: 1-0
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 1-0 | 2-1 3-4 4-5 5-6 6-7 7-9
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncServer honest       |   gotta serve 2-1
            ChainSyncServer honest       |   tip is      7-9
            GDD | Window: 0 -> 4
                Selection: G
                Candidates:
                    honest: G
                    adversary 3: G
                Candidate suffixes (bounds):
                    honest: G
                    adversary 3: G
                Density bounds:
                    honest: 0/5[ ], point: G, latest: 0
                    adversary 3: 0/5[ ], point: G, latest: unknown
                New candidate tips:
                    honest: G
                    adversary 3: G
                Losing peers: []
                Setting loeFrag: G
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncClient honest       | Validated header: 1-0
            ChainSyncServer honest       |   last intersection is 2-1
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 2-1 | 3-4 4-5 5-6 6-7 7-9
            ChainSyncClient honest       | Gave LoP token to 1-0 compared to BlockNo 0
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncServer honest       |   gotta serve 3-4
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            BlockFetchLogic | PeersFetch [TraceLabelPeer (HonestPeer 1) (Right [At (Block {blockPointSlot = SlotNo 0, blockPointHash = (testHashFromList [0])})]),TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible)]
            ChainSyncClient honest       | Received instruction: RunNormally
            BlockFetchServer honest      | handling BlockFetch
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Starting batch for slice G | 1-0
            ChainSyncClient honest       | Downloaded header: 2-1
            BlockFetchServer honest      | done handling BlockFetch
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Sending 1-0
            BlockFetchServer honest      | done handling SendBlocks
            ChainSyncClient honest       | Validated header: 2-1
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncClient honest       | Gave LoP token to 2-1 compared to BlockNo 1
            BlockFetchServer honest      | Batch is done
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | ChainSel starvation ended at 00:00.040 thanks to 1-0
            BlockFetchLogic | PeersFetch [TraceLabelPeer (HonestPeer 1) (Right [At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [0,0])})]),TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible)]
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            BlockFetchServer honest      | handling BlockFetch
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Starting batch for slice 1-0 | 2-1
            BlockFetchServer honest      | done handling BlockFetch
            ChainSyncClient honest       | Downloaded header: 3-4
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Sending 2-1
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | Current chain: G
            ChainSyncClient honest       | Validated header: 3-4
            ChainDB                      | LoE fragment: G
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Batch is done
            ChainSyncClient honest       | Gave LoP token to 3-4 compared to BlockNo 2
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | Added to current chain; now: G | 1-0
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is 3-4
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 3-4 | 4-5 5-6 6-7 7-9
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncServer honest       |   gotta serve 4-5
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is 4-5
            ChainSyncClient honest       | Downloaded header: 4-5
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 4-5 | 5-6 6-7 7-9
            ChainDB                      | Current chain: G | 1-0
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainDB                      | LoE fragment: G
            ChainSyncServer honest       |   gotta serve 5-6
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncClient adversary 3  | Received instruction: JumpTo 3-4
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncClient adversary 3  | Offering jump to 3-4
            ChainSyncClient honest       | Validated header: 4-5
            ChainSyncServer honest       |   last intersection is 5-6
            ChainDB                      | Added to current chain; now: G | 1-0 2-1
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 5-6 | 6-7 7-9
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncClient honest       | Gave LoP token to 4-5 compared to BlockNo 3
            ChainSyncServer honest       |   gotta serve 6-7
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncServer adversary 3  | handling MsgFindIntersect
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  | done handling MsgFindIntersect
            ChainSyncServer adversary 3  |   no intersection found
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainDB                      | ChainSel starvation started at 00:00.040
            ChainSyncClient adversary 3  | Rejected jump to 3-4
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Downloaded header: 5-6
            ChainSyncClient adversary 3  | Received instruction: JumpTo 2-1
            ChainSyncClient adversary 3  | Offering jump to 2-1
            ChainSyncClient honest       | Validated header: 5-6
            ChainSyncServer adversary 3  | handling MsgFindIntersect
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  | done handling MsgFindIntersect
            ChainSyncClient honest       | Gave LoP token to 5-6 compared to BlockNo 4
            ChainSyncServer adversary 3  |   no intersection found
            ChainSyncClient adversary 3  | Rejected jump to 2-1
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncClient adversary 3  | Received instruction: JumpTo 1-0
            ChainSyncClient adversary 3  | Offering jump to 1-0
            ChainSyncClient honest       | Downloaded header: 6-7
            ChainSyncServer adversary 3  | handling MsgFindIntersect
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  | done handling MsgFindIntersect
            ChainSyncClient honest       | Waiting for SlotNo 7 beyond forecast horizon
            ChainSyncServer adversary 3  |   intersection found: 1-0
            ChainSyncClient adversary 3  | Accepted jump to 1-0
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            ChainSyncClient adversary 3  | Received instruction: JumpToGoodPoint 1-0
            ChainSyncClient adversary 3  | Offering jump to 1-0
            ChainSyncServer adversary 3  | handling MsgFindIntersect
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  | done handling MsgFindIntersect
            ChainSyncServer adversary 3  |   intersection found: 1-0
            ChainSyncClient adversary 3  | Accepted jump to good point: 1-0
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            ChainSyncClient adversary 3  | Received instruction: RunNormally
            ChainSyncServer adversary 3  | handling MsgRequestNext
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP G | BP G
            ChainSyncServer adversary 3  |   last intersection is 1-0
            ChainSyncServer adversary 3  |   intersection is further than our header point
            ChainSyncServer adversary 3  |   cannot serve at this point; waiting for node state and starting again
            GDD | Window: 1 -> 5
                Selection: G | 1-0 2-1
                Candidates:
                    honest: 5-6
                    adversary 3: 1-0
                Candidate suffixes (bounds):
                    honest: 1-0 | 2-1 3-4 4-5
                    adversary 3: 1-0
                Density bounds:
                    honest: 3/3[+], point: 2-1, latest: 7, has header after sgen
                    adversary 3: 0/5[ ], point: 1-0, latest: 0
                New candidate tips:
                    honest: 5-6
                    adversary 3: 1-0
                Losing peers: []
                Setting loeFrag: 1-0
            BlockFetchLogic | PeersFetch [TraceLabelPeer (HonestPeer 1) (Right [At (Block {blockPointSlot = SlotNo 4, blockPointHash = (testHashFromList [0,0,0])}),At (Block {blockPointSlot = SlotNo 5, blockPointHash = (testHashFromList [0,0,0,0])}),At (Block {blockPointSlot = SlotNo 6, blockPointHash = (testHashFromList [0,0,0,0,0])})]),TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible)]
            ChainDB                      | Requested ChainSel run
            BlockFetchServer honest      | handling BlockFetch
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainDB                      | Current chain: G | 1-0 2-1
            BlockFetchServer honest      | Starting batch for slice 2-1 | 3-4 4-5 5-6
            ChainDB                      | LoE fragment: G | 1-0
            BlockFetchServer honest      | done handling BlockFetch
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | Sending 3-4
            BlockFetchServer honest      | done handling SendBlocks
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | BP is behind
            BlockFetchServer honest      |   cannot serve at this point; waiting for node state and starting again
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | BP is behind
            BlockFetchServer honest      |   cannot serve at this point; waiting for node state and starting again
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | BP is behind
            BlockFetchServer honest      |   cannot serve at this point; waiting for node state and starting again
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 3-4
            BlockFetchServer honest      | BP is behind
            BlockFetchServer honest      |   cannot serve at this point; waiting for node state and starting again
            ChainDB                      | Current chain: G | 1-0 2-1
            ChainDB                      | LoE fragment: G | 1-0
            ChainDB                      | Added to current chain; now: G | 1-0 2-1 3-4
            ChainSyncClient honest       | Accessing SlotNo 7, previously beyond forecast horizon
            ChainSyncClient honest       | Validated header: 6-7
            ChainSyncClient honest       | Gave LoP token to 6-7 compared to BlockNo 5
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is 6-7
            ChainSyncServer honest       |   intersection is before our header point
            ChainSyncServer honest       |   fragment ahead: 6-7 | 7-9
            ChainSyncServer honest       | done handling MsgRequestNext
            ChainSyncServer honest       |   gotta serve 7-9
            ChainSyncServer honest       |   tip is      7-9
            ChainSyncClient honest       | Downloaded header: 7-9
            ChainSyncClient honest       | Validated header: 7-9
            ChainSyncClient honest       | Gave LoP token to 7-9 compared to BlockNo 6
            ChainSyncClient honest       | Waiting for next instruction from the jumping governor
            ChainSyncClient honest       | Received instruction: RunNormally
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
            ChainSyncServer honest       |   last intersection is 7-9
            ChainSyncServer honest       |   chain has been fully served
            ChainSyncServer honest       | done handling MsgRequestNext
            GDD | Window: 1 -> 5
                Selection: 1-0 | 2-1 3-4
                Candidates:
                    honest: 7-9
                    adversary 3: 1-0
                Candidate suffixes (bounds):
                    honest: 1-0 | 2-1 3-4 4-5
                    adversary 3: 1-0
                Density bounds:
                    honest: 3/3[+], point: 2-1, latest: 9, has header after sgen, idling
                    adversary 3: 0/5[ ], point: 1-0, latest: 0
                New candidate tips:
                    honest: 7-9
                    adversary 3: 1-0
                Losing peers: []
                Setting loeFrag: 1-0
            BlockFetchLogic | PeersFetch [TraceLabelPeer (HonestPeer 1) (Right [At (Block {blockPointSlot = SlotNo 7, blockPointHash = (testHashFromList [0,0,0,0,0,0])}),At (Block {blockPointSlot = SlotNo 9, blockPointHash = (testHashFromList [0,0,0,0,0,0,0])})]),TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible)]
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.070
Tick:
    number: 4
    duration: 0.011609304825s
    peer: adversary 3
    state: TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G
    current chain: 1-0 | 2-1 3-4
    candidate fragment: G | 1-0
    jumping states:
    honest: Disengaged DisengagedDone
    adversary 3: Dynamo 1-0 SlotNo 0

            ChainSyncServer adversary 3  | handling MsgRequestNext
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G
            ChainSyncServer adversary 3  |   last intersection is 1-0
            ChainSyncServer adversary 3  |   intersection is before our header point
            ChainSyncServer adversary 3  |   fragment ahead: 1-0 | 2-2[1] 3-8
            ChainSyncServer adversary 3  | done handling MsgRequestNext
            ChainSyncServer adversary 3  |   gotta serve 2-2[0,1]
            ChainSyncServer adversary 3  |   tip is      4-9[0,1,2x0]
            ChainSyncServer adversary 3  | handling MsgRequestNext
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G
            ChainSyncServer adversary 3  |   last intersection is 2-2[0,1]
            ChainSyncClient adversary 3  | Downloaded header: 2-2[0,1]
            ChainSyncServer adversary 3  |   intersection is before our header point
            ChainSyncServer adversary 3  |   fragment ahead: 2-2[0,1] | 3-8
            ChainSyncServer adversary 3  | done handling MsgRequestNext
            ChainSyncServer adversary 3  |   gotta serve 3-8[0,1,0]
            ChainSyncServer adversary 3  |   tip is      4-9[0,1,2x0]
            GDD | Window: 1 -> 5
                Selection: 1-0 | 2-1 3-4
                Candidates:
                    honest: 7-9
                    adversary 3: 1-0
                Candidate suffixes (bounds):
                    honest: 1-0 | 2-1 3-4 4-5
                    adversary 3: 1-0
                Density bounds:
                    honest: 3/3[+], point: 2-1, latest: 9, has header after sgen, idling
                    adversary 3: 0/5[ ], point: 1-0, latest: 2
                New candidate tips:
                    honest: 7-9
                    adversary 3: 1-0
                Losing peers: []
                Setting loeFrag: 1-0
            ChainSyncServer adversary 3  | handling MsgRequestNext
            ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G
            ChainSyncClient adversary 3  | Validated header: 2-2[0,1]
            ChainSyncServer adversary 3  |   last intersection is 3-8[0,1,0]
            ChainSyncServer adversary 3  |   intersection is exactly our header point
            ChainSyncServer adversary 3  |   cannot serve at this point; waiting for node state and starting again
            ChainSyncClient adversary 3  | Gave LoP token to 2-2[0,1] compared to BlockNo 1
            BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineInFlightThisPeer)]
            ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
            ChainSyncClient adversary 3  | Received instruction: JumpToGoodPoint 1-0
┌──────────────────────────────────────────────────────────────────────────────┐
└─ 00:00.081
Tick:
    number: 5
    duration: 0.1s
    peer: honest
    state: TP 7-9 | HP 7-9 | BP 7-9
    current chain: 1-0 | 2-1 3-4
    candidate fragment: 1-0 | 2-1 3-4 4-5 5-6 6-7 7-9
    jumping states:
    honest: Disengaged DisengagedDone
    adversary 3: Dynamo DynamoStarted SlotNo 0

            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            ChainSyncServer honest       | handling MsgRequestNext
            ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Sending 4-5
            ChainSyncServer honest       |   last intersection is 7-9
            BlockFetchServer honest      | done handling SendBlocks
            ChainSyncServer honest       |   chain has been fully served
            ChainSyncServer honest       | done handling MsgRequestNext
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Sending 5-6
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | ChainSel starvation ended at 00:00.081 thanks to 4-5
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Batch is done
            BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineInFlightThisPeer)]
            BlockFetchServer honest      | done handling SendBlocks
            BlockFetchServer honest      | handling BlockFetch
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Starting batch for slice 5-6 | 6-7 7-9
            BlockFetchServer honest      | done handling BlockFetch
            BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineInFlightThisPeer)]
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Sending 6-7
            BlockFetchServer honest      | done handling SendBlocks
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            BlockFetchServer honest      | Sending 7-9
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | Current chain: 1-0 | 2-1 3-4
            BlockFetchServer honest      | handling SendBlocks
            BlockFetchServer honest      |   state is TP 7-9 | HP 7-9 | BP 7-9
            ChainDB                      | LoE fragment: 1-0
            BlockFetchServer honest      | Batch is done
            BlockFetchServer honest      | done handling SendBlocks
            ChainDB                      | Current chain: 1-0 | 2-1 3-4
            ChainDB                      | LoE fragment: 1-0
            ChainDB                      | Current chain: 1-0 | 2-1 3-4
            ChainDB                      | LoE fragment: 1-0
            ChainDB                      | Current chain: 1-0 | 2-1 3-4
            ChainDB                      | LoE fragment: 1-0
            ChainDB                      | ChainSel starvation started at 00:00.081
            BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
00:10.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
00:10.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
00:20.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
00:20.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
00:30.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
00:30.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
00:40.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
00:40.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
00:50.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
00:50.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
01:00.081 BlockFetchLogic | PeerStarvedUs (HonestPeer 1)
01:00.081 BlockFetchLogic | PeersFetch [TraceLabelPeer (AdversarialPeer 3) (Left FetchDeclineChainNotPlausible),TraceLabelPeer (HonestPeer 1) (Left FetchDeclineAlreadyFetched)]
╶──────────────────────────────────────────────────────────────────────────────╴
Finished running point schedule
SelectedChain: 1-0 | 2-1 3-4
TipBlock: 3-4
PeerSimulatorResults:


No peers were disconnected
The immutable tip is too old: 9
9 > 7
Use --quickcheck-replay="(SMGen 14174175684299620761 11115350297627667705,4)" to reproduce.
```
</details>


The story is the following:
- Two peers, honest and adversary.
- Honest serves its chain entirely, while the adversary only serves some of its headers.
- The honest peer is chosen as the dynamo for CSJ, which we can see in tick 1:
  ```
  jumping states:
      honest: Dynamo DynamoStarted G
      adversary 3: Jumper _ Happy FreshJumper Nothing
  ```
- The honest peer serves all of its chain, and therefore gets disengaged from CSJ. We can see that in tick 3:
  ```
  ChainSyncServer honest       | handling MsgRequestNext
  ChainSyncServer honest       |   state is TP 7-9 | HP 7-9 | BP 3-4
  ChainSyncServer honest       |   last intersection is 7-9
  ChainSyncServer honest       |   chain has been fully served
  ChainSyncServer honest       | done handling MsgRequestNext
  ```
  and in the begining of tick 4:  
  ```
  jumping states:
    honest: Disengaged DisengagedDone
    adversary 3: Dynamo 1-0 SlotNo 0
  ```
- In fact, at this point, the adversary is just starting as a dynamo, and we first have to reset the state of its ChainSync client to the known intersection. This happens in the same tick (4):
  ```
  ChainSyncClient adversary 3  | Waiting for next instruction from the jumping governor
  ChainSyncClient adversary 3  | Received instruction: JumpToGoodPoint 1-0
  ```
- And then... nothing, the ChainSync client goes quiet.

@jasagredo and I tracked it down, and it turns out that the ChainSyncClient for the adversary is stuck in the `drainThePipe` function. This is because its corresponding ScheduledChainSyncServer is just not answering the `MsgRequestNext`, by design: indeed, the next header to be served is ahead of the header point, so it cannot be served, but it is not the tip point either, so we cannot say `MsgAwaitReply`. We could actually see that in tick 4, slightly before the instruction to `JumpToGoodPoint 1-0`:
```
ChainSyncServer adversary 3  | handling MsgRequestNext
ChainSyncServer adversary 3  |   state is TP 4-9[0,1,2x0] | HP 3-8[0,1,0] | BP G
ChainSyncServer adversary 3  |   last intersection is 3-8[0,1,0]
ChainSyncServer adversary 3  |   intersection is exactly our header point
ChainSyncServer adversary 3  |   cannot serve at this point; waiting for node state and starting again
```

So far so good. The test fails because nothing unblocks the situation, so we stay stuck at this intersection forever. In fact, the ChainSync mini-protocol of the adversary stays in `StCanAwait` for ever. It should be dislodged from it by timeouts, but those have been disabled! (See the very beginning of the first output I pasted.)

<details>
  <summary>Reminder of the ChainSync mini-protocol state machine</summary>
  
</details>

Now at this point I am not entirely sure whether there was a deep reason to disable those timeouts or if it was a bit of an oversight. Here is the corresponding test case in `Uniform.hs`:
```haskell
-- | Test that the leashing attacks do not delay the immutable tip after. The
-- immutable tip needs to be advanced enough when the honest peer has offered
-- all of its ticks.
--
-- See Note [Leashing attacks]
prop_leashingAttackTimeLimited :: Property
prop_leashingAttackTimeLimited =
  forAllGenesisTest

    (disableCanAwaitTimeout . disableBoringTimeouts <$>
      genChains (QC.choose (1, 4)) `enrichedWith` genTimeLimitedSchedule
    )

    defaultSchedulerConfig
      { scTrace = True
      , scEnableLoE = True
      , scEnableLoP = True
      , scEnableCSJ = True
      , scEnableBlockFetchTimeouts = False
      }

    shrinkPeerSchedules

    theProperty
```
Note in particular the `disableBoringTimeouts` and `disableCanAwaitTimeout`. The latter is probably self-explanatory. The former is:
```haskell
disableBoringTimeouts :: GenesisTest blk schedule -> GenesisTest blk schedule
disableBoringTimeouts gt =
    gt
      { gtChainSyncTimeouts =
          (gtChainSyncTimeouts gt)
            { mustReplyTimeout = Nothing
            , idleTimeout = Nothing
            }
      }
```

The conclusion is double:

- I think there is a mix-up between `canAwait` and `mustReply`; at least, there was one in my head and it is consistent with this piece of code: it is important to disable the `mustReply` timeout because, in the context of the peer simulator tests, the chains are finite, so an honest node will serve all of it, then send `MsgAwaitReply`, and then will not be able to do anything else, and it should not get disconnected for this, of course. I think that's the sense of this `disableCanAwaitTimeout` function, but really it should be `disableMustReplyTimeout` (look at the state machine again if need be!). I suppose the mix-up comes from the fact that the important message is `MsgAwaitReply`. I think I might be behind this mistake. It is an easy fix; I will add some documentation in the process.

- We cannot just disable the timeouts like this. In particular, the `canAwait` timeout must be enabled to a value that has an observable effect in the duration of the test (so maybe something like 10s is enough, maybe even much shorter than that). This part of the problem is slightly more delicate to fix, because there might be a reason why they were disabled in the first place, and the people that wrote this test are on PTO for the winter, and because choosing a good value for the timeouts, that is not too high but also doesn't just break good tests, is non-trivial. Still, I will whip something up!

Niols:
Come to think of it, I think the mix-up is that “can await” sounds much nicer and less urgent than “must reply”, at least in my head, even though those terms don't per se carry any time meaning. In practice, in `StCanAwait`, we expect a very prompt message, but it _can_ be a message that say that we will take more time (`MsgAwaitReply`), and in `StMustReply`, the message _has to_ be a proper reply (`MsgRoll*`), but it _can_ take much longer to come back.

Niols:
For the probability of the test failure, I now measured over night 22 test failures over 2,000,000 test cases. This means that it is very unlikely to happen, but that if my fix does pass the same amount of test cases over the next night, then we can be pretty confident that it was a good fix.

Niols:
Approximate probability that the bug appears at least once after a certain amount of tests:

image

geo2a (approved):
Approving to unblock the merge queue initiated by @dnadales.

PR #1182, opened 2024-07-10
Remove capacity override from the forging functions

nfrisby:
It's undesirable for a node to make blocks that aren't full. Moreover, this parameter was only ever set to "no override" in our codebase and in the `cardano-node` repo. So it seems good and harmless to remove this.

Note well that the capacity limit on the mempool is a separate parameter from the one removed by this PR.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20240710_104628_nick.frisby_remove_cap_override_forge.md
+++ b/ouroboros-consensus-cardano/changelog.d/20240710_104628_nick.frisby_remove_cap_override_forge.md
@@ -1,0 +1,22 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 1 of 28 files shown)


amesgen (approved):
LGTM after changelog fragments etc.

This functionality seems to have been introduced based on https://github.com/IntersectMBO/ouroboros-network/issues/3225 (implemented in https://github.com/IntersectMBO/ouroboros-network/pull/3238), but I think it was never used in cardano-node (in contrast to the mempool size override, see https://github.com/IntersectMBO/cardano-node/pull/3273 and https://github.com/IntersectMBO/cardano-node/pull/3343, which will have to change in light of #1175 as one can only tweak the "multiplicity"/"block count" then).

PR #1183, opened 2024-07-15
More verbose tracing when snapshotting

jasagredo:
# Description

Trace a starting event when snapshotting as well as tracing the time spent in snapshotting.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
@@ -37,6 +37,7 @@ import           Codec.Serialise.Encoding (Encoding)
 import           Control.Monad (forM, void)
 import           Control.Monad.Except (ExceptT (..))
 import           Control.Tracer
+import           Data.Functor.Contravariant ((>$<))
 import qualified Data.List as List
 import           Data.Maybe (isJust, mapMaybe)
 import           Data.Ord (Down (..))
@@ -51,6 +52,7 @@ import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy
 import           Ouroboros.Consensus.Util.CBOR (ReadIncrementalErr,
                      decodeWithOrigin, readIncremental)
+import           Ouroboros.Consensus.Util.Enclose
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Versioned
 import           System.FS.API.Lazy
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Enclose.hs
```diff
@@ -32,3 +36,18 @@ encloseWith ::
...
+  => Tracer m EnclosingTimed
+  -> m a
+  -> m a
+encloseTimedWith tracer action = do
+    before <- getMonotonicTimeNSec
+    traceWith tracer RisingEdge
+    res <- action
+    after <- getMonotonicTimeNSec
+    traceWith tracer (FallingEdgeWith (after - before))
+    pure res
```
amesgen, line 53: Maybe rather use [`getMonotonicTime`](https://hackage.haskell.org/package/si-timers-1.5.0.0/docs/Control-Monad-Class-MonadTime-SI.html#v:getMonotonicTime) and [`diffTime`](https://hackage.haskell.org/package/si-timers-1.5.0.0/docs/Control-Monad-Class-MonadTime-SI.html#v:diffTime) for more descriptive types instead of `Word64`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -119,8 +122,7 @@ takeSnapshot tracer hasFS encLedger oldest =
         if List.any ((== number) . dsNumber) snapshots then
           return Nothing
         else do
-          writeSnapshot hasFS encLedger snapshot oldest
-          traceWith tracer $ TookSnapshot snapshot tip
+          encloseTimedWith (TookSnapshot snapshot tip >$< tracer) $ writeSnapshot hasFS encLedger snapshot oldest
```
amesgen, line 125: Long line

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Enclose.hs
jasagredo: Fair, will do

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Enclose.hs
```diff
@@ -32,3 +35,18 @@ encloseWith ::
...
 encloseWith tracer action =
     traceWith tracer RisingEdge *> action <* traceWith tracer FallingEdge
+
+type EnclosingTimed = Enclosing' DiffTime
+
+encloseTimedWith ::
+     MonadMonotonicTime m
+  => Tracer m EnclosingTimed
+  -> m a
+  -> m a
```
amesgen, line 45: There might be variants of this combinator, also see https://github.com/IntersectMBO/ouroboros-network/pull/3752#discussion_r876038163. I don't think we should do anything about this here; we can add more combinators here on a per-need basis.

amesgen (approved):
LGTM (after fixing the formatting)

PR #1184, opened 2024-07-16
Add tasty-rerun to test-suites

jasagredo:
This allows us to do `cabal test all --test-options=--rerun` to only rerun failed tests.


ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/TestEnv.hs
```diff
@@ -26,8 +28,12 @@ defaultMainWithTestEnv :: TestEnvConfig -> TestTree -> IO ()
 defaultMainWithTestEnv testConfig testTree = do
     cryptoInit
     withStdTerminalHandles $
-      defaultMainWithIngredients (testEnvIngredient : defaultIngredients) $
-        withTestEnv testConfig testTree
+      defaultMainWithIngredients
+        ( rerunningTests [ listingTests, consoleTestReporter ]
+        : testEnvIngredient
+        : defaultIngredients
+        )
```
amesgen, line 35: I am not exactly sure whether this is necessary, but given that
```haskell
defaultIngredients = [listingTests, consoleTestReporter]
```
it seems surprising to list these twice here. What about
```suggestion
        [rerunningTests (testEnvIngredient : defaultIngredients)]
```
instead?

jasagredo: Ah indeed good point. I just copied what was there in the Readme 😄

amesgen (approved):

PR #1185, opened 2024-07-19
Fix formatting scripts for MSYS2

jasagredo:
The path from `pwd` doesn't work, it is a `/c/Users/...` path.

Also using this opportunity to ensure `LF` line endings.


amesgen (approved):
Confirmed that the scripts still work on Linux (in the Nix shell) :+1:

Issue #526, opened 2023-11-28
Add far future blocks to the ChainSync client test


nfrisby:
Context: https://github.com/input-output-hk/ouroboros-consensus/pull/525#discussion_r1407679130

We're currently merging a PR that adds near-future blocks. A reasonable next increment would add far-future blocks. But that's a little more delicate since such a block truncates each execution of the test.

PR #1186, opened 2024-07-22
ChainSync client: allow to prevent historical `MsgRollBackward`/`MsgAwaitReply`

amesgen:
This PR enables the ChainSync client to disconnect from peers who send *historical `MsgRollBackward` and `MsgAwaitReply` messages*, ie rollbacks that no honest caught-up peer would send as the corresponding blocks are already immutable for them, and `MsgAwaitReply`s that would indicate a chain whose tip is way behind the wall-clock.

This is relevant for the ChainSync Jumping (CSJ) optimization (part of Ouroboros Genesis): CSJ handles `MsgAwaitReply` and rollbacks to before a point that a peer previously acknowledged by "disengaging" that peer, ie running full ChainSync with them. This is the right thing to do when we are almost caught-up; however, during syncing the historical chain, this causes extra network load (albeit symmetrical) and CPU load.

We define a `MsgRollBackward`/`MsgAwaitReply` to be *historical* if the Genesis State Machine has not yet concluded that we are caught up and the wall-clock time of the slot of the oldest rewound header (if any) or the previous tip of the candidate fragment, respectively, is older than a configuration value, the *historicity cutoff*. In this PR, we set it to the duration of one mainnet stability window plus one hour as extra margin (so `36 h + 1 h = 37 h`). Future work might include getting this duration out of the HFC.

We also modify the existing ChainSync client test to test for false positives; false negatives can be tested by manually modifying the code (such as subtracting a constant from `historicityCutoff`, or introducing bugs in the actual implementation); we defer a standalone test for false negatives (similar to https://github.com/IntersectMBO/ouroboros-consensus/issues/526).

The PR is intended to be reviewed commit by commit.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
@@ -81,6 +81,9 @@ import           Ouroboros.Consensus.Fragment.InFuture (CheckInFuture,
                      ClockSkew)
 import qualified Ouroboros.Consensus.Fragment.InFuture as InFuture
 import           Ouroboros.Consensus.Ledger.Extended (ExtLedgerState (..))
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck
+                     (HistoricityCheck)
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck as HistoricityCheck
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
 import qualified Ouroboros.Consensus.Network.NodeToClient as NTC
 import qualified Ouroboros.Consensus.Network.NodeToNode as NTN
@@ -491,6 +494,11 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
               let gsmMarkerFileView =
                     case ChainDB.cdbsHasFSGsmDB $ ChainDB.cdbsArgs finalArgs of
                         SomeHasFS x -> GSM.realMarkerFileView chainDB x
+                  historicityCheck getGsmState =
+                    case gcHistoricityCutoff llrnGenesisConfig of
+                      Nothing                -> HistoricityCheck.noCheck
+                      Just historicityCutoff ->
+                        HistoricityCheck.mkCheck systemTime getGsmState historicityCutoff
               fmap (nodeKernelArgsEnforceInvariants . llrnCustomiseNodeKernelArgs)
                 $ mkNodeKernelArgs
                     registry
@@ -501,6 +509,7 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
                     rnTraceConsensus
                     btime
                     (InFutureCheck.realHeaderInFutureCheck llrnMaxClockSkew systemTime)
+                    historicityCheck
                     chainDB
                     llrnMaxCaughtUpAge
                     (Just durationUntilTooOld)
@@ -744,6 +753,7 @@ mkNodeKernelArgs ::
   -> Tracers m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
   -> BlockchainTime m
   -> InFutureCheck.SomeHeaderInFutureCheck m blk
+  -> (m GSM.GsmState -> HistoricityCheck m blk)
   -> ChainDB m blk
   -> NominalDiffTime
   -> Maybe (GSM.WrapDurationUntilTooOld m blk)
@@ -761,6 +771,7 @@ mkNodeKernelArgs
   tracers
   btime
   chainSyncFutureCheck
+  chainSyncHistoricityCheck
   chainDB
   maxCaughtUpAge
   gsmDurationUntilTooOld
@@ -778,6 +789,7 @@ mkNodeKernelArgs
       , chainDB
       , initChainDB             = nodeInitChainDB
       , chainSyncFutureCheck
+      , chainSyncHistoricityCheck
       , blockFetchSize          = estimateBlockSize
       , mempoolCapacityOverride = NoMempoolCapacityBytesOverride
       , miniProtocolParameters  = defaultMiniProtocolParameters
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
@@ -24,6 +24,8 @@ import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
                      (CSJConfig (..), CSJEnabledConfig (..),
                      ChainSyncLoPBucketConfig (..),
                      ChainSyncLoPBucketEnabledConfig (..))
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck
+                     (HistoricityCutoff (..))
 import qualified Ouroboros.Consensus.Node.GsmState as GSM
 import           Ouroboros.Consensus.Storage.ChainDB (ChainDbArgs)
 import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
@@ -45,6 +47,7 @@ data GenesisConfig = GenesisConfig {
     gcChainSyncLoPBucketConfig :: !ChainSyncLoPBucketConfig
   , gcCSJConfig                :: !CSJConfig
   , gcLoEAndGDDConfig          :: !(LoEAndGDDConfig ())
+  , gcHistoricityCutoff        :: !(Maybe HistoricityCutoff)
   }
 
 -- TODO justification/derivation from other parameters
@@ -58,6 +61,10 @@ enableGenesisConfigDefault = GenesisConfig {
           csjcJumpSize = 3 * 2160 * 20 -- mainnet forecast range
         }
     , gcLoEAndGDDConfig = LoEAndGDDEnabled ()
+      -- Duration in seconds of one Cardano mainnet Shelley stability window
+      -- (3k/f slots times one second per slot) plus one extra hour as a
+      -- safety margin.
+    , gcHistoricityCutoff = Just $ HistoricityCutoff $ 3 * 2160 * 20 + 3600
     }
 
 -- | Disable all Genesis components, yielding Praos behavior.
@@ -66,6 +73,7 @@ disableGenesisConfig = GenesisConfig {
       gcChainSyncLoPBucketConfig = ChainSyncLoPBucketDisabled
     , gcCSJConfig                = CSJDisabled
     , gcLoEAndGDDConfig          = LoEAndGDDDisabled
+    , gcHistoricityCutoff        = Nothing
     }
 
 -- | Genesis-related arguments needed by the NodeKernel initialization logic.
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
@@ -17,6 +17,7 @@ import           Data.Map.Strict (Map)
 import           Data.Proxy (Proxy (..))
 import           Network.TypedProtocol.Codec (AnyMessage)
 import           Ouroboros.Consensus.Block (Header, Point)
+import           Ouroboros.Consensus.BlockchainTime (RelativeTime (..))
 import           Ouroboros.Consensus.Config (TopLevelConfig (..))
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
                      (LedgerSupportsProtocol)
@@ -103,7 +104,9 @@ basicChainSyncClient
       { InFutureCheck.proxyArrival = Proxy
       , InFutureCheck.recordHeaderArrival = \_ -> pure ()
       , InFutureCheck.judgeHeaderArrival = \_ _ _ -> pure ()
-      , InFutureCheck.handleHeaderArrival = \_ -> pure Nothing
+      , InFutureCheck.handleHeaderArrival = \_ ->
+          -- We are not inspecting header slot time in the Genesis tests.
+          pure $ pure $ RelativeTime 0
       }
 
 -- | Create and run a ChainSync client using 'bracketChainSyncClient' and
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
@@ -26,6 +26,7 @@ import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
                      ChainSyncLoPBucketConfig, ChainSyncStateView (..),
                      Consensus, bracketChainSyncClient, chainSyncClient)
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client as CSClient
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck as HistoricityCheck
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
 import           Ouroboros.Consensus.Node.GsmState (GsmState (Syncing))
 import           Ouroboros.Consensus.Util (ShowProxy)
@@ -85,6 +86,10 @@ basicChainSyncClient
       , CSClient.cfg
       , CSClient.chainDbView
       , CSClient.someHeaderInFutureCheck = dummyHeaderInFutureCheck
+        -- Preventing historical MsgRollBack and MsgAwaitReply messages is
+        -- motivated by preventing additional load from CSJ-disengaged peers; we
+        -- do not care about this in these tests.
+      , CSClient.historicityCheck        = HistoricityCheck.noCheck
       }
     CSClient.DynamicEnv {
         CSClient.version             = maxBound
--- a/ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
+++ b/ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
@@ -38,10 +38,10 @@ import           Ouroboros.Consensus.Util.STM (Fingerprint (..),
 import           Ouroboros.Consensus.Util.Time (secondsToNominalDiffTime)
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import qualified Ouroboros.Network.AnchoredFragment as AF
-import qualified Ouroboros.Network.AnchoredSeq as AS
 import           Ouroboros.Network.Block (ChainUpdate (AddBlock, RollBack),
                      Tip (TipGenesis), tipFromHeader)
 import           Ouroboros.Network.ControlMessage (ControlMessage (Continue))
+import qualified Ouroboros.Network.Mock.Chain as Chain
 import           Ouroboros.Network.Protocol.ChainSync.ClientPipelined
 import           Ouroboros.Network.Protocol.ChainSync.Codec (codecChainSyncId)
 import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
@@ -105,11 +105,12 @@ oneBenchRun
         CSClient.getCurrentChain       = pure $ AF.Empty AF.AnchorGenesis
       , CSClient.getHeaderStateHistory =
             pure
-          $ HeaderStateHistory.HeaderStateHistory
-          $ AS.Empty
-          $ HV.genesisHeaderState ()
+          $ HeaderStateHistory.fromChain
+              topConfig
+              (oracularLedgerDB GenesisPoint)
+              Chain.Genesis
       , CSClient.getIsInvalidBlock     = pure invalidBlock
-      , CSClient.getPastLedger         = pure . oracularLedgerDB
+      , CSClient.getPastLedger         = pure . Just . oracularLedgerDB
       }
 
     headerInFutureCheck ::
@@ -174,9 +175,9 @@ inTheYearOneBillion = SystemTime {
         * 1e9
   }
 
-oracularLedgerDB :: Point B -> Maybe (Extended.ExtLedgerState B)
+oracularLedgerDB :: Point B -> Extended.ExtLedgerState B
 oracularLedgerDB p =
-    Just Extended.ExtLedgerState {
+    Extended.ExtLedgerState {
         Extended.headerState = HV.HeaderState {
             HV.headerStateTip      = case pointToWithOriginRealPoint p of
                 Origin       -> Origin
--- a/ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
+++ b/ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
@@ -22,6 +22,7 @@ import qualified Ouroboros.Consensus.HeaderStateHistory as HeaderStateHistory
 import qualified Ouroboros.Consensus.HeaderValidation as HV
 import qualified Ouroboros.Consensus.Ledger.Extended as Extended
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client as CSClient
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck as HistoricityCheck
 import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
 import           Ouroboros.Consensus.MiniProtocol.ChainSync.Server
                      (chainSyncServerForFollower)
@@ -128,6 +129,7 @@ oneBenchRun
               , CSClient.cfg                     = topConfig
               , CSClient.tracer                  = nullTracer `asTypeOf` contramap show debugTracer
               , CSClient.someHeaderInFutureCheck = headerInFutureCheck
+              , CSClient.historicityCheck        = HistoricityCheck.noCheck
               , CSClient.mkPipelineDecision0     =
                     pipelineDecisionLowHighMark 10 20
               }
--- a/ouroboros-consensus/changelog.d/20240722_170344_alexander.esgen_historical_rollbacks.md
+++ b/ouroboros-consensus/changelog.d/20240722_170344_alexander.esgen_historical_rollbacks.md
@@ -1,0 +1,4 @@
+### Breaking
+
+- Added functionality to disallow historical `MsgRollBackward`s and
+  `MsgAwaitReply`s in the ChainSync client.
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderStateHistory.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderStateHistory.hs
@@ -1,6 +1,10 @@
+{-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeFamilies               #-}
@@ -18,6 +22,11 @@ module Ouroboros.Consensus.HeaderStateHistory (
   , current
   , rewind
   , trim
+    -- * 'HeaderStateWithTime'
+  , HeaderStateWithTime (..)
+  , castHeaderStateWithTime
+  , mkHeaderStateWithTime
+  , mkHeaderStateWithTimeFromSummary
     -- * Validation
   , validateHeader
     -- * Support for tests
@@ -30,39 +39,44 @@ import qualified Data.List.NonEmpty as NE
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.BlockchainTime (RelativeTime)
 import           Ouroboros.Consensus.Config
+import           Ouroboros.Consensus.HardFork.Abstract (HasHardForkHistory (..))
+import           Ouroboros.Consensus.HardFork.History (Summary)
+import qualified Ouroboros.Consensus.HardFork.History.Qry as Qry
 import           Ouroboros.Consensus.HeaderValidation hiding (validateHeader)
 import qualified Ouroboros.Consensus.HeaderValidation as HeaderValidation
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Protocol.Abstract
-import           Ouroboros.Network.AnchoredSeq (AnchoredSeq (..))
+import           Ouroboros.Consensus.Util.CallStack (HasCallStack)
+import           Ouroboros.Network.AnchoredSeq (Anchorable, AnchoredSeq (..))
 import qualified Ouroboros.Network.AnchoredSeq as AS
 import           Ouroboros.Network.Mock.Chain (Chain)
 import qualified Ouroboros.Network.Mock.Chain as Chain
 
--- | Maintain a history of 'HeaderState's.
+-- | Maintain a history of 'HeaderStateWithTime's.
 newtype HeaderStateHistory blk = HeaderStateHistory {
       unHeaderStateHistory ::
            AnchoredSeq
              (WithOrigin SlotNo)
-             (HeaderState blk)
-             (HeaderState blk)
+             (HeaderStateWithTime blk)
+             (HeaderStateWithTime blk)
     }
   deriving (Generic)
 
-deriving instance (BlockSupportsProtocol blk, HasAnnTip blk)
-                => Eq (HeaderStateHistory blk)
-deriving instance (BlockSupportsProtocol blk, HasAnnTip blk)
-                => Show (HeaderStateHistory blk)
-deriving instance (BlockSupportsProtocol blk, HasAnnTip blk)
-                => NoThunks (HeaderStateHistory blk)
+deriving stock instance (BlockSupportsProtocol blk, HasAnnTip blk)
+                      => Eq (HeaderStateHistory blk)
+deriving stock instance (BlockSupportsProtocol blk, HasAnnTip blk)
+                      => Show (HeaderStateHistory blk)
+deriving newtype instance (BlockSupportsProtocol blk, HasAnnTip blk)
+                        => NoThunks (HeaderStateHistory blk)
 
-current :: HeaderStateHistory blk -> HeaderState blk
+current :: HeaderStateHistory blk -> HeaderStateWithTime blk
 current = either id id . AS.head . unHeaderStateHistory
 
 -- | Append a 'HeaderState' to the history.
-append :: HeaderState blk -> HeaderStateHistory blk -> HeaderStateHistory blk
+append :: HeaderStateWithTime blk -> HeaderStateHistory blk -> HeaderStateHistory blk
 append h (HeaderStateHistory history) = HeaderStateHistory (history :> h)
 
 -- | Trim the 'HeaderStateHistory' to the given size, dropping the oldest
@@ -82,7 +96,8 @@ cast ::
      )
   => HeaderStateHistory blk -> HeaderStateHistory blk'
 cast (HeaderStateHistory history) =
-    HeaderStateHistory $ AS.bimap castHeaderState castHeaderState history
+      HeaderStateHistory
+    $ AS.bimap castHeaderStateWithTime castHeaderStateWithTime history
 
 -- | \( O\(n\) \). Rewind the header state history
 --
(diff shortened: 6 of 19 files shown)


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -909,6 +926,20 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo
       , llrnMaxCaughtUpAge = secondsToNominalDiffTime $ 20 * 60   -- 20 min
       , llrnMaxClockSkew =
           InFuture.defaultClockSkew
+      , llrnMaxRollbackAge = case gcCSJConfig rnGenesisConfig of
+          -- If CSJ is disabled, we are either syncing from trusted peers (Praos
+          -- mode), or the LoP already handles historical rollbacks.
```
nfrisby, line 931:
```suggestion
          -- mode), or the LoP already indirectly handles historical rollbacks.
```

nfrisby: This prohibition would still benefit the syncing Genesis node, even without CSJ. And honest nodes should never trigger it.

So: do we really need to disable it based on this config?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -909,6 +926,20 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo
...
+          CSJDisabled  -> Nothing
+          -- If CSJ is enabled, we prevent historical rollbacks: When peers roll
+          -- back before their last acknowledged jump, we disengage them
+          -- (causing us to run full ChainSync with them). This is fine when we
+          -- are almost caught-up, but it is wasteful while syncing historical
+          -- parts of the chain.
+          --
+          -- Here, "historical" is defined to be the length of a Shelley
```
nfrisby, line 939: Perhaps add a comment explaining that this magic number is specific to Cardano mainnet.

Seems like the duration itself should be an `llrn` parameter, at least for the sake of nodes that aren't running Cardano mainnet (eg testnets). If so, that parameter should explain that the user should choose a duration that is slightly larger than the largest Chain Growth property's window of any era of their chain.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1847,6 +1865,49 @@ checkLoP ConfigEnv{tracer} DynamicEnv{loPBucket} hdr kis@KnownIntersectionState{
...
+checkHistoricalRollback ::
+     forall m blk. (IOLike m, HasHeader blk, GetHeader blk)
+  => ConfigEnv m blk
+  -> WhetherJustFoundIntersection blk
+  -> Point blk
+  -> m ()
+checkHistoricalRollback cfgEnv justFoundIntersection rollbackPoint =
+    case justFoundIntersection of
+      JustFoundIntersection intersectionPoint
+        -- Honest peers will send a rollback to the intersection point right
```
nfrisby, line 1879: I wonder if it'd be simpler/cleaner over all to embellish the `recvMsgIntersectFound` handlers to explicitly require/expect that `MsgRollBack` 🤔

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1847,6 +1865,49 @@ checkLoP ConfigEnv{tracer} DynamicEnv{loPBucket} hdr kis@KnownIntersectionState{
...
+  where
+    doHistoricalRollbackCheck = do
+        -- TODO Using the ledger at the chain tip is sound due to the
+        -- cross-chain lemma (Property 17.3 in the Consensus report), but we try
+        -- not to rely on it, so this should be changed.
+        currentLedger <- atomically $ do
+          chain <- getCurrentChain chainDbView
+          getPastLedger chainDbView (castPoint $ AF.headPoint chain) >>= \case
+            Nothing -> error "impossible"
+            Just st -> pure $ ledgerState st
```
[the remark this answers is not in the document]
nfrisby: Yeah, I had generally been anticipating that the LedgerDB ought to cache the `Summary` of each ledger state in it. This sounds like a reasonable step in that same direction.

Edit: However, also see my alternative approach suggested in a subsequent comment (in `HistoricalRollbacks.hs`).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1877,6 +1938,19 @@ castEarlyExitIntersects =
...
+-- | Whether we just negotiated an intersection. This is used to specially
+-- handle the first @MsgRollBackward@ that is sent right after.
+data WhetherJustFoundIntersection blk =
+    -- | We just received and successfully processed a @MsgIntersectFound@. The
+    -- ChainSync network spec guarantees that the reply to the next
+    -- @MsgRequestNext@ will be @MsgRollBackward@ to the intersection point.
+    JustFoundIntersection !(Point blk)
+  | -- | The last intersection was negotiated in the past, ie we received at
+    -- least one reply to @MsgRequestNext@ in the meantime.
+    IntersectionWasFoundInThePast
```
nfrisby, line 1950: I suggest renaming this, since "the past" is a relatively ambiguous term, easy to interpret in a variety of ways

Perhaps: `ExpectingRequiredMsgRollBack | NotExpectingRequiredMsgRollBack`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricalRollbacks.hs
```diff
@@ -0,0 +1,134 @@
...
+  -> m GsmState
+     -- ^ Get the current 'GsmState'.
+     --
+     -- This is used to disable the historical rollback check when we are caught
+     -- up. The rationale is extra resilience against disconnects between honest
+     -- nodes in disaster scenarios with very low chain density.
+  -> NominalDiffTime
+     -- ^ The maximum age of a rollback at arrival time.
+     --
+     -- This should be set to at least the maximum duration duration (across all
```
nfrisby, line 109:
```suggestion
     -- This should be set to at least the maximum duration (across all
     ```

File: ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricalRollbacks.hs [hunk 7]
```diff
@@ -0,0 +1,134 @@
...
+            , arrivalTime
+            , maxRollbackAge
+            }
+      where
+        -- Use slot 0 in case we are rolling back to Genesis.
+        rollbackSlot  = fromWithOrigin 0 $ pointSlot rollbackPoint
+        qry           = Qry.slotToWallclock rollbackSlot
+        hfSummary     = hardForkSummary lcfg lst
+        -- This can't fail as per the precondition on 'onRollBackward'.
+        (slotTime, _) = Qry.runQueryPure qry hfSummary
```
nfrisby, line 134: I think there's a different approach that would work and maybe be simpler/clearer, but would require some yak shaving.

The idea would be to record the UTC time of a header's slot when validating that header. We necessarily have it, because of the future check. So maybe we could simply add it as a field of the `HeaderState` data type, for example.

Then the roll back handler (ie `HeaderState.rollback`) could return the oldest header state that was rolled back (if any). This check here would merely need to compare that oldest-rolled-back UTC time against the current wall clock.

Moreover: this would actually avoid the required-msg-rollback-after-intersection wrinkle since such a `MsgRollBack` doesn't actually rollback any blocks. (And if it somehow did, that'd be a proper roll back, not dismissable as just part of an intersection response.)

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -594,6 +597,60 @@ runChainSync skew securityParam (ClientUpdates clientUpdates)
...
+              where
+                time = clientTimeForNewMaxSlot newMaxSlot
+            f _   _ _            = error "bad clockUpdates"
+
+    -- For the historical rollback check. This is calculated by considering the
+    -- 'ChainUpdate' which rolls back to the earliest slot relative to the
+    -- client wallclock time in the given tick.
+    maxRollbackAge :: NominalDiffTime
+    maxRollbackAge =
+        case foldMap (Just . Max) rollbackAges of
```
nfrisby, line 640: `foldl' max 0 rollbackAges`

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -594,6 +597,60 @@ runChainSync skew securityParam (ClientUpdates clientUpdates)
...
+        NewMaxClientAndServerSlot cslot sslot ->
+          toOnset cslot `max` toSkewedOnset sslot
+
+    clockUpdates :: Schedule NewMaxSlot
+    clockUpdates =
+        mkClockUpdates
+          (ClientUpdates clientUpdates)
+          (ServerUpdates serverUpdates)
+
+    clientTimeForTick :: Tick -> RelativeTime
```
nfrisby, line 616: Maybe add a comment here that at least recommends rereading the module header for context on how ticks/time/skew/etc works in this test.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Yeah, that would also work, I somewhat shyed away from it as it would explicitly *require* this behavior from our server (currently, both on `main` and with the current state of this PR, we tolerate servers that don't actually send this, also see https://github.com/IntersectMBO/ouroboros-consensus/pull/1181), but I think it should be fine. Will think about it, also given that https://github.com/IntersectMBO/ouroboros-consensus/pull/1186#discussion_r1688176610 *might* supersede this.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -281,6 +286,10 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
       -- | Maximum clock skew
     , llrnMaxClockSkew :: ClockSkew
 
+      -- | If we receive a ChainSync rollback to a slot that is older than this
+      -- upon arrival, we will disconnect. If 'Nothing', disable the check.
```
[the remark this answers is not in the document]
amesgen: Adopted :+1: (with a "not" added to the last sentence)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
amesgen: Added it to `GenesisConfig` separately and moved the comment there :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricalRollbacks.hs
amesgen: Hmm, this sounds interesting; it would of course invalidate existing ledger snapshots (or require some rather ad-hoc code for migration).

> Then the roll back handler (ie `HeaderState.rollback`) could return the oldest header state that was rolled back (if any). This check here would merely need to compare that oldest-rolled-back UTC time against the current wall clock.

Any particular reason for not using the `HeaderState` of the tip of the result of `HeaderState.rollback`? Just because it would avoid the "required-msg-rollback-after-intersection wrinkle"?

> Moreover: this would actually avoid the required-msg-rollback-after-intersection wrinkle since such a `MsgRollBack` doesn't actually rollback any blocks. (And if it somehow did, that'd be a proper roll back, not dismissable as just part of an intersection response.)

This would be equivalent to just not doing the historical rollback check if the rollback isn't actually rolling back any blocks even in the current state of the PR, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Always grateful for better naming suggestions, will change. (I dropped the `Required` for now as this PR does not yet actually *require* the rollback, it just tolerates it.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: I implemented this [here](https://github.com/IntersectMBO/ouroboros-consensus/tree/amesgen/historical-rollbacks-alt), but it has test failures:

<details>
<summary>Details for the test failure</summary>

```
ouroboros-consensus
  ChainSyncClient
    chainSync: FAIL (0.10s)
      *** Failed! Falsified (after 1 test and 16 shrinks):
      ChainSyncClientSetup {securityParam = SecurityParam 3, clientUpdates = ClientUpdates {getClientUpdates = Schedule {getSchedule = fromList [(Tick {tickToWord64 = 1},[AddBlock (TestBlockWith {tbHash = (testHashFromList [2]), tbSlot = SlotNo 1, tbValid = Valid, tbPayload = ()}),AddBlock (TestBlockWith {tbHash = (testHashFromList [2,0]), tbSlot = SlotNo 2, tbValid = Valid, tbPayload = ()}),AddBlock (TestBlockWith {tbHash = (testHashFromList [2,0,0]), tbSlot = SlotNo 3, tbValid = Valid, tbPayload = ()}),AddBlock (TestBlockWith {tbHash = (testHashFromList [2,0,0,2]), tbSlot = SlotNo 4, tbValid = Valid, tbPayload = ()})])]}}, serverUpdates = ServerUpdates {getServerUpdates = Schedule {getSchedule = fromList [(Tick {tickToWord64 = 1},[]),(Tick {tickToWord64 = 2},[]),(Tick {tickToWord64 = 3},[AddBlock (TestBlockWith {tbHash = (testHashFromList [3]), tbSlot = SlotNo 1, tbValid = Invalid, tbPayload = ()}),SwitchFork Origin [TestBlockWith {tbHash = (testHashFromList [2]), tbSlot = SlotNo 1, tbValid = Valid, tbPayload = ()}]]),(Tick {tickToWord64 = 4},[]),(Tick {tickToWord64 = 5},[]),(Tick {tickToWord64 = 6},[]),(Tick {tickToWord64 = 7},[AddBlock (TestBlockWith {tbHash = (testHashFromList [2,3]), tbSlot = SlotNo 2, tbValid = Invalid, tbPayload = ()})]),(Tick {tickToWord64 = 8},[SwitchFork (At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])})) [TestBlockWith {tbHash = (testHashFromList [2,0]), tbSlot = SlotNo 2, tbValid = Valid, tbPayload = ()}],SwitchFork (At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])})) [TestBlockWith {tbHash = (testHashFromList [2,0]), tbSlot = SlotNo 2, tbValid = Valid, tbPayload = ()}],SwitchFork Origin [TestBlockWith {tbHash = (testHashFromList [0]), tbSlot = SlotNo 1, tbValid = Valid, tbPayload = ()},TestBlockWith {tbHash = (testHashFromList [0,0]), tbSlot = SlotNo 2, tbValid = Valid, tbPayload = ()},TestBlockWith {tbHash = (testHashFromList [0,0,0]), tbSlot = SlotNo 3, tbValid = Valid, tbPayload = ()}]])]}}, startTick = Tick {tickToWord64 = 3}, invalidBlocks = InvalidBlocks {getInvalidBlocks = Schedule {getSchedule = fromList [(Tick {tickToWord64 = 1},[]),(Tick {tickToWord64 = 2},[]),(Tick {tickToWord64 = 3},[]),(Tick {tickToWord64 = 4},[(testHashFromList [3]),(testHashFromList [0,2,0,0,1,0,2,0,1,3])]),(Tick {tickToWord64 = 5},[(testHashFromList [0,2,0,0,1,3]),(testHashFromList [0,2,0,0,1,0,3])]),(Tick {tickToWord64 = 6},[]),(Tick {tickToWord64 = 7},[]),(Tick {tickToWord64 = 8},[(testHashFromList [2,3])])]}}, clientSlowBy = SlotLengthTenths 0}
      ChainSyncClientSetup:
      securityParam: 3
      clientSlowBy: 0s
      --
      clockUpdates:
      1: c4|s_
      3: c_|s1
      7: c_|s2
      8: c_|s3
      --
      clientUpdates:
      1: AddBlock (H:[2],S:1,B:1), AddBlock (H:[2,0],S:2,B:2), AddBlock (H:[2,0,0],S:3,B:3), AddBlock (H:[2,0,0,2],S:4,B:4)
      --
      serverUpdates:
      3: AddBlock (H:[3],S:1,B:1), SwitchFork <- Origin -> (H:[2],S:1,B:1)
      7: AddBlock (H:[2,3],S:2,B:2)
      8: SwitchFork <- (Point 1, [2]) -> (H:[2,0],S:2,B:2), SwitchFork <- (Point 1, [2]) -> (H:[2,0],S:2,B:2), SwitchFork <- Origin -> (H:[0],S:1,B:1) (H:[0,0],S:2,B:2) (H:[0,0,0],S:3,B:3)
      --
      startTick: Tick {tickToWord64 = 3}
      invalidBlocks:
      4: [3], [0,2,0,0,1,0,2,0,1,3]
      5: [0,2,0,0,1,3], [0,2,0,0,1,0,3]
      8: [2,3]


      Client chain: Origin ] (H:[2],S:1,B:1) :> (H:[2,0],S:2,B:2) :> (H:[2,0,0],S:3,B:3) :> (H:[2,0,0,2],S:4,B:4)
      Server chain: Origin ] (H:[0],S:1,B:1) :> (H:[0,0],S:2,B:2) :> (H:[0,0,0],S:3,B:3)
      Synced fragment: (S:1; H:(testHashFromList [2])) ]
      Trace:
      (6,40s) | Protocol: Send MsgFindIntersect [At (Block {blockPointSlot = SlotNo 4, blockPointHash = (testHashFromList [2,0,0,2])}),At (Block {blockPointSlot = SlotNo 3, blockPointHash = (testHashFromList [2,0,0])}),At (Block {blockPointSlot = SlotNo 2, blockPointHash = (testHashFromList [2,0])}),At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])})]
      (7,40s) | Protocol: Recv MsgIntersectFound At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])}) Tip (SlotNo 1) (testHashFromList [2]) (BlockNo 1)
      (7,40s) | Client: TraceFoundIntersection (At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])})) (Tip (SlotNo 4) (testHashFromList [2,0,0,2]) (BlockNo 4)) (Tip (SlotNo 1) (testHashFromList [2]) (BlockNo 1))
      (8,40s) | Protocol: Send MsgRequestNext
      (8,40s) | Protocol: Recv MsgRollBackward Origin Tip (SlotNo 3) (testHashFromList [0,0,0]) (BlockNo 3)

      Exception: NoRollbackToIntersection (At (Block {blockPointSlot = SlotNo 1, blockPointHash = (testHashFromList [2])})) (Right Origin)
      Use --quickcheck-replay="(SMGen 7391498360415002818 16079556119013927611,30)" to reproduce.
```

</details>

This is because the [reference implementation](https://github.com/IntersectMBO/ouroboros-network/blob/29899dffaaab43e82cbee381c8889563083cc2b8/ouroboros-network-protocols/testlib/Ouroboros/Network/Protocol/ChainSync/Examples.hs#L161-L167) does not adhere to the [spec](https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf#section.3.7):



Should be easy to fix though.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -76,6 +84,31 @@ getLedgerDB ::
...
+        -- @ledgerDb@ as these are not newer than the tip slot of the current
+        -- ledger state (Property 17.1 in the Consensus report).
+        summary = hardForkSummary (configLedger cfg) currentLedgerState
+        mkHeaderStateWithTime' =
+              mkHeaderStateWithTimeFromSummary summary
+            . headerState
+    pure
+      . HeaderStateHistory
+      . LedgerDB.ledgerDbBimap headerState mkHeaderStateWithTime'
+      $ ledgerDb
```
amesgen, line 110: This reasoning could be avoided by using `mkHeaderStateWithTime` on each `ExtLedgerState` individually; which would construct (at most) `k` many `Summary`s. However, this optimized implementation is IMO still easy to understand, and only constructs at most one `Summary`.

For context: This function is currently called once per new ChainSync client.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -594,6 +597,65 @@ runChainSync skew securityParam (ClientUpdates clientUpdates)
...
+        rollbackAges :: [NominalDiffTime]
+        rollbackAges =
+          [ clientTime `diffRelTime` toSkewedOnset oldestRewound
+          | (tick, updates) <- Map.toList $ getSchedule serverUpdates
+          , let clientTime = clientTimeForTick tick
+          , SwitchFork rollbackPoint _blks <- updates
+          , -- Here, we make use of the fact that the blocks generated for this
+            -- test have dense slot numbers (ie there are no empty slots).
+            let oldestRewound =
+                  withOrigin firstSlot succ $ pointSlot rollbackPoint
```
amesgen, line 654: A future test for false negatives (eg one that ensures that running with `maxRollbackAge - 1` sometimes fails) would catch if the generator changed in the future; but I think this is fine for now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Superseded by using the approach from https://github.com/IntersectMBO/ouroboros-consensus/pull/1186#discussion_r1688176610

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Superseded by using the approach from https://github.com/IntersectMBO/ouroboros-consensus/pull/1186#discussion_r1688176610

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricalRollbacks.hs
amesgen: The PR has been rewritten to this approach: the payload of the `HeaderStateHistory` is now `HeaderStateWithTime` which is a `HeaderState` and a `RelativeTime` of the header slot. This approach avoids invalidating ledger snapshots.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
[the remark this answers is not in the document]
amesgen: Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/1186#discussion_r1688176610

nfrisby (approved):
Looks good! Thanks for switching approaches.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
```diff
@@ -0,0 +1,163 @@
...
+-- guaranteed to arise).
+--
+-- For example, on Cardano mainnet today, the Praos Chain Growth property
+-- implies that @3k/f@ (=129600) slots (=36 hours) will contain at least @k@
+-- (=2160) blocks. (Byron has a smaller stability window, namely @2k@ (=24 hours
+-- as the Byron slot length is 20s). Thus a peer rolling back a header that is
+-- older than 36 hours or signals that it doesn't have more headers is either
+-- violating the maximum rollback or else isn't a caught-up node. Either way, a
+-- syncing node should not be connected to that peer.
+newtype HistoricityCutoff = HistoricityCutoff {
```
amesgen, line 105: To reviewers: Please say if you don't like this name. A more descriptive but more mouthful alternative would be `MaxRollbackAndAwaitReplyAge` for example.

dnadales: Yeah, "historicity" might be too generic, but maybe once one is familiar with its meaning in this context that won't be an issue anymore. I'm perfectly ok with the longer alternative, and I'd prefer it if it'd be up to me, but I'm fine either way.

dnadales (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
```diff
@@ -84,6 +86,10 @@ basicChainSyncClient
       , CSClient.cfg
       , CSClient.chainDbView
       , CSClient.someHeaderInFutureCheck = dummyHeaderInFutureCheck
+        -- Preventing historical MsgRollBacks and MsgAwaitReplies is motivated
```
nfrisby, line 89: minor: MsgRollBack and MsgAwaitReply messages

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs
```diff
@@ -103,7 +109,9 @@ basicChainSyncClient
       { InFutureCheck.proxyArrival = Proxy
       , InFutureCheck.recordHeaderArrival = \_ -> pure ()
       , InFutureCheck.judgeHeaderArrival = \_ _ _ -> pure ()
-      , InFutureCheck.handleHeaderArrival = \_ -> pure Nothing
+      , InFutureCheck.handleHeaderArrival = \_ ->
```
nfrisby, line 112: Would you explain this change --- seems unrelated based on the diff I've reviewed so far.

ouroboros-consensus/bench/ChainSync-client-bench/Main.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE LambdaCase     #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
nfrisby, line 3: TODO remove this flag

ouroboros-consensus/changelog.d/20240722_170344_alexander.esgen_historical_rollbacks.md
```diff
@@ -0,0 +1,4 @@
+### Breaking
+
+- Added functionality to prevent historical `MsgRollBackward`s and
```
nfrisby, line 3: I'd say "disallow" instead of "prevent"

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderStateHistory.hs
```diff
@@ -107,12 +124,81 @@ cast (HeaderStateHistory history) =
...
+     , TipInfo blk ~ TipInfo blk'
+     )
+  => HeaderStateWithTime blk -> HeaderStateWithTime blk'
+castHeaderStateWithTime hswt = HeaderStateWithTime {
+      hswtHeaderState = castHeaderState $ hswtHeaderState hswt
+    , hswtSlotTime    = hswtSlotTime hswt
+    }
+
+mkHeaderStateWithTimeFromSummary ::
+     forall blk. (HasCallStack, HasAnnTip blk)
```
nfrisby, line 176: minor: unnecessary `forall`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderStateHistory.hs
```diff
@@ -107,12 +124,81 @@ cast (HeaderStateHistory history) =
...
+        hswtHeaderState = hst
+      , hswtSlotTime    = slotTime
+      }
+  where
+    (slotTime, _) = Qry.runQueryPure qry summary
+    qry           = Qry.slotToWallclock slot
+    slot          = fromWithOrigin 0 $ pointSlot $ headerStatePoint hst
+
+mkHeaderStateWithTime ::
+     forall blk. (HasCallStack, HasHardForkHistory blk, HasAnnTip blk)
```
nfrisby, line 192: minor: unnecessary `forall`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderStateHistory.hs
```diff
@@ -107,12 +124,81 @@ cast (HeaderStateHistory history) =
...
-  -> HeaderStateHistory blk -> Maybe (HeaderStateHistory blk)
-rewind p (HeaderStateHistory history) = HeaderStateHistory <$>
-    AS.rollback
+  -> HeaderStateHistory blk
+  -> Maybe (HeaderStateHistory blk, Maybe (HeaderStateWithTime blk))
+rewind p (HeaderStateHistory history) = do
+    (prefix, suffix) <- AS.splitAfterMeasure
       (pointSlot p)
-      ((== p) . either headerStatePoint headerStatePoint)
+      ((== p) . join either (headerStatePoint . hswtHeaderState))
```
nfrisby, line 132: Does `(headerStatePoint . hswtHeaderState) . either id id` work? That might be easier for many readers.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1277,14 +1286,14 @@ knownIntersectionStateTop cfgEnv dynEnv intEnv =
...
-                rewoundHistory =
+                (rewoundHistory, _oldestRewound) =
                   fromMaybe (error "offerJump: cannot rewind history") mRewoundHistory
                 -- If the tip of jTheirFragment does not match the tip of
                 -- jTheirHeaderStateHistory, then the history needs rewinding.
                 historyNeedsRewinding =
                      (/= AF.headPoint (jTheirFragment ji)) $
                      castPoint $
-                     either headerStatePoint headerStatePoint $
+                     join either (headerStatePoint . hswtHeaderState) $
```
nfrisby, line 1296: `f . either id id` here too?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1769,11 +1793,12 @@ checkValid ::
   => ConfigEnv m blk
   -> InternalEnv m blk arrival judgment
   -> Header blk
+  -> RelativeTime
```
nfrisby, line 1796: Please add ` -- ^ onset of the header's slot`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -2177,6 +2215,7 @@ instance Eq ChainSyncClientException where
     InvalidIntersection{}            == _ = False
     InvalidBlock{}                   == _ = False
     InFutureHeaderExceedsClockSkew{} == _ = False
+    HistoricityError{}             == _ = False
```
nfrisby, line 2218: minor: `==` alignment

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
```diff
@@ -0,0 +1,163 @@
...
+import           Ouroboros.Consensus.Node.GsmState (GsmState (..))
+
+{-------------------------------------------------------------------------------
+  Interface
+-------------------------------------------------------------------------------}
+
+-- | Interface for the ChainSync client for deciding whether @MsgRollBackward@s
+-- and @MsgAwaitReply@s are historical.
+data HistoricityCheck m blk = HistoricityCheck {
+    -- | Determine whether the received @MsgRollBackward@ is historical. The
```
nfrisby, line 42: This comment seems stale in a couple ways, not that MsgAwaitReply is also involved.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
```diff
@@ -0,0 +1,163 @@
...
+
+-- | ChainSync historicity checks are performed for @MsgRollBackward@s and
+-- @MsgAwaitReply@s, see 'HistoricityCheck'.
+data HistoricalChainSyncMessage =
+    HistoricalMsgRollBackward
+  | HistoricalMsgAwaitReply
+  deriving stock (Show, Eq)
+
+data HistoricityException =
+  -- | We received a @MsgRollBackward@ or a @MsgAwaitReply@ at a time too early
```
nfrisby, line 59: I'd rephrase "at a time too early" to "while their candidate chain was too old" or something like that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
```diff
@@ -0,0 +1,163 @@
...
+  -- INVARIANT: @'historicityCutoff' < 'arrivalTime' `diffRelTime` 'slotTime'@
+  forall blk. HasHeader blk => HistoricityException {
+      historicalMessage :: HistoricalChainSyncMessage
+      -- | Depending on 'historicalMessage':
+      --
+      --  * 'HistoricalMsgRollBackward': The oldest header that was rewound.
+      --
+      --  * 'HistoricalMsgAwaitReply': The tip of the candidate fragment.
+    , historicalPoint   :: !(Point blk)
+      -- | The time corresponding to the slot of the oldest header that was
```
nfrisby, line 71: This comment's current wording doesn't work for `MsgAwaitReply`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
```diff
@@ -0,0 +1,163 @@
...
+      -- rolled back.
+    , slotTime          :: !RelativeTime
+      -- | When the @MsgRollBackward@ was received.
```
nfrisby, line 74: ibid

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
nfrisby: I think "Historicity" strikes a good balance.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -76,6 +85,31 @@ getLedgerDB ::
...
+     ( IOLike m
+     , HasHardForkHistory blk
+     , HasAnnTip blk
+     , IsLedger (LedgerState blk)
+     )
+  => ChainDbEnv m blk -> STM m (HeaderStateHistory blk)
+getHeaderStateHistory cdb@CDB{cdbTopLevelConfig = cfg} = do
+    ledgerDb <- getLedgerDB cdb
+    let currentLedgerState = ledgerState $ LedgerDB.ledgerDbCurrent ledgerDb
+        -- This summary can convert all tip slots of the ledger states in the
```
nfrisby, line 101: Unless there's a Chain Growth violation, right?

That's probably fine for now --- lots goes wrong with CG is violated.

But ideally the ChainSel logic could be calculating these incrementally, such that this function is merely a lookup.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -403,6 +401,15 @@ runChainSync skew securityParam (ClientUpdates clientUpdates)
             -- Note that this tests passes in the exact difference between the
             -- client's and server's clock as the tolerable clock skew.
 
+        historicityCheck :: HistoricityCheck m TestBlock
+        historicityCheck =
+            HistoricityCheck.mkCheck
+              clientSystemTime
+              -- The historicity check check is disabled when we use 'CaughtUp'
```
nfrisby, line 408: "check check"

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
```diff
@@ -58,6 +61,10 @@ enableGenesisConfigDefault = GenesisConfig {
           csjcJumpSize = 3 * 2160 * 20 -- mainnet forecast range
         }
     , gcLoEAndGDDConfig = LoEAndGDDEnabled ()
+      -- Duration in seconds of one Cardano mainnet Shelley stability window
+      -- (3k/f slots times one second per slot) plus one extra hour as a
+      -- safety margin.
+    , gcHistoricityCutoff = Just $ HistoricityCutoff $ 3 * 2160 * 20 + 3600
```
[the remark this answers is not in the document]
amesgen: Yes, this is specific to Cardano mainnet; like many other constants in this area (e.g. the timeouts, protocol-level size limits, the GSM [MaxCaughtUpAge](https://github.com/IntersectMBO/ouroboros-consensus/blob/f7d78f8e7d493ebe0ca3312a93b05f05838b3dd2/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs#L948)).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
amesgen: Thanks, changed to
```haskell
    -- | Determine whether the received message is historical. Depending on the
    -- first argument, the second argument is:
    --
    --  * 'HistoricalMsgRollBackward': The oldest state that was rolled back.
    --    (Note that rollbacks of depth zero are hence never historical.).
    --
    --  * 'HistoricalMsgAwaitReply': The state corresponding to the tip of the
    --    candidate fragment when @MsgAwaitReply@ was sent.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
amesgen: Changed to
```haskell
      -- | The time corresponding to the slot of 'historicalPoint'.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/HistoricityCheck.hs
amesgen: Changed to
```haskell
      -- | When the offending 'historicalMessage' was received.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
amesgen: This should work even in case of a chain growth violations. Note that we use the ledger state *at the tip* of the LedgerDB in order to translate the time of tip slots of all previous ledger states.

nfrisby: Yep, I had it backwards. I think I missed the "not" in the "not newer" you wrote.

PR #1187, opened 2024-07-24
Make it build with ghc 9.10

erikd:


cabal.project
```diff
@@ -38,3 +38,18 @@ import: ./asserts.cabal
...
+  -- All these cardano-ledger packages have been fixed on master but not
+  -- yet released to CHaP.
+  , cardano-ledger-alonzo:plutus-ledger-api
+  , cardano-ledger-alonzo-test:plutus-ledger-api
+  , cardano-ledger-babbage:plutus-ledger-api
+  , cardano-ledger-binary:plutus-ledger-api
+  , cardano-ledger-conway:plutus-ledger-api
+  -- Upstream maintainer seems reluctant to fix this.
+  -- https://github.com/phadej/latex-svg/issues/10
+  , latex-svg-image:base
```
jasagredo, line 52: Notice we are using our own version from CHaP, not the one on Hackage

cabal.project
```diff
@@ -38,3 +38,18 @@ import: ./asserts.cabal
 if(os(windows))
   constraints:
     bitvec -simd
+
+allow-newer:
```
jasagredo, line 42: Couldn't we wrap this in a conditional for the GHC version?

erikd: I hope to remove these before  pull it out of draft mode.

cabal.project
erikd: Thanks, I will try updating the one in CHaP.

jasagredo: It is a fork of mine (I opened a PR/issue against the upstream one long ago and it is still waiting there). Let me give you permissions for pushing

erikd: Add a revision for the one in CHaP. PR incoming for https://github.com/jasagredo/latex-svg

erikd: PR submitted.

Now using the fixed version from CHaP.

cabal.project
erikd: It is some time before the `ledger` packages are released to CHaP so I have made them conditional on GHC version.

PR #1188, opened 2024-07-25
Nix/Hydra/GHA: add GHC 9.10, drop GHC 9.8

amesgen:
Also drop GHC 9.8 as it is broken due to shipping a buggy `stm` version, see https://github.com/input-output-hk/io-sim/pull/166.


amesgen:
Not merging this in order to prevent extra Hydra load until we can drop GHC 8.10

jasagredo (approved):

PR #1189, opened 2024-07-25
Update docusaurus

jasagredo:
# Description

Update docusaurus webpage. Also separate the headers in Glossary from the `:`, otherwise the right side navigation bar omits the first word:
image


docs/website/contents/for-developers/Glossary.md
```diff
@@ -5,7 +5,7 @@ Notes on the use and maintenance of this glossary:
 - Definitions are preceded by a colon (`:`), eg `:Slot` to make it easy to find the using `Ctrl-F`.
 - When adding a new definition, add a colon in front of it. If the definition has synonyms, separate them by commas (prefixing a colon for each synonym).
 
-## :Active Slot Coefficient, :f
+## : Active Slot Coefficient, :f
```
amesgen, line 8: > Also separate the headers in Glossary from the :, otherwise the right side navigation bar omits the first word

Hmm, these changes are unfortunate, as the colon was intentionally set that way:

> - Definitions are preceded by a colon (`:`), eg `:Slot` to make it easy to find the using `Ctrl-F`.
> - When adding a new definition, add a colon in front of it. If the definition has synonyms, separate them by commas (prefixing a colon for each synonym).

A `;` seems to not cause this weird behavior, maybe change `:` to `;` everywhere?

amesgen (changes requested):

docs/website/contents/for-developers/Glossary.md
jasagredo: Done!

amesgen (approved):

Issue #1192, opened 2024-07-29
UTxO-HD heap increase in latest benchmarks


jasagredo:
Live data increased from 3800MB to 4200MB in value workload

![image](https://github.com/user-attachments/assets/d1fd9fa9-8d39-4490-b859-1b47f11cb7f8)

and from 3400MB to 4600MB in plutus workload

![image](https://github.com/user-attachments/assets/f33ecf02-3b13-4345-85da-08748e3827df)

jasagredo:
Local cluster runs seem unreliable to reproduce this behavior. As the difference is more drastic in the `forge-stress-pre-plutus` workload we are running the following clusters:
- `utxo-hd-9.1` with `-hT`
![image (2)](https://github.com/user-attachments/assets/3a7fc9dd-3652-4273-bc36-ac49da9bb77b)
- `utxo-hd-9.1` with `-hi` and `-finfo-table-map -fdistinct-constructor-tables`
- `9.1.0` with `-hT`

jasagredo:
As a side note, #1194 doesn't seem to have any notable benefit in the memory usage.

jasagredo:
The graphs above show that on local runs, there is only an 80 MB difference on BabbageTxOut which cannot be unpacked in GHC 8.10.7, see https://github.com/IntersectMBO/ouroboros-consensus/blob/440828510cd40f8f0cded91c89a21ed6fdbf7b9a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs#L94

jasagredo:
This is blocked on us being able to get profiling information from the cluster runs. We need the help from the Perf team.

jasagredo:
Done in #1232

PR #1194, opened 2024-07-29
Use even simpler diffs

jasagredo:
Just holding two maps, one for deletes one for inserts. This should
remove the layer of indirection from `Delta`.

Related to #1192


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Orphans.hs
```diff
@@ -8,6 +8,7 @@
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE UndecidableInstances       #-}
 
+
```
jorisdral, line 11:
```suggestion
```

ouroboros-consensus/test/consensus-test/Test/Consensus/Ledger/Tables/DiffSeq.hs
```diff
@@ -9,15 +9,15 @@ module Test.Consensus.Ledger.Tables.DiffSeq (tests) where
 
 import           Control.Monad (liftM)
 import qualified Data.FingerTree.RootMeasured.Strict as RMFT
-import           Data.Map.Diff.Strict (Delta (..), Diff)
-import           Data.Map.Diff.Strict.Internal (DeltaHistory (..), Diff (..))
+{- import           Data.Map.Diff.Strict (Delta (..), Diff)
+import           Data.Map.Diff.Strict.Internal (DeltaHistory (..), Diff (..)) -}
```
jorisdral, line 13:
```suggestion
```

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/V1/BackingStore.hs
```diff
@@ -325,14 +325,11 @@ instance (Ord k, QC.Arbitrary k, QC.Arbitrary v)
...
-  arbitrary =
-    QC.oneof [
-        Diff.Insert <$> QC.arbitrary
-      , pure Diff.Delete
-      ]
+instance (Ord k, QC.Arbitrary k, QC.Arbitrary v)
+                       => QC.Arbitrary (Diff.UtxoDiff k v) where
+
+  arbitrary = Diff.UtxoDiff <$> QC.arbitrary <*> QC.arbitrary
+  shrink (Diff.UtxoDiff m1 m2) = (flip Diff.UtxoDiff m2 <$> QC.shrink m1) ++ (Diff.UtxoDiff m1 <$> QC.shrink m2)
```
jorisdral, line 332: Neat trick I use often to coast on existing shrinkers:

```suggestion
  shrink (Diff.UtxoDiff m1 m2) = [ Diff.UtxoDiff m1' m2' | (m1', m2') <- shrink (m1, m2) ]
```

ouroboros-consensus/test/consensus-test/Test/Consensus/Ledger/Tables/Diff.hs
```diff
@@ -80,41 +162,44 @@ prop_applySum x ds = F.foldl' applyDiff x ds === applyDiff x (foldMap' id ds)
...
 ------------------------------------------------------------------------------}
 
-deriving newtype instance (Ord k, Arbitrary k, Arbitrary v)
-                       => Arbitrary (Diff k v)
+instance (Ord k, Arbitrary k, Arbitrary v)
+                       => Arbitrary (UtxoDiff k v) where
 
-instance Arbitrary v => Arbitrary (Delta v) where
+  arbitrary = UtxoDiff <$> QC.arbitrary <*> QC.arbitrary
+  shrink (UtxoDiff m1 m2) = (flip UtxoDiff m2 <$> QC.shrink m1) ++ (UtxoDiff m1 <$> QC.shrink m2)
```
jorisdral, line 196:
```suggestion
  shrink (UtxoDiff m1 m2) = [ UtxoDiff m1' m2' | (m1', m2') <- shrink (m1, m2) ]
```

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -313,7 +313,6 @@ library
     containers >=0.5 && <0.7,
     contra-tracer,
     deepseq,
-    diff-containers >=1.2,
```
jorisdral, line 316: does this mean we're not using `diff-containers` anywhere anymore?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
```diff
@@ -0,0 +1,177 @@
...
+import qualified Data.Set as Set
+import           GHC.Generics
+import           NoThunks.Class
+import           Prelude hiding (null)
+
+{------------------------------------------------------------------------------
+  Types
+------------------------------------------------------------------------------}
+
+data UtxoDiff k v = UtxoDiff (Map k v) (Map k v)
```
jorisdral, line 45: What do the fields mean?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
```diff
@@ -0,0 +1,177 @@
...
+
+instance Ord k => Semigroup (UtxoDiff k v) where
+  -- right biased
+  (UtxoDiff m1 m2) <> (UtxoDiff m3 m4) =
+     UtxoDiff
+        -- all deletes except the ones cancelling m2
+        (Map.union m1 (m3 `Map.difference` m2))
+        -- inserts from the first component without the latest deletes and inserts
+        -- from the second component
+        (Map.unionWith const (m2 `Map.difference` m3) m4)
```
jorisdral, line 61:
```suggestion
        (Map.union (m2 `Map.difference` m3) m4)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
```diff
@@ -0,0 +1,177 @@
...
+
+keysSet :: Ord k => UtxoDiff k v -> Set k
+keysSet (UtxoDiff m1 m2) = Map.keysSet m1 <> Map.keysSet m2
+
+{------------------------------------------------------------------------------
+  Construction
+------------------------------------------------------------------------------}
+
+diff :: Map k v -> Map k v -> UtxoDiff k v
+diff = UtxoDiff
```
jorisdral, line 79: Why does this works? Seems counterintuitive. Should it not be something like:

```suggestion
diff :: Map k v -> Map k v -> UtxoDiff k v
diff m1 m2 = UtxoDiff (m1 `Map.difference` m2) (m2 `Map.difference` m1)
```

If it's just the application of the `UtxoDiff` constructor that you need, then I'd suggest exposing that

ouroboros-consensus/test/consensus-test/Test/Consensus/Ledger/Tables/Diff.hs
```diff
@@ -54,24 +71,89 @@ lawsTestOne p tts =
   Applying diffs
 ------------------------------------------------------------------------------}
 
-type K = Int
-type V = Char
+type K = Word16
+type V = Word16
```
jorisdral, line 75: The arbitrary instance or `Word16` generates a value uniformly over the whole range `[minBound .. maxBound]`, so the probability of generating the same key twice is very small. I think you should use a generator with a higher probability of duplicates to see if the tests still pass

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
```diff
@@ -0,0 +1,177 @@
...
+        -- we cannot have added something that was deleted
+        dels = stripPrefix m1 m3
+    in
+      case (dels, stripPrefix m2 m4) of
+        -- Violation of UTXO-property
+        (Nothing, _) -> Nothing
+        -- m2 ⊄ m4 because some have been cancelled
+        (Just dels', Nothing) ->
+          let
+            (notCancelled, cancelled) = Map.partitionWithKey (\k _ -> k `Map.member` m4) m2
```
jorisdral, line 158: This looks like it could be a rather costly operation: doing a `Map.member` for each key. Maybe using `Map.difference` or something like that is cheaper

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
```diff
@@ -0,0 +1,177 @@
...
+{-------------------------------------------------------------------------------
+  Filter
+-------------------------------------------------------------------------------}
+
+filterOnlyKey :: (k -> Bool) -> UtxoDiff k v -> UtxoDiff k v
+filterOnlyKey f (UtxoDiff m1 m2) =
+  UtxoDiff (Map.filterWithKey (const . f) m1) (Map.filterWithKey (const . f) m2)
+
+{-------------------------------------------------------------------------------
+  Reductive instances
```
jorisdral, line 142: The correctness of these instances relies on a relation between the two argument `UtxoDiffs`, so *technically* these instances are unlawful. My main argument against providing unlawful instances is that it introduces (unforeseen) ways in which code can go wrong, and that it will be hard to debug when it does. Having fewer assumptions, or assumptions that we can statically guarantee is probably nicest

Some options:
* Clearly document the assumptions we make, and possibly let the instances thrown an error when the UTXO property is violated.
* Don't provide instance for `UtxoDiff k v`, but rather something like `Maybe (UtxoDiff k v)`, and then the UTXO property violation error case is represented as `Nothing` (and the using code can act on the error)
* (my preference): stop using `fingertree-rm`, and use a normal finger tree instead, which just requires `Monoid` and `Semigroup` instances (which are lawful)

jorisdral (commented):
Okay, looks promising. Props for getting the reductive/cancellative instances to work! Here are some comments

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/UtxoDiff.hs
jasagredo: Done, with something isomorphic to `Maybe`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/Utils.hs
```diff
@@ -270,12 +274,15 @@ attachEmptyDiffs l1 = over l1 $ ltmap rawAttachEmptyDiffs (ltprj l1)
 -- second argument is @TrackingMK v2 d2@, it should be the case that @applyDiff
 -- v1 d2 == v2@.
 rawPrependTrackingDiffs ::
-      Ord k
+      (HasCallStack, Ord k)
    => TrackingMK k v
    -> TrackingMK k v
    -> TrackingMK k v
```
jasagredo, line 280: This should return a Maybe

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/Utils.hs
```diff
@@ -121,11 +122,14 @@ forgetTrackingDiffs l = over l $ ltmap rawForgetTrackingDiffs (ltprj l)
 --
 
 rawPrependDiffs ::
-     Ord k
+     (HasCallStack, Ord k)
   => DiffMK k v -- ^ Earlier differences
   -> DiffMK k v -- ^ Later differences
   -> DiffMK k v
```
jasagredo, line 128: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/DiffSeq.hs
```diff
@@ -201,24 +201,34 @@ noSlotBoundsIntersect (SlotNoUB sl1) (SlotNoLB sl2) = sl1 <= sl2
...
 instance (Ord k, Eq v) => RootMeasured (RootMeasure k v) (Element k v) where
   measureRoot (Element _ d) =
-      RootMeasure 1 d (Sum $ Diff.numInserts d) (Sum $ Diff.numDeletes d)
+      RootMeasure 1 d (Sum $ UtxoDiff.numInserts d) (Sum $ UtxoDiff.numDeletes d)
 
 instance (Ord k, Eq v) => Semigroup (RootMeasure k v) where
   RootMeasure len1 d1 n1 m1 <> RootMeasure len2 d2 n2 m2 =
-      RootMeasure (len1 <> len2) (d1 <> d2) (n1 <> n2) (m1 <> m2)
+    case AUtxoDiff d1 <> AUtxoDiff d2 of
+      NotAUtxoDiff s -> error s
```
jasagredo, line 209: We can keep this if `extend` preserves the invariant.

jasagredo:
This PR seems to have NO EFFECT on the memory consumption of UTxO-HD.

PR #1201, opened 2024-08-01
reduce the frequency of client and server diverging before they reach startTick

### LINKED ISSUE #529: ChainSync client test: decrease number of tests without any sent headers
> Upon inspection, I was surpsised that ~70% of the cases had `SomeNotEarly_SomeEarly Zero Zero`, which implies the test never involved downloading a header!
>
> My initial guess is that the client and server chain are diverging before they ever reach `startTick`.
>
> I suspect we want to tune the generator to visit that less often. But that can be a separate Issue.

_Originally posted by @nfrisby in https://github.com/input-output-hk/ouroboros-consensus/pull/525#discussion_r1408513738_

> Yeah, that seems be the case, when always setting `startTick` to `1`, the `Zero Zero` case shrinks to ~9% of all cases.

fraser-iohk:
This PR changes the generator for `ChainSyncClientSetup` such that `startTick` is now heavily weighted to pick a lower value (closer to 1) rather than a linear range between 1 and `maxStartTick`. This causes fewer randomly-generated `ChainSyncClientSetup`s to produce test runs in which no headers are sent due to immediate divergence.

Before: 

```
ouroboros-consensus
  ChainSyncClient
    chainSync: OK (5.69s)                                                              
      +++ OK, passed 10000 tests:
      67.35% ForkTooDeep
      27.89% NoMoreIntersection
       0.25% RolledBackPastIntersection
      
      TickArrivalTimeStats (10000 in total):
      67.40% OnlyNotEarly_SomeEarly Zero Zero
      10.76% OnlyNotEarly_SomeEarly Zero Many
       6.72% OnlyNotEarly_SomeEarly One Many
       5.95% OnlyNotEarly_SomeEarly One One
       3.75% OnlyNotEarly_SomeEarly Many Zero
       2.89% OnlyNotEarly_SomeEarly One Zero
       1.38% OnlyNotEarly_SomeEarly Many Many
       1.07% OnlyNotEarly_SomeEarly Many One
       0.08% OnlyNotEarly_SomeEarly Zero One
```

After:
```
ouroboros-consensus
  ChainSyncClient
    chainSync: OK (8.51s)                                                              
      +++ OK, passed 10000 tests:
      84.58% NoMoreIntersection
      10.51% ForkTooDeep
       0.62% RolledBackPastIntersection
      
      TickArrivalTimeStats (10000 in total):
      19.50% OnlyNotEarly_SomeEarly One Many
      16.81% OnlyNotEarly_SomeEarly One Zero
      15.85% OnlyNotEarly_SomeEarly Zero Many
      13.83% OnlyNotEarly_SomeEarly One One
      13.25% OnlyNotEarly_SomeEarly Many Zero
      10.58% OnlyNotEarly_SomeEarly Zero Zero
       5.39% OnlyNotEarly_SomeEarly Many Many
       4.79% OnlyNotEarly_SomeEarly Many One

```

Note that the `NoMoreIntersection` case is now significantly more common than the `ForkTooDeep` case, rather than the inverse.

Closes #529

### DIFF
--- a/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -766,7 +766,7 @@ instance Arbitrary ChainSyncClientSetup where
           , lastTick (getClientUpdates clientUpdates) - 1
           , lastTick (getServerUpdates serverUpdates) - 1
           ]
-    startTick <- choose (1, maxStartTick)
+    startTick <- chooseExponential 3 (1, maxStartTick)
     let trapBlocks =
           [ blockHash b
           | AddBlock b <- joinSchedule (getServerUpdates serverUpdates)
@@ -827,6 +827,13 @@ instance Arbitrary ChainSyncClientSetup where
     [ cscs { clientSlowBy = SlotLengthTenths y }
     | let SlotLengthTenths x = clientSlowBy
     , y <- shrink x
+    ]
+
+chooseExponential :: Enum a => Int -> (a, a) -> Gen a
+chooseExponential decayFactor (low, up) =
+  frequency
+    [ (decayFactor, pure low)
+    , (1, chooseExponential decayFactor (succ low, up))
     ]
 
 prettyChainSyncClientSetup :: ChainSyncClientSetup -> String


ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -829,6 +829,13 @@ instance Arbitrary ChainSyncClientSetup where
     , y <- shrink x
     ]
 
+chooseExponential :: Enum a => Int -> (a, a) -> Gen a
+chooseExponential decayFactor (low, up) =
+  frequency
+    [ (1, pure low)
+    , (1, chooseExponential decayFactor (succ low, up))
+    ]
```
amesgen, line 837: `decayFactor` is completely unused by this function, it is just passed around recursively. I imagine it is supposed to influence the weights in some way?

Also: what about moving this to `Test.Util.QuickCheck` as well as adding documentation? Maybe take inspiration from [hedgehog](https://hackage.haskell.org/package/hedgehog-1.5/docs/Hedgehog-Range.html#g:5)?

amesgen (changes requested):
Can confirm that this achives the goal of #529:

#### Before #1201

```
      TickArrivalTimeStats (10000 in total):
      69.09% OnlyNotEarly_SomeEarly Zero Zero
      10.61% OnlyNotEarly_SomeEarly Zero Many
       5.89% OnlyNotEarly_SomeEarly One Many
       5.52% OnlyNotEarly_SomeEarly One One
       3.82% OnlyNotEarly_SomeEarly Many Zero
       2.70% OnlyNotEarly_SomeEarly One Zero
       1.26% OnlyNotEarly_SomeEarly Many Many
       1.01% OnlyNotEarly_SomeEarly Many One
       0.10% OnlyNotEarly_SomeEarly Zero One
```

#### With #1201

```
      TickArrivalTimeStats (10000 in total):
      18.16% OnlyNotEarly_SomeEarly One Many
      16.79% OnlyNotEarly_SomeEarly Zero Zero
      15.15% OnlyNotEarly_SomeEarly Zero Many
      14.41% OnlyNotEarly_SomeEarly One Zero
      13.38% OnlyNotEarly_SomeEarly One One
      12.87% OnlyNotEarly_SomeEarly Many Zero
       4.72% OnlyNotEarly_SomeEarly Many Many
       4.52% OnlyNotEarly_SomeEarly Many One
```

While better, the `Zero Zero` case still occurs ~17% of the time, so I guess `startTick` could be even smaller on average.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs
fraser-iohk: oops, was faffing around with `decayFactor` and accidentally committed it while it did nothing lmao

I don't really think this a great function, and it probably shouldn't be reused elsewhere? `hedgehog`'s `Range.exponential` is exactly what we want here, but I don't want to badly reimplement a bunch of `Hedgehog.Range` in our codebase 😅

amesgen: Makes sense, then let's keep this as an ad-hoc helper here for now :+1:

amesgen (approved):
LGTM

Needs eg an empty force push to make the changelog CI check go green after Javier added the https://github.com/IntersectMBO/ouroboros-consensus/labels/no%20changelog label (we should at some point try to make this check be triggered by label changes).

PR #1211, opened 2024-08-08
Expose over-the-wire encoded transaction size

crocodile-dentist:
# Description

This change introduces a new method wireSizeTx for the LedgerSupportsMempool class. It provides actual CBOR encoded transaction size as it is when transmitted over the network, which the difffusion layer could exploit.

Also note that:

IntersectMBO/cardano-ledger#4521
IntersectMBO/ouroboros-network#4926

### DIFF
--- a/ouroboros-consensus-diffusion/changelog.d/20240808_125632_marcin.wojtowicz_tx_wire_size.md
+++ b/ouroboros-consensus-diffusion/changelog.d/20240808_125632_marcin.wojtowicz_tx_wire_size.md
@@ -1,0 +1,25 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Pass wireTxSize binding to txSubmission{Inbound, Outbound} which
+  enables transaction submission handlers to conveniently work with
+  sizes of byte encoded transactions for network dissemination.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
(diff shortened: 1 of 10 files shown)


ouroboros-consensus-diffusion/changelog.d/20240808_125632_marcin.wojtowicz_tx_wire_size.md
```diff
@@ -0,0 +1,23 @@
...
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- use wireTxSize
```
jasagredo, line 16: Very succint. At least explain what this is for in one line.

ouroboros-consensus/changelog.d/20240808_101916_marcin.wojtowicz_tx_wire_size.md
```diff
@@ -0,0 +1,23 @@
...
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+### Breaking
+
+* Added txWireSize method to LedgerSupportsMempool to provide
```
jasagredo, line 21: I think it doesn't matter but we use `-` not `*`

ouroboros-consensus-diffusion/changelog.d/20240808_125632_marcin.wojtowicz_tx_wire_size.md
crocodile-dentist: Pardon, and done.

ouroboros-consensus/changelog.d/20240808_101916_marcin.wojtowicz_tx_wire_size.md
crocodile-dentist: As it goes, when in Rome... :)

Issue #1102, opened 2024-05-14
Use `criterion` instead of `tasty-bench` in `mempool-bench`


jorisdral:
See https://github.com/IntersectMBO/ouroboros-consensus/pull/1095#issuecomment-2102458996

PR #1218, opened 2024-08-15
Use  `criterion` instead of `tasty-bench`  in `mempool-bench`

RenateEilers:
# Description

Use  `criterion` instead of `tasty-bench`  in `mempool-bench`, as described in #1102 .


jasagredo:
The criterion port can't be dropped in without also reworking CI. The benchmarks job runs the executable with tasty-bench specific flags (--timeout, --baseline, --fail-if-slower), and criterion doesn't support those. tasty-bench's built-in baseline comparison is the thing that actually gates our PRs today, and criterion has no equivalent, so finishing this would mean rewriting the CI job and rebuilding the regression check ourselves (or wiring up the github-action-benchmark publishing path that's currently stubbed out). That's a fair bit of work for the payoff.

And the payoff is small. The benchmark adds synthetic TestBlock transactions to a mocked mempool, so it measures the overhead of the mempool machinery rather than realistic transaction cost. On top of that it's known to be noisy on CI runners, which is why the regression threshold is set as loose as 2x. So even in its current form it only catches gross regressions, and perRunEnv wouldn't change that picture enough to justify the migration and CI rework.

Given all that, we'd rather leave the benchmark on tasty-bench as-is and close this out. Happy to revisit if we ever decide to invest in proper benchmark publishing and trend tracking, since that's the context where criterion's model would actually pull its weight.

PR #1219, opened 2024-08-16
db-analyser: support reapplication in `--benchmark-ledger-ops`

amesgen:
The first commit is just boring refactoring.

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
@@ -49,11 +49,12 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.Forecast (forecastFor)
 import           Ouroboros.Consensus.HeaderValidation (HasAnnTip (..),
-                     HeaderState (..), headerStatePoint, tickHeaderState,
-                     validateHeader)
-import           Ouroboros.Consensus.Ledger.Abstract (LedgerCfg, LedgerConfig,
-                     applyBlockLedgerResult, applyChainTick, tickThenApply,
-                     tickThenApplyLedgerResult, tickThenReapply)
+                     HeaderState (..), headerStatePoint, revalidateHeader,
+                     tickHeaderState, validateHeader)
+import           Ouroboros.Consensus.Ledger.Abstract
+                     (ApplyBlock (reapplyBlockLedgerResult), LedgerCfg,
+                     LedgerConfig, applyBlockLedgerResult, applyChainTick,
+                     tickThenApply, tickThenApplyLedgerResult, tickThenReapply)
 import           Ouroboros.Consensus.Ledger.Basics (LedgerResult (..),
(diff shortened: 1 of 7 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -569,8 +570,10 @@ benchmarkLedgerOps ::
      ( HasAnalysis blk
      , LedgerSupportsProtocol blk
      )
-  => Maybe FilePath -> Analysis blk StartFromLedgerState
-benchmarkLedgerOps mOutfile AnalysisEnv {db, registry, startFrom, cfg, limit} = do
+  => Maybe FilePath
+  -> LedgerApplicationMode
```
amesgen, line 574: Currently, the labels/field names in the CSV/JSON output of this run isn't influenced by this. Is this acceptable? The mode that is used is at least printed at the beginning of every run, e.g.
```
Started BenchmarkLedgerOps (Just "timings.csv") LedgerReapply
```

dnadales: Do you mean that the labels or field names do not indicate which kind of application was performed (full validation or reapply)? If so, I think it's fine. We could consider adding this to the run's metadata, but that's part of beacon now. 

CC: @mgmeier

dnadales (approved):

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
mgmeier: Thanks for the ping. To "document the evidence" mentioning the application mode is necessary, IMHO.
I don't think the field (both CSV and `SlotDataPoint`) semantics change with that mode, so they should actually keep their name(s).

The mode should be part of the metadata, though. Out of curiosity - what was the default up to now? What will be the default if no CLI arg is given for the mode?

`beacon` is exclusively based on the `Metadata` and `SlotDataPoint` types in `db-analyser`, so documenting this for beacon should be trivial. For CSV output and its lack of metadata though... my best bet would be the filename...

amesgen: > Out of curiosity - what was the default up to now? What will be the default if no CLI arg is given for the mode?

The default was and is full application, will point that out more clearly :+1:

> `beacon` is exclusively based on the `Metadata` and `SlotDataPoint` types in `db-analyser`, so documenting this for beacon should be trivial. For CSV output and its lack of metadata though... my best bet would be the filename...

Good point, I added the information to `Metadata`, such that the JSON object will contain an extra field like this:

```json
  "ledgerApplicationMode": "full-application"
```
or like this:
```json
  "ledgerApplicationMode": "reapplication"
```

I think it is fine to not do something about the CSV output for now.

mgmeier: Thanks @amesgen , LGTM!

PR #1221, opened 2024-08-20
db-synthesizer: relax check whether a dir "looks like a ChainDB"

amesgen:
db-synthesizer has a sanity check when (over)writing a directory to check whether it actually plausibly contains a ChainDB. Previously, this check was only succeeding when the directory would contain exactly three entries named `immutable`/`ledger`/`volatile`. This is somewhat annoying, as ChainDBs usually contain the folder `gsm` as well as files like `clean`/`lock`/`protocolMagicId` in addition.

This PR relaxes the check, as discussed with @mgmeier: We now only check whether the subdirectories are a subset of `immutable`/`ledger`/`volatile`/`gsm`, and fail if any other subdirectories are present. 

Follow-up to #1206


mgmeier (approved):
perfect solution!

dnadales (approved):

PR #1222, opened 2024-08-21
Snapshot-converter

jasagredo:
Add the snapshot-converter tool

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -668,6 +668,32 @@ executable immdb-server
     ouroboros-consensus-cardano:unstable-cardano-tools,
     with-utf8,
 
+executable snapshot-converter
+  import: common-exe
+  hs-source-dirs: app
+  main-is: snapshot-converter.hs
+  build-depends:
+    base,
+    bytestring,
+    cardano-crypto-class,
+    cardano-crypto-wrapper,
+    cardano-lmdb-simple,
+    cborg,
+    filepath,
+    fs-api,
+    optparse-applicative,
+    ouroboros-consensus,
+    ouroboros-consensus-cardano,
+    ouroboros-consensus-cardano:unstable-cardano-tools,
+    serialise,
+    sop-core,
+    text,
+    transformers,
+    with-utf8,
+
+  other-modules:
+    DBAnalyser.Parsers
+
 test-suite tools-test
   import: common-test
   type: exitcode-stdio-1.0
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/BackingStore/Impl/LMDB.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/BackingStore/Impl/LMDB.hs
@@ -17,6 +17,12 @@ module Ouroboros.Consensus.Storage.LedgerDB.V1.BackingStore.Impl.LMDB (
   , newLMDBBackingStore
     -- * Errors
   , LMDBErr (..)
+    -- * Internals exposed for @snapshot-converter@
+  , DbState (..)
+  , LMDBMK (..)
+  , getDb
+  , initLMDBTable
+  , withDbStateRWMaybeNull
   ) where
 
 import           Cardano.Slotting.Slot (SlotNo, WithOrigin (At))
--- a/ouroboros-consensus-cardano/app/snapshot-converter.hs
+++ b/ouroboros-consensus-cardano/app/snapshot-converter.hs
@@ -1,0 +1,266 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE DeriveAnyClass      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+module Main (main) where
+
+import           Cardano.Crypto.Init (cryptoInit)
+import           Cardano.Tools.DBAnalyser.HasAnalysis (mkProtocolInfo)
+import qualified Codec.CBOR.Read as CBOR
+import qualified Codec.CBOR.Write as CBOR
+import           Codec.Serialise
+import           Control.Monad
+import           Control.Monad.IO.Class
+import           Control.Monad.Trans.Except
+import qualified Data.ByteString.Lazy as BSL
+import           Data.Foldable
+import           Data.Functor
+import           Data.Text hiding (foldl')
+import qualified Data.Text as Text
+import qualified Database.LMDB.Simple as LMDB
+import qualified Database.LMDB.Simple.Cursor as LMDB.Cursor
+import           DBAnalyser.Parsers
+import           Main.Utf8
+import           Options.Applicative
+import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Config
+import           Ouroboros.Consensus.Ledger.Basics
+import           Ouroboros.Consensus.Ledger.Extended
+import           Ouroboros.Consensus.Ledger.Tables.Utils
+import           Ouroboros.Consensus.Node.ProtocolInfo
+import           Ouroboros.Consensus.Storage.LedgerDB.Impl.Common
+import           Ouroboros.Consensus.Storage.LedgerDB.Impl.Snapshots
+import qualified Ouroboros.Consensus.Storage.LedgerDB.V1.BackingStore.Impl.LMDB as Disk
+import qualified Ouroboros.Consensus.Storage.LedgerDB.V1.BackingStore.Impl.LMDB.Bridge as LMDB.Bridge
+import           Ouroboros.Consensus.Util.CBOR
+import           Ouroboros.Consensus.Util.IOLike
+import           System.FilePath (isRelative)
+import           System.FS.API
+import           System.FS.API.Lazy
+import           System.FS.IO
+
+data Format
+    = Legacy
+    | Mem
+    | LMDB
+    deriving (Show, Read)
+
+data Config = Config
+    { from    :: Format
+    -- ^ Which format the input snapshot is in
+    , inpath  :: FsPath
+    -- ^ Path to the input snapshot
+    , to      :: Format
+    -- ^ Which format the output snapshot must be in
+    , outpath :: FsPath
+    -- ^ Path to the output snapshot
+    }
+
+getCommandLineConfig :: IO (Config, BlockType)
+getCommandLineConfig =
+    execParser $
+        info
+            ((,) <$> parseConfig <*> blockTypeParser <**> helper)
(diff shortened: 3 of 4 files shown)


ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+                ]
+            )
+        <*> ( fromMaybe (error "Malformed output path") . fsFromFilePath (MountPoint ".")
+                <$> strArgument
+                    ( mconcat
+                        [ help "Output dir/file"
+                        , metavar "PATH-OUT"
+                        ]
+                    )
+            )
```
jorisdral, line 102: I'd call `mode` a `format` instead.

jorisdral, line 102: What would happen if the user puts in an absolute path instead of a relative path? I'm unsure what would happen. We should either tell the user to only pass in only relative paths (and check that they do), or check that it's fine to use absolute paths

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+
+-- Helpers
+
+(</>) :: FsPath -> Text -> FsPath
+a </> b = mkFsPath (Prelude.map unpack $ fsPathToList a ++ [b])
```
jorisdral, line 107: The newest (unreleased) version of `fs-sim` actually has this helper already: https://input-output-hk.github.io/fs-sim/fs-api/System-FS-API-Types.html#v:-60--47--62-

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/BackingStore/Impl/LMDB.hs
```diff
@@ -17,6 +17,11 @@ module Ouroboros.Consensus.Storage.LedgerDB.V1.BackingStore.Impl.LMDB (
   , newLMDBBackingStore
     -- * Errors
   , LMDBErr (..)
+  , getDb
+  , DbState (..)
+  , initLMDBTable
+  , withDbStateRWMaybeNull
+  , LMDBMK (..)
```
jorisdral, line 24: These do not belong to the `Errors` section

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -668,6 +668,31 @@ executable immdb-server
     ouroboros-consensus-cardano:unstable-cardano-tools,
     with-utf8,
 
+executable db-converter
```
jorisdral, line 671: Technically: `snapshot-converter`

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+fromCodecMK :: forall v k. CodecMK k v -> LMDB.Cursor.PeekPoke k v
+fromCodecMK (CodecMK encKey encVal decKey decVal) =
+    LMDB.Cursor.PeekPoke
+        { LMDB.Cursor.kPeek = LMDB.Bridge.peekMDBVal decKey
+        , LMDB.Cursor.vPeek = LMDB.Bridge.peekMDBVal decVal
+        , LMDB.Cursor.kPoke = LMDB.Bridge.pokeMDBVal encKey
+        , LMDB.Cursor.vPoke = LMDB.Bridge.pokeMDBVal encVal
+        }
```
jorisdral, line 124: This definition exists already in the `LMDB.Bridge` module

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+    = SnapshotError ReadIncrementalErr
+    | TablesCantDeserializeError DeserialiseFailure
+    | TablesTrailingBytes
+    | UnsuitableSnapshot String
+
+instance Show Error where
+    show (SnapshotError err) = "Couldn't deserialize the snapshot. Are you running the same node version that created the snapshot? " <> show err
+    show (TablesCantDeserializeError err) = "Couldn't deserialize the tables: " <> show err
+    show TablesTrailingBytes = "Malformed tables, there are trailing bytes!"
+    show (UnsuitableSnapshot s) = "The input snapshot does not seem to correspond to the input mode:\n\t" <> s
```
jorisdral, line 136:
```suggestion
data Error
    = SnapshotError ReadIncrementalErr
    | TablesCantDeserializeError DeserialiseFailure
    | TablesTrailingBytes
    | SnapshotFormatMismatch Mode Mode

instance Show Error where
    show (SnapshotError err) = "Couldn't deserialize the snapshot. Are you running the same node version that created the snapshot? " <> show err
    show (TablesCantDeserializeError err) = "Couldn't deserialize the tables: " <> show err
    show TablesTrailingBytes = "Malformed tables, there are trailing bytes!"
    show (SnapshotFormatMismatch expected) = "The input snapshot does not seem to correspond to the input format:\n\t" <> show expected
```

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+main = withStdTerminalHandles $ do
+    cryptoInit
+    (conf, blocktype) <- getCommandLineConfig
+    case blocktype of
+        ByronBlock args   -> run conf args
+        ShelleyBlock args -> run conf args
+        CardanoBlock args -> run conf args
+  where
+    run conf args = do
+        ccfg <- configCodec . pInfoConfig <$> mkProtocolInfo args
```
jorisdral, line 256: Do we need to support byron or shelley blocks? Would it be sufficient to use the `protocolInfoCardano` instead?

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+        let fs = SomeHasFS $ ioHasFS $ MountPoint "."
+        putStrLn "Loading..."
+        eState <- runExceptT $ load conf fs ccfg
```
jorisdral, line 259: Is there a reason that the `load` code runs in `ExceptT`? It looks like the custom `Error`s in this module are fatal exceptions, so it would be fine IMO to `throwIO` them instead if we are not going to catch them.

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+    show TablesTrailingBytes = "Malformed tables, there are trailing bytes!"
+    show (UnsuitableSnapshot s) = "The input snapshot does not seem to correspond to the input mode:\n\t" <> s
+
+checkSnapshot :: Mode -> FsPath -> SomeHasFS IO -> ExceptT Error IO ()
+checkSnapshot m p (SomeHasFS fs) = case m of
+    Legacy ->
+        want (doesFileExist fs) p "Input mode is Legacy but the provided path is NOT a file"
+    Mem ->
+        mapM_
+            (\(doCheck, extra :: [Text], err) -> want (doCheck fs) (foldl' (</>) p extra) err)
```
jorisdral, line 144: Can probably benefit from extracting it into a local where clause

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+            , (doesFileExist, ["tables", "tvar"], "Input mode is Mem but the provided path DOES NOT contain a \"tables/tvar\" file")
+            ]
+    LMDB ->
+        mapM_
+            (\(doCheck, extra :: [Text], err) -> want (doCheck fs) (foldl' (</>) p extra) err)
+            [ (doesDirectoryExist, [], "Input mode is LMDB but the provided path is NOT a directory")
+            , (doesFileExist, ["state"], "Input mode is LMDB but the provided path DOES NOT contain a \"state\" file")
+            , (doesDirectoryExist, ["tables"], "Input mode is LMDB but the provided path DOES NOT contain a \"tables\" directory")
+            , (doesFileExist, ["tables", "data.mdb"], "Input mode is LMDB but the provided path DOES NOT contain a \"tables/data.mdb\" file")
+            ]
```
jorisdral, line 157: There is a bit of duplication in the messages here

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+    checkSnapshot LMDB inpath fs
+    extLedgerSt <- withExceptT SnapshotError $ readExtLedgerState fs (decodeDiskExtLedgerState ccfg) decode (inpath </> "state")
+    values <- liftIO $ do
+        dbEnv <- LMDB.openEnvironment (fsToFilePath (MountPoint ".") (inpath </> "tables")) defaultLMDBLimits
+        Disk.LMDBMK _ dbBackingTables <- LMDB.readWriteTransaction dbEnv (Disk.getDb (K2 "utxo"))
+        LMDB.readOnlyTransaction dbEnv $
+            LMDB.Cursor.runCursorAsTransaction'
+                LMDB.Cursor.cgetAll
+                dbBackingTables
+                (fromCodecMK $ getLedgerTables $ codecLedgerTables @(LedgerState blk))
```
jorisdral, line 201: `TablesCantDeserialiseError` should probably be thrown here too in case the key/value types are wrong

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+    pure (extLedgerSt `withLedgerTables` LedgerTables (ValuesMK values))
```
jorisdral, line 202: You could probably reuse the existing `BackingStore` initialisation followed by a range read, so you don't have to duplicate the code here

ouroboros-consensus-cardano/app/db-converter.hs
```diff
@@ -0,0 +1,266 @@
...
+    dbState <- LMDB.readWriteTransaction dbEnv $ LMDB.getDatabase (Just "_dbstate")
+    dbBackingTables <-
+        LMDB.readWriteTransaction dbEnv $
+            lttraverse Disk.getDb (ltpure $ K2 "utxo")
+    LMDB.readWriteTransaction dbEnv $
+        Disk.withDbStateRWMaybeNull dbState $ \case
+            Nothing ->
+                ltzipWith3A Disk.initLMDBTable dbBackingTables codecLedgerTables (projectLedgerTables state)
+                    $> ((), Disk.DbState{Disk.dbsSeq = pointSlot $ getTip state})
+            Just _ -> liftIO $ throwIO $ Disk.LMDBErrInitialisingAlreadyHasState
```
jorisdral, line 244: Idem here

ouroboros-consensus-cardano/app/db-converter.hs
jasagredo: I'd rather avoid the BackingStore here. It will open another instance of the same data.

ouroboros-consensus-cardano/app/db-converter.hs
jasagredo: I asked the user now to input relative paths, although I still don't check them. I don't know how to check them with `fs-api`

ouroboros-consensus-cardano/app/db-converter.hs
jasagredo: I am just mirroring db-analyser here. It is not really a big deal

ouroboros-consensus-cardano/app/db-converter.hs
jasagredo: Done

PR #1225, opened 2024-08-21
Mempool: reject txs that don't fit in an empty mempool

amesgen:
Follow-up to #1168 that makes sure that adding a tx exceeding the per-tx limit does not cause a deadlock which prevents txs from being added to the mempool until the node is restarted.

We accomplish this by validating such transactions and relying on the per-tx limit to reject them.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
@@ -189,7 +189,21 @@ pureTryAddTx cfg txSize wti tx is
   , let curTotalRefScriptSize = isTotalRefScriptSize is
         newTxRefScriptSize    = txRefScriptSize cfg (isLedgerState is) tx
         maxTotalRefScriptSize = 1024 * 1024 -- 1MiB
-  , curTotalRefScriptSize + newTxRefScriptSize Prelude.<= maxTotalRefScriptSize
+        -- In case the tx exceeds the per-tx limit, let it be rejected by tx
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -189,7 +189,17 @@ pureTryAddTx cfg txSize wti tx is
   , let curTotalRefScriptSize = isTotalRefScriptSize is
         newTxRefScriptSize    = txRefScriptSize cfg (isLedgerState is) tx
         maxTotalRefScriptSize = 1024 * 1024 -- 1MiB
-  , curTotalRefScriptSize + newTxRefScriptSize Prelude.<= maxTotalRefScriptSize
+        mempoolStaysBelowCapacity =
+          curTotalRefScriptSize + newTxRefScriptSize Prelude.<= maxTotalRefScriptSize
```
nfrisby, line 193: Just a note, perhaps worth including as a comment.

There is potential for overflow in this check, which would cause a false `False` in this first disjunct. But it could in practice only overflow if `newTxRefScriptSize` were huge, which would set the second disjunct to `True`. So it's fine as-is.

nfrisby (approved):
Approved. Only one suggestion for a small addition in the comment, regarding overflow.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
amesgen: Changed accordingly :+1:

PR #1227, opened 2024-08-25
mempool-bench: move more pre-calculation into withResource

nfrisby:
PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1175 makes a change that alters the cost of the size of the pre-generated txs. This PR pre-emptively removes that cost from the benchmarked section of code.

### DIFF
--- a/ouroboros-consensus/bench/mempool-bench/Main.hs
+++ b/ouroboros-consensus/bench/mempool-bench/Main.hs
@@ -54,31 +54,35 @@ main = withStdTerminalHandles $ do
           where
             benchAddNTxs n =
                 withResource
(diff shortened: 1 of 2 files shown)


jasagredo (approved):
Changes in benchmarks and tests do not need a corresponding changelog entry

amesgen (approved):
I think this will be superseded by #1102 (https://github.com/IntersectMBO/ouroboros-consensus/pull/1095#issuecomment-2102458996) by moving the entire mempool construction out of the measurement, but I see no reason not to merge this PR now.

dnadales (approved):

PR #1228, opened 2024-08-26
Bump `fs-sim` and `fs-api` dependencies to `^>=0.3`

jorisdral:
# Description

`fs-sim` and `fs-api` are now released on Hackage. This PR updates dependencies to start using the Hackage versions.


cabal.project
```diff
@@ -53,3 +53,12 @@ if impl(ghc >= 9.10)
...
     , plutus-ledger-api ^>=1.31
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/fs-sim
+  tag: 984e07de44c3aa42f3720b18aa6307d83851b165
+  --sha256: 0kfhmdmyy8gqxzbrsm5ayvhv9hbiacvwfh36l4liw2nwjfp8ipz5
+  subdir:
+    fs-api
+    fs-sim
```
jorisdral, line 64: I'll remove this once we the Hackage versions are downloadable -- currently, the packages are not yet visible

jasagredo (approved):
Unsure why you need the TMVar in the NodeDBs, but its fine.

jorisdral:
> Unsure why you need the TMVar in the NodeDBs, but its fine.

Because the simulated `HasFS` now requires you to pass in a `TMVar` instead of a `TVar`, it's a breaking change in `fs-sim-0.3.0.0`

PR #1230, opened 2024-08-27
Mempool test: generate txs larger than the entire mempool 

### LINKED ISSUE #1226: Test that we never block on adding a tx to an empty mempool
Since #49, adding a tx to the mempool is guarded by a lock
https://github.com/IntersectMBO/ouroboros-consensus/blob/6a8def970161c599bdb90909f36ae615950eeb76/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs#L76
Therefore, while a tx is waiting to be added to the mempool, no other tx can enter the mempool. In order to avoid a deadlock, it must be guaranteed that **adding a tx can not block forever**. Such a deadlock is problematic both for the local node (as a restart is necessary), as well as for the entire network, as only empty blocks might otherwise be minted.

This requirement can be deduced from the following two properties:

 1. Transactions are removed from the mempool over time due to transactions becoming invalid (most commonly, due to being included in a minted block).
 2. Add a transaction to an *empty* mempool never blocks.

As the mempool will eventually become empty, no transaction can block forever. 

The goal of this issue is to test the second property.

---

At the moment (6a8def970161c599bdb90909f36ae615950eeb76), this property is true for the block size capacity due to #21: As the capacity of the mempool is positive, the conditional `mempoolsize < mempoolcapacity` is always true when adding a tx to an empty mempool.

See #1225 for a recent change restoring this behavior for the ad-hoc reference scripts size limit (added in #1166) after #1168.

As of #1175, this property will be guaranteed by performing the per-tx size checks as part of the check whether to add the tx to the mempool, as the per-tx limit is smaller than the per-block limit, and the mempool capacity encompasses at least one block.

---

One strategy to test this would be to enrich the `TestBlock` used in `Test.Consensus.Mempool` with a per-tx size limit, and then add a simple test which tries to add overly large txs to an empty mempool.

amesgen:
Closes #1226

In addition to the previously valid/invalid txs (purely based on the UTxO ledger rules), we add an optional per-tx limit to the mock block.

As a second step, we generate very large txs that are larger than an entire mempool, in order to test that we do *not* block when adding them (just like the other txs), which is important as explained in #1226.

---

One way to validate this PR is to introduce a bug that would cause us to block on such transactions, and observe that the test now indeed catches that.

For example, retrying when the per-tx limited is not satisfied (this is basically what happened in #1168 and was fixed by #1225) via
```diff
diff --git a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
index 372ea15c29..31abe25fbf 100644
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
@@ -170,25 +170,8 @@ pureTryAddTx ::
   -> TryAddTx blk
 pureTryAddTx cfg wti tx is =
   case runExcept $ txMeasure cfg (isLedgerState is) tx of
-    Left err ->
-      -- The transaction does not have a valid measure (eg its ExUnits is
-      -- greater than what this ledger state allows for a single transaction).
-      --
-      -- It might seem simpler to remove the failure case from 'txMeasure' and
-      -- simply fully validate the tx before determining whether it'd fit in
-      -- the mempool; that way we could reject invalid txs ASAP. However, for a
-      -- valid tx, we'd pay that validation cost every time the node's
-      -- selection changed, even if the tx wouldn't fit. So it'd very much be
-      -- as if the mempool were effectively over capacity! What's worse, each
-      -- attempt would not be using 'extendVRPrevApplied'.
-      TryAddTx
-        Nothing
-        (MempoolTxRejected tx err)
-        (TraceMempoolRejectedTx
-         tx
-         err
-         (isMempoolSize is)
-        )
+    Left _err ->
+      NotEnoughSpaceLeft
     Right txsz
       -- Check for overflow
       --
```
or here
```diff
diff --git a/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs b/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
index 743e11bc61..e01d8cfe5a 100644
--- a/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
+++ b/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
@@ -452,10 +452,7 @@ instance TxLimits (SimpleBlock c ext) where
   -- But not 'maxbound'!, since the mempool sometimes holds multiple blocks worth.
   blockCapacityTxMeasure _cfg _st = IgnoringOverflow simpleBlockCapacity
 
-  txMeasure cfg _st =
-        fmap IgnoringOverflow
-      . checkTxSize (simpleLedgerMockConfig cfg)
-      . simpleGenTx
+  txMeasure _cfg _st = pure . IgnoringOverflow . txSize . simpleGenTx
 
 simpleBlockCapacity :: ByteSize32
 simpleBlockCapacity = ByteSize32 512
```
will cause many test failures with `FailureDeadlock [Labelled (ThreadId []) (Just "main")]'` via `io-sim`'s nice deadlock detection.

---

Stacked on top of #1175 to avoid boring rebase work

### DIFF
--- a/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/State.hs
+++ b/ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/State.hs
@@ -1,35 +1,61 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Ouroboros.Consensus.Mock.Ledger.State (
+    -- * Config for the mock ledger
+    MockConfig (..)
+  , defaultMockConfig
     -- * State of the mock ledger
-    MockError (..)
+  , MockError (..)
   , MockState (..)
   , updateMockState
   , updateMockTip
   , updateMockUTxO
+    -- * Supporting definitions
+  , checkTxSize
+  , txSize
     -- * Genesis state
   , genesisMockState
   ) where
 
 import           Cardano.Binary (toCBOR)
 import           Cardano.Crypto.Hash
-import           Codec.Serialise (Serialise)
+import           Codec.Serialise (Serialise, serialise)
 import           Control.Monad (guard)
 import           Control.Monad.Except (Except, throwError, withExcept)
+import qualified Data.ByteString.Lazy as BL
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.Typeable (Typeable)
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Ledger.SupportsMempool (ByteSize32 (..))
 import           Ouroboros.Consensus.Mock.Ledger.Address
 import           Ouroboros.Consensus.Mock.Ledger.UTxO
 import           Ouroboros.Consensus.Util (ShowProxy (..), repeatedlyM)
+import           Test.Util.Orphans.Serialise ()
+
+{-------------------------------------------------------------------------------
+  Config of the mock block
+-------------------------------------------------------------------------------}
+
+-- | Parameters needed to validate blocks/txs
+data MockConfig = MockConfig {
+    mockCfgMaxTxSize :: !(Maybe ByteSize32)
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NoThunks)
+
+defaultMockConfig :: MockConfig
+defaultMockConfig = MockConfig {
+      mockCfgMaxTxSize = Nothing
+    }
 
 {-------------------------------------------------------------------------------
   State of the mock ledger
@@ -50,6 +76,7 @@ data MockError blk =
     -- validate in the second 'SlotNo'.
   | MockUtxoError UtxoError
   | MockInvalidHash (ChainHash blk) (ChainHash blk)
+  | MockTxSizeTooBig ByteSize32 ByteSize32
   deriving (Generic, NoThunks)
 
 deriving instance StandardHash blk => Show (MockError blk)
@@ -59,13 +86,14 @@ deriving instance Serialise (HeaderHash blk) => Serialise (MockError blk)
 instance Typeable blk => ShowProxy (MockError blk) where
 
 updateMockState :: (GetPrevHash blk, HasMockTxs blk)
-                => blk
+                => MockConfig
+                -> blk
                 -> MockState blk
                 -> Except (MockError blk) (MockState blk)
-updateMockState blk st = do
+updateMockState cfg blk st = do
     let hdr = getHeader blk
     st' <- updateMockTip hdr st
-    updateMockUTxO (blockSlot hdr) blk st'
+    updateMockUTxO cfg (blockSlot hdr) blk st'
 
 updateMockTip :: GetPrevHash blk
               => Header blk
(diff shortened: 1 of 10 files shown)


amesgen:
I think accepting the mempool benchmark changes is fine as this PR doesn't actually change anything about the mempool, but rather just about the ledger rules that are being used.

nfrisby:
Doh --- I didn't review this before I rebased the target branch :'(

I'll try to rebase this once https://github.com/IntersectMBO/ouroboros-consensus/pull/1175 successfully merged.

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/State.hs
```diff
@@ -1,35 +1,61 @@
...
 import           NoThunks.Class (NoThunks)
 import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Ledger.SupportsMempool (ByteSize32 (..))
 import           Ouroboros.Consensus.Mock.Ledger.Address
 import           Ouroboros.Consensus.Mock.Ledger.UTxO
 import           Ouroboros.Consensus.Util (ShowProxy (..), repeatedlyM)
+import           Test.Util.Orphans.Serialise ()
+
+{-------------------------------------------------------------------------------
+  State of the mock cnfig
```
nfrisby, line 45: `Config of the mock block`

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
```diff
@@ -347,7 +356,7 @@ ppTestSetup TestSetup { testInitialTxs
     ["Initial contents of the Mempool:"]     <>
     (map ppTestTxWithHash testInitialTxs)    <>
     ["Total size:"]                          <>
-    [show $ foldMap txSize $ testInitialTxs] <>
+    [show $ foldMap genTxSize $ testInitialTxs] <>
```
nfrisby, line 359: minor: `<>` alignment

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
```diff
@@ -492,6 +508,7 @@ genValidTxs = go []
           (tx, ledger') <- genValidTx ledger
           go (tx:txs) (n - 1) ledger'
 
+-- | Generate a valid transaction (but ignoring any per-tx size limits).
```
nfrisby, line 511: This comment is a bit confusing. Does `mustBeValid` not check the per-tx size limit?

nfrisby (approved):
Just a few minor changes

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
amesgen: It does not as we use `testLedgerConfigNoSizeLimits` in this particular case. The current approach is to first generate "non-large" txs, and then use that to set the mempool size (already before this PR) and the per-tx limit.

nfrisby: We had a call. It all looks good. Alex is going to add a GHC-style `Note` explaining how the piece interact (ie the new config argument).

amesgen: Added the following note:
```
Note [Transaction size limit]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

An important property of the mempool is that adding a transaction that can never
fit into the mempool must not block, also see
https://github.com/IntersectMBO/ouroboros-consensus/issues/1226. We test this
while generating a TestSetupWithTxs by always including a transaction that is
larger than the entire mempool, and setting the per-tx size limit such that just
this transaction is invalid due to its size, but not impacting the validity of
any other transactions. Therefore, we disable the size limit in e.g.
'genValidTx' to only capture UTxO-related validity for them by using an
appropriate ledger config ('testLedgerConfigNoSizeLimits').
```

PR #1237, opened 2024-09-04
Run mempool benchmarks in a separate GHA job

jorisdral:
We can then mark the build+test job as required, while leaving the benchmark job as non-required


amesgen (approved):
Thanks :+1:

PR #1241, opened 2024-09-06
db-analyser: add block size to --benchmark-ledger-ops and --show-block-header-size

nfrisby:
For example, the analysis in PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1240 can be simplified via this PR.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20240906_081159_nick.frisby_db_analyser_add_blocksize.md
+++ b/ouroboros-consensus-cardano/changelog.d/20240906_081159_nick.frisby_db_analyser_add_blocksize.md
@@ -1,0 +1,22 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-cardano/changelog.d/20240906_081159_nick.frisby_db_analyser_add_blocksize.md
```diff
@@ -0,0 +1,22 @@
...
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Added a `blockBytes` column to the output of --benchmark-ledger-ops and --show-header-size.
```
amesgen, line 22: As this is a change in `unstable-cardano-tools`, it doesn't need a changelog entry, but it also doesn't hurt.

amesgen (approved):

PR #1245, opened 2024-09-12
db-analyser: StoreLedgerStateAt multiple slots

nfrisby:


nfrisby:
This is draft because I want to enrich the command-line parser to accept either an integer or a list.

PR #1251, opened 2024-09-17
Add a new `GetFuturePParams` query

dnadales:
Closes #1204 

Depends on https://github.com/IntersectMBO/ouroboros-network/pull/4960

PR #1262, opened 2024-09-24
For node 10.0 release: Integrate ledger & o-network

lehins:
# Description

This PR integrates latest ledger and o-network for node 10.0 release. Notably, the following changed:
* Support for NodeToNodeV_12 and below was removed from o-network as these are unable to cross the HF boundary
* NodeToNodeV_14 was added to supported versions to track adoption rate
* o-network does not track pipelining support as this was only necessary in previous NodeToNode versions. To preserve generality of consensus, this functionality was moved here.
* Added new GetFuturePParams query
* Account for removal of minVersion in SL.Globals
* o-network and ledger package bumps


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -163,18 +163,15 @@ mkShelleyLedgerConfig ::
   -> EpochInfo (Except HardFork.PastHorizonException)
   -> MaxMajorProtVer
   -> ShelleyLedgerConfig era
-mkShelleyLedgerConfig genesis transCtxt epochInfo mmpv =
+mkShelleyLedgerConfig genesis transCtxt epochInfo _mmpv =
```
dnadales, line 166: Can we remove this parameter?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1988,7 +1978,7 @@ invalidBlockRejector ::
   -> Watcher m
          (WithFingerprint (HeaderHash blk -> Maybe (InvalidBlockReason blk)))
          Fingerprint
-invalidBlockRejector tracer version getIsInvalidBlock getCandidate =
+invalidBlockRejector tracer _version getIsInvalidBlock getCandidate =
```
dnadales, line 1981: It seems we could remove this parameter.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
nfrisby: I think we can. It'll inflate the diff a bit, but that's the only downside I foresee.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -342,7 +343,10 @@ instance MaxTxSizeUTxO (ShelleyEra c) where
       SL.ApplyTxError . pure
     $ ShelleyEra.UtxowFailure
     $ ShelleyEra.UtxoFailure
-    $ ShelleyEra.MaxTxSizeUTxO x y
+    $ ShelleyEra.MaxTxSizeUTxO
+    -- TODO is this in the right order?
```
nfrisby, line 347: It's called like this `validateMaybe (maxTxSizeUTxO txsz limit) $ ...`

So, `y` is the "expected" and `x` is the "supplied".

✔️ --- so please remove this TODO (and maybe rename `x` and `y` and add some Haddock on `maxTxSizeUTxO` now that the constructor is using record syntax).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -2010,10 +2000,7 @@ invalidBlockRejector tracer version getIsInvalidBlock getCandidate =
         mapM_ (uncurry disconnect)
           $ firstJust
                 (\hdr -> (hdr,) <$> isInvalidBlock (headerHash hdr))
-          $ (   case isPipeliningEnabled version of
-                    ReceivingTentativeBlocks    -> drop 1
-                    NotReceivingTentativeBlocks -> id
-            )
+          $ drop 1
```
amesgen, line 2003: I think we'd want a comment here that calls out that this `drop 1` exists because of diffusion pipelining

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -67,6 +62,7 @@ instance SupportedNetworkProtocolVersion (ShelleyBlock proto era) where
       , (NodeToClientV_15, ShelleyNodeToClientVersion7)
       , (NodeToClientV_16, ShelleyNodeToClientVersion8)
       , (NodeToClientV_17, ShelleyNodeToClientVersion9)
+      -- , (NodeToClientV_18, ShelleyNodeToClientVersion10)
```
nfrisby, line 65: Why is this commented out? Just a WIP thing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -1591,7 +1588,7 @@ checkKnownInvalid ::
   -> InternalEnv m blk arrival judgment
   -> Header blk
   -> m ()
-checkKnownInvalid cfgEnv dynEnv intEnv hdr = case scrutinee of
+checkKnownInvalid cfgEnv dynEnv intEnv hdr = case headerPrevHash hdr of
```
amesgen, line 1591: Again, I think we want a comment here that calls out that this drop 1 exists because of diffusion pipelining

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -612,10 +611,10 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout lopBucke
       bracketWithPrivateRegistry
         (chainSyncHeaderServerFollower
            (getChainDB kernel)
-           ( case isPipeliningEnabled version of
-              ReceivingTentativeBlocks    -> ChainDB.TentativeChain
-              NotReceivingTentativeBlocks -> ChainDB.SelectedChain
-           )
+           ChainDB.TentativeChain
```
nfrisby, line 614: We do not want to remove the argument from `chainSyncHeaderServerFollower`.

So please remove this comment.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -612,10 +611,10 @@ mkApps kernel Tracers {..} mkCodecs ByteLimits {..} genChainSyncTimeout lopBucke
...
+           -- ^ isPipeliningenabled was removed from o-n-api,
+           -- can we remove this argument, or eliminate the type
+           -- itself unless it is used for other purposes, eg. tests?
```
amesgen, line 617: We want to keep the `ChainType` type (of which `TentativeChain` is a constructor) as it is used for different kinds of `Follower`s, but the `ChainType` argument can be removed from `chainSyncHeaderServerFollower`.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
amesgen, line 1: The following can be removed in this file due to the ledger version bump:
https://github.com/IntersectMBO/ouroboros-consensus/blob/c1a681d6d3e584d9943b8e685c7bfa12bed2697f/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L583-L585

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
nfrisby: On some level, I think this PR already goes too far, removing some logic that is necessary to be able to run a net that doesn't use block diffusion pipelining. `mainnet` does, but these Consensus are supposed to be more general than that.

For `ouroboros-consensus-diffusion`, I could agree with it, since that is glue between Consensus and the Diffusion Layer, which has itself removed the ability to distinguish. However, a separate net might have a default of no pipelining, which this code would then be insufficient for.

The changes I'm already seeing in `ouroboros-consensus` I disagree with -- just because the Diffusion Layer no longer considers pipelining support (it doesn't really assume one way or the other) doesn't mean the Consensus Layer should only support one option.

So in other words, I think the various "is-pipelining-enabled" `Bool` isomorphs should have been pulled out as a new config parameter instead of collapsed to the value used by `mainnet`.

amesgen (commented):
LGTM!

Only minor comments, which shouldn't be hard to address, but can could also be postponed to a later PR

Also needs changelog entries

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: (I agree, modulo my general caveat about the Consensus Layer retaining the ability to run without block diffusion pipelining.)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
nfrisby: That said, all of these changes are quite small, so reintroducing this ability after the fact won't be significant tech debt.

So getting the release out probably takes priority over undoing these changes.

nfrisby (dismissed):
I'm Requesting Changes only because there are couple TODO-ish comments that should not be merged.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
crocodile-dentist: Shelley...Version10 constructor is added in the next PR with Damian's commits

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
dnadales: Maybe there's not much we could do, but I find the use of the `Mismatch` constructor misleading: It's not that we expected `limit` but got `txsz`; instead, we exceeded a specific threshold value.

nfrisby: I agree. But that does seem like Ledger's decision, not ours.

nfrisby:
For the record, I know you'll be working on this first thing Monday. But Monday is a holiday for me.

So please feel free to dismiss my "Changes requested" review once the concerns described in my comments have been addressed. cc @dnadales @crocodile-dentist

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
dnadales: Another out-of-scope comment: shouldn't `MaxTxSizeUTxO` be called something like `MaxTxSizeUTxOError`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
dnadales: I guess we want more than simply a comment that states:

```
We drop the last ? header in the fragment because of diffusion pipelining
```

It'd be great if we could explain why we drop 1. I don't know if you could provide a couple of sentences @amesgen

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
dnadales: So maybe we want to leave it after all.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
dnadales: Do we want to restore [this](https://github.com/IntersectMBO/ouroboros-consensus/pull/1262/files#diff-2f5ac08e9601d9f879119ec18969c373486813f7448c70d053c5e4ff70d0f17fL1617) comment (with some modifications)?

```
    -- When pipelining, the tip of the candidate is forgiven for being an￼
    -- invalid block, but not if it extends any invalid blocks.
    -- Disconnect if the parent block of `hdr` is known to be invalid.		
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
dnadales: Or we can leave it as per https://github.com/IntersectMBO/ouroboros-consensus/pull/1262/files#r1797178683

amesgen: This particular parameter is unrelated to diffusion pipelining, so it can definitely be removed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: I think this is no longer relevant as @crocodile-dentist implemented the approach suggested in https://github.com/IntersectMBO/ouroboros-consensus/pull/1262#discussion_r1797178683.

I might add a more extensive comment why we drop exactly 1 header in a separate PR; this doesn't have to happen here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: Resolved as in https://github.com/IntersectMBO/ouroboros-consensus/pull/1262#discussion_r1798970918

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
amesgen: Resolved as in https://github.com/IntersectMBO/ouroboros-consensus/pull/1262#discussion_r1798970918

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
amesgen: It still could be removed as testing whether diffusion pipelining is (not) enabled no longer relies on the `NodeToNodeVersion`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
amesgen: Cherry-picking from #1251/merging #1251 into this PR sounds useful then :+1:

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
crocodile-dentist: Nick, I took the path of least resistance when proposing these adaptations since I'm not privy to the design intent and the issue this caused.; however, now I do see your point and I've refactored to expose pipelining support as a configuration option at the consensus layer since the diffusion layer is now agnostic to it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
crocodile-dentist: fyi, I've left it in case o-network introduced something in a future NodeToNode version and you'd need to distinguish based on it here. The patch now and later is/would be smaller, and you never know when version information could come in handy.

amesgen (approved):

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
lehins: If you look at the type of `Mismatch` it has a phantom type that indicates the less then equal relation:
```haskell
  | MaxTxSizeUTxO
      !(Mismatch 'RelLTEQ Integer)
```

https://github.com/IntersectMBO/cardano-ledger/blob/cea57499d22c8acfce2a6b2d3c5c7ba59caa147e/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs#L177-L178

So, with `Mismatch` we can now easily infer just by looking at the predicate type definition of what is the current vs expected value and their expected relationship.

> Another out-of-scope comment: shouldn't MaxTxSizeUTxO be called something like MaxTxSizeUTxOError?

It is not an *Error*, it is a predicate failure. Historically we never added *PredicateFailure* to every constructor in every predicate failure. So, no, IMHO it should not be called that.

PR #1207, opened 2024-08-05
Bump plutus and ledger dependencies for cardano-node-9.2 release

Jimbo4350:
- Bump dependencies in preparation for the cardano-node-9.2 release
- [x] updated Ledger dependencies.
- [x] updated Plutus dependencies
- [x] updated ouroboros-network dependencies,
- Propagate code changes resulting from the dependency bumps

PR #1266, opened 2024-09-26
Amend leftover changelogs from previous release

jasagredo:
In particular we broke PVP in `ouroboros-consensus-diffusion` as there were breaking changes.


amesgen (approved):
Thanks :+1:

---

FTR: These are exactly the changelog entries that were erroneously not incorporated in #1207

The only exception is `ouroboros-consensus-protocol/changelog.d/20240207_130614_fraser.murray_startup_sanity_checks.md` which documented a change that didn't actually made it into #874.

PR #1273, opened 2024-10-02
Reimplement #1265 using `modifyMVar`

jasagredo:


nfrisby (approved):
Looks good! Maybe we should also pull the tracing out of the critical section, since we don't control what downstream code runs when we emit a tracer.

🤞 this one helps John

PR #1284, opened 2024-10-16
Use mlocked KES

tdammers:
# Description

Integrates the new mlocked KES functionality from `cardano-crypto-class`.

In the process, we also change the way KES keys are passed into block forging threads.


tdammers:
Superseded by #1374.

PR #1285, opened 2024-10-18
Praos headers validation properties and generators

abailly:
# Description

This PR introduces generators, properties, and an executable to be able to QuickCheck Praos headers validation in a somewhat exhaustive and explicit way. Here is a brief summary: 

* Generators produce `GeneratorContext` with various keys and parameters used in the forging of headers, and valid Praos `Header StandardCrypto` from such a context
* From a valid `Header` one can apply a `Mutation` that invalidates one particular rule for headers validation, using `Gen`erators too
* `Header` and `GeneratorContext` have JSON serialisers, with keys and bytes being output as base64-encoded strings
* the `gen-header` program can be used to generate a `Sample` of headers and context to be stored as part of a regression test suite, and to provide reference test vectors for other projects interested in validating headers

The JSON output looks like the following:

```
{
  "context": {
    "coldSignKey": "WCDQ24vfOrWlhgg65ZbHHxtyBt+QAGBoUx+IhgP1jYdHMg==",
    "kesSignKey": "WQJgmgx5dL86RoXLVEB18TH...=",
    "nonce": "ggFYII+ejLTnuete4jqw0uaCyPDzh1xOLhiTO/RSqyZ3v/wt",
    "praosSlotsPerKESPeriod": 9496,
    "vrfSignKey": "WEBzzMvKExinGbhtUaOcCMvRUw77NVnEfq85OV+MSvEb5m+NMN2tyMLgmaZNOhuAifLI3K/FQVZkEU8Rmb7A6EAj"
  },
  "headers": [
    {
      "header": "gooaAK6SORoAud8OWCA+ah5Aq6Nyw4uKwFeVWEMJ4Dw1tcOrDFFciVxibi06a...=",
      "mutation": "NoMutation"
    },
    {
      "header": "gooaAEIEphoA2MMuWCDqcIOxbDmAD0BU8mv3D7DauIjLAGG/o301UMqD5...=",
      "mutation": "MutateColdKey"
    },
```

TODO:

* [x] Don't restyle existing files (added by @nfrisby)
* [x] (_deferred to other PR_) Complete `Mutation` type to cover all kind of potential validation errors
* [x] Include more components in the `GeneratorContext` (eg. stake distribution, KES period...) 
* [x] Add `generate` and `validate` commands to `gen-header`
* [x] Add documentation
* [ ] ~Cover more Eras (?)~

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20241029_062000_abailly_header_validation_test.md
+++ b/ouroboros-consensus-cardano/changelog.d/20241029_062000_abailly_header_validation_test.md
@@ -1,3 +1,0 @@
-### Patch
-
-- Introduce generators and properties to test header validation logic
--- a/ouroboros-consensus-cardano/changelog.d/20241029_062000_abailly_header_validation_test.md
+++ b/ouroboros-consensus-cardano/changelog.d/20241029_062000_abailly_header_validation_test.md
@@ -1,0 +1,3 @@
+### Patch
+
+- Introduce generators and properties to test header validation logic
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -356,7 +356,7 @@ test-suite shelley-test
     ouroboros-consensus:{ouroboros-consensus, unstable-consensus-testlib},
     ouroboros-consensus-cardano,
     ouroboros-consensus-diffusion:unstable-diffusion-testlib,
-    ouroboros-consensus-protocol,
+    ouroboros-consensus-protocol:ouroboros-consensus-protocol,
     sop-core,
     strict-sop-core,
     tasty,
@@ -497,6 +497,7 @@ library unstable-cardano-tools
     Cardano.Tools.DBTruncater.Run
     Cardano.Tools.DBTruncater.Types
     Cardano.Tools.GitRev
+    Cardano.Tools.Headers
     Cardano.Tools.ImmDBServer.Diffusion
     Cardano.Tools.ImmDBServer.MiniProtocols
 
@@ -553,6 +554,7 @@ library unstable-cardano-tools
     ouroboros-consensus-cardano,
     ouroboros-consensus-diffusion ^>=0.18,
     ouroboros-consensus-protocol ^>=0.9,
+    ouroboros-consensus-protocol:unstable-protocol-testlib,
     ouroboros-network,
     ouroboros-network-api,
     ouroboros-network-framework ^>=0.14,
@@ -662,9 +664,30 @@ test-suite tools-test
   hs-source-dirs: test/tools-test
   main-is: Main.hs
   build-depends:
+    aeson,
     base,
     ouroboros-consensus:{ouroboros-consensus, unstable-consensus-testlib},
-    ouroboros-consensus-cardano,
+    ouroboros-consensus-cardano:{ouroboros-consensus-cardano,unstable-cardano-tools},
+    ouroboros-consensus-protocol:unstable-protocol-testlib,
+    QuickCheck,
     tasty,
     tasty-hunit,
+    tasty-quickcheck,
+    text,
     unstable-cardano-tools,
+  other-modules:
+    Test.Cardano.Tools.Headers
+
+executable gen-header
+  import: common-exe
+  hs-source-dirs: app
+  main-is: gen-header.hs
+  build-depends:
+    base,
+    cardano-crypto-class,
+    optparse-applicative,
+    ouroboros-consensus-cardano:unstable-cardano-tools,
+    with-utf8,
+  other-modules:
+    GenHeader.Parsers
+    Paths_ouroboros_consensus_cardano
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -356,7 +356,7 @@ test-suite shelley-test
     ouroboros-consensus:{ouroboros-consensus, unstable-consensus-testlib},
     ouroboros-consensus-cardano,
     ouroboros-consensus-diffusion:unstable-diffusion-testlib,
-    ouroboros-consensus-protocol:ouroboros-consensus-protocol,
+    ouroboros-consensus-protocol,
     sop-core,
     strict-sop-core,
     tasty,
@@ -553,8 +553,8 @@ library unstable-cardano-tools
     ouroboros-consensus ^>=0.21,
     ouroboros-consensus-cardano,
     ouroboros-consensus-diffusion ^>=0.18,
-    ouroboros-consensus-protocol ^>=0.9,
     ouroboros-consensus-protocol:unstable-protocol-testlib,
+    ouroboros-consensus-protocol ^>=0.9,
     ouroboros-network,
     ouroboros-network-api,
     ouroboros-network-framework ^>=0.14,
@@ -664,17 +664,18 @@ test-suite tools-test
   hs-source-dirs: test/tools-test
   main-is: Main.hs
   build-depends:
+    QuickCheck,
     aeson,
     base,
     ouroboros-consensus:{ouroboros-consensus, unstable-consensus-testlib},
-    ouroboros-consensus-cardano:{ouroboros-consensus-cardano,unstable-cardano-tools},
+    ouroboros-consensus-cardano:{ouroboros-consensus-cardano, unstable-cardano-tools},
     ouroboros-consensus-protocol:unstable-protocol-testlib,
-    QuickCheck,
     tasty,
     tasty-hunit,
     tasty-quickcheck,
     text,
     unstable-cardano-tools,
+
   other-modules:
     Test.Cardano.Tools.Headers
 
@@ -688,8 +689,10 @@ executable gen-header
     optparse-applicative,
     ouroboros-consensus-cardano:unstable-cardano-tools,
     with-utf8,
+
   other-modules:
     GenHeader.Parsers
     Paths_ouroboros_consensus_cardano
+
   autogen-modules:
     Paths_ouroboros_consensus_cardano
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -691,3 +691,5 @@ executable gen-header
   other-modules:
     GenHeader.Parsers
     Paths_ouroboros_consensus_cardano
+  autogen-modules:
+    Paths_ouroboros_consensus_cardano
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
@@ -2,10 +2,14 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications #-}
-{-# OPTIONS_GHC -Wno-missing-export-lists #-}
 
 -- | Tooling to generate and validate (Praos) headers.
-module Cardano.Tools.Headers where
+module Cardano.Tools.Headers (
+    Options (..)
+  , ValidationResult (..)
+  , run
+  , validate
+  ) where
 
 import           Cardano.Crypto.DSIGN (deriveVerKeyDSIGN)
 import           Cardano.Crypto.VRF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
@@ -8,12 +8,9 @@
 module Cardano.Tools.Headers where
 
 import           Cardano.Crypto.DSIGN (deriveVerKeyDSIGN)
-import           Cardano.Crypto.Hash (Blake2b_256, hashToBytes)
 import           Cardano.Crypto.VRF
                      (VRFAlgorithm (deriveVerKeyVRF, hashVerKeyVRF))
-import           Cardano.Ledger.Api (ConwayEra, StandardCrypto, VRF)
-import           Cardano.Ledger.BaseTypes (BoundedRational (boundRational),
-                     PositiveUnitInterval, mkActiveSlotCoeff)
+import           Cardano.Ledger.Api (ConwayEra, StandardCrypto)
 import           Cardano.Ledger.Coin (Coin (..))
 import           Cardano.Ledger.Compactible (toCompact)
 import           Cardano.Ledger.Keys (VKey (..), hashKey)
@@ -22,17 +19,12 @@ import           Cardano.Prelude (ExitCode (..), exitWith, forM_, hPutStrLn,
                      stderr)
 import           Control.Monad.Except (runExcept)
 import qualified Data.Aeson as Json
-import qualified Data.ByteString.Base16 as Hex
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Map as Map
 import           Data.Maybe (fromJust)
-import           Data.Text (unpack)
-import           Data.Text.Encoding (decodeUtf8)
-import           Debug.Trace (trace)
 import           Ouroboros.Consensus.Block (validateView)
 import           Ouroboros.Consensus.Protocol.Praos (Praos,
                      doValidateKESSignature, doValidateVRFSignature)
-import qualified Ouroboros.Consensus.Protocol.Praos.Views as Views
 import           Ouroboros.Consensus.Shelley.HFEras ()
 import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock,
                      mkShelleyHeader)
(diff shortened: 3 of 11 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
```diff
@@ -0,0 +1,89 @@
...
+    sample <- generateSamples
+    LBS.putStr $ Json.encode sample <> "\n"
+
+data ValidationResult = Valid Mutation | Invalid Mutation String
+    deriving (Eq, Show)
+
+validate :: GeneratorContext -> MutatedHeader -> ValidationResult
+validate context MutatedHeader{header, mutation} =
+    case (runExcept $ validateKES >> validateVRF, mutation) of
+        (Left err, mut) | ctor err == expectedError mut -> Valid mut
```
nfrisby, line 66: It seems preferable to avoid the stringification currently involved in this equivalence check.

Have you already assessed how easy it is to use the existing error types?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
```diff
@@ -0,0 +1,89 @@
...
+        (Right _, NoMutation) -> Valid NoMutation
+        (Right _, mut) -> Invalid mut $ "Expected error from mutation " <> show mut <> ", but validation succeeded"
+  where
+    GeneratorContext{praosSlotsPerKESPeriod, nonce, coldSignKey, vrfSignKey} = context
+    -- TODO: get these from the context
+    maxKESEvo = 63
+    coin = fromJust . toCompact . Coin
+    slotCoeff = mkActiveSlotCoeff $ fromJust $ boundRational @PositiveUnitInterval $ 1
+    ownsAllStake vrfKey = IndividualPoolStake 1 (coin 1) vrfKey
+    poolDistr = Map.fromList [(poolId, ownsAllStake hashVRFKey)]
```
nfrisby, line 77: Perhaps helpful to note that the generator context doesn't necessarily need a full stake distribution: it only needs the relative stake of the vrf keys in the headers you expect that context to validate.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
```diff
@@ -0,0 +1,89 @@
...
+
+run :: Options -> IO ()
+run Options = do
+    sample <- generateSamples
+    LBS.putStr $ Json.encode sample <> "\n"
+
+data ValidationResult = Valid Mutation | Invalid Mutation String
+    deriving (Eq, Show)
+
+validate :: GeneratorContext -> MutatedHeader -> ValidationResult
```
nfrisby, line 63: Do you have a sense yet of how difficult it would be to reuse the implementation's validation logic instead of re-specifying it here? Maybe we could do some rearranging upstream (eg factor out a "worker") to make it easier to invoke in this incomplete calling context.

Unless you were intending for this code to be independent?

nfrisby: Esgen also mentioned that maybe the Agda spec would be amenable to reuse here?

nfrisby: EG you're missing the `chainChecks` and the opcert checks. But I suppose that makes sense considering you're not so much interested in the actual incoming ledger/header state.

But maybe your generator could also generate a header state and the ledger view alongside each header? IE all three ingredients necessary to invoke `Ouroboros.Consensus.HeaderValidation.validateHeader`.

nfrisby:
> * [ ] Cover more Eras (?) 

So far, headers are very much specific to protocols more so than era. Byron (RealPBFT), Shelley (TPraos), and Babbage (proper Praos) were the only eras that altered the protocol.

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -0,0 +1,306 @@
...
+genHeaderBody :: GeneratorContext -> Gen (HeaderBody StandardCrypto, KESPeriod)
+genHeaderBody context = do
+    hbBlockNo <- BlockNo <$> arbitrary
+    hbSlotNo <- SlotNo . getPositive <$> arbitrary
+    hbPrev <- BlockHash . HashHeader <$> genHash
+    let hbVk = VKey $ deriveVerKeyDSIGN coldSignKey
+    let rho' = mkInputVRF hbSlotNo nonce
+        hbVrfRes = VRF.evalCertified () rho' vrfSignKey
+        hbVrfVk = deriveVerKeyVRF vrfSignKey
+    hbBodySize <- choose (1000, 90000)
```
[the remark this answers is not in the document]
abailly: Yes, certainly. There's probably something like a `GeneratorConfig` on top of `GeneratorContext` to be defined.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
abailly: Yes, this is already fixed in a newer version with `expectedError :: Mutation -> (PraosValidationErr -> Bool) -> Bool`

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
abailly: I agree the `validate` function in its current form is cumbersome, but ultimately it should just be a thin wrapper over calling the VRF and KES validation functions from `Praos` module. Creating the necessary environment to directly call `validateHeader` seemed like a more ambitious task at this stage, and something that could be done in a later PR to avoid "scope creep" as there really are 2 independent stages in the header validation.

re Agda spec: that's definitely something I think would be interesting to reuse as testing against an independent specification of the validation logic makes a lot of sense. This means the specification should be modular enough and separate the validation steps perhaps?

abailly:
Thanks a lot @nfrisby for taking the time to review this draft, it comforts me this can be useful and gives useful directions to improve it. The reason for this comment about eras is that the code pretty much pins down the block header's era to `Conway` which might or might not be desirable.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
amesgen: > re Agda spec: that's definitely something I think would be interesting to reuse as testing against an independent specification of the validation logic makes a lot of sense. This means the specification should be modular enough and separate the validation steps perhaps?

The spec is structured in STS style, with a DAG of rules, see e.g. Figure 6 in the Agda spec [here](https://ouroboros-consensus.cardano.intersectmbo.org/docs/about-ouroboros/References). I guess that when exposing the spec as runnable Haskell functions (which I think shouldn't be *that* far off after #1270), this could be modular to e.g. have one function per STS sub-rule.[^1]

While being more granular that what is present in the current implementation, there is not directly something that corresponds exactly to VRF + KES validation. OCERT corresponds to KES validation, and PRTCL does VRF validation, but also a few more things like nonce evolution (not very complicated).

[^1]: (Of course, we could also structure the existing Haskell code in such a way that different STS sub-rules can be invoked independently.

abailly: Thanks for the links. Of course, modelling the tests after the specification's transitions or logic rules would be desirable. BTW, having a hyperlinked HTML rendering of the Agda spec would be much more convenient than a PDF. Or did I miss the link somehow?

amesgen: > BTW, having a hyperlinked HTML rendering of the Agda spec would be much more convenient than a PDF. Or did I miss the link somehow?

Good point, we don't have that ATM, might to it later when I find the time :+1:

abailly:
Took me a while to have CI passing and fix some issues in the generators' logic but I think it's now ready for a proper review. 
While the mutations do not cover all the validation logic, I think it's better to have a first partial version, and complete it in a later PR.

ouroboros-consensus-cardano/changelog.d/20241029_062000_abailly_header_validation_test.md
jasagredo, line 1: No need to add this changelog. We don't keep track of test code, nor `unstable-*` libraries or executables in changelogs. You probably want to add the `no changelog` label to make the check pass in this case.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -662,9 +664,35 @@ test-suite tools-test
...
+    optparse-applicative,
+    ouroboros-consensus-cardano:unstable-cardano-tools,
+    with-utf8,
+
+  other-modules:
+    GenHeader.Parsers
+    Paths_ouroboros_consensus_cardano
+
+  autogen-modules:
+    Paths_ouroboros_consensus_cardano
```
jasagredo, line 698: We don't usually include the `Paths_` modules in the cabal files.

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -0,0 +1,463 @@
...
+instance Json.FromJSON Mutation where
+    parseJSON = \case
+        "NoMutation" -> pure NoMutation
+        "MutateKESKey" -> pure MutateKESKey
+        "MutateColdKey" -> pure MutateColdKey
+        "MutateKESPeriod" -> pure MutateKESPeriod
+        "MutateKESPeriodBefore" -> pure MutateKESPeriodBefore
+        "MutateCounterOver1" -> pure MutateCounterOver1
+        "MutateCounterUnder" -> pure MutateCounterUnder
+        _ -> fail "Invalid mutation"
```
jasagredo, line 196: These two seem very standard, doesn't deriving the instances yield the same result?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Headers.hs
```diff
@@ -0,0 +1,77 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-missing-export-lists #-}
```
jasagredo, line 5: We don't usually enable this option. I don't know if it is enabled anywhere else.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
abailly: I need this to be able to retrieve the current version to display in https://github.com/abailly/ouroboros-consensus/blob/67917c2bc6e1c3b10960a8d1933bf2e701c5ad0b/ouroboros-consensus-cardano/app/GenHeader/Parsers.hs#L19
Is there another, better, way to do this?

jasagredo: Ah I didn't see that use, sorry. You are indeed right, if you need that then you need to declare the `Paths_` module. However I'm unsure if that would be the right version to consider or should you print some other version number. The library version number is not really representative of what protocol it understands.

abailly: What would be the best option?

abailly:
@jasagredo I have removed changelog entry, but how do I prevent the CI from failing?

jasagredo:
I mentioned it in the suggestion above. I have set the `no changelog` label now. Next time you push it will pass.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
jasagredo: Probably @amesgen is the one that knows the answer more precisely here.

abailly:
> I mentioned it in the suggestion above. I have set the `no changelog` label now. Next time you push it will pass.

I understand, point is: _I_ cannot do this. I would suggest finding another way, perhaps looking at commit messages, otherwise this will annoy external contributors.

jasagredo:
Ah I didn't realize you couldn't do it. Indeed that is annoying for contributors hmmm. We will try to come up with a better solution.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
amesgen: I think the version number is fine at least for now, especially as this tool is quite new and will probably be evolving for some time. We could also include the git rev:
https://github.com/IntersectMBO/ouroboros-consensus/blob/e86b921443bd6e8ea25e7190eb7cb5788e28f4cc/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/GitRev.hs#L11-L13

abailly: I will do it in another PR if you don't mind, this is a first stab at generators for validation and I would rather like to have this as a baseline before adding more stuff, if possible

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -0,0 +1,456 @@
...
+    let poolId = coerce $ hashKey $ VKey $ deriveVerKeyDSIGN coldSignKey
+        ocertCounters = Map.fromList [(poolId, ocertCounter)]
+    activeSlotCoeff <- genActiveSlotCoeff
+    pure $ GeneratorContext{..}
+
+genActiveSlotCoeff :: Gen ActiveSlotCoeff
+genActiveSlotCoeff = do
+    choose (1, 100) >>= \n -> pure $ activeSlotCoeff (n % 100)
+  where
+    activeSlotCoeff = mkActiveSlotCoeff . fromJust . boundRational @PositiveUnitInterval
```
jasagredo, line 369: I usually understand it better if 0-arity `Gen`erators are just the `arbitrary` method of QuickCheck instead of a separate function, and then (>0)-arity `Gen`erators are clear that they cannot be generated on their own but they need some more data to be generated before.

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -0,0 +1,456 @@
...
+    -- c0 <=  kp -  praosMaxKESEvo
+    SlotNo <$> arbitrary `suchThat` (> threshold)
+  where
+    threshold = (ocertKESPeriod + praosMaxKESEvo + 1) * praosSlotsPerKESPeriod
+
+genHash :: Gen (Hash Blake2b_256 a)
+genHash = coerce . hashWith id <$> gen32Bytes
+
+gen32Bytes :: Gen ByteString
+gen32Bytes = BS.pack <$> vectorOf 32 arbitrary
```
jasagredo, line 456: Same as above, maybe even using some newtypes.

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -0,0 +1,456 @@
...
+
+generateSamples :: Int -> IO Sample
+generateSamples n = generate (resize n genSample)
+
+-- FIXME: Should be defined according to some Era
+testVersion :: Version
+testVersion = natVersion @MaxVersion
+
+data Sample = Sample {sample :: ![(GeneratorContext, MutatedHeader)]}
+    deriving (Show, Eq)
```
jasagredo, line 86: Why is this not a newtype?

jasagredo (approved):
Just some minor comments. 

Also yesterday we discussed it in our meeting and you can actually just push an empty changelog fragment in your PR to go around the CI check. I think it makes sense as that says "I am declaring that this PR has no relevant changes worth to be included in the changelog".

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
abailly: If you don't mind I would rather avoid that dance: these generators are used only locally and adding a `newtype` declaration at this stage seems overkill to me.

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
abailly: I suspect this is a leftover from some refactoring.

ouroboros-consensus-protocol/changelog.d/20241029_062000_abailly_header_validation_test.md
```diff
@@ -0,0 +1,3 @@
+### Patch
+
+- Introduce generators and properties to test header validation logic
```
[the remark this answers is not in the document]
abailly: Sure, but these changes make sense in relationship with the fact we are testing them, and pointing out that in the changelog seems to make sense to me

ouroboros-consensus-protocol/src/unstable-protocol-testlib/Test/Ouroboros/Consensus/Protocol/Praos/Header.hs
abailly: The `Arbitrary` instance is never used, as all the tests explicitly call generator functions. I like to "drive by needs" and only write the code that's necessary, for which I have written tests.

Furthermore, I think that in this case, we are more interested in specific scenarios, and I can imagine that we might want another generator for Context, eg. to highlight specific corner cases, or test different aspects related to `Header` construction and validation.

jasagredo (approved):
You are required to sign your commits before I the PR can be merged.

jasagredo:
The error in CI is a flaky tests that has not been flaky for a very long time and we are recently seeing more instances of it. Let's pray that the signed force-push makes it go away...

abailly:
Ok, I will sign the commits

PR #1290, opened 2024-10-22
Use hackage `rawlock`

jasagredo:
Replace the in-tree `RAWLock` with the hackage version, which is safer in presence of exceptions.


jasagredo:
The `nix`-based actions will fail because hackage.nix is not yet updated.

ouroboros-consensus/changelog.d/js-rawlock.md
```diff
@@ -0,0 +1,3 @@
+## Non-breaking
+
+* Use [`rawlock`](https://hackage.haskell.org/package/rawlock) instead of the in-tree implementation.
```
amesgen, line 3: It is a breaking change according to PVP as we remove stuff, right?

amesgen (approved):
LGTM tomorrow once hackage.nix got updated :+1:

ouroboros-consensus/changelog.d/js-rawlock.md
jasagredo: Hmmmm. Technically yes. But it comes from the `Util` subtree 😄 . But yeah, will change that once I push the hackage.nix update too

Issue #416, opened 2023-10-11
Remove `TriggerHardForkAtEpoch`


amesgen:
Due to #395, we have good reason to use case is **fundamentally**/**crucially** relying on `TriggerHardForkAtEpoch`. Hence, we would like to remove them to remove them to simplify the possible trajectories of the HFC (also see [this HWW section](https://input-output-hk.github.io/ouroboros-consensus/docs/for-developers/HardWonWisdom#the-safe-zone-semantics)).

Goals of this ticket:

 - Remove the `TriggerHardForkAtEpoch` constructor, and adapt documentation accordingly.
 - Introduce a different way for users to specify that they wish to directly skip to a later era. (The current interface/configuration options could be kept as is for a migration period.)
 - Provide guidance to migrate away from the remaining use cases of `TriggerHardForkAtEpoch` mentioned in #395.

amesgen:
Today, we discussed the following idea that should supersed supersede `TriggerXXXHardForkAtEpoch` (where `XXX` is an era): Introducing configuration that would cause the ledger to increase the (major) protocol version to a specific value.

For example, the node `.yaml` file could contain configuration like
```yaml
ProtocolVersionUpdates:
- epoch: 5
  protocolVersion: 10
```
and the Ledger would hence update the major protocol version to 10 in epoch 5[^1].

This way, "scripted" era transitions on a testnet (e.g. Babbage at protocol version `8` for the first two epochs, then Conway at protocol version `9` for two epochs, then Conway at protocol version `10` for Chang+1 which unlocks new Plutus functionalities) are easily possible without actually having to submit any governance-related transactions. In particular, intra-era HFs (like to Chang+1 in the example) are not possible with the current `TriggerHardForkAtEpoch` mechanism.

The downside of this approach is that it makes Consensus/the HFC simpler at the expense of making Ledger more complex, so this definitely requires further thought/discussion whether this pays its weight.

[^1]: This requires a bit of thought re what should happen if the protocol version is already larger than 10, or whether this update should take precedence over a regular governance-induced update to the protocol version.

nfrisby:
I'd summarize the state of https://github.com/IntersectMBO/ouroboros-consensus/issues/395 and this 416 Issue as follows.

1. See Issue 395 for reports of how this version is being used: mostly to skip eras in epoch 0, but also a few scheduled transitions in epoch 1 or 2 eg.

[...]

PR #1295, opened 2024-10-28
HFC: generalize cross era ticking

### LINKED ISSUE #345: [FEAT] - Generalize how the HFC handles era transitions

amesgen:
Closes #345 

Essentially a revived version of #358


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -117,53 +120,110 @@ instance CanHardFork xs => IsLedger (LedgerState (HardForkBlock xs)) where
...
-            --   applicable here.
-            State.mostRecentTransitionInfo cfg extended
-        , tickedHardForkLedgerStatePerEra = l'
+            -- TODO justify
+            case State.match (State.tip st) st' of
+              -- TODO Previously, this might have been 'TransitionKnown' if the
+              -- next era uses 'TriggerHardForkAtEpoch'. It /seems/
+              -- inconsequential that we know a bit less here.
+              Left _  -> TransitionUnknown (ledgerTipSlot lst)
+              Right _ -> State.mostRecentTransitionInfo cfg st
```
amesgen, line 133: FTR this is the tricky party. There will be a difference in behavior here e.g. when someone configures `TriggerHardForkAtEpoch` with a non-zero argument. (Ideally we want to remove that functionality, see #416.)

An alternative we discussed was to let the cross-era ticking function return a `TransitionInfo`, but this also seems very ad-hoc.

The other alternative is to use `WhetherTickedOrNot` as in #358.

amesgen: There is movement on #416 due to @nfrisby's work, so it probably makes sense to wait for that.

PR #1306, opened 2024-11-14
Fixes assignment of NodeToClientV_19 to ShelleyNodeToClientVersion11

crocodile-dentist:
NodeToClientV_19 should map to ShelleyNodeToClientVersion11 for GetBigLedgerPeerSnapshot query.


jasagredo:
You need to update the golden files. Run `cabal run shelley-test -- --accept`

amesgen (changes requested):
Thanks, I missed this in #1067!

We also need to introduce `CardanoNodeToClientVersion15` and use it here
https://github.com/IntersectMBO/ouroboros-consensus/blob/a58e3fa9f5fdf949e1576f3bae798ce40de279dd/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L514
where `CardanoNodeToClientVersion15` is a copy of 
https://github.com/IntersectMBO/ouroboros-consensus/blob/a58e3fa9f5fdf949e1576f3bae798ce40de279dd/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs#L482-L494
with `ShelleyNodeToClientVersion11`.

All of this is fine because `NodeToClientV_19` hasn't been released yet.

amesgen (approved):

amesgen:
Pushed a minor fix in the tests and squashed the commits :+1:

Issue #1028, opened 2024-03-27
Change `OneEraGenTxId` to just wrap a `ShortByteString`


amesgen:
Currently, `OneEraGenTxId` (used for the HFC's `GenTxId`) is an n-ary sum:
https://github.com/IntersectMBO/ouroboros-consensus/blob/8c485b24161981d34bf3d0c1d58cafeac141d359/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs#L127
Since #1017, motivated by #1009, we already compare these based on the underlying hash, ignoring the index of the n-ary sum:
https://github.com/IntersectMBO/ouroboros-consensus/blob/8c485b24161981d34bf3d0c1d58cafeac141d359/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs#L164-L175
As mentioned in https://github.com/IntersectMBO/ouroboros-consensus/issues/1009#issuecomment-2017825556, the logical next step is to change `OneEraGenTxId` to
```haskell
newtype OneEraGenTxId xs = OneEraGenTxId { getOneEraGenTxId :: ShortByteString }
```
to get rid of the n-ary sum completely.

This will require changes to the serialization format, and we will need to keep compatibility with the previous format for some time (it might be fine to not do this for Byron).

PR #1017, opened 2024-03-25
Add `ConvertRawTxId`, use it for comparing `OneEraGenTxId`s

amesgen:
Closes #1009

PR #1311, opened 2024-11-19
Change serialization of `OneEraGenTxId` to treat it as though it's simply a `ShortByteString`

fraser-iohk:
# Description

This PR changes the serialization of `OneEraGenTxId` so that it's serialized and deserialized as an era-independent `ShortByteString` rather than as an era-tag-prefixed `ShortByteString`. As part of this, I've had to make a breaking change to the representation of `TxId (GenTx ByronBlock)`, since there is now no way to identify which kind of Byron-era transaction ID we're deserializing – `TxId`, `CertificateId`, `UpId`, or `VoteId` – and these are all now represented simply in `ByronGenTxId` as a `Hash`.


fraser-iohk:
what should happen in the case that the blessed `GenTxId` era is disabled in the config? should we pick another era arbitrarily? should there be a way to order fallbacks?

```
          *** Failed! Falsified (after 15 tests):
          WithVersion (HardForkNodeToClientEnabled HardForkSpecificNodeToClientVersion4 (EraNodeToClientEnabled ByronNodeToClientVersion1 :* EraNodeToClientEnabled ShelleyNodeToClientVersion6 :* EraNodeToClientEnabled ShelleyNodeToClientVersion9 :* EraNodeToClientEnabled ShelleyNodeToClientVersion1 :* EraNodeToClientEnabled ShelleyNodeToClientVersion5 :* EraNodeToClientDisabled :* EraNodeToClientEnabled ShelleyNodeToClientVersion5 :* Nil)) (HardForkGenTxId {getHardForkGenTxId = S (S (S (Z (WrapGenTxId {unwrapGenTxId = txid: TxId {unTxId = SafeHash "fb4ea1312616d1a61645991aa8012a3aca2178e121759fb833177201080c5673"}}))))})
          DeserialiseFailure 0 "HardForkEncoderDisabledEra (SingleEraInfo {singleEraName = \"Babbage\"})"
          When deserialising 5820fb4ea1312616d1a61645991aa8012a3aca2178e121759fb833177201080c5673
          Use --quickcheck-replay="(SMGen 1921722396195416107 8896093690400693515,14)" to reproduce.
          Use -p '/Roundtrip.SerialiseNodeToClient.roundtrip GenTxId/' to rerun this test only.
```

amesgen:
> what should happen in the case that the blessed `GenTxId` era is disabled in the config? should we pick another era arbitrarily? should there be a way to order fallbacks?

What about picking an era that is supported by all `CardanoNodeToXVersion`s, eg Shelley should work I think?

Related: The next Network release (#1314) will have #5002, such that we can remove all old `CardanoNodeToClientVersion`s that do not support all currently known eras.

fraser-iohk:
yeah, I saw the changes in network and kinda feel like this is irrelevant since we're dropping support for older `NodeToClientVersion`s. does picking Shelley have any other risks that would be relevant? I did think about messing with the `Arbitrary` instances to always pick a sensible era instead

amesgen:
> does picking Shelley have any other risks that would be relevant?

I can't think of anything; everything that uses our code shouldn't care about this at all due to #1017; and I don't think that external services (like custom tx submission clients) do either.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -530,9 +567,10 @@ instance CardanoHardForkConstraints c
       , (NodeToClientV_17, CardanoNodeToClientVersion13)
       , (NodeToClientV_18, CardanoNodeToClientVersion14)
       , (NodeToClientV_19, CardanoNodeToClientVersion15)
+      , (NodeToClientV_20, CardanoNodeToClientVersion16)
```
amesgen, line 570: If we merge this PR before the next node release, we don't need to add `NodeToClientV_20`, as `NodeToClientV_19` hasn't been released yet, so it can be used to enable this change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs
```diff
@@ -164,8 +166,41 @@ instance SerialiseHFC xs
 
 instance SerialiseHFC xs
       => SerialiseNodeToClient (HardForkBlock xs) (GenTxId (HardForkBlock xs)) where
-  encodeNodeToClient = dispatchEncoder `after` (getOneEraGenTxId . getHardForkGenTxId)
-  decodeNodeToClient = fmap (HardForkGenTxId . OneEraGenTxId) .: dispatchDecoder
+  encodeNodeToClient cc v (HardForkGenTxId (OneEraGenTxId txid)) =
```
amesgen, line 169: When we talked about this recently, IIRC we concluded that `OneEraGenTxId` should wrap a `ShortByteString` directly instead of an n-ary sum as described in #1028, right?

fraser-iohk:
added the `no changelog` label for `-diffusion`, whose only change is a new instance for `HasBlessedGenTxIdEra`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
```diff
@@ -210,6 +218,13 @@ isHardForkNodeToClientEnabled :: HardForkNodeToClientVersion xs -> Bool
 isHardForkNodeToClientEnabled HardForkNodeToClientEnabled {} = True
 isHardForkNodeToClientEnabled _                              = False
 
+
+-- absolutely horrible. should be HasLatestStableGenTxIdEra probably.
+-- especially nasty because it's a K () rather than an equivalent (but
+-- non-existent) 'Void :: (Type -> Type) -> Type'
```
amesgen, line 224: > it's a `K ()` rather than an equivalent (but non-existent) `Void :: (Type -> Type) -> Type`

That's `Proxy`, right? It's argument is poly-kinded, so can be instantiated with `Type -> Type`. However, `K ()` also sounds fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs
```diff
@@ -164,8 +165,33 @@ instance SerialiseHFC xs
...
 instance SerialiseHFC xs
       => SerialiseNodeToClient (HardForkBlock xs) (GenTxId (HardForkBlock xs)) where
-  encodeNodeToClient = dispatchEncoder `after` (getOneEraGenTxId . getHardForkGenTxId)
-  decodeNodeToClient = fmap (HardForkGenTxId . OneEraGenTxId) .: dispatchDecoder
+  -- This instance can be massively simplified after we drop support for
+  -- 'NodeToClientVersion's earlier than 'NodeToClientV_19', since we no longer
+  -- need to handle the cases where 'ShortByteString's are serialised with
+  -- an era tag ('encodeNS').
+
+  encodeNodeToClient _cc v (HardForkGenTxId (OneEraGenTxId txid)) =
```
amesgen, line 173: Can we share the NodeToClient and NodeToNode logic?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
```diff
@@ -210,6 +218,13 @@ isHardForkNodeToClientEnabled :: HardForkNodeToClientVersion xs -> Bool
...
+class HasBlessedGenTxIdEra (xs :: [Type]) where
+  blessedGenTxIdEra :: NS (K ()) xs
```
amesgen, line 226: Just brainstorming an alternative: Instead of having a typeclass for this, we could also "hardcode" this choice in the serialization logic (e.g. sth like: "use the first index when `xs` has length 1, otherwise, use the second index). The downside is that one can't customize this for a hypothetical different `HardForkBlock` that would need to customize this, but that seems unlikely.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
```diff
@@ -135,7 +137,8 @@ instance Inject GenTx where
   inject _ = injectNS' (Proxy @GenTx)
 
 instance Inject WrapGenTxId where
-  inject _ = injectNS' (Proxy @WrapGenTxId)
+  inject _ ix w = WrapGenTxId $ HardForkGenTxId $ OneEraGenTxId $
+    hcollapse $ hcmap proxySingle (K . toRawTxIdHash . unwrapGenTxId) $ injectNS ix w
```
amesgen, line 141: I was briefly confused why we `injectNS` and then immediately `hcollapse` again, but we only need to do this to get an `ConvertRawTxId` dictionary :smile: 

What about unifying the implementation here and in `Inject WrapHeaderHash`?

amesgen (commented):
Looks good! (Sorry for not sending this review earlier, just noticed it while closing tabs)

I think it makes sense to also give a self-contained overview of what changes here for external implementations of mini-protocols involving tx ids, such that they can easily adapt to this change without having to read the code (ideally, we would have CDDL specs as tracked by #7, but we aren't there yet).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
```diff
@@ -124,7 +123,8 @@ newtype OneEraEnvelopeErr           xs = OneEraEnvelopeErr           { getOneEra
 newtype OneEraForgeStateInfo        xs = OneEraForgeStateInfo        { getOneEraForgeStateInfo        :: NS WrapForgeStateInfo        xs }
 newtype OneEraForgeStateUpdateError xs = OneEraForgeStateUpdateError { getOneEraForgeStateUpdateError :: NS WrapForgeStateUpdateError xs }
 newtype OneEraGenTx                 xs = OneEraGenTx                 { getOneEraGenTx                 :: NS GenTx                     xs }
-newtype OneEraGenTxId               xs = OneEraGenTxId               { getOneEraGenTxId               :: NS WrapGenTxId               xs }
+newtype OneEraGenTxId               xs = OneEraGenTxId               { getOneEraGenTxId               :: ShortByteString                 }
```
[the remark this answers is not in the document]
fraser-iohk: should `OneEraGenTxId` and `OneEraHash` be one single type? they have the same representation 🤔

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Serialisation/Common.hs
fraser-iohk: yeah, that's a good point actually! fixed 👍 

(been working a bunch with the `barbies` library recently, and used to having `Void` from there lying around lol)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Nary.hs
fraser-iohk: lol yeah I'll change this definition to be the same as WrapHeaderHash, this is needlessly complicating it

cabal.project
```diff
@@ -57,3 +57,20 @@ if impl (ghc >= 9.12)
...
+
+source-repository-package
+  type: git
+  location: https://github.com/IntersectMBO/plutus
+  tag: be9ccfc7f8ecc6ebc577dcf3374a30530ecdb168
+  --sha256: sha256-R7t5Luc1d9l2tXKg5Jgqye+vQAEONwCrQ9/JDkFCu9M=
+  subdir:
+    plutus-core
+    plutus-ledger-api
+    plutus-tx
```
amesgen, line 69: Why do we need this?

amesgen (commented):
(There still are a few unaddressed comments from the last review.)

In particular, once we have a self-contained overview of the change as mentioned above, we might want to ping people in the node diversity channel on Discord to give them a heads up and gather feedback, as any tool implementing LocalTxSubmission or TxSubmission2 will have to adapt to this serialization simplification.

Eg maybe this would be an opportunity to contribute to https://cardano-scaling.github.io/cardano-blueprint/mempool/txsubmission2.html cc @ch1bo

PR #1314, opened 2024-11-20
Integrate `ouroboros-network` 

neilmayhew:
# Description

Integrate `ouroboros-network` in preparation for `cardano-node-10.2.0` release

See IntersectMBO/cardano-node#6040 for overall progress


jasagredo:
FYI, the cardano-ledger commit in the SRP is gone. I suspect the new one is https://github.com/IntersectMBO/cardano-ledger/commit/3fe02217a185bc76eef917b7af479f39fb07d90a or https://github.com/IntersectMBO/cardano-ledger/commit/12f6aa6f094af5dab722edf03d4ab3f4ec99aa48 (the merge commit of the [PR](https://github.com/IntersectMBO/cardano-ledger/pull/4790)).

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
```diff
@@ -34,57 +42,143 @@ import           Ouroboros.Consensus.Util.Args
...
+-- | Genesis configuration flags and low-level args, as parsed from config file or CLI
+data GenesisConfigFlags = GenesisConfigFlags
+  { gcfEnableCSJ             :: Bool
+  , gcfEnableLoEAndGDD       :: Bool
+  , gcfEnableLoP             :: Bool
+  , gcfBlockFetchGracePeriod :: Maybe Integer
+  , gcfBucketCapacity        :: Maybe Integer
+  , gcfBucketRate            :: Maybe Integer
+  , gcfCSJJumpSize           :: Maybe Integer
+  , gcfGDDRateLimit          :: Maybe DiffTime
```
lehins, line 76: Configuration should always try and use the types that will be used in the end. This way we can protect against invalid values when parsing the config instead of allowing for dangerous values that can lead to overflows. For example, by using `SlotNo`, we here we can protect against a very large value being placed in the config that will end up overflowing
```suggestion
  , gcfBlockFetchGracePeriod :: Maybe DiffTime
  , gcfBucketCapacity        :: Maybe Integer
  , gcfBucketRate            :: Maybe Integer
  , gcfCSJJumpSize           :: Maybe SlotNo
  , gcfGDDRateLimit          :: Maybe DiffTime
```

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
```diff
@@ -34,57 +42,143 @@ import           Ouroboros.Consensus.Util.Args
...
+    -- eras.
+    defaultCSJJumpSize = 2 * 2160 -- Byron forecast range
+
+    -- Limiting the performance impact of the GDD.
+    defaultGDDRateLimit        = 1.0 -- seconds
+
+    gbfcGracePeriod = fromInteger $ fromMaybe defaultBlockFetchGracePeriod gcfBlockFetchGracePeriod
+    csbcCapacity    = fromInteger $ fromMaybe defaultCapacity gcfBucketCapacity
+    csbcRate        = fromInteger $ fromMaybe defaultRate gcfBucketRate
+    csjcJumpSize    = fromInteger $ fromMaybe defaultCSJJumpSize gcfCSJJumpSize
```
lehins, line 159: @facundominguez I have a couple of suggestions and some minor concerns about this:

* `csbcCapacity` is already an `Integer`, so `fromInteger` is redundant
* using `maybe` instead of `fromMaybe` allows for default values to have the expected type.
* All conversions with `to/fromInteger` and `fromIntegral` should have type annotations, because that way it is easier to spot dangers of overflow
* `SlotNo` is backed by `Word64`, so there is danger of overflow here, so it would be better to use that type instead in `gcfCSJJumpSize`, as I suggested above.

```suggestion
    gbfcGracePeriod = maybe defaultBlockFetchGracePeriod (fromInteger @DiffTime) gcfBlockFetchGracePeriod
    csbcCapacity    = fromMaybe defaultCapacity gcfBucketCapacity
    csbcRate        = maybe defaultRate (fromInteger @Rational) gcfBucketRate
    csjcJumpSize    = maybe defaultCSJJumpSize (fromInteger @SlotNo) gcfCSJJumpSize
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
```diff
@@ -228,47 +237,39 @@ prop_leashingAttackStalling =
...
-
+  where
     dropElemsAt :: [a] -> [Int] -> [a]
-    dropElemsAt xs [] = xs
-    dropElemsAt xs (i:is) =
-      let (ys, zs) = splitAt i xs
-       in ys ++ dropElemsAt (drop 1 zs) is
+    dropElemsAt xs is' =
+      let is = Set.fromList is'
+      in map fst $ filter (\(_, i) -> not $ i `Set.member` is) (zip xs [0..])
```
lehins, line 252: @nbacquey List comprehensions seem to be more readable, more compact and won't rely on list fusion to be efficient.

Also switching to `IntSet` will be more efficient and there are special functions in containers `notMember x s`, that can be used instead of `not $ member x s`
```suggestion
      let is = IntSet.fromList is'
      in [x | (x, i) <- zip xs [0..], i `IntSet.notMember` is]
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query/Types.hs
```diff
@@ -39,7 +39,7 @@ data IndividualPoolStake c = IndividualPoolStake {
 fromLedgerIndividualPoolStake :: SL.IndividualPoolStake c -> IndividualPoolStake c
 fromLedgerIndividualPoolStake ips = IndividualPoolStake {
       individualPoolStake    = SL.individualPoolStake ips
-    , individualPoolStakeVrf = SL.individualPoolStakeVrf ips
+    , individualPoolStakeVrf = SL.fromVRFVerKeyHash $ SL.individualPoolStakeVrf ips
```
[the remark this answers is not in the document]
neilmayhew: I tried that first, but it has a big knock-on effect on the rest of the code so I decided to go for minimal disruption instead.

Also, this is a wrapper type that insulates node clients from changes in ledger so it makes sense to keep it as little changed as possible.

jasagredo:
I removed the `no changelog` label. Please only apply it at the very end when merging this iff there are changes in non-testing libraries (those whose name is not prefixed by `unstable-`) that do not require a changelog entry. Otherwise we risk introducing changes in the meantime which don't end up in the Changelog.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/Uniform.hs
jasagredo: Done in 67cf19b0258e54f1c9c33bc562fcebbbe9ec733a

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
jasagredo: Done in 67cf19b0258e54f1c9c33bc562fcebbbe9ec733a

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Genesis.hs
jasagredo: Done in 67cf19b0258e54f1c9c33bc562fcebbbe9ec733a

neilmayhew:
This is now based on #1179, and will be rebased when that is merged

amesgen (approved):

Issue #892, opened 2024-01-18
[FEAT] - checksum when deserializing the ledger snapshot file


nfrisby:
**Internal/External**
*Internal*

**Describe the feature you'd like**
The deserialization of the ledger snapshot file should fail if what was read from disk differs from what was written to disk (eg disk corruption flipped a bit).

**Describe alternatives you've considered**
This requested feature is not strictly necessary: some interaction with the rest of the network will (eventually) fail if the node has the wrong ledger state. But it could be arbitrarily later, which would be especially confusing. Also, this would limit one of the ubiquitous "possible corruption" explanations for failures (such as `VRFKeyBadProof`).

Limitation: the checksum would not ensure that the ledger state was serialized to disk correctly---only that the bitstream written to disk was the same read from disk.

nfrisby:
@jorisdral had some suggestions about how some utilities from `fs-api` might help here?

jorisdral:
We could use the file read/write functions from [`System.FS.CRC`](https://input-output-hk.github.io/fs-sim/fs-api/System-FS-CRC.html). They are used by the other storage components already, e.g., 

https://github.com/IntersectMBO/ouroboros-consensus/blob/706c4b162072a4ccf4d484007130f080f5c1a183/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs#L500

Now that we're talking about checksums, though it is orthogonal to this issue, the `System.FS.CRC` module is consensus specific and should probably be moved to consensus instead.

PR #1319, opened 2024-11-21
Calculate and compare CRC when writing and reading ledger snapshots

geo2a:
Fixes https://github.com/IntersectMBO/ouroboros-consensus/issues/892

Integration into `cardano-node`: https://github.com/IntersectMBO/cardano-node/pull/6047. This uses the branch [geo2a/issue-892-checksum-snaphot-file-release-ouroboros-consensus-0.21.0.0-backport](https://github.com/IntersectMBO/ouroboros-consensus/tree/geo2a/issue-892-checksum-snaphot-file-release-ouroboros-consensus-0.21.0.0-backport) which is the backport of this PR onto the most resent release of the `ouroboros-consensus` package.

In this PR, we change the reading and writing disk snapshots of ledger state. When a snapshot is taken and written to disk, an additional file with the `.checksum` extension is written alongside it. The checksum file contains a string that represent the CRC32 checksum of the snapshot.

The checksum is calculated incrementally, alongside writing the snapshot to disk. When a snapshot is read from dist, the checksum is again calculated and compared to the tracked one. If the checksum is different, `readSnaphot` returns the `ReadSnapshotDataCorruption` error, indicating data corruption.

The checksum is calculated incrementally, alongside reading a writing the data. On write, we use the [`hPutAllCRC`](https://input-output-hk.github.io/fs-sim/fs-api/src/System.FS.CRC.html#hPutAllCRC) function from `fs-sim`, and on read we modify the [readIncremental](https://github.com/IntersectMBO/ouroboros-consensus/blob/892-checksum-snaphot-file/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/CBOR.hs#L191) function to compute the checksum as data is read.

To enable seamless integration into `cardano-node`, we make the check optional. When initialising the ledger state from a snapshot in `initLedgerDB`, we issue a warning in case the checksum file is missing for a snapshot, but do not fail as in case of invalid snapshots.

The `db-analyser` tool ignores the checksum files by default when reading the snapshots. We add `--disk-snapshot-checksum` flag to enabled the check. When writing a snapshot to disk, e.g. as a result of the `--store-ledger` analysis, `db-analyser` will always write
calculate the checksum and write it into the snapshot's `.checksum` file.

**Tests**

There state machine test in `Test.Ouroboros.Storage.LedgerDB.OnDisk` is relevant to this feature, and has caught a number of silly mistakes in the process of its implementation, for example forgetting to delete a checksum file when the snapshot is deleted.

The model in the test does not track checksums, and I do not think it can (or should) be augmented to do that. Howerver, the `Snap` and `Restore` events are now parameterised by the checksum flag, and the values for the flag are randomised when generating these events. This leads to testing the following properties:
- this feature is backwards-compatible, i.e. the `Restore` events will always lead to restoring from a snapshot, even if `Snap` events do not write checksum files (i.e. their flag is `NoDoDiskSnapshotChecksum` ~= `False`).
- If the interpretation of the `DoDiskSnapshotChecksum` flag changes in the code base and becomes strict, i.e. hard fail if the checksum file is missing, this test will discover that.

**Effects on Performance**:

Running `db-analyser` to read a ledger snapshot and store the snapshot of the state at the next slot shows a difference of 2 seconds on my machine. See a comment below for the logs.

To precisely evaluate the effects, we need a micro-benchmark of the reading and writing of snapshots with and without the checksum calculation.

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
@@ -74,7 +74,7 @@ import qualified Ouroboros.Consensus.Storage.ImmutableDB as ImmutableDB
 import           Ouroboros.Consensus.Storage.LedgerDB (DiskSnapshot (..),
                      writeSnapshot)
 import           Ouroboros.Consensus.Storage.Serialisation (encodeDisk)
-import           Ouroboros.Consensus.Util ((..:))
+import           Ouroboros.Consensus.Util (Flag (..), (..:))
 import qualified Ouroboros.Consensus.Util.IOLike as IOLike
 import           Ouroboros.Network.SizeInBytes
 import           System.FS.API (SomeHasFS (..))
@@ -102,19 +102,19 @@ runAnalysis analysisName = case go analysisName of
       pure result
   where
     go :: AnalysisName -> SomeAnalysis blk
-    go ShowSlotBlockNo                                = mkAnalysis $ showSlotBlockNo
-    go CountTxOutputs                                 = mkAnalysis $ countTxOutputs
-    go ShowBlockHeaderSize                            = mkAnalysis $ showHeaderSize
-    go ShowBlockTxsSize                               = mkAnalysis $ showBlockTxsSize
-    go ShowEBBs                                       = mkAnalysis $ showEBBs
-    go OnlyValidation                                 = mkAnalysis @StartFromPoint $ \_ -> pure Nothing
-    go (StoreLedgerStateAt slotNo lgrAppMode)         = mkAnalysis $ storeLedgerStateAt slotNo lgrAppMode
-    go CountBlocks                                    = mkAnalysis $ countBlocks
-    go (CheckNoThunksEvery nBks)                      = mkAnalysis $ checkNoThunksEvery nBks
-    go TraceLedgerProcessing                          = mkAnalysis $ traceLedgerProcessing
-    go (ReproMempoolAndForge nBks)                    = mkAnalysis $ reproMempoolForge nBks
-    go (BenchmarkLedgerOps mOutfile lgrAppMode)       = mkAnalysis $ benchmarkLedgerOps mOutfile lgrAppMode
-    go (GetBlockApplicationMetrics nrBlocks mOutfile) = mkAnalysis $ getBlockApplicationMetrics nrBlocks mOutfile
+    go ShowSlotBlockNo                                   = mkAnalysis $ showSlotBlockNo
+    go CountTxOutputs                                    = mkAnalysis $ countTxOutputs
+    go ShowBlockHeaderSize                               = mkAnalysis $ showHeaderSize
+    go ShowBlockTxsSize                                  = mkAnalysis $ showBlockTxsSize
+    go ShowEBBs                                          = mkAnalysis $ showEBBs
+    go OnlyValidation                                    = mkAnalysis @StartFromPoint $ \_ -> pure Nothing
+    go (StoreLedgerStateAt slotNo lgrAppMode doChecksum) = mkAnalysis $ storeLedgerStateAt slotNo lgrAppMode doChecksum
+    go CountBlocks                                       = mkAnalysis $ countBlocks
+    go (CheckNoThunksEvery nBks)                         = mkAnalysis $ checkNoThunksEvery nBks
+    go TraceLedgerProcessing                             = mkAnalysis $ traceLedgerProcessing
+    go (ReproMempoolAndForge nBks)                       = mkAnalysis $ reproMempoolForge nBks
+    go (BenchmarkLedgerOps mOutfile lgrAppMode)          = mkAnalysis $ benchmarkLedgerOps mOutfile lgrAppMode
+    go (GetBlockApplicationMetrics nrBlocks mOutfile)    = mkAnalysis $ getBlockApplicationMetrics nrBlocks mOutfile
 
     mkAnalysis ::
          forall startFrom. SingI startFrom
@@ -382,8 +382,9 @@ storeLedgerStateAt ::
      )
   => SlotNo
   -> LedgerApplicationMode
+  -> Flag "DoDiskSnapshotChecksum"
   -> Analysis blk StartFromLedgerState
-storeLedgerStateAt slotNo ledgerAppMode env = do
+storeLedgerStateAt slotNo ledgerAppMode doChecksum env = do
     void $ processAllUntil db registry GetBlock startFrom limit initLedger process
     pure Nothing
   where
@@ -422,7 +423,7 @@ storeLedgerStateAt slotNo ledgerAppMode env = do
     storeLedgerState ledgerState = case pointSlot pt of
         NotOrigin slot -> do
           let snapshot = DiskSnapshot (unSlotNo slot) (Just "db-analyser")
-          writeSnapshot ledgerDbFS encLedger snapshot ledgerState
+          writeSnapshot ledgerDbFS doChecksum encLedger snapshot ledgerState
           traceWith tracer $ SnapshotStoredEvent slot
         Origin -> pure ()
       where
--- a/ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
+++ b/ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
@@ -1,0 +1,14 @@
+### Breaking
+
+- When writing a ledger state snapshot to disk, calculate the state's CRC32 checksum and write it to a separate file, which is named the same as the snapshot file, plus the `.checksum` extension.
+- When reading a snapshot file in `readSnapshot`, calculate its checksum and compare it to the value in the corresponding `.checksum` file. Return an error if the checksum is different or invalid. Issue a warning if the checksum file does not exist, but still initialise the ledger DB.
+- To support the previous item, change the error type of the `readSnapshot` from `ReadIncrementalErr` to the extended `ReadSnaphotErr`.
+- Checksumming the snapshots is controlled via the `doChecksum :: Flag "DoDiskSnapshotChecksum"` parameter of `initFromSnapshot`. Ultimately, this parameter comes from the Node's configuration file via the `DiskPolicy` data type.
+- Extend the `DiskPolicyArgs` data type to enable the node to pass `Flag "DoDiskSnapshotChecksum"` to Consensus.
+
+### Non-breaking
+
+- Make `Ouroboros.Consensus.Util.CBOR.readIncremental` optionally compute the checksum of the data as it is read.
+- Introduce an explicit `Ord` instance for `DiskSnapshot` that compares the values on `dsNumber`.
+- Introduce a new utility newtype `Flag` to represent type-safe boolean flags. See ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs.
+- Use `Flag "DoDiskSnapshotChecksum"` to control the check of the snapshot checksum file in `takeSnapshot`, `readSnapshot` and `writeSnapshot`.
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
@@ -232,9 +232,11 @@ initFromDisk LgrDbArgs { lgrHasFS = hasFS, .. }
         lgrConfig
         lgrGenesis
         (streamAPI immutableDB)
+        doDiskSnapshotChecksum
     return (db, replayed)
   where
     ccfg = configCodec $ getExtLedgerCfg $ LedgerDB.ledgerDbCfg lgrConfig
+    LedgerDB.DiskPolicyArgs _ _ doDiskSnapshotChecksum = lgrDiskPolicyArgs
 
 -- | For testing purposes
 mkLgrDB :: StrictTVar m (LedgerDB' blk)
@@ -280,11 +282,12 @@ takeSnapshot ::
      , IsLedger (LedgerState blk)
      )
   => LgrDB m blk -> m (Maybe (LedgerDB.DiskSnapshot, RealPoint blk))
-takeSnapshot lgrDB@LgrDB{ cfg, tracer, hasFS } = wrapFailure (Proxy @blk) $ do
+takeSnapshot lgrDB@LgrDB{ cfg, tracer, hasFS, diskPolicy } = wrapFailure (Proxy @blk) $ do
     ledgerDB <- LedgerDB.ledgerDbAnchor <$> atomically (getCurrent lgrDB)
     LedgerDB.takeSnapshot
       tracer
       hasFS
+      (LedgerDB.onDiskShouldChecksumSnapshots diskPolicy)
       (encodeDiskExtLedgerState ccfg)
       ledgerDB
   where
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -31,10 +33,13 @@ import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Ledger.Inspect
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.Storage.ImmutableDB.Stream
+import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy
+                     (pattern NoDoDiskSnapshotChecksum)
 import           Ouroboros.Consensus.Storage.LedgerDB.LedgerDB
 import           Ouroboros.Consensus.Storage.LedgerDB.Query
 import           Ouroboros.Consensus.Storage.LedgerDB.Snapshots
 import           Ouroboros.Consensus.Storage.LedgerDB.Update
+import           Ouroboros.Consensus.Util (Flag)
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Network.Block (Point (Point))
 import           System.FS.API
@@ -102,6 +107,7 @@ initLedgerDB ::
   -> LedgerDbCfg (ExtLedgerState blk)
   -> m (ExtLedgerState blk) -- ^ Genesis ledger state
   -> StreamAPI m blk blk
+  -> Flag "DoDiskSnapshotChecksum"
   -> m (InitLog blk, LedgerDB' blk, Word64)
 initLedgerDB replayTracer
              tracer
@@ -110,14 +116,16 @@ initLedgerDB replayTracer
              decHash
              cfg
              getGenesisLedger
-             stream = do
+             stream
+             doDoDiskSnapshotChecksum = do
     snapshots <- listSnapshots hasFS
-    tryNewestFirst id snapshots
+    tryNewestFirst doDoDiskSnapshotChecksum id snapshots
   where
-    tryNewestFirst :: (InitLog blk -> InitLog blk)
+    tryNewestFirst :: Flag "DoDiskSnapshotChecksum"
+                   -> (InitLog blk -> InitLog blk)
                    -> [DiskSnapshot]
                    -> m (InitLog blk, LedgerDB' blk, Word64)
-    tryNewestFirst acc [] = do
+    tryNewestFirst _ acc [] = do
         -- We're out of snapshots. Start at genesis
         traceWith replayTracer ReplayFromGenesis
         initDb <- ledgerDbWithAnchor <$> getGenesisLedger
@@ -126,8 +134,7 @@ initLedgerDB replayTracer
         case ml of
           Left _  -> error "invariant violation: invalid current chain"
           Right (l, replayed) -> return (acc InitFromGenesis, l, replayed)
-    tryNewestFirst acc (s:ss) = do
-        -- If we fail to use this snapshot, delete it and try an older one
+    tryNewestFirst doChecksum acc allSnapshot@(s:ss) = do
         ml <- runExceptT $ initFromSnapshot
                              replayTracer
                              hasFS
@@ -136,14 +143,23 @@ initLedgerDB replayTracer
                              cfg
                              stream
                              s
+                             doChecksum
         case ml of
+          -- If a checksum file is missing for a snapshot,
+          -- issue a warning and retry the same snapshot
+          -- ignoring the checksum
+          Left (InitFailureRead ReadSnapshotNoChecksumFile{}) -> do
+            traceWith tracer $ SnapshotMissingChecksum s
+            tryNewestFirst NoDoDiskSnapshotChecksum acc allSnapshot
+          -- If we fail to use this snapshot for any other reason, delete it and try an older one
           Left err -> do
             when (diskSnapshotIsTemporary s) $
               -- We don't delete permanent snapshots, even if we couldn't parse
               -- them
               deleteSnapshot hasFS s
             traceWith tracer $ InvalidSnapshot s err
-            tryNewestFirst (acc . InitFailure s err) ss
+            -- reset checksum flag to the initial state after failure
+            tryNewestFirst doChecksum (acc . InitFailure s err) ss
           Right (r, l, replayed) ->
             return (acc (InitFromSnapshot s r), l, replayed)
 
@@ -170,10 +186,11 @@ initFromSnapshot ::
   -> LedgerDbCfg (ExtLedgerState blk)
   -> StreamAPI m blk blk
   -> DiskSnapshot
+  -> Flag "DoDiskSnapshotChecksum"
   -> ExceptT (SnapshotFailure blk) m (RealPoint blk, LedgerDB' blk, Word64)
-initFromSnapshot tracer hasFS decLedger decHash cfg stream ss = do
+initFromSnapshot tracer hasFS decLedger decHash cfg stream ss doChecksum = do
     initSS <- withExceptT InitFailureRead $
-                readSnapshot hasFS decLedger decHash ss
+                readSnapshot hasFS decLedger decHash doChecksum ss
     let replayStart = castPoint $ getTip initSS
     case pointToWithOriginRealPoint replayStart of
       Origin -> throwError InitFailureGenesis
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -12,6 +14,7 @@
 module Ouroboros.Consensus.Storage.LedgerDB.Snapshots (
     DiskSnapshot (..)
     -- * Read from disk
+  , ReadSnapshotErr (..)
   , SnapshotFailure (..)
   , diskSnapshotIsTemporary
   , listSnapshots
@@ -34,13 +37,18 @@ import qualified Codec.CBOR.Write as CBOR
 import           Codec.Serialise.Decoding (Decoder)
 import qualified Codec.Serialise.Decoding as Dec
 import           Codec.Serialise.Encoding (Encoding)
-import           Control.Monad (forM, void)
-import           Control.Monad.Except (ExceptT (..))
+import           Control.Monad (forM, void, when)
+import           Control.Monad.Except (ExceptT (..), throwError, withExceptT)
 import           Control.Tracer
+import           Data.Bits
+import qualified Data.ByteString.Builder as BS
+import qualified Data.ByteString.Char8 as BSC
+import qualified Data.ByteString.Lazy as BSL
+import           Data.Char (ord)
 import           Data.Functor.Contravariant ((>$<))
 import qualified Data.List as List
 import           Data.Maybe (isJust, mapMaybe)
-import           Data.Ord (Down (..))
+import           Data.Ord (Down (..), comparing)
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.Word
@@ -56,6 +64,7 @@ import           Ouroboros.Consensus.Util.Enclose
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Versioned
 import           System.FS.API.Lazy
+import           System.FS.CRC (CRC (..), hPutAllCRC)
 import           Text.Read (readMaybe)
 
 {-------------------------------------------------------------------------------
@@ -66,7 +75,7 @@ data SnapshotFailure blk =
     -- | We failed to deserialise the snapshot
     --
     -- This can happen due to data corruption in the ledger DB.
-    InitFailureRead ReadIncrementalErr
+    InitFailureRead ReadSnapshotErr
 
     -- | This snapshot is too recent (ahead of the tip of the chain)
   | InitFailureTooRecent (RealPoint blk)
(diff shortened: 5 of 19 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -236,7 +267,7 @@ snapshotToPath = mkFsPath . (:[]) . snapshotToFileName
 snapshotFromPath :: String -> Maybe DiskSnapshot
 snapshotFromPath fileName = do
     number <- readMaybe prefix
-    return $ DiskSnapshot number suffix'
+    return $ DiskSnapshot number suffix' Nothing
```
jasagredo, line 270: This will be used when listing snapshots, so on startup all of them will have a CRC `Nothing` therefore making this check void. My guess is that we want to have a `<slotno>[_<suffix>].crc` file next to the snapshot and read that one maybe?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -116,14 +118,14 @@ takeSnapshot tracer hasFS encLedger oldest =
...
+            snapshot = DiskSnapshot number Nothing Nothing
         snapshots <- listSnapshots hasFS
         if List.any ((== number) . dsNumber) snapshots then
           return Nothing
         else do
-          encloseTimedWith (TookSnapshot snapshot tip >$< tracer)
-              $ writeSnapshot hasFS encLedger snapshot oldest
-          return $ Just (snapshot, tip)
+          snapshotWithCRC <- writeSnapshot hasFS encLedger snapshot oldest
+          encloseTimedWith (TookSnapshot snapshotWithCRC tip >$< tracer) $ pure ()
```
jasagredo, line 127: This `encloseTimedWith` no longer encloses the time it takes to write the snapshot.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -161,15 +163,25 @@ data DiskSnapshot = DiskSnapshot {
...
       -- number with an underscore, e.g., @4492799_last_Byron@. This suffix acts
       -- as metadata for the operator of the node. Snapshots with a suffix will
       -- /not be trimmed/.
-    , dsSuffix :: Maybe String
+    , dsSuffix   :: Maybe String
+
+    -- | Snapshots can carry a checksum, which is filled in once they
+    --   have been written to disk. When reading a snapshot, the checksum
+    --   of the read data will be checked to match to detect data corruption.
+    , dsChecksum :: Maybe CRC
```
jasagredo, line 177: My understanding is that `DiskSnapshot` represents "the name of the snapshot" and this CRC field represents the contents. I think that there should be a `DiskSnapshot` datatype and on top a:

```
data DiskSnapshotWithCRC = DiskSnapshotWithCRC {
  snapshotName :: !DiskSnapshot
  snapshotCRC :: !CRC
}
```
And whenever we have one of those, the CRC must be present (not a `Maybe`).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +192,51 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
   -> (ExtLedgerState blk -> Encoding)
   -> DiskSnapshot
-  -> ExtLedgerState blk -> m ()
+  -> ExtLedgerState blk -> m DiskSnapshot
 writeSnapshot (SomeHasFS hasFS) encLedger ss cs = do
-    withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h ->
-      void $ hPut hasFS h $ CBOR.toBuilder (encode cs)
+    withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h -> do
+      (_, crc) <- hPutAllCRC hasFS h $ CBOR.toLazyByteString (encode cs)
+      pure ss{dsChecksum = Just crc}
```
jasagredo, line 239: This can produce a CRC as a return type instead (the rest of the DiskSnapshot is actually an input parameter already). And I would argue that we don't want to return this but instead write it to a file.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
amesgen: A variant of `encloseTimedWith` that allows to return a result seems sensible, as mentioned in https://github.com/IntersectMBO/ouroboros-consensus/pull/1183#discussion_r1679251466

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -161,15 +163,25 @@ data DiskSnapshot = DiskSnapshot {
...
+    --   of the read data will be checked to match to detect data corruption.
+    , dsChecksum :: Maybe CRC
     }
-  deriving (Show, Eq, Ord, Generic)
+  deriving (Show, Eq, Generic)
+
+-- | TODO: CRC should really just have an @'Ord'@ instance as it's just a newtype
+instance Ord DiskSnapshot where
+  compare (DiskSnapshot dsNumberL dsSuffixL dsChecksumL) (DiskSnapshot dsNumberR dsSuffixR dsChecksumR) =
+    compare (dsNumberL, dsSuffixL, getCRC <$> dsChecksumL) (dsNumberR, dsSuffixR, getCRC <$> dsChecksumR)
```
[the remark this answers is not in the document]
geo2a: Indeed!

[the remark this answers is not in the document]
geo2a: We actually don't need an `Ord` instance for `DiskSnapshot` at all.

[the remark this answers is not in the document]
jasagredo: Don't we? When we startup the node we want to start with the most recent snapshot and that relies on Ord I think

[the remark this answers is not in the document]
geo2a: Right, that makes sense. I've just removed the instance and ran `cabal build`. It did not cross my mind to check if the downstream packages still build!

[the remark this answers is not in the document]
jorisdral: The `LedgerDB` itself decides the load order of snapshots, and it uses `dsNumber` only.

https://github.com/IntersectMBO/ouroboros-consensus/blob/f196bfccb0cddc4d86f6b45ff5bfa835d6549c9f/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs#L114-L115

https://github.com/IntersectMBO/ouroboros-consensus/blob/f196bfccb0cddc4d86f6b45ff5bfa835d6549c9f/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs#L217-L223

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/CBOR.hs
```diff
@@ -186,24 +187,25 @@ readIncremental :: forall m a. IOLike m
...
-    go _ _ (CBOR.R.Done leftover _ a) =
+        go hasFS h (updateCRC bs checksum) dec'
+    go _ _ checksum (CBOR.R.Done leftover _ a) =
         return $ if BS.null leftover
-                   then Right a
+                   then Right (a, checksum)
                    else Left $ TrailingBytes leftover
-    go _ _ (CBOR.R.Fail _ _ err) =
+    go _ _ _ (CBOR.R.Fail _ _ err) =
         return $ Left $ ReadFailed err
```
jorisdral, line 209: Use bang patterns on `checksum` to be extra sure that this function will unbox it -- it's a `Word32` under the hood, which can be readily unboxed

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
jorisdral: We also put checksums in a separate file in `lsm-tree`. We use a small abstraction for it:

https://github.com/IntersectMBO/lsm-tree/blob/d11c709cb3c3b91d9f66e94620017e0edd997528/src/Database/LSMTree/Internal/CRC32C.hs#L239-L262

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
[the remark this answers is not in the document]
geo2a: I've double-checked that `cardano-node` builds without the `Ord` instance for `DiskSnapshot`, so it is we can either:
- remove the instance and keep `listSnaphot` as is
- have an explicit instance that compares on `dsNumber` and make `listSnapshot` use that

I think I lean toward the second solution, please let me know if anyone has objections.

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
```diff
@@ -0,0 +1,13 @@
...
+- Rename `DiskSnapshot` to `DiskSnapshotName`; `DiskSnapshot` is now a record of a `DiskSnapshotName` and a CRC checksum of the data.
+- When writing a ledger state snapshot to disk, calculate the state's CRC checksum and write it to a separate file, which is named the same as the snapshot file, plus the `.crc` extension.
+- When reading a snapshot file, calculate its checksum and compare it to the value in the corresponding `.crc` file. Return an error if either the checksum file does not exist or the checksum is wrong.
+- To support the previous item, change the error type of the `readSnapshot` from `ReadIncrementalErr` to the extended `ReadSnaphotErr`.
+
+- Make `Ouroboros.Consensus.Util.CBOR.readIncremental` compute the checksum of the data as it is read.
+
+### Non-breaking
+
+- Introduce an explicit `Ord` instance for `DiskSnapshotName` that compares the values on `dsnNumber`.
```
jasagredo: I think `DiskSnapshotName` is gone?

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
```diff
@@ -0,0 +1,13 @@
+### Breaking
+
+- Rename `DiskSnapshot` to `DiskSnapshotName`; `DiskSnapshot` is now a record of a `DiskSnapshotName` and a CRC checksum of the data.
```
jasagredo: I think this renaming is gone?

ouroboros-consensus-cardano/changelog.d/20241127_163624_georgy.lukyanov_892_checksum_snaphot_file.md
```diff
@@ -0,0 +1,5 @@
+### Breaking
+
+- The changes in `ouroboros-consensus-cardano` renamed `DiskSnapshot` to `DiskSnapshotName` and introduced a new type `DiskSnapshot` that carries a checksum. This change is propagated here as well.
```
jasagredo: I think you mean `ouroboros-consensus` not `ouroboros-consensus-cardano`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +194,99 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
+  -> ExceptT ReadSnaphotErr m (ExtLedgerState blk)
+readSnapshot someHasFS decLedger decHash snapshotName = do
+  snapshotCRC <- readCRC someHasFS (snapshotToChecksumPath snapshotName)
+  ExceptT $
+    readIncremental someHasFS decoder (snapshotToPath snapshotName) >>= \case
+       Left e  -> pure $ Left (ReadSnaphotFailed e)
+       Right (ledgerState, checksumAsRead) ->
+           if checksumAsRead /= snapshotCRC
+           then pure $ Left ReadSnaphotDataCorruption
+           else pure (Right ledgerState)
```
jasagredo, line 228: Perhaps this could be written inside `ExceptT`?

```suggestion
  (ledgerState, checsumAsRead) <- 
    withExceptT 
      ReadSnapshotFailed 
      (readIncremental someHasFS decoder (snapshotToPath snapshotName))
  if checksumAsRead /= snapshotCRC
   then throwE ReadSnaphotDataCorruption
   else pure ledgerState
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +194,99 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
   => SomeHasFS m
   -> (ExtLedgerState blk -> Encoding)
   -> DiskSnapshot
   -> ExtLedgerState blk -> m ()
 writeSnapshot (SomeHasFS hasFS) encLedger ss cs = do
-    withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h ->
-      void $ hPut hasFS h $ CBOR.toBuilder (encode cs)
+    crc <- withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h -> do
+      (_, crc) <- hPutAllCRC hasFS h $ CBOR.toLazyByteString (encode cs)
+      pure crc
```
jasagredo:
```suggestion
    crc <- withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h -> 
      snd <$> hPutAllCRC hasFS h $ CBOR.toLazyByteString (encode cs)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -217,10 +298,13 @@ deleteSnapshot (SomeHasFS HasFS{..}) = removeFile . snapshotToPath
 -- | List on-disk snapshots, highest number first.
 listSnapshots :: Monad m => SomeHasFS m -> m [DiskSnapshot]
 listSnapshots (SomeHasFS HasFS{..}) =
-    aux <$> listDirectory (mkFsPath [])
+   aux <$> listDirectory (mkFsPath [])
```
jasagredo, line 300: I think we usually have 4 spaces before a where so I'd keep the space here, not particularly important.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -217,10 +298,13 @@ deleteSnapshot (SomeHasFS HasFS{..}) = removeFile . snapshotToPath
...
   where
     aux :: Set String -> [DiskSnapshot]
-    aux = List.sortOn (Down . dsNumber) . mapMaybe snapshotFromPath . Set.toList
+    aux = List.sortOn Down . mapMaybe snapshotFromPath . Set.toList
+
+snapshotToChecksumFileName :: DiskSnapshot -> String
+snapshotToChecksumFileName = (<> ".checksum") . snapshotToFileName
```
jasagredo, line 306: Wasn't it called `.crc`? otherwise the changelog has to change.

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
jasagredo: Ah you were faster than I was 😄

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
```diff
@@ -0,0 +1,12 @@
+### Breaking
+
+- When writing a ledger state snapshot to disk, calculate the state's CRC checksum and write it to a separate file, which is named the same as the snapshot file, plus the `.checksum` extension.
+- When reading a snapshot file, calculate its checksum and compare it to the value in the corresponding `.checksum` file. Return an error if either the checksum file does not exist or the checksum is different or invalid.
```
jasagredo, line 4: Perhaps if the file was missing we could emit a warning but not fail 🤔

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +194,98 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
      forall m blk. MonadThrow m
   => SomeHasFS m
   -> (ExtLedgerState blk -> Encoding)
   -> DiskSnapshot
   -> ExtLedgerState blk -> m ()
 writeSnapshot (SomeHasFS hasFS) encLedger ss cs = do
-    withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h ->
-      void $ hPut hasFS h $ CBOR.toBuilder (encode cs)
+    !crc <- withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h -> do
+      (_, !crc) <- hPutAllCRC hasFS h $ CBOR.toLazyByteString (encode cs)
```
jasagredo, line 285: One of these two bangs should be redundant I think. Perhaps looking at the Core output can show whether that is the case.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +194,98 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
+      pure crc
```
jasagredo, line 286:
```suggestion
    !crc <- withFile hasFS (snapshotToPath ss) (WriteMode MustBeNew) $ \h -> 
      snd <$> hPutAllCRC hasFS h $ CBOR.toLazyByteString (encode cs)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
jasagredo: Also I think it doesn't matter that much because the next line will write a file with the contents of the CRC so it will be evaluated soon.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -180,32 +194,98 @@ diskSnapshotIsPermanent = isJust . dsSuffix
...
+    withFile hasFS (snapshotToChecksumPath ss) (WriteMode AllowExisting) $ \h -> do
+      void $ hPutAll hasFS h . BS.toLazyByteString . BS.word32HexFixed $ getCRC crc
```
jasagredo, line 288:
```suggestion
    withFile hasFS (snapshotToChecksumPath ss) (WriteMode AllowExisting) $ \h -> 
      void $ hPutAll hasFS h . BS.toLazyByteString . BS.word32HexFixed $ getCRC crc
```
Unary-expression do 🙅 😄

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
geo2a: Yes, great idea. I think that is actually necessary, as otherwise we'd need to delete the snapshots or create the checksums manually when starting an updated node on top of the old node's state. I'll make this change.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
geo2a: Agreed, both are redundant.

geo2a:
I've compared the performance of reading and then writing a snapshot using `db-analyser`. The feature branch seems to be 2s slower, but it is not clear if it is significant.

I would like to spend some more time looking at the microbenchmarks that we currently have and maybe adding one for reading-writing ledger state snapshots.

<details>

<summary>Logs of `db-analyser` with timings</summary>

# Feature 101.325841s

    time $(cabal list-bin db-analyser) --db ~/Workspace/IOG/cardano-node-data/mainnet/state/mainnet-db cardano --config ~/Workspace/IOG/cardano-node-data/mainnet/mainnet-config.json --analyse-from 112526498 --store-ledger 112526499 --verbose
    [0.459957s] TraceImmutableDBEvent (ChunkValidationEvent (StartedValidatingChunk 5285 5285))
    [0.869950s] TraceImmutableDBEvent (ChunkValidationEvent (ValidatedChunk 5285 5285))
    [0.871115s] TraceImmutableDBEvent (ValidatedLastLocation 5285 (Tip {tipSlotNo = SlotNo 114172812, tipIsEBB = IsNotEBB, tipBlockNo = BlockNo 9827240, tipHash = fcb4668176a65b29cf64c31a0bf7c4d4558f61d17774c4f9c3c1366435be8a7f}))
    [59.829891s] Started StoreLedgerStateAt (SlotNo 112526499) LedgerReapply
    [60.150611s] TraceImmutableDBEvent (TraceCacheEvent (TraceCurrentChunkHit 5285 0))
    [60.150929s] TraceImmutableDBEvent (TraceCacheEvent (TraceCurrentChunkHit 5285 0))
    [60.151203s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkMiss 5209 0))
    [60.179937s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkHit 5209 1))
    [60.180243s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkHit 5209 1))
    [101.325411s] Snapshot stored at SlotNo 112526500
    [101.325568s] Snapshot was created at SlotNo 112526500 because there was no block forged at requested SlotNo 112526499
    [101.325841s] Done
    ImmutableDB tip: At (Block {blockPointSlot = SlotNo 114172812, blockPointHash = fcb4668176a65b29cf64c31a0bf7c4d4558f61d17774c4f9c3c1366435be8a7f})
    [101.325945s] TraceImmutableDBEvent DBClosed
    $(cabal list-bin db-analyser) --db  cardano --config  --analyse-from 11252649  102.92s user 6.79s system 107% cpu 1:41.95 total

# Main 99.685827s

     time $(cabal list-bin db-analyser) --db ~/Workspace/IOG/cardano-node-data/mainnet/state/mainnet-db cardano --config ~/Workspace/IOG/cardano-node-data/mainnet/mainnet-config.json --analyse-from 112526498 --store-ledger 112526499 --verbose
    [0.466244s] TraceImmutableDBEvent (ChunkValidationEvent (StartedValidatingChunk 5285 5285))
    [0.879898s] TraceImmutableDBEvent (ChunkValidationEvent (ValidatedChunk 5285 5285))
    [0.881057s] TraceImmutableDBEvent (ValidatedLastLocation 5285 (Tip {tipSlotNo = SlotNo 114172812, tipIsEBB = IsNotEBB, tipBlockNo = BlockNo 9827240, tipHash = fcb4668176a65b29cf64c31a0bf7c4d4558f61d17774c4f9c3c1366435be8a7f}))
    [57.514297s] Started StoreLedgerStateAt (SlotNo 112526499) LedgerReapply
    [57.837578s] TraceImmutableDBEvent (TraceCacheEvent (TraceCurrentChunkHit 5285 0))
    [57.837830s] TraceImmutableDBEvent (TraceCacheEvent (TraceCurrentChunkHit 5285 0))
    [57.838031s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkMiss 5209 0))
    [57.858723s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkHit 5209 1))
    [57.858963s] TraceImmutableDBEvent (TraceCacheEvent (TracePastChunkHit 5209 1))
    [99.685293s] Snapshot stored at SlotNo 112526500
    [99.685447s] Snapshot was created at SlotNo 112526500 because there was no block forged at requested SlotNo 112526499
    [99.685721s] Done
    ImmutableDB tip: At (Block {blockPointSlot = SlotNo 114172812, blockPointHash = fcb4668176a65b29cf64c31a0bf7c4d4558f61d17774c4f9c3c1366435be8a7f})
    [99.685827s] TraceImmutableDBEvent DBClosed
    $(cabal list-bin db-analyser) --db  cardano --config  --analyse-from 11252649  102.60s user 6.13s system 108% cpu 1:40.31 total


</details>

ouroboros-consensus/changelog.d/20241128_084625_georgy.lukyanov_892_checksum_snaphot_file.md
geo2a: The solution I've chosen: in `initLedgerDB`, we catch the error if the `.checksum` file does not exist, issue a warning and retry reading the snapshot without the checksum file.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
```diff
@@ -450,3 +454,17 @@ withFuse (Fuse name m) (Electric io) = do
...
  deriving anyclass (Exception)
+
+{-------------------------------------------------------------------------------
+  Type-safe boolean flags
+-------------------------------------------------------------------------------}
+
+-- | Type-safe boolean flags with type level tags
+--
+-- It is recommended to create pattern synonyms for the true and false values.
+-- Use 'coerce' to unwrap for use in e.g. if-statements.
```
jasagredo, line 465: Couldn't we define a field name for the newtype and remove the need for using `coerce`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
```diff
@@ -112,12 +114,13 @@ initLedgerDB replayTracer
              getGenesisLedger
              stream = do
     snapshots <- listSnapshots hasFS
-    tryNewestFirst id snapshots
+    tryNewestFirst DiskSnapshotChecksum id snapshots
```
jasagredo, line 117: This should be an argument on the diskpolicy, and eventually come from the node.

ouroboros-consensus-cardano/changelog.d/20241129_122139_georgy.lukyanov_892_checksum_snaphot_file.md
jasagredo, line 1: db-analyser (and anything else in the `unstable-*` folders) is not tracked on the changelog, please delete this changelog entry.

jasagredo (commented):
I think it looks good in general, but we should propagate this configuration option to the node and expose it on the JSON/YAML config file.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
geo2a: Yes, this was an idea I had as well, thanks for confirming! Will do.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util.hs
geo2a: I just like the `coerce` use here, but I'm not opposed to an explicit field name either. I'll implement the explicit name approach.

ouroboros-consensus-cardano/changelog.d/20241129_122139_georgy.lukyanov_892_checksum_snaphot_file.md
geo2a: I've removed the changelog fragment for `ouroboros-consensus-cardano`, and now I get a CI failure saying that I need to add one, or mark the whole PR with `no changelog`.

Would it be possible to refine this check to not require a changelog fragment if only the unstable folders have been modified?

Just to be clear, I'm happy with adding `no changelog`, just reporting a paper cut.

jasagredo: Yes indeed that should be the right thing to do. The check is very ad-hoc but it should check the unstable folders.

For now feel free to tag it as no changelog and let's change the script soon.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
```diff
@@ -136,14 +144,23 @@ initLedgerDB replayTracer
...
+          -- If we fail to use this snapshot for any other reason, delete it and try an older one
           Left err -> do
             when (diskSnapshotIsTemporary s) $
               -- We don't delete permanent snapshots, even if we couldn't parse
               -- them
               deleteSnapshot hasFS s
             traceWith tracer $ InvalidSnapshot s err
-            tryNewestFirst (acc . InitFailure s err) ss
+            -- always reset checksum flag after failure
+            tryNewestFirst DoDiskSnapshotChecksum (acc . InitFailure s err) ss
```
jasagredo, line 163: Only if it was enabled from the start!

```suggestion
            tryNewestFirst doChecksum (acc . InitFailure s err) ss
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -149,6 +161,9 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
   Internal: reading from disk
 -------------------------------------------------------------------------------}
 
+-- | Name of a disk snapshot.
+--
+--   The snapshot itself does not have to exist.
```
jasagredo, line 166:
```suggestion
-- The snapshot itself does not need to exist.
```

jasagredo: Or

```suggestion
-- The snapshot itself might not yet exist on the disk.
```

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -422,7 +423,7 @@ storeLedgerStateAt slotNo ledgerAppMode env = do
     storeLedgerState ledgerState = case pointSlot pt of
         NotOrigin slot -> do
           let snapshot = DiskSnapshot (unSlotNo slot) (Just "db-analyser")
-          writeSnapshot ledgerDbFS encLedger snapshot ledgerState
+          writeSnapshot ledgerDbFS NoDoDiskSnapshotChecksum encLedger snapshot ledgerState
```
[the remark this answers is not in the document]
geo2a: Great catch! Fixed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -222,6 +222,7 @@ initFromDisk ::
 initFromDisk LgrDbArgs { lgrHasFS = hasFS, .. }
              replayTracer
              immutableDB = wrapFailure (Proxy @blk) $ do
+    let LedgerDB.DiskPolicyArgs _ _ doDiskSnapshotChecksum = lgrDiskPolicyArgs
```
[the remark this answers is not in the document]
geo2a: Good idea, moved to `where`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
geo2a: I went with the second option.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
geo2a: Fixed! Well caught!

jasagredo (approved):

PR #1328, opened 2024-11-22
Two tiny cleanups

amesgen:
Noticed while doing other stuff, see the commit descriptions


geo2a (approved):
The changes look good to me, though I still have a very limited understanding of what's going on here! Perhaps it's best if somebody else reviews too.

PR #1342, opened 2024-12-10
Update CI changelog checking script to ignore `unstable` components

jasagredo:
With this update to the script it should not be necessary to push a changelog fragment nor set the label whenever a change to `unstable-*` components is made.


geo2a (approved):
Thanks for taking care of this!

Issue #1343, opened 2024-12-12
Integrate the Haskell code generate from `agda-spec` into the build system and CI


geo2a:
The PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1315 by @javierdiaz72 adds an executable specification of a part of the consensus layer in Agda.

We need to integrate the spec into the repo's build system and CI:
- [x] ~check the spec in CI~ this is already done on `main` when building the PDF spec
- [x] generate the Haskell code from the spec and run the unit-tests for it. ~That probably involves integrating the generated code into the top-level Cabal project.~
- [x] ~generate the pdf from the spec (optionally in CI?), which means forwarding the relevant bits to ./nix/pdfs.nix.~ this is already done on `main`.

javierdiaz72:
> That probably involves integrating the generated code into the top-level Cabal project.

I would say that the right thing to do is to effectively integrate the code generation part into the top-level Cabal project since, as a workaround, I created an "empty" `cabal.project` file in `Spec/hs-src` with the only purpose of the top-level one not interfering with my work.

geo2a:
I've tried a couple of things today to integrate the the Agda spec and the generated code into the existing Nix build system. Much has already been done by Alex Esgen, i.e. we seem to actually build the pdfs in CI and thus implicitly also type-check the Agda development.

So what remains to be done is the code generation.

## Nix `mkDerivation` approach

A similar setup exists in the [formal-ledger-specifications](https://github.com/IntersectMBO/formal-ledger-specifications).

 After exploring the Nix rabbit hole for a while, I've found a sequence of steps that would allow to expose the extracted code at the top level:

-   add an attribute for the extracted code into `nix/agda.nix`:

        hsSrc = pkgs.stdenv.mkDerivation {
          inherit (locales) LANG LC_ALL LOCALE_ARCHIVE;
          pname = "agda-spec-hs-src";
          version = "0.1";
          src = ../docs/agda-spec;
          meta = { };
          buildInputs = [ agda pkgs.ghc pkgs.cabal ];
          buildPhase = ''
            export CABAL_DIR=$out/haskell/.cabal
            OUT_DIR=$out make hs
          '';
          doCheck = true;
          checkPhase = ''
            test -n "$(find $out/haskell/ -type f -name '*.hs')"
            # OUT_DIR=$out make hsTest
          '';
          dontInstall = true;
        };
-   this attribute will become exposed in the top-level flake, build it:

[...]

PR #1315, opened 2024-11-21
Generate Haskell code from the Agda spec

javierdiaz72:
This PR resolves #1312 and https://github.com/IntersectMBO/ouroboros-consensus/issues/1343.

PR #1344, opened 2024-12-13
Integrate the Haskell code emitted from `agda-spec` into CI

geo2a:
Fixes https://github.com/IntersectMBO/ouroboros-consensus/issues/1343
Note that this PR targets the branch of https://github.com/IntersectMBO/ouroboros-consensus/pull/1315 and not `main`.

This PR adds Nix derivations that:
- generate Haskell code from the Agda spec using the `Makefile`-based build system in `./docs/agda-spec/`
- wrap the generated code into Nix using `cabal2nix`

In combination with the existing set-up in `flake.nix`, we can now build the generated code and run it's tests using the following flake output:
```
nix build .#hydraJobs.x86_64-linux.native.agda-spec.hsExe
```
which will also be done in CI, i.e. see the logs from testing this PR
- the job [x86_64-linux.native.agda-spec.hsExe](https://ci.iog.io/build/6263029/log) builds and runs tests for the generated code

some other jobs, e.g. the linting of the unit tests for the generated code, are failing, but we can fix those in https://github.com/IntersectMBO/ouroboros-consensus/pull/1315


javierdiaz72 (approved):
LGTM!

PR #1346, opened 2024-12-18
Use a script to check cabal files in GH actions

geo2a:
# Description

I've noticed that the GH actions workflow step that runs  `cabal check` inlines the script we have in `./scropts`. This PR calls the script and also modifies it to exit on the first `cabal check` failure.


jasagredo:
You shouldn't need the no changelog flag anymore 😄

jasagredo (approved):

Issue #1347, opened 2024-12-18
check-changelogs.sh is looking for changelog.d in the wrong directory


nfrisby:
PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1342 happily changed the script to ignore unstable source code.
    
But the whitelist of stable source code is also incorrectly being used to determine where the changelog.d fragment is.

PR #1348, opened 2024-12-18
check-changelogs.sh: bugfix

nfrisby:
Fixes https://github.com/IntersectMBO/ouroboros-consensus/issues/1347


nfrisby:
This fix is a bit hacky --- I'm happy to replace it with someone else's better fix. EG My bash skills would require some brushing off to use an associate array here instead of this simple invariant we currently happen to have, for example.

PR #1349, opened 2024-12-19
Make `ChainSelStarvation` carry an `Enclosed`

jasagredo:
The definition of this tracer, carrying a `Time`, was basically isomorphic to carrying an `Enclosed`.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
@@ -541,15 +541,14 @@ getChainSelMessage starvationTracer starvationVar chainSelQueue =
     startStarvationMeasure = do
       prevStarvation <- atomically $ swapTVar starvationVar ChainSelStarvationOngoing
       when (prevStarvation /= ChainSelStarvationOngoing) $
-        traceWith starvationTracer . ChainSelStarvationStarted =<< getMonotonicTime
+        traceWith starvationTracer $ ChainSelStarvation RisingEdge
 
     terminateStarvationMeasure :: ChainSelMessage m blk -> m ()
     terminateStarvationMeasure = \case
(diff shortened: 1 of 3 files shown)


Niols (approved):
LGTM.

Note that the previous `ChainSelStarvation*` events carried a time. This was added the first time we saw very short starvations, got extremely, and wanted to be sure to measure their exact length, without potential delays introduced in the tracing system. After discussion with @jasagredo, we decided that it was most likely not necessary at this point; it can always be re-introduced subsequently. The time is of course still written in the `starvationVar`, because that is used by the BlockFetch logic.

Niols:
(Impressive that you stumble about this time limited leashing attack bug again, considering how rare it is! It also makes one wonder how we didn't notice it beforehand.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -938,11 +937,10 @@ data TraceIteratorEvent blk
...
 -- them, even though we generally expect to be CPU-bound.
-data TraceChainSelStarvationEvent blk
-    -- | A ChainSel starvation started at the given time.
-  = ChainSelStarvationStarted Time
-
-    -- | The last ChainSel starvation ended at the given time as a block wth the
-    -- given point has been received.
-  | ChainSelStarvationEnded Time (RealPoint blk)
+--
+-- Somehow it happens regularly during syncing for very short times.
```
nfrisby, line 941: > Somehow it happens regularly during syncing for very short times.

This sentence seems a little out of place. Should it be a TODO (investigate why we're seeing these hiccups)? An Issue?

nfrisby (approved):
The change seems fine, since it seems unlikely we'll truly need more than the centisecond precision the node's logging infrastructure gives on tracer times in the log.

I have no strong preference either way, but I haven't been the one debugging these error traces. So I'll defer to you and Niols.

PR #1354, opened 2025-01-08
Release Consensus packages for Node 10.2.0

geo2a:


amesgen (approved):
Looks great!

CI should go green once you do an (empty) force push (We should really change the trigger of that GHA job such that it re-runs when labels are added...)

Also, as mentioned in the release process doc, consider opening a CHaP PR (not mandatory though):

> It is recommended to already open a draft PR to [CHaP](https://github.com/IntersectMBO/cardano-haskell-packages) and make sure that CI passes to avoid later surprises (eg due to missing bounds).

PR #1355, opened 2025-01-09
Fix create-release.hs script

geo2a:
This PR fixes the `./scripts/release/create-release.hs` script to:
- traverse packages in prescribed order
- do not include zeros in version patterns for dependencies

When cutting the previous release (https://github.com/IntersectMBO/ouroboros-consensus/pull/1354), I've noticed that the scripts does not update the dependencies in the `.cabal` files. This was due to traversing the packages in a wrong order because of storing them in `Data.Map` and also because of a quirk in the version patterns.

I've verified that the updated scripts has the expected behaviour on on the repo state pre https://github.com/IntersectMBO/ouroboros-consensus/pull/1354. I do not think it's worth it to add tests for the script. Hopefully, I have only fixed bugs and not introduced new ones!

### DIFF
--- a/scripts/release/create-release.hs
+++ b/scripts/release/create-release.hs
@@ -13,9 +13,9 @@
     with-utf8,
 -}
 
-{-# LANGUAGE BlockArguments      #-}
-{-# LANGUAGE GHC2021             #-}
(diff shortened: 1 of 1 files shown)


jasagredo (approved):
Thanks!

PR #1356, opened 2025-01-09
Remove legacy NodeToClient serialization code

amesgen:
#1314 removed various old `NodeToClientVersion`s, so we can remove lots of now-dead code.


dnadales (approved):
Thanks!

PR #1358, opened 2025-01-11
Investigation: failing faster when validating txs with missing inputs

amesgen:
In various common scenarios, the mempool will (re)apply a tx to a ledger state that does not contain all of the necessary inputs (usually because the tx has already been applied before). Therefore, it is desirable for tx (re) application to fail fast in these cases.

The purpose of this PR is just to summarize effort to look into improvements here. Ideally, no modification to Consensus would be necessary in the end; Ledger should be able to do everything on their side.

Specifically, this PR tries two things:

 - Try to use `reapplyTx` instead of `applyTx` when the tx is already in the mempool. The idea here is to avoid crypto work that will be unnecessary as tx validation will fail due to missing inputs eventually anyways.

   [This tx](https://adastat.net/blocks/3478fe87fd494cb7ecc35b4e1f800f7e5c54775eb6dcbeda8bb4df158e8b05c0) is a concrete example for a tx where this saves time.

 - Try to restrict the UTxO set before calling the ledger to the potentially needed inputs of the tx. That's exactly what UTxO HD does, so this PR branch is targeting #1267.

---

To test/benchmark this, this PR modifies the `--repro-mempool-and-forge` db-analyser pass. When processing a block, it adds all of its transactions to the mempool, and then individuall tries to add each tx again, measuring how long it takes until the mempool is rejected. The output data is stored in `readd-txs.csv`.

I ran `db-analyser` like this
```console
cabal run db-analyser -- --v2-in-mem --db /path/to/db --no-snapshot-checksum-on-read \
  --repro-mempool-and-forge 1 --analyse-from 133660855 --num-blocks-to-process 100000 \
  cardano --config /path/to/config.json
```
(takes ~30min) and used the mean of two runs.

Plots are created using a very ad-hoc [script](https://gist.github.com/amesgen/3286e36c300f7f715edf072e9d15e977).

Here, GHC mutator time is used, and `mut_{baseline,patched}` refer to versions against current `main` (ie without UTxO HD), and `mut_utxohd_{baseline,patched}` to versions in this branch. `_baseline` means that the mempool hasn't been patched to use `reapplyTx`, whereas `_patched` means exactly that.

<details><summary>Show plot :chart_with_upwards_trend:  </summary>

plot

</details>

---

Thoughts:

 - The mempool patch definitely helps, both with and without UTxO HD, but only for certain txs.
 - UTxO HD seems is slower when comparing means/medians.
 - UTxO HD seems to significantly help for certain txs, visible both in "`mut_baseline` vs `mut_utxohd_baseline`" and "`mut_patched` vs `mut_utxohd_patched`". 
 - Eyeballing seems to indicate that there are significantly fewer big outliers with UTxO HD.

---

A next step would be to profile individual txs to understand why they are currently relatively "slow to fail". I did this already for a few txs (that's e.g. how I saw that the tx mentioned above is slow to fail due to crypto without the mempool patch) like this:

 - Create a ledger snapshot for the preceding block of the tx.
 - Run the `db-analyser` pass with `--num-blocks-to-process 1`, and make use of `{start,stop}ProfTimer` (see the code for an example).
 - Load the result into https://www.speedscope.app/ (requires `+RTS -pj`).

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
@@ -38,8 +38,8 @@ import           Control.Monad.Trans.Except (runExcept)
 import           Control.Tracer
 import qualified Data.Foldable as Foldable
 import qualified Data.List.NonEmpty as NE
-import           Data.Set (Set)
-import qualified Data.Set as Set
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 import           Data.Typeable
 import           GHC.Generics (Generic)
 import           Ouroboros.Consensus.Block
@@ -80,7 +80,7 @@ data InternalState blk = IS {
       -- 'MempoolSnapshot' (see 'snapshotHasTx').
       --
       -- This should always be in-sync with the transactions in 'isTxs'.
-    , isTxIds        :: !(Set (GenTxId blk))
+    , isTxIds        :: !(Map (GenTxId blk) (Validated (GenTx blk)))
 
       -- | The cached ledger state after applying the transactions in the
       -- Mempool against the chain's ledger state. New transactions will be
@@ -150,7 +150,7 @@ initInternalState ::
   -> InternalState blk
 initInternalState capacityOverride lastTicketNo cfg slot st = IS {
       isTxs          = TxSeq.Empty
-    , isTxIds        = Set.empty
+    , isTxIds        = Map.empty
     , isLedgerState  = st
     , isTip          = castPoint $ getTip st
     , isSlotNo       = slot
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -279,17 +279,25 @@ validateNewTransaction
...
              }
         )
   where
+    res = case Map.lookup (txId tx) isTxIds of
+      Nothing  -> applyTx cfg wti isSlotNo tx st
+      Just vtx -> do
+        _ <- reapplyTx cfg isSlotNo vtx st
+        -- that must have failed, since every tx must consume at least one
+        -- input and this one is necessarily a double-spend
+        pure $ error "unreachable"
```
jasagredo, line 299: Am I reading this wrong, or will this only fire when double-adding transactions into the mempool? That will be indeed a fail-fast situation, but it won't help when the tx is not in the mempool but instead has already been applied to the state, no?

amesgen: > Am I reading this wrong, or will this only fire when double-adding transactions into the mempool?

Yes, this will indeed only fire in the scenario you describe.

> but it won't help when the tx is not in the mempool but instead has already been applied to the state, no?

Exactly :+1: For the purpose of this PR (benchmarking when/why tx application is failing slowly) this is enough and was convenient and very simple to implement, but we indeed want the final optimization to be more general, see https://github.com/IntersectMBO/cardano-ledger/issues/4849#issuecomment-2607834134.

PR #1364, opened 2025-01-23
Draft the Consensus Whitepaper

### LINKED ISSUE #1363: Consensus Whitepaper
As part of creating the Cardano Blueprint, we need to draft a Consensus whitepaper. This document will give an overview of the Consensus component and make it easier to understand and provide guidance for:

- People implementing alternative Cardano nodes.
- Users of the Consensus functionality.
- New contributors to ouroboros-consensus.

In addition, this document will help us elicit additional requirements for the Consensus documentation. 

We should aim for an 8-10 pages high-level overview that:
- describes the components of Consensus,
- explains the Consensus layer, and how it relates to the Diffusion (aka Network) and Ledger layers,
- contains diagrams, 
- would make it possible to develop a toy model of the Consensus layer.

geo2a:
Closes #1363

### DIFF
--- a/docs/website/contents/for-developers/Whitepaper.md
+++ b/docs/website/contents/for-developers/Whitepaper.md
@@ -27,6 +27,12 @@ Pros: enables massive code reuse and forces the code to be structured in the way
 
 Cons: abstraction is a double-edged sword and may be difficult to encode in some programming languages. Engineers require extended onboarding to be productive.
 
+## Replaying history of multiple eras
+
+When replaying the chain in `ouroboros-consensus`, as I understand, we use the same code of the HardFork combinator that is used when actually producing/validating these blocks. This is nice because we do not have to have separate code path for the historic chain and caught-up protocol participation.
+
+But obviously we don not have to use the HFC. What would be a reasonable way to enable a non-HFC node to catch up with the chain? Would this hypothetical node be able to enjoy code reuse for the historic and non-historic blocks? In the worst case, we should be able to hard-code the historical eras.
+
 ## The Hard Fork Combinator: a uniform way to support multiple eras}
 
 Ideally, this would be a short section that should outline the core ideas behind the HFC without a single line of Haskell. The purpose should be to demonstrate the benefits of having an abstract interface for describing mixed-era blocks. The interested reader would be then referred to an extended, Haskell-enabled document that described the HFC in its full glory.
--- a/docs/website/contents/for-developers/Whitepaper.md
+++ b/docs/website/contents/for-developers/Whitepaper.md
@@ -56,6 +56,8 @@ flowchart TD
 
 In order to switch to an alternative chain, Consensus needs to evaluate the validity of such a chain. In order to evaluate the validity of a block, we need to have a Ledger state at the predecessor block. As applying blocks is not an inversible operation, this is usually solved by maintaining the last `k` ledger states in memory.
 
+Here and further, `k` is the **security parameter**. In the Praos consensus algorithm, `k` is the parameter of the Common Prefix theorem that bounds how many blocks would exist on a well-connected honest node's selections after its intersection with another well-connected honest node. See the Ouroboros Praos [paper](https://iohk.io/en/research/library/papers/ouroboros-praos-an-adaptively-secure-semi-synchronous-proof-of-stake-protocol/) for more details. For the purposes of implementing a Cardano node today, it is sufficient to know that `k`'s value on mainnet is 2160, and it denotes the amount of ledger states the implementation needs to keep track of in order to perform chain selection.
+
 In the Haskell reference implementation, the Storage layer is implemented by the `ChainDB` components, which comprises the `ImmutableDB`, `VolatileDB` and `LedgerDB` subcomponents:
 
 | Component   | Responsibility                                                                                            |
@@ -65,7 +67,7 @@ In the Haskell reference implementation, the Storage layer is implemented by the
 |             | efficiently switch to a different suffix of the chain if needed                                           |
 | LedgerDB    | Maintaining the last `k` (2160 on Cardano mainnet) ledger states in memory to facilitate chain selection. |
 
-**TODO**: refile requirements on access patterns for Immutable and Volatile databases. Read/write access? When do we need random access and where we do not? What does it mean specifically to efficiently switch chains?
+**TODO**: refine requirements on access patterns for Immutable and Volatile databases. Read/write access? When do we need random access and where we do not? What does it mean specifically to efficiently switch chains?
 
 ## Requirements and Constraints for the Consensus and Storage Layers
 
--- a/docs/website/contents/for-developers/Whitepaper.md
+++ b/docs/website/contents/for-developers/Whitepaper.md
@@ -1,0 +1,34 @@
+In this technical report, we describe the necessary components comprising the consensus layer of a Cardano blockchain node. The main goal of this report is to provide guidance to software engineers that intend to implement the Consensus layer of a Cardano node from scratch.
+
+# Introduction
+
+# Single-era Consensus Layer
+
+This section describes the components of the Consensus layer of the Cardano Node, as if Cardano only ever had one era. While this assumption greatly simplifies the implementation of the Consensus layer, one must keep that if such an assumption is made, in may not be straightforward or even possible to implement a node that is capable of syncing with Cardano mainnet. However, we still argue that this section is useful as an educational material.
+
+We need to take both advantage of the rigorous structure we have in the code base, but at the same time take care not to expose the non-Haskell target audience to the full power of abstraction it provides. We can achieve it by instantiating the abstractions we have at concrete types. We may also want to monomorphise, i.e. remove the typeclass constraints as much as possible after instantiating them. We may then proceed to obfuscate the Haskell syntax as something else.
+
+The key typeclass we need to instantiate is ConsensusProtocol.
+
+# Multi-era Considerations
+
+With the blokchain network evolving, the block format and ledger rules are bound to change. In Cardano, every significant change starts a new "era". There are several ways to deal with multiple eras in the node software, associated here with some of the DnD alignments:
+
+* Chaotic Evil: the node software only ever implements one era. When the protocol needs to be updated, all participants must update the software or risk being ejected from the network. Most importantly, the decision to transition to the new era needs to happen off-chain.
+Pros: the simplest possible node software.
+Cons: on-chain governance of the hard-fork is impossible, as the software has no way of knowing where the era boundary is and does not even have such a concept. Additionally, special care is needed to process history blocks: chain replay is likely to be handled by special code, separate from the active era's logic.
+
+* Chaotic Good: the node software supports the current era and the next era. Once the next era is adopted, a grace period is allowed for the participants to upgrade. The decision to upgrade may happen on chain.
+Pros: allows for on-chain governance of the hard fork.
+Cons: supporting two eras is more difficult than one: higher chances of bugs that will cause the network to fork in an unintended way. Like in the previous case, special care is needed to process historic blocks.
+
+* True Neutral: software is structured in such a way that is supports all eras.
+Pros: enables massive code reuse and forces the code to be structured in the way that allows for abstract manipulation of blocks of different eras. The on-chain governance of hard-forks is reflected in the code, and ideally in the types as well, making it more likely that unintended scenarios are either impossible or easily discover able through type checking and testing.
+
+Cons: abstraction is a double-edged sword and may be difficult to encode in some programming languages. Engineers require extended onboarding to be productive.
+
+## The Hard Fork Combinator: a uniform way to support multiple eras}
+
+Ideally, this would be a short section that should outline the core ideas behind the HFC without a single line of Haskell. The purpose should be to demonstrate the benefits of having an abstract interface for describing mixed-era blocks. The interested reader would be then referred to an extended, Haskell-enabled document that described the HFC in its full glory.
+
+
--- a/docs/website/contents/for-developers/Whitepaper.md
+++ b/docs/website/contents/for-developers/Whitepaper.md
@@ -2,8 +2,6 @@ In this document, we describe the necessary components comprising the Consensus 
 
 This document is a work in progress. We strive to provide a set of requirements for and responsibilities of the Consensus layer that is agnostic of a particular implementation. However, we still very much are informed by the Haskell implementation in [ouroboros-consensus](https://github.com/IntersectMBO/ouroboros-consensus) as it is the only one.
 
-# Introduction
-
 # What does Consensus and Storage need to do: responsibilities and requirements
 
 ## Responsibilities of the Consensus Layer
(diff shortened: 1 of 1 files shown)


abailly:
Would be great if we avoided LaTeX IMO: we want to publish this at least on the web, and make it straightforward to use for various people. Markdown is just fine, it's simple, it supports simple diagrams and even math equations out of the box and we can always use pandoc to get a PDF if we want.

ch1bo:
This is coming together well!

As this seems to be fairly independent of the concrete implementation, would you be interested in contributing this to https://github.com/cardano-scaling/cardano-blueprint?

It could serve as a perfect starting point for conversations with other implementers of Cardano consensus in other languages (cc @abailly, @michele-nuzzi), where together we could make it an implementation-independent asset for how the consensus on Cardano works. Of course we would need to add more fidelity and detail (while discussing what is really implementation specific or not?) over time, but I see it as an ongoing collaborative effort of multiple people.

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+The following table outlines the informal components of the Haskell reference implementation and their responsibilities:
+
+| Component                                      | Responsibility                                                                      |
+|:-----------------------------------------------|:------------------------------------------------------------------------------------|
+| Forging Loop (aka The Mint)                    | Extend the Cardano chain by minting new blocks                                      |
+| `Mempool`                                      | Maintain a list of valid transactions to include in new blocks                      |
+| `ChainSync` and `BlockFetch` clients           | Gather candidate chains from peers                                                  |
+| `ChainSync` and `BlockFetch` servers           | Provide peers with access to `ChainDB` (see Storage Layer)                          |
+| `ChaineSel`                                    | Among all forks that can be fetched from peers, must identify the best chain.       |
```
jasagredo, line 34:
```suggestion
| `ChainSel`                                    | Among all forks that can be fetched from peers, must identify the best chain.       |
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+| N2N `TxSubmission`, N2C `TxSubmission` servers | Accept transactions from peers and store them into `Mempool` for further processing |
+|                                                |                                                                                     |
+|                                                |                                                                                     |
+|                                                |                                                                                     |
```
jasagredo, line 38:
```suggestion
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Responsibilities of the Storage Layer
+
+The Consensus layer needs to keep track of a significant amount of data to maintain the Cardano network. Some of this data is "hot" and relates to the candidate chains and other data is cold and relates to the historical chain.
```
jasagredo, line 42: If you make this distinction, I would clarify what data is Hot in the diagram and what is cold

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+In order to switch to an alternative chain, Consensus needs to evaluate the validity of such a chain. In order to evaluate the validity of a block, we need to have a Ledger state at the predecessor block. As applying blocks is not an inversible operation, this is usually solved by maintaining the last `k` ledger states in memory.
+
+In the Haskell reference implementation, the Storage layer is implemented by the `ChainDB` components, which comprises the `ImmutableDB`, `VolatileDB` and `LedgerDB` subcomponents:
+
+| Component   | Responsibility                                                                                            |
+|:------------|:----------------------------------------------------------------------------------------------------------|
+| ImmutableDB | Store definitive blocks on disk and provide iterator access to them for new syncing peers.                |
+| VolatileDB  | Store non-definitive blocks on disk, provide them to peers (random access?),                              |
```
jasagredo, line 64: The starting point of a stream is random, then from that point onwards I think it is sequential. This might have changed with Genesis

jasagredo: (This means that I agree with saying this is `random access`)

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+|             | efficiently switch to a different suffix of the chain if needed                                           |
+| LedgerDB    | Maintaining the last `k` (2160 on Cardano mainnet) ledger states in memory to facilitate chain selection. |
+|             |                                                                                                           |
```
jasagredo, line 67:
```suggestion
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+Each ChainSync client maintains an upstream peer's candidate chain.
+
+The protocol state is also independently maintained alongside each candidate chain by ChainSync.
+
+### Details
+
+It disconnects if the peer violates the mini protocol, if the peer sends an invalid header, or if switching to their chain would require rolling back more than kcp blocks of the local selection.
+
+*TODO*(georgy): what is kcp here?
```
jasagredo, line 115: `kcp` is `k`

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+It's able to their validate headers past the intersection with the local selection because the parts of the ledger state necessary to validate a header were completely determined some positive number of slots ago on that header's chain; see _forecasting_ within the ledger rules and _snapshots_ within the ledger state.
```
jasagredo, line 117:
```suggestion
It's able to validate headers past the intersection with the local selection because the parts of the ledger state necessary to validate a header were completely determined some positive number of slots ago on that header's chain; see _forecasting_ within the ledger rules and _snapshots_ within the ledger state.
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+ChainSync blocks while the candidate chain is past the forecast range of the local selection.
```
jasagredo, line 119: We use block to mean blocking and blockchain blocks. I would use `stall` here, as at first I though "Chainsync blocks? which blocks are those?"

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+The forecast range must be great enough for the peer to rescue the local node from the worse-case scenario.
+The Praos paper bounds this to needing at most kcp+1 headers from the peer, which that paper also bounds to requiring at most scg slots of forecast range.
```
jasagredo, line 121: `scg` is I think the stability window which on mainnet, by the Chain Growth Property of Praos is $3k/f=129600$ slots. (k = 2160, f = 0.05)

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+The Genesis paper is also satisfied by a forecast range of scg.
+(The ChainSync server is much simpler than the client; see _followers_ below.)
+
+## ChainSync server
+
+ChainSync server provides an iterator into the ChainDB for downstream peers to be able to download headers and blocks.
+
+#### Details
+
+Moreover — because the node must serve the whole chain and not only the historical chain — each ChainSync server actually uses a _follower_ abstraction, which is implemented via iterators and additionally supports the fact that ChainSel might have to rollback a follower if it's already within kcp of the local selection's tip.
```
jasagredo, line 131:
```suggestion
Moreover — because the node must serve the whole chain and not only the historical chain — each ChainSync server actually uses a _follower_ abstraction, which is implemented via iterators and additionally supports the fact that ChainSel might have to rollback a follower if it's already within kcp of the local selection's tip, i.e. if it is streaming the volatile part of the chain.
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+The combined ledger and protocol state is maintained alongside the local selection by ChainSel, so that blocks can be validated.
+
+## ChainDB
+
+A Praos node must not introduce unnecessary delays between receiving a block and forwarding it along.
+It is therefore an important observation that the ChainDB does not require the Durability property of ACID: upstream peers will always be available to replace any blocks a node loses.
+
+## LedgerDB
+
+In both ChainSel and ChainSync, rollbacks require maintenance of/access to the past kcp+1 states, not only the tip's state — access to any such state must be fast enough to avoid disrupting the semantics of the worst-case delay parameter Delta assumed in the Praos paper's security argument.
```
jasagredo, line 163:
```suggestion
In both ChainSel and ChainSync, rollbacks require maintenance of/access to the past `k+1` states, not only the tip's state — access to any such state must be fast enough to avoid disrupting the semantics of the worst-case delay parameter Delta assumed in the Praos paper's security argument.
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+The Mempool is bounded via a multi-dimensional notion of size such that it never contains more transactions than could fit in N blocks.
+Peers synchronize their Mempools via the TxSubmission protocol.
+This mini protocol leverages the fact that the Mempool is a sequence as opposed to an unordered set; a simple integer suffices as the iterator state for the TxSub client.
+(Recall that transactions flow from client to server, since the orientation is determined by the flow of blocks and pending transactions naturally flow opposite of blocks.)
+
+## The Mint aka Block Forge
+
+Whenever the wall clock enters a new slot, the node checks whether its configured minting credentials (if any) were elected by the protocol state and forecasted ledger state to lead this slot.
+If so, it mints a block that extends its local selection (or its immediate predecessor if the tip somehow already occupies this slot or greater) and contains the longest prefix of the Mempool's transactions that can fit.
+That block is then sent to ChainSel, as if it had been fetched.
```
jasagredo, line 179:
```suggestion
That block is then sent to ChainSel, as if it had been fetched, to make sure other nodes will be able to adopt it. Only reason why a block would be rejected at this point is if there was a mismatch between the Block application and Mempool transaction validation logic.
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+When (if) the node selects that block, the Mempool will react as it does for any change in the local selection: it discards any transactions that are no longer valid in the updated hypothetical block the Mempool continuously represents.
```
jasagredo, line 181: This is fine, but it alludes to a behavior of the mempool that was not described above, namely resyncing with the ledgerdb.

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Some Important Optimizations
+
+Both both blocks and transactions can be applied much more quickly if they are known to have been previously validated.
+For blocks, the outcome will be exact same, since they can only validly extend a single chain.
+A transaction, though, might be validated against a different chain/ledger state than the first time, so many checks still need to happen.
+But many "static" checks don't need to be repeated, since they'd fail regardless of the ledger state.
+
+Despite block reapplication being much faster than initial validation, the node should not need to reapply their entire historical chain whenever it is restarted.
+The node instead occasionally persists its oldest ledger state (ie the kcp+1st).
```
jasagredo, line 192:
```suggestion
The node instead occasionally persists its oldest ledger state in the ledger db (ie the `k+1`st, the immutable tip of the chain).
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+On startup, the node only needs to deserialize that snapshotted ledger state and then replay the best chain amongs the persisted blocks it has that extends this ledger state in order to re-establish its kcp+1 ledger states.
```
jasagredo, line 193:
```suggestion
On startup, the node only needs to deserialize that snapshotted ledger state and then replay the best chain amongs the persisted blocks it has that extends this ledger state in order to re-establish its k+1 ledger states in the ledgerdb (and k volatile blocks in its selected chain).
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+# Multi-era Considerations
+
+In Cardano, an _era_ is an informal concept that unites a set of Cardano features, such as ledger rules or consensus protocols. In Cardano, the on-chain governance ultimately decides when to adopt backwards-incompatible changes, by incrementing the major component of the protocol version. The first such _era transition_ switched to the Praos protocol.
```
jasagredo, line 197: I think it might be worth linking to this https://github.com/cardano-foundation/CIPs/blob/master/CIP-0059/feature-table.md

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+In this section, we discuss the challenges imposed by the need to support multiple Cardano eras. We start with the table outlining the different aspects that need to considered when implementing a multi-era node. We than transition to discussing possible approaches to supporting a multiple-era blockchain network in the node software. Finally, we discuss the Hard Fork Combinator --- the approach implemented in the Haskell reference implementation.
```
jasagredo, line 199:
```suggestion
In this section, we discuss the challenges imposed by the need to support multiple Cardano eras. We start with the table outlining the different aspects that need to considered when implementing a multi-era node. We then transition to discussing possible approaches to supporting a multiple-era blockchain network in the node software. Finally, we discuss the Hard Fork Combinator --- the approach implemented in the Haskell reference implementation.
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+| Transition to a new era  | slot number of the era boundary unknown            | during the era transition:                                                                           |
+|                          |                                                    | * switch the consensus protocol, block format, ledger rules                                          |
+|                          |                                                    | * translate transactions received from prevoios-era peers                                            |
+|                          |                                                    | into the format of the current era for them included in a new block                                  |
+
+## Approaches to handle protocol changes
+
+With the blokchain network evolving, the block format and ledger rules are bound to change. In Cardano, every significant change starts a new "era". There are several ways to deal with multiple eras in the node software, associated here with some of the DnD alignments:
```
jasagredo, line 213:
```suggestion
With the blockchain network evolving, the block format and ledger rules are bound to change. In Cardano, every significant change starts a new "era". There are several ways to deal with multiple eras in the node software, associated here with some of the DnD alignments:
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+Cons: abstraction is a double-edged sword and may be difficult to encode in some programming languages. Engineers require extended onboarding to be productive.
+
+## Replaying history of multiple eras
+
+Cardano has the peculiarity of being a multi-era network, in which at given
+points in the chain, new backwards-incompatible features were added to the
+ledger. Consensus, as it needs to be able to replay the whole chain, needs to
+implement some mechanism to switch the logic used for each of the eras, in
+particular the Ledger layer exposes different implementations for each one of
+the ledger eras.
```
jasagredo, line 235: This probably could go at the beginning of this section? (where I linked the feature table) possibly reworded to match the context there.

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+
+## The Hard Fork Combinator: a uniform way to support multiple eras}
+
+*Hard Fork Combinator (HFC)* is mechanism that handles era transitions, including changes to ledger rules, block formats and even consensus protocols. It automates translations between different eras and provides a minimal interface for defining specific translations that obey rigorous laws. The HFC was primarily introduced to handle the fundamental complexity that arises when the translation between wall clocks and slot onsets depends on the ledger state (since on-chain governance determines when the era transition happens).
```
jasagredo, line 239:
```suggestion
*Hard Fork Combinator (HFC)* is a mechanism that handles era transitions, including changes to ledger rules, block formats and even consensus protocols. It automates translations between different eras and provides a minimal interface for defining specific translations that obey rigorous laws. The HFC was primarily introduced to handle the fundamental complexity that arises when the translation between wall clocks and slot onsets depends on the ledger state (since on-chain governance determines when the era transition happens).
```

docs/website/contents/for-developers/Whitepaper.md
```diff
@@ -0,0 +1,244 @@
...
+It also handles the comparatively simple bookkeeping of the protocol, ledger, codecs, and so on changing on the era boundary — ie a variety of blocks, transactions, etc co-existing on the same chain in a pre-specified sequence but not according to a pre-specified schedule.
+Lastly, it handles the fact that ticking and forecasting can cross era boundaries, which requires translation of ledger states, protocol states, and pending transactions from one era to the next.
+The HFC cannot automatically infer the implementation of these necessities, but it automates as much as possible against a minimal interface that requires the definition of the specific translations etc.
+
```
jasagredo, line 243: I miss some actual description of *what* the hard fork combinator does.

> The HFC lifts types parametrized by eras into sum types which can be manipulated in an abstract way. As so, for example, it lifts the LedgerState of an era into a HardForkLedgerState which is a sum type of LedgerState of the CardanoEras.
>
> Such sum types fulfill the same interfaces as the single-era types, and the consensus layer is designed around such interfaces, allowing the same code to run a multi-era network without modifications.
>
> These sum types are paired with the functions and mechanisms that make possible transitioning to a different summand in the type (translating between eras) or combine properties of those different summands (the forecasting window over an era transition is a combination of the forecasting window of the old era and the new era).

Something along those lines ^ ?

docs/website/contents/for-developers/Whitepaper.md
geo2a: I'll put a TODO to refine the requirements on access patterns for Immutable and Volatile databases so that we can discuss that further!

docs/website/contents/for-developers/Whitepaper.md
geo2a: I will use the term "stability window". This is probably already too much detail for this document though.

geo2a:
Superseded by https://github.com/cardano-scaling/cardano-blueprint/pull/7

PR #1374, opened 2025-01-29
Use mlocked KES internally

tdammers:
This changes Consensus such that mlocked KES keys are used internally.

This is important groundwork for supporting KES agents in the future. In this form, the code will still load KES keys from disk, which is unsound, but the internal machinery is ready to also accept KES keys from other sources, and once loaded, KES keys will be handled appropriately (kept in mlocked RAM at all times, securely erased when expired).

This also involves a restructuring of the `HotKey` data structure, which now manages not only a KES SignKey, but also the corresponding OpCert. This is necessary for two reasons:

- With a KES agent, the OpCert will be provided along with the SignKey; this is the easiest way to make sure that the OpCert always matches the SignKey it is used with
- With the new structure, KES keys can be replaced in a live node kernel, without having to restart it and reload the entire configuration. Because of this, the HotKey, which manages the dynamic part of the node kernel's signing mechanism, needs to be able to replace not just the SignKey (which it already did, in order to handle key evolution), but also the OpCert (which will not change when a SignKey evolves, but it will change when a new SignKey is provided).

Supersedes #1284.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -108,6 +109,10 @@ data family CodecConfig blk :: Type
 -- avoid circular dependencies.
 data family StorageConfig blk :: Type
 
+-- | Credentials needed for block forging. In eras that use KES, this will be
+-- a pair of KES sign key and OpCert; in other eras, it should be 'Void'.
+type family BlockForgingCredentials blk :: Type
```
jasagredo, line 114: This is defined here, and for the HFBlock, but nowhere else?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -77,6 +77,8 @@ deriving stock   instance CanHardFork xs => Show (LedgerState (HardForkBlock xs)
 deriving stock   instance CanHardFork xs => Eq   (LedgerState (HardForkBlock xs))
 deriving newtype instance CanHardFork xs => NoThunks (LedgerState (HardForkBlock xs))
 
+type instance BlockForgingCredentials (HardForkBlock '[blk]) = BlockForgingCredentials blk
```
jasagredo, line 80: Ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Unary.hs
```diff
@@ -11,6 +11,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
```
jasagredo, line 14: Why is this necessary?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
```diff
@@ -99,6 +103,12 @@ hardForkCanBeLeader =
       SomeErasCanBeLeader
     . hmap (WrapCanBeLeader . canBeLeader)
 
+hardForkFinalize :: (Monad m, All Top xs)
+                 => NonEmptyOptNP (BlockForging m) xs -> m ()
+hardForkFinalize blockForging =
+  -- pure ()
```
jasagredo, line 109: Remove

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
```diff
@@ -15,8 +15,11 @@ module Ouroboros.Consensus.HardFork.Combinator.Forging (
   , hardForkBlockForging
   ) where
 
+import           Control.Monad (void)
```
jasagredo, line 18: I _think_ this will be a redundant import in some versions. We usually use `import qualified Control.Monad as Monad` and use `Monad.void`. This is until we drop GHC 8.10 I think.

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
```diff
@@ -170,7 +171,10 @@ data SimpleStdHeader c ext = SimpleStdHeader {
     , simpleBodySize :: SizeInBytes
     }
   deriving stock    (Generic, Show, Eq)
-  deriving anyclass (Serialise, NoThunks)
+  deriving anyclass (NoThunks)
+
+deriving anyclass instance KnownNat (Hash.SizeHash (SimpleHash c)) =>
```
jasagredo, line 176: If this constraint appears in many more places I wonder if it would be wise to use an alias `KnownHashSize`

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wno-orphans            #-}
```
jasagredo, line 1: Move this after the LANGUAGE pragmas.

Also if you could annotate in a comment why we need it?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -618,6 +627,40 @@ instance PraosCrypto PraosMockCrypto where
...
+    return . SignKeyDSIGNMEd448 $ genKeyDSIGN seed
+
+  cloneKeyDSIGNMWith _ sk = return sk
+
+  getSeedDSIGNMWith _ _ _ =
+    error "getSeed not implemented for Ed448"
+
+  forgetSignKeyDSIGNMWith _ _ =
+    -- This is fake mlocking, so just don't forget anything
+    return ()
```
jasagredo, line 656: Are we using this still? I see above you changed to ED25519DSIGN

sop-extras/changelog.d/20250130_111443_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,3 @@
+### Non-Breaking
+
+- Add `ctraverse`, `traverse_`, and `HTraverse_`
```
jasagredo, line 3:
```suggestion
- Add `ctraverse`, `traverse_`, and `HTraverse_` to `OptNP`
```

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -744,5 +744,4 @@ test-suite doctest
   default-language: Haskell2010
   ghc-options: -Wno-unused-packages
   build-depends:
-    base,
-    latex-svg-image,
```
jasagredo, line 748: Why was this removed?

It is needed for the doctests

ouroboros-consensus/changelog.d/20250130_093251_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,9 @@
+### Breaking
+
+- Use new mlocked KES API to represent KES sign keys internally. This ensures
```
jasagredo, line 3: It would be good to link to the new API pr wherever it is (cardano-base?) or to a document that explains the design.

ouroboros-consensus/changelog.d/20250130_093251_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,9 @@
...
+  that KES keys are securely erased when replaced with a newer evolution or a
+  fresh key, and that they will not spill to disk or swap.
+- Add `finalize` method to `BlockForging`, and use it where necessary to clean
+  up when a block forging thread terminates (see `forkLinkedWatcherFinalize`)
+- Add orphan instances to make Ed448 look like a proper mlocked DSIGN algorithm
```
jasagredo, line 8: Same, are we using this?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -245,16 +251,31 @@ instance Crypto c => ChainOrder (PraosChainSelectView c) where
...
-    -- genesis stakeholder delegate cold key) to the online KES key.
-    praosCanBeLeaderOpCert     :: !(OCert.OCert c),
-    -- | Stake pool cold key or genesis stakeholder delegate cold key.
+  { -- | Stake pool cold key or genesis stakeholder delegate cold key.
     praosCanBeLeaderColdVerKey :: !(SL.VKey 'SL.BlockIssuer c),
-    praosCanBeLeaderSignKeyVRF :: !(SL.SignKeyVRF c)
+    praosCanBeLeaderSignKeyVRF :: !(SL.SignKeyVRF c),
+    praosCanBeLeaderCredentialsSource :: !(PraosCredentialsSource c)
+    -- praosCanBeLeaderOCert :: !(OCert.OCert c),
+    -- praosCanBeLeaderKESKey :: !(SL.SignKeyKES c)
```
jasagredo, line 259: Remove

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -148,60 +183,126 @@ sign = sign_
...
       }
 
+mkShelleyHotKey :: forall m c. (Crypto c, IOLike m)
+                => OCert.OCert c
+                -> SL.SignKeyKES c
+                -> Absolute.KESPeriod
+                -> Word64
+                -> m (HotKey c m)
+mkShelleyHotKey ocert sk startPeriod maxEvolutions =
+  mkHotKey ocert sk startPeriod maxEvolutions
```
jasagredo, line 293: Why define this if all arguments are the same?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -148,60 +183,126 @@ sign = sign_
...
+mkHotKey ocert initKey startPeriod maxKESEvolutions = do
+  hotKey <- mkEmptyHotKey maxKESEvolutions (pure ())
+  set hotKey ocert initKey 0 startPeriod
+  return hotKey
+
+-- Create a new 'HotKey' and initialize it to the given initial KES key. The
+-- initial key should be at the given evolution.
+mkHotKeyEv ::
+     forall m c. (Crypto c, IOLike m)
+  => Word
```
jasagredo, line 223:
```suggestion
mkHotKeyAtEvolution ::
     forall m c. (Crypto c, IOLike m)
  => Evolution
```

Could be worth creating a `newtype Evolution = Evolution Word`

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -129,16 +139,41 @@ data HotKey c m = HotKey {
...
       --
       -- PRECONDITION: the key is not poisoned.
       --
       -- POSTCONDITION: the signature is in normal form.
     , sign_      :: forall toSign. (SL.KESignable c toSign, HasCallStack)
-                 => toSign -> m (SL.SignedKES c toSign)
+                 => toSign
+                 -> m (SL.SignedKES c toSign)
+      -- | Securely erase the key and release its memory.
+    , forget :: m ()
```
jasagredo, line 153:
```suggestion
                 -> m (SL.SignedKES c toSign)

      -- | Securely erase the key and release its memory.
    , forget :: m ()
```

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -129,16 +139,41 @@ data HotKey c m = HotKey {
       -- | Return 'KESInfo' of the signing key.
     , getInfo    :: m KESInfo
       -- | Return 'True' when the signing key is poisoned because it expired.
+    , getOCertMaybe :: m (Maybe (OCert.OCert c))
     , isPoisoned :: m Bool
```
jasagredo, line 143:
```suggestion
      -- | Some haddock here
    , getOCertMaybe :: m (Maybe (OCert.OCert c))
      -- | Return 'True' when the signing key is poisoned because it expired.
    , isPoisoned :: m Bool
```

jasagredo (changes requested):

jasagredo:
I do wonder if this impacts the forging loop. I have been following the call hierarchy a bit, and it turns out that:

- `checkShouldForge` → `updateForgeState` → `HotKey.evolve`
- `forgeBlock` → `forgeShelleyBlock` → `mkHeader` → `forgePraosFields` → `HotKey.sign`

Will these invoke the KES agent? if so, they might impact the forging loop and therefore block production and diffusion.

tdammers:
> I do wonder if this impacts the forging loop. I have been following the call hierarchy a bit, and it turns out that:
> 
>     * `checkShouldForge` → `updateForgeState` → `HotKey.evolve`
> 
>     * `forgeBlock` → `forgeShelleyBlock` → `mkHeader` → `forgePraosFields` → `HotKey.sign`
> 
> 
> Will these invoke the KES agent? if so, they might impact the forging loop and therefore block production and diffusion.

No.

The way the KES agent connectivity will work is that `mkHotKey` will spawn a separate thread that connects to a KES agent and replaces the credentials stored inside the `HotKey` whenever the KES agent sends any. `evolve` and `sign` will block for a short amount of time while the ke is being replaced, but other than that, the block forging will be completely independent from the KES agent connection.

The only situation where this can block the block forging is when the block forging thread first starts up; at this point, it will not have a KES key until a connection to the KES agent is established and a KES key is received. This is different from the current situation, where the KES key is loaded from disk before the block forging thread starts up.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
tdammers: I think this is a leftover from a previous design that we don't actually need anymore. Will delete.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Unary.hs
tdammers: Looks like a leftover.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
tdammers: Could you confirm this somehow? I've never heard of this before, and I'd hate to make the code more verbose for no good reason.

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Protocol/Praos.hs
tdammers: Indeed this is no longer needed.

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
tdammers: Good idea.

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Protocol/Praos.hs
tdammers: Will no longer need this.

ouroboros-consensus/changelog.d/20250130_093251_tdammers_mlocked_kes_rebase.md
tdammers: Nope.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -148,60 +183,126 @@ sign = sign_
...
+              KESKeyPoisoned ->
+                error "trying to sign with a poisoned key"
+              KESKey _ key -> do
+                let evolution = kesEvolution kesStateInfo
+                SL.signedKES () evolution toSign key
+      , forget = do
+          modifyMVar_ varKESState $ poisonState
+      , set = \newOCert newKey evolution startPeriod@(Absolute.KESPeriod start) -> do
+          modifyMVar_ varKESState $ \oldState -> do
+            _ <- poisonState oldState
```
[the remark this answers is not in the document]
tdammers: To replace the sign key. The idea here is that we can initialize a HotKey in a poisoned state, and put a fresh key in later.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
tdammers: Yes, but that would have to live in `cardano-base` (and/or `cardano-ledger`), where `updateKES` is defined.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -148,60 +183,126 @@ sign = sign_
...
 kesKeyIsPoisoned :: KESKey c -> Bool
 kesKeyIsPoisoned KESKeyPoisoned = True
-kesKeyIsPoisoned (KESKey _)     = False
+kesKeyIsPoisoned (KESKey _ _)     = False
 
 data KESState c = KESState {
-      kesStateInfo :: !KESInfo
-    , kesStateKey  :: !(KESKey c)
+      kesStateInfo  :: !KESInfo
+    , kesStateKey   :: !(KESKey c)
```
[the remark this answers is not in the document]
tdammers: No idea, must be some sort of editing or rebasing artifact.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
tdammers: Note that "because it expired" is wrong - a key can be in a poisoned state for a variety of reasons, e.g. because no key has ever been set for this 'HotKey', or because it has been explicitly erased.

kevinhammond:
Can you quantify “for a short time” please.  SPOs are quite sensitive to timing issues during block forging.  if it’s significant, there might be operational procedures that could mitigate the issue (eg if it’s related to the timing of KES key evolution or KES agent restart)?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
jasagredo: I was just pointing to the fact that that the haddocks were mixed up, you introduced `getOCertMaybe` in between `isPoisoned` and its haddock.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/MiniProtocol/LocalTxSubmission/Server.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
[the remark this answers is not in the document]
tdammers: Yes.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -854,24 +855,26 @@ protocolInfoCardano paramsCardano
...
 
+        (ocert, sk) <- instantiatePraosCredentials (praosCanBeLeaderCredentialsSource canBeLeader)
+
+        let startPeriod :: Absolute.KESPeriod
+            startPeriod = Absolute.ocertKESPeriod ocert
+
+        let maxKESEvo :: Word64
+            maxKESEvo = assert (tpraosMaxKESEvo == praosMaxKESEvo) praosMaxKESEvo
+
+        hotKey :: HotKey.HotKey c m <- HotKey.mkHotKey
```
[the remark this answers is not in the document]
tdammers: Turns out we don't actually need this type annotation.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -827,8 +828,8 @@ protocolInfoCardano paramsCardano
     -- credentials. If there are multiple Shelley credentials, we merge the
     -- Byron credentials with the first Shelley one but still have separate
     -- threads for the remaining Shelley ones.
-    blockForging :: m [BlockForging m (CardanoBlock c)]
-    blockForging = do
+    mkBlockForgings :: m ([BlockForging m (CardanoBlock c)])
```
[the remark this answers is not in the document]
tdammers: Mass editing artifact, sorry.

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -38,6 +38,9 @@ module Ouroboros.Consensus.Cardano.Node (
   , CardanoHardForkTriggers (.., CardanoHardForkTriggers', triggerHardForkShelley, triggerHardForkAllegra, triggerHardForkMary, triggerHardForkAlonzo, triggerHardForkBabbage, triggerHardForkConway)
   , CardanoProtocolParams (..)
   , MaxMajorProtVer (..)
+  , ProtocolParamsByron
+  , ProtocolParamsShelleyBased
+  , CheckpointsMap
```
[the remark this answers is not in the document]
tdammers: Probably a rebasing artifact? Not sure.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Common.hs
```diff
@@ -48,12 +48,7 @@ import           Ouroboros.Consensus.Storage.ImmutableDB
 -------------------------------------------------------------------------------}
 
 data ShelleyLeaderCredentials c = ShelleyLeaderCredentials
```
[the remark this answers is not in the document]
tdammers: They still contain a VRF sign key, and at least a reference to the actual credentials (either an unsound KES sign key, or the address of a KES agent that will serve KES keys dynamically).

The naming is also consistent other `LeaderCredentials` types, such as `ByronLeaderCredentials`, which doesn't use KES and just contains the keys needed to forge blocks in the Byron era, so I'd say if we change the name of `ShelleyLeaderCredentials`, we also need to change the name of `ByronLeaderCredentials`. Which of course we could do, but I'm trying to limit the scope of the changes here.

ouroboros-consensus-cardano/changelog.d/20250130_093803_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,18 @@
...
+
+- Use new mlocked KES API for all internal KES sign key handling.
+- Add finalizers to all block forgings (required by `ouroboros-consensus`).
+- Change `HotKey` to manage not only KES sign keys, but also the corresponding
+  OpCerts. This is in preparation for KES agent connectivity: with the new
+  design, the KES agent will provide both KES sign keys and matching OpCerts
+  together, and we need to be able to dynamically replace them both together.
+- Add finalizer to `HotKey`. This takes care of securely forgetting any KES
+  keys the HotKey may still hold, and will be called automatically when the
+  owning block forging terminates.
```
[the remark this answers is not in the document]
tdammers: It is, and it's also documented in the changelog there. Removing.

ouroboros-consensus-cardano/changelog.d/20250130_093803_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,18 @@
...
+- Change `ShelleyLeaderCredentials` to not contain the KES sign key itself
+  anymore. Instead, the `CanBeLeader` data structure now contains a
+  `praosCanBeLeaderCredentialsSource` field, which specifies how to obtain the
+  actual credentials (OpCert and KES SignKey).
+- The `KesKey` data type in `unstable-cardano-tools` has been renamed to
+  `UnsoundPureKesKey`, to reflect the fact that it uses the old, unsound KES
+  API (which does not use mlocking or secure forgetting).
```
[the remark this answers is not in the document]
tdammers: Why not? It's a breaking change to the public API of that library.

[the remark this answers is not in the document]
jasagredo: We purposefully named the testing libraries `unstable-*` to convey the information that they might change without notice. I'm trying to find somewhere we had written down that rule but I couldn't find it. Maybe we missed documenting it?

ouroboros-consensus/ouroboros-consensus.cabal
jasagredo: I'm surprised by this, I thought the `ghc-options: -Wno-unused-packages` on the line above would take precedence over your custom options, but probably I was wrong. User options are always the ones that dominate.

So yeah, a normal build won't complain (even if the package was unused to the eyes of Cabal) but if you specify `-Wunused-packages` explicitly then it will.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
jasagredo: Omit my comment for now and let's see what CI says. If we can avoid this change, even better.

tdammers:
> Can you quantify “for a short time” please. SPOs are quite sensitive to timing issues during block forging. if it’s significant, there might be operational procedures that could mitigate the issue (eg if it’s related to the timing of KES key evolution or KES agent restart)?

Depends on the scenario.

Evolving keys will temporarily block the block forging threads for the time it takes to evolve the key; however, in the current code, the block forging thread itself will evolve keys as needed before attempting to forge, so that delay is going to be the same, modulo thread switching overhead.

Pushing entirely new keys works differently. In the current situation, installing a new key requires shutting down the node kernel, reloading the configuration from disk, and restarting the node kernel. In the new situation, a KES agent can just push a new key, and it will be replaced in the live block forging threads. The block forging will be blocked for the time it takes to swap out the key, but since this is effectively just an in-memory pointer swap, it's going to be extremely fast - much faster than restarting the entire node kernel.

The only situation where an extra delay is introduced that wasn't there is when the node kernel first starts up. Right now, we load the KES key from disk before starting up the node kernel, and when the block forging thread starts, it already has the KES key. In the new situation, we will instead pass it the address of a KES agent, so the block forging thread will not have the KES key yet when it starts up, and it will block until it has established a KES agent connection and received a valid KES key. This process shouldn't take significantly longer than loading a KES key from disk (it may, in fact, even be faster, since the KES agent serves the key from RAM), but the delay occurs at a different point, after starting the block forging thread rather than before. It also means that if the KES agent connection fails, the block forging thread may remain blocked indefinitely.

In other words, failing to obtain a valid KES key will no longer prevent the block forging threads from starting, it will instead block them indefinitely.

TL;DR: I expect the overall delays to be shorter, but the order of operations during startup is different, causing potentially different failure modes.

cabal.project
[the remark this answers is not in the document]
tdammers: Agree. This PR depends on another PR anyway, so it's not in a merge-ready state. I just want to have it as ready as I can.

ouroboros-consensus-diffusion/changelog.d/20250130_100651_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,5 @@
+### Non-Breaking
+
+- Use registry to ensure block forging threads are finalized in node kernel.
+- Tests have been adapted to match the API changes in
+  `ouroboros-consensus` and `ouroboros-consensus-cardano`.
```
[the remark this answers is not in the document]
tdammers: Fair enough.

ouroboros-consensus-cardano/changelog.d/20250130_093803_tdammers_mlocked_kes_rebase.md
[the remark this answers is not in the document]
tdammers: Alright, makes sense.

kevinhammond:
Hi @tdammers 

Thanks!  I'm trying to make sure we can answer any questions that SPOs might have, so that we encourage adoption

> Evolving keys will temporarily block the block forging threads for the time it takes to evolve the key; however, in the current code, the block forging thread itself will evolve keys as needed before attempting to forge, so that delay is going to be the same, modulo thread switching overhead.
> 
> Pushing entirely new keys works differently. In the current situation, installing a new key requires shutting down the node kernel, reloading the configuration from disk, and restarting the node kernel. In the new situation, a KES agent can just push a new key, and it will be replaced in the live block forging threads. The block forging will be blocked for the time it takes to swap out the key, but since this is effectively just an in-memory pointer swap, it's going to be extremely fast - much faster than restarting the entire node kernel.

Part of the point is that SPOs can currently control exactly when the KES evolution takes place, so can avoid doing this when they're scheduled to make a block.  That is, they can avoid any timing impact if they are careful with their procedures.  In the new system, I think they would need to kill/suspend/block the KES agent to achieve the same effort (something that is presumably undesirable).

> In other words, failing to obtain a valid KES key will no longer prevent the block forging threads from starting, it will instead block them indefinitely

This situation is interesting since this could impact unattended operation/scheduled restarts.  It might be possible to mitigate this by having OS process monitors (i.e. make sure the KES agent is running before starting a block producer).  Would the node proceed "silently" in this case (i.e. would it look as if it was working, but just fail to make blocks?).
Do we know how it is currently detected (a process monitor can presumably check whether the block forging thread has started).  Basically, this looks like a change that needs to be considered by those writing e.g. CNCLI

(I wouldn't be too worried about small increases in startup time, of course - that can be accounted for).

Also, I'm assuming there's absolutely no need to have the KES agent running if you are e.g. operating relays/a full node wallet or some other service (I can't see why you would need to do this since KES keys are only needed for block producers!)

tdammers:
> Part of the point is that SPOs can currently control exactly when the KES evolution takes place, so can avoid doing this when they're scheduled to make a block. That is, they can avoid any timing impact if they are careful with their procedures. In the new system, I think they would need to kill/suspend/block the KES agent to achieve the same effort (something that is presumably undesirable).

OK, so I've been assuming that each KES evolution is valid for the exact duration of its corresponding 36-hour period, which implies that there is no point doing the evolution any earlier or later than the exact moment the period flips over. The code as it currently is is in line with this - the block forging threads will always check whether the KES key needs evolving before attempting to forge a block (this has been the case previously, and still is), and AFAICT there is no way for SPOs to control exactly when that happens. Please correct me if I'm mistaken about this.

Also note that the KES agent does not interact with the node in order to evolve KES keys - this happens independently on both sides. The only reason the KES agent evolves keys at all is to ensure forward security on that end (we need to erase old evolutions of the key *everywhere*, not just inside the node process), but it will not push out a new key just because it has evolved. Killing, suspending, or blocking the KES agent will change absolutely nothing about how keys are evolved inside the node process.

The only thing SPOs *can* control is when an entirely new KES key is installed, but this will still be the case with the KES agent system, because a KES agent will only push a key in two situations:

- When a node starts up and first connects to the agent (delaying this makes no sense, because without that first key, you can't forge anything anyway).
- When the user installs a new key into the agent. However, this requires a manual intervention, just like the process it replaces, so SPOs are still completely in control as to when this happens.

> This situation is interesting since this could impact unattended operation/scheduled restarts.

Absolutely, yes. Although you can say the same about invalid configurations that prevent a node process from starting up (or, worse, that would cause a node process to start up, but not become a block forging node).

> It might be possible to mitigate this by having OS process monitors (i.e. make sure the KES agent is running before starting a block producer).

It doesn't need to run before starting the block producer, it just needs to run at some point before the block producer is supposed to start producing blocks. If you start both together, having the agent start a couple milliseconds after the node process wouldn't be an issue.

The KES agent is definitely not supposed to be down for extended amounts of time though, so the plan is to run it as a systemd unit and hook into the OS' logging and process management infrastructure. But of course it can also run under some other process management system.

> Would the node proceed "silently" in this case (i.e. would it look as if it was working, but just fail to make blocks?).

There will be trace logging (currently working on that), and we could possibly add some functionality for querying the block forging state of a node, something like:

- "Not a block forging node": this node has not been configured as a block forging node
- "No agent": this node has been configured to load keys from a KES agent, but hasn't been able to make a connection on the configured address
- "No key": this node has successfully connected to a KES agent, but hasn't received any KES key yet
- "Forging": this node is currently forging blocks

One thing that might complicate matters a bit is that it's possible for a node to be actively forging blocks despite having temporarily lost the connection to its KES agent - as long as it still holds a valid KES key (or one that can be evolved into a valid one), it will happily continue, make regular reconnection attempts, and pick up any new KES keys once the connection comes back.

> Do we know how it is currently detected (a process monitor can presumably check whether the block forging thread has started).

No, I don't know how that works. I haven't looked into node / CLI source code at all so far. My guess would be that the best way to monitor this would be via trace events, or by directly querying the node kernel about its block forging threads, if any.

> Also, I'm assuming there's absolutely no need to have the KES agent running if you are e.g. operating relays/a full node wallet or some other service (I can't see why you would need to do this since KES keys are only needed for block producers!)

Correct. If you don't need to forge any blocks, you can ignore the KES agent entirely. Just like you wouldn't configure an initial KES key with the current setup, you would simply not configure a KES agent address, and the node would run as a non-block-forging instance.

cabal.project
```diff
@@ -26,6 +26,8 @@ packages:
   sop-extras
   strict-sop-core
 
+allow-newer: plutus-core:cardano-crypto-class
```
jasagredo, line 29: This is something that we will have to address for the 10.3 release, so I hope it disappears before we merge this PR.

ouroboros-consensus-cardano/changelog.d/20250130_093803_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,11 @@
+### Breaking
+
+- Use new mlocked KES API for all internal KES sign key handling.
+- Add finalizers to all block forgings (required by `ouroboros-consensus`).
+- Change `ShelleyLeaderCredentials` to not contain the KES sign key itself
+  anymore. Instead, the `CanBeLeader` data structure now contains a
+  `praosCanBeLeaderCredentialsSource` field, which specifies how to obtain the
+  actual credentials (OpCert and KES SignKey).
+- The `KesKey` data type in `unstable-cardano-tools` has been renamed to
```
jasagredo, line 9: We do not reflect the unstable libraries in the changelogs, as described in the project Readme

ouroboros-consensus-diffusion/changelog.d/20250130_100651_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,3 @@
+### Non-Breaking
+
+- Use registry to ensure block forging threads are finalized in node kernel.
```
jasagredo, line 3: "Use registry" is a bit ambiguous. I would instead say
```suggestion
- Ensure that block forging threads erase their keys on shutdown in node kernel.
```

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -126,19 +135,58 @@ data HotKey c m = HotKey {
...
+             -- ^ The new KES key
+          -> Word
+             -- ^ The new KES key's current evolution
+          -> Absolute.KESPeriod
+             -- ^ Start period (relative to the KES key's 0th evolution)
+          -> m ()
+
+      -- | Release any resources held by the 'HotKey'. Must be run exactly once
+      -- per 'HotKey'.
+    , finalize :: m ()
```
jasagredo, line 178: Hoy is `finalize` different from `forget`? Or how do those relate, should one call both?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
jasagredo: Even if we don't create the new type, please rename the function.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -245,16 +251,30 @@ instance Crypto c => ChainOrder (PraosChainSelectView c) where
...
-    praosCanBeLeaderSignKeyVRF :: !(SL.SignKeyVRF c)
+    praosCanBeLeaderSignKeyVRF :: !(SL.SignKeyVRF c),
+    praosCanBeLeaderCredentialsSource :: !(PraosCredentialsSource c)
   }
   deriving (Generic)
 
-instance Crypto c => NoThunks (PraosCanBeLeader c)
+instance (NoThunks (KES.UnsoundPureSignKeyKES (KES c)), Crypto c) => NoThunks (PraosCanBeLeader c)
+
+data PraosCredentialsSource c
```
jasagredo, line 263: Some haddocks would be nice here. There is no explanation neither here or in the `praosCanBeLeaderCredentialSource`. 

To clarify, I think such haddock should be either here or in both places.

cabal.project
tdammers: I hope so too, but it's not under my control. We need a newer `cardano-crypto-class`, so the only way to make that compatible with plutus is to upgrade the dependency there and then depend on the upgraded plutus ourselves.

ouroboros-consensus-diffusion/changelog.d/20250130_100651_tdammers_mlocked_kes_rebase.md
tdammers: It's not necessarily just erasing; finalizing may also involve shutting down a KES agent connection in the future.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
tdammers: `forget` erases the key itself, `finalize` deals with any other resources that may be associated with the HotKey.

You just pointed me at a bug, though: we provide a `forget` method that erases the key, but we are not actually using it anywhere, so we actually leak mlocked memory on shutdown.

cabal.project
tdammers: Actually, it looks like this is no longer needed.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
tdammers: Fixed that by renaming `finalize` to `finalize_`, and exporting a wrapper, `finalize`, that calls both.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
tdammers: Done.

cabal.project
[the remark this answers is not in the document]
tdammers: Actually, ledger is the last one that we need to wait for.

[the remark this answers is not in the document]
jasagredo: I think the update done in 10.3 (https://github.com/IntersectMBO/cardano-ledger/commit/f5b35692b4d92d664187781a98b7af3fab445bad) will probably be enough, so once the 10.3 branch (`aniketd/release-10-3`) is merged we could merge this one

cabal.project
jasagredo: It is a bit complicated because of some dancing in dependency bounds (a revision expanded the bounds, a later revision restricted them again). Let's see after the release for 10.3 I think this will not be needed. (but I'm surprised you don't seem to need it anymore in this branch?)

cabal.project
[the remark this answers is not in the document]
tdammers: Yes. Or, well, we also need there to be an actual release so we can pull that version from CHaP.

PR #1379, opened 2025-02-03
Integrate with ouroboros-network-0.19.0.2

coot:
# Description

Integration with `ouroboros-diffusion-0.19.0.2`.


jasagredo (approved):
@coot You need to do `nix flake update CHaP`. I already approve the PR so that it merges when you do so.

PR #1382, opened 2025-02-06
Improvements to UTxO-HD: mempool snapshotting

jasagredo:
This PR accumulates some improvements for UTxO-HD that result in benefits for mempool snapshotting, which is part of the critical path for block minting (and therefore block diffusion). Running in a full-block synthetic chain shows these times for 50 consecutive mempool snapshots:

- pre: at utxo-hd-main before this PR
- caf: pre + moving the Cardano tx out translations to a CAF
- no-diffs: caf + do not compute diffs when snapshotting the mempool
- no-shtxin: no-diffs + remove the `ShelleyTxIn` newtype

plot

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
@@ -62,6 +62,8 @@ import           Data.SOP.InPairs (InPairs (..))
 import qualified Data.SOP.InPairs as InPairs
 import qualified Data.SOP.Match as Match
 import           Data.SOP.Strict
+import           Data.SOP.Tails (Tails)
+import qualified Data.SOP.Tails as Tails
 import           Data.SOP.Telescope (Telescope (..))
 import qualified Data.SOP.Telescope as Telescope
 import           Data.Typeable
@@ -1113,6 +1115,14 @@ class ( Show (HardForkTxOut xs)
   txOutEjections      :: NP (K (NS WrapTxOut xs) -.-> WrapTxOut) xs
   default txOutEjections :: CanHardFork xs => NP (K (NS WrapTxOut xs) -.-> WrapTxOut) xs
   txOutEjections = composeTxOutTranslations $ ipTranslateTxOut hardForkEraTranslation
+
+  txOutTails :: Tails (InPairs.Fn2 WrapTxOut) xs
+  default txOutTails :: CanHardFork xs => Tails (InPairs.Fn2 WrapTxOut) xs
+  txOutTails =
+    Tails.inPairsToTails
+     $ InPairs.hmap
+      (\translator -> InPairs.Fn2 $ WrapTxOut . translateTxOutWith translator . unwrapTxOut)
+      (translateLedgerTables (hardForkEraTranslation @xs))
 
 instance (CanHardFork xs, HasHardForkTxOut xs)
       => CanUpgradeLedgerTables (LedgerState (HardForkBlock xs)) where
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
@@ -1112,13 +1112,20 @@ class ( Show (HardForkTxOut xs)
   -- that we only compute it once, then it is cached for the duration of the
   -- program, as we will use it very often when converting from the
   -- HardForkBlock to the particular @blk@.
+  --
+  -- This particular method is useful when our HardForkBlock uses
+  -- DefaultHardForkTxOut, so that we can implement inject and project.
   txOutEjections      :: NP (K (NS WrapTxOut xs) -.-> WrapTxOut) xs
   default txOutEjections :: CanHardFork xs => NP (K (NS WrapTxOut xs) -.-> WrapTxOut) xs
   txOutEjections = composeTxOutTranslations $ ipTranslateTxOut hardForkEraTranslation
 
-  txOutTails :: Tails (InPairs.Fn2 WrapTxOut) xs
-  default txOutTails :: CanHardFork xs => Tails (InPairs.Fn2 WrapTxOut) xs
-  txOutTails =
+  -- | This method is a null-arity method in a typeclass to make it a CAF, such
+  -- that we only compute it once, then it is cached for the duration of the
+  -- program, as we will use it very often when converting from the
+  -- HardForkBlock to the particular @blk@.
+  txOutTranslations :: Tails (InPairs.Fn2 WrapTxOut) xs
+  default txOutTranslations :: CanHardFork xs => Tails (InPairs.Fn2 WrapTxOut) xs
+  txOutTranslations =
     Tails.inPairsToTails
      $ InPairs.hmap
       (\translator -> InPairs.Fn2 $ WrapTxOut . translateTxOutWith translator . unwrapTxOut)
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
@@ -12,6 +12,7 @@ module Ouroboros.Consensus.Ledger.SupportsMempool (
   , ByteSize32 (..)
   , ConvertRawTxId (..)
   , GenTx
+  , ComputeDiffs (..)
   , GenTxId
   , HasByteSize (..)
   , HasTxId (..)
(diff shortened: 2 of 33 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -1114,6 +1116,14 @@ class ( Show (HardForkTxOut xs)
   default txOutEjections :: CanHardFork xs => NP (K (NS WrapTxOut xs) -.-> WrapTxOut) xs
   txOutEjections = composeTxOutTranslations $ ipTranslateTxOut hardForkEraTranslation
 
+  txOutTails :: Tails (InPairs.Fn2 WrapTxOut) xs
```
nfrisby, line 1119: Need a comment on this method, like for `txOutEjections` just above it.

nfrisby:
I edited the PR description a little.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
nfrisby: It's a little awkward that we never intend for any `instance` to override this default 🤔.

I suppose the only alternative would be to define the polymorphic "CAF" and give it a SPECIALISE pragma. But... then we'd have to make sure there's a SPECIALISE chain that reaches from the Consensus entry point all the way to every use, which is a real headache. So, the awkwardness seems worthwhile 👍

With the current approach, the dictionary _apparently_ handles that specialization chain for us. But I'm not exactly sure why it's able to do so, which makes me a little nervous. We have `instance CardanoHardForkConstraints c => HasHardForkTxOut (CardanoEras c)`, which means that instance function (ie the thing that builds the `HasHardForkTxOut` dictionary from the other dictionaries) _might_ get used in some code location that needs `HasHardForkTxOut` and already has a `CardanoHardForkConstraints` context. Since the `txOutTails` default does depend on that instance context (since it has `CanHardFork` in its own context, and our Cardano instance of that also requires `CardanoHardForkConstraints`), it wouldn't be shared among different uses of this instance function.

But, your measurements show that it does help. I just worry that we might accidentally spoil that someday, since it's not entirely transparent to us why tucking this polymorphic "CAF" inside a type class does actually end up using a proper CAF.

(This same concern applies to `txOutEjections` --- I just hadn't thought through the details when originally suggesting it.)

Edit: I guess as of the Ledger Team's monomorphic crypto work, the explanation will become much simpler, since `CardanoHardForkConstraints` only has one index, `c`! At that point, it'll be obvious that a sufficient requirement is for `CanHardFork XS` to be in-scope for the `HasHardForkTxOut XS` instance.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -79,6 +80,16 @@ data WhetherToIntervene
     -- them for it.
     deriving (Show)
 
+-- | Whether to keep track of the diffs produced by applying the transactions.
+--
+-- When getting a mempool snapshot, we will revalidate all the
+-- transactions but we won't do anything useful with the resulting
+-- state. We can safely omit computing the differences in this case.
```
nfrisby, line 87: Please amend this comment to also discuss:

- This is a worthwhile optimization (and has been measured as such), since snapshotting the mempool is on the critical path of block minting.
- Eventually, the UTxO HD plan has always been for the ledger rules to construct the differences, instead of the Consensus layer computing them retroactively via `calculateDifferences`. Once that's true, we hope this optimization will no longer be worthwhile. That's part of the reason that we're content to just use a boolean isomorph instead and and yield empty `Diff`s instead of making it a GADT that makes the precise codomain (or just having two different functions, one with `Diff`s in the codomain and one without).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -79,6 +80,16 @@ data WhetherToIntervene
...
+data ComputeDiffs
+  = ComputeDiffs
```
nfrisby, line 89: Comment saying that this option should be used with resyncing the mempool with an updated selection.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -79,6 +80,16 @@ data WhetherToIntervene
...
+  | IgnoreDiffs
```
nfrisby, line 90: Comment saying that this option should be used when snapshotting the mempool when minting a block.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/Utils.hs
```diff
@@ -306,3 +308,12 @@ restrictValues' ::
      (SameUtxoTypes l l'', SameUtxoTypes l' l'', HasLedgerTables l, HasLedgerTables l', HasLedgerTables l'')
   => l ValuesMK -> l' KeysMK -> LedgerTables l'' ValuesMK
 restrictValues' l1 l2 = ltliftA2 restrictValuesMK (ltprj l1) (ltprj l2)
+
+---
+
+unionValues ::
```
nfrisby, line 314: My intuition is that we should have a comment here explaining that collisions are impossible for this first phase of UTxO HD.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -82,6 +83,9 @@ data InternalState blk = IS {
       -- This should always be in-sync with the transactions in 'isTxs'.
     , isTxIds        :: !(Set (GenTxId blk))
 
+    , isTxKeys       :: !(LedgerTables (LedgerState blk) KeysMK)
```
nfrisby, line 86: These two fields definitely need comments, which can probably state a pretty simple INVARIANT with respect to `isTxs`. I'd imagine it can be exact for `isTxKeys` and just a little hand-wavy for `isTxValues`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -350,6 +363,43 @@ data RevalidateTxsResult blk =
...
+        reapplyTxs IgnoreDiffs cfg slot theTxs
+        $ applyDiffForKeysOnTables
+              values
+              (Foldable.foldMap' (getTransactionKeySets . txForgetValidated . fst) theTxs)
+              st
+
+  in snapshotFromIS $ IS {
+         isTxs          = TxSeq.fromList $ map unwrap val
+       , isTxIds        = Set.fromList $ map (txId . txForgetValidated . fst) val
+       , isTxKeys       = emptyLedgerTables
```
nfrisby, line 394: I think a comment could help explain why `isTxKeys` and `isTxValues` can soundly be empty here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -33,25 +32,32 @@ implGetSnapshotFor mpEnv slot ticked readUntickedTables = do
...
-         else newInternalState
-            $ revalidateTxsFor
-                capacityOverride
-                cfg
-                slot
-                ticked
-                values
-                (isLastTicketNo is)
-                (TxSeq.toList $ isTxs is)
+    else if pointHash (isTip is) == castHash (getTipHash ticked)
```
nfrisby, line 35: I think there should be one call to `computeSnapshot`, where the `values` has been conditionally recovered either from `isTxValues` or from `isTxKeys`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables.hs
```diff
@@ -295,13 +295,15 @@ valuesMKDecoder ::
...
  where
   go :: Int
-     -> CBOR.Decoder s (ValuesMK (TxIn l) (TxOut l))
-  go len =
-    ValuesMK . Map.fromList
-      <$> replicateM len (unpackError @(TxIn l, TxOut l) @ByteString <$> fromCBOR)
+     -> Map.Map (TxIn l) (TxOut l)
+     -> CBOR.Decoder s (Map.Map (TxIn l) (TxOut l))
+  go 0 m = pure m
+  go !len m = do
```
nfrisby, line 304: I think the only place you need bangs are `len` and `m`, yeah? Since you're using the `Data.Map.Strict` interface. (Which would plug the leak in `m` here, also.)

nfrisby (changes requested):
Mostly comments, fewer bangs, etc.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
jasagredo: Yeah, I think `CardanoHardForkConstraints` will basically disappear with the Crypto work!

>We have instance CardanoHardForkConstraints c => HasHardForkTxOut (CardanoEras c), which means that instance function (ie the thing that builds the HasHardForkTxOut dictionary from the other dictionaries) might get used in some code location that needs HasHardForkTxOut and already has a CardanoHardForkConstraints context. Since the txOutTails default does depend on that instance context (since it has CanHardFork in its own context, and our Cardano instance of that also requires CardanoHardForkConstraints), it wouldn't be shared among different uses of this instance function.

That is also my understanding but it shows clear improvements, not sure why

jasagredo:
I dismissed the request for changes because the request was mostly about comments and I added those.

Issue #1383, opened 2025-02-07
[FEAT] - Mempool could do operations at the "ledgerstate" level to prevent projection and injection


jasagredo:
The mempool makes some operations on ledger tables when adding transactions, take for example this left-heavy flamegraph from a db-analyser run on a full-blocks synthetic chain:

![Image](https://github.com/user-attachments/assets/5c812849-3f70-472d-a8e1-7fd6b24014c7)

Roughly 60% of the time of applying the transactions is spent in prepending the differences and in applying the diffs to the tables. And of that the 100% of it is spent in projecting and injecting the tables. 

The reason behind this is that `prependDiffs` for example is defined in terms of ledger tables:
```
rawPrependDiffs ::
     Ord k
  => DiffMK k v -- ^ Earlier differences
  -> DiffMK k v -- ^ Later differences
  -> DiffMK k v
rawPrependDiffs (DiffMK d1) (DiffMK d2) = DiffMK (d1 <> d2)

-- | Prepend diffs from the first ledger state to the diffs from the second
-- ledger state. Returns ledger tables.
prependDiffs' ::
     (SameUtxoTypes l l'', SameUtxoTypes l' l'', HasLedgerTables l, HasLedgerTables l', HasLedgerTables l'')
  => l DiffMK -> l' DiffMK -> LedgerTables l'' DiffMK
prependDiffs' l1 l2 = ltliftA2 rawPrependDiffs (ltprj l1) (ltprj l2)

-- | Prepend the diffs from @l1@ to @l2@. Returns @l2@.
prependDiffs ::
     (SameUtxoTypes l l', HasLedgerTables l, HasLedgerTables l')
  => l DiffMK -> l' DiffMK -> l' DiffMK
prependDiffs l1 l2 = ltwith l2 $ prependDiffs' l1 l2
```

We use `ltprj` and `ltwith` with go from ledger states to tables. This imposes in the HFC case a conversion between CardanoTxOut and ShelleyTxOut, which is a newtype-like conversion as tables are always in the latest era.

To mitigate this, we could lift these two operations to work in ledgerstates so that we don't extract the tables ever.

In any case the scale of time of this optimization should be negligible in the grand picture.

PR #1384, opened 2025-02-07
Move a couple of mempool operations to work on ledgerstates instead

jasagredo:
# Description

10.2
image

UTxO-HD before this:
image

UTxO-HD after this:
image

The changes are in line with what is described in #1383. There are still some more operations that could be lifted to ledger states.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
@@ -29,6 +29,7 @@ module Ouroboros.Consensus.HardFork.Combinator.Mempool (
 
 import           Control.Arrow (first, (+++))
 import           Control.Monad.Except
+import           Data.Functor.Identity
 import           Data.Functor.Product
 import           Data.Kind (Type)
 import qualified Data.Measure as Measure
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -185,6 +185,26 @@ class ( UpdateLedger blk
   -- transaction size.
   getTransactionKeySets :: GenTx blk -> LedgerTables (LedgerState blk) KeysMK
 
+  -- Mempools live in a single slot so in the hard fork block case
```
nfrisby, line 188: Could you explicitly point out here the caveat we had talked through the other day: that the cost savings here is non-zero, but relatively small compared to the eg network costs of mempool synchronization? Unlike PR https://github.com/IntersectMBO/ouroboros-consensus/pull/1382, in particular. As as of PR 1382, this code is no longer on the critical path of block (minting and) propagation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
```diff
@@ -224,6 +226,36 @@ instance ( CanHardFork xs
...
 
+  prependMempoolDiffs
+    (TickedHardForkLedgerState _ (State.HardForkState st1))
+    (TickedHardForkLedgerState tr (State.HardForkState st2))
+    = TickedHardForkLedgerState
+          tr
+        $ State.HardForkState
+        $ runIdentity
+         (Tele.alignExtend
+         (InPairs.hpure undefined)
```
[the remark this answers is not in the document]
jasagredo: Done!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
jasagredo: Done

PR #1390, opened 2025-02-11
Expose `ValidationPolicy` interface

jasagredo:
# Description

Exposes `ValidationPolicy` and `ComputeLedgerEvents` that can be passed to control whether events are computed by the ledger and whether we should enable validation in the ledger layer.

This allows clients of our library to customize these arguments when using our classes (for example someone using the consensus layer to apply blocks in a streamed fashion).


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -338,49 +360,56 @@ instance ShelleyCompatible proto era
...
+        -> Except
+             (SL.BlockTransitionError era)
+             (STS.EventReturnType ep (Core.EraRule "BBODY" era) (SL.NewEpochState era))
+      appBlk ep' =
         SL.applyBlockOpts
           STS.ApplySTSOpts {
               asoAssertions = STS.globalAssertionPolicy
-            , asoValidation = STS.ValidateAll
-            , asoEvents     = STS.EPReturn
+            , asoValidation = validationMode (validation opts)
```
geo2a, line 402: Same question: we had some hard-coded options, why are we changing them?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -302,18 +312,30 @@ instance ShelleyBasedEra era => IsLedger (LedgerState (ShelleyBlock proto era))
...
+           STS.SingEP ep
+        -> STS.EventReturnType ep (Core.EraRule "TICK" era) (SL.NewEpochState era)
+      appTick ep' =
         SL.applyTickOpts
           STS.ApplySTSOpts {
-              asoAssertions = STS.globalAssertionPolicy
-            , asoValidation = STS.ValidateAll
-            , asoEvents     = STS.EPReturn
+              asoAssertions = STS.AssertionsOff
+            , asoValidation = STS.ValidateNone
```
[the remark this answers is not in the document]
jasagredo: Good catch! 

Alexey confirmed validation doesn't matter for ticking, but the assertions should remain as they were, see https://github.com/IntersectMBO/cardano-ledger/pull/4889/files#diff-0c5c12bd6d27ec3878c3fa4117d3c53c4159b653ec664997deb97b10d93d443bR170

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -174,28 +174,31 @@ instance IsLedger (LedgerState ByronBlock) where
...
         }
 
+deriving instance Generic CC.ValidationMode
+instance NoThunks CC.ValidationMode
+deriving instance Eq CC.ValidationMode
+deriving instance Generic CC.BlockValidationMode
+instance NoThunks CC.BlockValidationMode
+deriving instance Generic CC.TxValidationMode
+instance NoThunks CC.TxValidationMode
+
```
jasagredo, line 192: No longer needed

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Init.hs
```diff
@@ -231,13 +231,13 @@ initStartingWith tracer cfg stream initDb = do
         let replayed' :: Word64
             !replayed' = replayed + 1
 
-            events :: [LedgerEvent blk]
-            events = inspectLedger
+            evs :: [LedgerEvent blk]
```
geo2a, line 234: ~Why is this renaming necessary?~ Ah, I see, because of the `events` field in `ApplyOpts`. I suggest to rename `ApplyOpts.events` to `ApplyOpts.computeEvents`/`ApplyOpts.requestEvents` and keep `events` for these.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -83,9 +111,10 @@ class ( IsLedger l
   --
   -- This is passed the ledger state ticked to the slot of the given block, so
   -- 'applyChainTickLedgerResult' has already been called.
-  applyBlockLedgerResult ::
+  applyBlockLedgerResultWithOpts ::
```
geo2a, line 114: I personally do not see the benefit of renaming `applyBlockLedgerResult` to `applyBlockLedgerResultWithOpts`, as we do not have a variant without options.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -114,76 +163,88 @@ class ApplyBlock (LedgerState blk) blk => UpdateLedger blk
 -------------------------------------------------------------------------------}
 
 -- | 'lrResult' after 'applyBlockLedgerResult'
-applyLedgerBlock ::
+applyLedgerBlockWithOpts ::
```
geo2a, line 166: Same here, I think the `WIthOpts` suffix does not carry any useful information.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -114,76 +163,88 @@ class ApplyBlock (LedgerState blk) blk => UpdateLedger blk
...
+     forall l blk.
      (ApplyBlock l blk, HasCallStack)
-  => LedgerCfg l
+  => ApplyOpts
+  -> LedgerCfg l
   -> blk
   -> Ticked l
   -> Except (LedgerErr l) l
-applyLedgerBlock = fmap lrResult ..: applyBlockLedgerResult
+applyLedgerBlockWithOpts = fmap lrResult ...: applyBlockLedgerResultWithOpts
```
geo2a, line 174: Same here, I think the WIthOpts suffix does not carry any useful information.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -114,76 +163,88 @@ class ApplyBlock (LedgerState blk) blk => UpdateLedger blk
...
-  => LedgerCfg l
+  => ApplyOpts
+  -> LedgerCfg l
   -> blk
   -> l
   -> Except (LedgerErr l) (LedgerResult l l)
-tickThenApplyLedgerResult cfg blk l = do
-  let lrTick = applyChainTickLedgerResult cfg (blockSlot blk) l
-  lrBlock <-   applyBlockLedgerResult     cfg            blk  (lrResult lrTick)
+tickThenApplyLedgerResultWithOpts opts cfg blk l = do
```
geo2a, line 195: Same here, I think the WIthOpts suffix does not carry any useful information.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -114,76 +163,88 @@ class ApplyBlock (LedgerState blk) blk => UpdateLedger blk
...
+tickThenReapplyLedgerResult evs cfg blk l =
+  let lrTick  = applyChainTickLedgerResult evs cfg (blockSlot blk) l
+      lrBlock = reapplyBlockLedgerResult   evs cfg            blk (lrResult lrTick)
   in LedgerResult {
       lrEvents = lrEvents lrTick <> lrEvents lrBlock
     , lrResult = lrResult lrBlock
     }
 
-tickThenApply ::
+tickThenApplyWithOpts ::
```
geo2a, line 219: Same here, I think the WIthOpts suffix does not carry any useful information.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -114,76 +163,88 @@ class ApplyBlock (LedgerState blk) blk => UpdateLedger blk
...
+     forall l blk.
      ApplyBlock l blk
-  => LedgerCfg l
+  => ApplyOpts
+  -> LedgerCfg l
   -> blk
   -> l
   -> Except (LedgerErr l) l
-tickThenApply = fmap lrResult ..: tickThenApplyLedgerResult
+tickThenApplyWithOpts = fmap lrResult ...: tickThenApplyLedgerResultWithOpts
```
geo2a, line 227: Same here, I think the WIthOpts suffix does not carry any useful information.

geo2a (commented):
Overall looks good! I've left one question and several naming suggestions.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
jasagredo: This line is exactly the key to this PR. Now you can choose which options are passed to `applyBlockLedgerResult` and it is at this point where they are set.

geo2a: Right, makes sense, thanks!

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -129,24 +129,35 @@ storeLedgerParser = do
...
     (mconcat [ long "no-snapshot-checksum-on-write"
              , help (unlines [ "Don't calculate the checksum and"
                              , "write the '.checksum' file"
                              , "when taking a ledger snapshot"
                              ])
              ])
-  pure $ StoreLedgerStateAt slot ledgerValidation doChecksum
+  pure $ StoreLedgerStateAt slot opts doChecksum
+
+pApplyOpts :: Parser ApplyOpts
```
nfrisby, line 142: If `db-analyser` is in fact the only intended use case of `ValidationOff`, then I would strongly prefer to not define `ValidationOpts` in the `ouroboros-consensus` interface and rather isolate that flag within `db-analyser`.

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -129,24 +129,35 @@ storeLedgerParser = do
...
+  val <- flag ValidationOn ValidationOff
+    (  long "full-ledger-validation"
+    <> help (  "Use full block application while applying blocks to ledger states, "
+            <> "also validating signatures and scripts. "
+            <> "This is much slower than block reapplication (the default)."
+            )
+    )
+  evs <- flag ComputeLedgerEvents OmitLedgerEvents
+    (  long "compute-ledger-events"
+    <> help (  "Although the events will not be emitted anywhere (yet), "
```
nfrisby, line 153: Since `db-analyser` is often about performance, I think it'd help if you disambiguated "compute such events". Right now (due to the "yet"), the ledger would merely allocate thunks for the events, but nothing would ever force those thunks, right?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -405,9 +409,7 @@ storeLedgerStateAt slotNo ledgerAppMode doChecksum env = do
           storeLedgerState oldLedger
           pure (Stop, oldLedger)
 
-    tickThenXApply = case ledgerAppMode of
-        LedgerReapply -> pure ..: tickThenReapply
-        LedgerApply   -> tickThenApply
+    tickThenXApply = tickThenApply appOpts
```
nfrisby, line 412: As in my other comment: I'd much prefer this code to stay the way it was instead of polluting the core `ouroboros-consensus` code with that flag.

Or am I overlooking something?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/Metadata.hs
```diff
@@ -42,7 +42,7 @@ data Metadata = Metadata {
 instance ToJSON Metadata where
   toEncoding = Aeson.genericToEncoding Aeson.defaultOptions
 
-getMetadata :: LedgerApplicationMode -> IO Metadata
+getMetadata :: ApplyOpts -> IO Metadata
```
nfrisby, line 45: Looks like this only uses `ValidationOpts`, no the full `ApplyOpts`.

Is that intended?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -70,6 +70,7 @@ analyse DBAnalyserConfig{analysis, confLimit, dbDir, selectDB, validation, verbo
                   (const True)
                   (Node.stdMkChainDbHasFS dbDir)
                   (Node.stdMkChainDbHasFS dbDir)
+                  (ApplyOpts ValidationOff OmitLedgerEvents)
```
nfrisby, line 73: What's the relation between this `ValidationOff` and the `maybeValidateAll` wrapper?

Might deserve an inline comment clarifying.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -226,7 +226,7 @@ instance ApplyBlock (LedgerState BlockA) BlockA where
       setTip (TickedLedgerStateA st) = st { lgrA_tip = blockPoint blk }
 
   reapplyBlockLedgerResult =
-      dontExpectError ..: applyBlockLedgerResult
+      dontExpectError ...: (applyBlockLedgerResult . ApplyOpts ValidationOff)
```
nfrisby, line 229: No use of the default here?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/B.hs
```diff
@@ -186,11 +186,11 @@ instance IsLedger (LedgerState BlockB) where
...
     VoidLedgerEvent (LedgerState BlockB)
 
-  applyChainTickLedgerResult _ _ = pureLedgerResult . TickedLedgerStateB
+  applyChainTickLedgerResult _ _ _ = pureLedgerResult . TickedLedgerStateB
 
 instance ApplyBlock (LedgerState BlockB) BlockB where
-  applyBlockLedgerResult   = \_ b _ -> return $ pureLedgerResult $ LgrB (blockPoint b)
-  reapplyBlockLedgerResult = \_ b _ ->          pureLedgerResult $ LgrB (blockPoint b)
+  applyBlockLedgerResult = \_ _ b _ -> return $ pureLedgerResult $ LgrB (blockPoint b)
+  reapplyBlockLedgerResult       = \_ _ b _ ->          pureLedgerResult $ LgrB (blockPoint b)
```
nfrisby, line 193: Weird spacing

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
```diff
@@ -351,14 +350,14 @@ instance MockProtocolSpecific c ext
...
+  applyChainTickLedgerResult _ _ _ = pureLedgerResult . TickedSimpleLedgerState
 
 instance MockProtocolSpecific c ext
       => ApplyBlock (LedgerState (SimpleBlock c ext)) (SimpleBlock c ext) where
-  applyBlockLedgerResult = fmap pureLedgerResult ..: updateSimpleLedgerState
+  applyBlockLedgerResult _opts = fmap pureLedgerResult ..: updateSimpleLedgerState
 
   reapplyBlockLedgerResult =
-      (mustSucceed . runExcept) ..: applyBlockLedgerResult
+      (mustSucceed . runExcept) ...: (applyBlockLedgerResult . ApplyOpts ValidationOff)
```
nfrisby, line 360: No default use here?

ouroboros-consensus/src/unstable-tutorials/Ouroboros/Consensus/Tutorial/Simple.lhs
```diff
@@ -130,34 +130,34 @@ Next, we instantiate the `ConsensusProtocol` for `SP`:
 
 > instance ConsensusProtocol SP where
 >   type SelectView    SP = BlockNo
->
+
```
nfrisby, line 133: Please don't remove these bird feet. I think the intention is for the rendered file to have blank lines in it instead of multiple separate listings.

Or am I overlooking some `lhs` idiom?

ouroboros-consensus/src/unstable-tutorials/Ouroboros/Consensus/Tutorial/WithEpoch.lhs
```diff
@@ -73,16 +73,16 @@ And imports, of course:
 >   (BlockSupportsProtocol (..))
 > import Ouroboros.Consensus.Protocol.Abstract
 >   (ConsensusConfig, SecurityParam, ConsensusProtocol (..))
->
+
```
nfrisby, line 76: Same request

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -216,7 +217,7 @@ initLgrDB k chain = do
     blockMapping = Map.fromList
       [(blockRealPoint b, b) | b <- Chain.toOldestFirst chain]
 
-    cfg = configLedgerDb $ testCfg k
+    cfg = configLedgerDb (testCfg k) (ApplyOpts ValidationOn OmitLedgerEvents)
```
nfrisby, line 220: I wonder if it'd be worth factoring out a `defaultApplyOpts` and using that everywhere this triplet occurs?

nfrisby (commented):
I'm questioning some things that might lead to a much different diff.

ouroboros-consensus/changelog.d/20250214_122203_jasataco_sts.md
```diff
@@ -0,0 +1,22 @@
...
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Expose `ApplyOpts` when calling ledger rules. This allows users of our library to choose whether to enable or disable validation as well as emit or do not emit ledger events.
```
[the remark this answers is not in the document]
nfrisby: Just an update: there's an ongoing Slack convo about this question.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -274,17 +276,25 @@ untickedShelleyLedgerTipPoint ::
   -> Point (ShelleyBlock proto era)
 untickedShelleyLedgerTipPoint = shelleyTipToPoint . untickedShelleyLedgerTip
 
+validationMode :: Validation -> STS.ValidationPolicy
+validationMode ValidationOn  = STS.ValidateAll
+validationMode ValidationOff = STS.ValidateNone
```
lehins: When I say we need to be able to use any validation level, I really mean any validation level that is provided by STS. This approach of redefining new validation types hides an important validation level `ValidateSuchThat`.
I suggest passing through the validation level provided by STS instead creating some isomorphic type, this will also ensure it will not go out of sync if there is a change in the validation level in STS.

jasagredo: Please check the latest commit on this PR. That is from an outdated version.

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -323,22 +315,32 @@ validationErrorImpossible = cantBeError . runExcept
...
     blkNo :: BlockNo
     blkNo = blockNo blk
 
+    byronOpts =
+      case doValidation of
+        STS.ValidateAll -> CC.fromBlockValidationMode CC.BlockValidation
+        STS.ValidateNone -> CC.fromBlockValidationMode CC.NoBlockValidation
+        STS.ValidateSuchThat _ ->
+          -- TODO @alexey
+          CC.fromBlockValidationMode CC.BlockValidation
```
lehins, line 342: Yes, this is a totally sensible mapping
```suggestion
        STS.ValidateSuchThat _ -> CC.fromBlockValidationMode CC.BlockValidation
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -101,10 +130,30 @@ class ( IsLedger l
   -- validation checks.
   reapplyBlockLedgerResult ::
        HasCallStack
-    => LedgerCfg l
+    => ComputeLedgerEvents
+    -> LedgerCfg l
     -> blk
     -> Ticked l
     -> LedgerResult l l
+  default reapplyBlockLedgerResult ::
```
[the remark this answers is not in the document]
jasagredo: I implemented this approach of explicit calls to default functions

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -73,6 +81,26 @@ data family Validated x :: Type
   Apply block to ledger state
 -------------------------------------------------------------------------------}
 
+-- | This option will decide whether we enable the validation in the Ledger
+-- layer or not.
+data Validation =
```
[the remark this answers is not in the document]
jasagredo: Now I just use `STS.ValidationPolicy`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -97,6 +103,17 @@ pureLedgerResult a = LedgerResult {
 -- Types that inhabit this family will come from the Ledger code.
 type family LedgerCfg l :: Type
 
+-- | Whether we tell the ledger layer to compute ledger events
+--
+-- At the moment events are not emitted in any case, so they are not
```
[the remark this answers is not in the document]
jasagredo: I tried to clarify this haddock. Let me know if it looks good

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
```diff
@@ -157,6 +158,8 @@ completeChainDbArgs ::
      -- ^ Immutable FS, see 'NodeDatabasePaths'
   -> (RelativeMountPoint -> SomeHasFS m)
      -- ^ Volatile  FS, see 'NodeDatabasePaths'
+  -> ApplyOpts
```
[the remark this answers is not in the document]
jasagredo: Now these are defaulted, but can be overridden with `enableLedgerEvents`. This is passed to the LedgerDbCfg which is then passed to `applyBlock`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Update.hs
```diff
@@ -110,26 +110,27 @@ toRealPoint (Weaken ap)      = toRealPoint ap
...
-applyBlock cfg ap db = case ap of
+applyBlock opts cfg ap db = case ap of
     ReapplyVal b ->
       return $
-        tickThenReapply cfg b l
+        tickThenReapply (eventsOpts opts) cfg b l
     ApplyVal b ->
       either (throwLedgerError db (blockRealPoint b)) return $ runExcept $
-        tickThenApply cfg b l
+        tickThenApply opts cfg b l
```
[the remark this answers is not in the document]
jasagredo: This will now never be off, we only set now the ledger events emission

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -338,49 +360,56 @@ instance ShelleyCompatible proto era
...
-                , asoEvents     = STS.EPReturn
-                }
+instance ShelleyCompatible proto era
+      => ThrowLedgerReapplyError (LedgerState (ShelleyBlock proto era)) where
+  reapplyResult err = Exception.throw $! ShelleyReapplyException @era err
 
 data ShelleyReapplyException =
-  forall era. Show (SL.BlockTransitionError era)
-  => ShelleyReapplyException (SL.BlockTransitionError era)
+  forall era. Show (ShelleyLedgerError era)
```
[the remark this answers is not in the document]
jasagredo: I reverted it back, and actually removed ShelleyLedgerError as it was a newtype not used anywhere else

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Ledger/Ledger.hs
```diff
@@ -323,22 +314,27 @@ validationErrorImpossible = cantBeError . runExcept
...
-      CC.ABOBBoundary raw' -> applyABoundaryBlock        cfg raw'         blkNo ls
+      CC.ABOBBlock    raw' -> applyABlock byronOpts cfg raw' blkHash blkNo ls
+      CC.ABOBBoundary raw' -> applyABoundaryBlock   cfg raw'         blkNo ls
   where
     blkNo :: BlockNo
     blkNo = blockNo blk
 
+    byronOpts =
+      case validationOpts opts of
+        ValidationOn  -> CC.fromBlockValidationMode CC.BlockValidation
```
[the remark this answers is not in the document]
jasagredo: Done

ouroboros-consensus/changelog.d/20250214_122203_jasataco_sts.md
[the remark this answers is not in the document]
jasagredo: Consensus users will never not validate blocks anymore. Only low-level users will be able to set the validation policy

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
jasagredo: Validation options are gone

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
jasagredo: I undid these changes

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
jasagredo: Defaulted now

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/B.hs
jasagredo: Indeed, fixed

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
jasagredo: Defaulted now

ouroboros-consensus/src/unstable-tutorials/Ouroboros/Consensus/Tutorial/Simple.lhs
jasagredo: Emacs did it for me. I fixed it thanks

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
jasagredo: No longer applicable

jasagredo:
This PR is ready for another round of reviews. The final change is that validation options are exposed for the low-level `applyBlockLedgerResultWithValidation` and hardcoded for everything else. And for the events, those are defaulted to `OmitLedgerEvents` but eventually we want to expose the function `enableLedgerEvents` to be used. It is however useless to do it now as there is no handler for those events.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -83,9 +88,28 @@ class ( IsLedger l
   --
   -- This is passed the ledger state ticked to the slot of the given block, so
   -- 'applyChainTickLedgerResult' has already been called.
+  --
+  -- Users of this function can set any validation level allowed by the
+  -- @small-steps@ package. See "Control.State.Transition.Extended".
+  applyBlockLedgerResultWithValidation ::
```
nfrisby, line 94: In my previous comments, I was envisioning something that tightly mimics `applySTSOpts` https://github.com/IntersectMBO/cardano-ledger/blob/a6f276a9e3df45d69be3a593d3db6cd9dfaa7a02/libs/small-steps/src/Control/State/Transition/Extended.hs#L506-L513 --- so that eg ledger devs can have their same familiar interface

```
applyStsOptsBlock :: forall ep.
     ApplySTSOpts ep
  -> LedgerCfg l
  -> blk
  -> Ticked l
  -> WithEvents ep (AuxLedgerEvent l) (l, [LedgerErr l]))

type family WithEvents ep ev a where
  WithEvents EPReturn  a = (a, [ev])
  WithEvents EPDiscard a = a
```

where `WithEvents` replaces https://github.com/IntersectMBO/cardano-ledger/blob/a6f276a9e3df45d69be3a593d3db6cd9dfaa7a02/libs/small-steps/src/Control/State/Transition/Extended.hs#L278C1-L280C28

---

We could even preserve their exact same signature --- except still taking a config, a block, and a ticked ledger state instead of their `RuleContext` --- if we wanted to expose their `sts`/`s` parameter as an associated type of `blk` and `l`, I suppose?


```
type STS_ApplyBlock blk l :: Type

applySTSOpts_ApplyBlock :: forall ep s.
  (s ~ STS_ApplyBlock blk l) =>
  LedgerCfg l ->
  blk ->
  Ticked l ->
  ApplySTSOpts ep ->
  BaseM s (EventReturnType ep s (State s, [PredicateFailure s]))
```

---

I'm relatively sure that there's no redundancy in that set of inputs, but if there were, then we could change the `ApplySTSOpts ep` argument to be ``ApplySTSOpts ep -> ApplySTSOpts ep`, which the method definition would use to override the default `ApplySTSOpts ep` implied by the other arguments?

nfrisby (commented):
Just sharing one key high-level idea.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Abstract.hs
jasagredo: - So Alexey said explicitly that assertions should use `globalAssertionPolicy` always and nothing else.
- As for changing the definition, I think the definition is almost entirely stable in the coming future.
- For the _why_, I could elaborate on the haddock

dnadales (approved):

PR #1401, opened 2025-02-19
Draft HardForks.md

nfrisby:
A document discussing how the node (mostly Ledger and Consensus) handle hard forks.

This is part of my going effort to justify simplifying the HFC and might also be useful for deriving some of the Cardano Blueprint.


nfrisby:
Rendered: https://github.com/IntersectMBO/ouroboros-consensus/blob/nfrisby/doc-triggers/docs/website/contents/for-developers/HardForks.md

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+Every hard fork is either an era transition or an _intra-era hard fork_ [^accidentals].
+Intra-era hard forks only introduce changes with a small effect on the code, eg a few easy-to-maintain conditionals.
+Some have been bugfixes that were small by chance, and others have been changes that were only small because they had been anticipated and intentionally deferred when the era was originally designed.
+
+Eras are a straight-forward mechanism that enables node developers to separate the concerns arising from features that are added to Cardano over time.
+The separation amongst eras inherently simplifies the development of the new features, and it also helps ensure that the Cardano node can sync the entire historical chain [^mithril] --- as well as refuse invalid alternative histories.
+
+### Which changes require a new era?
+
+It's ultimately the judgment of the node developers whether some changes should be achieved with an era transition rather than an intra-era hard fork.
```
nfrisby, line 40: @lehins I'd appreciate your feedback/contributions on the entire `Era` section, but especially within this `Which changes require a new era?` subsection.

nfrisby: I had a sentence about "structural changes" specifically, but I seemed to have lost it 🤷

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+
+There are other Cardano networks besides mainnet, notably the testnets.
+By their essential nature, though, testnets should also use mainnet's mapping between each major protocol version number and the corresponding pair of protocol rules and ledger rules.
+On the other hand, these other networks must be able to hard fork at different times than mainnet does --- the testnets should have hard forked successfully before mainnet hard forks!
+Since the testnet governance and mainnet governance are independent, the determination of the hard fork timing via on-chain governance immediately enables it to vary on different networks' chains.
+
+## Eras
+
+A hard fork is the result of the nodes' developers introducing backwards-incompatible changes [^change-decisions] and the community ultimately choosing to adopt them by increasing the major protocol version parameter via on-chain governance [^frame-rule].
```
lehins, line 21: It is important to note that hard forks happen atomically across the world. That is the whole reason why we need them. SPOs voting for them acts only acts as an indicator that majority of nodes have upgraded their software and are in agreement to do the protocol version bump, while the actual act of bumping the protocol version, i.e. the act of hard forking, happens at one particular instant for everyone.
```suggestion
A hard fork is the result of the nodes' developers introducing backwards-incompatible changes [^change-decisions] and the community ultimately voting via on-chain governance [^frame-rule] to indicate the majority readiness to adopt those changes at the same time by atomically increasing the major protocol version parameter for everyone at once.
```

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+When those developers design and implement such changes, they determine whether the changes are easy to maintain as variations within the existing code.
+If so, then the hard fork is ultimately toggling some conditionals within the code.
+If the changes are not so simple, the Cardano developers instead introduce a new _era_ within the code [^roadmap-disambiguation].
```
lehins, line 24: It is true, that most of the time it is complexity of the feature that drives the new era, but it is not a requirement. It is not necessarily the simplicity that drives this decision of including a change in an intra-era hard fork. The change could be as simple as prevent duplicates in a Set when it is being deserialized. But because of current limitations of our implementation we do require such change be introduced on a new era. This is a concrete example that we are planning for the next era that we can't introduce with an intra-era hard fork.
Another concrete example was Allegra era. The only change it had was introduction of a Timelock (two extra cases to a native script and a validity interval). It was a fairly simple change, but it required a new era.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+Several eras coexist sequentially on the Cardano chain; a hard fork that takes one step within that sequence is called an _era transition_.
+
+Every hard fork is either an era transition or an _intra-era hard fork_ [^accidentals].
+Intra-era hard forks only introduce changes with a small effect on the code, eg a few easy-to-maintain conditionals.
+Some have been bugfixes that were small by chance, and others have been changes that were only small because they had been anticipated and intentionally deferred when the era was originally designed.
+
+Eras are a straight-forward mechanism that enables node developers to separate the concerns arising from features that are added to Cardano over time.
+The separation amongst eras inherently simplifies the development of the new features, and it also helps ensure that the Cardano node can sync the entire historical chain [^mithril] --- as well as refuse invalid alternative histories.
+
+### Which changes require a new era?
```
lehins, line 38: Another reason for a new era is that sometimes it happens that new features that are being introduced require an initial value, which needs to be populated upon a hard fork. For example initial Constitution and the guardrails script in Conway. Such initial values are supplied through a Genesis file.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+
+  However, the era transitions themselves should generally be considered as "just a big hard fork".
+  The foundational data is instead the protocol version, especially because not all hard forks are era transitions.
+  (See [this GitHub comment](https://github.com/IntersectMBO/ouroboros-consensus/issues/416#issuecomment-2669347315) for the details of how this lesson was learned.)
+
+- Some codecs reflect Cardano's specific era structure.
+  That's  not strictly necessary, but a tagged union approach has made it trivial to combine the era-specific codecs into a full Cardano codec.
+  Specifically, top-level codecs fundamentally branch on a simple number located near the beginning of the bytestring and dispatch accordingly to the corresponding era's codecs.
+
+- (TODO Confirm whether these tags dont affect any hashes. If they do, to what extent does it actually matter?)
```
lehins, line 70: I am not aware of any tags. As such no hashes are affected.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+[^frame-rule]: *Clarification*.
+  There's no other technical reason for the community to increment the major protocol version.
+
+[^roadmap-disambiguation]: *Disambiguation*.
+  The [Cardano Roadmap](https://roadmap.cardano.org/en/) also defines some very broad "eras": Byron, Shelley, Goguen, Basho, and Voltaire.
+  These are more general than the eras that divide the actual evolution of the protocol rules and ledger rules within the node.
+  [CIP-0059](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0059/feature-table.md) maintains a table showing the correspondence; it lists roadmap eras in the "Phase" column.
+
+[^accidentals]: *Clarification*.
+  Accidental hard forks would necessarily be intra-era hard forks.
```
lehins, line 95: What is accidental hard fork? Are you talking about an uncontrolled hard fork due to divergence in validity of a block?
If that is the case then it can't be considered as an intra-era hard fork either, IMHO, because no protocol version update would happen. 

However, if you are talking about _accidental_ fork as it is described on wikipedia, then it is not even classified as a *hard* fork, it is part of the normal process of chain selection, which has a potential to span over era boundaries, as far as I understand it.

lehins (commented):
It is just my personal opinion, but the sense I felt from reading this document is that it downplays the role of HFC and overall importance of separation of logic by eras. In my mind it is an incredible achievement of the Cardano core team coming up with this type safe design that logically separates major upgrades.
Could have it been done differently? Of course! From what I understand, however, this document attempts to describe current state of affairs, as such it should emphasize the importance of separation by era, not the opposite.

docs/website/contents/for-developers/HardForks.md
nfrisby: Ha! Proof by "I've never heard of that" :D Excellent.

docs/website/contents/for-developers/HardForks.md
nfrisby: Hmm. Yeah, I was taking the "intra-era" part too literally. It's a good point that what we call an _intra-era hard fork_ necessarily involves a major protocol version bump, and that wouldn't be accidental.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
+# Hard Forks in Cardano
+
+## Introduction
+
+This document details how the Cardano node handles a [hard fork](https://en.wikipedia.org/wiki/Fork_(blockchain)#Hard_fork).
+
+Cardano mainnet is intended to hard fork when and only when the major component of the [`protocol version` protocol parameter](https://github.com/IntersectMBO/cardano-ledger/blob/2ff5fa4e8c6b773748799981ef44ef6a62cd5c92/libs/cardano-ledger-core/src/Cardano/Ledger/Core/PParams.hs#L350) changes.
```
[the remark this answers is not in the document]
nfrisby: CIP-1694 says "While they need to be treated specially by SPOs, hard forks are not differentiated from other protocol parameter changes."

That's the only difference between `HardForkInitiation` and `UpdateParameters`, right? That `HardForkInitiation` explicitly signals that SPOs need to be paying attention?

> `HardForkInitiation` is also used for updating minor protocol version, albeit it would not be considered a hard fork.

😬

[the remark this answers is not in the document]
nfrisby: Nevermind... I'm already seeing some additional special cases for `HardForkInitiation` within CIP-1694.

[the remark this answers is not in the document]
lehins: Yeah, SPOs for instance do not even vote on some of the protocol parameter updates and for those that they do vote on, there is a mechanism to override the default vote. Which is not the case with HardForkInitiation. Moreover they have different settings on the required thresholds. So, yeah totally different governance actions

docs/website/contents/for-developers/HardForks.md
nfrisby: > Moreover, in my personal opinion, this limitation of requiring a hard fork for serialization changes is also a feature. Because serialization changes have a very serious knock-on effect. All of the downstream tooling and hardware wallets are affected by serialization changes. For example we had to freeze all CDDL changes about 6 months before the Conway release, just so hardware wallets were ready for the hard fork.

Did you mean to write "requiring an era for serialization changes"? Otherwise, I'm not seeing how this explicitly relates to the question of whether some hard fork needs to be an era transition rather than "just" an intra-era hard fork.

> changes to a transaction format require changes to plutus context, thus requiring a new plutus ledger version.

I didn't follow these statements about Plutus. But I'm pretty naive about it. Are you ultimately referring to these data structures?

- https://plutus.cardano.intersectmbo.org/haddock/latest/plutus-ledger-api/src/PlutusLedgerApi.V1.Contexts.html#TxInfo
- https://plutus.cardano.intersectmbo.org/haddock/latest/plutus-ledger-api/src/PlutusLedgerApi.V2.Contexts.html#TxInfo
- https://plutus.cardano.intersectmbo.org/haddock/latest/plutus-ledger-api/src/PlutusLedgerApi.V3.Contexts.html#TxInfo

I see how some eras have affected them (eg Conway obviously shows up in V3), but I'm not seeing how the CDDL details of a tx would affect these types, eg.

I chased from `PlutusWithContext` https://github.com/IntersectMBO/cardano-ledger/blob/9f98a9e15d8220ccc4940935efdce8dc69a63c22/libs/cardano-ledger-core/src/Cardano/Ledger/Plutus/Evaluate.hs#L94-L117 and those `TxInfo`s seemed like the only ultimate data that was relevant. Maybe I missed a problematic hash or something?

Maybe the `CostModel? Its Haddock does explicitly mention hard fork boundaries.

lehins: > Did you mean to write "requiring an era for serialization changes"?

Yes. Sorry, that is exactly what I meant. 

> Are you ultimately referring to these data structures?

Correct

> I'm not seeing how the CDDL details of a tx would affect these types, eg.

It is an indirect relationship that stems from change to Tx, rather than from CDDL. Change to Tx requires CDDL changes and it requires similar change to Plutus Context, thus new Plutus version. There can also be direct relationship, for example we are planning on making inputs an ordered Set, which will affect both CDDL and Plutus context in interesting ways.

Cost Model nowadays (starting with Conway) can be updated at any point in time, so that is not really related to any future hard fork anymore
See this file, this is where era + plutus context mapping happens for Conway (there is similar functionality for previous eras): https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/src/Cardano/Ledger/Conway/TxInfo.hs

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,98 @@
...
+
+- The familiar and intuitive notion of era is useful for organization in general, and so the eras' names might appear within parameter names in configuration files, specification documents, announcements, etc.
+
+  However, the era transitions themselves should generally be considered as "just a big hard fork".
+  The foundational data is instead the protocol version, especially because not all hard forks are era transitions.
+  (See [this GitHub comment](https://github.com/IntersectMBO/ouroboros-consensus/issues/416#issuecomment-2669347315) for the details of how this lesson was learned.)
+
+- Some codecs reflect Cardano's specific era structure.
+  That's  not strictly necessary, but a tagged union approach has made it trivial to combine the era-specific codecs into a full Cardano codec.
+  Specifically, top-level codecs fundamentally branch on a simple number located near the beginning of the bytestring and dispatch accordingly to the corresponding era's codecs.
```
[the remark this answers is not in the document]
nfrisby: > If you are talking about deserializers here then there is no simple number that is added today to the beginning of any block or a transaction that does this sort of branching, nor should it.

That number is present when the block is on the network or stored on disk, etc. We call it "the HFC envelope". It is, by design, transparent to the ledger (the ledger rules generally don't need to consider other eras, except for era-translation functions eg), which is hopefully the only reason this text sticks out to you?

> Protocol version in the ledger state decides the era 

In today's node, the HFC ultimately determines what era some received block/header/transaction _claims_ to be in (according to its HFC envelope). The ledger state determines whether that claim is correct :)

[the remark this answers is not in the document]
nfrisby: I should elaborate on that last quip.

- The HFC tracks which era each ledger state is in. When Consensus ticks a ledger state across an epoch boundary that increments the major protocol version, then the ticked ledger state is now in the next era. (Modulo that block counting wart we've discussed before.)

- The HFC errors out when applying a block that claims to be in some era to a ledger state that is in an older era; the ledger rules aren't even invoked in this case: https://github.com/IntersectMBO/ouroboros-consensus/blob/b14da83d982224edc1700b0bed31ebd96c9ceaf8/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs#L77-L82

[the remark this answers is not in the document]
nfrisby: > Protocol version in the ledger state decides the era and deserializers can use this information to do the branching.

That's the hypothetical we had discussed in the Slack thread: that the Consensus layer might be able to provide the protocol version to deserializers. Currently it does not; instead, the HFC envelope tells it which era's parser to use.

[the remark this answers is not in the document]
nfrisby: > while the actual act of bumping the protocol version, i.e. the act of hard forking, happens at one particular instant for everyone

^^^ that quote is from a different comment thread

> Protocol version in the ledger state decides the era and deserializers can use this information to do the branching.

I fully acknowledge that I can't read your mind, but these two quotes together seem to indicate that you're envisioning that the node parses incoming blocks according to the protocol version of the node's most recent ledger state.

We can't do that, since arriving blocks might not extend the node's currently selection chain. In particular, if an arriving block extends an earlier prefix of the node's selection, then the block to be parsed might be in the "previous" era.

nfrisby:
I just pushed up a rewrite, added for now as a second file.

There's probably some fat to trim still and maybe another dimension of the design to better emphasize, but I think it has a useful organization and flow. [ProtocolVersioning.md](https://github.com/IntersectMBO/ouroboros-consensus/blob/nfrisby/doc-triggers/docs/website/contents/for-developers/ProtocolVersioning.md)

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
+# Introduction
```
jasagredo, line 1: Just a small detail, the outline of this MD document seems wrong, everything is a subsection of the introduction 
image

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
...
+  Tiebreakers are a gray area, since Praos does not inherently rely on them.
+- *Chain Data Model*, the parts of the Message Data Rung used to define the Chain Validity Predicates and the Chain Order.
+    - A change to the data model of the node's maintained state --- as opposed to its messages --- would not inherently trigger this.
+    - However, if such a change were meaningful (ie not just a refactor/optimization/etc), then some Chain Validity Predicates would necessarily also change.
+- *Chain Codecs*, the parts of the mapping from the CBOR Rung to the Chain Data Model that affect the hashes therein.
+  Two nodes won't even be able to exchange transactions, block headers, and/or blocks if they disagree on their identifying hashes.
+
+The most common change in Cardano's protocol so far has been the addition of a new kind/variant of transaction, which triggers all of these except the Chain Order.
+The Chain Order has only changed from Praos's length to Ouroboros Genesis's length-soon-after-the-intersection, which is slight.
```
jasagredo, line 99:
```suggestion
The Chain Order has only changed from Praos's length to Ouroboros Genesis's density-after-the-intersection, which is slight.
```

I would not relate Genesis and length artificially. Genesis is about density.

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
...
+Or the mini protocols --- excluding their payload --- could change.
+All such changes are conditional on the per-connection version negotiated by the Handshake mini protocol instead of the protocol version.
+
+## Ledger Eras
+
+Any Cardano node implemented in a strongly-typed language is likely going to declare distinct types for Chain Data Models that differ between protocol versions; more precise types yield stronger assurances from the typechecker.
+For the sake of such implementations, the ledger introduces a new era whenever the Chain Data Model changes.
+The [sequence of eras](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0059/feature-table.md) at the time of writing are: Byron [^byron-halves], Shelley, Allegra, Mary, Alonzo, Babbage, and Conway.
+
+A Cardano node implementation or other tooling implemented in a language prone to dynamic typing might find it convenient to ignore the era tags for the most part.
```
jasagredo, line 114: I appreciate the distinction between strongly-typed nodes and other nodes. Perhaps this could end up in the Blueprint

jasagredo:
In general the document looks good, and I think there explanation is quite clear and I appreciate the fact that no code is actually depicted here.

docs/website/contents/for-developers/ProtocolVersioning.md
nfrisby: I'm hung up on this a little, since `length-soon-after-the-intersection` _is_ density, assuming "soon" is a fixed value.

nfrisby:
- As of writing this latest draft, I started seriously considering changing the HFC envelope to contain the protocol version instead of the era tag.
- I eventually realized that the `cardano-cli $era build` command (et al) would pose a challenge: it would have to choose a protocol version to include in the new tx, but the user only specified an era. For today's era's, it'd be OK to use the lowest protocol version of that era, but --- with protocol versions in the CBOR --- it would be possible for intra-era hard forks to affect the codecs, and so this "use era's lowest protocol version" heuristic would no longer be sound.
- After some study and thought these for a few days, I think the answer is: TxSubmission and LocalTxSubmission should be exchanging transactions that don't have any tags. The node's TxSub and/or Mempool logic would introduce the tag based on its current ledger state when the untagged tx was received.
- The only corner case I can see where this would degrade behavior is (I'm describing this with era tags, but it's the same with protocol version tags):
       - The node's selection advances from era X to era Y.
       - An untagged transaction arrives that is valid in both eras; the node tags it with Y, its current selection's era.
       - The node switches to a (denser) selection that is still in era X.
       - This causes the Mempool to discard validated transactions tagged with a "future era", including Y.
   If the user could choose the tx's era tag (as they do with today's `cardano-cli`), then they could declare X, and so the tx could have survived the node's brief reversion to era X.

docs/website/contents/for-developers/ProtocolVersioning.md
nfrisby: Great to hear! The blueprint is certainly in the back of my mind as I'm writing this document.

lehins:
@nfrisby I don't fully understand why transactions would include any such tag?
As far as ledger is concerned a transaction either deserializes and validates, or it doesn't. So, there should be no surprise to a user. Also, more often than not transactions from previous protocol version will be accepted just fine for the next protocol version, whether it was an intra-era or cross-era hard fork. That is why today translation of transactions in the mempool works when we transition from one era to another.

This backwards compatibility is a nice property, because users could submit transactions and not worry about era transitions and HFCs, therefore it is unclear to me why during transaction submission today there is information about the era that the transaction was built for.

We, in ledger, when we change transaction serialization, we try to do it in the most graceful matter possible for any one era transition, in order to make the transition to the new era smooth. The only reason why we would ever want to change serialization during an intra era hard fork would be to patch some bug, which would have no affect on normal legitimate transactions.

If there is a desire from users to specify validity for protocol version, maybe we could provide ability for a user to supply an optional range of protocol versions that the transaction should be valid for. Considering such envelope information today wouldn't be signed it would be more of UX improvement/restriction, rather than some sort of guarantee, unless in the future era we make this protocol version range part of the TxBody, similarly to validity interval.

How does this scenario work today?
> - The node's selection advances from era X to era Y.
> - An untagged transaction arrives that is valid in both eras; the node tags it with Y, its current selection's era.
> - The node switches to a (denser) selection that is still in era X.
> - This causes the Mempool to discard validated transactions tagged with a "future era", including Y.

Cause I'd assume when a node advances from era X to Y all transactions in the mempool are upgraded to that new era Y? If there is a selection of another chain that is still in era X then transactions in the mempool cannot be downgraded to the era X from era Y, at least there is no such mechanism provided by the Ledger. Is the whole mempool reconstructed by deserializing all transactions in era X from their binary representation or how does it work?

jasagredo:
@lehins 

> As far as ledger is concerned a transaction either deserializes and validates, or it doesn't. 

Does this mean that transactions are also forward compatible in serialization? i.e. a transaction created in Shelley can be deserialized just fine as a Conway transaction? is there no special logic used when upgrading Transactions to later eras?

Or maybe I'm misunderstanding your comment here.

lehins:
> Does this mean that transactions are also forward compatible in serialization? i.e. a transaction created in Shelley can be deserialized just fine as a Conway transaction?

Yes, most of them will. There were some bug fixes and some aspects of a transaction since Shelley have been deprecated (like MIR Certs, etc), but majority of transactions will deserialize in Conway.

This is not always going to be true. The goal is to keep as much backwards compatibility from one era to the next one that follows, but there is no real need to support all eras all the way down to Shelley.

> is there no special logic used when upgrading Transactions to later eras?

Nope. Logic is take the bytes and deserialize that transaction in an era that you'd like to validate it in.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,134 @@
...
+It was the first era transition we experienced, and no subsequent transition changes nearly so much.
+The two code bases are drastically different, so it's hard to argue this should have been an intra-era hard fork.
+In particular, this is the only era transition that changed the duration of slots and the number of slots in an epoch.
+In the current codebase, any hard fork that changes those per-era constants must be an era transition, since the Consensus infrastructure was explicitly designed with that restriction as a simplifying assumption.
+A different design might relax that constraint, but it has not been restrictive so far: a second change to these constants has been discussed only rarely and never actually proposed, and deferring them until whatever justifies the next era transition wouldn't obviously be prohibitively disruptive.
+
+### How wide-reaching are eras?
+
+Eras are ultimately an implementation detail of the node, a consequence of architectural trade-offs.
+The node's behaviors currently reflects the eras in only two ways.
```
dnadales, line 78: I would expect these two ways to be mentioned after this sentence :)

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,134 @@
...
+
+Specifically, the node must be able to forecast the data necessary to validate a useful prefix of some header chain from the intersection of that header chain and the node's current selection, including whether each header arrived too early.
+
+- The familiar and intuitive notion of era is useful for organization in general, and so the eras' names might appear within parameter names in configuration files, specification documents, announcements, etc.
```
dnadales, line 82: This bullet seems a bit disconnected from the preceding paragraph.

docs/website/contents/for-developers/HardForks.md
```diff
@@ -0,0 +1,134 @@
...
+
+  *Remark*.
+  These eras tags are not present in any codecs for bytes that determine hashes used on chain.
+  For example, suppose the only difference between some transaction's serialization in two different eras is the numeric value of the era tag.
+  In that case --- unless the eras use different hashing functions --- both serializations would be assigned the same transaction ID, since the function that assigns transaction IDs is applied to the bytes inside of the envelope that carries the era tag.
+
+## Forecasting, forewarning
+
+Hard forks can affect the validity of block headers, and so must be sufficiently forecastable.
+For this reason, hard forks must not happen less than 36 hr (on mainnet) after the voting deadline, ie hard forks must have at least 36 hrs of forewarning --- just like other protocol parameter updates that affect headers (eg maximum block size).
```
dnadales, line 100: Maybe clarify that 36 hours correspond to 3k?

dnadales, line 100:
```suggestion
For this reason, hard forks must not happen until 36 hours (on mainnet) after the voting deadline has passed, ie hard forks must have at least 36 hours of forewarning --- just like other protocol parameter updates that affect headers (eg maximum block size).
```

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
+# Introduction
+
+The Cardano protocol will change over time.
+The community innovates applications and updates feature priorities, researchers refine Ouroboros, and developers improve the code.
+
+This document describes how Cardano accomodates these changes as a community, as a software engineering project, and as a distributed system.
+
+## The Protocol Version
+
+Each change to the Cardano protocol advances through the following steps before influencing the network's chain.
```
dnadales, line 10: I'm not sure "influencing" is the right term.

```suggestion
Each change to the Cardano protocol advances through the following steps before impacting the network's blockchain.
```

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Someone (you!) specifies a change that is backwards-incompatible but worthwhile.
+  For example, it could be a [Cardano Improvement Proposal](https://cips.cardano.org/) (CIP).
+- The Technical Steering Committee and Ouroboros researchers address the worth, safety, and future consequences of the proposed solution.
+- Cardano developers prepare a software release that includes the changes but only enables them for chains on which the protocol version has been incremented.
+- The community uses on-chain governance to increment the protocol version, thereby putting the change into effect.
+  (Stake pool operators must not vote to increment the protocol version before updating their software!)
```
dnadales, line 17: Is it sufficiently clear for the reader which protocol version increment we're referring to? I guess this refers to the protocol version included in the headers, which denotes the maximum protocol version our node supports.

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+Cardano philosophy mandates that the community has the final say regarding if and when to adopt some protocol change.
+They accordingly control the final step above: developers determine the exact meaning of each protocol version, but only the community increments the protocol version on mainnet.
+
+It is crucial that the community does not increment the protocol version before enough nodes have upgraded their software.
+Each protocol version increment corresponds to a _hard fork_, because the changes are backwards-incompatible: nodes running the old code will not be able to adopt blocks minted by the new code.
+Even if some block content happened to be valid in both the old and the new protocol, each software release explicitly refuses any block that (extends a ledger state that) has a protocol version greater than that release was intended to implement — see the MaxMajorPV constant.
```
dnadales, line 24: Is it worth adding a link to the location where `MaxMajorPV` is defined?

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+Even software released well after mainnet increments the protocol version must implement the protocol both with and without that increment's changes.
+The Cardano node must correctly process blocks with older protocol versions, eg when catching up to mainnet for the first time or after being offline for a long time.
+So every protocol change must remain conditional within all future releases, instead of only within the particular release that mainnet used for that increment of the protocol version.
```
dnadales, line 28: What does it mean for a protocol change to be conditional on a (future) release?

coot:
In the future, it would be nice to expand this document and describe the relation between protocol version, `BlockNodeToNodeVersion` and `NodeToNodeVersion`.

docs/website/contents/for-developers/ProtocolVersioning.md
```diff
@@ -0,0 +1,155 @@
+# Introduction
+
+The Cardano protocol will change over time.
```
nfrisby, line 3: GitHub does not allow threading of comments on the PR itself. But we've got a pretty big conversation happening there right now, so I'm going to arbitrarily choose this line in the diff in order to create a thread for that conversation.

---

@lehins wrote:

> We, in ledger, when we change transaction serialization, we try to do it in the most graceful matter possible for any one era transition, in order to make the transition to the new era smooth. The only reason why we would ever want to change serialization during an intra era hard fork would be to patch some bug, which would have no affect on normal legitimate transactions.

and

> This is not always going to be true. The goal is to keep as much backwards compatibility from one era to the next one that follows, but there is no real need to support all eras all the way down to Shelley.

Based on those two quotes, I'm thinking the tacit rule is that: Cardano users should only ever consider the latest protocol version, and maybe the two latest protocol versions when near a hard fork event.

And Alexey is saying that the ledger team tries, for each hard fork event, to minimize the disagreement between those two protocol versions regarding what some bytes might mean as a transaction.

This is a very helpful insight for me; I'll ponder it. I could very much see it simplifying my intuition about what's important here. On the other hand, there is some appeal in the user explicit declaring their intent through a version, as Alexey elaborated here:

> If there is a desire from users to specify validity for protocol version, maybe we could provide ability for a user to supply an optional range of protocol versions that the transaction should be valid for. Considering such envelope information today wouldn't be signed it would be more of UX improvement/restriction, rather than some sort of guarantee, unless in the future era we make this protocol version range part of the TxBody, similarly to validity interval.

---

Ultimately, the Mempool needs to know the byte size/ExUnits size/ref script size/etc of a tx before validating that tx. And as of UTxO HD it'll need to identify the set of TxIn before validating the tx. Etc. Parsing the tx to project out that data currently requires choosing which era's codec to use.

That's the only reason why "a tx is just bytes" doesn't trivially work for the Mempool with the HFC.

However, maybe that idea remains workable if we store txs "as bytes" even when they're already in the Mempool. That is, whenever the Mempool does anything with a tx (eg measure its sizes, revalidate it, etc) it applies the codec for the era (/protocol version) of the node's current selection to the tx's bytes --- forgetting any parsing it might have done before (unless the era is the same as last time, eg).

lehins: > However, maybe that idea remains workable if we store txs "as bytes" even when they're already in the Mempool. 

This is already possible. You can always get the original bytes of any transaction with this [`originalBytes`](https://github.com/IntersectMBO/cardano-ledger/blob/81637a1c2250225fef47399dd56f80d87384df32/libs/cardano-ledger-core/src/Cardano/Ledger/Hashes.hs#L372) function if you ever need to re-deserialize it another era.

lehins:
@jasagredo It is much simpler than you think. When you look at CDDL of a transaction, there is no indication of which era it was built for. Sometimes it can be inferred, eg. if a transaction contains votes than it surely was build for Conway.

So, the way translation works in ledger is we take the bytes of the transaction and try to deserialize it in the new era. In other words, there is no "translation" happens. We simply deserialize the submitted bytes in the era that it needs to be validated in and we retain those bytes. If that transaction is incompatible with this era you will get deserialization error.

The overlap of compatibility of transactions for all eras can be inferred form the CDDL specification. If two different eras have the same specification for the parts of the transaction, then it will deserialize for those two eras. Whether it validate or not that is a different story, because there are plenty of things that can affect validity of a transaction, eg. if cost models protocol parameter were updated then it would effectively invalidate all transactions that use older cost model for the plutus version that was updated.

PR #1407, opened 2025-02-28
Rework docs using haddock-project

jasagredo:
# Description

this PR might resolve.

Also note that:


jasagredo:
`haddock-project` is broken for type families. We are abandoning this approach for now.

PR #1409, opened 2025-03-03
Validation interface

lehins:
# Description

This PR contains integration of latest ledger master which was required for #1390 that is also included in this PR

Also note that:


lehins:
Close in favor of #1390

PR #1415, opened 2025-03-07
Do not delete the Ledger snapshot if it looks like a legacy snapshot

jasagredo:
# Description

this PR might resolve.

Also note that:


jasagredo:
Closed in favor of a more principle approach described in #1416

PR #1417, opened 2025-03-07
CSJ: prefer the Objector when replacing a Dynamo 

nfrisby:
See the various commits. The `CSJ: prefer the Objector when replacing a Dynamo` commit in particular explains the key change within this PR.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
@@ -158,7 +158,7 @@
 --
 -- If the objector disconnects or is disengaged, and there are FI* jumpers, then
 -- the one with the oldest intersection with the dynamo gets elected (i).
--- Otherwise, we are left with no dynamo.
+-- Otherwise, we are left with no objector.
 --
 -- If the dynamo rolls back to a point older than the last jump it requested, it
 -- is disengaged (j), a new dynamo is elected (e|f), and all the other peers
@@ -166,16 +166,15 @@
 --
 -- If the objector agrees with the dynamo, it is disengaged (k). If there are
 -- FI* jumpers, then one of them gets elected as the new objector (i).
--- Otherwise, we are left with no dynamo.
+-- Otherwise, we are left with no objector.
 --
 -- If the dynamo or the objector claim to have no more headers, they are
 -- disengaged (j|k), triggering the same chain of effect as described in the two
 -- previous points.
 --
--- The BlockFetch logic can ask to change the dynamo if it is not serving blocks
--- fast enough. If there are other non-disengaged peers, the dynamo (and the
--- objector if there is one, and all the other peers) is demoted to a happy
--- jumper (l+g+h) and a new dynamo is elected (e).
+-- The BlockFetch logic can ask to change the dynamo if it is not serving
+-- blocks fast enough. If there are other non-disengaged peers, all peers are
+-- demoted to happy jumpers (l+g+h) and a new dynamo is elected (e).
 --
 module Ouroboros.Consensus.MiniProtocol.ChainSync.Client.Jumping (
     Context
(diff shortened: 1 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -837,15 +865,21 @@ electNewDynamo ::
     LedgerSupportsProtocol blk
   ) =>
   Context m peer blk ->
-  STM m (Maybe (peer, ChainSyncClientHandle m blk))
+  STM m (TraceCsjReason -> TraceEventCsj peer blk, Maybe (peer, ChainSyncClientHandle m blk))
 electNewDynamo context = do
   peerStates <- cschcSeq (handlesCol context)
-  mDynamo <- findNonDisengaged peerStates
+  mDynamo <- do
+    -- prefer a 'Started' 'Objector', if any exists
```
nfrisby, line 872: Note that `rotateDynamo` does **not** call `electNewDynamo`.

This is desirable, since the Objector isn't responsible for Dynamo rotations.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -886,13 +892,30 @@ promoteToDynamo ::
...
-  let dynamoInitState = maybe DynamoStarted DynamoStarting mJumpInfo
-  writeTVar (cschJumping dynamo) $
-    Dynamo dynamoInitState $ pointSlot $ AF.headPoint fragment
+  jumping' <- readTVar (cschJumping dynamo) >>= \case
+    -- An 'Objector' that already 'Started' need not be disrupted.
+    --
+    -- TODO Intuitively, a 'Starting' 'Objector' also need not be disrupted,
+    -- but disrupting it wouldn't waste any @MsgRollForward@s. More concretely,
+    -- it's not obvious to me how to build a 'DynamoStarting' from a
+    -- 'Starting'.
```
amesgen, line 902: I agree that there is no need to have special logic for this case, and I think it is fine to "downgrade" this into a NOTE.

amesgen (approved):
LGTM, only very minor comments

Also great tracing/pretty-printing improvements, will be nice for any future test deep dives :clap:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
```diff
@@ -395,36 +411,39 @@ deriving anyclass instance
 -- requires writing to a TVar for every jumper.
 nextInstruction ::
   ( MonadSTM m ) =>
+  STM m retry ->
   PeerContext m peer blk ->
-  STM m (Instruction blk)
-nextInstruction context =
+  STM m (Strict.Either retry (Instruction blk))
```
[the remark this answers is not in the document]
nfrisby: I went with strict because I'm relying on `Void` being uninhabited. My intuition was to preclude the `NonStrict.Left _|_ :: NonStrict.Either Void a` junk term.

Does that seem excessive?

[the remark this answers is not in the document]
amesgen: Ah, that seems good then :+1: In particular, code like
```haskell
foo :: Strict.Either Void a -> Stuff
foo (Right a) = process a
```
do not raise pattern match warnings in contrast to lazy `Either`; which you currently don't do (instead, `Strict.either absurd` is used), but it still seems neat.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/Jumping.hs
nfrisby: I renamed `electNewDynamo` to `backfillDynamo` in order to emphasize this.

nfrisby:
I didn't realize the older commits weren't signed. So I did a rebase that merely signed them --- I'm expecting auto-merge to work now.

PR #1421, opened 2025-03-12
db-analyser: add DumpStakeDistributions pass

nfrisby:
A new pass in `db-analyser` and a bash/awk (😞) script for post-processing its output.


nfrisby:
Blue line: This graph shows that the 551 pools that have been in the top 90% of every epoch from 506 to 543 (ie 190 days) have controlled more than 88% of stake in every epoch.

Red line: Even if all of those pools' individual least-stake epochs were instead simultaneous, that lower bound is 81.77% cumulative stake. I think that suggests the stake is quite stable within these 551 pools over the past 190 days.

z-browser-22207339158-8616259868624-fc19dc45ff36764247ab5cf7a6b6aed7a96d13fd4e0fb1761b2cababbc32a8cf

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -271,7 +275,14 @@ instance (HasAnalysis blk, LedgerSupportsProtocol blk) => Show (TraceEvent blk)
     , "mutSnap " <> show mutSnap
     , "gcSnap " <> show gcSnap
     ]
-
+  show (DumpStakeDistribution eno pd) =
+      intercalate "\t"
+    $ (\ss -> show eno : show (SL.pdTotalActiveStake pd) : show (Map.size mp) : ss)
+    $ [ show (keyhash, SL.individualTotalPoolStake x, SL.individualPoolStake x)
+      | (keyhash, x) <- Map.assocs mp
```
amesgen, line 282: AFAIK `pdTotalActiveStake` and `individualTotalPoolStake` are only used for Conway voting stuff, so `individualTotalPoolStake / pdTotalActiveStake` does not necessarily equal `individualPoolStake`, which is used for leader election: https://github.com/IntersectMBO/ouroboros-consensus/blob/43c1fef4631d7a00879974e785c2291175a5f0dc/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L521 and the ledger peer info we give to Network https://github.com/IntersectMBO/ouroboros-consensus/blob/43c1fef4631d7a00879974e785c2291175a5f0dc/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/PeerSelection.hs#L48
So I think it makes sense to either ignore them here or make this explicit in the output somehow.

I looked at all epochs until 544, and so far, we apparently didn't have a difference here, no idea how likely it is for a difference to arise in the future.

See https://github.com/IntersectMBO/cardano-ledger/pull/4324#r1593847709, cc @lehins for confirmation.

scripts/genesis-stake-drift-analysis/README.md
amesgen, line 1: Idea for another thing to plot: the statistical distance/total variation distance of the stake distributions over time (potentially eliding the top 90%). This is what the researchers use in the [Praos paper](https://eprint.iacr.org/2017/573.pdf) (Definition 15, Theorem 11) for their notion of stake shift, as well as in their later paper "[Stake Shift in Major Cryptocurrencies: An Empirical Study](https://arxiv.org/pdf/2001.04187.pdf)".

Concretely, the stake shift $\sigma$ between two stake distributions measures the largest absolute difference between the combined stake of a fixed set of pools relative to the two distributions. So when we sample pools for Genesis with total stake $s$ from a somewhat old stake distribution from a snapshot, then these pools have at least $s-\sigma$ stake in today's stake distribution, where $\sigma$ is the stake shift between the old and today's stake distribution.

nfrisby: Great, thank you. I'll compute ${\displaystyle \Delta (X,Y)={\frac {1}{2}}\sum _{\alpha \in D}|\Pr[X=\alpha ]-\Pr[Y=\alpha ]|}$.  for the final stake distribution versus each individual past stake distribution.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
lehins: > AFAIK pdTotalActiveStake and individualTotalPoolStake are only used for Conway voting stuff

That statement is not quite correct. 

When `IndividualPoolStake` that is extracted from `PoolDistr` as it is done in consensus, will not have any voting related stake within it. However, the version of this in DRep pulser will have proposal deposits added to that stake. The issue was that the haddock for `individualTotalPoolStake` should not have stated anything about proposal deposits.

> I looked at all epochs until 544, and so far, we apparently didn't have a difference here, no idea how likely it is for a difference to arise in the future.

So, individualTotalPoolStake / pdTotalActiveStake should always equal individualPoolStake for leader election, unless you are looking into the version from pulser.

amesgen: Thanks, that is good to hear! I guess then the remaining mentions of proposal deposits [here](https://github.com/IntersectMBO/cardano-ledger/blob/81637a1c2250225fef47399dd56f80d87384df32/libs/cardano-ledger-core/src/Cardano/Ledger/State/PoolDistr.hs#L56) and [here](https://github.com/IntersectMBO/cardano-ledger/blob/81637a1c2250225fef47399dd56f80d87384df32/libs/cardano-ledger-core/src/Cardano/Ledger/State/PoolDistr.hs#L104) should also be removed?

scripts/genesis-stake-drift-analysis/README.md
nfrisby: ^^^ Done, as of my latest commit.

This is the result
image

Blue line is all pools that were in all epochs. Red line is pools that were in the top 90% of all epochs.

This is the cumulative stake of those two sets of pools per epoch: 
image

PR #1399, opened 2025-02-19
A more precise type for Nary.inject

nfrisby:
See the commit message of `consensus: a more precise type for Nary.inject`; the other commits are superficial.

PR #1435, opened 2025-03-26
More precise type for Index

jasagredo:


sop-extras/src/Data/SOP/Index.hs
```diff
@@ -18,6 +20,8 @@ module Data.SOP.Index (
   , indices
   , injectNS
   , injectNS'
+  , pattern IS
+  , pattern IZ
```
amesgen, line 23: [Bundle](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pattern_synonyms.html#import-and-export-of-pattern-synonyms) these?

sop-extras/src/Data/SOP/Index.hs
```diff
@@ -36,29 +39,42 @@ module Data.SOP.Index (
...
+
+pattern IS ::
+     ()
+  => xs ~ (x' ': xs')
+  => Index xs' x
+  -> Index xs x
+pattern IS z <- Index (S (Index -> z)) where
+  IS (Index z) = Index (S z)
+
+{-# COMPLETE IZ, IS #-}
```
amesgen, line 67: Hmm, while this approach is indeed makes the payload strict, GHC's pattern match coverage checker does not seem to detect that when using the pattern synonyms:
```haskell
test :: Index '[()] () -> ()
test IZ     = ()
test (IS a) = case a of {}
```
still requires the `IS` case, even though it is now definitely unreachable. :thinking: 

In contrast, keeping the previous definition of `Index` and making the `IS` payload strict should allow simplifying `test`.

Not a disaster, but something to be aware of.

amesgen (commented):
LGTM, main comment is that this change will not allow us to get rid of "superfluous" pattern matches.

Optional, as you mention https://github.com/IntersectMBO/ouroboros-consensus/pull/1399/files#r1961871210: You could simplify various definitions using SOP combinators now that `Index` is a newtype over `NS`. Candidates are both some of the definitions in `Data.SOP.Index` as well as the definition referred to in the review comment above.

sop-extras/src/Data/SOP/Index.hs
```diff
@@ -61,34 +59,31 @@ pattern IS ::
...
-    IZ     -> Dict
-    IS idx -> dictIndexAll p idx
+dictIndexAll :: forall c xs x. All c xs => Proxy c -> Index xs x -> Dict c x
+dictIndexAll p (Index idx) = hcollapse $ hcmap p (\Refl -> K Dict) idx
 
-injectNS :: forall f x xs. Index xs x -> f x -> NS f xs
-injectNS idx x = case idx of
-    IZ      -> Z x
-    IS idx' -> S (injectNS idx' x)
+injectNS :: forall f x xs. All Top xs => Index xs x -> f x -> NS f xs
```
amesgen, line 75: You can avoid the `All Top` constraint (same as `SList`) here and below by using `npToSListI` on `idx`, such that it doesn't need to be propagated.

sop-extras/src/Data/SOP/Index.hs
jasagredo: This is unfortunate. Do we want to proceed or shall we keep the previous definition and make it strict? I think it is fine this way because we can make use of `EmptyCase` to signal that match is inhabited.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Block.hs
```diff
@@ -205,7 +205,7 @@ distribAnnTip AnnTip{..} =
     distrib (WrapTipInfo info) =
         AnnTip annTipSlotNo annTipBlockNo info
 
-undistribAnnTip :: SListI xs => NS AnnTip xs -> AnnTip (HardForkBlock xs)
+undistribAnnTip :: forall xs. SListI xs => NS AnnTip xs -> AnnTip (HardForkBlock xs)
```
amesgen, line 208: Redundant change

sop-extras/src/Data/SOP/Index.hs
```diff
@@ -61,34 +59,31 @@ pattern IS ::
...
+     forall f a b x xs. All Top xs => (Coercible a (f x), Coercible b (NS f xs))
   => Proxy f -> Index xs x -> a -> b
 injectNS' _ idx = coerce . injectNS @f idx . coerce
 
-projectNP :: Index xs x -> NP f xs -> f x
-projectNP IZ       (x :* _ ) = x
-projectNP (IS idx) (_ :* xs) = projectNP idx xs
+projectNP :: All Top xs => Index xs x -> NP f xs -> f x
+projectNP (Index idx) np =
+  hcollapse $ hap (hmap (\f -> fn (\Refl -> K f)) np) idx
```
amesgen, line 85:
```suggestion
  hcollapse $ hliftA2 (\f Refl -> K f) np idx
```

amesgen (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Block.hs
jasagredo: It is needed for the `Index xs blk` in the `where` clause.

PR #1437, opened 2025-03-28
Expose `querySupportedVersions`

jasagredo:
This PR exposes a `querySupportedVersions` function that provides the list of versions in which a query is available, to be used by `cardano-api`

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
@@ -23,7 +23,6 @@ module Ouroboros.Consensus.Shelley.Ledger.Query (
   , NonMyopicMemberRewards (..)
   , StakeSnapshot (..)
   , StakeSnapshots (..)
-  , querySupportedVersion
     -- * Serialisation
   , decodeShelleyQuery
   , decodeShelleyResult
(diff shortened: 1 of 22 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -751,6 +752,10 @@ querySupportedVersion = \case
     v11 = ShelleyNodeToClientVersion11
     v12 = ShelleyNodeToClientVersion12
 
+querySupportedVersions :: BlockQuery (ShelleyBlock proto era) result -> [ShelleyNodeToClientVersion]
+querySupportedVersions q =
+  [ v | v <- [minBound..maxBound], queryIsSupportedOnVersion q v ]
```
amesgen, line 758: Does this function (and the other variant) cross the [fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -131,6 +134,17 @@ instance Show (SomeSecond BlockQuery blk) => Show (SomeSecond Query blk) where
   show (SomeSecond GetChainPoint)             = "Query GetChainPoint"
   show (SomeSecond GetLedgerConfig)           = "Query GetLedgerConfig"
 
+queryIsSupportedOnVersion :: Query blk result -> QueryVersion -> Bool
+queryIsSupportedOnVersion = \case
+  BlockQuery{} -> (>= QueryVersion1)
+  GetSystemStart{} -> (>= QueryVersion1)
+  GetChainBlockNo{} -> (>= QueryVersion2)
+  GetChainPoint{} -> (>= QueryVersion2)
+  GetLedgerConfig{} -> (>= QueryVersion3)
```
amesgen, line 143: Hmm, what is potentially confusing is that one needs to perform *two* checks for a `Query`: first, using this
```haskell
queryIsSupportedOnVersion ::
     Query blk result -> QueryVersion -> Bool
```
and if the query is a `BlockQuery`, then also using
```haskell
queryIsSupportedOnVersion ::
     BlockQuery (ShelleyBlock proto era) result -> ShelleyNodeToClientVersion -> Bool
```
on the payload.

---

Maybe we want to add an abstract
```haskell
blockQueryIsSupportedOnVersion :: Query blk result -> BlockNodeToClientVersion blk -> Bool
```
to `BlockSupportsLedgerQuery` such that it can be called here automatically?[^rich-return] E.g. the Shelley version defined above would be called automatically.

[^rich-return]: *Maybe* we want a richer return type instead of `Bool` similar to what you have in https://github.com/IntersectMBO/cardano-api/pull/790, but how necessary is that actually?

Then I think it would be easy to define a function
```haskell
queryIsSupportedOnVersion :: Query blk result -> NodeToClientVersion -> Bool
```
in Consensus that could be used by cardano-api or any other consumer, and we would own that logic.

amesgen (commented):
I think for consistency we would like to also add another variant of `queryIsSupportedOnVersion` for `HardForkNodeToClientVersion`. (Currently, this will always return true, but that might not be the case in the future.)

This should also be required by the abstract `blockQueryIsSupportedOnVersion` suggested below.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -135,6 +139,22 @@ instance All SingleEraBlock xs => BlockSupportsLedgerQuery (HardForkBlock xs) wh
...
+  blockQueryIsSupportedOnVersion q (HardForkNodeToClientDisabled x) = case q of
+    QueryIfCurrent (QZ q') -> blockQueryIsSupportedOnVersion q' x
+    QueryIfCurrent{}       -> False
+    QueryAnytime{}         -> False
+    QueryHardFork {}       -> False
+  blockQueryIsSupportedOnVersion q (HardForkNodeToClientEnabled hfv npversions) = case q of
+    QueryIfCurrent qc -> hfv >= HardForkSpecificNodeToClientVersion3 &&
+      go qc npversions
+    QueryAnytime{} -> hfv >= HardForkSpecificNodeToClientVersion3
+    QueryHardFork{} -> hfv >= HardForkSpecificNodeToClientVersion3
```
amesgen, line 151: I think all of these were actually added before `HardForkSpecificNodeToClientVersion3`, so I think you can replace these `hfv >= HardForkSpecificNodeToClientVersion3` with `True`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -73,7 +72,8 @@ import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.HeaderValidation
 import           Ouroboros.Consensus.Ledger.Basics
 import           Ouroboros.Consensus.Ledger.Extended
-import           Ouroboros.Consensus.Ledger.Query
+import           Ouroboros.Consensus.Ledger.Query hiding
+                     (queryIsSupportedOnVersion, querySupportedVersions)
```
amesgen, line 76:
```suggestion
import           Ouroboros.Consensus.Ledger.Query
```

amesgen (commented):
Looks great!

Some Haddocks might be nice, which explain e.g. that it is recommended invoke `blockQueryIsSupportedOnVersion` in the query encoder (maybe we could even provide a helper function for that), and provide an explanation why that is useful.

amesgen (approved):
LGTM (after fixing the innocent warning CI is complaining about)

amesgen:
(Force-pushed only for trivial hlint/stylish fixes and commit signing)

PR #1448, opened 2025-04-04
Update index-state to remove allow-newers

jasagredo:
Remove some `allow-newer`s


jasagredo:
Ah so there has been a breaking change :/

amesgen:
> Ah so there has been a breaking change :/

I think it is just a deprecation warning (`import Text.Builder` -> `import TextBuilder`) that CI fails on due to our use of `-Werror`.

jasagredo:
Closed in favor of #1449

PR #1449, opened 2025-04-07
Update index states

erikd:
# Description

* Update index-states
* Remove some `allow-newer` stanzas.

Also note that:

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -572,7 +572,7 @@ library unstable-cardano-tools
     sop-extras,
     strict-sop-core,
     text,
-    text-builder,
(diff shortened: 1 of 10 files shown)


ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -572,7 +572,7 @@ library unstable-cardano-tools
     sop-extras,
     strict-sop-core,
     text,
-    text-builder,
+    text-builder >=0.6.10,
```
erikd, line 575: I would have left a comment on this change, but the CI cabal formatting stuff wanted me to move the comment to just under the `build-depends` where it would have made zero sense.

jasagredo (approved):
We could try to get all the way to `text-builder-1` but let's merge this to unblock UTxO-HD.

PR #1466, opened 2025-04-15
consensus-testlib: add Test.CsjModel

nfrisby:
This PR model is the result of troubleshooting some test failures. I found a minor non-optimality in CSJ (recall that CSJ is fundamentally an optimization), and with the non-centralized definition of the existing CSJ implementation it was challenging for me to establish confidence that there weren't similar issues. I developed this centralized model for more legibility/clarity, and I intend for it to be used at least to test the non-centralized CSJ implementation.


ouroboros-consensus/src/unstable-consensus-testlib/Test/CsjModel/NonEmptySeq.hs
```diff
@@ -0,0 +1,94 @@
+{-# LANGUAGE BangPatterns #-}
```
nfrisby, line 1: This type doesn't use `AnchoredFragment` for two reasons.

- As a model, I'm trying to keep it as simple as possible. So far, I don't need headers at all, and I can't put points in an `AnchoredFragment` because of the `HasHeader` constraint.

- So far, the model deals with chains that reach all the way back to `Origin`, not _fragments_. As long as that's true, it seems misleading to use `AnchoredFragment`.

ouroboros-consensus/src/unstable-consensus-testlib/Test/CsjModel/NonEmptySeq.hs
```diff
@@ -0,0 +1,94 @@
...
+    module Test.CsjModel.NonEmptySeq
+  ) where
+
+import           Control.Monad (guard)
+import qualified Data.Maybe as L (Maybe (Just, Nothing))
+import           Data.Sequence (Seq)
+import qualified Data.Sequence as Seq
+
+-- | A non-empty 'Seq'
+newtype NonEmptySeq a = UnsafeNonEmptySeq (Seq a)
```
nfrisby, line 15: This type doesn't use `AnchoredFragment` for two reasons.

- As a model, I'm trying to keep it as simple as possible: so far, I don't need headers at all. And I can't put points in an `AnchoredFragment` because of the `HasHeader` constraint.

- So far, the model deals with chains that reach all the way back to `Origin`, not _fragments_. As long as that's true, it seems misleading to use `AnchoredFragment`.

nfrisby: The code no longer reaches all the way back to Genesis. It still doesn't involve headers explicitly, but perhaps `AnchoredSeq` could still be reused?

ouroboros-consensus/src/unstable-consensus-testlib/Test/CsjModel/Jumping.hs
```diff
@@ -0,0 +1,249 @@
...
+            Stop p                   ->
+                pure $ StopRunningNormally $ withOriginRealPointToPoint p
+  ,
+    jgOnAwaitReply      = step $ ChainSyncReply MsgAwaitReply
+  ,
+    jgOnRecvRollForward =
+        step . ChainSyncReply . MsgRollForwardSTART . castRealPoint
+  ,
+    jgOnRollForward     = \p ji -> atomically $ do
+        -- Suppress this stimulus if CSJ has already sent 'Stop'; this
```
nfrisby, line 164: Do `jgOnAwaitReply`, `jgOnRecvRollFordward`, and `jgOnRollBackward` also need to check for `Stop`, since the Diffusion Layer thread is calling the handlers as responses come in to the pipelined `MsgRequestNext` messages?

nfrisby:
Overnight, I ran:

- 4 copies of `-p\"/CSJ/&&/With some/&&/own thing/\" --quickcheck-shrinks 1000 --quickcheck-tests 10000`
- 28 copies of  `-p\"/CSJ/" --quickcheck-shrinks 1000 --quickcheck-tests 1000`

The testsuite has a 10x factor, so the three of the four CSJ tests ran 280000 times, and the fourth ran 680000 times.

All tests passed! 🙌 Encouraging.

ouroboros-consensus/src/unstable-consensus-testlib/Test/CsjModel/UglyStylishHaskellCppWorkaround.hs
```diff
@@ -0,0 +1,12 @@
+{-# LANGUAGE CPP #-}
```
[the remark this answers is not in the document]
nfrisby: image
image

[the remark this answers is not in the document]
amesgen: Our usual approach to fix this issue is to use
```haskell
import           Data.List as List (foldl')
```
and then use `List.foldl'`.

This doesn't require CPP on any GHC version.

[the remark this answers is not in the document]
nfrisby: Ah, this option is weird looking but otherwise much more attractive: https://github.com/haskell/core-libraries-committee/issues/167#issuecomment-1546692717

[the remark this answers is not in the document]
nfrisby: 🤣 same problem with `liftA2` and `base` 4.18.

So I'm switching to Lelechenko's recommend trick.

nfrisby:
I think the tree of the current tip commit is in pretty good shape.; passing huge number of `-p/CSJ/` tests locally. We discussed next steps in our call today.

- I'll pause and ramp up on Leios Innovation work.
- @amesgen will start reviewing this PR.
- The nearest milestone seems to be: split out a smaller PR that merely abstracts the ChainSync client over the CSJ `registerClient` function and reorganize the tests in this PR just slightly so that `CsjModel.hs` is also used in the Genesis tests (maybe the `TestSetup` determines which governor, 50-50 and shrink to `CsjModel.hs`?).
- Principal remaining work for a follow-up PR is to confirm that the gap between the CSJ implementation and `CsjModel.hs` is just the risk of unnecessary header downloads and nothing more than that. Secondary remaining work is the flaky "node restart during sync" test.

ouroboros-consensus/src/unstable-consensus-testlib/Test/CsjModel.hs
```diff
@@ -0,0 +1,1218 @@
...
+                pid' == pid
+             &&
+                -- Don't demote the Dynamo if we'll have to immediately
+                -- repromote the same peer.
+                any eligible (nonDynamos x)
+            eligible = \case
+                Jumped{}   -> True
+                Jumping{}  -> False
+                Objector{} -> True
+        pure (x, if not shouldDemote then [] else [(pid, Demoting)])
```
nfrisby, line 574: This `Demoting` isn't enough. Specifically:

- The Dynamo is adversarial, withholding blocks.
- But the Dynamo is serving the honest chain, and all Jumpers have fully accepted the latest jump request.
- The Dynamo has served some headers after that jump request, but not enough for another jump.

In this scenario, the Dynamo has the longest chain, so Devoted BlockFetch will favor them. Also, the Dynamo's ChainSync client is fully blocked until a new block arrives, which it won't. So it'll never send the `Demoted` stimulus---the `demoted` definition just below won't happen. So the CSJ state and the syncing node overall are now stuck.

I anticipate the smoothest way to fix this is to introduce a new `NonDynamo` which is `FormerDynamo`. The CSJ governor expects it to send MsgRollForward etc and also `Demoted`. Once it sends `Demoted`, it's just a Jumper again. The benefit is that this way a new Dynamo can be elected before this Dynamo being demoted raises `Demoted`.

PR #1482, opened 2025-04-24
db-immutaliser: more info on volatile candidates 

amesgen:
Enrich db-immutaliser to print info about the candidates in a VolatileDB. This is useful when analyzing a VolatileDB with a priori unknown content (especially when it contains various forks).

Concretely, it will print all maximal candidates, as well as a block tree illustrating the fork relation.

An invocation looks like this (also see the updated README section):
```
db-immutaliser --immutable-db /path/to/immutable --volatile-db /path/to/volatile --config /path/to/config.json \ 
  --verbose --dot-out /path/to/block-tree.dot --dry-run
```

Here, `/path/to/out-imm` must be the path to an appropriate ImmutableDB (we are only processing volatile blocks that are reachable from the immutable tip) that will *not* be modified here due to `--dry-run` (new in this PR). In the simple case where no blocks are immutable yet (which can be the case for short-lived testnets), this can just be an empty directory.

The resulting block tree can be visually rendered as a graph, see the README, or https://github.com/IntersectMBO/ouroboros-consensus/pull/1482#issuecomment-2828083404 for an example.

---

A potential alternative is to put this into a separate tool, but that could also happen later.

### DIFF
--- a/ouroboros-consensus-cardano/README.md
+++ b/ouroboros-consensus-cardano/README.md
@@ -335,6 +335,8 @@ Additional flags:
    immutable tip in the graphviz DOT format to a file. The `dot` CLI tool can
    then be used to graphically render this tree.
 
+ - `--dry-run`: Do not actually append anything to the ImmutableDB.
+
 ## db-synthesizer
 
 ### About
--- a/ouroboros-consensus-cardano/README.md
+++ b/ouroboros-consensus-cardano/README.md
@@ -326,6 +326,15 @@ cabal run db-immutaliser -- \
 The `config.json` should be in the same format (Node configuration) as for eg
 db-analyser.
 
+Additional flags:
+
+ - `--verbose`: Print additional information on the blocks reachable from the
+   immutable tip, as well as all possible volatile candidates.
+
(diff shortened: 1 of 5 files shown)


amesgen:
Example output for the block tree:
```
ImmTip
|
`- e8e6
   |
   `- 3c45
      |
      +- 1783
      |
      +- 7ef0
      |  |
      |  `- 0408
      |     |
      |     `- cba5
      |        |
      |        `- a56d
      |           |
      |           `- 6e71
      |              |
      |              `- 5c6f
      |                 |
      |                 +- 03e2
      |                 |  |
      |                 |  `- 5a74
      |                 |     |
      |                 |     `- 8ff2
      |                 |        |
      |                 |        `- 7d0d
      |                 |
      |                 `- 05da
      |                    |
      |                    `- d7e1
      |                       |
      |                       `- aad1
      |                          |
      |                          `- cbb0
      |                             |
      |                             `- c50a
      |
      `- f560
         |
         `- 3547
```

jasagredo (approved):
Very nice, a very straight-forward PR. Thanks Alex!

Issue #1077, opened 2024-04-25
Refine interface for node creating `BlockForging` closures for the sake of KES Agent


nfrisby:
The existing Consensus interface used by the node code to create `BlockForging` structures (eg upon init and SIGHUP) that are then passed to Consensus do not obviously suffice for the needs of the https://github.com/IntersectMBO/cardano-base/issues/312 (owned by @tdammers) refinement. This Issue is to create an alternative path that is more obviously compatible with the KES Agent's needs, and to establish confidence (eg via tests) that it is otherwise equivalent in behavior to the existing path.

Issue #558, opened 2022-09-15
Adapt KES usage to new, non-pure, interface


iquerejeta:
We are changing the way the KES keys are handled, to increase the guarantees that the key is not copied to disk and safely deleted by mlocking the secret data. These changes have modified the KES interface, which use to expose pure functions, to a non-pure interface. 

The changes are being performed in input-output-hk/cardano-base#255 and input-output-hk/cardano-base#317. 

To be able to work on this, the PRs mentioned above need to be merged.

tdammers:
PR: #1374

dnadales:
Seems to have been addressed in https://github.com/IntersectMBO/ouroboros-consensus/pull/1374#issuecomment-3991540049 We need to double-check this.

PR #1487, opened 2025-04-30
KES agent integration

jasagredo:
This changes Consensus such that mlocked KES keys are used internally.

This is important groundwork for supporting KES agents in the future. In this form, the code will still load KES keys from disk, which is unsound, but the internal machinery is ready to also accept KES keys from other sources, and once loaded, KES keys will be handled appropriately (kept in mlocked RAM at all times, securely erased when expired).

This also involves a restructuring of the `HotKey` data structure, which now manages not only a KES SignKey, but also the corresponding OpCert. This is necessary for two reasons:

- With a KES agent, the OpCert will be provided along with the SignKey; this is the easiest way to make sure that the OpCert always matches the SignKey it is used with
- With the new structure, KES keys can be replaced in a live node kernel, without having to restart it and reload the entire configuration. Because of this, the HotKey, which manages the dynamic part of the node kernel's signing mechanism, needs to be able to replace not just the SignKey (which it already did, in order to handle key evolution), but also the OpCert (which will not change when a SignKey evolves, but it will change when a new SignKey is provided).

Supersedes #1284.

Issue #558.

This adds KES Agent connectivity to consensus.

To use a KES Agent to source KES SignKeys and OpCerts, the `praosCredentialsSource` in the `PraosCanBeLeader` data structure can now be pointed to a domain socket address where it will look for a KES Agent.

Also covers #1077.


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
```diff
@@ -170,8 +168,12 @@ validateGenesis = first errsToString . SL.validateGenesis
 protocolInfoShelley ::
      forall m c.
       ( IOLike m
+      , AgentCrypto c
       , ShelleyCompatible (TPraos c) ShelleyEra
       , TxLimits (ShelleyBlock (TPraos c) ShelleyEra)
+      , MonadKESAgent m
+      , SerDoc.HasInfo (Agent.DirectCodec m) (KES.VerKeyKES (KES c))
+      , SerDoc.HasInfo (Agent.DirectCodec m) (KES.SignKeyKES (KES c))
```
jasagredo, line 176: Why is this doc stuff needed at all?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -74,6 +76,7 @@ data Tracers' remotePeer localPeer blk f = Tracers
   , gddTracer                     :: f (TraceGDDEvent remotePeer blk)
   , csjTracer                     :: f (TraceLabelPeer remotePeer (CSJumping.TraceEventCsj remotePeer blk))
   , dbfTracer                     :: f (CSJumping.TraceEventDbf remotePeer)
+  , kesAgentTracer                :: f KESAgentClientTrace
```
jasagredo, line 79: This tracer is defined but unused

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/AgentClient.hs
```diff
@@ -0,0 +1,183 @@
...
+            case mBundle of
+              Just (Agent.Bundle skpRef ocert) -> do
+                -- We take ownership of the key, so we acquire one extra reference,
+                -- preventing the key from being discarded after `handleKey`
+                -- finishes.
+                _ <- acquireCRef skpRef
+                withCRefValue skpRef $ \(SignKeyWithPeriodKES sk p) ->
+                  handleKey (convertOCert ocert) sk p
+                return Agent.RecvOK
+              _ -> undefined -- TODO
```
jasagredo, line 156: What to do with this?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -244,16 +255,57 @@ instance Crypto c => ChainOrder (PraosChainSelectView c) where
...
+              startPeriod
+              maxKESEvolutions
+
+instantiatePraosCredentials maxKESEvolutions (PraosCredentialsAgent path) = do
+  HotKey.mkDynamicHotKey
+      maxKESEvolutions
+      (Just $ \send -> do
+        let handleKey ocert sk p = do
+              send ocert sk p (OCert.ocertKESPeriod ocert)
+        runKESAgentClient nullTracer path handleKey
```
jasagredo, line 306: Here we should probably pass the tracer above?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/State.hs
```diff
@@ -175,9 +171,7 @@ data DynamoInitState blk
   deriving (Generic)
 
 deriving anyclass instance
-  ( HasHeader blk,
-    LedgerSupportsProtocol blk,
-    NoThunks (Header blk)
+  ( LedgerSupportsProtocol blk
```
jasagredo, line 174: I'm confused about why we can do this but it seems to work fine

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/MockCrypto.hs
```diff
@@ -29,23 +31,17 @@ import           Control.State.Transition.Extended (PredicateFailure)
...
---
--- We run the tests with this mock crypto, as it is easier to generate and
--- debug things. The code is parametric in the crypto, so it shouldn't make
--- much of a difference. This also has the important advantage
--- that we can reuse the generators from cardano-ledger-specs.
-data MockCrypto
-
 instance Crypto MockCrypto where
-  type KES      MockCrypto = MockKES 10
+  type KES      MockCrypto = MockKES 128
```
jasagredo, line 44: This was done in order to align with https://github.com/input-output-hk/kes-agent/blob/a212f98b9bc567ebf3c8472e62f28450990f7821/kes-agent/src/Cardano/KESAgent/Protocols/StandardCrypto.hs#L48

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
tdammers: SerDoc is a library for semi-automatically generating serialization documentation from types; this can then be used to generate documentation for `typed-protocols`. In a perfect world, we wouldn't need this here, but the KES agent protocols (or at least, some of them) are generic in the crypto, and if we want to be able to generate protocol documentation for those, we need to demand those `SerDoc` classes on them. This constraint then propagates up to the KES Agent API, and anything that uses it.
If this is a major complaint, I can restructure the code such that the protocols are no longer parametrized over the crypto, but always use the `StandardCrypto`. This would make it impossible to test the KES agent protocols against `MockCrypto`, but I don't think that would be a major concern to be honest.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
tdammers: Oh dear, yes... looks like I forgot to actually implement that part. I'll take a look ASAP.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/AgentClient.hs
tdammers: Ah, right; that's the case where the agent sends a "key drop" message. I'll take care of it.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
tdammers: Indeed, but I'm not entirely sure how to do this cleanly, since `instantiatePraosCredentials` is used from within the block forgings defined in `protocolInfoCardano`, which doesn't (and shouldn't) have `Tracers` in scope. The best I can come up with at this point is extend `protocolInfoCardano` such that the block forging generator it returns takes the tracer as an extra argument, but that change would then ripple to all call sites of `protocolInfoCardano`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
jasagredo: My concern here is that we don't use SerDoc anywhere else, so even if it is a nice feature we might not leverage it in the end. Shouldn't some documentation be generated somewhere making use of this feature? Or couldn't this be attached to the `kes-agent` executable instead of to the library and used there to generate whatever docs you want, leaving Consensus free of these constraints?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
jasagredo: If we need it, then so be it

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/MockCrypto.hs
jasagredo: This was reverted

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -183,8 +187,8 @@ genCoreNode startKESPeriod = do
     genKey <- genKeyDSIGN <$> genSeed (seedSizeDSIGN (Proxy @LK.DSIGN))
     delKey <- genKeyDSIGN <$> genSeed (seedSizeDSIGN (Proxy @LK.DSIGN))
     stkKey <- genKeyDSIGN <$> genSeed (seedSizeDSIGN (Proxy @LK.DSIGN))
-    vrfKey <- genKeyVRF   <$> genSeed (seedSizeVRF   (Proxy @(VRF c)))
-    kesKey <- unsoundPureGenKeyKES <$> genSeed (seedSizeKES (Proxy @(KES c)))
+    vrfKey <- genKeyVRF   <$> genSeed (seedSizeVRF   (Proxy @(VRF   c)))
+    kesKey <- unsoundPureGenKeyKES <$> genSeed (seedSizeKES   (Proxy @(KES   c)))
```
tdammers, line 191: Why the extra spaces?

ouroboros-consensus-protocol/changelog.d/20250130_101128_tdammers_mlocked_kes_rebase.md
```diff
@@ -0,0 +1,18 @@
...
+- Add finalizer to `HotKey`. This takes care of securely forgetting any KES
+  keys the HotKey may still hold, and will be called automatically when the
+  owning block forging terminates.
+- Change `PraosCanBeLeader` to not contain the KES sign key itself anymore.
+  Instead, it now contains a `PraosCredentialsSource` field, which
+  specifies how to obtain the actual credentials (OpCert and KES SignKey). For
+  now, the only supported method is passing an OpCert and an
+  UnsoundPureSignKeyKES, presumably loaded from disk
+  (`PraosCredentialsUnsound`); future iterations will add support for
+  connecting to a KES agent.
```
tdammers, line 18:
```suggestion
  specifies how to obtain the actual credentials (OpCert and KES SignKey).
  Two methods are supported:
  - Directly passing an OpCert and an UnsoundPureSignKeyKES, presumably loaded
    from disk (`PraosCredentialsUnsound`)
  - Passing a socket address to a KES Agent from which OpCerts and (sound)
    SignKeyKES can be obtained (`PraosCredentialsAgent`)
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/TPraos.hs
tdammers: Indeed that's a bit unfortunate. I had envisioned a setup where we could generate human-readable protocol documentation for all protocols via `typed-protocols-doc`, but due to recent changes to the `typed-protocols` API, and bugs in earlier versions of `typed-protocols-doc`, this is only viable for protocols that use the latest `typed-protocols`, which we currently don't.
The actual documentation generator would be a separate executable under the `kes-agent` project, but in order to do its thing, it does need a few constraints, and due to the way the code abstracts over crypto algorithms etc., some of those constraints "infect" other code that uses the same data structures. If the whole documentation generation thing isn't something we want to pursue, then yes, we can remove all that stuff, but that would also require removing it from `kes-agent`.

tdammers (changes requested):
Changelog needs to be fixed, rest looks good.

fraser-iohk:
This PR has been superceded by #1620

PR #1488, opened 2025-04-30
add Show instance for HardForkLedgerConfig

### LINKED ISSUE #1414: [FEAT] - Implement `Show (CardanoLedgerConfig StandardCrypto)`
**Internal**

**Describe the feature you'd like**
Using the `GetLedgerConfig` query in `cardano-cli` requires some way of presenting the ledger config to the user. It can be either a show instance or a custom in-line show logic. The query needs some way of presenting the result to the user before it can be integrated.

**Describe alternatives you've considered**

**Additional context / screenshots**

fraser-iohk:
# Description

Adds a `Show` instance for `HardForkLedgerConfig`

Fixes #1414

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Show.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Show.hs
@@ -1,0 +1,9 @@
+module Test.Consensus.Cardano.Show () where
+
+import           Ouroboros.Consensus.Cardano ()
+import           Ouroboros.Consensus.Cardano.Block
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+
+-- | This definition exists solely to ensure that 'CardanoLedgerConfig' has a working 'Show' instance that isn't accidentally removed or broken in future
+_showCardanoLedgerConfig :: CardanoLedgerConfig StandardCrypto -> String
+_showCardanoLedgerConfig = show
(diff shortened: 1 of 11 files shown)


amesgen (commented):
LGTM, what about adding a `Dict (Show (LedgerConfig CardanoBlock))` (or equivalent) to a test suite to make sure we don't accidentally regress here?

amesgen:
> > LGTM, what about adding a `Dict (Show (LedgerConfig CardanoBlock))` (or equivalent) to a test suite to make sure we don't accidentally regress here?
> 
> yeah, I thought about doing this -- do you think that going with an explicit `Dict` is clearer than just adding a `_showCardanoLedgerConfig = show` definition in there with a comment?

That also sounds good :+1:

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Show.hs
```diff
@@ -0,0 +1,9 @@
+module Test.Consensus.Cardano.Show () where
+
+import           Ouroboros.Consensus.Cardano ()
+import           Ouroboros.Consensus.Cardano.Block
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
```
amesgen, line 5: Would be really nice to get rid of the orphan instances (#588), but this is out of scope for this PR

amesgen (approved):

PR #1489, opened 2025-04-30
Drop GHC 8.10

amesgen:
The node already uses 9.6, so we can finally drop 8.10 and get rid of a lot of cruft.

I tried to remove everything I could find by grepping for a few ad-hoc regexes; maybe there is more, but we can remove that as we go.

Follow-up work (like upgrading from `Haskell2010` to `GHC2021`) can happen in later PRs as it needs more discussion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.6.6", "9.10.1", "9.12"]
+        ghc: ["9.6.6", "9.10.1", "9.12"]
         variant: [default, no-thunks]
         test-set: [all, no-thunks-safe]
         exclude:
@@ -193,7 +193,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.6.6", "9.10.1"]
+        ghc: ["9.6.6", "9.10.1"]
 
     steps:
     - uses: actions/checkout@v4
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -61,8 +61,7 @@ common common-bench
   -- We use this option to avoid skewed results due to changes in cache-line
   -- alignment. See
   -- https://github.com/Bodigrim/tasty-bench#comparison-against-baseline
-  if impl(ghc >=8.6)
-    ghc-options: -fproc-alignment=64
+  ghc-options: -fproc-alignment=64
 
 library
   import: common-lib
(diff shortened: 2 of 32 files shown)


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.6.6", "9.10.1", "9.12"]
+        ghc: ["9.6.6", "9.10.1", "9.12"]
```
jasagredo, line 41: We could bump 9.6 to 9.6.7 and 9.12 to 9.12.1. But it doesn't need to be now.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -358,8 +358,6 @@ library
 
   if !impl(ghc >=9.4)
     build-depends: data-array-byte
```
jasagredo, line 360: I think this can also be removed?

jasagredo:
I also see this:
```
-- ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs
...
    commonPrefix
      | null frags = AF.Empty sharedAnchor
      -- TODO use Foldable1 once all our GHCs support it
      | otherwise = L.foldl1' computeCommonPrefix (toList frags)
```

Not sure if GHC >=9.6.7 supports it

jasagredo:
Also worth double-checking this one:
```
-- Ouroboros.Consensus.Protocol.PBFT

#if __GLASGOW_HASKELL__ >= 908
-- GHC is a bit pickier for data family instances, but trying to remove this
-- one forces us to reorganize the Protocol.* modules. TODO eventually.
#endif
```

jasagredo:
And this one:
```
-- Ouroboros.Consensus.HardFork.Combinator.Embed.Unary

  Defaults (to ease implementation)

  It'd be nicer to use deriving-via here, but we cannot due to a GHC bug
  (resulting in @No family instance for ‘GenTx’@ errors).
  See <https://gitlab.haskell.org/ghc/ghc/issues/13154#note_224287> .
```

The fix seems to be included in multiple tags dating as far back as 8.10.4: https://gitlab.haskell.org/ghc/ghc/-/commit/cd9b94594440163a1a726300d300f76ff05cd15a

.github/workflows/ci.yml
amesgen: Will do that in a follow-up PR where I also update haskell.nix :+1: 

(also, there even are 9.10.2 and 9.12.2 already)

amesgen:
Thanks, addressed your comments (apart from https://github.com/IntersectMBO/ouroboros-consensus/pull/1489#issuecomment-2853647494 as we currently still need GHC 9.6), great finds.

jasagredo (approved):

Issue #1491, opened 2025-04-30
Enriched Genesis tests to confirm all Genesis components are disabled when GSM transitions to CaughtUp


nfrisby:
Today's test suite doesn't test this, and so didn't catch Issue #1490. It's straight-forward enough that I'm surprised code review didn't catch it. Regardless, it's tech debt to not cover that with tests.

Moreover: for GDD, LoP, LoE, etc the logic for disabling and re-enabling is trivial, but for CSJ the re-enable logic might not be trivial. There is at least one trivial option, but it's somewhat wasteful. So some tests should exercise this.

PR #1492, opened 2025-04-30
ChainSync: let GSM disable and re-enable CSJ; also enable LoP in PreSyncing

### LINKED ISSUE #1490: [BUG] - CSJ is always on, regardless of the GSM state
The Genesis specification declares that all Genesis components, including CSJ, should be disabled when the node finishes syncing.

The CSJ optimization is not influenced at all by the GSM state, so that's a bug.

It means that a Genesis-enabled node that finishes syncing will still only allow a single peer to run normal ChainSync at a time. If none of its peers are trying to attack this bug, then they'll all one-by-one be activated by CSJ (as its "dynamo"), send `MsgAwaitReply` soon enough, and so disengage from CSJ. In that case, this bug merely slightly prolongs the transition from syncing to caught-up and then also introduces a similar brief delay for any upstream peer that is connected to afterwards.

However, if an upstream peer behaves adversarially, by putting off sending `MsgAwaitReply` for as long as they can, then the bug becomes much more severe. Adversaries can choose that behavior while the victim is syncing, but Genesis's Limit on Patience (LoP) strictly bounds how long they can do it for. The LoP, however, is correctly disabled when the GSM transitions to `CaughtUp`. In the absence of the LoP, the adversarial peer can avoid sending `MsgAwaitReply` for a long time, up to several hours. During that time, any new upstream peers will be prevented by CSJ from syncing any headers. Any upstream peers that existed before the victim connected to the adversary are unaffected (because they already disengaged from CSJ). Because the Diffusion Layer churns every 30 minutes or so, a clever adversary might be able to effectively eclipse the victim. They won't be able to feed it an alternative chain, though---they can just ensure it sees the fresh honest chain's blocks much later than it should have.

The fix is simply to disable CSJ when the GSM transitions to `CaughtUp`, and also to re-enable CSJ when the GSM transitions back out of `CaughtUp`. The forward transition is trivial, since the GSM only transitions when all peers have sent `MsgAwaitReply` (and so are disengaged). The backwards transition, though, will require a small amount of new logic for CSJ to be able to re-engage a peer that it had previously disengaged.

nfrisby:
Fixes #1490. Contributes to #1491.

### DIFF
--- a/ouroboros-consensus/changelog.d/20250506_090817_nick.frisby_bugfix_csj_gsm.md
+++ b/ouroboros-consensus/changelog.d/20250506_090817_nick.frisby_bugfix_csj_gsm.md
@@ -1,0 +1,25 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Patch
+
+- Bugfix in Ouroboros Genesis. Added the @GsmState@ argument to the
+  @registerClient@ function for the ChainSync Jumping optimization. If the node
+  is in the @GSM.CaughtUp@ state, new peers now immediately disengage from CSJ.
+
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -410,8 +410,8 @@ bracketChainSyncClient
             , cschJumpInfo
             }
           insertHandle = atomicallyWithMonotonicTime $ \time -> do
-            initialGsmState <- getGsmState
-            updateLopBucketConfig lopBucket initialGsmState time
+            gsmState <- getGsmState
+            updateLopBucketConfig lopBucket gsmState time
             cschcAddHandle varHandles peer handle
           deleteHandle = atomically $ cschcRemoveHandle varHandles peer
       bracket_ insertHandle deleteHandle $ f Jumping.noJumping
@@ -424,13 +424,15 @@ bracketChainSyncClient
     acquireContext lopBucket cschState (CSJEnabledConfig jumpSize) = do
         tid <- myThreadId
         atomicallyWithMonotonicTime $ \time -> do
-          initialGsmState <- getGsmState
-          updateLopBucketConfig lopBucket initialGsmState time
+          gsmState <- getGsmState
+          updateLopBucketConfig lopBucket gsmState time
           cschJumpInfo <- newTVar Nothing
           context <- Jumping.makeContext varHandles jumpSize tracerCsj
-          Jumping.registerClient context peer cschState $ \cschJumping -> ChainSyncClientHandle
+          Jumping.registerClient gsmState context peer cschState $ \cschJumping -> ChainSyncClientHandle
             { cschGDDKill = throwTo tid DensityTooLow
             , cschOnGsmStateChanged = updateLopBucketConfig lopBucket
+                -- See Note [Updating the CSJ State when the GSM State Changes]
+                -- in the Haddocks of 'Jumping.registerClient'.
             , cschState
             , cschJumping
             , cschJumpInfo
--- a/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/CSJ.hs
+++ b/ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/CSJ.hs
@@ -1,0 +1,304 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | Narrow tests for ChainSync Jumping
+module Test.Consensus.MiniProtocol.ChainSync.CSJ (tests) where
+
+import qualified Control.Concurrent.Class.MonadSTM.Strict.TVar as TVar
+import           Control.Monad (void)
+import           Control.Monad.Class.MonadTimer (MonadTimer)
+import           Control.Monad.IOSim (runSim)
+import           Control.ResourceRegistry
+import           Control.Tracer (nullTracer)
+import           Data.Typeable
+import           Network.TypedProtocol.Channel
+import           Network.TypedProtocol.Driver.Simple
+import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.BlockchainTime
+import           Ouroboros.Consensus.Config
+import qualified Ouroboros.Consensus.HeaderStateHistory as HeaderStateHistory
+import           Ouroboros.Consensus.Ledger.Tables.Utils (forgetLedgerTables)
+import           Ouroboros.Consensus.MiniProtocol.ChainSync.Client
+                     (CSJConfig (..), CSJEnabledConfig (..), ChainDbView (..),
+                     ChainSyncLoPBucketConfig (..), ChainSyncStateView (..),
+                     ConfigEnv (..), Consensus, DynamicEnv (..),
+                     bracketChainSyncClient, chainSyncClient,
+                     newChainSyncClientHandleCollection)
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.HistoricityCheck as HistoricityCheck
+import qualified Ouroboros.Consensus.MiniProtocol.ChainSync.Client.InFutureCheck as InFutureCheck
+import qualified Ouroboros.Consensus.Node.GsmState as GSM
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion
+                     (NodeToNodeVersion)
+import           Ouroboros.Consensus.NodeId
+import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Consensus.Util.STM (Fingerprint (..),
+                     WithFingerprint (..))
+import qualified Ouroboros.Network.AnchoredFragment as AF
+import           Ouroboros.Network.ControlMessage (ControlMessage (..))
+import qualified Ouroboros.Network.Mock.Chain as MockChain
+import           Ouroboros.Network.Protocol.ChainSync.ClientPipelined
+import           Ouroboros.Network.Protocol.ChainSync.Codec (codecChainSyncId)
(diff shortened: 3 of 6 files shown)


nfrisby:
This is Draft because:

- No tests (see Issue #1491).
- I'm not yet sure the CSJ re-enable logic I've added here is respecting the CSJ logic's invariants.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +426,39 @@ bracketChainSyncClient
...
+                  -- not have 'Nothing' in its 'cschJumpInfo'.
+                  case gsmTransition of
+                    PreSyncingSyncing  -> pure $ pure ()
+                    SyncingCaughtUp    ->
+                      -- The GSM only transitions to CaughtUp if all peers have
+                      -- sent MsgAwaitReply, so all peers are already
+                      -- disengaged.
+                      pure $ pure ()
+                    CaughtUpPreSyncing -> do
+                      -- TODO do we want to do something more sophisticated
```
nfrisby, line 451: Highlighting a new code TODO

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +426,52 @@ bracketChainSyncClient
...
+                      -- The GSM only transitions out of CaughtUp if the tip of
+                      -- its selection is much older than expected (eg 20
+                      -- minutes). There are many possible explanations for why
+                      -- that could have happened.
+                      --
+                      -- Some of those explanations would permit some of the
+                      -- current upstream peers to be honest. For that reason,
+                      -- it would be unsound to leave them as disengaged peers:
+                      -- they might include the only upstream peers that
+                      -- satisfy the Honest Availability Assumption. TODO check
```
nfrisby, line 460: Highlighting a new code TODO

nfrisby:
TODO Now that the CSJ changes are actually so minor, we should split out the LoP changes. It'll make backporting the CSJ diff that much easier.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +424,70 @@ bracketChainSyncClient
...
+                  --
+                  -- Leaving the existing peers' CSJ states (disengaged,
+                  -- Dynamo, etc) as-is also makes sense for the other GSM
+                  -- transitions.
+                  --
+                  -- It seems plausible that a non-trivial extension of the CSJ
+                  -- behavior might handle the different possibilities of
+                  -- 'CaughtUpPreSyncing' better (and most likely adaptively).
+                  -- However, that new logic is not immediately obvious, and so
+                  -- it's not clear that the extra complexity is worthwhile.
```
nfrisby, line 472: The only motivator I can think of for trying to handle this transition more carefully is if there's a bug in the CaughtUp BlockFetch such that the node might have several honest peers that are stuck on the forecast horizon "arbitrarily far" behind the wall clock (the hypothetical bug would be that the node has honest peers but is for some reason failing to download their blocks for more than ~20 minutes). If the GSM `CaughtUpPreSyncing` transition ultimately unstucks the node, then its selection will finally advance, and now all of those honest peers will have to send all the headers between where the node had been stuck and the actual wall clock. Note that the node might have been stuck for much more than 20 minutes (ie for multiple cycles of the GSM anti-thrashing delay).

But it seems unwise to complicate CSJ to guard against a potential bug in CaughtUp's BlockFetch logic. IE we should assume that a node whose selection is far behind the honest chain simply has been eclipsed: ie has no honest peers, and so CSJ isn't obligated to reduce load on whatever upstream peers the node had _during_ the eclipse.

nfrisby: What do you think, @coot @crocodile-dentist ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +424,70 @@ bracketChainSyncClient
...
-            }
+          Jumping.registerClient gsmState context peer cschState $ \cschJumping ->
+            ChainSyncClientHandle
+              { cschGDDKill = throwTo tid DensityTooLow
+              , cschOnGsmStateChanged = updateLopBucketConfig lopBucket
+                  -- The GSM only transitions out of CaughtUp if the tip of its
+                  -- selection is much older than expected (eg 20 minutes).
+                  -- There are many possible explanations for why that could
+                  -- have happened, so it's not obvious what is the best thing
+                  -- to do for 'CaughtUpPreSyncing'.
```
amesgen, line 439: One other potential scenario: Someone ran a (caught-up) Daedalus wallet and closed their laptop lib, suspending their machine. Then, when they wake up again after a ~week. In that case, using CSJ would be beneficial.

However, I am not sure if this is a likely scenario; maybe Daedalus already takes care of restarting the node in this case.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +424,70 @@ bracketChainSyncClient
...
+                  -- However, that new logic is not immediately obvious, and so
+                  -- it's not clear that the extra complexity is worthwhile.
+                  -- Especially since the risk due to not re-engaging the
+                  -- existing peers seems minor: some thousands of headers
+                  -- downloaded from whichever honest upstream peers happened
+                  -- to join "just before" 'CaughtUpPreSyncing' (because they
+                  -- were left as Disengaged instead of being reset to
+                  -- Jumpers).
+                  --
+                  -- One key remark: the 'CaughtUpPreSyncing' transition does
```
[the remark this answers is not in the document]
amesgen: I am confident that the logic everywhere else is correct.

One potential point of discussion is whether to run the GDD in `PreSyncing`:
https://github.com/IntersectMBO/ouroboros-consensus/blob/5730b89d4c74ec9f1006374bf2350509a0ba43bb/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Genesis/Governor.hs#L133-L136
We could do it, but it doesn't seem essential; but still might be nice to be consistent with the LoP.

amesgen (commented):
Looks good, left a few comments

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -424,17 +424,70 @@ bracketChainSyncClient
...
+                  -- The GSM only transitions out of CaughtUp if the tip of its
+                  -- selection is much older than expected (eg 20 minutes).
+                  -- There are many possible explanations for why that could
+                  -- have happened, so it's not obvious what is the best thing
+                  -- to do for 'CaughtUpPreSyncing'.
+                  --
+                  -- - If the tip is 20 minutes old because the network itself
+                  --   is failing to mint new blocks---the probability of the
+                  --   leader schedule just being empty is high enough it'll
+                  --   plausibly happens one or twice per year---then the
```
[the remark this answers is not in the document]
nfrisby: Thanks for the details. I will not Resolve this comment, for the benefit of future people reading this PR.

But I no longer go into this level of detail in the comment.

ouroboros-consensus/changelog.d/20250506_090817_nick.frisby_bugfix_csj_gsm.md
```diff
@@ -0,0 +1,26 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
```
amesgen, line 7: HTML comment wrapper needs to be removed

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/CSJ.hs
```diff
@@ -0,0 +1,336 @@
...
+      , bftSignKey = SignKeyMockDSIGN 0
+            -- inconsequential for this test
+      , bftVerKeys = Map.empty
+      }
+  , topLevelConfigLedger      = testBlockLedgerConfigFrom eraParams
+  , topLevelConfigBlock       = TestBlockConfig numCoreNodes
+  , topLevelConfigCodec       = TestBlockCodecConfig
+  , topLevelConfigStorage     = TestBlockStorageConfig
+  , topLevelConfigCheckpoints = emptyCheckpointsMap
+  }
```
amesgen, line 336:
```suggestion
-- inconsequential for this test
topLevelCfg :: TopLevelConfig TestBlock
topLevelCfg = singleNodeTestConfig
```

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/CSJ.hs
```diff
@@ -0,0 +1,336 @@
...
+                varHandles
+                (pure GSM.CaughtUp)
+                peer
+                version
+                lopBucketConfig
+                (CSJEnabled csjConfig)
+                diffusionPipelining
+                (k . mkClient)
+
+        spawnConnection :: CoreNodeId -> m (TVar.StrictTVar m Bool)
```
amesgen, line 231: Create a `Bool`-isomorphic type to aid local understanding/readability?

amesgen (approved):
Looks great, only very minor comments :heart:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
coot: `BlockFetch` has 60s timeout on downloading a block, so if a request was made (to an honest node) it will either be fulfilled or it will timeout and the connection will be dropped. It could be that we have a bug in the block-fetch logic that doesn't issue a request - this should be covered by this test https://github.com/IntersectMBO/ouroboros-network/blob/main/ouroboros-network/testlib/Test/Ouroboros/Network/BlockFetch.hs#L180-L201

PR #1500, opened 2025-05-09
Introduce `fourmolu`

### LINKED ISSUE #1499: Introduce automatic code formatting
The Consensus team has decided to introduce a code formatter to enforce consistency of the code style automatically.

## Motivation

- reduce the cognitive load on the developers when working on the code: no need to think how the code is formatted
- dramatically simplify the code style document

## Implementation

The team has agreed to use the `fourmolu` tool with the following configuration:

- line length limit is set to 100 characters
- indentation is set to 4 spaces

further configuration settings, such as formatting of records, comments, etc. will be set to minimise the reformatting.

### Code style application locally

The `forumolu` tool will need to be available in the developer's environment, and the configuration file will be set by the `fourmolu.yaml` file in the repository root. The developer are free to choose how they apply the code style. Some options are:

- on every save by means of the language server protocol hook in the editor
- as a Git hook
- manually via a script invocation

### ~Automated code style application in CI~

Crossed-out for the initial roll out of `fourmolu` as it requires too much effort.

<details>
<summary>See detailed idea</summary>

In addition to *checking* the code style in CI, it is also convenient to *apply* it as well. In my experience, it minimises the possible frustration for external contributors, shortening the CI feedback loop.

The proposed workflow is:

- developer makes a change to the files and pushes a commit, either forgetting to run the code formatter or using an incorrect configuration
- the CI checks the style and, in case of a violation, pushes an additional commit that applies the code style to the violating files
- the developer, when they check the status of they CI build, see this new commit instead of a CI failure. They are not forced to swithc context and set-up the code formatting tool, and are free to carry on with their work.
- the formatting commit can be squashed when merging the PR

</details>

### Hiding the Big Formatting Commit from `git blame`

We can leverage `git blame --ignore-revs-file .git-blame-ignore-revs` to hide the formatting commit from `git blame`. GitHub seems to [support](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) this workflow out of the box in its blame view. Local `git blame` needs a little bit of setup:

```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

geo2a:
Fixes #1499 

# PR description

This PR introduces `fourmolu` as an automatic code formatter to replace `stylish-haskell`

Beyond the actual reformatting, I've introduced the following changes:

- add `fourmolu` to the Nix shell
- add the `scripts/ci/run-fourmolu.sh` scripts, which is based on the previously used script for `stylish-haskell`
- add a Hydra job to check the project with `fourmolu` (and pushed a dummy commit to check that the CI rejects improperly styled code)
- add The Big Reformatting Commit to `.git-blame-ignore-revs` and checked that the blame interface on GitHub respects it
- with a somewhat heavy heart, butchered the style guide. The layout-related sections are removed as the layout is now enforced by `fourmolu`.

Additionally, I've enabled `ImportQualifiedPost` in all cabal files, alongside with the `-Wprepositive-qualified-module` warning. This extension and warning could be removed once (if?) we transition to `GHC2021` as the default language.


nix/tools.nix
```diff
@@ -31,14 +31,14 @@ in
...
 
-  stylish-haskell = tool "stylish-haskell" "0.14.6.0" { };
-
   cabal-gild = tool "cabal-gild" "1.5.0.1" { };
 
   hlint = tool "hlint" "3.8" { };
 
   xrefcheck = tool "xrefcheck" "0.3.1" { };
 
+  fourmolu = tool "fourmolu" "0.17.0.0" { };
```
amesgen, line 40: Any particular reason for using 0.17.0.0, given that 0.18.0.0 is the latest version? This needs GHC 8.8 for building, but we already do that for eg `cabal-docspec` above.

I think that you will run into https://github.com/tweag/ormolu/issues/1166, but that is trivial to work around (just add the explicit fixity).

fourmolu.yaml
```diff
@@ -0,0 +1,6 @@
+indentation: 2
+column-limit: 100
```
amesgen, line 2: Enforcing a column limit has various idempotency issues. Usually minor, so no need to do sth immediately, but something to keep in mind if it gets too annoying.

docs/website/contents/for-developers/StyleGuide.md
amesgen, line 1: The section on `if-then-else` also seems superseded by fourmolu.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
```diff
@@ -1,3 +1,5 @@
+{- FOURMOLU_DISABLE -}
```
amesgen, line 1: Generally, Ormolu/Fourmolu require that formatting chunks (contiguous lines) of code before/after CPP chunks can still be formatted individually.

In the case of this file, it is trivial to arrange this, so we don't need to disable Fourmolu, just apply this diff:
```diff
diff --git a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
index 15b29b51d4..1deff8d080 100644
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
@@ -257,9 +257,9 @@ assertSameChunk a b
   | otherwise = throw $ NotSameChunk a b prettyCallStack
 #else
 assertSameChunk _ _ = id
-#endif
   where
     _ = keepRedundantConstraint (Proxy @HasCallStack)
+#endif
 
 assertWithinBounds :: HasCallStack => Word64 -> ChunkSize -> a -> a
 #if ENABLE_ASSERTIONS
@@ -268,9 +268,9 @@ assertWithinBounds ix sz
   | otherwise                 = throw $ NotWithinBounds ix sz prettyCallStack
 #else
 assertWithinBounds _ _ = id
-#endif
   where
     _ = keepRedundantConstraint (Proxy @HasCallStack)
+#endif
 
 assertChunkCanContainEBB :: HasCallStack => ChunkNo -> ChunkSize -> a -> a
 #if ENABLE_ASSERTIONS
@@ -279,6 +279,6 @@ assertChunkCanContainEBB chunk size
   | otherwise               = throw $ ChunkCannotContainEBBs chunk prettyCallStack
 #else
 assertChunkCanContainEBB _ _ = id
-#endif
   where
     _ = keepRedundantConstraint (Proxy @HasCallStack)
+#endif
```

nix/tools.nix
geo2a: Thanks Alex, I did not immediately realize that the it was the GHC version mismatch that caused the build failure for 0.18.0.0:

```
Resolving dependencies...
make-install-plan: dieVerbatim: user error (Error: make-install-plan: Could not resolve dependencies:
[__0] trying: fourmolu-0.18.0.0 (user goal)
[__1] trying: ghc-lib-parser-9.12.2.20250320 (dependency of fourmolu)
[__2] next goal: base (dependency of fourmolu)
[__2] rejecting: base-4.18.2.1/installed-4.18.2.1 (conflict: ghc-lib-parser =>
base>=4.19 && <4.22)
[__2] rejecting: base-4.21.0.0, base-4.20.0.1, base-4.20.0.0, base-4.19.2.0,
base-4.19.1.0, base-4.19.0.0, base-4.18.2.1, base-4.18.2.0, base-4.18.1.0,
base-4.18.0.0, base-4.17.2.1, base-4.17.2.0, base-4.17.1.0, base-4.17.0.0,
base-4.16.4.0, base-4.16.3.0, base-4.16.2.0, base-4.16.1.0, base-4.16.0.0,
base-4.15.1.0, base-4.15.0.0, base-4.14.3.0, base-4.14.2.0, base-4.14.1.0,
base-4.14.0.0, base-4.13.0.0, base-4.12.0.0, base-4.11.1.0, base-4.11.0.0,
base-4.10.1.0, base-4.10.0.0, base-4.9.1.0, base-4.9.0.0, base-4.8.2.0,
base-4.8.1.0, base-4.8.0.0, base-4.7.0.2, base-4.7.0.1, base-4.7.0.0,
base-4.6.0.1, base-4.6.0.0, base-4.5.1.0, base-4.5.0.0, base-4.4.1.0,
base-4.4.0.0, base-4.3.1.0, base-4.3.0.0, base-4.2.0.2, base-4.2.0.1,
base-4.2.0.0, base-4.1.0.0, base-4.0.0.0, base-3.0.3.2, base-3.0.3.1
(constraint from non-upgradeable package requires installed instance)
[__2] fail (backjumping, conflict set: base, fourmolu, ghc-lib-parser)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: fourmolu, ghc-lib-parser, base
```

fourmolu.yaml
geo2a: Would you rather suggest to reformat the code base and to not enforce the limit? I do not see how we would be able to prevent addition of longer lines in this case.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
geo2a: Right, I was not thinking straight yesterday and it appeared to me that the `where` clause could do something I didn't understand in this context. I do see clearly that it does not now! Thanks Alex.

geo2a: I think this CI error is absolutely hilarious:

```
src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs:52:1: error: [-Wunused-imports, Werror=unused-imports]
    The import of ‘Ouroboros.Consensus.Util.RedundantConstraints’ is redundant
      except perhaps to import instances from ‘Ouroboros.Consensus.Util.RedundantConstraints’
    To import instances alone, use: import Ouroboros.Consensus.Util.RedundantConstraints()
   |
52 | import Ouroboros.Consensus.Util.RedundantConstraints
```

geo2a: On a more serious note: as I understand, we need to either disable fourmolu or to duplicate the `where`-clause in both branches of `#if`. I think I prefer to disable fourmolu for this file.

fourmolu.yaml
amesgen: No, what you do in this PR looks good to me, just pointing out this potential downside of `column-limit`, but no immediate action required.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Internal.hs
geo2a: Sounds good, done!

CONTRIBUTING.md
```diff
@@ -203,19 +203,19 @@ and improvements are always welcome.
 
 ## Formatting the code
 
-We use `stylish-haskell` 0.14.6.0 for Haskell code formatting.
+We use `fourmolu` 0.17.0.0 for Haskell code formatting.
```
jasagredo, line 206: It is 0.18.0.0 in the end.

jasagredo (approved):
I think the final code looks good. Just a minor remark on CONTRIBUTING but overall I am fine with the result.

Issue #1430, opened 2025-03-24
[FEAT] - More thorough CDDL and Golden testing


jasagredo:
**Internal/External**

**Describe the feature you'd like**
With #1422 we have CDDL definitions for all data that consensus emits. We should implement the following tests based on them:

### Is Haskell serialization stable?

```mermaid
flowchart LR
    A[Haskell datatype] -->|define| B[Haskell Example]
    B -->|encode| C[CBOR file]
    C --> D{compare with golden file}
    D --> Ok
    D --> Fail
```

### Do Haskell serializers roundtrip?

```mermaid
flowchart LR
    A[Haskell datatype] -->|generate| B[Haskell value 1]
    B -->|encode| C[CBOR file 1]
    C -->|decode| D[Haskell value 2]
    B --> E{are equal}
    D --> E
    E --> Ok1[Ok]
    E --> Fail1[Fail]
    D -->|encode| F[CBOR file 2]
    F --> G{are equal}
    C --> G
    G --> Ok
    G --> Fail
```

### Are Haskell serializers and deserializers compliant with CDDL?

```mermaid
flowchart LR
    A[Haskell datatype] -->|generate| B[Haskell value]
    B -->|encode| C[CBOR file]
    C --> D{\`cddl validate\`}
    D --> Ok
    D --> Fail

    M[CDDL definition] --> D
    M -->|\`cddl generate\`| N[CBOR file]
    N --> O{decode}
    O --> Ok2[Ok]
    O --> Fail2[Fail]
```

PR #1501, opened 2025-05-09
Draft local state query serialization tests

ch1bo:
Experimented with how we would want to test against CDDLs in https://github.com/cardano-scaling/cardano-blueprint/ from this repository.

This is incomplete, but I'd like to gather some feedback already on:
- Ran into `UTCTime` not fully roundtripping because `dayOfYear` is clamped to `[1, 366]` - would we want that any `CBOR` that is valid against a `CDDL` to roundtrip, i.e. `CBOR -> result -> CBOR`?
- I just composed the `.cddl` files, while @jasagredo started using cddl module extensions. However that requires `cddlc` and I was not feeling like packaging that for NixOS. Do we really want/need `;# include`? If yes, is the `cddlc` gem the way to go or should we extend https://github.com/input-output-hk/cuddle?
- Similarly, validating `CBOR` against `CDDL` can only be done with the `cddl` gem, which is not open source (anymore?) and it's error output is just insane (thousands of repeated lines of errors). Do we want https://github.com/input-output-hk/cuddle/issues/7 instead?
- How are you feeling about using a git submodule for `cardano-blueprint`?


jasagredo:
> Ran into UTCTime not fully roundtripping because dayOfYear is clamped to [1, 366] - would we want that any CBOR that is valid against a CDDL to roundtrip, i.e. CBOR -> result -> CBOR?

https://github.com/IntersectMBO/ouroboros-consensus/issues/1430

> I just composed the .cddl files, while @jasagredo started using cddl module extensions. However that requires cddlc and I was not feeling like packaging that for NixOS. Do we really want/need ;# include? If yes, is the cddlc gem the way to go or should we extend https://github.com/input-output-hk/cuddle?

Concatenating is fine with simple examples, however I really doubt one can define the block CDDL keeping their sanity with only concat. I do think `cddlc` is the way to go. I do not think we should extend cuddle for this. I think it would be a waste of time re-implementing something that already exists just because we want to use our tool.

> Similarly, validating CBOR against CDDL can only be done with the cddl gem, which is not open source (anymore?) and it's error output is just insane (thousands of repeated lines of errors). Do we want
https://github.com/input-output-hk/cuddle/issues/7 instead?

I'm preparing a validator here: https://github.com/input-output-hk/cuddle/pull/68. As opposed to `cddlc`, this feature I think is worth re-implementing because of how slow `cddl` is, and its very confusing and verbose output (at least for me!).

> How are you feeling about using a git submodule for cardano-blueprint?

No objections from my side.

-------------------------------------------

One thing that I would like to note here is that GetSystemStart is presumably one of the easiest specs. I want to warn not to commit to something that works for this simple case and then becomes a horrible nightmare to adapt it to the multi thousand lines CDDLs for blocks and others.

PR #1507, opened 2025-05-14
ChainDB.getIsValid: remove caveat about blocks from the future

### LINKED ISSUE #647: Chain DB: enforce no overlap between previously applied and invalid blocks
The LgrDB maintains a set of previously applied *valid* blocks. The ChainDB maintains a map of invalid blocks.

A block from the future that is valid according to the ledger that exceeds the max clock skew will be *both* in the set of previously applied blocks (added in the LgrDB) and invalid blocks (add in the ChainDB after ledger validation).

We should have the invariant that such a block cannot be in both. Related: https://github.com/input-output-hk/ouroboros-network/pull/2402#discussion_r455161670

amesgen:
Closes #647

See the single commit's PR description


jasagredo (approved):

nfrisby (approved):
Thanks for catching that straggler.

PR #1412, opened 2025-03-05
UTxO-HD for node 10.4

jasagredo:
# Description

Implement UTxO-HD on top of the latest main.

Changelog fragments give a super-high-level description. Documentation on the website goes a bit more in depth.

PR #1512, opened 2025-05-19
`db-analyser --store-ledger`: fix two bugs

amesgen:
Suppose that `s0 < s1 < s2` are slots with blocks in the ImmutableDB, and no other slot `s` with `s0 < s < s2` has a block, and suppose that we run `db-analyser --store-ledger s2`. It is now expected that we store a ledger snapshot for slot `s2`, which was the case before #1412. However:

 - With `--v1-in-mem` or `--lmdb`, we store a ledger snapshot for slot `s1`.
 - With `--v2-in-mem`, we store a ledger snapshot for slot `s0`.

This PR fixes two unrelated bugs causing this (one specific to the V2 LedgerDB), see the individual commits. Now, in all cases, we store a ledger snapshot for slot `s2`.

Applied the `no changelog` label as only testing/internal code is affected.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
@@ -180,8 +180,8 @@ mkInternals ::
 mkInternals bss h = TestInternals {
       takeSnapshotNOW = \whereTo suff -> getEnv h $ \env -> do
           st <- (case whereTo of
-            TakeAtVolatileTip  -> anchorHandle
-            TakeAtImmutableTip -> currentHandle) <$> readTVarIO (ldbSeq env)
+            TakeAtImmutableTip -> anchorHandle
+            TakeAtVolatileTip  -> currentHandle) <$> readTVarIO (ldbSeq env)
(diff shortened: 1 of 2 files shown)


jasagredo (approved):
Good catch! I forgot about it but now I remember you mentioned this months ago, sorry 🙏

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -180,8 +180,8 @@ mkInternals ::
 mkInternals bss h = TestInternals {
```
geo2a, line 180: What do you mean by "this is a test-only function"? I went up the chain that calls `mkInternals` and, if I'm not mistaken, it is transitively called from `LedgerDB.openDB` which is the primary entry point into the LedgerDB.

jasagredo: It is created there, but the internals are ignored on production. See `f` in:
```
doOpenDB args initDb stream replayGoal =
    f <$> openDBInternal args initDb stream replayGoal
  where f (ldb, replayCounter, _) = (ldb, replayCounter)
```

Issue #1424, opened 2025-03-17
[FEAT] - Snapshot every Nth block


jasagredo:
**Internal**

**Describe the feature you'd like**
In order to make sure that multiple nodes take the same snapshots, it would be useful to implement an alternative snapshotting policy that decides based on the block number. I propose something like:
```
// currently
{ 
  "SnapshotInterval": 5000       // seconds
}

// proposal
{ 
  "SnapshotInterval": "5000s"       // seconds
}
{ 
  "SnapshotInterval": "100b"        // every @blockno `mod` 100 == 0@
}
```

We would have to pass to `onDiskShouldTakeSnapshot` also the current block number.

amesgen:
We talked about this today with @jpraynaud. It will most likely make sense to create snapshots based on the *slot* number, namely near the end of chunk files (encompassing `10*k = 21600` slots each), such that one can simply take the (complete) ImmutableDB chunks up till this point.

Additionally, it makes sense to give these snapshots a special suffix such that they are not deleted by the node istelf. Rather, it will be the job of the Mithril signer to delete them.

amesgen:
FTR: I created separate tickets for the two related refinements that we also discussed just now, such that this ticket purely tracks the aspect of predictable snapshots:

 - #1573
 - #1574

PR #1513, opened 2025-05-19
LedgerDB: prune on garbage collection instead of on every change

amesgen:
This is in preparation for #1424

Currently, we prune the LedgerDB (ie remove all but the last `k+1` states) every time we adopt a longer chain. This means that we can not rely on the fact that other threads (like the `copyAndSnapshot` ChainDB background) actually observe all immutable ledger states, just as described in the caveats of our `Watcher` abstraction.

However, a predictable ledger snapshotting rule (#1424) requires this property; otherwise, when the node is under high load and/or we are adopting multiple blocks in quick succession, the node might not be able to create a snapshot for its desired block.

This PR changes this fact: Now, when adopting new blocks, the LedgerDB is *not* immediately pruned. Instead, the a new dedicated background thread for ledger maintenance tasks (flushing/snapshotting/garbage collection) in the ChainDB will periodically (on every new immutable block) wake up and (in particular) garbage collect the LedgerDB based on a slot number.

Also, this makes the semantics more consistent with the existing garbage collection of previously-applied blocks in the LedgerDB, and also with how the ChainDB works, where we also don't immediately delete blocks from the VolatileDB once they are buried beneath `k+1` blocks.

See https://github.com/IntersectMBO/ouroboros-consensus/pull/1513#issuecomment-2944658608 for benchmarks demonstrating that the peak memory usage does not increase while syncing (where we now briefly might hold more than `k+1` ledger states in memory).

### DIFF
--- a/ouroboros-consensus/changelog.d/20250626_193647_alexander.esgen_ledgerdb_garbage_collect_states.md
+++ b/ouroboros-consensus/changelog.d/20250626_193647_alexander.esgen_ledgerdb_garbage_collect_states.md
@@ -1,0 +1,9 @@
+### Breaking
+
+- Changed pruning of immutable ledger states to happen on LedgerDB garbage
+  collection instead of directly on every block adoption. This is purely an
+  internal refactoring (with breaking API changes) supporting predictable ledger
+  snapshotting.
+
+- Avoid maintaining volatile ledger states during ledger replay, making it
+  slightly more efficient.
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
@@ -307,8 +307,8 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed fuse = do
 garbageCollect :: forall m blk. IOLike m => ChainDbEnv m blk -> SlotNo -> m ()
 garbageCollect CDB{..} slotNo = do
   VolatileDB.garbageCollect cdbVolatileDB slotNo
+  LedgerDB.garbageCollect cdbLedgerDB slotNo
   atomically $ do
-    LedgerDB.garbageCollect cdbLedgerDB slotNo
     modifyTVar cdbInvalid $ fmap $ Map.filter ((>= slotNo) . invalidBlockSlotNo)
   traceWith cdbTracer $ TraceGCEvent $ PerformedGC slotNo
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
@@ -19,11 +20,10 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.Background
     launchBgTasks
 
     -- * Copying blocks from the VolatileDB to the ImmutableDB
-  , copyAndSnapshotRunner
   , copyToImmutableDB
 
     -- * Executing garbage collection
-  , garbageCollect
+  , garbageCollectBlocks
 
     -- * Scheduling garbage collections
   , GcParams (..)
@@ -76,6 +76,7 @@ import qualified Ouroboros.Consensus.Storage.VolatileDB as VolatileDB
 import Ouroboros.Consensus.Util
 import Ouroboros.Consensus.Util.Condense
 import Ouroboros.Consensus.Util.IOLike
+import Ouroboros.Consensus.Util.STM (Watcher (..), forkLinkedWatcher)
 import Ouroboros.Network.AnchoredFragment (AnchoredSeq (..))
 import qualified Ouroboros.Network.AnchoredFragment as AF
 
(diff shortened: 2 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -207,8 +210,9 @@ mkInternals bss h =
         eFrk <- newForkerAtTarget h reg VolatileTip
         case eFrk of
           Left{} -> error "Unreachable, Volatile tip MUST be in LedgerDB"
-          Right frk ->
+          Right frk -> do
             forkerPush frk st >> atomically (forkerCommit frk) >> forkerClose frk
+            getEnv h pruneLedgerSeq
```
jasagredo, line 215: I don't see why we are still pruning in here? I thought we would only prune at garbage collection, which we do below. Couldn't a snapshot be started in between this line and the previous one and then we would close the handle while taking the snapshot?

jasagredo: Ah wait, if we commit the forker, and we try to take a snapshot, then we will take the snapshot of the committed imm tip, which will not be pruned ever

jasagredo: Still this is confusing, because we prune on every change it seems, and the title of the PR says precisely that we prune on GC only.

jasagredo: Ah nevermind. This is part of the testing internals, I was missing that.

jasagredo (approved):
Looks good.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -775,10 +788,11 @@ type LedgerSupportsLedgerDB blk =
...
--- we are indeed using
---   1. @0@ in places where it is necessary
---   2. the security parameter as is, in other places
-data LedgerDbPrune = LedgerDbPruneAll | LedgerDbPruneKeeping SecurityParam
+data LedgerDbPrune
+  = -- | Prune all states, keeping only the current tip.
+    LedgerDbPruneAll
+  | -- | Prune to only keep the last @k@ states.
+    LedgerDbPruneKeeping SecurityParam
+  | -- | Prune such that all (non-anchor) states are older than the given slot.
```
jasagredo, line 796:
```suggestion
  | -- | Prune such that all (non-anchor) states are younger than the given slot.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/DbChangelog.hs
```diff
@@ -47,8 +47,8 @@
 -- == Carrying states
 --
 -- The 'DbChangelog' contains an instantiation of the 'AnchoredSeq' data type to
--- hold the last \(k\) in-memory ledger states. This data type is impemented
--- using the /finger tree/ data structure and has the following time
+-- hold (at least) the last \(k\) in-memory ledger states. This data type is
+-- impemented using the /finger tree/ data structure and has the following time
```
jasagredo, line 51:
```suggestion
-- implemented using the /finger tree/ data structure and has the following time
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/DbChangelog.hs
```diff
@@ -402,6 +399,14 @@ prune LedgerDbPruneAll dblog =
 
   vol' =
     snd $ AS.splitAt nvol changelogStates
+prune (LedgerDbPruneBeforeSlot slot) dblog =
+  dblog{changelogStates = vol'}
+ where
+  DbChangelog{changelogStates} = dblog
+
+  -- The anchor of @vol'@ might still have a tip slot larger than @slot@, which
```
jasagredo, line 407:
```suggestion
  -- The anchor of @vol'@ might still have a tip slot smaller than @slot@, which
```

jasagredo: or `older`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -255,6 +256,12 @@ prune howToPrune (LedgerSeq ldb) = case howToPrune of
     (closeButHead before, LedgerSeq after)
    where
     (before, after) = (ldb, AS.Empty (AS.headAnchor ldb))
+  LedgerDbPruneBeforeSlot slot ->
+    (closeButHead before, LedgerSeq after)
+   where
+    -- The anchor of @vol'@ might still have a tip slot larger than @slot@,
```
jasagredo, line 262:
```suggestion
    -- The anchor of @vol'@ might still have a tip slot older than @slot@,
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
amesgen: Thanks for pointing out this inversion in a couple of places!

Changed, using "not older" instead of "younger" to account for the case of equality (we want to keep states with the same slot as the argument slot).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -265,8 +270,12 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed fuse = do
     --
     -- This is a synchronous operation: when it returns, the blocks have been
     -- copied to disk (though not flushed, necessarily).
-    withFuse fuse (copyToImmutableDB cdb) >>= scheduleGC'
+    gcSlotNo <- withFuse fuse (copyToImmutableDB cdb)
+    scheduleGC' gcSlotNo
 
+    -- See the Haddocks above as for why we garbage-collect the LedgerDB already
+    -- here (instead of as part of the scheduled GC).
+    whenJust (withOriginToMaybe gcSlotNo) $ LedgerDB.garbageCollect cdbLedgerDB
```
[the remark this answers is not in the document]
geo2a: oops, now I see that chainging this is the entire point of e4143672b!! Please disregard my previous comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -251,9 +251,15 @@ data LedgerDB m l blk = LedgerDB
   -- back as many blocks as the passed @Word64@.
   , getPrevApplied :: STM m (Set (RealPoint blk))
   -- ^ Get the references to blocks that have previously been applied.
-  , garbageCollect :: SlotNo -> STM m ()
-  -- ^ Garbage collect references to old blocks that have been previously
-  -- applied and committed.
+  , garbageCollect :: SlotNo -> m ()
+  -- ^ Garbage collect references to old state that is older than the given
```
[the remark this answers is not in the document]
amesgen: "State" was intended to be a generic term here, as garbage collection does not just clear old ledger states, but also old previously-applied points, and it feels slightly weird to describe these as "states".

amesgen:
Thanks for the great reviews, I hope I addressed your comments. Interesting changes:

 - d627e843fd712e3aa725c1843b65939b7bea6295 for removing `LedgerDbPruneKeeping` prompted by https://github.com/IntersectMBO/ouroboros-consensus/pull/1513#discussion_r2179421993
 - Found a bug in the LedgerDB state machine test (#1576) which this PR now depends on as I enriched it due to https://github.com/IntersectMBO/ouroboros-consensus/pull/1513#discussion_r2174925570

ouroboros-consensus/changelog.d/20250626_193647_alexander.esgen_ledgerdb_garbage_collect_states.md
```diff
@@ -0,0 +1,6 @@
+### Breaking
+
+- Changed pruning of immutable ledger states to happen on LedgerDB/ChainDB
```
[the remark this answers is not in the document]
amesgen: FTR we discussed this in the sync today, and decided that a new ChainDB background thread makes sense.

PR #1516, opened 2025-05-19
LedgerDB.V2: make sure to actually close handles

### LINKED ISSUE #1515: LedgerDB V2: handles aren't closed in the common case
The V2 LedgerDB (introduced as part of UTxO HD in #1412) maintains a `LedgerSeq` at its core:
https://github.com/IntersectMBO/ouroboros-consensus/blob/d582af5495a5a0a03c889841fb5c1e02e24b8ce3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs#L140-L142

https://github.com/IntersectMBO/ouroboros-consensus/blob/d582af5495a5a0a03c889841fb5c1e02e24b8ce3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs#L119-L122

Now, when we `prune` a `LedgerSeq`, we want to `close` the `LedgerTablesHandle`s from the prefix that we are removing. The current code *tries* to do this, it it is not effective. Namely, `prune` returns the `LedgerSeq` prefix containing states to prune:

https://github.com/IntersectMBO/ouroboros-consensus/blob/d582af5495a5a0a03c889841fb5c1e02e24b8ce3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs#L230-L238

And then in other places, we try to `close` the handles:

https://github.com/IntersectMBO/ouroboros-consensus/blob/d582af5495a5a0a03c889841fb5c1e02e24b8ce3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs#L154-L155
https://github.com/IntersectMBO/ouroboros-consensus/blob/d582af5495a5a0a03c889841fb5c1e02e24b8ce3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs#L163
https://github.com/IntersectMBO/ouroboros-consensus/blob/2c064714edc82636867ea8927faedbde5bec4aad/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs#L186-L187

The crucial observation is that `toOldestFirst` does **not** return the anchor of the `LedgerSeq`/`AnchoredSeq`, so we do not close this. In particular, in the very common case of adopting one block (such that the immutable tip also advances by one block), which is the common case both while syncing and caught-up, we do not close anything.

The fix is simply to also consider the anchor of the sequence of handles to close.

---

Currently, this isn't problematic as the only V2 backend is the in-memory backend at the moment, and for it, closing handles doesn't matter (as the GC will still claim the ledger state once it is no longer referenced). However, in the upcoming LSM tree backend, this will be crucial.

---

Concrete "demonstration" of the bug:

Apply this diff on d582af5495a5a0a03c889841fb5c1e02e24b8ce3 (`main` at time of writing):
<details>
<summary>Show diff</summary>

```diff
diff --git a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
index 770a85862d..c4e2ac3115 100644
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
@@ -90,7 +90,8 @@ newInMemoryLedgerTablesHandle ::
 newInMemoryLedgerTablesHandle someFS@(SomeHasFS hasFS) l = do
   !tv <- newTVarIO (LedgerTablesHandleOpen l)
   pure LedgerTablesHandle {
-      close =
+      close = do
+        !_ <- error "foo"
         atomically $ writeTVar tv LedgerTablesHandleClosed
     , duplicate = do
         hs <- readTVarIO tv
diff --git a/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs b/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
index 5015d44106..33f842b615 100644
--- a/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
+++ b/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
@@ -270,14 +270,14 @@ instance StateModel Model where
   arbitraryAction _ model@(Model chain secParam) =
     frequency $ [ (2, pure $ Some GetState)
                 , (2, pure $ Some ForceTakeSnapshot)
-                , (1, Some . DropAndRestore <$> QC.choose (0, fromIntegral $ AS.length chain))
+                , (0, Some . DropAndRestore <$> QC.choose (0, fromIntegral $ AS.length chain))
                 , (4, Some <$> do
                       let maxRollback =
                             min
                               (fromIntegral . AS.length $ chain)
                               (BT.unNonZero $ maxRollbacks secParam)
-                      numRollback  <- QC.choose (0, maxRollback)
-                      numNewBlocks <- QC.choose (numRollback, numRollback + 2)
+                      numRollback  <- pure 0
+                      numNewBlocks <- pure 1
                       let
                         chain' = case modelRollback numRollback model of
                           UnInit     -> error "Impossible"
```

</details>

Then, running `cabal run storage-test -- -p InMemV2` succeeds, even though we would expect it to fail if we would actually close the handles.

Concretely, this patch

 - modifies the close logic of the `LedgerTablesHandle` to always fail with an imprecise exception, and
 - modifies the LedgerDB to only generate `ValidateAndCommit` commands that add exactly one block, and disables `DropAndRestore`.

Then, if V2 were correctly closing pruned LedgerTablesHandles, then we would expect the test to fail; however, it passes just fine. Note that it *does* fail when keeping the test unmodified, showing that we at least *sometimes* close `LedgerTablesHandle`s.

amesgen:
Closes #1515 

The first three commits add a regression test by enrichting the existing LedgerDB state machine test, making sure that
```
  total number of all opened handles
- total number of closed handles
= states on the LedgerSeq in the end
```
The fourth commit then fixes the bug, causing the regression test to succeed.

Note that this bug fix reveals another bug #1551 (causing the `mock-test` failure for `LeaderSchedule.simple convergence`), which we silence by temporarily making the closing of an in-memory handle a no-op (as it is not *necessary* anyways, the GC will still collect the in-memory ledger state).

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs
@@ -143,7 +143,7 @@ implForkerCommit env = do
   LedgerSeq lseq <- readTVar foeLedgerSeq
   let intersectionSlot = getTipSlot $ state $ AS.anchor lseq
   let predicate = (== getTipHash (state (AS.anchor lseq))) . getTipHash . state
-  (discardedBySelection, LedgerSeq discardedByPruning) <- do
+  closeDiscarded <- do
     stateTVar
       foeSwitchVar
       ( \(LedgerSeq olddb) -> fromMaybe theImpossible $ do
@@ -153,17 +153,25 @@ implForkerCommit env = do
           -- Join the prefix of the selection with the sequence in the forker
           newdb <- AS.join (const $ const True) olddb' lseq
           -- Prune the resulting sequence to keep @k@ states
-          let (l, s) = prune (LedgerDbPruneKeeping (foeSecurityParam env)) (LedgerSeq newdb)
-          pure ((toClose, l), s)
+          let (closePruned, s) = prune (LedgerDbPruneKeeping (foeSecurityParam env)) (LedgerSeq newdb)
+              closeDiscarded = do
+                closePruned
+                -- Do /not/ close the anchor of @toClose@, as that is also the
+                -- tip of @olddb'@ which will be used in @newdb@.
+                case toClose of
+                  AS.Empty _ -> pure ()
+                  _ AS.:< closeOld' -> closeLedgerSeq (LedgerSeq closeOld')
+                -- Finally, close the anchor of @lseq@ (which is a duplicate of
+                -- the head of @olddb'@).
+                close $ tables $ AS.anchor lseq
+          pure (closeDiscarded, s)
       )
 
   -- We are discarding the previous value in the TVar because we had accumulated
   -- actions for closing the states pushed to the forker. As we are committing
   -- those we have to close the ones discarded in this function and forget about
   -- those releasing actions.
-  writeTVar foeResourcesToRelease $
-    mapM_ (close . tables) $
-      AS.toOldestFirst discardedBySelection ++ AS.toOldestFirst discardedByPruning
+  writeTVar foeResourcesToRelease closeDiscarded
  where
   ForkerEnv
     { foeLedgerSeq
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
@@ -96,7 +96,11 @@ newInMemoryLedgerTablesHandle tracer someFS@(SomeHasFS hasFS) l = do
   pure
     LedgerTablesHandle
       { close = do
-          atomically $ writeTVar tv LedgerTablesHandleClosed
+          -- Temporarily a no-op until
+          -- https://github.com/IntersectMBO/ouroboros-consensus/issues/1551 has
+          -- been fixed.
+
+          -- atomically $ writeTVar tv LedgerTablesHandleClosed
           traceWith tracer V2.TraceLedgerTablesHandleClose
       , duplicate = do
           hs <- readTVarIO tv
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
@@ -51,6 +51,7 @@ import Ouroboros.Consensus.Ledger.SupportsProtocol
 import qualified Ouroboros.Consensus.Ledger.Tables.Diff as Diff
 import Ouroboros.Consensus.Storage.LedgerDB.API
 import Ouroboros.Consensus.Storage.LedgerDB.Snapshots
+import qualified Ouroboros.Consensus.Storage.LedgerDB.V2.Args as V2
 import Ouroboros.Consensus.Storage.LedgerDB.V2.LedgerSeq
 import Ouroboros.Consensus.Util.CBOR (readIncremental)
 import Ouroboros.Consensus.Util.CRC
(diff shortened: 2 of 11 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs
```diff
@@ -153,17 +153,25 @@ implForkerCommit env = do
...
+              closeDiscarded = do
+                closePruned
+                -- Do /not/ close the anchor of @toClose@, as that is also the
+                -- tip of @olddb'@ which will be used in @newdb@.
+                case toClose of
+                  AS.Empty _ -> pure ()
+                  _ AS.:< closeOld' -> closeLedgerSeq (LedgerSeq closeOld')
+                -- Finally, close the anchor of @lseq@ (which is a duplicate of
+                -- the head of @olddb'@).
+                close $ tables $ AS.anchor lseq
```
amesgen, line 166: I *think* this is correct, but it is rather non-obvious.

It would be "nicer" if `newdb` would use the anchor state of `lseq` instead of the head state of `olddb'` (but we would need a special purpose `AS.join` for this as this isn't possible in general when the anchor is "strictly smaller" than the element type). In that case, we could simply write
```haskell
closeDiscarded = closePruned *> closeLedgerSeq (LedgerSeq closeOld)
```
instead of what I have here.

jasagredo (approved):
Very nice catch, Alex. The fix looks very reasonable.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
```diff
@@ -85,18 +86,21 @@ newInMemoryLedgerTablesHandle ::
...
   , SerializeTablesWithHint l
   ) =>
+  Tracer m V2.FlavorImplSpecificTrace ->
   SomeHasFS m ->
   LedgerTables l ValuesMK ->
   m (LedgerTablesHandle m l)
-newInMemoryLedgerTablesHandle someFS@(SomeHasFS hasFS) l = do
+newInMemoryLedgerTablesHandle tracer someFS@(SomeHasFS hasFS) l = do
   !tv <- newTVarIO (LedgerTablesHandleOpen l)
+  traceWith tracer V2.TraceLedgerTablesHandleCreate
```
jasagredo, line 95: I seem to recall we already had these kind of traces.

jasagredo: Ah actually we have "some" datatype for traces but we don't emit it anywhere as it is useless.
```haskell
data FlavorImplSpecificTrace
  = FlavorImplSpecificTraceInMemory
  | FlavorImplSpecificTraceOnDisk
  deriving (Show, Eq)
```

Perahaps we can even delete it...

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -462,19 +468,20 @@ openLedgerDB ::
...
   LedgerDbCfg (ExtLedgerState TestBlock) ->
   SomeHasFS IO ->
-  IO (LedgerDB' IO TestBlock, TestInternals' IO TestBlock)
+  IO (LedgerDB' IO TestBlock, TestInternals' IO TestBlock, IO NumOpenHandles)
 openLedgerDB flavArgs env cfg fs = do
   (stream, volBlocks) <- dbStreamAPI (ledgerDbCfgSecParam cfg) env
   let getBlock f = Map.findWithDefault (error blockNotFound) f <$> readTVarIO (dbBlocks env)
   replayGoal <- fmap (realPointToPoint . last . Map.keys) . atomically $ readTVar (dbBlocks env)
   rr <- unsafeNewRegistry
+  (tracer, getNumOpenHandles) <- mkTrackOpenHandles
```
jasagredo, line 477: Ah very nice relying on the trace to count the handles 👌

jasagredo (approved):
No negative comments. I think we can merge this if you want.

PR #1525, opened 2025-05-27
Integrate newer cardano-ledger version

jasagredo:
This needs a new NTC version in ouroboros-network, and also a new release of cardano-ledger (see the commit in the SRP)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query/Types.hs
```diff
@@ -77,3 +98,442 @@ fromLedgerPoolDistr pd =
   PoolDistr
     { unPoolDistr = Map.map fromLedgerIndividualPoolStake $ SL.unPoolDistr pd
     }
+
+--------------------------------------------------------------------------------
+-- LegacyPParams
```
amesgen, line 103: Maybe move this to a separate module as before #1356?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query/Types.hs
```diff
@@ -77,3 +98,442 @@ fromLedgerPoolDistr pd =
...
+--
+-- After node 10.5, Ledger started encoding the protocol version in a TList
+-- instead of as two contiguous terms. These codecs respect the old version.
+--
+-- This can be deleted once we cross the Dijkstra HF
```
amesgen, line 108: Not necessarily Dijkstra, the next intra-era HF would also suffice

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Simple.hs
```diff
@@ -42,12 +42,12 @@ instance SerialiseNodeToClient blk TriggerHardFork where
   encodeNodeToClient _ _ triggerHardFork = case triggerHardFork of
     TriggerHardForkAtVersion v -> encodeListLen 2 <> encodeWord8 0 <> toCBOR v
     TriggerHardForkAtEpoch e -> encodeListLen 2 <> encodeWord8 1 <> toCBOR e
-    TriggerHardForkNotDuringThisExecution -> encodeListLen 2 <> encodeWord8 2
+    TriggerHardForkNotDuringThisExecution -> encodeListLen 1 <> encodeWord8 2
```
amesgen, line 45: Just confirming: Doing this fix without special backwards-compat dances is fine because it is only used for `DebugLedgerConfig`?

jasagredo: Indeed :+1:

lehins (commented):
Looks good :+1:

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -370,6 +388,7 @@ instance
       , (NodeToClientV_18, CardanoNodeToClientVersion14)
       , (NodeToClientV_19, CardanoNodeToClientVersion15)
       , (NodeToClientV_20, CardanoNodeToClientVersion16)
+      , (NodeToClientV_21, CardanoNodeToClientVersion17)
```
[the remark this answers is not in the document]
jasagredo: I don't really know what we should do here because NodeToClientV_21 is purely a stub I created in a branch in network. A new release of network will probably officially provide NodeToClientV_21 so I'm a bit hesitant on making this the last released version now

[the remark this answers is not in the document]
amesgen: Right, so we bump it when we remove the s-r-p due to Networking having made a release? That reasonable (but I would also be fine with bumping this now), probably useful to record that somewhere so we don't forget.

PR #1542, opened 2025-05-30
Reorganize our documentation

dnadales:
This PR reorganizes the Consensus documentation, integrating information from various sources such as the Consensus report, Haddock comments, and existing documentation. The documentation is structured using the Diátaxis framework. Implementation-specific concepts will be moved to the [Cardano Blueprints](https://cardano-scaling.github.io/cardano-blueprint/).


nfrisby:
The added text looks good and nice to me.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -106,3 +108,32 @@ The `additionalEnvelopeChecks` for the `HardForkBlock` ensures that the block's
...
+The [`HeaderState`](https://github.com/intersectmbo/ouroboros-consensus/blob/c573f0639584623bd143f39e722340e412859aa1/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HeaderValidation.hs#L185) is defined as a data structure containing:
+
+- The **chain tip**, which includes:
+  - Slot number
+  - Block number
+  - Additional `TipInfo` specific to the block type (e.g., whether it's an Epoch Boundary Block in Byron)
+- The **chain-dependent state** of the consensus protocol ([`ChainDepState`](#consensusprotocol)), which:
+  - Is protocol-specific
+  - Is updated with new block headers
+  - Supports rollback
```
amesgen, line 139: What does "Supports rollback" mean?

dnadales: Does [this](https://github.com/IntersectMBO/ouroboros-consensus/pull/1542/commits/010b374c54f4d2f485ab114f702db6ec3b7a8f95) make sense?

amesgen: Yes :+1: ("Supports rollback" sounded a bit like there was something inherent about `ChainDepState` that makes it rollback-able.)

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+This allows a stake pool operator to replace a compromised hot key with a new one and still have their blocks take precedence.
+
+- 2. **VRF Tiebreaker**:
+If the `opcert` check is inconclusive (which is common when two pools are elected in the same slot), the chain with the lower VRF value at its tip is preferred.
+This avoids always picking the first block to arrive, which could encourage centralization to reduce latency (see ["The Frankfurt Problem"](https://github.com/IntersectMBO/ouroboros-consensus/blob/40d77fdd74a9b2b2a1d112a0b836b5cb8026c88c/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs#L227)). The VRF value used for tiebreaking (non-range extended) is uncorrelated to the leader VRF value and typically results in a uniformly random decision. Depending on the `ChainOrderConfig` for Praos, there are two [flavors](https://github.com/intersectmbo/ouroboros-consensus/blob/010b374c54f4d2f485ab114f702db6ec3b7a8f95/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs#L75) that determine when this tie-breaker comparison takes place.
```
nfrisby, line 209: centralization -> geographic centralization?

centralization -> network centralization?

centralization -> physical collocation?

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+    - `UnrestrictedVRFTiebreaker`: With this flavor, VRF tiebreakers are always compared. This has been the standard behavior for all eras before Conway.
+    - `RestrictedVRFTiebreaker`: This flavor restricts the VRF tiebreaker comparison to situations where the slot numbers of the chain tips differ by at most a specified maximum distance (`maxDist`).
+The primary motivation for this restriction is to favor blocks that were diffused earlier (in earlier slots) over those diffused later, even if the later block has a "better" VRF tiebreaker value. This aims to mitigate [issues](https://github.com/IntersectMBO/ouroboros-network/issues/2913) caused by poorly configured or resource-constrained pools that might diffuse blocks later.
```
nfrisby, line 212: > The primary motivation for this restriction is to favor blocks that were diffused earlier (in earlier slots) over those diffused later

The "in earlier slots" clarification seems wrong to me, but I haven't thought about this in a while.

- Suppose I'm elected in X and somebody else is elected in Y, such that Y > X + maxDist.
- If a node selects the other person's block before the receive mine, then they would not switch to mine even though mine is from an earlier slot.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+Consider the following select views where chains `A`, `B`, and `C` have the same length:
+
+|                 | A | B | C |
+|-----------------|---|---|---|
+| Issuer          | x | y | x |
+| Slot            | 0 | i | 0 |
+| `opcert` number | 2 | j | 1 |
+| VRF             | 3 | 2 | 1 |
+
+Lower-case letters stand for arbitrary values (two letters designate the same value if and only if they are same letter).
```
nfrisby, line 227: > two letters designate the same value if and only if they are same letter

It's an irrelevant distraction, but it'd be harmless if different letters in different rows had the same value.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+
+In this example we have:
+
+- `B` is preferred over `A`, since `B` has lower VRF than `A`.
+- `C` is preferred over `B`, since `C` has lower VRF than `B`.
+- However `C` is **not** preferred over `A`, since they have the same issuer and slot, and therefore we prefer the chain with the highest `opcert` number (2), therefore `A` is preferred over `C`.
```
nfrisby, line 233: If you have the same issuer and the same slot, then the VRF value must be equivalent.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+Also, the `RestrictedVRFTiebreaker` flavour breaks the transitivity of chain ordering. To see this consider the following example where chains `D`, `E`, and `F` have the same length and different issuers, and assume `maxDist = 5` slots:
+
+|              | D | E | F |
+| ------------ | - | - | - |
+| Slot         | 0 | 3 | 6 |
+| VRF          | 3 | 2 | 1 |
+
+
+We have that:
+- `E` is preferred over `D`, since `E` has lower VRF than `D` and `|0 - 3| < 5`.
```
nfrisby, line 244: If the name is `maxFoo`, then I assume that's an _inclusive_ upper bound, and so it should be `<=5` instead of `<5`.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+- `F` is preferred over `E`, since `F` has lower VRF than `E` and `|3 - 6| < 5`
+- However, `D` is **not** preferred over `F`, but instead `D` and `F` are equally preferred since `|0 - 6| > 5` which implies that the VRF values of `D` and `F` are not used.
+
+Despite the non-transitivity, the fundamental Consensus properties, such as [Common Prefix](TODO-ref!), are not affected. This is because the primary factor for chain selection for a [caught-up](TODO-ref!) node remains the chain length, with longer chains always being preferred. However, a non-transitive chain ordering brings the following complications:
+
+- **Implementation**: The use of `sortBy` from base in `chain` selection, which is typically expected to work with transitive relations, could become a concern. While preliminary property tests suggest it works for the current non-transitive order, there's a theoretical risk that future GHC implementations might interact non-trivially.
```
nfrisby, line 250: from base -> from [`base`](https://hackage.haskell.org/package/base)

theoretical -> potential

interact non-trivially -> behave in surprising ways

docs/website/contents/explanation/consensus-protocol.md
dnadales: From the three options above, I'd go with either "geographic centralization" or "physical collocation". I think it's better to mention centralization since the contrast with the aspiration of desentralization is more evident (or immediate).

docs/website/contents/explanation/consensus-protocol.md
dnadales: Thanks. For the last suggestion, I used "might behave in unexpected ways". Let me know if you prefer your version.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+- **Reasoning**: The non-transitive order can be very confusing for reasoning about anything related to chain order, as transitivity is an implicitly assumed property of orders.
```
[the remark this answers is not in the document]
dnadales: What about:

> The non-transitive order can be very confusing for reasoning about anything related to chain order, as transitivity is an implicitly assumed property when working with order relations.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -176,4 +191,79 @@ Although headers are already validated by the `ChainSync` client, they are re-va
...
+    - `UnrestrictedVRFTiebreaker`: With this flavor, VRF tiebreakers are always compared. This has been the standard behavior for all eras before Conway.
+    - `RestrictedVRFTiebreaker`: This flavor restricts the VRF tiebreaker comparison to situations where the slot numbers of the chain tips differ by at most a specified maximum distance (`maxDist`).
+The primary motivation for this restriction is to favor blocks that were diffused earlier over those diffused later, even if the later block has a "better" VRF tiebreaker value. This aims to mitigate [issues](https://github.com/IntersectMBO/ouroboros-network/issues/2913) caused by poorly configured or resource-constrained pools that might diffuse blocks later.
+
+### On the Transitivity of Praos Chain Order
+
+Praos chain ordering is **not transitive**, regardless of the VRF tiebreaker flavor.
```
amesgen, line 216: This is technically true because of https://github.com/IntersectMBO/ouroboros-consensus/issues/1075#issuecomment-3035911537, but this is just a bug that should be fixed.

I think it is important to mention two aspects of the chain order here:

 - The `Ord` instance on `PraosChainSelectView` must be a total order (and in particular transitive). The fact that is currently isn't in a weird edge case (#1075) is just a bug.
 - `preferCandidate` is *not* transitive when using `RestrictedVRFTiebreaker`, and this is completely fine.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -176,4 +191,79 @@ Although headers are already validated by the `ChainSync` client, they are re-va
...
+We have that:
+- `E` is preferred over `D`, since `E` has lower VRF than `D` and `|0 - 3| <= 5`.
+- `F` is preferred over `E`, since `F` has lower VRF than `E` and `|3 - 6| <= 5`
+- However, `D` is **not** preferred over `F`, but instead `D` and `F` are equally preferred since `|0 - 6| > 5` which implies that the VRF values of `D` and `F` are not used.
+
+Despite the non-transitivity, the fundamental Consensus properties, such as [Common Prefix](TODO-ref!), are not affected. This is because the primary factor for chain selection for a [caught-up](TODO-ref!) node remains the chain length, with longer chains always being preferred.
+The Ouroboros Praos authors make no assumptions about how nodes resolve ties.
+However, a non-transitive chain ordering brings the following complications:
+
+- **Implementation**: The use of `sortBy` from [`base`](https://hackage.haskell.org/package/base) in chain selection, which is typically expected to work with transitive relations, could become a concern. While preliminary property tests suggest it works for the current non-transitive order, there's a potential risk that future GHC implementations might behave in unexpected ways.
```
amesgen, line 252: This is no longer a concern due to #1063, also see https://github.com/IntersectMBO/ouroboros-consensus/pull/891#issuecomment-2061575181

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -176,4 +191,79 @@ Although headers are already validated by the `ChainSync` client, they are re-va
...
+- **Reasoning**: The non-transitive order can be very confusing for reasoning about anything related to chain order, as transitivity is an implicitly assumed property when working with order relations.
+This, in turn, leads to "obvious" properties failing to hold. For instance, the expectation that observing a node's selection over time yields a strictly improving sequence may not hold, as different observers could disagree on whether each selection is "strictly better" than the previous one. This non-objectivity can have practical effects, particularly for diffusion pipelining, which relies on a clear, consistent chain order.
+- **Potential for Cycles**: The non-transitivity can conceptually give rise to "cycles" in preference, such as `A < B < C = A`. However, in practice, the node's implementation guarantees that it will never end up changing its selection in such a cycle because blocks already in the `VolatileDB` are not added again.
```
amesgen, line 255: It might be useful to mention here that the "Consistency with `Ord`" requirement of the `ChainOrder` class guarantees that we never can have `preferCandidate A B`, `preferCandidate B C` *and* `preferCandidate C A` (so no "cycles of preference").

docs/website/contents/explanation/consensus-protocol.md
amesgen: I am not actually sure what exactly you want to demonstrate here, given that you are not using the example from https://github.com/IntersectMBO/ouroboros-consensus/issues/1075#issuecomment-3035911537. It (or variants thereof) are the only source of non-transitivity of `UnrestrictedVRFTiebreaker`.

docs/website/contents/explanation/consensus-protocol.md
```diff
@@ -0,0 +1,267 @@
...
+| Slot         | 0 | 3 | 6 |
+| VRF          | 3 | 2 | 1 |
+
+
+We have that:
+- `E` is preferred over `D`, since `E` has lower VRF than `D` and `|0 - 3| < 5`.
+- `F` is preferred over `E`, since `F` has lower VRF than `E` and `|3 - 6| < 5`
+- However, `D` is **not** preferred over `F`, but instead `D` and `F` are equally preferred since `|0 - 6| > 5` which implies that the VRF values of `D` and `F` are not used.
+
+Despite the non-transitivity, the fundamental Consensus properties, such as [Common Prefix](TODO-ref!), are not affected. This is because the primary factor for chain selection for a [caught-up](TODO-ref!) node remains the chain length, with longer chains always being preferred. However, a non-transitive chain ordering brings the following complications:
```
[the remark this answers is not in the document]
amesgen: > However, if I recall correctly, the rules that the paper proves the security argument for does use a specific tiebreaker: favor first to arrive.

The proofs in the [Praos paper](https://eprint.iacr.org/2017/573.pdf) actually don't assume anything about the tiebreaker (so in particular work when the ties are broken by the adversary):

> Function $\mathrm{maxvalid}(C, C)$: Returns the longest chain from $`C \cup \{C\}`$. Ties are broken in favor of $C$, if it has maximum length, or arbitrarily otherwise.

(There are other papers that assume specific tiebreakers, eg [this one](https://eprint.iacr.org/2017/241.pdf) which proves security in systems where there only are either empty or multi-leader slots.)

fraser-iohk (commented):
all the documentation changes look good (modulo typos), I've also left a comment on `ChainOrder` but that's more of an aside and changing that probably shouldn't be considered as part of this PR

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -217,6 +217,13 @@ instance TranslateProto singleProto singleProto where
 -- include also recording\/storing arrival information and using that instead
 -- of\/in addition to the 'Ord' instance.
 class Ord sv => ChainOrder sv where
+
+  -- | The 'ChainOrder' type class's 'preferCandidate' function
+  -- consumes both the 'SelectView's of competing chain tips and the
+  -- 'ChainOrderConfig' to make the final decision on which chain is
+  -- strictly preferred. This design allows for flexible and
+  -- protocol-specific tie-breaking rules that go beyond simple chain
+  -- length comparisons.
```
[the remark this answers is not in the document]
amesgen: Yeah, I think this has come up in the past; I think everyone agreed that this would be neat (I don't remember why we didn't follow through with it; probably because it was *relatively* minor and `SelectView` containing the `BlockNo` has been the case forever). Eg concretely
```haskell
data SelectView proto = SelectView {
    svBlockNo :: BlockNo
  , svTiebreaker :: TiebreakerView proto
  }
```
where `TiebreakerView` is an associated type family of `ConsensusProtocol` with `ChainOrder (TiebreakerView proto)` as a superclass constraint, and instances

 - `Ord (TiebreakerView proto) => Ord (SelectView proto)`.
 - `ChainOrder (TiebreakerView proto) => ChainOrder (SelectView proto)`.

---

I was actually considering doing sth like this in the context of Peras, as doing weighted comparisons between chains is naturally modeled by adding the additional weight to the `BlockNo` component, and it is a bit weird that it currently isn't directly accessible.

docs/website/contents/explanation/design-goals.md
```diff
@@ -0,0 +1,61 @@
...
+The consensus protocol is designed to be independent of any specific ledger implementation.
+Since multiple ledgers (such as Shelley and its Shelley-based successors) can utilize the same consensus protocol (TPraos or Praos), the consensus protocol is defined based on what it expects or requires from the ledger rather than being tightly coupled to a specific one.
+This approach makes the consensus protocol abstract and reusable across different ledgers.
+
+## Testability
+
+Ensuring the thorough testability of the consensus layer is a critical design goal.
+As a core component of the Cardano Node, which manages the cryptocurrency, the consensus layer must adhere to strict correctness standards.
+Currently, we extensively employ property-based testing.
+Whenever possible, we should abstract over IO, enabling simulations of various failures (such as disk or network errors) to verify system recovery capabilities.
```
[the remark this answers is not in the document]
dnadales: Good point. What do you think about adding [this footnote](https://github.com/IntersectMBO/ouroboros-consensus/pull/1690/files#diff-c4cb03aeac01b8ab2f46929d9594f5e92eb2e2547b5c0e0ddf6fbe064d9ac91aR30)?

PR #1543, opened 2025-06-02
ChainDB: minor refactorings/simplifications

amesgen:
The commits are intended to be reviewed individually, see their commit messages.

No semantic changes.


jasagredo (approved):
I have no objections. Each commit looks self contained and sufficiently motivated. All looks right. Thanks!

Issue #1205, opened 2024-08-02
Enroll Consensus into the CIP mechanism


dnadales:
We should define a process for proposing changes to Consensus to enable the community to contribute to the design and specification of new features. Such a process should be crystalized in the form of a CIP.

We can use the corresponding CIPs for [Plutus](https://cips.cardano.org/cip/CIP-35) and [Ledeger](https://cips.cardano.org/cip/CIP-84) as inspiration. 

The CIP should be submitted according to the process defined [here](https://cips.cardano.org/cip/CIP-1).

PR #1559, opened 2025-06-11
DONOTMERGE a place to discuss my draft enrollment CIP

nfrisby:
DO NOT MERGE, this is just so we can have threaded conversations about this draft text

The initial content of this PR is just the content of this GitHub Issue Comment https://github.com/IntersectMBO/ouroboros-consensus/issues/1205#issuecomment-2962799831

### DIFF
--- a/WIP-enrollment-CIP.md
+++ b/WIP-enrollment-CIP.md
@@ -11,7 +11,7 @@ My relevant suggestions are essentially just a few sentences.
 > In our experience, the cognitive load of all the requirements on the consensus protocol is enormous.
 > Many newcomers would bounce off of that intimidating list and a even diligent newcomer with a relatively small change would probably find most of that list to be irrelevant.
 > An "open door" policy among the Cardano consensus experts, on the other hand, seems much more plausible to increase the likelihood of promising consensus CIPs being developed.
-> The experts welcome newcomers' to sketch their idea and will happily offer the initial guidance of a specific subset of relevant concerns to consider.
+> The experts welcome newcomers to sketch their idea and will happily offer the initial guidance of a specific subset of relevant concerns to consider.
 > 
 > The only apparent risk to this plan is if there's an overwhelming influx of dead-end ideas.
 > That doesn't seem likely at the moment, and we would adapt the scheme if it happens.


WIP-enrollment-CIP.md
```diff
@@ -0,0 +1,9 @@
+In my current estimation, compared to the Plutus- and Ledger-enrolling CIPs above, any CIP enrolling Consensus would be trivial.
```
nfrisby, line 1: @perturbing said https://github.com/IntersectMBO/ouroboros-consensus/pull/1559#issuecomment-2963187656:

> > any CIP enrolling Consensus would be trivial.

> Can you expand on this statement?

By "it would be trivial" I mean it would just be the quoted paragraphs below, which is the slightest fraction of the Plutus and Ledger enrollment CIPs.

WIP-enrollment-CIP.md
```diff
@@ -0,0 +1,9 @@
...
+We've simply had many less contributors and many fewer design iterations, each of which has been a massive investment.
```
nfrisby, line 2: @perturbing said https://github.com/IntersectMBO/ouroboros-consensus/pull/1559#issuecomment-2963187656:

> > We've simply had many less contributors and many fewer design iterations, each of which has been a massive investment.
> 
> Note that in the future, there will be another node (Amaru), this project is making heavy investments in their knowledge of the consensus stack. So it is not far-fetched that other contributors to the consensus stack can be expected. A result of this will be suggestions on how to improve it as well.
> 
> It makes sense that thus far only plutus and ledger were involved in such an "enrollment" CIP process, as both have direct users. With a second node, there is a direct user of consensus that can produce improvements. So, in this case, is the above addition you made enough to guide a CIP for consensus?

Yes, it's my intention that "available experts on the existing implementation" included CF. I should reword it, eg to "experts on the existing implementationS".

I'm hoping @abailly et al hold similar Office Hours as we do, and so I could included that in the list of relevant meetings as well.

> I am not an expert, but I can imagine that a proposal for consensus might require some checks or basic requirements to call it a proper improvement (which might require significant effort from the author). Similarly, in the plutus "enrollment" CIP, it tries to express what things should be checked and thought about when dealing with plutus improvements, this so that the experts can easily verify that the CIP really improves things and also gives a solid argument for it.
> 
> Given this, is such an CIP also not needed for consensus?

In my experience, the cognitive load of "all the things consensus needs to worry about" is just _huge_. An exhaustive list wouldn't help people---it'd be intimidating and people would bounce off it. My opinion is that inviting them to talk to us and trying to share a specific subset of our concerns with them to start with is much more likely to lead to useful consensus CIPs.

If people find that argument compelling, I could add a paragraph saying as much.

nfrisby: I pushed up a commit with the edits implied by our discussion in this thread 👍  Thanks!

abailly: What you describe does not seem very specific to Cardano, it's the standard situation for any significantly used piece of distributed system: there are a lot of stakeholders whose needs and requirements one needs to address, and yes, that's complex like any social system. 
Perhaps the problem is that this load is concentrated on the shoulders of a handful of persons, or even one person, you? I have worked for company running software comprising millions of line of codes and responsible for trading billions of dollars every day and one thing I have learned there is that no-one can know everything and the knowledge has to spread and codified in the form of documents, processes, etc.

WIP-enrollment-CIP.md
```diff
@@ -0,0 +1,17 @@
...
+My relevant suggestions are essentially just a few sentences.
+
+> The authors of any CIP that explicitly affects the Cardano consensus protocol will need to communicate early and often with some combination of the IOG Researchers and the available experts on the various existing implementations.
+> We suggest beginning by scheduling a discussion on the agenda of a session of INTERSECT's Consensus Technical Working Group.
+> The discussion there will likely involve scheduling additional discussions with the aforementioned Researchers and experts (eg the IOE Consensus Team holds weekly office hours for this sort of communication; Amaru does not hold Office Hours yet, but is planning to).
```
abailly, line 7: This CIP could be the seed from which to grow a proper process that would outlive the people actually implementing consensus now, which requires defining rules and process to change the rules. 
What kind of discussion? Over what kind of artefacts? Who would be part of these discussions? What would be the requirements for a CIP on Consensus to even be discussed by experts? What kind of SLA would experts agree on? etc.

abailly: To draw the analogy with research: you want to do some research, you publish a paper in some venue, the reviewers review your paper, possibly requesting changes and explanations, you update your paper, you get final approval or rejection. 
There's a social contract and a clear process with some relatively well-known rules and SLAs (eg. you have to review papers throughly, and with deadlines).
In our case, a CIP could require some supporting code, perhaps a formal spec with some proofs about the implementation, conformance with previously known or provided test vectors, etc.

rphair (commented):
(GitHub rejected part of this submission, reposted immediately below... @nfrisby please delete this)

rphair (commented):
responding to invitation in https://github.com/IntersectMBO/ouroboros-consensus/pull/1559#discussion_r2140461956:

I think this draft already has many of the points of what would be a complete CIP for the category... and I agree with all statements from other reviewers so far about why a CIP for the `Consensus` category would be welcome.

My only suggestion would be something helpful to the CIP process itself (though generally irrelevant to actual `Consensus` CIPs):

PR #1563, opened 2025-06-17
Only open a V1 ValueHandle when we perform UTxO operations

jasagredo:
# Description

Supersedes #1560


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
```diff
@@ -687,7 +689,7 @@ newForkerByRollback ::
   Word64 ->
   m (Either GetForkerError (Forker m l blk))
 newForkerByRollback h rr n = getEnv h $ \ldbEnv -> do
-  withReadLock (ldbLock ldbEnv) (acquireAtTarget ldbEnv rr (Left n)) >>= traverse (newForker h ldbEnv)
+  withReadLock (ldbLock ldbEnv) (acquireAtTarget ldbEnv (Left n) >>= traverse (newForker h ldbEnv rr))
```
jasagredo, line 692: Note there is a read lock wrapping this whole function body, so in reality there are two read locks at the same time, the outer one and the one that goes "into the forker until we open tables or close it".

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
```diff
@@ -783,21 +757,28 @@ newForker ::
...
           , foeSwitchVar = ldbChangelog ldbEnv
           , foeSecurityParam = ledgerDbCfgSecParam $ ldbCfg ldbEnv
           , foeTracer =
               LedgerDBForkerEvent . TraceForkerEventWithKey forkerKey >$< ldbTracer ldbEnv
           }
-  atomically $ modifyTVar (ldbForkers ldbEnv) $ Map.insert forkerKey forkerEnv
+  atomically $ do
+    -- We need to make sure to release this read access when we drop the value
+    -- handle, so in 'closeForkerEnv' if it wasn't promoted or when promoting it
+    -- in 'getValueHandle'.
```
dnadales, line 779:
```suggestion
-- We need to ensure that read access is released when the value handle is dropped,
-- either in 'closeForkerEnv' (if it wasn’t promoted) or when promoting it in 'getValueHandle'.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/Forker.hs
```diff
@@ -40,15 +42,25 @@ import Ouroboros.Consensus.Storage.LedgerDB.V1.DiffSeq
...
-  -- ^ Local, consistent view of backing store
+  { foeBackingStoreValueHandle ::
+      !( StrictMVar
+           m
+           ( Either
+               (LedgerDBLock m, LedgerBackingStore m l, ResourceRegistry m)
+               (LedgerBackingStoreValueHandle m l)
+           )
+       )
+  -- ^ Either the ingredients to create a value handle or a value handle, i.e. a
```
dnadales, line 61: What does "ingredients" mean in this context?

dnadales (approved):

PR #1572, opened 2025-06-26
LSM trees initial integration

jasagredo:
# Description

this PR might resolve.

Also note that:


ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/ByronHFC.hs
```diff
@@ -292,7 +295,7 @@ instance HasCanonicalTxIn '[ByronBlock] where
     { getByronHFCTxIn :: Void
     }
     deriving stock (Show, Eq, Ord)
-    deriving newtype (NoThunks, MemPack)
+    deriving newtype (NoThunks, MemPack, LSM.SerialiseKey)
```
jasagredo, line 298: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs
```diff
@@ -74,7 +80,7 @@ instance
     { getCardanoTxIn :: SL.TxIn
     }
     deriving stock (Show, Eq, Ord)
-    deriving newtype NoThunks
+    deriving newtype (NoThunks, LSM.SerialiseKey)
```
jasagredo, line 83: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs
```diff
@@ -112,13 +118,27 @@ data CardanoTxOut c
   deriving stock (Show, Eq, Generic)
   deriving anyclass NoThunks
 
+type instance LSMTxOut (LedgerState (CardanoBlock c)) = LSM.RawBytes
```
jasagredo, line 121: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs
```diff
@@ -112,13 +118,27 @@ data CardanoTxOut c
...
+
+instance LSM.SerialiseValue LSM.RawBytes where
```
jasagredo, line 123: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs
```diff
@@ -112,13 +118,27 @@ data CardanoTxOut c
...
+  serialiseValue = id
+  deserialiseValue = id
+
+deriving via LSM.ResolveAsFirst LSM.RawBytes instance LSM.ResolveValue LSM.RawBytes
```
jasagredo, line 127: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Ledger.hs
```diff
@@ -181,6 +201,8 @@ instance
             :* (Fn $ const $ Comp $ K . ConwayTxOut <$> unpackM)
             :* Nil
         )
+    -- TODO, we can extract the tip before this function! The class would be
+    -- IndexedMemPack (NS (Flip LedgerState EmptyMK) (CardanoEras c)) (CardanoTxOut c)
```
jasagredo, line 205: Make an issue about this

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -77,11 +79,13 @@ import Control.Monad.Except
 import Control.State.Transition (PredicateFailure)
 import Data.Data (Proxy (Proxy))
 import Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Database.LSMTree as LSM
```
jasagredo, line 82: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -404,3 +408,34 @@ instance Core.TranslateEra ConwayEra WrapTx where
...
+--     LSM.ResolveValue (SL.ShelleyTxOut AllegraEra)
+
+-- instance LSM.SerialiseValue (SL.ShelleyTxOut MaryEra) where
+--   serialiseValue = serialiseLSMViaMemPack
+--   deserialiseValue = deserialiseLSMViaMemPack
+
+-- deriving via
+--   LSM.ResolveAsFirst (SL.ShelleyTxOut MaryEra)
+--   instance
+--     LSM.ResolveValue (SL.ShelleyTxOut MaryEra)
```
jasagredo, line 441: Why didnt I need these in the end?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -103,6 +104,7 @@ import Data.MemPack
 import qualified Data.Text as T
 import qualified Data.Text as Text
 import Data.Word
+import qualified Database.LSMTree as LSM
```
jasagredo, line 107: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -26,10 +26,12 @@ module Ouroboros.Consensus.Shelley.Node
   , validateGenesis
   ) where
 
+import Cardano.Ledger.Core as Core
 import qualified Cardano.Ledger.Shelley.API as SL
 import Cardano.Protocol.Crypto (Crypto)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import Database.LSMTree as LSM
```
jasagredo, line 34: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -64,6 +64,7 @@ import qualified Data.Text as T (pack)
 import Data.Typeable
 import Data.Void (Void)
 import Data.Word
+import qualified Database.LSMTree as LSM
```
jasagredo, line 67: Import from consensus which re-exports it

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
```diff
@@ -67,6 +67,7 @@ import Data.SOP.Strict
 import qualified Data.SOP.Tails as Tails
 import qualified Data.SOP.Telescope as Telescope
 import Data.Void (Void)
+import qualified Database.LSMTree as LSM
```
jasagredo, line 70: Import from consensus which re-exports it

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -33,6 +34,7 @@ import qualified Data.SOP.Tails as Tails
 import qualified Data.SOP.Telescope as Telescope
 import Data.Void (Void, absurd)
 import Data.Word
+import qualified Database.LSMTree as LSM
```
jasagredo, line 37: Import from consensus which re-exports it

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -105,7 +106,7 @@ data Forker m l blk = Forker
 
     forkerReadTables :: !(LedgerTables l KeysMK -> m (LedgerTables l ValuesMK))
   -- ^ Read ledger tables from disk.
-  , forkerRangeReadTables :: !(RangeQueryPrevious l -> m (LedgerTables l ValuesMK))
+  , forkerRangeReadTables :: !(RangeQueryPrevious l -> m (LedgerTables l ValuesMK, Maybe (TxIn l)))
```
jasagredo, line 109: Explain why the new argument

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -13,6 +14,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
jasagredo, line 17: Necessary?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -225,44 +234,22 @@ mkInternals bss h =
...
-          isDir <- doesDirectoryExist fs $ mkFsPath d'
-          if isDir
-            then truncateRecursively d'
-            else withFile fs (mkFsPath d') (AppendMode AllowExisting) $ \h -> hTruncate fs h 0
-      )
-      dirs
+implIntTruncateSnapshots :: MonadThrow m => SnapshotManagement m m blk st -> SomeHasFS m -> m ()
+implIntTruncateSnapshots snapManager (SomeHasFS fs) = do
+  snapshotsMapM_ snapManager $
+    \pre -> withFile fs (snapshotToStatePath pre) (AppendMode AllowExisting) $
```
jasagredo, line 251: Mention we are only truncating the state file, not the tables

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -474,6 +441,7 @@ data LedgerDBEnv m l blk = LedgerDBEnv
   --
   --  * Modify 'ldbSeq' while holding a write lock, and then close the removed
   --    handles without any locking.
+  , ldbSession :: !(Maybe (ResourceKey m, ResourceKey m))
```
jasagredo, line 444: Perhaps we don't even need this?

ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Ledger/Block.hs
```diff
@@ -86,6 +88,7 @@ import Data.Kind (Type)
 import Data.Proxy
 import Data.Typeable
 import Data.Word
+import qualified Database.LSMTree as LSM
```
jasagredo, line 91: Import from consensus which re-exports it

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/FollowerPromptness.hs
```diff
@@ -1,9 +1,11 @@
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DerivingVia #-}
```
jasagredo, line 3: Unnecesary?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/V1/BackingStore.hs
```diff
@@ -33,6 +33,7 @@ import Data.MemPack
 import qualified Data.SOP.Dict as Dict
 import qualified Data.Set as Set
 import Data.Typeable
+import Ouroboros.Consensus.Ledger.Extended
```
jasagredo, line 36: Unnecessary

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/V1/BackingStore.hs
```diff
@@ -154,7 +155,8 @@ setupBSEnv mkBsArgs mkShfs cleanup = do
 
   let bsi = BS.newBackingStoreInitialiser mempty mkBsArgs (BS.SnapshotsFS shfs)
 
-  bsVar <- newMVar =<< bsi (BS.InitFromValues Origin emptyOTLedgerState emptyLedgerTables)
+  bsVar <-
```
jasagredo, line 158: Unnecessary

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Comment what it contains!

cabal.project
```diff
@@ -97,8 +115,17 @@ source-repository-package
...
+  tag: ea66e96173dc484d82b987acb0df60396f7aa60b
+  --sha256: sha256-3RNqHbDinNz6KOIeCottY153wx7Pkr9WqpbVAvJADSc=
   subdir: plutus-core
           plutus-tx
           plutus-ledger-api
+
+source-repository-package
+  type: git
+  location: https://github.com/jasagredo/lsm-tree
+  tag: 0fe8b227b78a78503d0201755d50435635da5a22
```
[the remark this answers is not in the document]
jasagredo: Ah indeed. I just talked to Joris and we merged it. I will update it to: 84857d4f921a265f267df175e74ba6c652d2b1a0 which is in master now so it is immutable.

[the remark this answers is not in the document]
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,415 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Implementation of the 'LedgerTablesHandle' interface with LSM trees.
+module Ouroboros.Consensus.Storage.LedgerDB.V2.LSM
```
geo2a, line 19: I think we may want to have our own tests for this backend. Even though `lsm-tree` has a comprehensive test suite, I feel like the code in this file is involved enough to at least have some lightweight testing.

For example, the range read seems not to be covered by the `LedgerDB`-level test suite. I ran the tests with coverage as follows:

```
cabal test storage-test --enable-coverage --test-options "-p LSM"
...
Test suite storage-test: RUNNING...
ouroboros-storage
  Storage
    LedgerDB
      StateMachine
        LSM: OK (7.35s)
          +++ OK, passed 1000 tests.
...
```

And the coverage report (see attached, specifically the `Ouroboros.Consensus.Storage.LedgerDB.V2.LSM` module) seems to indicate that the range read was never executed (`hpc` highlights such code in yellow):
lsm-coverage

that agrees with the LedgerDB-level model actions.

[hpc_index.html.txt](https://github.com/user-attachments/files/21349573/hpc_index.html.txt)

jasagredo: Deferred to #1601

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -320,6 +321,32 @@ instance ShelleyCompatible proto era => UpdateLedger (ShelleyBlock proto era)
 type instance TxIn (LedgerState (ShelleyBlock proto era)) = SL.TxIn
 type instance TxOut (LedgerState (ShelleyBlock proto era)) = Core.TxOut era
 
+-- | We use this newtype only to alter MemPack serialization. LSM trees use an
+-- index that looks at the first 8 bytes so it is important to put the index
+-- first such that we avoid all TxIns from the same tx to imply a collision in
+-- the LSM index.
```
dnadales, line 327:
```suggestion
-- | We use this newtype only to alter MemPack serialization. LSM trees use an
-- index that looks at the first 8 bytes, so it is important to put the index
-- first to avoid all TxIns from the same tx to cause a collision in
-- the LSM index.
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -320,6 +321,32 @@ instance ShelleyCompatible proto era => UpdateLedger (ShelleyBlock proto era)
...
+newtype LSMTxIn = LSMTxIn {lsmTxIn :: SL.TxIn}
+
+instance MemPack LSMTxIn where
+  packedByteCount = packedByteCount . lsmTxIn
+  packM (LSMTxIn (SL.TxIn txid txix)) = packM txix >> packM txid
+  unpackM = do
```
dnadales, line 333: We might want to consider avoiding the `do` notation, especially to be consistent with the definition above.

ouroboros-consensus/changelog.d/20250729_124008_javier.sagredo_lsm1.md
```diff
@@ -0,0 +1,27 @@
...
+
+- A bullet item for the Non-Breaking category.
+
+-->
+### Breaking
+
+- Implement LSM-trees backend for LedgerDB V2 handles.
+- Group snapshot management functions in the new datatype `SnapshotManager`.
+- `forkerRangeRead` now returns also the maximal key found in the backend.
+- Deleted `LedgerSupportsOnDiskLedgerDB` constraint and created `LedgerSupports(V1|V2|LSM|LMDB)LedgerDB`.
```
dnadales, line 25:
```suggestion
- Delete `LedgerSupportsOnDiskLedgerDB` constraint and created `LedgerSupports(V1|V2|LSM|LMDB)LedgerDB`.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -81,22 +81,44 @@ import Prelude hiding (read)
...
   { close :: !(m ())
   , duplicate :: !(m (LedgerTablesHandle m l))
-  -- ^ It is expected that this operation takes constant time.
-  , read :: !(LedgerTables l KeysMK -> m (LedgerTables l ValuesMK))
-  , readRange :: !((Maybe (TxIn l), Int) -> m (LedgerTables l ValuesMK))
-  , readAll :: !(m (LedgerTables l ValuesMK))
+  -- ^ Create a copy of the handle.
+  --
+  -- When applying diffs to a table, we will first duplicate the handle, then
+  -- apply the diffs in the copy. It is expected that this operation takes
```
dnadales, line 91: Do we apply a list of diffs? I so, how can they take constant time? 🤔

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -81,22 +81,44 @@ import Prelude hiding (read)
...
+  , read :: !(l EmptyMK -> LedgerTables l KeysMK -> m (LedgerTables l ValuesMK))
+  -- ^ Read values for the given keys from the tables, and deserialize them as
+  -- if they were from the same era as the given ledger state.
+  , readRange :: !(l EmptyMK -> (Maybe (TxIn l), Int) -> m (LedgerTables l ValuesMK, Maybe (TxIn l)))
+  -- ^ Read the requested number of values, possibly starting from the given
+  -- key, from the tables, and deserialize them as if they were from the same
+  -- era as the given ledger state.
+  --
+  -- The returned value contains both the read values as well as the last key
+  -- retrieved. This is necessary because the LSM backend uses an alternative
```
dnadales, line 102: Some questions that might be worth answering in this comment:

- Is the last key retrieved part of the returned map? 
- What is the user expected to do with the returned "last key retrieved"?
- When does this function returns `Nothing` ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -81,22 +81,44 @@ import Prelude hiding (read)
...
   -- ^ Push some diffs into the ledger tables handle.
   --
   -- The first argument has to be the ledger state before applying
   -- the block, the second argument should be the ledger state after
   -- applying a block. See 'CanUpgradeLedgerTables'.
-  , takeHandleSnapshot :: !(l EmptyMK -> String -> m CRC)
+  --
+  -- Note 'CanUpgradeLedgerTables' is only used in the InMemory backend.
+  , takeHandleSnapshot :: !(l EmptyMK -> String -> m (Maybe CRC))
+  -- ^ Take a snapshot of a handle. The given ledger state is used to decide the
```
dnadales, line 118: Why not passing the era index only? Or is there any other information in the ledger state that we need?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,420 @@
...
+                res <- LSM.lookups t (V.fromList keys')
+                pure $
+                  LedgerTables $
+                    ValuesMK $
+                      Map.fromList [(k, fromLSMTxOut st v) | (k, LSM.Found v) <- zip keys' (V.toList res)]
+            )
+      , readRange = implReadRange tv
+      , readAll = \st ->
+          let readAll' m = do
+                (v, n) <- implReadRange tv st (m, 100000)
```
dnadales, line 212: Is there any particular reason why we chose this constant? Would it make sense to add it as a parameter, and pass it in some top-level function with a corresponding note?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -105,11 +106,15 @@ data Forker m l blk = Forker
...
   -- ^ Range-read ledger tables from disk.
   --
-  -- This range read will return as many values as the 'QueryBatchSize' that
-  -- was passed when opening the LedgerDB.
+  -- This range read will return as many values as the 'QueryBatchSize' that was
+  -- passed when opening the LedgerDB.
+  --
+  -- The second component of the returned tuple is the maximal key found by the
```
dnadales, line 115: Maybe here as well we'd want to clarify if the maximal key is part of the map, what is the user expected to do with this key, and what happens if it returns `Nothing`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -205,6 +209,25 @@ data MetadataErr
...
+-- | Management of snapshots for the different LedgerDB backends.
+--
+-- The LedgerDB V1 takes snapshots in @ReadLocked m@, hence the two different
+-- @m@ and @n@ monad types.
+data SnapshotManager m n blk st = SnapshotManager
+  { listSnapshots :: m [DiskSnapshot]
+  , deleteSnapshot :: DiskSnapshot -> m ()
+  , takeSnapshot ::
+      Maybe String ->
+      -- \^ The (possibly empty) suffix for the snapshot name
```
dnadales, line 221: What is the prefix this suffix is appended to?

dnadales (commented):
Looks good 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: The operation that takes constant time is duplicating the handle, the one that has the Haddock attached to it, not applying the diffs.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: An era index would work, but we already have the ledger state in scope so we can just pass that one and use the sop functions with it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
jasagredo: The prefix is the usual prefix for the snapshots, namely the slot number of the snapshot.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
jorisdral: Quick back of the envelope calculation, if we assume keys are ~34 bytes and values are ~60 bytes. `1_000_000` of these key-value pairs would amount to ~100 MB of disk data.

```
ghci> (1000000*(34+60))/1024/1024
89.6453857421875
```

Seeing as `readAll` reads all the disk data into memory anyway, which is inherently memory and CPU intensive, maybe the read size parameter doesn't matter so much. Still, could be good to test out different values

jorisdral:
Oh, BTW, I forgot to add it to my review but I don't think that currently the `lsm-tree` session and `HasBlockIO` are closed explicitly when the ledger DB are closed. They should be closed by the registry as a backup mechanism, but maybe they should be closed explicitly too in `LedgerDB.closeDB`?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -320,6 +321,32 @@ instance ShelleyCompatible proto era => UpdateLedger (ShelleyBlock proto era)
 type instance TxIn (LedgerState (ShelleyBlock proto era)) = SL.TxIn
 type instance TxOut (LedgerState (ShelleyBlock proto era)) = Core.TxOut era
 
+-- | We use this newtype only to alter MemPack serialization. LSM trees use an
+-- index that looks at the first 8 bytes so it is important to put the index
+-- first such that we avoid all TxIns from the same tx to imply a collision in
+-- the LSM index.
+newtype LSMTxIn = LSMTxIn {lsmTxIn :: SL.TxIn}
```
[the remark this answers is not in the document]
amesgen: Drive-by question: I probably sound like a broken record, but compact indices require that the ["keys must be uniformly distributed"](https://intersectmbo.github.io/lsm-tree/lsm-tree-1.0.0.0-inplace/Database-LSMTree.html#t:FencePointerIndexType). How difficult is it for adversaries to grind hashes that result in (non-trivial) performance degradation?

(I guess/hope that, similar to Bloom filter poisoning concerns, this currently is pretty irreleant as Cardano's throughput is so low, but that will not be true indefinitely.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
jasagredo: readAll should be gone once we have streaming tables, so I wouldn't worry too much, but I will leave this comment open just in case

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,420 @@
...
+  ( IOLike m
+  , HasLedgerTables l
+  , LedgerSupportsLSMLedgerDB l
+  ) =>
+  Tracer m V2.FlavorImplSpecificTrace ->
+  ResourceRegistry m ->
+  Session m ->
+  (ResourceKey m, UTxOTable m l) ->
+  m (LedgerTablesHandle m l)
+newLSMLedgerTablesHandle tracer rr session (resKey, t0) = do
```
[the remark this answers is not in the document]
jasagredo: It is only used to create duplicates, but is that a problem?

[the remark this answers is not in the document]
jorisdral: I saw that `session` is only passed as an argument to `newLSMLedgerTablesHandle` -- it's never used otherwise. I think you can safely remove it and the code will still work

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Args.hs
```diff
@@ -1,34 +1,53 @@
...
+data HandleArgs f m
   = InMemoryHandleArgs
-  | LSMHandleArgs Void
-  deriving (Generic, NoThunks)
+  | LSMHandleArgs (LSMHandleArgs f m)
+
+data LSMHandleArgs f m = LSMArgs
+  { lsmFilePath :: HKD f FilePath
+  -- ^ The file path relative to the fast storage directory in which the LSM
+  -- trees database will be located.
```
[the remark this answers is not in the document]
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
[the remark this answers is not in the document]
jasagredo: Ah wait, you are right! I am just passing it around

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -81,22 +81,49 @@ import Prelude hiding (read)
...
+  -- encoding of the values based on the current era.
+  --
+  -- It returns a CRC only on backends that support it, as the InMemory backend.
```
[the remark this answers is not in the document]
fraser-iohk: already discussed and it's probably unnecessary

cabal.project
```diff
@@ -90,3 +95,9 @@ constraints:
   plutus-core < 1.51,
   plutus-ledger-api < 1.51,
   plutus-tx < 1.51,
+
+source-repository-package
+  type: git
+  location: https://github.com/jasagredo/digest
+  tag: 329fc2a911878ffe47472751cb40aae20ab2c00a
+  --sha256: sha256-84f8dFee9EfWbQ5UTLZ9MrsZ3JVojNhzfTGmWof6wHU=
```
amesgen, line 103: This can be conditional to just be used on Windows, right? (Might make sense to add a comment why it is necessary)

Also, what is the plan to make this unnecessary? Release this to CHaP for now, and then try to upstream?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -617,13 +703,13 @@ mkTrackOpenHandles = do
...
-checkNoLeakedHandles (Environment _ testInternals _ _ _ getNumOpenHandles _) = do
-  expected <- liftIO $ NumOpenHandles <$> LedgerDB.getNumLedgerTablesHandles testInternals
-  actual <- liftIO getNumOpenHandles
-  QC.assertWith (actual == expected) $
-    "leaked handles, expected " <> show expected <> ", but actual " <> show actual
+checkNoLeakedHandles :: Environment -> IO (NumOpenHandles, NumOpenHandles)
+checkNoLeakedHandles (Environment _ testInternals _ _ _ getNumOpenHandles _ _) = do
+  expected <- NumOpenHandles <$> LedgerDB.getNumLedgerTablesHandles testInternals
+  actual <- getNumOpenHandles
+  pure (actual, expected)
```
amesgen, line 715: Why did this need to be changed?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -112,13 +116,53 @@ prop_sequential ::
...
+--   QC.monadic (QC.ioProperty . Monad.void . flip runStateT env) propM
+
+-- QC.withMaxSuccess maxSuccess $
+-- QC.monadic $
+--   QC.ioProperty $
+--     mask $ \restore -> do
+--       ref <- lift $ initialEnvironment fsOps getLmdbDir mkTestArguments =<< initChainDB
+--       (_, env@(Environment _ testInternals _ _ _ _ clean)) <- runPropertyStateT (runActions as) ref
+--       checkNoLeakedHandles env
+--       lift $ closeLedgerDB testInternals >> clean
```
amesgen, line 165: Leftover?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -762,12 +769,72 @@ instance
...
+  -- > class SerialiseValue k where
+  -- >   serialiseKey :: v -> ByteArray
+  -- >   deserialiseKey :: ByteArray -> v
+  --
+  -- So we cannot pass a deserialisation hint like we do with
+  -- 'IndexedMemPack', so this option must be discarded:
+  --
+  -- > type Option1 = Table TxIn TxOut
+  -- > instance SerialiseValue TxOut where
+  -- >   deserialiseValue = {- we want to use mempack, but we have no deserialisation hint -}
```
amesgen, line 808: Did you consider using `reflection`, ie [`Given`](https://hackage.haskell.org/package/reflection-2.1.9/docs/Data-Reflection.html#t:Given)?

Ie then you could define
```haskell
instance Given Hint => SerialiseValue TxOut where
  deserialiseValue = {- code -}
    where 
      hint = given @Hint
```
and then at every call site of a function `foo` requiring `SerialiseValue TxOut`, you write 
```haskell
give hint $ foo arg1 arg2
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -762,12 +769,72 @@ instance
...
+  -- > -- Roughly what I mean by "consensus deserialises manually outside of lsm-tree"
+  -- > readTxOuts hint txIns table = fromLSMTxOut hint <$> lookups table txIns
+  --
+  -- So we will define `LSMTxOut (LedgerState (CardanoBlock StandardCrypto))` to
+  -- `ByteArray`.
+  type LSMTxOut l
+
+  toLSMTxOut :: Proxy l -> TxOut l -> LSMTxOut l
+  fromLSMTxOut :: l EmptyMK -> LSMTxOut l -> TxOut l
+  lsmIndex :: Proxy l -> LSM.FencePointerIndexType
```
amesgen, line 828: What are possible reasons to use different `FencePointerIndexType`s for different blocks? Is it "use `CompactIndex` if possible, otherwise use `OrdinaryIndex`"? Are there blocks that don't support `CompactIndex`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -762,12 +769,72 @@ instance
...
+  lsmSnapLabel :: Proxy l -> String
```
amesgen, line 829: Replace with `ShowProxy l`?

cabal.project
jasagredo: Good call

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
amesgen: More generally, what seems a bit dubious to me is that we need to define `SerialiseValue`/`SerialiseKey` for every block. Even without something like `reflection`, couldn't we deserialize into `ByteArray`, and then just invoke the `IndexedMemPack` logic in `newLSMLedgerTablesHandle`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
jasagredo: @jorisdral probably has more insights. In general I think blocks whose txins are hashes should use the CompactIndex, and blocks whose txins are something else like consecutive ints (like test blocks) should not use the CompactIndex.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
fraser-iohk: oh I've never looked at this library before, this is neat

jasagredo: >Even without something like reflection, couldn't we deserialize into ByteArray, and then just invoke the IndexedMemPack logic in newLSMLedgerTablesHandle?

Done!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
jasagredo: > Replace lsmSnapLabel with ShowProxy l?

Done, it turns out the limitations are on `SnapshotName` not `SnapshotLabel`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
jasagredo: Due to the complications we face with the CompactIndex, we are for now going to integrate using the OrdinaryIndex and maybe reconsider doing this in a follow-up PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -81,22 +81,49 @@ import Prelude hiding (read)
...
+  -- if they were from the same era as the given ledger state.
+  , readRange :: !(l EmptyMK -> (Maybe (TxIn l), Int) -> m (LedgerTables l ValuesMK, Maybe (TxIn l)))
+  -- ^ Read the requested number of values, possibly starting from the given
+  -- key, from the tables, and deserialize them as if they were from the same
+  -- era as the given ledger state.
+  --
+  -- The returned value contains both the read values as well as the last key
+  -- retrieved. This is necessary because the LSM backend uses an alternative
+  -- serialization format and the last key in the returned Map might not be the
+  -- last key read.
```
amesgen, line 104: Currently, this is only not true for LSM because the TxIx uses little endian encoding, right?

That makes me wonder: Isn't the same true for LMDB, given that it uses the same serialization of keys? If so, it actually also needs to be updated (or we change the serialization to be big endian).

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -64,26 +69,54 @@ openLedgerDB ::
...
+          snapManager
       )
+      snapManager
       emptyStream
       genesisPoint
   pure (ledgerDB, intLedgerDB)
 openLedgerDB lgrDbArgs@LedgerDB.LedgerDbArgs{LedgerDB.lgrFlavorArgs = LedgerDB.LedgerDbFlavorArgsV2 args} = do
+  (snapManager, bss') <- case args of
+    V2.V2Args V2.InMemoryHandleArgs -> pure (InMemory.snapshotManager lgrDbArgs, V2.InMemoryHandleEnv)
+    V2.V2Args (V2.LSMHandleArgs (V2.LSMArgs path genSalt mkFS)) -> do
```
amesgen, line 89: It seems sensible to actually allow to run db-analyser with LSM and eg replay mainnet as a non-trivial usage. Did you already do this maybe elsewhere/locally?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,478 @@
...
+          table <- allocate rr (\_ -> LSM.duplicate t) LSM.closeTable
+          newLSMLedgerTablesHandle tracer rr table
+      , read = \st (LedgerTables (KeysMK keys)) -> do
+          vec <- VM.unsafeNew (Set.size keys)
+          _ <-
+            Foldable.foldl'
+              (\m x -> m >>= \i -> VM.write vec i (toTxInBytes (Proxy @l) x) >> pure (i + 1))
+              (pure 0)
+              keys
+          vec' <- V.freeze vec
```
amesgen, line 253: There is [`create`](https://hackage-content.haskell.org/package/vector-0.13.2.0/docs/Data-Vector-Generic.html#v:create) to do this safely (and even more efficiently, as it can use `unsafeFreeze`).

Also, are we sure this is faster than
```haskell
V.fromListN (Set.size) $ toTxInBytes (Proxy @l) <$> Set.toList keys 
```
to warrant the extra complexity?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,478 @@
...
+ where
+  cursorFromStart = LSM.withCursor table (LSM.take num)
+  -- Here we ask for one value more and we drop one value because the
+  -- cursor returns also the key at which it was opened.
+  cursorFromKey k = fmap (V.drop 1) $ LSM.withCursorAtOffset table (toTxInBytes (Proxy @l) k) (LSM.take $ num + 1)
+
+foldMWithKey :: Monad m => (a -> k -> v -> m a) -> a -> Map.Map k v -> m a
+foldMWithKey f z m = foldM step z (Map.toAscList m)
+ where
+  step acc (k, v) = f acc k v
```
amesgen, line 317: Isn't this basically the same as `ifoldlM` which you already use above?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -1050,7 +1053,20 @@ stdLowLevelRunNodeArgsIO
...
+            case srnLedgerDbBackendArgs of
+              V1LMDB args -> LedgerDbFlavorArgsV1 args
+              V2InMemory -> LedgerDbFlavorArgsV2 (V2.V2Args V2.InMemoryHandleArgs)
+              V2LSM path ->
+                LedgerDbFlavorArgsV2
+                  ( V2.V2Args
+                      ( V2.LSMHandleArgs
+                          ( V2.LSMArgs
+                              (mkFsPath $ splitDirectories path)
+                              LSM.stdGenSalt
```
amesgen, line 1065: I think it would be a bit more in line with existing consumers of randomness to use `llrnRng`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,478 @@
...
+newtype TxInBytes = TxInBytes {unTxInBytes :: LSM.RawBytes}
+
+toTxInBytes :: MemPack (TxIn l) => Proxy l -> TxIn l -> TxInBytes
+toTxInBytes _ txin =
+  let barr = packByteArray True txin
+   in TxInBytes $ LSM.RawBytes (VP.Vector 0 (PBA.sizeofByteArray barr) barr)
+
+fromTxInBytes :: MemPack (TxIn l) => Proxy l -> TxInBytes -> TxIn l
+fromTxInBytes _ (TxInBytes (LSM.RawBytes (VP.Vector _ _ barr))) =
+  unpackError barr
```
amesgen, line 143: I guess it is fine to throw an imprecise exception here. We will likely enter a crash restart loop, so we might want to have special logging for this case. Maybe document this and potentially wrap the exception?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,478 @@
...
+      , readAll = \st ->
+          let readAll' m = do
+                (v, n) <- implReadRange t st (m, 100000)
+                maybe (pure v) (fmap (ltliftA2 unionValues v) . readAll' . Just) n
+           in readAll' Nothing
+      , pushDiffs = const (implPushDiffs t)
+      , takeHandleSnapshot = \_ snapshotName -> do
+          LSM.saveSnapshot
+            (fromString snapshotName)
+            (LSM.SnapshotLabel $ Text.pack $ "UTxO table: " ++ showProxy (Proxy @(LedgerBlock l)))
```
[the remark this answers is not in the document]
jasagredo: For 2. I was able to create an overlapping instance
```
instance {-# OVERLAPPING #-} ShowProxy CardanoBlock where
  showProxy _ = "CardanoBlock"
```

Does that sound like a better alternative?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
jasagredo: Very good observation. I guess it actually means that we were doing the reads from LMDB wrong all this time? It is not crucial for the functioning of the node, because range reads are only used for user queries and it only affects LMDB backends, but I think you are right!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -166,6 +166,8 @@ class
   -- these errors does not depend on the type of the block.
   type LedgerErr l :: Type
 
+  type LedgerBlock l :: Type
```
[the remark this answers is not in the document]
jasagredo: Actually I removed the extra string in the snapshot label, so I removed this type family

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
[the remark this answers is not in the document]
jasagredo: Actually I removed the extra string in the snapshot label, so this is not applicable anymore

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
jasagredo: Not sure what are you pointing out with this. If you mean that I should add an option to run with LSM trees in db-analyser, the idea is to do this on a follow-up https://github.com/IntersectMBO/ouroboros-consensus/issues/1592

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -197,6 +224,20 @@ inMemV2TestArguments secParam _ =
...
+  SecurityParam ->
+  FilePath ->
+  TestArguments IO
+lsmTestArguments secParam fp =
+  TestArguments
+    { argFlavorArgs =
+        LedgerDbFlavorArgsV2 $
+          V2Args $
+            LSMHandleArgs $
+              LSMArgs (mkFsPath $ FilePath.splitDirectories fp) LSM.stdGenSalt (LSM.stdMkBlockIOFS fp)
```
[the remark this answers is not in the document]
jasagredo: Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
amesgen: Yeah, it is depressing that one has to make these kinds of judgement calls of simplicity vs (potential) predictable performance. *When* convenient staging for predictable code generation :stuck_out_tongue:

(Using `create` should be a clear win though.)

---

Also, looking at this again, using `foldM_` seems slightly cleaner

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -752,7 +752,7 @@ newForker h ldbEnv rr st = do
   let tr = LedgerDBForkerEvent . TraceForkerEventWithKey forkerKey >$< ldbTracer ldbEnv
   traceWith tr ForkerOpen
   lseqVar <- newTVarIO . LedgerSeq . AS.Empty $ st
-  (k, toRelease) <- allocate rr (\_ -> newTVarIO (pure ())) (readTVarIO Monad.>=> id)
+  (k, toRelease) <- allocate rr (\_ -> newTVarIO (close (tables st))) (Monad.join . readTVarIO)
```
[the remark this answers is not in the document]
amesgen: Also, what about async exception safety here? What if we get an async exception after having opened the forker via `getStateRef`, but before `allocate` has finished here?

jasagredo:
I plan on merging this as a PR stack:
- #1572	←
- #1644	
- #1641	
- #1643	
- #1640	
- #1639

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,462 @@
...
+newtype TxOutBytes = TxOutBytes {unTxOutBytes :: LSM.RawBytes}
+
+toTxOutBytes :: IndexedMemPack (l EmptyMK) (TxOut l) => l EmptyMK -> TxOut l -> TxOutBytes
+toTxOutBytes st txout =
+  let barr = indexedPackByteArray True st txout
+   in TxOutBytes $ LSM.RawBytes (VP.Vector 0 (PBA.sizeofByteArray barr) barr)
+
+fromTxOutBytes :: IndexedMemPack (l EmptyMK) (TxOut l) => l EmptyMK -> TxOutBytes -> TxOut l
+fromTxOutBytes st (TxOutBytes (LSM.RawBytes (VP.force -> (VP.Vector _ _ barr)))) =
+  case indexedUnpackLeftOver' st barr of
```
amesgen, line 115: Is it really expected that we `VP.force` *and* allow leftovers?

More generally (as also briefly discussed on Slack some time ago), it seems rather easy to just invoke `mempack` in such a way that it is aware of the offset and length of the `ByteArray`, meaning that we can avoid the copy via `VP.force`.

Same idea, different guise: We could write an appropriate [`Buffer`](https://hackage-content.haskell.org/package/mempack-0.1.2.0/docs/Data-MemPack-Buffer.html#t:Buffer) instance for `RawBytes`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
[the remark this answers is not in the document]
jasagredo: We are not using the CompactIndex, so I will close this comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
```diff
@@ -0,0 +1,466 @@
...
+  showTypeOf _ = "Table"
+  wNoThunks _ _ = pure Nothing
+
+data LSMClosedExn = LSMClosedExn
+  deriving (Show, Exception)
+
+instance Buffer LSM.RawBytes where
+  bufferByteCount (LSM.RawBytes (VP.Vector _ l _)) = l
+  buffer (LSM.RawBytes (VP.Vector (I# off) _ (PBA.ByteArray barr))) _ g =
+    g $ plusAddr# (byteArrayContents# barr) off
```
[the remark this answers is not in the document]
jasagredo: Implemented in https://github.com/IntersectMBO/ouroboros-consensus/compare/26e8bf26142c7e1a6a97cca5861ac4898271987b..3a525ea94d9d0bd36d89a2a78afc91f036dc6ef8

cabal.project
jasagredo: I think we can close this as the digest patch that removes crc32c for windows is released in CHaP.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/IndexedMemPack.hs
```diff
@@ -71,12 +83,30 @@ indexedUnpackFail idx b = do
...
+  forall idx a b.
+  (IndexedMemPack idx a, Buffer b, HasCallStack) => idx -> b -> Either SomeError a
+indexedUnpackEither idx = first fromMultipleErrors . runFailAgg . indexedUnpackFail idx
+{-# INLINEABLE indexedUnpackEither #-}
+
+unpackEither ::
+  forall a b.
+  (MemPack a, Buffer b, HasCallStack) => b -> Either SomeError a
+unpackEither = first fromMultipleErrors . runFailAgg . unpackFail
+{-# INLINEABLE unpackEither #-}
```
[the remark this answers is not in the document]
jasagredo: It was more because of the `fromMultipleErrors` than the `runFail`. In the end I just tried to mimick what was there in Data.MemPack

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/IndexedMemPack.hs
```diff
@@ -53,6 +59,12 @@ indexedPackByteArray isPinned idx a =
     (indexedPackM idx a)
 {-# INLINE indexedPackByteArray #-}
 
+indexedUnpack ::
+  forall idx a b.
+  (Buffer b, IndexedMemPack idx a, HasCallStack) => idx -> b -> Either SomeError a
+indexedUnpack idx = first fromMultipleErrors . runFailAgg . indexedUnpackFail idx
+{-# INLINEABLE indexedUnpack #-}
```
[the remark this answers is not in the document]
jasagredo: Aha indeed

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -55,9 +55,14 @@ parseDBAnalyserConfig =
...
       , flag' V2InMem $
           mconcat
-            [ long "v2-in-mem"
+            [ long "in-mem"
             , help "use v2 in-memory backend"
             ]
+      , flag' V2LSM $
+          mconcat
+            [ long "lsm"
+            , help "use v2 LSM backend"
```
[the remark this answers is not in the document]
jasagredo: I think I would rather keep it until we remove V1. It is only in the help text so it should be fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LedgerSeq.hs
```diff
@@ -80,22 +80,49 @@ import Prelude hiding (read)
...
-  , readAll :: !(m (LedgerTables l ValuesMK))
+  -- ^ Create a copy of the handle.
+  --
+  -- When applying diffs to a table, we will first duplicate the handle, then
+  -- apply the diffs in the copy. It is expected that duplicating the handle
+  -- takes constant time.
+  , read :: !(l EmptyMK -> LedgerTables l KeysMK -> m (LedgerTables l ValuesMK))
+  -- ^ Read values for the given keys from the tables, and deserialize them as
+  -- if they were from the same era as the given ledger state.
+  , readRange :: !(l EmptyMK -> (Maybe (TxIn l), Int) -> m (LedgerTables l ValuesMK, Maybe (TxIn l)))
```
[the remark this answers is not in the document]
jasagredo: I think this is clearly explained in the haddock for the function just below.

> ```
> -- ^ Read the requested number of values, possibly starting from the given
> -- key,
> ```

> ```
> -- ^ ... If the function returns
> -- Nothing, it means the read returned no results, or in other words, we
> -- reached the end of the ledger tables.
> ```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/LSM.hs
bladyjoker, line 1: Looked at from `fromTxInBytes` and `fromTxOutBytes` and `error` calls therein. Couple of questions:

1. Should we introduce module level `Error` type and handle such conditions gracefully?

```
data Error = TxOutDecodingError SomeError | TxInDecodingError SomeError | InternalError Text
```

Even though we're calling `from` functions in `IOLike m`, it would be very useful to have structured error types such that we can eventually also provide machine readable errors in agreed upon format etc.
 
2. What's our take on `error` generally?

nix/haskell.nix
```diff
@@ -89,9 +89,13 @@ in
     nativeBuildInputs = [
       final.fd
       final.cabal-docspec
-      (hsPkgs.ghcWithPackages
-        (ps: [ ps.latex-svg-image ] ++ lib.filter (p: p ? components.library)
-          (lib.attrValues (haskell-nix.haskellLib.selectProjectPackages ps))))
+      (hsPkgs.shellFor {
```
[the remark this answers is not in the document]
jasagredo: See the commit message in https://github.com/IntersectMBO/ouroboros-consensus/pull/1572/commits/a0d614eedacdf77af771a9578132fa58ca66ef23

cabal.project
```diff
@@ -49,3 +49,38 @@ if impl (ghc >= 9.12)
...
+  allow-newer:
+    -- https://github.com/phadej/regression-simple/pull/14
+    , regression-simple:base
+
+source-repository-package
+  type: git
+  location: https://github.com/IntersectMBO/cardano-ledger
+  tag: fb09078fa55015c881303a2ddb609c024cec258f
+  --sha256: sha256-9Y9CRiyMn0AWD+C4aNVMaJgrj3FDAYfCX4VrLvtoMaI=
+  subdir:
```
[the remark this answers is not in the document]
bladyjoker: When do we expect cardano-ledger to land in Chap? Do we tend to merge like this or do we coordinate with Ledger folks to get this in Chap?

Curious...is there automation that uploads to Chap on every master merge? What's the story there?

PR #1575, opened 2025-06-30
LedgerDB: implement predictable snapshotting

### LINKED ISSUE #1424: [FEAT] - Snapshot every Nth block

### LINKED ISSUE #1573: Optional random delay when creating snapshots
As the performance impact of a snapshot is still non-trivial (albeit improved recently), cf #868, it is desirable to avoid scenarios where all nodes in the system perform a snapshot at the same time.

In the current snapshotting scheme, this is achieved implicitly due to the ad-hoc policy used for when to snapshot, which depends on the startup time, and so when nodes create snapshots is *usually* largely uncorrelated.

With predictable snapshotting (#1424), this will change, so a simple mitigation is to introduce a random delay before we do the snapshot (while still snapshotting for a particular slot). A random delay of eg 5-10 minutes sounds fine.

This mechanism is only needed when caught-up; when syncing, it would increase peak memory usage if we had to keep around ledger states from 10 minutes ago (as we might have adopted hundreds of thousands of blocks in the meantime).

---

In the long term (with lsm-tree and LedgerHD), snapshotting should become very cheap, so this functionality can then be removed again.

amesgen:
Closes #1424, https://github.com/IntersectMBO/ouroboros-consensus/issues/1573

Based on top of #1513, see there for the relation to this change.

Note that we need #1573 before this can be released in a node.

---

This PR replaces the previous logic for when to create snapshots (It would be possible to preserve it, but I don't see a big motivation). Concretely, `SnapshotFrequencyArgs` contains

 - `sfaInterval :: SlotNo`: Create snapshots every `sfaInterval` many slots.

   Default: `2*k = 4320` slots, so 72 min on mainnet as before.
 - `sfaOffset :: SlotNo`: Allows to determine the offset of where snapshots are taken, see below.

   Default: `0`
 - `sfaRateLimit :: DiffTime`: A minimum duration between snapshots (used to avoid excessive snapshots while syncing).

   Default: 10 minutes (previous value was 6 minutes, which seemed a bit low, so I increased it somewhat. Maybe it should be increased even more now that we no longer have the [`substantialAmountOfBlocksWereProcessed`](https://github.com/IntersectMBO/ouroboros-consensus/blob/b4b17499e1c2118a5cac1f7e7bc5b8f83cf5bcc4/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs#L552) check.)

Concretely, the node will try to create snapshots for the last immutable blocks before the slots

    sfaOffset, sfaOffset + sfaInterval, sfaOffset + 2 * sfaInterval, sfaOffset + 3 * sfaInterval, ...

but can skip creating some of these depending on the `sfaRateLimit` (which can be disabled by setting it to a non-positive value). Also see the Haddocks.

For example, setting `sfaInterval = 10*2160*20` (one mainnet Shelley epoch) and `sfaOffset = 172800` will cause the node to create snapshots for the last block in every Shelley epoch (because the first Shelley slot is `4492800`, and ``4492800 `mod` (10*2160*20) = 172800``. By tweaking `sfaOffset`, one can take snapshots eg right before the midway point in each epoch.

There is some code that could be shared between V1 and V2 (already even before this PR), but given the upcoming removal of V1, this seems acceptable for now.

---

Also includes a test running a ChainDB in `IOSim` to test that everything is hooked up correctly (in particular regarding the background threads).

See https://github.com/IntersectMBO/ouroboros-consensus/pull/1513#issuecomment-2944658608 for sync regression tests.


amesgen:
I integrated this into 10.5 for testing:

 - Consensus branch: [`amesgen/predictable-snapshots-10.5`](https://github.com/IntersectMBO/ouroboros-consensus/tree/amesgen/predictable-snapshots-10.5)
 - Node branch: [`amesgen/test-snapshot-stuff`](https://github.com/IntersectMBO/cardano-node/tree/amesgen/test-snapshot-stuff)

   Currently, this treats `SnapshotInterval` (which previously was a number of seconds) as a number of slots. This is convenient, and should be fine for mainnet AFAICT, but maybe we still want to add some kind of warning (see also the TODO in the code there).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -475,95 +464,144 @@ data SnapshotPolicy = SnapshotPolicy
...
+  -- 'onDiskSnapshotSelector' must be a subset of this list.
+  }
+  deriving stock Show
+
+-- | Determines when/how often we take ledger snapshots.
+--
+-- We only write snapshots for ledger states that are /immutable/. Concretely,
+-- for every slot @s@ out of
+--
+-- > sfaOffset, sfaOffset + sfaInterval, sfaOffset + 2 * sfaInterval, sfaOffset + 3 * sfaInterval
```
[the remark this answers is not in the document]
amesgen: I added the same example as in the PR description :+1:

changelog.d/20260407_141821_georgy.lukyanov_predictable_snapshots.md
```diff
@@ -0,0 +1,32 @@
...
+
+- `tryTakeSnapshot` now accepts a `Time` argument.
+- Renamed `ldbLastSnapshotWrite` to `ldbLastSnapshotRequestedAt`
+- Added a `delay` argument to `implTryTakeSnapshot`.
+- Renamed `ledgerDbMaintenaceThread` to `ledgerDbMaintenanceThread`
+
+
+### Non-Breaking
+
+- add `cdbSnapshotDelayRNG` filed to `ChainDbEnv`.
```
geo2a: filed -> field

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -334,14 +334,15 @@ implTryTakeSnapshot ::
...
+  (SnapshotDelayRange -> m DiffTime) ->
   m ()
-implTryTakeSnapshot snapManager env = do
-  timeSinceLastWrite <- do
-    mLastWrite <- readTVarIO $ ldbLastSnapshotWrite env
-    for mLastWrite $ \lastWrite -> do
-      now <- getMonotonicTime
-      pure $ now `diffTime` lastWrite
-  RAWLock.withReadAccess (ldbOpenHandlesLock env) $ \() -> do
+implTryTakeSnapshot snapManager env snapshotRequestTime getRandomDelay = do
```
geo2a: Calculate `snapshotRequestTime` inside this function using `now`, as it was before.

changelog.d/20260407_141821_georgy.lukyanov_predictable_snapshots.md
geo2a: Done.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
geo2a: Fixed.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -840,13 +842,15 @@ openChainDB ::
   (ChainDB.RelativeMountPoint -> SomeHasFS m) ->
   -- | Volatile FS, see 'NodeDatabasePaths'
   (ChainDB.RelativeMountPoint -> SomeHasFS m) ->
+  StdGen ->
+  -- | RNG used to randomise snapshot delays
```
jasagredo, line 845: This comment should point upwards

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -362,23 +358,27 @@ data LedgerDB m l blk = LedgerDB
   --
   --  * The set of previously applied points.
   , tryTakeSnapshot ::
+      l ~ ExtLedgerState blk =>
```
jasagredo: Why do we need this now?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -494,21 +502,6 @@ decodeLBackwardsCompatible _ decodeLedger decodeHash =
...
-data SnapshotInterval
-  = DefaultSnapshotInterval
-  | RequestedSnapshotInterval DiffTime
-  | DisableSnapshots
-  deriving stock (Eq, Generic, Show)
-
--- | Number of snapshots to be stored on disk. This is either the default value
--- as determined by the @'SnapshotPolicy'@, or it is provided by the user. See the
--- @'SnapshotPolicy'@ documentation for more information.
-data NumOfDiskSnapshots
```
jasagredo, line 507: Ah where did this go?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -541,95 +534,177 @@ data SnapshotPolicy = SnapshotPolicy
...
+data SnapshotFrequency
+  = SnapshotFrequency SnapshotFrequencyArgs
+  | DisableSnapshots
+  deriving stock (Show, Eq)
+
 data SnapshotPolicyArgs = SnapshotPolicyArgs
-  { spaInterval :: !SnapshotInterval
-  , spaNum :: !NumOfDiskSnapshots
+  { spaFrequency :: SnapshotFrequency
+  , spaNum :: OverrideOrDefault Word
```
jasagredo: I think we should bring back the datatype (or a new newtype)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -638,8 +713,25 @@ defaultSnapshotPolicy
...
+  | -- | A snapshot request was completed
+    SnapshotRequestCompleted
   | -- | A snapshot was written to disk.
     TookSnapshot DiskSnapshot (RealPoint blk) EnclosingTimed
   | -- | An old or invalid on-disk snapshot was deleted
     DeletedSnapshot DiskSnapshot
   deriving (Generic, Eq, Show)
+
+{-------------------------------------------------------------------------------
+  Utility (could live in O.C.Util.Args)
```
jasagredo: Let's move it now to not have to move it in the future

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
```diff
@@ -306,34 +310,74 @@ implGarbageCollect env slotNo = atomically $ do
...
-  Maybe (Time, Time) ->
-  Word64 ->
-  m SnapCounters
-implTryTakeSnapshot snapManager env copyBlocks mTime nrBlocks =
-  if onDiskShouldTakeSnapshot (ldbSnapshotPolicy env) (uncurry (flip diffTime) <$> mTime) nrBlocks
-    then do
+  (SnapshotDelayRange -> m DiffTime) ->
+  m ()
+implTryTakeSnapshot snapManager env copyBlocks getRandomDelay = do
+  snapshotRequestTime <- getMonotonicTime
```
jasagredo: You could name this `now`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
```diff
@@ -306,34 +310,74 @@ implGarbageCollect env slotNo = atomically $ do
...
+    [] -> pure ()
+    _ -> do
       copyBlocks
-      void $
-        withReadLock
+
+      delayBeforeSnapshotting <- getRandomDelay (onDiskSnapshotDelayRange (ldbSnapshotPolicy env))
+      let nonEmptySnapshotSlots =
+            case NonEmpty.nonEmpty $ snapshotSlots of
+              Nothing -> error "impossible: empty handles, see pattern-match above"
```
jasagredo: Couldn't we eliminate this by calling nonEmpty above?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
```diff
@@ -551,6 +595,12 @@ data LedgerDBEnv m l blk = LedgerDBEnv
   , ldbQueryBatchSize :: !QueryBatchSize
   , ldbResolveBlock :: !(ResolveBlock m blk)
   , ldbGetVolatileSuffix :: !(GetVolatileSuffix m blk)
+  , ldbLastSnapshotRequestedAt :: !(StrictTVar m (Maybe Time))
```
jasagredo: Let's rename to `ldbLastSuccessfulSnapshotRequestedAt` or `Completed`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -331,26 +337,67 @@ implTryTakeSnapshot ::
...
-  Maybe (Time, Time) ->
-  Word64 ->
-  m SnapCounters
-implTryTakeSnapshot snapManager env copyBlocks mTime nrBlocks =
-  if onDiskShouldTakeSnapshot (ldbSnapshotPolicy env) (uncurry (flip diffTime) <$> mTime) nrBlocks
-    then do
+  (SnapshotDelayRange -> m DiffTime) ->
+  m ()
+implTryTakeSnapshot snapManager env copyBlocks getRandomDelay = do
+  snapshotRequestTime <- getMonotonicTime
```
jasagredo: Same as above, name it `now`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -331,26 +337,67 @@ implTryTakeSnapshot ::
...
-          snapManager
-          (ldbSnapshotPolicy env)
-      (`SnapCounters` 0) . Just <$> maybe getMonotonicTime (pure . snd) mTime
-    else
-      pure $ SnapCounters (fst <$> mTime) nrBlocks
+
+      delayBeforeSnapshotting <- getRandomDelay (onDiskSnapshotDelayRange (ldbSnapshotPolicy env))
+      let slotsOfHandles =
+            case NonEmpty.nonEmpty $ map fst handles of
+              Nothing -> error "impossible: empty handles, see pattern-match above"
```
jasagredo: Same as above

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -421,6 +468,10 @@ data LedgerDBEnv m l blk = LedgerDBEnv
   -- in tests can release such resources. These are the resource keys for the
   -- LSM session and the resource key for the BlockIO interface.
   , ldbGetVolatileSuffix :: !(GetVolatileSuffix m blk)
+  , ldbLastSnapshotRequestedAt :: !(StrictTVar m (Maybe Time))
```
jasagredo: Same as above

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/LedgerSnapshots.hs
```diff
@@ -0,0 +1,525 @@
...
+import Test.Util.Orphans.IOLike ()
+import Test.Util.QuickCheck
+import Test.Util.TestBlock
+import Test.Util.Tracer (recordingTracerTVar)
+
+tests :: TestTree
+tests =
+  testGroup
+    "LedgerSnapshots"
+    [ testProperty "InMemV1" $ prop_ledgerSnapshots inMemV1
```
jasagredo, line 71: We can omit V1Mem, it is unused. If anything we can check LMDB

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/LedgerSnapshots.hs
```diff
@@ -0,0 +1,525 @@
...
+
+  for_ (tsBlocksToAdd testSetup) \blk -> do
+    ChainDB.addBlock_ chainDB Punishment.noPunishment blk
+    threadDelay 1
+    atomically $ isEmptyTMVar isSnapshottingTMVar >>= guard
+
+  toutImmutableTip <-
+    AF.castAnchor . AF.anchor <$> atomically (ChainDB.getCurrentChain chainDB)
+  toutTrace <- getTrace
+  toutFinalSnapshots <- LedgerDB.defaultListSnapshots lgrHasFS -- TODO
```
jasagredo: Alex already mentions passing the snapshot manager around. Let's do that

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
geo2a: Done.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
geo2a: We don't, I've removed it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
geo2a: I'll bring it back.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
geo2a: Done.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
geo2a: Done.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
geo2a: Renamed to `ldbLastSuccessfulSnapshotRequestedAt`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
geo2a: Renamed to `ldbLastSuccessfulSnapshotRequestedAt`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
geo2a: I've brought back `NumOfDiskSnapshots` as a newtype.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1.hs
geo2a: Good idea, done.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
geo2a: Done.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/LedgerSnapshots.hs
geo2a: These tests run in IOSim. Can we test LMDB in IOSim?

jasagredo: Aha good catch. We can't. Let's use V1Mem which is a mem mirror of LMDB, in hopes that we can get rid of all this soon.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/LedgerSnapshots.hs
geo2a: I've tried threading the snapshot manager through, but it feels like a lot of ceremony for zero benefit. Do we want test anything but the default logic here?

jasagredo (approved):
Thanks for addressing the comments!

PR #1576, opened 2025-07-02
LedgerDB.StateMachine test: actually test rollbacks

amesgen:
This PR fixes a bug in the LedgerDB state machine test that caused us to never generate non-trivial rollbacks, only chain extensions.

The first commit adds appropriate command labelling, demonstrating the problem:
```
Rollback depths (526356 in total):
100.0000% 0
```

The second commit fixes the bug causing this: The precondition is overzeleaos, so all generated `ValidateAndCommit` actions that have a positive rollback are rejected:
```
Actions rejected by precondition (584429 in total):
100.0000% ValidateAndCommit
```

The third commit fixes a trivial bug in the testing infrastructure where rollbacks were not accounted for.

With this PR, we now get a variety or rollback depths:

```
Rollback depths (898533 in total):
44.2447% 0
25.9243% 1
14.6631% 2
 7.9446% 3
 4.3239% 4
 2.1126% 5
 0.7868% 6
```


jasagredo (approved):
Very nice and localized fix. Thanks Alex!

PR #1583, opened 2025-07-08
Peras chain weight microbenchmark

geo2a:
Closes https://github.com/tweag/cardano-peras/issues/61

This PR contains a micro benchmark for Peras chain weight calculation as implemented by the by the `Ouroboros.Consensus.Storage.PerasCertDB.API.boostedWeightForFragment` function.

### DIFF
--- a/docs/website/contents/for-developers/Benchmarks.md
+++ b/docs/website/contents/for-developers/Benchmarks.md
@@ -1,6 +1,11 @@
 # Consensus benchmarks
 
 We are in the process of adding component level microbenchmarks for Consensus.
+
+We check for regressions in performance on CI.
+
+## Mempool Benchmark
+
 We started with microbenchmarks for adding transactions to the mempool. The
 mempool benchmarks can be run using the following command.
 
@@ -8,4 +13,10 @@ mempool benchmarks can be run using the following command.
 cabal new-run ouroboros-consensus:mempool-bench
 ```
 
-We check for regressions in performance on CI. We might publish benchmark results in this site shortly.
+## ChainSync Client Benchmark
+
+To aid the refactoring of the ChainSync client, we added a benchmark for it in [PR#823](https://github.com/IntersectMBO/ouroboros-consensus/pull/823). The benchmark could be invoked as follows:
+
+```sh
+cabal new-run ouroboros-consensus:ChainSync-client-bench -- 10 10
+```
--- a/docs/website/contents/for-developers/Benchmarks.md
+++ b/docs/website/contents/for-developers/Benchmarks.md
@@ -20,3 +20,13 @@ To aid the refactoring of the ChainSync client, we added a benchmark for it in [
 ```sh
 cabal new-run ouroboros-consensus:ChainSync-client-bench -- 10 10
 ```
+
+## PerasCertDB Benchmark
+
+We have a microbenchmark for the boosted chain fragment weight calculation, which could be run as follows:
+
+```sh
+cabal run ouroboros-consensus:PerasCertDB-bench -- +RTS -T -A32m -RTS
+```
+
+We request GHC runtime system statistics with `-T` to get a memory usage estimate, and also request a large nursery with `-A32m` to minimise garbage collection. See `tasty-bench` [documentation](https://github.com/Bodigrim/tasty-bench?tab=readme-ov-file#troubleshooting) for more tips.
--- a/ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
+++ b/ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
@@ -1,0 +1,102 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
+
+-- | This module contains benchmarks for Peras chain weight calculation as implemented by
+--   the by the 'Ouroboros.Consensus.Storage.PerasCertDB.API.boostedWeightForFragment'
+--   function.
+--
+--   We benchmark the calculation on a static sequence of chain fragments of increasing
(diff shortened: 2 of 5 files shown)


geo2a:
Closes https://github.com/orgs/tweag/projects/78/views/1

docs/website/contents/for-developers/Benchmarks.md
```diff
@@ -1,11 +1,22 @@
...
 mempool benchmarks can be run using the following command.
 
 ```sh
 cabal new-run ouroboros-consensus:mempool-bench
 ```
 
-We check for regressions in performance on CI. We might publish benchmark results in this site shortly.
+## ChainSync Client Benchmark
+
+During the work on Ouroboros Genesis, we added a ChainSync benchmark which uses test blocks:
```
amesgen, line 18: I don't think this work was done as part of the Genesis SoW FWIW (IIRC #823 was prompted by Nick working with the ChainSync client in #525 and then being motivated to do some refactorings for maintainability).

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
```diff
@@ -0,0 +1,81 @@
...
+import Data.Map.Strict qualified as Map
+import Numeric.Natural (Natural)
+import Ouroboros.Consensus.Block (PerasWeight (PerasWeight))
+import Ouroboros.Consensus.Storage.PerasCertDB.API
+  ( PerasWeightSnapshot (..)
+  , boostedWeightForFragment
+  )
+import Ouroboros.Network.AnchoredFragment qualified as AF
+import Test.Tasty.Bench
+import Test.Util.TestBlock (TestBlock)
```
amesgen, line 24: Hmm, for purely comparative measurements, `Test.Util.TestBlock` should be fine, but for absolute estimations (useful for deciding how much we need to optimize this stuff), it is a bit degenerate, eg because `HeaderHash TestBlock ~ NonEmpty Word64` which is pretty different from what Cardano blocks use as a header hash, namely (short) bytestrings.

`Test.Ouroboros.Storage.TestBlock` is a bit more realistic (it uses `Int`s as hashes); we could also define a super simple benchmark block here, but we might need to implement a bunch of classes.

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
+
+-- | This module contains benchmarks for Peras chain weight calculation as implemented by
+--   the by the 'Ouroboros.Consensus.Storage.PerasCertDB.API.boostedWeightForFragment'
+--   function.
+--
+--   We benchmark the calculation on a static sequence of chain fragments of increasing
+--   length, ranging from 0 to around 2160, with a sampling rate of 100. The chain fragments
```
amesgen, line 9: In the ChainSync client, while syncing, we usually have fragments of length `k + 3*k = 8640` (because usually contain the entire volatile candidate (`k` headers) as well as one forecast worth of blocks (on average `3*k` headers).

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
```diff
@@ -0,0 +1,81 @@
...
+ where
+  genesisFragment :: AF.AnchoredFragment TestBlock
+  genesisFragment = AF.Empty AF.AnchorGenesis
+
+  addSuccessorBlock :: AF.AnchoredFragment TestBlock -> AF.AnchoredFragment TestBlock
+  addSuccessorBlock = \case
+    AF.Empty _ -> (AF.Empty AF.AnchorGenesis) AF.:> (TestBlock.firstBlock 0)
+    (xs AF.:> x) -> (xs AF.:> x) AF.:> TestBlock.successorBlock x
+
+-- | Given a chain fragment, construct a weight snapshot where there's a boosted block every 300 slots
```
amesgen, line 67: Currently, this benchmark creates chains where every slot has a block, so every 300th block is boosted.

With the (draft) Peras parameters for mainnet, we would have a Peras boost block every 90 slots (if everything goes well), so due to `f=1/20` we would expect every 4-5 blocks under nominal chain density.

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
```diff
@@ -0,0 +1,81 @@
...
+ where
+  -- NOTE: we do not use the 'env' combinator to set up the test data since
+  --       it requires 'NFData' for 'AF.AnchoredFragment'. While the necessary
+  --       instances could be provided, we do not think is necessary for this
+  --       benchmark, as the input data is rather small.
+  inputs :: [(Natural, (PerasWeightSnapshot TestBlock, AF.AnchoredFragment TestBlock))]
+  inputs =
+    getEveryN (samplingRate benchmarkParams) $
+      take (fromIntegral $ k benchmarkParams) $
+        zip [0 ..] $
```
amesgen, line 46:
```suggestion
      zip [0 .. k benchmarkParams] $
```

amesgen (commented):
Looks great, some comments on increasing "realism"

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
geo2a: :golfing: :+1:

ouroboros-consensus/bench/PerasCertDB-bench/Main.hs
geo2a: Good point, fixed!

PR #1587, opened 2025-07-11
storage-test: move TestBlock to testlib

geo2a:
This PR moves the `Test.Ouroboros.Storage.TestBlock` module form `storage-test` to `unstable-consensus-testlib` so that it could be used in a wider context. Specifically, I'd like to use it for Peras-related benchmarking in https://github.com/IntersectMBO/ouroboros-consensus/pull/1583


amesgen (approved):
Long-term, we of course want #562, but this is still useful for now :+1:

PR #1591, opened 2025-07-16
Refactor: `SelectView` = `BlockNo` × `TiebreakerView`

amesgen:
This PR does not change any behavior; most of the diff is purely mechanical.

Previously, the `SelectView`s of all `ConsensusProtocol`s contained the `BlockNo` as the most important criterion for comparing chains, with only tiebreaking behavior (among chains of equal length) being different across different protocols.

This PR makes this explicit: `SelectView` is now longer an associated type family of `ConsensusProtocol`, but rather an ordinary data type
```haskell
data SelectView p = SelectView
  { svBlockNo :: !BlockNo
  , svTiebreakerView :: !(TiebreakerView p)
  }
```
where `TiebreakerView` is now an associated type family of `ConsensusProtocol`.

This PR also removes the `WithBlockNo` type that the HFC was already using to always attach a `BlockNo` to its `HardForkSelectView`. The code can be simpler now that `SelectView`s always automatically contain an explicit `BlockNo`.

Also see https://github.com/IntersectMBO/ouroboros-consensus/pull/1542#discussion_r2205299482

## Primary motivation

The main motivation for this is to enable weighted chain selection in Peras (https://github.com/tweag/cardano-peras/issues/62), which can now be easily modeled by adding the weight of a chain to the block number component, at least morally; we probably want to eventually have more type safety here to make it hard to "forget" to account for the weight, such as defining
```haskell
data WeightedSelectView p = WeightedSelectView
  { wsvWeight :: !PerasWeight
  , wsvTiebreakerView :: !(TiebreakerView p)
  }
```

If the only goal were to minimize the diff, then it would also be possible to add a function `(BlockNo -> BlockNo) -> SelectView p -> SelectView p`, but it seems more honest to "properly" extract out the block number.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -270,4 +264,31 @@ instance Ord sv => ChainOrder (SimpleChainOrder sv) where
...
+  }
+  deriving stock Generic
+
+deriving stock instance Show (TiebreakerView p) => Show (SelectView p)
+deriving stock instance Eq (TiebreakerView p) => Eq (SelectView p)
+
+instance NoThunks (TiebreakerView p) => NoThunks (SelectView p)
+
+-- | First compare block numbers, then compare the 'TiebreakerView'.
+deriving stock instance Ord (TiebreakerView p) => Ord (SelectView p)
```
fraser-iohk, line 282: should this be an explicitly written-out `Ord` instance? it would be bad if someone decided to reorder the fields in `SelectView` and catastrophically broke chain selection lol

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -270,4 +265,36 @@ instance Ord sv => ChainOrder (SimpleChainOrder sv) where
...
   preferCandidate _cfg ours cand = ours < cand
 
+deriving via SimpleChainOrder () instance ChainOrder ()
 deriving via SimpleChainOrder BlockNo instance ChainOrder BlockNo
+
+data SelectView p = SelectView
+  { svBlockNo :: !BlockNo
+  , svTiebreakerView :: !(TiebreakerView p)
+  }
+  deriving stock Generic
```
dnadales, line 275: Probably unrelated to this PR, but I was wondering if we could elaborate on why the block number (which will become a weigth) and the notion of a tie-breaker cannot be unified into a single weight notion.

dnadales (commented):
Looks good. I left a couple of comments. Let me know when the TODOs are addressed and a new review round is required 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -32,23 +26,20 @@ import Ouroboros.Consensus.TypeFamilyWrappers
...
-  -- chains even across eras, as the chain ordering is fully captured by
-  -- 'SelectView' and its 'ChainOrder' instance.
+-- | How to compare chains of equal length across eras.
+data AcrossEraTiebreaker :: Type -> Type -> Type where
+  -- | No preference.
+  NoTiebreakerAcrossEras :: AcrossEraTiebreaker x y
+  -- | Two eras using the same 'TiebreakerView', so use the corresponding
+  -- (identical) tiebreaker.
   --
   -- We use the 'ChainOrderConfig' of the 'SelectView' in the newer era (with
```
geo2a, line 36:
```suggestion
  -- We use the 'ChainOrderConfig' of the 'TiebreakerView' in the newer era (with
```

geo2a (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
amesgen: I am not exactly sure what unification you have in mind, is it something different than what we have in current main, where everybody defines SelectView as a "single" thing determining the chain order for themselves?

> Do we want to keep the block number and tiebreaker separately because of the separation that we currently have between sorting candidates and preferring a chain?

It is unrelated to that distinction; the motivation is 

 1. (minor) to make it explicit that `BlockNo` must be the primary way of comparing chains; even before this PR, `ConsensusProtocol`s didn't have the freedom to use arbitrary `SelectView`s, and
 2. nicely supports weighted chain comparisons, see "Primary motivation" in the PR description, as well as #1594.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -205,7 +211,7 @@ instance TranslateProto singleProto singleProto where
   translateChainDepState _ = id
 
 -- | The chain order of some type; in the Consensus layer, this will always be
--- the 'SelectView' of some 'ConsensusProtocol'.
+-- the 'TiebreakerView'/'SelectView' of some 'ConsensusProtocol'.
```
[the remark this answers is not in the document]
amesgen: Because we have instances for both of them, where `ChainOrder SelectView` uses `ChainOrder TiebreakerView`. I explicitly mentioned this also here :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
dnadales: FTR, we discussed this offline with @amesgen. The key constraint we have is that we rely on a transitive `Ord` instance when selecting candidates. Therefore, we cannot unify the sorting of `Views` in a single function.  I think that we have is indeed two comparison functions that operate on `SelectView`s. It's just that we're explicit about the part of the field that is used for breaking ties (which I think makes sense)

PR #1603, opened 2025-07-22
Support LSM in snapshot-converter and streaming ledger tables

### LINKED ISSUE #1592: Add LSM to snapshot-converter and db-analyser
In case I defer it on the integration PR, it is necessary that we add support for the snapshot-converter to create LSM trees snapshots and for db-analyser to process the chain with LSM trees.

### LINKED ISSUE #1602: Streaming snapshot-converter
Currently snapshot-converter loads the whole UTxO set into memory and then dumps it in the output format. Instead, we should provide a way of producing and consuming streams of UTxOs.

This will later be used in #1574

jasagredo:
Closes #1592 
Closes #1602

### DIFF
--- a/ouroboros-consensus-cardano/README.md
+++ b/ouroboros-consensus-cardano/README.md
@@ -20,6 +20,8 @@ This package also contains a few executables:
 
 * `app/immdb-server.hs`: serve an immutable DB via ChainSync and BlockFetch.
 
+* `app/snapshot-converter.hs`: converts snapshots among different storage formats.
+
 ### Assertions
 
 Our top level `cabal.project` enables assertions in both our local packages
@@ -378,3 +380,23 @@ To point a node to a running ImmDB server, use a topology file like
   ]
 }
 ```
+
+
+## snapshot-converter
+
+## About
+
+This tool converts snapshots among the different backends supported by the node.
+
+## Running the tool
+
+Invoking the tool follows the same simple pattern always:
+
+```sh
+cabal run snapshot-converter -- <IN> <OUT> --config /path/to/cardano/config.json
+```
+
+The `<IN>` and `<OUT>` parameters depend on the input and output format, receiving options:
+- `--mem-in PATH`/`--mem-out PATH` for InMemory
+- `--lmdb-in PATH`/`--lmdb-out PATH` for LMDB
+- `--lsm-database-in DB_PATH --lsm-snapshot-in PATH`/`--lsm-database-out DB_PATH --lsm-snapshot-out PATH` for LSM-trees.


ouroboros-consensus-cardano/changelog.d/20250904_111240_javier.sagredo_lsm_sc_da.md
jasagredo, line 1: This is deliberately empty because StreamingLedgerTables it is only used by snapshot-converter

ouroboros-consensus/changelog.d/20250904_111212_javier.sagredo_lsm_sc_da.md
jasagredo, line 1: This is deliberately empty because StreamingLedgerTables it is only used by snapshot-converter

geo2a (commented):
In general, the PR looks great, but I have complains about documentation and testing. 

-   The `Ouroboros.Consensus.Util.StreamingLedgerTables` needs mode docs.
    -   I feel like the `Yield` and `Sink` type aliases deserve a haddock, which would explain what they do and why they are structured like they are.
    -   What are the `Int` fields in the data constructors of `YieldArgs`? Probably chunk sizes?
    -   Same for `SinkArgs`, every field needs a haddock
    -   It would also be great to have a haddock for function in the yield/sink section, explaining in a few words what they do, i.e. read/write a chunk. I know, it's all "in the types", but I think a short doc string would still be beneficial.
-   need to update <https://github.com/IntersectMBO/ouroboros-consensus/blob/main/docs/website/contents/references/miscellaneous/utxo-hd/migrating.mdx#convert-the-existing-ledger-snapshots-with-snapshot-converter>, as the CLI interface of `snapshot-converter` has changed
-   I see that we do not have integration tests for `snapshot-converter`. It's indeed probably not necessary to test it in CI. However, I think it would be nice to have a short description of the manual testing scenario, i.e. something like:
    -   take an in-memory/LMDB snapshot with the node of version lower than X
    -   convert to the new version with the following command
    -   run the node of version X or newer to make sure the snapshot is picked up

jasagredo:
Georgy and I briefly discussed the first point in his review last week. These are solved in later PRs in this same train of PRs, so I will omit adding the documentation now.

jasagredo:
Due to other things that are out of my control (like ledger serialization changes) this pattern:
> I see that we do not have integration tests for snapshot-converter. It's indeed probably not necessary to test it in CI. However, I think it would be nice to have a short description of the manual testing scenario, i.e. something like:
>    - take an in-memory/LMDB snapshot with the node of version lower than X
>    - convert to the new version with the following command
>    - run the node of version X or newer to make sure the snapshot is picked up

will not always be applicable. `snapshot-converter` is only an "upgrader" in 10.4.x to 10.5.x, the UTxO-HD release. Afterwards it is again just a converter.

I think the description I added in the readme is enough, because I think snapshot-converter itself is (1) small (2) ad-hoc.

PR #1620, opened 2025-08-05
Integrate KES agent functionality into ouroboros-consensus

fraser-iohk:
includes the following squashed commit messages: 

- Update to use newest cardano-crypto-class with unsound pure KES implementation

- Use mlocked KES

- Add KES agent connectivity

- Rebase cleanup

- Handle drop-key messages from KES Agent

- Provide KESAgentClientTrace to BlockForging

- Revert change to MockCrypto and require DSIGN only when running the KES agent

- Bump kes-agent SRP to remove SerDoc dependency

# Description

this PR might resolve.

Also note that:


ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
```diff
@@ -382,16 +386,21 @@ instance
   Protocol info
 -------------------------------------------------------------------------------}
 
+type instance KESTracer (ShelleyBlock proto era) = KESAgentClientTrace
```
jasagredo, line 389: I don't see why this is necessary, if it is always the same for all Shelley Blocks and it is only used in the ShelleyBasedHardFork and the Embed.Binary modules.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Api/Protocol/Types.hs
```diff
@@ -64,10 +68,16 @@ instance IOLike m => Protocol m ByronBlockHFC where
...
     ( inject $ protocolInfoByron params
-    , pure . map inject $ blockForgingByron params
+    , \_ -> pure . map inject $ blockForgingByron params
     )
 
-instance (CardanoHardForkConstraints StandardCrypto, IOLike m) => Protocol m (CardanoBlock StandardCrypto) where
+instance
+  ( CardanoHardForkConstraints StandardCrypto
+  , IOLike m
+  , MonadKESAgent m
```
jasagredo, line 77: Should `MonadKESAgent` be part of `IOLike`?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Run.hs
```diff
@@ -158,7 +159,7 @@ synthesize genTxs DBSynthesizerConfig{confOptions, confShelleyGenesis, confDbDir
           flavargs
           $ ChainDB.defaultArgs
 
-    forgers <- blockForging
+    (_, forgers) <- allocate registry (const $ mkForgers nullTracer) (mapM_ BlockForging.finalize)
```
jasagredo, line 162: We rely on the registry finalizing these keys. Are we manually finalizing them somewhere? If so, we should call `release` on the `ResourceKey`.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -87,6 +90,7 @@ data Tracers' remotePeer localPeer blk f = Tracers
   , csjTracer ::
       f (TraceLabelPeer remotePeer (CSJumping.TraceEventCsj remotePeer blk))
   , dbfTracer :: f (CSJumping.TraceEventDbf remotePeer)
+  , kesAgentTracer :: f KESAgentClientTrace
```
jasagredo, line 93: This should probably appear in the changelog too

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -496,9 +496,13 @@ forkBlockForging ::
...
-  forkLinkedWatcher registry label $
-    knownSlotWatcher btime $
-      \currentSlot -> withRegistry (\rr -> withEarlyExit_ $ go rr currentSlot)
+  forkLinkedWatcherFinalize
+    registry
+    label
+    ( knownSlotWatcher btime $
+        \currentSlot -> withRegistry (\rr -> withEarlyExit_ $ go rr currentSlot)
+    )
+    (finalize blockForging)
```
jasagredo, line 505: Did we store this `blockForging` in some registry? Should we instead `release` it from there?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -134,19 +149,46 @@ data HotKey c m = HotKey
...
       (KES.Signable (KES c) toSign, HasCallStack) =>
-      toSign -> m (KES.SignedKES (KES c) toSign)
+      toSign ->
+      m (KES.SignedKES (KES c) toSign)
   -- ^ Sign the given @toSign@ with the current signing key.
   --
   -- PRECONDITION: the key is not poisoned.
   --
   -- POSTCONDITION: the signature is in normal form.
+  , forget :: m ()
```
jasagredo, line 171: If we don't want the user to call this directly, perhaps we should also use `_` at the end?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -157,66 +199,184 @@ sign = sign_
...
+        modifyMVar_ varKESState $ poisonState
+
+  forM_ initialStateMay $ \(newOCert, newKey, evolution, startPeriod) ->
+    set newOCert newKey evolution startPeriod
+
+  finalizer' <- case keyThreadMay of
+    Just keyThread -> do
+      keyThreadAsync <- async $ do
+        labelThisThread "HotKey receiver"
+        keyThread set unset
```
jasagredo, line 329: This is convoluted, we spawn a new thread to which we give the `set` and `unset` functions, but I didn't find what that thread is supposed to do. Seems we always call this with `Nothing` as the `keyThreadMay`?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -240,16 +254,67 @@ instance Crypto c => ChainOrder (PraosTiebreakerView c) where
...
+  HotKey.mkHotKey
+    ocert
+    sk
+    startPeriod
+    maxKESEvolutions
+instantiatePraosCredentials maxKESEvolutions tr (PraosCredentialsAgent _ path) = do
+  HotKey.mkDynamicHotKey
+    maxKESEvolutions
+    ( Just $ \handleKey handleDrop -> do
+        runKESAgentClient tr path handleKey handleDrop
```
jasagredo, line 315: Ah I see now that here we call with a `Just`.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -240,16 +254,67 @@ instance Crypto c => ChainOrder (PraosTiebreakerView c) where
...
+    )
+    (pure ())
```
jasagredo, line 317: Shouldn't the finalizer do _something_?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Api/Protocol/Types.hs
fraser-iohk: it could in theory, but MonadKESAgent is currently (in theory) Praos-specific and defined in `ouroboros-consensus-diffusion` (which is admittedly a bit weird) -- I can try to move it into `ouroboros-consensus` if that would make more sense

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
fraser-iohk: it's only a type family so `KESAgentClientTrace` can be defined in `-protocols`, but we never do anything with a  `KESTracer blk` except pass them around -- I've changed it to just be an unconstrained type variable in `Embed.Binary`

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
fraser-iohk: afaik this should be fine -- we don't need to instruct the KES agent to forget the key since that occurs at the `BlockForging` level, and the `HotKey`'s finalizer already cancels the async it spawns

PR #1622, opened 2025-08-05
Nixify cabal-docspec check

amesgen:
Currently, this check is only run on GHA on `main` (so it isn't run on PRs). This is fine as we currently only use it to render some images in the UTxO HD docs, but for Peras, we actually want to use it for checking the correctness of some API examples.

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -2,6 +2,7 @@ inputs: final: prev:
 
 let
   inherit (prev) lib;
+  fs = lib.fileset;
   inherit (final) haskell-nix;
 
   forAllProjectPackages = cfg: args@{ config, lib, ... }: {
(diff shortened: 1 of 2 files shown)


nix/haskell.nix
```diff
@@ -74,4 +75,31 @@ let
...
+      fileset = fs.unions [
+        (fs.fileFilter (f: f.hasExt "cabal" || f.hasExt "hs") ./..)
+      ];
+    };
+
+    nativeBuildInputs = [
+      final.fd
+      final.cabal-docspec
+      (hsPkgs.ghcWithPackages
+        (ps: [ ps.latex-svg-image ] ++ lib.filter (p: p ? components.library)
```
amesgen, line 93: It should also be possible to avoid mentioning `latex-svg-image` here (eg by making sure we also depend on test-only dependencies), but this doesn't seem like a priority.

dnadales (approved):

PR #1624, opened 2025-08-08
GetFilteredVoteDelegatees: special-case empty query set

rrruko:
This implements an API change that would be convenient for [amaru](https://github.com/pragma-org/amaru) development.

The motivation is that there are currently several queries that accept a set of values to be used to filter the result. For example, `GetDRepState` takes a set of DRep credentials and returns the corresponding DRep states. Many of these queries implement a special case where passing an empty set behaves as if the set of all possible values was queried, thus returning "all of the data".

Likewise, it would be useful to enable querying the full map from `GetFilteredVoteDelegatees` by simply passing an empty set, instead of the current behavior, which is to just return an empty map in response to an empty set. This change would make the query strictly more useful since there's no point making a query that you already know will return an empty map, whereas querying for the full map under the current behavior is non-trivial.

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20250811_072709_rukokarasu_get_filtered_vote_delegatees.md
+++ b/ouroboros-consensus-cardano/changelog.d/20250811_072709_rukokarasu_get_filtered_vote_delegatees.md
@@ -1,0 +1,22 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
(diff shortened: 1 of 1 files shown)


amesgen (commented):
Thanks, looks great! 

 - Can you add this behavior to the Haddock string of `GetFilteredVoteDelegatees`?
 - We require signed commits. (If that is annoying, I can also rebase and add my signature; you retain authorship credit :smile: )
 - Can you add a [changelog fragment](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/docs/website/contents/for-developers/ReleaseProcess.md#installing-scriv)?

Also happy to do this for you if this is too annoying.

amesgen (approved):

PR #1640, opened 2025-08-22
Fix leaking handles in uncommitted forkers

jasagredo:


ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -280,6 +285,7 @@ instance StateModel Model where
       Action Model (ExtLedgerState TestBlock EmptyMK, ExtLedgerState TestBlock EmptyMK)
     Init :: SecurityParam -> Action Model ()
     ValidateAndCommit :: Word64 -> [TestBlock] -> Action Model ()
+    OpenAndCloseForker :: Action Model ()
```
amesgen, line 288: Add a comment that we only test this for its potential "side effects"?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs
```diff
@@ -165,6 +165,17 @@ implForkerCommit env = do
...
                 -- Finally, close the anchor of @lseq@ (which is a duplicate of
                 -- the head of @olddb'@).
+                --
+                -- Note if the resource registry used to create the Forker is
+                -- ephemeral as it should (one created on each Chain selection
+                -- or each Forging loop iteration), this first duplicated state
+                -- will be closed by the resource registry closing down, so this
+                -- will be a double release, which is fine. We prefer keeping
+                -- this action just in case some client doesn't respect the
+                -- condition of registries being ephemeral.
```
amesgen, line 175: Hmm, do we document the requirement that one should use a `Forker` with a short-lived `ResourceRegistry` somewhere? This is especially weird as eg the LedgerDB state machine test uses a long-lived registry for all forkers.

Alternatively, I think this comment could be rephrased to say that we close this just to release resources as early as possible, even if it might be closed soon after due to the registry getting closed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -757,7 +758,7 @@ newForker h ldbEnv rr st = do
   let tr = LedgerDBForkerEvent . TraceForkerEventWithKey forkerKey >$< ldbTracer ldbEnv
   traceWith tr ForkerOpen
   lseqVar <- newTVarIO . LedgerSeq . AS.Empty $ st
-  (k, toRelease) <- allocate rr (\_ -> newTVarIO (pure ())) (readTVarIO Monad.>=> id)
+  (k, toRelease) <- allocate rr (\_ -> newTVarIO (close (tables st))) (Monad.join . readTVarIO)
```
amesgen, line 761: Just to make sure: The `close` here actually also isn't strictly necessary, as it will also be closed by the `ResourceRegistry`; just like in `implForkerCommit`, right?

The LedgerDB state machine test is failing if I change this back to `pure ()`, we *don't* use a short-lived `ResourceRegistry` for `OpenAndCloseForker`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/Forker.hs
jasagredo: I have reworded it a bit

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -573,13 +572,14 @@ getEnvSTM (LDBHandle varState) f =
...
+  ResourceRegistry m ->
   (LedgerSeq m l -> t (StateRef m l)) ->
   m (t (StateRef m l))
-getStateRef ldbEnv project =
+getStateRef ldbEnv reg project =
   RAWLock.withReadAccess (ldbOpenHandlesLock ldbEnv) $ \() -> do
     tst <- project <$> readTVarIO (ldbSeq ldbEnv)
     for tst $ \st -> do
-      tables' <- duplicate $ tables st
+      (_, tables') <- allocate reg (\_ -> duplicate $ tables st) close
```
amesgen, line 582: Hmm, do we also want to return the `ResourceKey` here, such that the table can be closed via it? I think that would be more in line with our general push to close things via their `ResourceKey` if they were allocated within a registry. Especially if the registry is long-lived.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -573,13 +572,14 @@ getEnvSTM (LDBHandle varState) f =
 getStateRef ::
   (IOLike m, Traversable t) =>
   LedgerDBEnv m l blk ->
+  ResourceRegistry m ->
```
amesgen, line 575: Update the documentation of the function, which currently still says

> The caller is responsible for closing the handle.

amesgen (approved):
LGTM

Not in scope for this PR, but the story around safely and correctly releasing resources for forkers is quite complicated (even more so with this PR), with some parts being handled by a `ResourceRegistry` and some parts by a special `StrictTVar m (m ())`. I wonder whether we could make some general improvements here; maybe the concept of a temporary registry a la [`runWithTempRegistry`](https://hackage-content.haskell.org/package/resource-registry-0.1.1.0/docs/Control-ResourceRegistry.html#v:runWithTempRegistry) could be used to streamline this?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: While this is true, it turns out that the closing function of a handle will already track the resource key inside, so just by passing the handle we already will invoke the `release` on it.

However, for being more explicit I think your suggestion is ok. Let me implement it

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Done

geo2a (approved):

PR #1643, opened 2025-08-25
Define SnapshotManager

jasagredo:


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -205,6 +206,25 @@ data MetadataErr
...
+-- | Management of snapshots for the different LedgerDB backends.
+--
+-- The LedgerDB V1 takes snapshots in @ReadLocked m@, hence the two different
+-- @m@ and @n@ monad types.
+data SnapshotManager m n blk st = SnapshotManager
+  { listSnapshots :: m [DiskSnapshot]
+  , deleteSnapshot :: DiskSnapshot -> m ()
+  , takeSnapshot ::
+      Maybe String ->
+      -- \^ The (possibly empty) suffix for the snapshot name
```
geo2a, line 218: Nit: fourmolu does not like the invalid haddock here. Consider converting `-- \^` to `--`.

geo2a (approved):
Seems like a sensible refactoring, I do not see anything that does not look right :+1:

PR #1651, opened 2025-08-28
Gate noThunksInvariant behind expensive-invariants build flag

### LINKED ISSUE #1649: Allow to check `TVar` invariants
The goal of this issue is to easily enable custom invariant checking for `StrictTVar`s etc. that are always tested on CI (and not just nightly CI).

### Status quo

We use the [`strict-checked-vars`](https://github.com/IntersectMBO/io-classes-extra/blob/main/strict-checked-vars/README.md) library to get variants of `StrictTVar`s/`StrictMVar`s which allow one to check invariants every time their value is updated. This library provides two flags, [`checktvarinvariants` and `checkmvarinvariants`](https://github.com/IntersectMBO/io-classes-extra/blob/b2356a34b30a9431bbcdf1cf7a5b71e00e9e5bbe/strict-checked-vars/strict-checked-vars.cabal#L40-L48) (disabled by default), which one needs to enable to actually get invariant-checking behavior.

 - When building for production (in [`cardano-node`](https://github.com/IntersectMBO/cardano-node)), these flags are disabled, causing no runtime overhead.
 - When building for our nightly nothunks tests, we enable them
   https://github.com/IntersectMBO/ouroboros-consensus/blob/698f67ea5a967b000b2f63a46b272afdd970aaa2/.github/workflows/cabal.project.no-thunks.local#L4

By default, the invariant we check is the [`nothunks`](https://github.com/input-output-hk/nothunks/) property; also see the [description of this module](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus/Ouroboros-Consensus-Util-NormalForm-StrictTVar.html).

Notably, in regular CI, we don't actually enable these invariants, because the `nothunks` checks are way too expensive for the regular CI.

### Desired change

It is annoying that we can't have invariants for eg `TVar`s which are checked in regular CI, as well as when running tests locally. Therefore, the goal is to gate the `nothunks` checks behind an additional flag (say `expensive-invariants`), which is only enabled on nightly CI, and to then enable `checktvarinvariants` and `checkmvarinvariants` even on regular CI.

There are also one more place where we call [`checkInvariant`](https://github.com/IntersectMBO/io-classes-extra/blob/b2356a34b30a9431bbcdf1cf7a5b71e00e9e5bbe/strict-checked-vars/src/Control/Concurrent/Class/MonadSTM/Strict/TVar/Checked.hs#L168) with `nothunks` checks apart from the `TVar`s/`MVar`s; this should also be gated behind the new `expensive-invariants` flag:
https://github.com/IntersectMBO/ouroboros-consensus/blob/698f67ea5a967b000b2f63a46b272afdd970aaa2/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L2193
(https://github.com/IntersectMBO/ouroboros-consensus/pull/1615 will add something similar)

---

Motivated by https://github.com/IntersectMBO/ouroboros-consensus/pull/1615#discussion_r2303642573

agustinmista:
This PR closes #1649 by:

* Defining the `expensive-invariants` build flag (disabled by default)
* Gating the use of `unsafeNoThunks` (the expensive check to avoid) behind the `ENABLE_EXPENSIVE_INVARIANTS` CPP flag (implied by `expensive-invariants`)
* Breaking `noThunkInvariants` into its own module to reduce the amount of CPP needed to compile without warnings 
* Tweaking the cabal overrides used in CI to:
  + Run with `+checktvarinvariants` and `+checkmvarinvariants` by default (via `asserts.cabal`)
  + Run `no-thunks` CI with `+expensive-invariants` in addition to the default flags

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -28,6 +28,11 @@ flag asserts
   manual: False
   default: False
 
+flag expensive-invariants
+  description: Enable checks for expensive invariants
+  manual: True
+  default: False
+
 common common-lib
   default-language: Haskell2010
   ghc-options:
@@ -45,6 +50,9 @@ common common-lib
   if flag(asserts)
     ghc-options: -fno-ignore-asserts
     cpp-options: -DENABLE_ASSERTIONS
+
+  if flag(expensive-invariants)
+    cpp-options: -DENABLE_EXPENSIVE_INVARIANTS
 
 common common-test
   import: common-lib
@@ -284,6 +292,7 @@ library
     Ouroboros.Consensus.Util.LeakyBucket
     Ouroboros.Consensus.Util.MonadSTM.NormalForm
     Ouroboros.Consensus.Util.MonadSTM.StrictSVar
+    Ouroboros.Consensus.Util.NormalForm.Invariant
     Ouroboros.Consensus.Util.NormalForm.StrictMVar
     Ouroboros.Consensus.Util.NormalForm.StrictTVar
     Ouroboros.Consensus.Util.Orphans
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/NormalForm/StrictMVar.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/NormalForm/StrictMVar.hs
@@ -9,8 +9,12 @@
 -- 'NoThunks' invariant. See 'newMVarWithInvariant' and
 -- 'newEmptyMVarWithInvariant'.
 --
--- Use the @checkmvarinvariants@ cabal flag from the @strict-checked-vars@
--- package to enable or disable invariant checks at compile time.
+-- Due to their expensive nature, checks for the 'NoThunks' invariant are
+-- disabled by default and can be enabled at compile-time via the
+-- @expensive-invariants@ flag from the @ouroboros-consensus@ package. To
+-- disable invariant checks entirely (i.e., both 'NoThunks' and custom ones),
+-- use the @checkmvarinvariants@ cabal flag from the @strict-checked-vars@
+-- package.
 --
 -- The exports of this module (should) mirror the exports of the
 -- "Control.Concurrent.Class.MonadMVar.Strict.Checked" module from the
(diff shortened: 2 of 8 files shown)


.github/workflows/cabal.project.default.local
```diff
@@ -1,5 +1,8 @@
 package cardano-crypto-praos
   flags: -external-libsodium-vrf
 
+package strict-checked-vars
+  flags: +checktvarinvariants +checkmvarinvariants
```
amesgen, line 5: I think we can move this to [`asserts.cabal`](https://github.com/IntersectMBO/ouroboros-consensus/blob/6b130b1eb11db3b367c7cc3cd3bb2bb4496d6b8a/asserts.cabal).

This file is only used by the GitHub Actions CI; while `asserts.cabal` is also used on Hydra and when building/testing locally.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/NormalForm/StrictMVar.hs
```diff
@@ -85,7 +92,11 @@ newEmptyMVarWithInvariant inv =
 -------------------------------------------------------------------------------}
 
 noThunksInvariant :: NoThunks a => a -> Maybe String
+#ifdef CHECK_NOTHUNKS_INVARIANT
 noThunksInvariant = fmap show . unsafeNoThunks
+#else
+noThunksInvariant = const Nothing
+#endif
```
amesgen, line 99: >  It might be a good idea to extract noThunksInvariant into its own module Ouroboros.Consensus.Util.NormalForm.Invariant. This would allows us to reduce the amount of CPP needed to conditionally take the different build paths without raising warnings.

Agreed, that sounds like a good idea :+1:

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -46,6 +51,10 @@ common common-lib
     ghc-options: -fno-ignore-asserts
     cpp-options: -DENABLE_ASSERTIONS
 
+  if flag(expensive-invariants)
+    cpp-options:
+      -DCHECK_NOTHUNKS_INVARIANT
```
amesgen, line 56: > Will there ever be more than one expensive-invariant? If not, I'd rather consolidate the build and CPP flag under the "same" name (either checknothunksinvariants/CHECK_NOTHUNK_INVARIANTS, or expensive-invariants/EXPENSIVE_INVARIANTS).

Good point, I think I would be weakly in favor of the latter option. I hope we won't need to introduce additional levels of differentiation here :sweat_smile:

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -28,6 +28,11 @@ flag asserts
   manual: False
   default: False
 
+flag expensive-invariants
+  description: Enable checks for expensive invariants
+  manual: False
```
amesgen, line 33: I think we want the flag to manual, to avoid the solver turning it on automatically somehow (though I don't think it would happen currently as it doesn't control any dependencies).

geo2a (approved):

Issue #94, opened 2023-05-22
Draft a CSJ liveness argument


amesgen:
One of the main things we have to convince ourselves is that CSJ does not reduce liveness, i.e. that CSJ does not change the property that we can not be delayed for prolonged periods from concluding we are caught up.

The goal of this ticket is to draft an argument for this, which also includes discussing how different approaches of disabling CSJ (a fixed slot before the wall clock; or when peers tell they have no more headers) affect it.

PR #1655, opened 2025-09-02
[WIP] Add basic API for certificate validation

amesgen:
Closes https://github.com/tweag/cardano-peras/issues/94


agustinmista:
Hey @amesgen @tbagrel1 

The PR still needs a lot of work, but I think it would be good to get some early feedback on [b31fc2a](https://github.com/IntersectMBO/ouroboros-consensus/pull/1655/commits/b31fc2a522c341e26092c7ce75dac7f7ffdf512a). I pushed a bunch of changes adapting call sites to use `ValidatedPerasCert`s whenever needed, as implied by the two TODOs left in [3425d67](https://github.com/IntersectMBO/ouroboros-consensus/pull/1655/commits/3425d6764e16f0971f313a51defad05ac108b4b2):

* Adapting `PerasCertDB.addCert` to take a `ValidatedPerasCert`, and
* Adapting `ChainDB.addPerasCertAsync` to take a `ValidatedPerasCert`

Based on my current understanding, there is a tension between keeping either `PerasCert`s or `ValidatedPerasCert`s around, depending on what we want to enforce:
1) Validating certs early on, and using those everywhere
  + Somewhat better assurance that we don't ever try to operate on potentially-invalid certs
  + Potentially useful in the cases where the extra evidence they carry allows us to avoid recomputing stuff (e.g. the block's boost)
  + Needs a "big" refactor to adapt all call sites, including the "generation" of the now needed validation evidence in some property tests 
2) Validating certs when needed, but using the original `PerasCert` everywhere
  + Might open the door for bugs caused by operating on invalid certs, not sure how likely this is, though
  + When useful, validated certs (or at least their useful evidence) might need to be passed around manually all the way down to the consumer
  + Likely as smaller refactor, as we can probably unwrap `ValidatedPerasCert`s in a couple of places and keep the original type signature

Do you think (1) is the right approach here? Or is (2) something more akin to what we would want at this stage in the dev cycle?

amesgen:
I think (1) is indeed the right approach, the benefits seem compelling even at this stage.

(In the standup, we just briefly talked about some of the implications this has once we store certificates on disk, such as whether we also want to store the weight boost of a certificate on disk, or recompute it on startup. But this is not important for now.)

> including the "generation" of the now needed validation evidence in some property tests

This is actually a desired outcome of this change, see the last paragraph in the issue description of #94. To keep this PR minimal, still always using `boostPerCert` sounds good as you currently do; enriching the generators could be follow-up work.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -68,8 +69,15 @@ makePerasCertPoolWriterFromChainDB ::
...
     { opwObjectId = perasCertRound
-    , opwAddObjects =
-        mapM_ $ ChainDB.addPerasCertAsync chainDB
+    , opwAddObjects = \certs -> do
+        let perasRoundNo = PerasRoundNo 999
+        -- TODO ^ not the value we want, but the value we have
+        case sequence (validatePerasCert defaultPerasCfg perasRoundNo <$> certs) of
+          Left validationError ->
+            throw (userError (show validationError))
+          -- TODO   ^ refine this exception
```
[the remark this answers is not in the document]
agustinmista: There's also the question of whether we want to short-circuit and throw an exception upon the first validation error (at it happens now with `sequence`), or to be pedantic and collect all the validation errors in the current batch and report them at once.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -215,7 +217,9 @@ implGetCertSnapshot PerasCertDbEnv{pcdbVolatileState} =
 implGarbageCollect ::
   forall m blk.
   (IOLike m, StandardHash blk) =>
-  PerasCertDbEnv m blk -> SlotNo -> m ()
+  PerasCertDbEnv m blk ->
+  SlotNo ->
+  m ()
```
agustinmista, line 222: This signature didn't change, so I'm not sure why this change crept in.

If fourmolu running on some lenient/not-strict mode?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/PerasCertDB/StateMachine.hs
```diff
@@ -97,7 +97,9 @@ instance StateModel Model where
         -- Do not add equivocating certificates.
         AddCert cert -> all p model.certs
          where
-          p cert' = perasCertRound cert /= perasCertRound cert' || cert == cert'
+          p cert' =
+            perasCertRound cert /= perasCertRound (validatedPerasCert cert')
+              || cert == validatedPerasCert cert'
```
agustinmista: This predicate might be a bit contrived now, as we are comparing properties between both validated and unvalidated certs.

Maybe I'm being too paranoid, but this could potentially break some internal invariant if the cert validation becomes more complex in the future -- thinking concretely whether the notion of equality between `ValidatedPerasCert` will always imply equality between their inner `PerasCerts`.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/PerasCertDB/StateMachine.hs
```diff
@@ -117,7 +119,12 @@ instance RunModel Model (StateT (PerasCertDB IO TestBlock) IO) where
       lift $ PerasCertDB.closeDB perasCertDB
     AddCert cert -> do
       perasCertDB <- get
-      lift $ PerasCertDB.addCert perasCertDB cert
+      let validatedCert =
+            ValidatedPerasCert
+              { validatedPerasCert = cert
+              , validatedPerasCertBoost = boostPerCert
```
agustinmista: Another hardcoded use of the boost to keep in mind

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -23,32 +24,32 @@ import Ouroboros.Consensus.Util.IOLike
 makePerasCertPoolReaderFromSnapshot ::
   (IOLike m, StandardHash blk) =>
   STM m (PerasCertSnapshot blk) ->
-  ObjectPoolReader PerasRoundNo (PerasCert blk) PerasCertTicketNo m
+  ObjectPoolReader PerasRoundNo (ValidatedPerasCert blk) PerasCertTicketNo m
```
amesgen, line 27: This type signature should stay unchanged

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/PerasCert.hs
amesgen, line 1: This file should also not have any changes

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -144,7 +144,7 @@ implAddCert ::
   , StandardHash blk
   ) =>
   PerasCertDbEnv m blk ->
-  PerasCert blk ->
+  ValidatedPerasCert blk ->
```
amesgen, line 147: Currently, this function uses `boostPerCert`; rather, it should now use `validatedPerasCertBoost` of the argument here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -68,9 +76,31 @@ makePerasCertPoolWriterFromChainDB ::
...
-        mapM_ $ ChainDB.addPerasCertAsync chainDB
+    , opwAddObjects = \certs -> do
+        let validatedCerts = validatePerasCerts certs
+        mapM_ (ChainDB.addPerasCertAsync chainDB) validatedCerts
     , opwHasObject = do
         certSnapshot <- atomically $ ChainDB.getPerasCertSnapshot chainDB
         pure $ PerasCertDB.containsCert certSnapshot
     }
+
+-- TODO following the same approach as 'ChainSyncClientException', but why not using GADT syntax?
```
amesgen, line 87: No particular reason, feel free to use that one if you prefer it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -647,7 +647,9 @@ addBlockRunner fuse cdb@CDB{..} = forever $ do
             ChainSelAddPerasCert cert _varProcessed ->
               traceWith cdbTracer $
                 TraceAddPerasCertEvent $
-                  PoppedPerasCertFromQueue (perasCertRound cert) (perasCertBoostedBlock cert)
+                  PoppedPerasCertFromQueue
+                    (perasCertRound (validatedPerasCert cert))
+                    (perasCertBoostedBlock (validatedPerasCert cert))
```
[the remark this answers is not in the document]
agustinmista: Decided to use a type class to overload these for both kinds of certificates.

amesgen (commented):
LGTM (I can't approve because it is "my" PR), only minor comments

PR #1660, opened 2025-09-03
Address PeerSharing & KeepAlive todos in ouroboros-consensus-diffusion

coot:


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs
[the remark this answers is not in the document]
coot: True, that's by accident, but I'll include it in a separate commit.  It just makes things more readable.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs
```diff
@@ -652,10 +652,12 @@ doValidateKESSignature praosMaxKESEvo praosSlotsPerKESPeriod stakeDistribution o
 
   currentIssueNo :: Maybe Word64
   currentIssueNo
-    | Map.member hk ocertCounters = Map.lookup hk ocertCounters
-    | Set.member (coerceKeyRole hk) (Map.keysSet stakeDistribution) =
-        Just 0
-    | otherwise = Nothing
+    | r@Just {} <- Map.lookup hk ocertCounters
+      = r
+    | Set.member (coerceKeyRole hk) (Map.keysSet stakeDistribution)
```
amesgen, line 657: Just in case you feel like it because you already are in the area; also happy to merge what you have:
```suggestion
  | Map.member (coerceKeyRole hk) stakeDistribution
```
Could also use the `Alternative` instance of `Maybe` :thinking:

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -540,7 +540,7 @@ data Apps m addr bCS bBF bTX bKA bPS a b = Apps
 --
 -- They don't depend on the instantiation of the protocol parameters (which
 -- block type is used, etc.), hence the use of 'RankNTypes'.
-data ByteLimits bCS bBF bTX bKA = ByteLimits
+data ByteLimits addr bCS bBF bTX bKA bPS = ByteLimits
```
amesgen, line 543: I think it would be more in line to remove this and rather quantify in the type of `blPeerSharing`, also see the comment

```haskell
-- They don't depend on the instantiation of the protocol parameters (which
-- block type is used, etc.), hence the use of 'RankNTypes'.
```
just above.

```suggestion
data ByteLimits bCS bBF bTX bKA bPS = ByteLimits
```

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs
coot: :thinking: `Alternative` - how?

amesgen: Ie like this:
```haskell
  currentIssueNo =
    Map.lookup hk ocertCounters
      <|> (0 <$ Map.lookup (coerceKeyRole hk) stakeDistribution)
```

coot: I agree, it's a bit too far.

PR #1662, opened 2025-09-03
Add some details on the expected file formats

abailly:
This tiny PR documents in a bit more details how the arguments to `db-synthesizer` program look like. In particular, it wasn't obvious what the "bulk credentials" structure was and I got it only after looking at the example. HTH.


jasagredo (approved):
I was going to mention that it might be better to edit the Readme instead, but the readme explicitly says to run the tool without arguments to see the usage instructions. Therefore I am ok with this change.

PR #1673, opened 2025-09-11
[Peras 1] Introduce basic types to support Peras

tbagrel1:
This pull request introduces some base datatypes and functionality required for **Peras**, along with supporting tests, benchmarks, and documentation. 

Specifically, it adds:  

* `PerasWeight` and `PerasCert`, which associate a weight with a block.  
* `PerasWeightSnapshot`, a type for capturing the known weight state of blocks
* Functions for computing weights of chains and fragments.  
* Tests, microbenchmarks, documentation, and runnable examples for weight-related utilities.  

---

### **Peras types and weight computation**  

* Added module `Ouroboros.Consensus.Block.SupportsPeras`, defining types and logic for Peras weights and certificates.  
  * Key elements are re-exported via `Ouroboros.Consensus.Block`.  
* Added module `Ouroboros.Consensus.Peras.Weight`, introducing `PerasWeightSnapshot` and weight computation functions for chains and fragments.  

---

### **Documentation**  

* Added glossary entries for **Peras**, **weight boost**, and **chain fragment weight** in `glossary.md`.  
* Created a new benchmarks page explaining how to run Peras-related microbenchmarks and their purpose.  

---

### **Tests and benchmarks**  

* Introduced benchmark suite `PerasCertDB-bench`, including a weight computation microbenchmark (`bench/PerasCertDB-bench/Main.hs`).  
* Added module `Test.Consensus.Peras.WeightSnapshot` with property-based tests for weight computation on chains and fragments.  

---

### **Regression**

The new code added with this PR is not actually used, so it shouldn't introduce regressions.

PR #1674, opened 2025-09-11
[Peras 2] Introduce `PerasCertDB` and related tests

tbagrel1:
This pull request introduces the `PerasCertDB` datatype, used to hold and access all `PerasCert`s known by a node that could affect chain selection.

In the mid/near-term future, this will be enriched to persist certificates across restarts, and probably also to store historical certificates.

---

### **Peras types**

- Added modules `Ouroboros.Consensus.Storage.PerasCertDB{,.API,.Impl}`, notably defining the types`PerasCertDB`, `PerasCertSnapshot` (read-only snapshot of certs contained in the DB), and `AddPerasCertResult`; alongside their respectives methods

---

### **Tests**

- Added modules `Test.Ouroboros.Storage.PerasCertDB{,.StateMachine,.Model}` for q-s-m testing of the `PerasCertDB` datatype. The corresponding tests have been included in the test suite defined by `Test.Ouroboros.Storage`

---

### **Regression**

As the [previous PR](https://github.com/IntersectMBO/ouroboros-consensus/pull/1673), all the components introduced by this PR are not used at the moment by the rest of the code, so no regression is expected.

### DIFF
--- a/ouroboros-consensus/changelog.d/20250917_101832_thomas.bagrel_peras_cert_db.md
+++ b/ouroboros-consensus/changelog.d/20250917_101832_thomas.bagrel_peras_cert_db.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Add modules `Ouroboros.Consensus.Storage.PerasCertDB{,.API,.Impl}`, notably defining the types`PerasCertDB`, `PerasCertSnapshot` (read-only snapshot of certs contained in the DB), and `AddPerasCertResult`; alongside their respective methods
+- Add modules `Test.Ouroboros.Storage.PerasCertDB{,.StateMachine,.Model}` for q-s-m testing of the `PerasCertDB` datatype. The corresponding tests are included in the test suite defined by `Test.Ouroboros.Storage`


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -0,0 +1,56 @@
...
+  }
+  deriving NoThunks via OnlyCheckWhnfNamed "PerasCertDB" (PerasCertDB m blk)
+
+data AddPerasCertResult = AddedPerasCertToDB | PerasCertAlreadyInDB
+  deriving stock (Show, Eq)
+
+data PerasCertSnapshot blk = PerasCertSnapshot
+  { containsCert :: PerasRoundNo -> Bool
+  -- ^ Do we have the certificate for this round?
+  , getCertsAfter :: PerasCertTicketNo -> [(ValidatedPerasCert blk, PerasCertTicketNo)]
```
dnadales, line 48: The result type might deserve some explanation. I guess for each tuple `(cert, ticketNo)` in the result, `ticketNo` is the ticket number of `cert`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -0,0 +1,56 @@
...
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Consensus.Storage.PerasCertDB.API
+  ( PerasCertDB (..)
+  , AddPerasCertResult (..)
+
+    -- * 'PerasCertSnapshot'
+  , PerasCertSnapshot (..)
+  , PerasCertTicketNo
```
dnadales, line 12: Why do we need ticket numbers for Peras certificates?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -0,0 +1,297 @@
...
+      keepCert cert =
+        pointSlot (getPerasCertBoostedBlock cert) >= NotOrigin slot
+
+{-------------------------------------------------------------------------------
+  Implementation-internal types
+-------------------------------------------------------------------------------}
+
+-- | Volatile Peras certificate state, i.e. certificates that could influence
+-- chain selection by boosting a volatile block.
+data PerasVolatileCertState blk = PerasVolatileCertState
```
dnadales, line 249: Perhaps we should add some invariant checking here. For instance:

- Should the different maps have all the same size? 
- For every certificate in `pvcsCerts`, is there a corresponding weight in `pvcsWeightByPoint`?
- Should the range of `pvcsCerts` and `pvcsCertsByTicket` be the same? 
- etc.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -0,0 +1,297 @@
...
+    <&> forgetFingerprint
+    <&> \PerasVolatileCertState
+           { pvcsCerts
+           , pvcsCertsByTicket
+           } ->
+        PerasCertSnapshot
+          { containsCert = \r -> Map.member r pvcsCerts
+          , getCertsAfter = \ticketNo ->
+              let (_, certs) = Map.split ticketNo pvcsCertsByTicket
+               in [(cert, tno) | (tno, cert) <- Map.toAscList certs]
```
dnadales, line 213: Why not returning the pair of maps? Do we need them in list form?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/PerasCertDB/StateMachine.hs
```diff
@@ -0,0 +1,147 @@
...
+        frequency
+          [ (1, pure $ Some CloseDB)
+          , (20, Some <$> genAddCert)
+          , (20, pure $ Some GetWeightSnapshot)
+          , (5, Some . GarbageCollect . SlotNo <$> arbitrary)
+          ]
+    | otherwise = pure $ Some OpenDB
+   where
+    genAddCert = do
+      roundNo <- PerasRoundNo <$> arbitrary
```
dnadales, line 69: By using `arbitrary` on this data type, aren't we generating few cases in which there's a collision in `roundNo`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -0,0 +1,297 @@
...
+        writeTVar pcdbVolatileState $
+          WithFingerprint
+            PerasVolatileCertState
+              { pvcsCerts =
+                  Map.insert roundNo cert pvcsCerts
+              , -- Note that the same block might be boosted by multiple points.
+                pvcsWeightByPoint =
+                  addToPerasWeightSnapshot boostedPt (getPerasCertBoost cert) pvcsWeightByPoint
+              , pvcsCertsByTicket =
+                  Map.insert pvcsLastTicketNo' cert pvcsCertsByTicket
```
bladyjoker, line 175: Thinking generally, `certsByTicket` sounds like a specific "index" over a certs "table" which has a `roundNo` as a primary key (if I read this correctly).

Sometimes I opt to maintain indices such that they point to object 'identifiers', rather than objects themselves `Map TicketNo RoundNo` so to speak. You probably see that there's 2 lookups that need to happen here to go from ticketNo to a cert.

Sometimes that makes sense, for example 1-to-may indices (which this isn't), if there are updates (which there aren't), and if objects are copied (which here they aren't...me thinks).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
amesgen: Yeah, that sounds like a good refactoring; I don't think this is really leaking implementation details.

PR #1678, opened 2025-09-12
[Peras 3] Make `ChainDB` aware of `PerasCertDB`, and switch to weighted chain selection

tbagrel1:
# Description

This PR introduces:

 - **Weighted chain comparisons**: We compare chains based on their *weight* instead of their length. Non-trivial weight is given by Peras certificates as introduced in https://github.com/IntersectMBO/ouroboros-consensus/pull/1673.
 - **Weighted chain selection**: Select the *weightiest* chain instead of the *longest* chain.
 - **Weighted immutability criterion**: Define the immutable tip in terms of *weight* instead of length.

### Commits

The commits are intended to be reviewed individually:

 - *ChainDB: expose PerasCertDB functionality* 

   The ChainDB maintains a PerasCertDB internally, and exposes most of its functionality through its public API. This is analogous to how the LedgerDB is managed by the ChainDB.

 - *`SecurityParam`: mention weighted nature*

   Purely a documentation update to mention the richer dynamics under Peras, plus a small helper function.

 - *O.C.Peras.Weight: add `takeVolatileSuffix`*

   In Praos, the volatile suffix of a chain is defined to be the `k` most recent blocks. Analogously, in Peras, the volatile suffix of a chain is defined to be the longest suffix with weight at most `k`.

   This commits adds an appropriate function (via a binary search), as well as documentation and tests.

 - *ChainDB: define `getCurrentChain` in terms of weight*

   This makes use of the previous commit in the ChainDB. Note that https://github.com/IntersectMBO/ouroboros-consensus/pull/1619 guarantees that the LedgerDB automatically uses the same notion of immutability.

   Note that this means that the immutable tip will be less than `k` blocks behind the tip when Peras is working well, ie every Peras round is giving rise to a certificate. Concretely, for plausible parameters, ie a round length of $U = 90$ slots and a Peras boost of $B = 15$, the length of the volatile suffix decreases from $k=2160$ to

   $$\frac{k}{1+\frac{B}{U\cdot f}} \approx 499$$

   on average.

 - *ChainDB.StateMachine: check immutable tip monotonicity*

   To make sure that the more refined immutability criterion doesn't introduce any surprises, we add a postcondition to the ChainDB q-s-m test that the immutable tip never recedes. 

 - *GSM: allow `candidateOverSelection` to be stateful* and *ChainSel: make `rollbackExceedsSuffix` weight-aware*

   Preparatory refactorings for when the chain comparison will become weighted and hence additionally depend on the `PerasWeightSnapshot`.

 - *Peras.SelectView: initialize, expose `WeightedSelectView`*

   This introduces the notion of a *weighted* `SelectView`, making use of #1591 

   https://github.com/IntersectMBO/ouroboros-consensus/blob/90131b282b4e002e9ddc0d7196043c50b255a1d4/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/SelectView.hs#L35-L48

   This will used in place of the "normal" `SelectView`. Eventually, we could also remove the normal `SelectView` and replace it with `WeightedSelectView`. However, to keep this already big PR more focused, we propose to do this in the future.

 - *Introduce weighted chain comparisons*

   The core change in this commit is in the `O.C.Util.AnchoredFragment` to the `preferAnchoredCandidate` and `compareAnchoredFragments` functions. These now

    - have a slightly strengthened precondition, namely that they intersect (as we otherwise can't meaningfully compare their weight), that we ensure is satisfied anywhere, and
    - take `PerasWeightSnapshot` as an additional argument.

   We keep the old implementation in case the `PerasWeightSnapshot` is empty; which is semantically unnecessary, but is a trivial way to make sure that no performance regression is introduced.

   The rest of the diff of this commit is simply due to adapting the respective call sites in a rather straightforward fashion. 

 - *Integrate weighted BlockFetch decision logic*

   This commit plugs the weighted chain comparison logic into the BlockFetch decision logic. This relies on a (merged, but not yet released, hence the s-r-p) Network change, see https://github.com/IntersectMBO/ouroboros-network/pull/5161 for a detailed description.

 - *ChainDB: implement chain selection for Peras certificates*

   We allow new Peras certificates to be added to the ChainDB (and therefore to the managed PerasCertDB). If the certificate is boosting a block that is not on the current selection, we perform chain selection for it, potentially switching to a fork containing it if it now is weightier than our selection.

 - *MockChainSel: switch to weighted chain selection*

   This logic is for example used in tests.

 - *ChainDB q-s-m: test weighted chain selection*

   We enrich the model ChainDB implementation with weighted chain selection, and add a new command with a simple generator for adding certificates. We also enrich labelling with the `TagSwitchedToShorterChain` tag that shows that we sometimes *do* switch to a shorter chain; which would be a bug without Peras.

   A follow-up PR (https://github.com/IntersectMBO/ouroboros-consensus/pull/1670) will further improve the generators of this test.

---

### **Regression**

As of today, there is no way to generate certificates, so `PerasCertDB` is always empty, and the change to the chain selection algorithm is therefore invisible: in the absence of certificates, the Peras weight of a chain is equal to its length. Therefore, this PR does not introduce any semantic changes.

Furthermore, from a performance standpoint, care is taken to ensure that if there are no certificates, we use the unweighted Praos logic, meaning that there is no change in performance when Peras is disabled.

### DIFF
--- a/ouroboros-consensus-diffusion/changelog.d/20250919_095435_thomas.bagrel_weighted_chain_selec.md
+++ b/ouroboros-consensus-diffusion/changelog.d/20250919_095435_thomas.bagrel_weighted_chain_selec.md
@@ -1,0 +1,3 @@
+### Breaking
+
+- In module `Ouroboros.Consensus.Node.GSM`, `GSMView` now has a monadic `getCandidateOverSelection :: STM m (selection -> chainSyncState -> CandidateVersusSelection)` instead of the previous pure `candidateOverSelection`. This is due to the fact that chain comparisons now depend on the set of Peras certificates (if Peras is enabled).
--- a/ouroboros-consensus/changelog.d/20250917_144846_nicolas.bacquey_weighted_chain_selec.md
+++ b/ouroboros-consensus/changelog.d/20250917_144846_nicolas.bacquey_weighted_chain_selec.md
@@ -1,0 +1,7 @@
+### Breaking
+
+- Make the `ChainDB` aware of the `PerasCertDB`, and modify the chain selection function accordingly. In practice, it means that the candidate fragment is now selected based on its Peras weight, instead of its length.
+
+  Note that if Peras is disabled (which is the default), there is no observable difference.
+
+- Add module `Ouroboros.Consensus.Peras.SelectView`, which introduces a `WeightedSelectView` to correctly measure the length of a chain fragment.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/SelectView.hs
```diff
@@ -0,0 +1,137 @@
...
+import Ouroboros.Consensus.Protocol.Abstract
+import Ouroboros.Network.AnchoredFragment (AnchoredFragment)
+import qualified Ouroboros.Network.AnchoredFragment as AF
+
+{-------------------------------------------------------------------------------
+  Weighted select views
+-------------------------------------------------------------------------------}
+
+-- | Information from a non-empty chain fragment for a weighted chain comparison
+-- against other fragments with the same anchor.
```
dnadales: I wonder if this would be a good point to explain why the fragments having the same anchor is important in Peras.

dnadales (commented):
I reviewed the following commits:

- [X] ChainDB: expose PerasCertDB functionality
- [X] SecurityParam: mention weighted nature
- [X] O.C.Peras.Weight: add takeVolatileSuffix
- [X] ChainDB: define getCurrentChain in terms of weight
- [X] ChainDB.StateMachine: check immutable tip monotonicity
- [X] GSM: allow candidateOverSelection to be stateful
- [X] ChainSel: make rollbackExceedsSuffix weight-aware
- [X] Peras.SelectView: initialize, expose WeightedSelectView

I'll try to continue tomorrow

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -83,29 +91,44 @@ import Ouroboros.Network.Protocol.LocalStateQuery.Type
...
   , ConsensusProtocol (BlockProtocol blk)
   ) =>
   ChainDbEnv m blk ->
   STM m (AnchoredFragment (HeaderWithTime blk))
-getCurrentChainWithTime CDB{..} =
-  AF.anchorNewest (unNonZero k) . icWithTime <$> readTVar cdbChain
+getCurrentChainWithTime cdb@CDB{..} =
+  getCurrentChainLike cdb $ icWithTime <$> readTVar cdbChain
+
+getCurrentChainLike ::
```
geo2a: I think it's worth adding a comment explaining that this function is a generisation of `getCurrentChain(WithTime)`. Also, the haddoc of `getCurrentChain` needs to be updates to mention weights.

I'm attaching a patch with my suggestions:
``` patch
diff --git a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
index 353bae1f6..7b0af0a4e 100644
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
@@ -75,16 +75,16 @@ import Ouroboros.Network.Protocol.LocalStateQuery.Type

 -- | Return the last @k@ headers.
 --
--- While the in-memory fragment ('cdbChain') might temporarily be longer than
+-- While the in-memory fragment ('cdbChain') might temporarily have more weight than
 -- @k@ (until the background thread has copied those blocks to the
--- ImmutableDB), this function will never return a fragment longer than @k@.
+-- ImmutableDB), this function will never return a fragment heavier than @k@.
 --
 -- The anchor point of the returned fragment will be the most recent
 -- \"immutable\" block, i.e. a block that cannot be rolled back. In
 -- ChainDB.md, we call this block @i@.
 --
--- Note that the returned fragment may be shorter than @k@ in case the whole
--- chain itself is shorter than @k@ or in case the VolatileDB was corrupted.
+-- Note that the returned fragment may have weight less than @k@ in case the whole
+-- chain itself weights less than @k@, or in case the VolatileDB was corrupted.
 -- In the latter case, we don't take blocks already in the ImmutableDB into
 -- account, as we know they /must/ have been \"immutable\" at some point, and,
 -- therefore, /must/ still be \"immutable\".
@@ -113,6 +113,8 @@ getCurrentChainWithTime ::
 getCurrentChainWithTime cdb@CDB{..} =
   getCurrentChainLike cdb $ icWithTime <$> readTVar cdbChain

+-- | This function is the generalised helper for 'getCurrentChain' and 'getCurrentChainWithTime'.
+-- See 'getCurrentChain' for the explanation of it's behaviour.
 getCurrentChainLike ::
   forall m blk h.
   ( IOLike m
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -1026,3 +1057,26 @@ data TraceIteratorEvent blk
...
+data TraceAddPerasCertEvent blk
+  = -- | The Peras certificate from the given round boosting the given block was
+    -- added to the queue. The size of the queue is included.
+    AddedPerasCertToQueue PerasRoundNo (Point blk) (Enclosing' Word)
+  | -- | The Peras certificate from the given round boosting the given block was
+    -- popped from the queue.
+    PoppedPerasCertFromQueue PerasRoundNo (Point blk)
+  | -- | The Peras certificate from the given round boosting the given block was
+    -- too old, ie its slot was older than the current immutable slot (the third
+    -- argument).
```
dnadales: Maybe we want to wrap the the `Point blk` arguments in a newtype to clarify the thing each one denotes, eg `(CurrentImmutableSlot blk)`, where

```haskell
newtype CurrentImmutableSlot blk = CurrentImmutableSlot (Point blk) ...
```

But this might be too heavy handed. I just worry the trace messages might be confusing otherwise.

bladyjoker (commented):
Reviewed:

[x] ChainDB: expose PerasCertDB functionalityy

LGTM

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Weight.hs
```diff
@@ -291,9 +294,87 @@ totalWeightOfFragment weightSnap frag =
...
+-- buried under at least weight @k@ (also counting the weight boost of that
+-- point).
+--
+-- >>> :{
+-- weights :: [(Point Blk, PerasWeight)]
+-- weights =
+--   [ (BlockPoint 2 "foo", PerasWeight 2)
+--   , (GenesisPoint,       PerasWeight 3)
+--   , (BlockPoint 3 "bar", PerasWeight 2)
+--   , (BlockPoint 2 "foo", PerasWeight 2)
```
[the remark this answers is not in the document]
amesgen: Let's add a link to `mkPerasWeightSnapshot`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Config/SecurityParam.hs
```diff
@@ -3,24 +3,39 @@
...
 -- i.e., k == 1: we can roll back at most one block
 --       k == 2: we can roll back at most two blocks, etc
 --
 -- NOTE: This talks about the number of /blocks/ we can roll back, not
 -- the number of /slots/.
+--
+-- In weightiest-chain protocols (Ouroboros Peras), we interpret this as the
+-- maximum amount of weight we can roll back.
+--
+-- i.e. k == 30: we can roll back at most 30 unweighted blocks, or two blocks
```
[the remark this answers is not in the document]
amesgen: Done in c5433d63c2306d191bbcb045d5588886ff2b2961

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/WeightSnapshot.hs
```diff
@@ -59,12 +61,26 @@ prop_perasWeightSnapshot testSetup =
...
           | frag <- tsFragments
           ]
+      , conjoin
+          [ conjoin
+              [ counterexample ("Incorrect volatile suffix for " <> condense frag) $
+                  takeVolatileSuffixReference frag =:= volSuffix
+              , counterexample ("Volatile suffix must be a suffix of" <> condense frag) $
+                  AF.headPoint frag =:= AF.headPoint volSuffix
+                    .&&. AF.withinFragmentBounds (AF.anchorPoint volSuffix) frag
+              , counterexample ("Volatile suffix of " <> condense frag <> " must contain at most k blocks") $
```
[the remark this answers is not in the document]
amesgen: Ah, this is actually already achieved by using the [`le`](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus-unstable-consensus-testlib/Test-Util-QuickCheck.html#v:le) combinator in the line below:
https://github.com/IntersectMBO/ouroboros-consensus/blob/374ef153e20d83ad3d42d850ce560b67034ac578/ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/QuickCheck.hs#L126-L128

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs
```diff
@@ -76,59 +78,67 @@ forksAtMostKBlocks k ours theirs = case ours `AF.intersect` theirs of
...
+            if blockPoint tip == AF.castPoint (AF.anchorToPoint anchor')
+              then EQ
+              else GT
+          (_ :> tip, _ :> tip') ->
+            -- Case 4
+            compare
+              (selectView cfg (getHeader1 tip))
+              (selectView cfg (getHeader1 tip'))
+  | otherwise =
+      case AF.intersect frag1 frag2 of
```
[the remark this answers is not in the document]
amesgen: This is done implicitly by the `ChainOrder (WithEmptyFragment a)` instance.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
amesgen: Adopted in 953f3fb89adb18161ea17d9535efe8b1eda93756, also updated the documentation in the `ChainDB` API record.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -593,6 +604,27 @@ addBlockToAdd tracer (ChainSelQueue{varChainSelQueue, varChainSelPoints}) punish
...
+addPerasCertToQueue tracer ChainSelQueue{varChainSelQueue} cert = do
+  varProcessed <- newEmptyTMVarIO
+  traceWith tracer $ addedToQueue RisingEdge
+  queueSize <- atomically $ do
+    writeTBQueue varChainSelQueue $ ChainSelAddPerasCert cert varProcessed
+    lengthTBQueue varChainSelQueue
+  traceWith tracer $ addedToQueue $ FallingEdgeWith $ fromIntegral queueSize
+  pure
+    AddPerasCertPromise
+      { waitPerasCertProcessed = atomically $ takeTMVar varProcessed
```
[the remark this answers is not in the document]
amesgen: Great catch, using `readTMVar` for idempotency seems much less surprising for users here, adopted in ddb411988e430b692ab9a406156713b6253e3f67

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Weight.hs
```diff
@@ -291,9 +317,89 @@ totalWeightOfFragment weightSnap frag =
...
+    AnchoredFragment h -> -- The length lb suffix.
+    AnchoredFragment h
+  go lb ub lbFrag
+    | lb + 1 == ub = lbFrag
+    | hasAtMostWeightK midFrag = go mid ub midFrag
+    | otherwise = go lb mid lbFrag
+   where
+    mid = (lb + ub) `div` 2
+    midFrag = AF.anchorNewest mid frag
+
```
bladyjoker: In my attempts to reason about this I stumbled on a refactored version that helped my comprehension.

1. I'd consider refactoring with `takeLongestSuffix` such that one can provide property tests on that function alone. I would even argue that it might belong in AnchoredFragment library (especially if you can postulate other useful predicates)?

2. The refactoring moved the `| hasAtMostWeightK frag = frag` condition into `takeLongestSuffix` as it's key to its correctness.

```haskell
takeVolatileSuffix ::
  forall blk h.
  (StandardHash blk, HasHeader h, HeaderHash blk ~ HeaderHash h) =>
  PerasWeightSnapshot blk ->
  -- | The security parameter @k@ is interpreted as a weight.
  SecurityParam ->
  AnchoredFragment h ->
  AnchoredFragment h
takeVolatileSuffix snap secParam frag
  | Map.null $ getPerasWeightSnapshot snap =
      -- Optimize the case where Peras is disabled.
      AF.anchorNewest (unPerasWeight k) frag
  | otherwise = takeLongestSuffix hasAtMostWeightK frag
 where
  k :: PerasWeight
  k = maxRollbackWeight secParam

  hasAtMostWeightK :: AnchoredFragment h -> Bool
  hasAtMostWeightK f = totalWeightOfFragment snap f <= k

takeLongestSuffix ::
  HasHeader h =>
  -- | Monotone predicate (for smaller fragments it's False and for bigger it's True)
  (AnchoredFragment h -> Bool) ->
  AnchoredFragment h ->
  AnchoredFragment h
takeLongestSuffix p frag
  | p frag = frag
  | otherwise = go 0 (fromIntegral $ AF.length frag) (AF.Empty $ AF.headAnchor frag)
 where
  go lb ub lbFrag
    | lb + 1 == ub = lbFrag
    | p midFrag = go mid ub midFrag
    | otherwise = go lb mid lbFrag
   where
    mid = (lb + ub) `div` 2
    midFrag = AF.anchorNewest mid frag -- Take the n newest elements from the anchored sequence.
```

amesgen: Ah, I am seeing https://github.com/IntersectMBO/ouroboros-consensus/pull/1678#discussion_r2444733101 only now, sorry. My version also has a `join` call (plus `fromJust`) that we can't statically prove can't error.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/BlockFetch/ClientInterface.hs
```diff
@@ -258,45 +259,24 @@ mkBlockFetchConsensusInterface
...
-        (WithOrigin BlockNo, WithOrigin SlotNo)
-      anchorBlockNoAndSlot frag =
-        (AF.anchorToBlockNo a, AF.anchorToSlotNo a)
-       where
-        a = AF.anchor frag
+      -- 4. Therefore, we check whether the candidate fragments still intersects
+      --    with our fragment; if not, then it is only a matter of time until the
+      --    ChainSync client disconnects from that peer.
+      case AF.intersectionPoint ours cand of
+        -- REVIEW: Hmm, maybe we want to change 'preferAnchoredCandidates' to
```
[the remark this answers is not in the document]
amesgen: Let's open an issue for this once this PR is merged

cabal.project
```diff
@@ -84,3 +84,18 @@ source-repository-package
...
+    , bin:QuickCheck
+
+source-repository-package
+   type: git
+   location: https://github.com/IntersectMBO/ouroboros-network
+   tag: b07a86ed853b63881b5a83e57508902f1562ac01
+   --sha256: sha256-n/XX0+cQegq2a1cAfmGx30T64eix4oEXzpVEFCKqmg0=
+   subdir:
+     ouroboros-network-api
+     ouroboros-network
```
amesgen: Mention that this isn't on `main`, but rather a commit from `main` cherry-picked to a backported branch

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Config/SecurityParam.hs
```diff
@@ -3,24 +3,42 @@
...
+-- In longest-chain protocols, we interpret this as the number of rollbacks we
+-- support.
 --
 -- i.e., k == 1: we can roll back at most one block
 --       k == 2: we can roll back at most two blocks, etc
 --
 -- NOTE: This talks about the number of /blocks/ we can roll back, not
 -- the number of /slots/.
+--
+-- In weightiest-chain protocols (such as Ouroboros Peras), we interpret this as
```
bladyjoker: Somehow overloading the meaning of a SecurityParameter implicitly makes me feel a bit uncomfortable. We collate the measure of "meters" with that of "grams" so to speak.

`k` in Praos means "count", and `k` in Peras means "weight", and there's an explicit relationship between them. 

Do you see a way to perhaps be more explicit about this? Separating into Praos and Peras sub modules etc.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -1021,3 +1057,26 @@ data TraceIteratorEvent blk
 newtype TraceChainSelStarvationEvent blk
   = ChainSelStarvation (Enclosing' (RealPoint blk))
   deriving (Generic, Eq, Show)
+
+data TraceAddPerasCertEvent blk
+  = -- | The Peras certificate from the given round boosting the given block was
```
bladyjoker: Consider placing these in the Peras subdirectory you have?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs
```diff
@@ -59,76 +60,96 @@ compareHeadBlockNo ::
...
-        -- This case is symmetric to the previous
-        if blockPoint tip == AF.castPoint (AF.anchorToPoint anchor')
-          then EQ
-          else GT
-      (_ :> tip, _ :> tip') ->
-        -- Case 4
-        compare
-          (selectView cfg (getHeader1 tip))
-          (selectView cfg (getHeader1 tip'))
+compareAnchoredFragments cfg weights frag1 frag2
```
bladyjoker: Is there a concise common body based on AF.intersect that can be used in both cases? If so, we can perhaps shorten this and generalize it?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs
```diff
@@ -59,76 +60,96 @@ compareHeadBlockNo ::
...
-  -- | How many blocks can it fork?
-  Word64 ->
-  -- | Our chain.
+-- | Check that we can switch from @ours@ to @theirs@ by rolling back our chain
+-- by at most @k@ weight.
+--
+-- If @ours@ and @cand@ do not intersect, this returns 'False'. If they do
+-- intersect, then we check that the suffix of @ours@ after the intersection has
+-- total weight at most @k@.
+forksAtMostKWeight ::
```
bladyjoker: Similar story, perhaps this could be decomposed into a generalized `forksAtMost` and then further provide the Vanilla version along with Peras version?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs
```diff
@@ -142,27 +163,39 @@ preferAnchoredCandidate ::
...
+              (selectView cfg (getHeader1 ourTip))
+              (selectView cfg (getHeader1 theirTip))
+  | otherwise =
+      case AF.intersect ours cand of
+        Nothing -> error "precondition violated: fragments must intersect"
+        Just (_oursPrefix, _candPrefix, oursSuffix, candSuffix) ->
+          preferCandidate
+            (projectChainOrderConfig cfg)
+            (weightedSelectView cfg weights oursSuffix)
+            (weightedSelectView cfg weights candSuffix)
```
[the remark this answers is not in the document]
amesgen: > Is there a version of this function that for both cases Vanilla/Peras uses the same body? Can the vanilla case be expressed using the AF.intersect as well? Not sure how the cases in vanilla body relates to the Peras cases.

Yes, the long-term goal here is to just always use the Peras logic. We even did this at first while developing on a branch, but switched to this approach (where we transparently use the Praos logic when there are no certificates) to make it trivial to argue that our changes are not introducing any (performance) regressions.

> I suspect there's an opportunity to generalize this function over the 'type of select view'.

Yes, though if sharing as much code as possible is the goal, then using the Peras logic would be the way to go.

PR #1681, opened 2025-09-12
[Peras 5] Adapt the HFC time translation layer for Peras

tbagrel1:
Fixes https://github.com/tweag/cardano-peras/issues/65

## Changes

- Changes in the HFC types:
  * `EraParams` now keeps track of an optional Peras round length.
  * `Bound` now keeps track of an optional Peras round number.
  * In the `Serialise` instances for `EraParams` and `Bound`, the encoders generate different CBOR depending on whether of not the Peras-relate components are present. The decoders act differently depending on the length of the CBOR list.
  * The `EraSummary`, while not modified directly, is now Peras-aware via `EraParams` and `Bound`:
    - in a valid summary, Peras round length must divide the epoch size. See the changes in `EraSummary`s haddock and in `invariantSummary`'s implementation.
- Changes in the HFC time translation queries:
  * Two new top-level queries are exposed from `Ouroboros.Consensus.HardFork.History.Qry`:

    ```haskell
    perasRoundNoToSlot :: PerasRoundNo -> Qry SlotNo
    slotToPerasRoundNo :: SlotNo -> Qry PerasRoundNo
    ```

  * add a roundtrip test that ensures that converting Peras round number to a slot and then back is an identity is added into the `Test.Consensus.HardFork.Summary` module.
  * add a Peras-specific test into `Test.Consensus.HardFork.History` module.

## Suggested review order

Even though this PR affects many files, the bulk of the changes is localised. I suggest the following review order:

- First review the datatype and codec changes in `ouroboros-consensus/Ouroboros/Consensus/HardFork/History/EraParams.hs` and `ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Summary.hs`
- Proceed to review the time translation query changes in `ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs`
- Then review the newly added tests in `consensus-test/Test/Consensus/HardFork/History.hs` and `consensus-test/Test/Consensus/HardFork/Summary.hs`

### DIFF
--- a/ouroboros-consensus-cardano/changelog.d/20250919_101630_thomas.bagrel_hfc_era_peras.md
+++ b/ouroboros-consensus-cardano/changelog.d/20250919_101630_thomas.bagrel_hfc_era_peras.md
@@ -1,0 +1,28 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Add `eraPerasRoundLength` parameters to `{Byron,Shelley}EraParams` structs.
+
+### Non-Breaking
+
+- `EraSummary` is now indirectly Peras-aware via `EraParams`:
+  - In a valid summary, Peras round length must divide the epoch size.
--- a/ouroboros-consensus-diffusion/changelog.d/20250919_101623_thomas.bagrel_hfc_era_peras.md
+++ b/ouroboros-consensus-diffusion/changelog.d/20250919_101623_thomas.bagrel_hfc_era_peras.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Update code using `EraParams` now that it has a new field `eraPerasRoundLength` for Byron and Shelley eras.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
--- a/ouroboros-consensus/changelog.d/20250918_114333_thomas.bagrel_hfc_era_peras.md
+++ b/ouroboros-consensus/changelog.d/20250918_114333_thomas.bagrel_hfc_era_peras.md
@@ -1,0 +1,35 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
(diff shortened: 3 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/EraParams.hs
```diff
@@ -136,17 +146,65 @@ data EraParams = EraParams
   , eraSlotLength :: !SlotLength
   , eraSafeZone :: !SafeZone
   , eraGenesisWin :: !GenesisWindow
+  , eraPerasRoundLength :: !(PerasEnabled PerasRoundLength)
```
dnadales: Why not using `Maybe` instead? This add cognitive load, as the reader now has to know what `PerasEnabled` is.

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
```diff
@@ -503,7 +519,13 @@ data EventTime = EventTime
   { eventTimeSlot :: SlotNo
   , eventTimeEpochNo :: EpochNo
   , eventTimeEpochSlot :: Word64
+  -- ^ Relative slot withing the current epoch round,
+  --   needed to be able to advance the epoch number
   , eventTimeRelative :: RelativeTime
+  , eventTimePerasRoundNo :: PerasRoundNo
+  , eventTimeSlotInPerasRound :: Word64
+  -- ^ Relative slot withing the current Peras round,
```
dnadales:
```suggestion
  -- ^ Relative slot within the current Peras round,
```

dnadales (approved):

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
agustinmista: Fixed in f0e7a56d23e4a5da846b820e3edf9497d3887fdd

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/EraParams.hs
agustinmista: I don't have a strong opinion about this but, if we used a `Maybe`, then a programmer would still have to reach out to the docs/source code to see what that optional value is encoding. This might be just a tad more self-documenting than that.

geo2a: Initially I used `StrictMaybe`, but in one of the review rounds Alex suggested we used a custom type as it was more descriptive at call-sites.

In tests, I also had to follow a pattern of wrapping the era end into another `Maybe` (see [here](https://github.com/IntersectMBO/ouroboros-consensus/blob/1be66fb83180909e57d678f05460abd143f9b1aa/ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/Summary.hs#L197), which would be two layers of `Maybe`.

I think we have several such small custom types in the code base, so I though it would be alright. I can refactor it back to be a `Maybe` if you have a strong opinion.

PR #1692, opened 2025-09-23
Reintroduce `encodedSizeExpr` in `Header` instance

teodanciu:
# Description

In my previous [PR](https://github.com/IntersectMBO/ouroboros-consensus/pull/1687) , I [removed](https://github.com/IntersectMBO/ouroboros-consensus/pull/1687/files#diff-cc67072c8a9a8cae57404fc5f4ff58c83ef8c6238043df252402a83218ff1219L246) the implementation of `encodedSizeExpr` from the `Header` instance. 
On @lehins advice, I'm reintroducing it - even though it's not clear if it's actually used - it's better to remove it systematically if it turns out that's the case, rather than change it accidentally.


lehins:
> Eg if this value is incorrect, does it impact correctness, or just performance?

@amesgen Nobody knows why that function exists. I've been asking people around for years and yet to find an answer what is the purpose of that function is and who uses it. Hence is my suggestion to completely remove it from the serialization interface

PR #1694, opened 2025-09-25
Bump CHaP index-state, add `kes-agent-crypto` to build-depends

fraser-iohk:
This PR bumps the CHaP index-state to pick up the new version of `kes-agent` and adds `kes-agent-crypto` (now split into a separate package) to build-depends where necessary


ouroboros-consensus-protocol/ouroboros-consensus-protocol.cabal
```diff
@@ -80,6 +80,7 @@ library
     io-classes ^>=1.8.0,
     io-sim,
     kes-agent,
+    kes-agent-crypto,
```
amesgen, line 83: Do we want version bounds?

amesgen (approved):

ouroboros-consensus-protocol/ouroboros-consensus-protocol.cabal
fraser-iohk: I think we probably want version bounds once it's been audited, but at the moment it's fine left free

fraser-iohk: I assume (?) that the audited version will be released as v1.0, so maybe a `<1` on both `kes-agent` and `kes-agent-crypto` would be appropriate

PR #1698, opened 2025-09-29
Peras/experimental-object-diffusion-inbound-v2

tbagrel1:
Co-authored-by: nbacquey <nbacquey@users.noreply.github.com>

# Description

this PR might resolve.

Also note that:

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -193,7 +193,9 @@ library
     Ouroboros.Consensus.MiniProtocol.LocalTxSubmission.Server
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.PerasCert
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Outbound
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.PerasCert
     Ouroboros.Consensus.Node.GsmState
     Ouroboros.Consensus.Node.InitStorage
     Ouroboros.Consensus.Node.NetworkProtocolVersion
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -605,6 +605,7 @@ test-suite consensus-test
     Test.Consensus.MiniProtocol.ChainSync.CSJ
     Test.Consensus.MiniProtocol.ChainSync.Client
     Test.Consensus.MiniProtocol.LocalStateQuery.Server
+    Test.Consensus.MiniProtocol.ObjectDiffusion.Smoke
     Test.Consensus.Peras.WeightSnapshot
     Test.Consensus.Util.MonadSTM.NormalForm
     Test.Consensus.Util.Versioned
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -607,6 +607,7 @@ test-suite consensus-test
     Test.Consensus.MiniProtocol.ChainSync.CSJ
     Test.Consensus.MiniProtocol.ChainSync.Client
     Test.Consensus.MiniProtocol.LocalStateQuery.Server
+    Test.Consensus.MiniProtocol.ObjectDiffusion.PerasCert.Smoke
     Test.Consensus.MiniProtocol.ObjectDiffusion.Smoke
     Test.Consensus.Peras.WeightSnapshot
     Test.Consensus.Util.MonadSTM.NormalForm
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -191,6 +191,9 @@ library
     Ouroboros.Consensus.MiniProtocol.LocalStateQuery.Server
     Ouroboros.Consensus.MiniProtocol.LocalTxMonitor.Server
     Ouroboros.Consensus.MiniProtocol.LocalTxSubmission.Server
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Outbound
     Ouroboros.Consensus.Node.GsmState
     Ouroboros.Consensus.Node.InitStorage
     Ouroboros.Consensus.Node.NetworkProtocolVersion
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -191,7 +191,13 @@ library
     Ouroboros.Consensus.MiniProtocol.LocalStateQuery.Server
     Ouroboros.Consensus.MiniProtocol.LocalTxMonitor.Server
     Ouroboros.Consensus.MiniProtocol.LocalTxSubmission.Server
-    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V1
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Decision
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Policy
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Registry
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.State
+    Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Types
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.PerasCert
     Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Outbound
@@ -351,6 +357,7 @@ library
     primitive,
     psqueues ^>=0.2.3,
     quiet ^>=0.2,
+    random,
     rawlock ^>=0.1.1,
     resource-registry ^>=0.1,
     semialign >=1.1,
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
@@ -1,0 +1,225 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2
+  ( -- * ObjectDiffusion Inbound client
+    objectSubmissionInboundV2
+
+    -- * PeerObjectAPI
+  , withPeer
+  , PeerObjectAPI
+
+    -- * Supporting types
+  , module V2
+  , ObjectChannelsVar
+  , newObjectChannelsVar
+  , ObjectObjectPoolSem
+  , newObjectObjectPoolSem
+  , SharedObjectStateVar
+  , newSharedObjectStateVar
+  , ObjectDecisionPolicy (..)
+  , defaultObjectDecisionPolicy
+  ) where
+
+import Control.Exception (assert)
+import Control.Monad (unless, when)
+import Control.Monad.Class.MonadThrow
+import Control.Monad.Class.MonadTimer.SI
+import Control.Tracer (Tracer, traceWith)
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map.Strict as Map
+import qualified Data.Sequence.Strict as StrictSeq
+import qualified Data.Set as Set
+import Network.TypedProtocol
+import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Policy
+import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Registry
+import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.State
+import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Types as V2
+import Ouroboros.Network.Protocol.ObjectDiffusion.Inbound
+
+-- | A object-submission inbound side (server, sic!).
+--
+-- The server blocks on receiving `ObjectDecision` from the decision logic. If
+-- there are object's to download it pipelines two requests: first for object's second
+-- for objectId's. If there are no object's to download, it either sends a blocking or
+-- non-blocking request for objectId's.
+objectSubmissionInboundV2 ::
+  forall objectId object idx m.
+  ( MonadDelay m
+  , MonadThrow m
+  , Ord objectId
+  ) =>
+  Tracer m (TraceObjectDiffusionInbound objectId object) ->
+  ObjectDiffusionInitDelay ->
+  ObjectDiffusionObjectPoolWriter objectId object idx m ->
+  PeerObjectAPI m objectId object ->
+  ObjectDiffusionServerPipelined objectId object m ()
+objectSubmissionInboundV2
+  tracer
+  initDelay
+  ObjectDiffusionObjectPoolWriter{objectId}
+  PeerObjectAPI
+    { readObjectDecision
+    , handleReceivedObjectIds
+    , handleReceivedObjects
+    , submitObjectToObjectPool
+    } =
+    ObjectDiffusionServerPipelined $ do
+      case initDelay of
+        ObjectDiffusionInitDelay delay -> threadDelay delay
+        NoObjectDiffusionInitDelay -> return ()
+      serverIdle
+   where
+    serverIdle ::
+      m (ServerStIdle Z objectId object m ())
+    serverIdle = do
+      -- Block on next decision.
+      object@ObjectDecision
+        { objectsToRequest = objectsToRequest
+        , objectsToObjectPool = ObjectsToObjectPool{listOfObjectsToObjectPool}
+        } <-
+        readObjectDecision
+      traceWith tracer (TraceObjectInboundDecision object)
+
+      let !collected = length listOfObjectsToObjectPool
+
+      -- Only attempt to add OBJECTs if we have some work to do
+      when (collected > 0) $ do
+        -- submitObjectToObjectPool traces:
+        -- \* `TraceObjectDiffusionProcessed`,
+        -- \* `TraceObjectInboundAddedToObjectPool`, and
+        -- \* `TraceObjectInboundRejectedFromObjectPool`
+        -- events.
+        mapM_ (uncurry $ submitObjectToObjectPool tracer) listOfObjectsToObjectPool
+
+      -- TODO:
+      -- We can update the state so that other `object-submission` servers will
+      -- not try to add these objects to the objectpool.
+      if Map.null objectsToRequest
+        then serverReqObjectIds Zero object
+        else serverReqObjects object
+
+    -- Pipelined request of objects
+    serverReqObjects ::
+      ObjectDecision objectId object ->
+      m (ServerStIdle Z objectId object m ())
+    serverReqObjects object@ObjectDecision{objectsToRequest = objectsToRequest} =
+      pure $
+        SendMsgRequestObjectsPipelined
+          objectsToRequest
+          (serverReqObjectIds (Succ Zero) object)
+
+    serverReqObjectIds ::
+      forall (n :: N).
+      Nat n ->
+      ObjectDecision objectId object ->
+      m (ServerStIdle n objectId object m ())
+    serverReqObjectIds
+      n
+      ObjectDecision{objectIdsToRequest = 0} =
+        case n of
+          Zero -> serverIdle
+          Succ _ -> handleReplies n
+    serverReqObjectIds
+      -- if there are no unacknowledged objectIds, the protocol requires sending
+      -- a blocking `MsgRequestObjectIds` request.  This is important, as otherwise
+      -- the client side wouldn't have a chance to terminate the
+      -- mini-protocol.
+      Zero
+      ObjectDecision
+        { objectIdsToAcknowledge = objectIdsToAck
+        , objectPipelineObjectIds = False
+        , objectIdsToRequest = objectIdsToReq
+        } =
+        pure $
+          SendMsgRequestObjectIdsBlocking
+            objectIdsToAck
+            objectIdsToReq
+            -- Our result if the client terminates the protocol
+            (traceWith tracer TraceObjectInboundTerminated)
+            ( \objectIds -> do
+                let objectIds' = NonEmpty.toList objectIds
+                    objectIdsSeq = StrictSeq.fromList $ fst <$> objectIds'
+                    objectIdsMap = Map.fromList objectIds'
+                unless (StrictSeq.length objectIdsSeq <= fromIntegral objectIdsToReq) $
+                  throwIO ProtocolErrorObjectIdsNotRequested
+                handleReceivedObjectIds objectIdsToReq objectIdsSeq objectIdsMap
+                serverIdle
+            )
+    serverReqObjectIds
+      n@Zero
+      ObjectDecision
+        { objectIdsToAcknowledge = objectIdsToAck
+        , objectPipelineObjectIds = True
+        , objectIdsToRequest = objectIdsToReq
+        } =
+        pure $
+          SendMsgRequestObjectIdsPipelined
+            objectIdsToAck
+            objectIdsToReq
+            (handleReplies (Succ n))
+    serverReqObjectIds
+      n@Succ{}
+      ObjectDecision
+        { objectIdsToAcknowledge = objectIdsToAck
+        , objectPipelineObjectIds
+        , objectIdsToRequest = objectIdsToReq
+        } =
+        -- it is impossible that we have had `object`'s to request (Succ{} - is an
+        -- evidence for that), but no unacknowledged `objectId`s.
+        assert objectPipelineObjectIds $
+          pure $
+            SendMsgRequestObjectIdsPipelined
+              objectIdsToAck
+              objectIdsToReq
+              (handleReplies (Succ n))
+
+    handleReplies ::
+      forall (n :: N).
+      Nat (S n) ->
+      m (ServerStIdle (S n) objectId object m ())
+    handleReplies (Succ n'@Succ{}) =
+      pure $
+        CollectPipelined
+          Nothing
+          (handleReply (handleReplies n'))
+    handleReplies (Succ Zero) =
+      pure $
+        CollectPipelined
+          Nothing
+          (handleReply serverIdle)
+
+    handleReply ::
+      forall (n :: N).
+      m (ServerStIdle n objectId object m ()) ->
+      -- continuation
+      Collect objectId object ->
+      m (ServerStIdle n objectId object m ())
+    handleReply k = \case
+      CollectObjectIds objectIdsToReq objectIds -> do
+        let objectIdsSeq = StrictSeq.fromList $ fst <$> objectIds
+            objectIdsMap = Map.fromList objectIds
+        unless (StrictSeq.length objectIdsSeq <= fromIntegral objectIdsToReq) $
+          throwIO ProtocolErrorObjectIdsNotRequested
+        handleReceivedObjectIds objectIdsToReq objectIdsSeq objectIdsMap
+        k
+      CollectObjects objectIds objects -> do
+        let requested = Map.keysSet objectIds
+            received = Map.fromList [(objectId object, object) | object <- objects]
+
+        unless (Map.keysSet received `Set.isSubsetOf` requested) $
+          throwIO ProtocolErrorObjectNotRequested
+
+        mbe <- handleReceivedObjects objectIds received
+        traceWith tracer $ TraceObjectDiffusionCollected (objectId `map` objects)
+        case mbe of
+          -- one of `object`s had a wrong size
+          Just e ->
+            traceWith tracer (TraceObjectInboundError e)
+              >> throwIO e
+          Nothing -> k
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
@@ -1,0 +1,532 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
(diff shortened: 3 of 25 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.md
```diff
@@ -0,0 +1,187 @@
...
+
+- the inbound peer that submitted the object to pool might have acked it already at the moment the object is rejected by the pool, but the rejection indicates that the outbound peer which sent us the object is adversarial, and we should disconnect from it anyway. So there is no harm done by having acked the object to the adversarial outbound peer, as we won't want to re-download this object from it again (or any other object whatsoever).
+- any other inbound peer that has this ID available from its outbound peer won't be able to ack it because this ID isn't in **their** `dpsObjectsOwtPool`, and is not in the pool either, so we will be able to download it from these other peers until we find a valid one.
+
+As in TxSubmission, acknowledgement is done by indicating to the outbound peer the length of the (longest) prefix of the oustanding FIFO that we no longer care about (i.e. for which all IDs are eligible to acknowledgment by the definition above). The field `dpsOutstandingFifo` on the inbound peer is supposed to mirror exactly the state of the FIFO of the outbound peer, bar eventual discrepancies due to in-flight information.
```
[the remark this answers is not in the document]
tbagrel1: The outbound peer is not allowed to reply partially to one of our requests. It could (maybe) reply out of order to different requests we made, but if request A asks for 30 specific objects, reply to request A should contain exactly the 30 objects, otherwise we disconnect immediately from that peer for protocol fault.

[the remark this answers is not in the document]
agustinmista: That makes sense. I had missed the part that the requested objects would have to be sent all in the same reply and not in individual ones.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Download attribution process in `makeDecisions`
+
+When making decisions, we first divide the peers in two groups:
+
+- Those who are currently executing a decision, i.e., those for which the (previous) decision in the decision channel verifies `pdStatus == DecisionBeingActedUpon`. These are further called _frozen peers_.
+- Those who are not currently executing a decision, i.e., those for which the (previous) decision in the decision channel verifies `pdStatus == DecisionUnread || pdStatus == DecisionCompleted`. The former are the ones who didn't have time to read the previous decision yet, so it makes sense to recompute a more up-to-date decision for them. The latter are the ones who have completed executing the previous decision, so it also makes sense to compute a new decision for them. These two categories of peers are further called _active peers_.
```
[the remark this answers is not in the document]
agustinmista: Thanks. Yes, I think this answers my question.

Maybe a follow up question: let's say that all our inbound peers are blocked waiting for IDs. Would this make the decision thread enter a temporary busy waiting loop where it does nothing but to update its decisions over and over, or would it block as well until some condition flag switches state? 

Sorry in advance if this is nonsensical :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
```diff
@@ -0,0 +1,173 @@
...
+    ObjectDiffusionInboundPipelined $ goIdle Zero
+   where
+    goIdle :: forall (n :: N). Nat n -> InboundStIdle n objectId object m ()
+    goIdle n = WithEffect $ do
+      ctrlMsg <- atomically controlMessageSTM
+      traceWith tracer $ TraceObjectDiffusionInboundReceivedControlMessage ctrlMsg
+      case ctrlMsg of
+        -- The peer selection governor is asking us to terminate the connection.
+        Terminate ->
+          pure $ terminateAfterDrain n
```
agustinmista: There seem to be no tracing on this path to signal termination. Is this expected?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
```diff
@@ -0,0 +1,173 @@
...
+          psaOnDecisionCompleted
+          pure $ goIdle n
+        else do
+          psaOnRequestIds pdNumIdsToAck pdNumIdsToReq
+          psaOnDecisionCompleted
+          pure $
+            SendMsgRequestObjectIdsPipelined
+              pdNumIdsToAck
+              pdNumIdsToReq
+              (goIdle (Succ n))
```
agustinmista: Ditto `gotReqIdsBlocking`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
+{-# LANGUAGE BlockArguments #-}
```
agustinmista: I would avoid enabling this extension if it's only being used in L280 and L284.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,325 @@
...
+
+-- | Semaphore to guard access to the ObjectPool
+newtype ObjectPoolSem m = ObjectPoolSem (TSem m)
+
+newObjectPoolSem :: MonadSTM m => m (ObjectPoolSem m)
+newObjectPoolSem = ObjectPoolSem <$> atomically (newTSem 1)
+
+--
+-- DecisionPeerState, DecisionGlobalState
+--
```
agustinmista: You might want to turn this into a haddock section:

```-- * DecisionPeerState, DecisionGlobalState```

File: ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs [hunk 7]
```diff
@@ -0,0 +1,325 @@
...
+
+data DecisionContext peerAddr objectId object = DecisionContext
+  { dcRng :: StdGen
+  , dcHasObject :: (objectId -> Bool)
+  , dcDecisionPolicy :: DecisionPolicy
+  , dcGlobalState :: DecisionGlobalState peerAddr objectId object
+  , dcPrevDecisions :: Map peerAddr (PeerDecision objectId object)
+  }
+  deriving stock Generic
+  deriving anyclass NFData
```
agustinmista: Could you maybe add some documentation here?

Concretely, I would like to understand if we need both `DecisionGlobalState` and `DecisionContext`, as their names somewhat imply certain overlapping in their responsibilities.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,325 @@
...
+
+data PeerDecisionStatus
+  = DecisionUnread
+  | DecisionBeingActedUpon
+  | DecisionCompleted
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NFData, NoThunks)
+
+-- | A placeholder when no decision has been made, at the beginning of a loop.
+-- Nothing should be read from it except its status.
```
agustinmista: This feels a bit strange. Would it make sense to inject a `Maybe` somewhere to account for this?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+  ) =>
+  DecisionContext peerAddr objectId object ->
+  -- | New decisions
+  Map peerAddr (PeerDecision objectId object)
+makeDecisions DecisionContext{dcRng, dcHasObject, dcDecisionPolicy, dcGlobalState, dcPrevDecisions} =
+  let
+    -- A subset of peers are currently executing a decision. We shouldn't update the decision for them
+    frozenPeersToDecisions = Map.filter (\PeerDecision{pdStatus} -> pdStatus == DecisionBeingActedUpon) dcPrevDecisions
+
+    -- We do it in two steps, because computing the acknowledgment tell which objects from dpsObjectsAvailableIds sets of each peer won't actually be available anymore (as soon as we ack them),
```
agustinmista: NIT: wrap this line

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+  Map peerAddr (PeerDecision objectId object) ->
+  ( Map peerAddr (PeerDecision objectId object)
+  , Map peerAddr (Set objectId)
+  )
+computeAck poolHasObject DecisionPolicy{dpMaxNumObjectIdsReq, dpMaxNumObjectsOutstanding} DecisionGlobalState{dgsPeerStates} frozenPeersToDecisions =
+  let
+    -- We shouldn't create a new decision for peers that are currently executing a decision
+    filteredPeerStates = Map.withoutKeys dgsPeerStates (Map.keysSet frozenPeersToDecisions)
+    (decisions, peerToIdsToAck) =
+      Map.foldlWithKey' computeAckForPeer (Map.empty, Map.empty) filteredPeerStates
```
agustinmista: It seems like you're unconditionally inserting each input peer processed by this fold into the accumulator map, which makes me think that this could be defined more precisely via [Map.mapAccumWithKey](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html#v:mapAccumWithKey). 

I think such an implementation _should_ be more perfomant (single traversal (O(n)) vs. having to manually insert peers into the accumulator (O(n.log(n))) _and_ have the added benefit of preserving `length (fmap computeAckForPeer filteredPeerStates) == length filteredPeerStates` via parametricity.

Would this be possible?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+          + dpsNumIdsInflight
+
+      pdNumIdsToReq =
+        (fromIntegral dpMaxNumObjectsOutstanding - futureFifoSizeOnOutboundPeer)
+          `min` dpMaxNumObjectIdsReq
+
+      -- TODO: in the case where pdNumIdsToReq == 0, we know we actually won't be able to ack anything
+      -- during this round. So it might make sense to set pdNumIdsToAck = 0 and idsToAck = mempty as well?
+      -- If we do this change, we could add an assert in `V2.hs` that whenever pdNumIdsToReq == 0, then pdNumIdsToAck == 0 as well
+      -- /!\ We should also revise documentation in V2.md accordingly
```
agustinmista: Could you open and link an issue here?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+   in shuffle' peerPairs (length peerPairs) rng
+
+data DownloadPickState peerAddr objectId
+  = DownloadPickState
+  { totalNumObjectsToReq :: !NumObjectsReq
+  , objectMultiplicity :: ObjectMultiplicity
+  , peersToObjectsToReq :: Map peerAddr (Set objectId)
+  }
+
+-- | This function could just be pure if it hadn't be for the rng used to order peers
```
agustinmista: I think this comment no longer applies

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+  peerToIdsToAck =
+    peersToObjectsToReq
+   where
+    -- We order the peers that are not currently executing a decision
+    orderedPeers = orderPeers rng (dgsPeerStates `Map.withoutKeys` Map.keysSet frozenPeersToDecisions)
+
+    -- We want to map each objectId to the sorted list of peers that can provide it
+    -- For each peer we also indicate how many objects it has in flight at the moment
+    -- We filter out here the objects that are already in pool
+    objectsToSortedProviders :: Map objectId [(peerAddr, NumObjectsReq)]
```
agustinmista: Is the order of the elements important here? I think you could use a set instead of a list here (`peerAddr` already implements `Ord`), and turn the (O(n*2)) fold below (L217-L221) into O(n.log(n)).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+        DownloadPickState
+          { totalNumObjectsToReq = 0
+          , objectMultiplicity = 0
+          , peersToObjectsToReq = Map.empty
+          }
+        objectsToProvidersAndExpectedMultiplicities
+
+    -- This function decides whether or not we should select a given peer as provider for the current objectId
+    -- it takes into account if we are expecting to obtain the object from other sources (either inflight/owt pool already, or if the object will be requested from already selected peers in this given round)
+    howToFoldProviders ::
```
agustinmista: I would rename this function to something depicting its purpose, e.g. `pickProviderForObjectId`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,333 @@
...
+        peerToChannelAndDecision
+
+      -- Return values for tracing purposes
+      let counters = makeObjectDiffusionCounters globalState
+      return (newDecisions, counters)
+
+    traceWith decisionTracer (TraceDecisionLogicDecisionsMade newDecisions)
+    traceWith countersTracer counters
+
+-- `5ms` delay
```
agustinmista: Could this parameter be integrated into `DecisionPolicy`?

Also, 5ms is potentially up to 200 times a second. Isn't this _way_ too often? I would think that maybe somewhere in the realm of 10~20 times a second would be enough, but just guessing.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -354,6 +359,8 @@ library
     primitive,
     psqueues ^>=0.2.3,
     quiet ^>=0.2,
+    random,
+    random-shuffle,
```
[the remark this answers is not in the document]
tbagrel1: We shuffle the peers in `makeDecisions` to make sure we don't prioritize the same one (because of internal map ordering) when selecting providers.

This is because we don't a meaningful metric we can use to order pairs

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
tbagrel1: IIRC, I've been said (by Alex?) that we only need to improve tracing if the need shows up.  
But maybe are they some more precise guidelines for tracing somewhere?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
```diff
@@ -0,0 +1,173 @@
...
+    terminateAfterDrain = \case
+      Zero -> SendMsgDone ()
+      Succ n -> CollectPipelined Nothing $ \_ignoredMsg -> terminateAfterDrain n
+
+    goCollect :: Nat n -> PeerDecision objectId object -> InboundStIdle n objectId object m ()
+    goCollect Zero decision =
+      goReqObjects Zero decision
+    goCollect (Succ n) decision =
+      CollectPipelined
+        (Just $ goReqObjects (Succ n) decision)
```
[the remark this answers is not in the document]
tbagrel1: Yep the constructor has no real documentation, the only way to understand is to see how `CollectPipelined` is used by `run___` in `ouroboros-network`.

The first argument is what we should do if there is no reply available to be read immediately. If you use `Nothing`, it means you will always wait for a reply (and follow with the action provided as second argument). If you use `Just fallback`, in case there are no reply available, it will shortcut to `fallback` immediately and not use the second argument.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
```diff
@@ -0,0 +1,173 @@
...
+        ( \case
+            CollectObjectIds numIdsRequested ids -> WithEffect $ do
+              -- TODO: Add checks and validation
```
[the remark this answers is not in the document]
tbagrel1: For checks, indeed, as said above, this is a leftover comment as we now have all the checks implemented in functions of `State.hs` that this module calls through the `PeerStateAPI`.

For validation, we've made with @nbacquey that it can be done as early (upon reception) or as late (upon adding the object to the DB) as we want, with no impact on the correctness of the protocol implem.

I just don't understand your second sentence, as `V2.hs` is the _client side_ (and also inbound side, FTR).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
```diff
@@ -0,0 +1,173 @@
...
+          psaOnDecisionCompleted
+          pure $
+            SendMsgRequestObjectIdsBlocking
+              pdNumIdsToAck
+              pdNumIdsToReq
+              ( \objectIds -> WithEffect $ do
+                  -- TODO: Add checks and validation
+                  psaOnReceiveIds pdNumIdsToReq (NonEmpty.toList objectIds)
+                  pure $ goIdle Zero
+              )
```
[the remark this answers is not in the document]
tbagrel1: Here too, it will _always_ go idle in both branches (as it is the next step in order).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+module Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.V2.Decision
+  ( PeerDecision (..)
+  , makeDecisions
+
+    -- * Internal API exposed for testing
+  , DecisionContext (..)
```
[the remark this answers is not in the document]
tbagrel1: Actually, this is again a leftover comment that is no longer relevant. Now `DecisionContext` is an argument of `makeDecisions`, so it should definitely be public.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
tbagrel1: Sounds fair.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,325 @@
...
+  -- in which we submit them to the objectpool. It is also the order
+  -- in which we acknowledge them.
+  , dpsObjectsAvailableIds :: !(Set objectId)
+  -- ^ Set of known object ids which can be requested from this peer.
+  , dpsObjectsInflightIds :: !(Set objectId)
+  -- ^ The set of requested objects (by their ids).
+  -- , dpsObjectsRequestedButNotReceivedIds :: !(Set objectId)
+  -- ^ A subset of `dpsOutstandingFifo` which were unknown to the peer
+  -- (i.e. requested but not received). We need to track these `objectId`s
+  -- since they need to be acknowledged.
```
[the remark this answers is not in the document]
tbagrel1: It should be removed :) It's a vestige from the original TXSubmissionV2 implem

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs
tbagrel1: This comment has been written when all fields values except `pdStatus` were `error "..."`. Now it's plenty safe to read any field of these decisions.

And I don't think it would make sense to use a `Maybe`. Because `pdStatus` already guard the reading of such decision.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
tbagrel1: At first glance, I think we should be able to do that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+      -- /!\ We should also revise documentation in V2.md accordingly
+
+      pdCanPipelineIdsRequests = not . StrictSeq.null $ dpsOutstandingFifo'
+
+      peerDecision =
+        PeerDecision
+          { pdNumIdsToAck
+          , pdNumIdsToReq
+          , pdCanPipelineIdsRequests
+          , pdObjectsToReqIds = Set.empty -- we don't decide this here
```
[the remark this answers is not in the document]
tbagrel1: Probably the best solution type-wise would be to have a sub-datatype for all the fields related to acknowledgment. That way `computeAck` would return `AckDecision` and we would join `pdAckDecision :: AckDecision` and `pdObjectToReqIds` to form a `PeerDecision`.

Still, I'm not entirely convinced that we need to go that far, as we are never exposing a half-made decision through the `PeerStateAPI`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+  StdGen ->
+  (objectId -> Bool) ->
+  DecisionPolicy ->
+  DecisionGlobalState peerAddr objectId object ->
+  -- | Frozen peers and their previous decisions
+  Map peerAddr (PeerDecision objectId object) ->
+  -- | map from peer to the set of ids that will be acked for that peer on next requestIds
+  -- we should treat these ids as not available anymore for the purpose of picking objects to request
+  Map peerAddr (Set objectId) ->
+  -- | new global state (with just RNG updated), and objects to request from each peer
```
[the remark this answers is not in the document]
tbagrel1: Again, old comment that no longer applies :) sorry for this

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
tbagrel1: Yes, we need to preserve order, that's why I'm using a list :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+    howToFoldProviders ::
+      objectId ->
+      ObjectMultiplicity ->
+      DownloadPickState peerAddr objectId ->
+      (peerAddr, NumObjectsReq) ->
+      DownloadPickState peerAddr objectId
+    howToFoldProviders objectId expectedMultiplicity st@DownloadPickState{totalNumObjectsToReq, objectMultiplicity, peersToObjectsToReq} (peerAddr, numObjectsInFlight) =
+      let
+        -- see what has already been attributed to this peer
+        objectsToReq = Map.findWithDefault Set.empty peerAddr peersToObjectsToReq
```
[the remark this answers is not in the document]
tbagrel1: One important point of the `makeDecision` implem is that we want it to be as clear as possible, and only optimize for performance if we need to.

The stuff we copied from TxSubmissionV2 was probably well-optimized, but very hard to read, so here we tried to go for the more explicit (albeit less optimized) approach to avoid reproducing that.

Do you think that this particular case would look as clear/readable with `Map.updateLookupWithKey`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,333 @@
...
+  -- * The IDs of the received objects are a subset of `dpsObjectsInflightIds`
+  -- POSTCONDITIONS:
+  -- * The IDs of the received objects are removed from `dpsObjectsInflightIds`
+  -- * `dpsObjectsOwtPool` contains the received objects
+  }
+
+-- | A bracket function which registers / de-registers a new peer in
+-- `DecisionGlobalStateVar` and `PeerDecisionChannelsVar`s, which exposes `PeerStateAPI`.
+-- `PeerStateAPI` is only safe inside the `withPeer` scope.
+withPeer ::
```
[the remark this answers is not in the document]
tbagrel1: Agree!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/State.hs
```diff
@@ -0,0 +1,452 @@
...
+    peerState <- atomically $ ((Map.! peerAddr) . dgsPeerStates) <$> readTVar globalStateVar
+    hasObject <- atomically opwHasObject
+    checkProtocolErrors hasObject peerState numIdsInitiallyRequested receivedIds
+    globalState' <- atomically $ do
+      stateTVar
+        globalStateVar
+        ( \globalState ->
+            let globalState' = onReceiveIdsImpl peerAddr numIdsInitiallyRequested receivedIds globalState
+             in (globalState', globalState')
+        )
```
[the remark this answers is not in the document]
tbagrel1: @nbacquey I would like your input here :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/State.hs
```diff
@@ -0,0 +1,452 @@
...
+          decisionTracer
+          ( TraceDecisionLogicGlobalStateUpdated
+              "submitObjectsToPool.updateStateWhenObjectAddedToPool"
+              globalState'
+          )
+   where
+    updateStateWhenObjectAddedToPool ::
+      objectId ->
+      DecisionGlobalState peerAddr objectId object ->
+      DecisionGlobalState peerAddr objectId object
```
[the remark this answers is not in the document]
tbagrel1: Fine with that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2.hs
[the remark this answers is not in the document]
agustinmista: After discussing whether it makes sense to include validation as part of the protocol itself (thus allowing nodes to disconnect from adversarial peers as soon as possible), we noted that the benefit of such a feature will heavily depend on the redundancy factor (the more redundancy, the smallest the probability of choosing all adversarial peers when downloading an object), so there's also a tradeoff to be considered here.

We should keep this in mind when tuning the protocol parameters in the future.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Types.hs
agustinmista: After some discussion, see https://github.com/IntersectMBO/ouroboros-consensus/pull/1698#discussion_r2518714097

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,345 @@
...
+  ( Ord peerAddr
+  , Ord objectId
+  ) =>
+  DecisionContext peerAddr objectId object ->
+  -- | New decisions
+  Map peerAddr (PeerDecision objectId object)
+makeDecisions DecisionContext{dcRng, dcHasObject, dcDecisionPolicy, dcGlobalState, dcPrevDecisions} =
+  let
+    -- A subset of peers are currently executing a decision. We shouldn't update the decision for them
+    frozenPeersToDecisions = Map.filter (\PeerDecision{pdStatus} -> pdStatus == DecisionBeingActedUpon) dcPrevDecisions
```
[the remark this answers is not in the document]
agustinmista: Let's keep it unchanged :smile:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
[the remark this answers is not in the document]
agustinmista: I initially thought that the key being present in the map was an invariant of this function, but it seems that we are selecting providers while simultaneously populating the map.

Maybe a different way to look at it could be to create a map with all the final peers (not sure whether it's possible to know that in advance or not), and then map a function that selects the provider of an object. This should allow us to to the lookup/update in place.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/V2/Decision.hs
nbacquey: I think I made it even simpler than that, with `Map.map`. Let me know what you think

PR #1723, opened 2025-10-21
Add pure Peras voting rules

agustinmista:
Closes https://github.com/tweag/cardano-peras/issues/116

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
@@ -1,0 +1,398 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | Pure Peras voting rules
+--
+-- This module implements the Peras voting rules in a pure fasion, along with
+-- the necessary inpure machinery to retrieve their inputs. These rules are
+-- translated as verbatim as possible from:
+--
+-- https://github.com/cardano-foundation/CIPs/blob/master/CIP-0140/README.md#rules-for-voting-in-a-round
+module Ouroboros.Consensus.Peras.Voting
+  ( PerasVotingView (..)
+  , mkPerasVotingView
+  , isPerasVotingAllowed
+  , PerasVotingRule (..)
+  , VoteReason (..)
+  , NoVoteReason (..)
+  , perasVR1A
+  , perasVR1B
+  , perasVR2A
+  , perasVR2B
+  , perasVR1
+  , perasVR2
+  , perasVotingRules
+  )
+where
+
+import Cardano.Slotting.Slot (WithOrigin)
+import Data.Bifunctor (bimap)
+import Data.Typeable (Typeable)
+import Ouroboros.Consensus.Block (WithOrigin (..))
+import Ouroboros.Consensus.Block.Abstract
+  ( GetHeader (..)
+  , Header
+  , SlotNo (..)
+  , StandardHash
+  , castPoint
+  , succWithOrigin
+  )
+import Ouroboros.Consensus.Block.SupportsPeras
+  ( HasPerasCertRound (..)
+  , PerasRoundNo (..)
+  , ValidatedPerasCert
+  , getPerasCertBoostedBlock
+  , getPerasCertRound
+  , onPerasRoundNo
+  )
+import Ouroboros.Consensus.BlockchainTime.WallClock.Types
+  ( WithArrivalTime (..)
+  )
+import Ouroboros.Consensus.HardFork.History.EraParams (fromPerasEnabled)
+import qualified Ouroboros.Consensus.HardFork.History.Qry as HF
+import qualified Ouroboros.Consensus.HardFork.History.Summary as HF
+import Ouroboros.Consensus.Peras.Params
+  ( PerasBlockMinSlots (..)
+  , PerasCertArrivalThreshold (..)
+  , PerasCooldownRounds (..)
+  , PerasIgnoranceRounds (..)
+  , PerasParams (..)
+  )
+import Ouroboros.Consensus.Util.Pred
+  ( Evidence
+  , Explainable (..)
+  , ExplanationMode (..)
+  , Pred (..)
+  , evalPred
+  )
+import Ouroboros.Network.AnchoredFragment (AnchoredFragment)
+import qualified Ouroboros.Network.AnchoredFragment as AF
+
+{-------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
```diff
@@ -0,0 +1,196 @@
...
+  Set (AnchoredFragment blk) ->
+  WithArrivalTime (ValidatedPerasCert blk) ->
+  WithArrivalTime (ValidatedPerasCert blk) ->
+  (SlotNo -> PerasRoundNo) ->
+  (SlotNo -> SlotNo) ->
+  (SlotNo -> Point blk) ->
+  (Point blk -> WithArrivalTime (ValidatedPerasCert blk) -> Set (AnchoredFragment blk) -> Bool) ->
+  (WithArrivalTime (ValidatedPerasCert blk) -> SlotNo) ->
+  (PerasVotingView blk -> STM m a) ->
+  STM m a
```
agustinmista: This is a sketch for now. The idea is to derive most of these inputs from the context (e.g. a ChainDB instance)

agustinmista (commented):
Some implementation details are yet to be defined, especially around `PerasVotingView`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
```diff
@@ -0,0 +1,196 @@
...
+    , perasVotingRule2B r t
+    ]
+
+-- | If either VR-1A and VR-1B hold, or VR-2A and VR-2B hold, then voting is allowed.
+isPerasVotingAllowed :: StandardHash blk => SlotNo -> PerasVotingView blk -> Bool
+isPerasVotingAllowed s t =
+  or
+    [ perasVotingRule1 s t,
+      perasVotingRule2 (pvvRound t s) t
+    ]
```
amesgen: I think we want a somewhat richer return type, eg `Either NoVoteReason VoteReason`. This can then be used for tracing to signal why we did (not) vote, eg

 - We did not vote because the previous round didn't give rise to a quorum.
 - We did not vote because the cert from the previous round arrived too late.
 - We voted for the first time after a cooldown period.

(More abstract perspective: We essentially want to explain why a predicate succeeded/failed, so there is some overlap with general certificates for the truthiness of logical terms, as well as libraries like [`explainable-predicates`](https://hackage.haskell.org/package/explainable-predicates), but we should probably use sth ad-hoc; maybe there is room for a mini-DSL here)

Later, we could set up conformance testing against the Agda implementation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -43,6 +44,7 @@ import Quiet (Quiet (..))
 
 newtype PerasRoundNo = PerasRoundNo {unPerasRoundNo :: Word64}
   deriving Show via Quiet PerasRoundNo
+  deriving Semigroup via Sum Word64
```
amesgen: Hmm, for comparison, [`SlotNo`](https://base.cardano.intersectmbo.org/cardano-slotting/Cardano-Slotting-Slot.html#t:SlotNo) doesn't provide a particular `Semigroup`, but rather gives `Num`. `Num` of course isn't ideal, but unless you are opposed, I think we could follow suit.

(Also, *if* we add `Semigroup`, we should also add `Monoid` I think.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
```diff
@@ -0,0 +1,299 @@
...
+-- | Interface needed to evaluate the Peras voting rules
+data PerasVotingView cert = PerasVotingView
+  { pvvPerasParams :: !PerasParams
+  -- ^ Peras protocol parameters.
+  , pvvCurrRoundNo :: !PerasRoundNo
+  -- ^ The current Peras round number.
+  , pvvLatestCertSeen :: !cert
+  -- ^ The most recent certificate seen by the voter.
+  , pvvLatestCertOnChain :: !cert
+  -- ^ The most recent certificate present in some chain.
```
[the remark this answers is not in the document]
agustinmista: Good point.

I have pushed [some changes](https://github.com/IntersectMBO/ouroboros-consensus/pull/1723/commits/9869f7a29b38c18df1658d446d4e0c7c0ba4e0c3) to address this. If you agree with them, I will squash them down into the rest of the PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -164,22 +169,57 @@ makePerasCfg _ =
...
+  Cert0 -> a
+  Cert cert -> f cert
+
+-- | Eliminate the origin certificate case for a round number
+--
+-- NOTE: this assumes that the origin certificate corresponds to round number 0
+withOriginCertRoundNo ::
+  (HasPerasCertRound cert, Num (RoundNo cert)) =>
+  WithOriginCert cert -> RoundNo cert
+withOriginCertRoundNo = withOriginCert 0 getPerasCertRound
```
amesgen: Related to the previous comment: I think it would be a bit more in line to treat the round number of `Cert0` specially, ie assume that it is `Origin` which is the slot before slot `0`.

This is similar to how slot numbers work in the current code base: The `pointSlot` of a `Point blk` has type `WithOrigin SlotNo`, which is `Origin` when the point is the `GenesisPoint`. This makes arithmetic involving slots a bit annoying sometimes, but we have some utility functions like `succWithOrigin :: WithOrigin SlotNo -> SlotNo` that make it bearable.

For example, checking VR-1A would look very roughly like this:
```haskell
  VR1A
    := (pvvCurrRoundNo pvv :==: succWithOrigin latestCertSeenRoundNo)
    :/\: case pvvLatestCertSeen pvv of
           Cert0 -> Bool True
           Cert c -> latestCertSeenArrivalSlot c :<=: latestCertSeenRoundStart c + _X
```

Does this make sense, given that you might have already tried sth like this?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting.hs
```diff
@@ -0,0 +1,351 @@
...
+  vr2a =
+    -- NOTE we use 'succWithOrigin' and '-1' to handle the 'Origin' case (i.e.
+    -- when we have never seen a certificate before) correctly, essentially
+    -- treating the 'Origin' certificate as being from round -1.
+    succWithOrigin (getPerasCertRound latestCertSeen) - 1 + _R :<=: currRoundNo
+
+  latestCertSeen = pvvLatestCertSeen pvv
+  currRoundNo = pvvCurrRoundNo pvv
+
+  _R = coerce (perasIgnoranceRounds (pvvPerasParams pvv))
```
agustinmista: Action point: avoid using coerce here

amesgen (approved):
Looks great, only one comment about an underflow

PR #1727, opened 2025-10-22
Make DijkstraGenesis optional in unstable-cardano-tools

jasagredo:
Forwards-port of #1725


jasagredo:
I set the no-changelog label because this only relates to unstable tools.

PR #1730, opened 2025-10-23
[DO NOT MERGE] frisby/issue 1701 demo sandbox exe

nfrisby:
THIS PR WILL NEVER BE MERGED

I'm only opening it in order to facilitate conversation with @sterraf, who's trying out my WIP demo branch.


myManifest.json
```diff
@@ -0,0 +1,55 @@
+[
```
nfrisby: @sterraf Here's an example `myManifest.json` file.

sterraf:
@nfrisby  After building and testing with `./script`, I seem to see that EBs are CBOR-encoded as indefinite maps. (I see that this saves at least  1 byte per reference.) 

Are you planning to modify the CDDL accordingly?

sterraf:
> FYI @sterraf I just pushed up this file https://github.com/IntersectMBO/ouroboros-consensus/blob/nfrisby/issue-1701-demo-sandbox-exe/README-Leios-October-demo.md , which might be useful to read depending on what you're trying to do with this branch.

I got stuck with an already very stale commit, so I just took some data generated by the demo at that point and used it in my own tests. Not sure if I'll be able to keep doing experiments after tomorrow's call 😕

In any case, thanks a lot!

nfrisby:
Closed as stale. Superseded by `leios-prototype` branch.

PR #1739, opened 2025-10-29
Store previous epoch nonce in header state

agustinmista:
Closes https://github.com/tweag/cardano-peras/issues/150


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs
```diff
@@ -333,6 +335,7 @@ instance Serialise PraosState where
         <*> fromCBOR
         <*> fromCBOR
         <*> fromCBOR
+        <*> pure SNothing
         <*> fromCBOR
         <*> fromCBOR
```
agustinmista: I assumed that storing the previous epoch nonce is something we only want to do in memory and not on disk. Is this correct?

agustinmista:
> Apart from changing the target branch from `peras-staging-old` to `peras-staging`, is there anything more required to do here?

I think we are good to go. I will merge this after it passes the checks :smile:

PR #1768, opened 2025-11-20
[Peras 14] Add VoteDB and Vote aggregation structure for Peras

tbagrel1:
This PR introduces the API and implementation of an in-memory database to store Peras votes received from other peers via ObjectDiffusion. 

NOTES;

* To keep things compartmentalized, `O.C.Peras.Vote.Aggregation` implements the tricky parts of the vote aggregation logic within a given round, exposing only the interesting event of having _just_ reached quorum via the `VoteGeneratedCert` and `VoteDidntGenerateCert` pattern synonyms.

* Having multiple winners in the same round should, in principle, be impossible under the right parameterization (i.e. quorum threshold >= 75%). Since this will be ultimately become a tunable parameter, we resort to throw a `MultipleWinnersInRound` exception instead of a bare call to `error`. In addition, selecting the committee via Local Sortition in the future will add an extra level of statistical uncertainty that needs to be accounted for. This is because the _actual_ size of voting committee might differ from the _expected_ one, so one needs to end up asking for a stronger quorum to be reached before declaring a round winner. This is now kept as a separate `PerasQuorumStakeSafetyMargin` parameter.

* This PR is already rather large, so state-machine testing is implemented separately in [Peras 15](https://github.com/IntersectMBO/ouroboros-consensus/pull/1809).

### DIFF
--- a/ouroboros-consensus/changelog.d/20251224_142406_agustin.mista_vote_db.md
+++ b/ouroboros-consensus/changelog.d/20251224_142406_agustin.mista_vote_db.md
@@ -1,0 +1,25 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Add Peras vote aggregation logic.
+- Add PerasVoteDB API and in-memory implementation.
+- Add ObjectDiffusion instances for Peras votes using the PerasVotedDB as backend.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -108,13 +192,92 @@ class
...
+    { pvaTarget = (roundNo, point)
+    , pvaVotes = existingVotes
+    , pvaTotalStake = initialStake
+    , pvaMaybeCert = mExistingCert
+    }
+  vote =
+    if getPerasVoteRound vote == roundNo && getPerasVoteVotedBlock vote == point
+      then
+        let pvaTotalStake = initialStake + vpvVoteStake (forgetArrivalTime vote)
+            pvaVotes = Set.insert vote existingVotes
```
agustinmista: This insert not actually inserting anything (because the vote already exists) needs to be taken into consideration when updating the stake.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -108,13 +198,112 @@ class
...
     PerasCert blk ->
     Either (PerasValidationErr blk) (ValidatedPerasCert blk)
 
+  validatePerasVote ::
+    PerasCfg blk ->
+    PerasVote blk ->
+    PerasVoteStakeDistr ->
+    Either (PerasValidationErr blk) (ValidatedPerasVote blk)
+
+const_PERAS_QUORUM_THRESHOLD :: PerasVoteStake
```
agustinmista: Even if it's not a tunable parameter, I think it would be good to move this to `Ouroboros.Consensus.Peras.Params`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -108,13 +198,112 @@ class
...
+  }
+  deriving stock (Generic, Eq, Ord, Show)
+  deriving anyclass NoThunks
+
+data PerasVoteAggregateStatus blk
+  = PerasVoteAggregateQuorumNotReached {pvasVoteAggregate :: !(PerasVoteAggregate blk)}
+  | PerasVoteAggregateQuorumReachedAlready
+      {pvasVoteAggregate :: !(PerasVoteAggregate blk), pvasCert :: ValidatedPerasCert blk}
+  deriving stock (Generic, Eq, Ord, Show)
+  deriving anyclass NoThunks
```
agustinmista: I like this way better!

What I am still a bit concerned about, though, is that complete and incomplete voting aggregates are still represented by the same type, and that could make it easier to introduce bugs in the future.

Maybe we can define a GADT indexed by the voting state to make these two cases different at the type level, while still allowing for a single type to be carried around in most places:

```haskell
type data PerasVotingStatus = NoQuorum | Quorum

data PerasVotingTally blk status where
  NoQuorum ::
    !(PerasVoteAggregate) -> 
    PerasVotingTally NoQuorum
  Quorum :: 
    !(PerasVoteAggregate blk) ->
    !(ValidatedPerasCert blk) ->
    PerasVotingTally Quorum
```

This way, we could enforce the voting code to only accept the expected state transitions:

* NoQuorum -> NoQuorum
* NoQuorum -> Quorum
* Quorum -> Quorum
(note that Quorum -> NoQuorum shouldn't be possible)

As well as allowing future client code using these aggregates to specify when it expects voting to have reached a quorum, as opposed to handling it dynamically via a `Maybe` and/or some `error "impossible case"`. E.g., instead of `pvasMaybeCert`, we could have:

```haskell
perasVotingCert :: PerasVotingTally blk Quorum -> ValidatedPerasCert blk
perasVotingCert (Quorum _ cert) = cert
```

Which can only be used to extract the cert from the aggregate bundle after voting has reached a quorum.

---

Going one step further, we could also encode the number of votes that we have seen after reaching a quorum in the type-level `Quorum` constructor:

```haskell
type data PerasVotingStatus = NoQuorum | Quorum Nat
```

This would allow us to enrich the state transitions to:

* NoQuorum -> NoQuorum
* NoQuorum -> Quorum Z
* Quorum n -> Quorum (S n)

And we could use `Quorum 0` as the witness for the state transition that produces a certificate. This way, _I think_ we could able to get rid of the `PerasVoteResult` type* or, alternatively, turn it into a thin wrapper over `PerasVotingTally` to hide the existential:

```haskell
data PerasVotingTally blk status where
  NoQuorum ::
    !(PerasVoteAggregate) -> 
    PerasVotingTally NoQuorum
  Quorum ::
    !(SNat n) ->
    !(PerasVoteAggregate blk) ->
    !(ValidatedPerasCert blk) ->
    PerasVotingTally (Quorum n)

-- constructor not exported
newtype PerasVoteResult blk = PerasVoteResult (Some (PerasVoteTally blk))

-- instead we could expose only the things the client need, e.g.:
-- to extract the certificate if we have _just_ reached a quorum, and never again
perasVoteResultEmittedCert :: PerasVoteResult blk -> Maybe (ValidatedPerasCert blk)
perasVoteResultEmittedCert (PerasVoteResult (Some tally)) =
  case tally of
    Quorum Z _ cert -> Just cert
    _ -> Nothing
```

[*] depending on whether the future client code might also be interested in reacting to the event of adding a certificate that was already in the VoteDB (I think it wouldn't, as that's sounds as something purely relevant to tracing, which is already being handled by the VoteDB implementation)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,716 @@
...
+updatePerasRoundVoteState ::
+  forall blk.
+  StandardHash blk =>
+  WithArrivalTime (ValidatedPerasVote blk) ->
+  PerasCfg blk ->
+  PerasRoundVoteState blk ->
+  Either (UpdateRoundVoteStateError blk) (PerasRoundVoteState blk)
+updatePerasRoundVoteState vote cfg roundState =
+  if getPerasVoteRound vote /= getPerasVoteRound roundState
+    then error "updatePerasRoundVoteTallys: vote round does not match aggregate round"
```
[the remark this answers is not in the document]
tbagrel1: I'm not really sure. The fact that _the `roundNo` in a `RoundState`/`target` in a `TargetState` is consistent with the key it is given in the map_ is an invariant that is not enforced statistically, so I think it makes sense to still have an error case for it (i.e. it works as an assert if we mess up somewhere in the implementation).

Moving such a function to a local definition would make it harder to test, and I think that would be a bigger issue maybe?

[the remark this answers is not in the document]
agustinmista: That's fair. I just wanted to understand if this error call came from splitting a function with internal invariants into two smaller ones, or if it was something external.

Not sure if we will ever want to test this function in isolation. More likely through the election API via `quickcheck-dynamic`, no?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -83,6 +139,39 @@ data ValidatedPerasCert blk = ValidatedPerasCert
   deriving stock (Show, Eq, Ord, Generic)
   deriving anyclass NoThunks
 
+instance
+  ( HasId (PerasCert blk)
+  , Ord (IdOf (PerasCert blk))
+  , Eq (IdOf (PerasCert blk))
+  , Show (IdOf (PerasCert blk))
+  , NoThunks (IdOf (PerasCert blk))
+  , Serialise (IdOf (PerasCert blk))
```
nbacquey: Out of curiosity, do you need all those instances when you already know `HasId (PerasCert blk)` and `type IdOf (ValidatedPerasCert blk) = IdOf (PerasCert blk)`?
I'm not saying you should change anything, but it seems redundant to me. Maybe you could say something like `IdOf (PerasCert blk) ~ IdOf (ValidatedPerasCert blk)` in the constraints?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -83,6 +139,39 @@ data ValidatedPerasCert blk = ValidatedPerasCert
...
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving anyclass NoThunks
+
+instance
+  ( HasId (PerasVote blk)
+  , Ord (IdOf (PerasVote blk))
+  , Eq (IdOf (PerasVote blk))
+  , Show (IdOf (PerasVote blk))
+  , NoThunks (IdOf (PerasVote blk))
+  , Serialise (IdOf (PerasVote blk))
```
nbacquey: Same remark as above: could it be simplified with something like `IdOf (PerasVote blk) ~ IdOf (ValidatedPerasVote blk)`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -218,3 +394,51 @@ instance
   HasPerasCertBoost (WithArrivalTime cert)
   where
   getPerasCertBoost = getPerasCertBoost . forgetArrivalTime
+
+class HasPerasVoteRound vote where
```
nbacquey: Do you think it would be a worthwhile idea to get rid of the triplicate boilerplate with a single `HasPerasVote` class? You would then get the 3 `getPerasVoteStuff` with `pvVoteStuff . getPerasVote`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
tbagrel1: Hum HLS was complaining when I only had `HasId (PerasCert blk) =>` and suggested that I introduced the apparently-redundant constraints.

I'll try with `(HasId (PerasCert blk), IdOf (PerasCert blk) ~ IdOf (ValidatedPerasCert blk)) =>`! This may be the part that I was missing.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -133,12 +241,25 @@ instance StandardHash blk => BlockSupportsPeras blk where
...
+    deriving stock (Generic, Eq, Ord, Show)
+    deriving anyclass NoThunks
+
   -- TODO: enrich with actual error types
   -- see https://github.com/tweag/cardano-peras/issues/120
   data PerasValidationErr blk
     = PerasValidationErr
     deriving stock (Show, Eq)
 
+  data PerasForgeErr blk
```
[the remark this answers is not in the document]
tbagrel1: `PerasForgeErrMismatchedTarget` is when `forgePerasCert` is given votes for a different target than the one we are trying to forge a certificate for.

But yeah I agree, we would need to document these.

> Also, even if we eventually ensure by construction that quorum > 0.75, shouldn't we need some PerasForgeErrMultipleCerts?

No, because `forgePerasCert` is a function from `BlockSupportsPeras` that has no access to the state or voteDB, it is only there to handle the crypto part that depends on the concrete block type. However, there is a dedicated error variant in both `UpdateRoundVoteStateError` and `VoteDbError` for when two certificates are forged in a same round.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,716 @@
...
+-- May fail if the candidate is elected winner but forging the certificate fails.
+updateCandidateVoteState ::
+  StandardHash blk =>
+  PerasCfg blk ->
+  WithArrivalTime (ValidatedPerasVote blk) ->
+  PerasTargetVoteState blk 'Candidate ->
+  Either
+    (PerasForgeErr blk)
+    (Either (PerasTargetVoteState blk 'Candidate) (PerasTargetVoteState blk 'Winner))
+updateCandidateVoteState cfg vote oldState =
```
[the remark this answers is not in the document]
tbagrel1: I don't think this is an issue if in a single place in the code we have a nested `Either`. I think we already have many ad-hoc types in this module, and this new one wouldn't be very meaningful IMO. But I agree with @nbacquey suggestion, we could use `do` for error-handling and `Right/Left` explicit use for possible returned variants

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,716 @@
...
+updateLoserVoteState ::
+  StandardHash blk =>
+  PerasCfg blk ->
+  WithArrivalTime (ValidatedPerasVote blk) ->
+  PerasTargetVoteState blk 'Loser ->
+  Either (PerasTargetVoteState blk 'Loser) (PerasTargetVoteState blk 'Loser)
+updateLoserVoteState cfg vote oldState =
+  let newVoteTally = updateTargetVoteTally vote (ptvsVoteTally oldState)
+      aboveQuorum = voteTallyAboveQuorum cfg newVoteTally
+   in if aboveQuorum
```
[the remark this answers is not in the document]
tbagrel1: In other parts of the code we raise an `error` as it is not supposed to happen (i.e. the `error` is here to enforce an internal invariant of the implementation, something that shouldn't ever fail except if we made a mistake in the impl).

Here however, we are in a place where a loser could theoretically go above quorum even if our implementation is valid and bug-free. That's why we return an `Either` where the `Left` case indicates an error that is not our responsibility. E.g., if the vote validation logic is faulty, a loser could go above quorum even with a perfectly valid implementation on our end (as voteDB implementers).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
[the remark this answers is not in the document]
agustinmista: Sounds good to me. 

If you're feeling fancy, you could also pull `PerasTargetVoteState` out of the inner Either and return and NS over the two possible return variants of the same indexed type:

```haskell
Either 
  (PerasForgeErr blk)
  (NS (PerasTargetVoteState blk) '[Candidate, Winner])
```

This makes it more evident that you're returning a new vote state constrained to only 2 of the 3 existing variants.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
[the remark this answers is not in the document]
agustinmista: > E.g., if the vote validation logic is faulty

If we assume that the vote validation logic _could_ be faulty. Wouldn't we need to turn `error`s into `Either`s in other places as well? Or is this the _only_ place where that problem could manifest? Looks a bit like a slippery slope of responsibility delegation to me :thinking:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,405 @@
...
+                  ForgingCertError err
+              Right
+                ( PerasRoundVoteStateQuorumReachedNowWithCert cert
+                  , pvsRoundVoteStates'
+                  ) ->
+                  pure
+                    ( AddedPerasVoteAndGeneratedNewCert cert
+                    , pvsRoundVoteStates'
+                    )
+              Right (_, pvsRoundVoteStates') ->
```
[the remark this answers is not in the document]
agustinmista: Previously, `AddedPerasVoteAndGeneratedNewCert` was the only exported constructor (a pattern synonym actually), so there was nothing else to pattern match against. 

I refactored this a bit so that we have two separate public patterns for a peras vote state that the client code can match against.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
```diff
@@ -0,0 +1,456 @@
...
+  | RoundVoteStateForgingCertError
+      (PerasForgeErr blk)
+
+-- | Add a vote to an existing round aggregate.
+--
+-- PRECONDITION: the vote's round must match the aggregate's round.
+--
+-- May fail if the state transition is invalid (e.g., a loser going above
+-- quorum) or if forging the certificate fails.
+updatePerasRoundVoteState ::
```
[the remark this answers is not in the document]
agustinmista: It's a bit subtle, but `updateTargetVoteTally` silently discards votes when there's an existing one with a matching (roundNo, voterId).

I refactored the code a bit to make it more evident, but indeed, I think we want this to be at least a traceable incident.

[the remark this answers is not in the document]
agustinmista: This is currently being taken care by the VoteDB internal state, where there is a separate `Set (VoteId blk)` being kept around to check for [this very case](https://github.com/IntersectMBO/ouroboros-consensus/pull/1768/changes#diff-c436c8fd1e2ea25396a5ca2d43d7788cfc835d6db7021aac6ac199ebf9980f5cR207).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,405 @@
...
+          let pvsVoteIds' = Set.insert voteId pvsVoteIds
+              pvsLastTicketNo' = succ pvsLastTicketNo
+              pvsVotesByTicket' = Map.insert pvsLastTicketNo' vote pvsVotesByTicket
+              fp' = succ fp
+
+          (res, pvsRoundVoteStates') <-
+            case updatePerasRoundVoteStates vote perasCfg pvsRoundVoteStates of
+              Left (RoundVoteStateLoserAboveQuorum winnerState loserState) ->
+                throwSTM $
+                  EquivocatingCertError
```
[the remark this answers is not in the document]
agustinmista: > I don't think EquivocatingCertError is the right error to throw here

I renamed this case to `MultipleWinnersInRound` to make it more evident what's going on.

> it should be an implementation error, rather than an exception.

I think an exception is more fitting at the moment, given that this being an implementation error is currently up to parameterization.

I'm leaving the discussion about `>= 1-quorum` vs `>= quorum` open for the rest of the open conversations.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -44,11 +65,41 @@ import Ouroboros.Consensus.Util
...
+instance HasId perasObj => HasId (WithArrivalTime perasObj) where
+  type IdOf (WithArrivalTime perasObj) = IdOf perasObj
+  getId = getId . forgetArrivalTime
+
 newtype PerasRoundNo = PerasRoundNo {unPerasRoundNo :: Word64}
   deriving Show via Quiet PerasRoundNo
   deriving stock Generic
   deriving newtype (Enum, Eq, Ord, Num, Bounded, NoThunks, Serialise)
 
+newtype PerasVoteStake = PerasVoteStake {unPerasVoteStake :: Rational}
```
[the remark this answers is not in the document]
agustinmista: I think the rest of the open comments all stem from here:

We never say that the quorum threshold, nor the individual vote stake must be between 0 and 1.

If we could cap the total stake of the system so that any value can be mapped into [0,1], then we could assert loser states against a stronger consistency condition (`< 1-quorum` instead of `< quorum`, assuming that the threshold `> 0.5`).

The question is whether we can expect to know the total stake of the system (or maybe just of the committee if we make some assumptions like it being a representative sample of the population) in advance.

tbagrel1:
I think we should add  thresholdSafetyMargin and undraft it :)

Le mer. 28 janv. 2026 à 10:22, Agustín Mista ***@***.***> a
écrit :

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
```diff
@@ -0,0 +1,456 @@
...
+-- May fail if the loser goes above quorum by adding the vote.
+updateLoserVoteState ::
+  StandardHash blk =>
+  PerasCfg blk ->
+  WithArrivalTime (ValidatedPerasVote blk) ->
+  PerasTargetVoteState blk 'Loser ->
+  Either (PerasTargetVoteState blk 'Loser) (PerasTargetVoteState blk 'Loser)
+updateLoserVoteState cfg vote oldState =
+  let newVoteTally = updateTargetVoteTally vote (ptvsVoteTally oldState)
+      aboveQuorum = voteTallyAboveQuorum cfg newVoteTally
```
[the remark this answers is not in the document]
agustinmista: This is also limited by the future use of LS for committee selection. 

I think this check is already a good effort to raise an exception against the very unlikely event of having multiple winners in a round.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
dnadales: Would it make sense to add `assert`s for the `PRECONDITION`s  in this module?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,404 @@
...
+  deriving stock (Show, Eq, Generic)
+
+{-------------------------------------------------------------------------------
+  Exceptions
+-------------------------------------------------------------------------------}
+
+data PerasVoteDbError blk
+  = -- | Attempted to use a closed database
+    ClosedDBError PrettyCallStack
+  | -- | Attempted to add a vote that would lead to multiple winners for the same round
```
dnadales: Could we clarify what each of the two points and stakes mean (maybe with field names or a newtype wrapper)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -0,0 +1,126 @@
...
+  -- see https://github.com/tweag/cardano-peras/issues/73
+  -- see https://github.com/tweag/cardano-peras/issues/120
+  case traverse (validatePerasVote perasParams distr) votes of
+    Left validationErr -> throw (PerasVoteValidationError validationErr)
+    Right validatedVotes -> return validatedVotes
+
+-- | Add a list of 'PerasVote's into an object pool.
+--
+-- NOTE: we first validate the votes, throwing an exception if any of
+-- them are invalid. We then wrap them with their arrival time, and finally add
```
dnadales: Don't we want to collect all of the failures, instead of an all of nothing approach?

dnadales (commented):
Look great. I left some comments to discuss.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -0,0 +1,126 @@
...
+  (StandardHash blk, MonadThrow m) =>
+  PerasVoteStakeDistr ->
+  [PerasVote blk] ->
+  m [ValidatedPerasVote blk]
+validatePerasVotes distr votes = do
+  let perasParams = mkPerasParams
+  -- TODO pass down 'BlockConfig' when all the plumbing is in place
+  -- see https://github.com/tweag/cardano-peras/issues/73
+  -- see https://github.com/tweag/cardano-peras/issues/120
+  case traverse (validatePerasVote perasParams distr) votes of
```
bladyjoker: How many of these votes do we expect to have? Consider what I suggested before, all such reasoning could be made super transparent with your Predicate DSL. Now that we have it, why not use it? In that sense, the `traverse` becomes `Predicate.all (#== Bool True) (isValidVote <$> votes)`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -0,0 +1,126 @@
...
+  -- But `StrictTVar m PerasVoteStakeDistr` might not be the best choice for that.
+  StrictTVar m PerasVoteStakeDistr ->
+  SystemTime m ->
+  PerasVoteDB m blk ->
+  ObjectPoolWriter (PerasVoteId blk) (PerasVote blk) m
+makePerasVotePoolWriterFromVoteDB distrVar systemTime perasVoteDB =
+  ObjectPoolWriter
+    { opwObjectId = getPerasVoteId
+    , opwAddObjects = \votes -> do
+        distr <- readTVarIO distrVar
```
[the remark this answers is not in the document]
bladyjoker: I like that we're providing a `distrVar` which is a TVar, now the obvious question, do `systemTime` and `distr` have to be consistent? If so why not provide a `systemTimeVar` and read both in an atomic block?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,64 @@
...
+  ( MonadSTM (STM)
+  )
+
+data PerasVoteDB m blk = PerasVoteDB
+  { addVote :: WithArrivalTime (ValidatedPerasVote blk) -> m (AddPerasVoteResult blk)
+  -- ^ Add a Peras vote to the database. The result indicates whether
+  -- the vote was actually added, or if it was already present.
+  , getVoteSnapshot :: STM m (PerasVoteSnapshot blk)
+  -- ^ Interface to read the known votes, mostly for diffusion
+  , getForgedCertForRound :: PerasRoundNo -> STM m (Maybe (ValidatedPerasCert blk))
```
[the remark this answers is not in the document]
bladyjoker: Great comment. I'd like to expand a bit on this. It's important to note that this is not just a matter of "in memory" vs. "stored on disk". It's about the very essential properties of this component and about **ACID** properties.

The `getSnapshot` is one way, sometimes quite inefficient, and often prone to errors, to deal with the DB in ACID form. 

In the current form, one can't really say the following

```haskell
afterAddVoteItIsInSnapshot voteDb = do
  addVote voteDb someVote
  snap <- getSnapshot voteDb
  assert $ someVote `isIn` snap
```

Because, the interface doesn't enforce ACID between `addVote` and `getSnapshot`. This is going to be difficult to test and guarantee correctness.

An example of that is Mempool too, check out [this test](https://github.com/IntersectMBO/ouroboros-consensus/blob/36cf8a7eb8953a831e1b356aa6dcd0a894411574/ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs#L120) which claims something that can only be claimed for the very mocked version being tested and not the **real** Mempool.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,62 @@
...
+  -- ^ Get the certificate if quorum was reached for the given round.
+  , garbageCollect :: PerasRoundNo -> m ()
+  -- ^ Garbage-collect state strictly older than the given slot number.
+  , closeDB :: m ()
```
bladyjoker: nit: Somehow I would rather this is 'outside' like `openDB`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,424 @@
...
+            }
+
+{-------------------------------------------------------------------------------
+  Implementation-internal types
+-------------------------------------------------------------------------------}
+
+-- | Volatile Peras certificate state, i.e. certificates that could influence
+-- chain selection by boosting a volatile block.
+--
+-- INVARIANT: See 'invariantForPerasVolatileVoteState'.
```
bladyjoker: This is awesome! However, I find this to be less relevant than the same thing just expressed at the API level. I want to see `propFoo` listed clearly next to the API and then the implementation simply needs to conforms and the testsuite is going to assert all these properties.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,424 @@
...
+data PerasVolatileVoteState blk = PerasVolatileVoteState
```
bladyjoker: nit; Why `Volatile`? This is simply `State` and we know that the state changes given some actions.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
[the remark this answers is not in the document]
nbacquey: All vote validations during a fixed voting round (several rounds, actually) will use the same stake distribution. I suggest we use the same stake distribution that is used to check election proofs, since we face basically the same challenges, and don't want to reinvent the wheel

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -4,59 +4,117 @@
...
+  (res, traceEvents) <- atomically stm
+  mapM_ (traceWith tracer) traceEvents
+  pure res
+
+{------------------------------------------------------------------------------}
+
+-- TODO: for consistency I made all functions of the API return `STMWithTracing`
+-- but for the getters it might make sense to keep them as simple `STM m a`.
+-- If that is the case, we can get rid of the two variants for
+-- `STMWithTracing` and just keep the `WithTracing` one
```
ch1bo: This adds 50+ lines of utility for essentially two cases:
- `addVote` produces always `AddingPerasVote`, and `IgnoredVoteAlreadyInDB`, `AddedPerasVote`, `GeneratedPerasCert` depending on `addPerasVoteRes` .. which is exactly the return value of that function
- `garbageCollect` produces `GarbageCollected` (what a surprise! 🙂)

So unless I'm missing the big decision logic still incoming, which has intricate decisions to be traced, .. it's not worth it. And for the occasional use, the return value can be the trace / carry as much info as you would have traced.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
fraser-iohk: could we add this reasoning to the function's haddocks, so the intended use remains clear if we ever use `addPerasVotes` in a different way in future?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,120 @@
...
+      PerasVoteTicketNo ->
+      STMWithTracing (TraceEvent blk) m (Map PerasVoteTicketNo (WithArrivalTime (ValidatedPerasVote blk)))
+  , getForgedCertForRound ::
+      PerasRoundNo ->
+      STMWithTracing (TraceEvent blk) m (Maybe (ValidatedPerasCert blk))
+  -- ^ Get the certificate if quorum was reached for the given round.
+  , garbageCollect :: PerasRoundNo -> STMWithTracing (TraceEvent blk) m ()
+  -- ^ Garbage-collect state strictly older than the given slot number.
+  }
+  deriving NoThunks via OnlyCheckWhnfNamed "PerasVoteDB" (PerasVoteDB m blk)
```
bladyjoker: I would remove the explicit tracing at the API level. The implementation should provide tracing unconditionally as we do elsewhere. Furthermore, adding tracing to the API means that now you have to document or layout props of what tracing event happens at which call.

Generally, I'd actually like to see tracing become part of the API that callers can rely on, as it enables event driver programming, however, let's keep it the way we do it for now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,348 @@
...
+  pvdeState <-
+    newTVarWithInvariantIO
+      (either Just (const Nothing) . invariantForPerasVoteDbState)
+      initialPerasVoteDbState
+  let env =
+        PerasVoteDbEnv
+          { pvdeTracer
+          , pvdeState
+          }
+  traceWith pvdeTracer OpenedPerasVoteDB
```
bladyjoker: I'd even remove this entire trace, there's no "close", semantically it's dubious what it denotes.

bladyjoker (changes requested):
Heyoo, I promise just a bit more on this and we can get it approved too!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,120 @@
...
+zeroPerasVoteTicketNo = PerasVoteTicketNo 0
+
+data AddPerasVoteResult blk
+  = PerasVoteAlreadyInDB
+  | AddedPerasVoteButDidntGenerateNewCert
+  | AddedPerasVoteAndGeneratedNewCert (ValidatedPerasCert blk)
+  deriving stock (Generic, Eq, Ord, Show)
+  deriving anyclass NoThunks
+
+data TraceEvent blk
```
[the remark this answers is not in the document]
agustinmista: I think this is based on the precedent of other `TraceEvent`s being defined for the `ChainDB`, `VolatileDB`, `ImmutableDB` and `PerasCertDB`, so I guess these all rely on qualified imports.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,157 @@
...
+  MonadSTM m =>
+  PerasVoteDB m blk ->
+  PerasVoteTicketNo ->
+  m Bool
+prop_getVotesAfterMonotonic db ticketNo =
+  atomically $ do
+    votes <- getVotesAfter db ticketNo
+    let tickets = Map.keys votes
+    pure $
+      tickets == List.sort tickets
```
bladyjoker: Hmm, wasn't aware `Map.keys` returns an ordered list! Great stuff

```
O(n). Return all keys of the map in ascending order. 
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
bladyjoker: Freaking love this! Thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,352 @@
...
+-------------------------------------------------------------------------------}
+
+data TraceEvent blk
+  = AddVote
+      (PerasVoteId blk)
+      (WithArrivalTime (ValidatedPerasVote blk))
+      (AddPerasVoteResult blk)
+  | GarbageCollected
+      PerasRoundNo
+  deriving stock (Show, Eq, Generic)
```
bladyjoker: Quite concise and complete, thank you!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/Impl.hs
```diff
@@ -0,0 +1,352 @@
...
+    } = args
+
+{-------------------------------------------------------------------------------
+  API implementation
+-------------------------------------------------------------------------------}
+
+-- TODO: we will need to update this method with non-trivial validation logic
+-- see https://github.com/tweag/cardano-peras/issues/120
+implAddVote ::
+  ( IOLike m
```
bladyjoker: Curios if we want to restrict `m` to IOLike?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -0,0 +1,157 @@
...
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Consensus.Storage.PerasVoteDB.API
+  ( PerasVoteDB (..)
+  , AddPerasVoteResult (..)
+  , PerasVoteTicketNo
+  , zeroPerasVoteTicketNo
+
+    -- * Invariants
```
bladyjoker: TODO: Include in a testsuite

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
```diff
@@ -0,0 +1,501 @@
...
+  = PerasRoundVoteStateQuorumNotReached
+      { prvsRoundNo :: !PerasRoundNo
+      , prvsCandidateStates :: !(Map (Point blk) (PerasTargetVoteState blk 'Candidate))
+      }
+  | PerasRoundVoteStateQuorumReachedAlready
+      { prvsRoundNo :: !PerasRoundNo
+      , prvsExcessVotes :: !Word64
+      , prvsLoserStates :: !(Map (Point blk) (PerasTargetVoteState blk 'Loser))
+      , prvsWinnerState :: !(PerasTargetVoteState blk 'Winner)
+      }
```
bladyjoker: Ermm, can we avoid records in sums to not have partial accessors? Aka, add define 2 new records for each variant.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -155,6 +185,59 @@ data ValidatedPerasVote blk = ValidatedPerasVote
...
+votesReachQuorum ::
+  StandardHash blk =>
+  PerasCfg blk ->
+  [ValidatedPerasVote blk] ->
+  Maybe (ValidatedPerasVotesReachingQuorum blk)
+votesReachQuorum cfg votes =
+  case votes of
+    -- We need at least one vote to determine who these votes are for, so we
+    -- can't vacuously reach a quorum, even if the quorum threshold is 0.
+    [] -> Nothing
```
bladyjoker: NonEmpty?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
agustinmista: Testing these invariants would require generating interesting `PerasVoteDB`s to be used as inputs. One easy way to do so is by following the API and generating sequences of commands on top of an empty DB.

While we can definitely do that in isolation, it also sounds an awful lot like approach already taken by the state-machine tests defined in [Peras 15](https://github.com/IntersectMBO/ouroboros-consensus/pull/1809), so I'd like to see if we can reuse some* of that machinery to instantiate these invariants in separate PR.

Or maybe your comment is just suggesting we add a comment with a TODO here? :sweat_smile: 

[*] Note that we can't easily do it directly as part of the state machine test postcondition because these invariants are metamorphic relations involving more than one API call at a time.

bladyjoker (commented):
After reading through the Aggregation module I realized the intent was a bit obscured due to many indirections. I provided a draft of an alternative formulation that is imo quite concise which I find quite important for such a key part of the logic.

Let me know if I'm missing something.

PR #1772, opened 2025-11-24
[Peras 6] Improve generation of Peras events in ChainDB state machine tests

### LINKED ISSUE #1745: Allow for k>2 in ChainDB state machine tests
There is a misalignment between the ChainDB model and SUT that pops up as a counterexample to the `ChainDB q-s-m` property when setting `k>2`. 

More concretely, in the presence of EBBs, block numbers can be very misleading, e.g., the tip block number of a shorter chain can have a higher block number than that of a longer one. Moreover, the SUT implements an [optimization](https://github.com/IntersectMBO/ouroboros-consensus/blob/c3b1ab6453b812f31ff2e5344a127467f8f5db55/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs#L754) where it doesn't consider candidates that are strictly shorter than the current chain. This "optimization" turns out to change semantics with frequent EBBs.

To avoid hitting this mismatch, the test suite currently limits `k` to be `<=2`, which prevents testing weighted chain selection more effectively (currently blocking #1682). The goal of this ticket is to find a way to allow for `k` to take larger values without constantly triggering the aforementioned mismatch between model and SUT.

---

The concrete details to reproduce (courtesy of @amesgen):

`ouroboros-consensus` commit: `417989948178ad80e0af282a4ea8878df53d90ad` (but more recent ones should also work if they didn't somehow touch the generation logic)

Change the security parameter to `5` here https://github.com/IntersectMBO/ouroboros-consensus/blob/c3b1ab6453b812f31ff2e5344a127467f8f5db55/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs#L1616

```
cabal run ouroboros-consensus:storage-test -- -p "ChainDB q-s-m" --quickcheck-replay="(SMGen 7556261789993583380 2507775810768318359,32)"
```

Succinct overview of the test, ie its sequence of commands:
```
Genesis :> A :> C :> B :> D :> F
                  :> E :> G

k = 5

AddBlock A  BlockNo 0  IsEBB     -4719434653394739275
AddBlock B  BlockNo 1  IsEBB     1432075052124753162
AddBlock C  BlockNo 1  IsNotEBB  3900252851851256110
AddBlock D  BlockNo 2  IsNotEBB  -8964932805968377556
AddBlock E  BlockNo 2  IsNotEBB  3485505698117617698
AddBlock F  BlockNo 2  IsEBB     -1585419396674147681
AddBlock G  BlockNo 3  IsNotEBB  -1773151708669127551
```

 - The model selects `G` in the end, which makes sense from the perspective that it has the highest block number.
 - The SUT stays at `F` and doesn't adopt `G`, as it has an [optimization](https://github.com/IntersectMBO/ouroboros-consensus/blob/c3b1ab6453b812f31ff2e5344a127467f8f5db55/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs#L754) meaning that the candidate with `G` is filtered out.

agustinmista:
This PR introduces a sequence of changes to progressively improve the chances of observing possible-but-currently-uncommon Peras events in the ChainDB state machine tests. Concretely, we are interested in observing the chain selection algorithm switching from a longer to a shorter chain. For this, we rely on the following changes:

1. Tweaking the `Cmd` generation frequencies to produce more `AddBlock` and `AddPerasCerts` commands.
2. Enhancing the `Cmd` generator to "remember" previously generated blocks so that generated gaps can actually be filled sometimes (this previously relied entirely on randomness).
3. Generating a geometrically-distributed security parameter (k) on the fly, while first addressing #1745. This includes a rationale for why we have chosen to sample k using a geometric distribution with `p=0.5`.

Please check the commit messages to find the empirical effect of these changes between commits. In short, we improve the chances of observing `TagSwitchedToShorterChain` from ~0.03% to about ~1.8%.

Closes #1745

### DIFF
--- a/ouroboros-consensus/changelog.d/20251124_140035_agustin.mista_improve_chaindb_qsm_tests.md
+++ b/ouroboros-consensus/changelog.d/20251124_140035_agustin.mista_improve_chaindb_qsm_tests.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Allow generating k>2 in ChainDB state machine tests on the fly.
+- Improve chances of switching to a shorter chain in ChainDB state machine tests.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->


ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -1832,26 +1852,30 @@ runCmdsLockstep loe (SmallChunkInfo chunkInfo) cmds =
...
+                          (Mock.numOpenHandles (nodeDBsImm fses) === 0)
+                        .&&. counterexample
+                          "VolatileDB is leaking file handles"
+                          (Mock.numOpenHandles (nodeDBsVol fses) === 0)
+                        .&&. counterexample
+                          "LedgerDB is leaking file handles"
+                          (Mock.numOpenHandles (nodeDBsLgr fses) === 0)
+                        .&&. counterexample
+                          "There were registered clean-up actions"
+                          (remainingCleanups === 0)
```
agustinmista: NOTE: this hunk only contains formatting changes. 

The formatter in my editor is _very_ annoyed by this block not being properly formatted, so I need to remember to manually bypass format-on-save every time I touch this file. 

I would appreciate if we can merge this hunk, but I can also revert it back if nobody agrees :smiling_face_with_tear:

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -176,10 +178,18 @@ import Test.Util.WithEq
...
+newtype Persistent a = Persistent {unPersistent :: a}
+  deriving (Eq, Show, Functor)
+
 -- | Commands
 data Cmd blk it flr
-  = AddBlock blk
-  | AddPerasCert (ValidatedPerasCert blk)
+  = -- | Add a block, with (possibly) some gap blocks before it being created.
+    AddBlock blk (Persistent [blk])
+  | -- | Add a Peras cert for a block, with (possibly) some gap blocks before it being created.
```
geo2a:
```suggestion
  | -- | Add a Peras cert for a block, with (possibly) some gap blocks before it was created.
```

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -176,10 +178,18 @@ import Test.Util.WithEq
...
+-- can carry generator state forward between commands. See 'GenState' below for
+-- more details.
+newtype Persistent a = Persistent {unPersistent :: a}
+  deriving (Eq, Show, Functor)
+
 -- | Commands
 data Cmd blk it flr
-  = AddBlock blk
-  | AddPerasCert (ValidatedPerasCert blk)
+  = -- | Add a block, with (possibly) some gap blocks before it being created.
```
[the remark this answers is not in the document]
agustinmista: Solved in https://github.com/IntersectMBO/ouroboros-consensus/pull/1772/commits/a6d4a005aadaa17721c2cd5fc10bc1d3cfc49af6

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -760,12 +774,69 @@ type FollowerRef blk m r = Reference (Opaque (TestFollower m blk)) r
...
+-- However, 'quickcheck-state-machine' does not provide much support for this,
+-- so we manually carry it around as part of the evolving SUT's model--even if
+-- it's technically not part of the actual model we are trying to test against.
+--
+-- TODO: Explore if this can be improved by tweaking the API of
+-- 'quickcheck-state-machine' to allow for the same functionality to exist
+-- under the hood.
+data GenState blk
+  = GenState
+  { seenBlocks :: Map (HeaderHash blk) blk
```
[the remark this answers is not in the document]
agustinmista: Added a similar explanation (that doesn't focus on the before vs. after but rather on why we want this) in https://github.com/IntersectMBO/ouroboros-consensus/pull/1772/commits/a6d4a005aadaa17721c2cd5fc10bc1d3cfc49af6

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -912,9 +912,9 @@ generator ::
 generator loe genBlock m@Model{..} =
   At
     <$> frequency
-      [ (30, genAddBlock)
+      [ (100, genAddBlock)
```
dnadales: We might want to add here a comment similar to what's mentioned on `983856001a0f87dae23fef584270b9b5b4b395b6` about the justification for this frequency change.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -1767,56 +1768,75 @@ genBlk chunkInfo Model{..} =
         )
       ]
 
+genSecurityParam :: Gen SecurityParam
+genSecurityParam =
+  SecurityParam
+    . unsafeNonZero
+    . fromIntegral
+    . (+ 2) -- shift to the right to avoid degenerate cases
+    <$> geometric 0.5 -- range in [0, +inf); mean = 1/p = 2
```
dnadales: Maybe here it'd be a good point to explain why we chose a geometric distribution for generating `k`.

dnadales (commented):
Looks good to me. I left some comments where I think the code would benefit from additional explanations.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -176,10 +177,18 @@ import Test.Util.WithEq
...
+-- can carry generator state forward between commands. See 'GenState' below for
+-- more details.
+newtype Persistent a = Persistent {unPersistent :: a}
+  deriving (Eq, Show, Functor)
+
 -- | Commands
 data Cmd blk it flr
-  = AddBlock blk
-  | AddPerasCert (ValidatedPerasCert blk)
+  = -- | Add a block, with (possibly) some gap blocks leading to it.
```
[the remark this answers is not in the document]
geo2a: If I understand correctly, the gap blocks were a thing from the very inception of this test. This PR just makes their use a little more structured.

I think a reference to the excellent haddock of GenState would be great there.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
agustinmista: Added an adapted version of the analysis we included in the commit message of https://github.com/IntersectMBO/ouroboros-consensus/pull/1772/commits/842c8a9660a776644f3078e99262dd59cd79cba3 in https://github.com/IntersectMBO/ouroboros-consensus/pull/1772/commits/d521432cf90f1035a59f64e241cdb30189206642

agustinmista:
Thanks @dnadales for the review! I added some extra documention to the code (derived mainly from the existing commit messages) in https://github.com/IntersectMBO/ouroboros-consensus/commit/d521432cf90f1035a59f64e241cdb30189206642.

Let me know if you're happy with these changes and I will squash them into their corresponding commits.

geo2a (approved):

PR #1780, opened 2025-11-26
[Peras 7] Introduce and propagate feature flags down to NodeKernelArgs

agustinmista:
This PR brings in `cardano-base` and propagates a set of `CardanoFeatureFlag`s from the top-level `RunNodeArgs` down to the `NodeKernelArgs`. 

This is currently needed by upcoming PRs that need to conditionally enable certain behavior based on the `PerasFlag` feature flag (e.g, spawning voting and object diffusion threads, deciding when a node can consider itself as CaughtUp, etc.).


jasagredo (approved):
This looks correct to me.

geo2a (approved):

PR #1785, opened 2025-11-28
[Peras 10] Add pure voting rules

agustinmista:
This PR makes two main contributions:

* Adds a simple boolean predicate evaluator DSL that allows computing boolean predicates while returning evidence for the outcome. This is needed by the Peras voting rules to be able to trace why we did or did not enter/exit a cooldown period with good granularity. This DSL is accompanied by some unit tests.

* Implements the Peras voting rules (from the [CIP-0140](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0140/README.md#rules-for-voting-in-a-round)) in a pure fashion, given a `PerasVotingView` which abstracts the impure bits away and can be mocked-up for testing. This is  tested using randomly generated `PerasVotingViews` against a simpler model implemented directly over booleans that the real implementation must conform to. In the future, this model could be reified directly from the Agda spec.

### DIFF
--- a/ouroboros-consensus/changelog.d/20251216_133810_agustin.mista_pure_voting_rules.md
+++ b/ouroboros-consensus/changelog.d/20251216_133810_agustin.mista_pure_voting_rules.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+### Non-Breaking
+
+- Add explainable boolean predicate DSL and evaluator.
+- Add pure Peras voting rules and mocked up conformance tests.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+  , latestCertSeen :: !(WithOrigin cert)
+  -- ^ The most recent certificate seen by the voter.
+  , latestCertOnChain :: !(WithOrigin cert)
+  -- ^ The most recent certificate present in some chain.
+  , certRoundStart :: cert -> SlotNo
+  -- ^ Get the slot number at the start of the Peras round of a certificate.
+  , arrivalSlot :: cert -> SlotNo
+  -- ^ Get the arrival slot number of a certificate
+  , candidateExtendsCert :: cert -> Bool
+  -- ^ Does the candidate block extend the one boosted by a certificate?
```
tbagrel1: The signature, function name and documentation seem a bit misleading, since the "candidate block" appear nowhere here. I would have expected the candidate block to be a parameter of this function.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+-- | Construct a 'PerasVotingView'.
+--
+-- NOTE: this assumes that the client code computes all the needed inputs
+-- within the same STM transaction, or the results may be inconsistent.
+mkPerasVotingView ::
```
tbagrel1: At this point I still don't see what is the "candidate block" wrt. this function.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+    -- Does the candidate block extend the one boosted by a certificate?
+    --
+    -- This can be trivially tested by checking whether the certificate is
+    -- within the bounds of the chain prefix leading to the candidate block.
+    --
+    -- NOTE: in the case of an extremely old certificate boosting a block
+    -- beyond the immutable prefix, this could incorrectly return false even
+    -- if the voting candidate technically extends the certificate point.
+    -- However, this a boring case that we can safely ignore. Conversely,
+    -- the case of a certificate that's too new to be voted for is covered
```
tbagrel1: This is not very clear to me if this is a precondition to using this function ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+    , currRoundNo
+    , latestCertSeen
+    } =
+    VR2A := vr2a
+   where
+    vr2a =
+      -- NOTE: we use 'succWithOrigin' and '-1' to handle the 'Origin' case
+      -- (i.e. when we have never seen a certificate before) correctly,
+      -- treating the 'Origin' certificate as being from round -1.
+      (succWithOrigin (getPerasCertRound <$> latestCertSeen) - 1 + _R)
```
tbagrel1: Why don't we use a `Origin -> Bool True` case here as we did previously? IMO it would make the code clearer.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Pred.hs
```diff
@@ -0,0 +1,234 @@
...
+infixr 5 `Not`
+
+infix 5 :>:
+
+infix 5 :<=:
+
+infix 5 :==:
+
+-- | Sufficient evidence to show that a predicate is either true or false
+type Evidence a = Pred a
```
tbagrel1: Do we really need this alias?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Pred.hs
```diff
@@ -0,0 +1,234 @@
...
+-- Right (Not (B := 4 :==: 5))
+--
+-- >>> evalPred $ Not (a :/\: c) -- failure because both a~True and c~True
+-- Left (Not ((A := Bool True) :/\: (C := 10 :>: 5)))
+evalPred :: Pred tag -> Either (Evidence tag) (Evidence tag)
+evalPred = \case
+  tag := p' ->
+    lift (tag :=) id p'
+  p@(Bool b) ->
+    boolean b p
```
tbagrel1:
```suggestion
  p@(Bool True) -> Right p
  p@(Bool False) -> Left p
```

tbagrel1 (commented):
I really like the general idea of making voting rules as clear as possible, with a clearly defined view interface!

But I feel like it's hard to get away from implementation details sometimes. Typically `mkPerasVotingView` is probably the most brittle part of this infrastructure (given its many preconditions), and I don't think we have a very satisfying way to aleviate this, except more documentation.

On a different note, I like the usability/user interface of the predicate DSL, I think it definitely suits this use-case. But I think we are in a weird middle with no type-level safety over evidence construction, so I'm not sure what the evidence feature brings for us in its current form. I may have missed something though.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+        pureQuery (fromPerasEnabledOrFail <$> HF.perasRoundNoToSlot roundNo)
+
+      fromPerasEnabledOrFail =
+        fromPerasEnabled
+          (error "mkPerasVotingView: Peras is disabled in the current era")
+
+    -- A slightly safer version of 'roundStart' that works with certificates.
+    --
+    -- NOTE: Instead of using 'roundStart' directly, this makes it more
+    -- harder for the user of the voting interface to accidentally pass a
```
nbacquey: typo: `more harder`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+    VR2B -> "VR-2B"
+  explain Deep = \case
+    VR1A ->
+      "voter has seen the certificate for the previous round in time"
+    VR1B ->
+      "the block being voted upon extends the most recently certified block"
+    VR2A ->
+      "the last certificate seen is sufficiently old"
+    VR2B ->
+      "the last certificate on chain is exactly one or more cooldown periods old"
```
nbacquey: `exactly one or more` is unclear. I would say something like `is exactly k cooldown periods old, for some positive integer k`. Maybe `k` could be part of the `VR2B` constructor, with something like
```haskell
data PerasVotingRule = VR1A | VR1B | VR2A | VR2B Int
```

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,319 @@
...
+tests =
+  adjustQuickCheckTests (* 100) $
+    testGroup
+      "Peras voting rules"
+      [ testProperty "isPerasVotingAllowed" prop_isPerasVotingAllowed
+      ]
+
+{-------------------------------------------------------------------------------
+  Model conformance test property
+-------------------------------------------------------------------------------}
```
nbacquey: Same remark about variable names here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+-- | Interface needed to evaluate the Peras voting rules
+data PerasVotingView cert = PerasVotingView
+  { perasParams :: PerasParams
+  -- ^ Peras protocol parameters.
+  , currRoundNo :: !PerasRoundNo
+  -- ^ The current Peras round number.
+  , latestCertSeen :: !(WithOrigin cert)
+  -- ^ The most recent certificate seen by the voter.
+  , latestCertOnChain :: !(WithOrigin cert)
+  -- ^ The most recent certificate present in some chain.
```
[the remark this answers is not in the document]
agustinmista: According to the CIP: 

```
cert∗ to the certificate with the highest round number present in C pref
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+
+    -- Does the candidate block extend the one boosted by a certificate?
+    --
+    -- This can be trivially tested by checking whether the certificate is
+    -- within the bounds of the chain prefix leading to the candidate block.
+    --
+    -- NOTE: in the case of an extremely old certificate boosting a block
+    -- beyond the immutable prefix, this could incorrectly return false even
+    -- if the voting candidate technically extends the certificate point.
+    -- However, this a boring case that we can safely ignore. Conversely,
```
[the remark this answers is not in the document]
agustinmista: Discussed in [this issue](https://github.com/tweag/cardano-peras/issues/186)

We will add a TODO here to make sure we don't forget this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+  { perasParams :: PerasParams
+  -- ^ Peras protocol parameters.
+  , currRoundNo :: !PerasRoundNo
+  -- ^ The current Peras round number.
+  , latestCertSeen :: !(WithOrigin cert)
+  -- ^ The most recent certificate seen by the voter.
+  , latestCertOnChain :: !(WithOrigin cert)
+  -- ^ The most recent certificate present in some chain.
+  , certRoundStart :: cert -> SlotNo
+  -- ^ Get the slot number at the start of the Peras round of a certificate.
```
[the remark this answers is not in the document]
agustinmista: This is no longer part of the interface :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: Turned all 'succWithOrigin` into explicit `case` expressions.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Pred.hs
```diff
@@ -0,0 +1,246 @@
...
+--
+-- >>> evalPred (b :\/: (Not c)) show  -- failure because both b~False and c~True
+-- EFalse "((B := 4 :==: 5) :\/: Not (C := 10 :>: 5))"
+--
+-- >>> evalPred (Not (a :/\: b)) show  -- success because b~False
+-- ETrue "(Not (B := 4 :==: 5))"
+--
+-- >>> evalPred (Not (a :/\: c)) show  -- failure because both a~True and c~True
+-- EFalse "(Not ((A := Bool True) :/\: (C := 10 :>: 5)))"
+evalPred :: Pred tag -> (forall res. Evidence res tag -> r) -> r
```
tbagrel1: I have no strong opinion on this, so feel free to ignore, but we could have kept the `Either`, right?

Is the CPS-style adding any benefit?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,537 @@
...
+-- | View of the latest certificate seen by the voter
+data LatestCertSeenView cert
+  = LatestCertSeenView
+  { lcsCert :: !cert
+  -- ^ Latest certificate seen by the voter
+  , lcsArrivalSlot :: !SlotNo
+  -- ^ Slot number at which this certificate was received
+  , lcsRoundStartSlot :: !SlotNo
+  -- ^ Starting slot number of the round where this certificate was received
+  , lcsCandidateBlockExtendsCert :: !Bool
```
tbagrel1: I'm still not a fan of having a field of the structure that refers to something external/implicit (the candidate block). But there is probably a good reason to do so here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,537 @@
...
+    -- 'chainAtCandidateBlock' fragment.
+    --
+    -- NOTE: in the case of an extremely old certificate boosting a block beyond
+    -- the immutable prefix, this could incorrectly return false even if the
+    -- voting candidate technically extends the certificate's boosted block.
+    -- This case /should not/ affect the voting rules: this value is only used
+    -- for the "happy path", and the incorrect value can only happen during a
+    -- cooldown period where the "happy path" rules are not relevant.
+    --
+    -- Nevertheless, we should fix this by enhancing our view of the latest
```
tbagrel1: I would prefer if `mkPerasVotingView` could take a boolean (or use a newtype around `lastCertSeen`) to indicate whether or not it is pointing to a block beyond the immutable horizon.

This way we put the responsiblity for a fix/mitigation on the caller, not the internal implementation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,537 @@
...
+    -- certificate seen to know whether it is older than the volatile suffix:
+    -- TODO: https://github.com/tweag/cardano-peras/issues/186
+    candidateBlockExtendsCert cert =
+      AF.withinFragmentBounds
+        (castPoint (getPerasCertBoostedBlock cert))
+        chainAtCandidateBlock
+
+-- | Reason for voting being disallowed
```
tbagrel1: I don't think the newtypes around `Evidence` are very useful here.

I would either make a simple sum type `data VotingStatusWithEvidence = Vote (Evidence True PerasVotingRule) | NoVote (Evidence False PerasVotingRule)`,

or define `type VotingStatusWithEvidence b = Evidence b PerasVotingRule` and return `Either (VotingStatusWithEvidence 'True) (VotingStatusWithEvidence 'False)`.

But I think the first option is better

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,313 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Test that the Peras voting rules can correctly decide when to vote.
+--
+-- NOTE: in this file, we use uncommon variable names such as `_L` or `_X`
```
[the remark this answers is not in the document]
tbagrel1: (same above in the actual implementation)

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,313 @@
...
+    ( vr1a && vr1b || vr2a && vr2b
+    , (vr1a, vr1b, vr2a, vr2b)
+    )
+   where
+    vr1a =
+      vr1a1 && vr1a2
+    vr1a1 =
+      case latestCertSeen of
+        NotOrigin cert ->
+          currRoundNo + 1 == getPerasCertRound (lcsCert cert)
```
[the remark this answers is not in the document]
tbagrel1: Btw, if I'm right that this is a mistake, we should probably add a test for that :)

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,313 @@
...
+
+genPerasRoundNo :: Gen PerasRoundNo
+genPerasRoundNo = do
+  n <- arbitrary
+  pure (PerasRoundNo n)
+
+-- * Mocked certificate type
+
+-- NOTE: we could also use the real 'WithArrivalTime (ValidatedPerasCert blk)'
+-- here. However, this one is much easier to derive a 'Function' instance for,
```
tbagrel1: Does the justification still holds now that the view is pure? Otherwise we should probably use the official type :)

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,313 @@
...
+-- we generate certificates relative to a given Peras round (the current one).
+genTestCert :: PerasRoundNo -> Gen TestCert
+genTestCert roundNo = do
+  arrivalTime <- RelativeTime <$> genNominalDiffTime50Years
+  offset <- choose @Integer (-10, 3)
+  -- NOTE: here we need to be careful not to underflow the round number
+  let roundNo' =
+        PerasRoundNo $
+          fromIntegral $
+            max 0 $
```
tbagrel1: Are you sure `max 0` will help for underflow? Since `0 + (-1) = maxBound` for unsigned types

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,537 @@
...
+-- | Monad for querying Peras-related information from a hard fork summary.
+--
+-- This covers two possible error cases:
+-- 1. the hard fork summary does not cover the needed point in time, or
+-- 2. Peras is not enabled at the needed point in time.
+newtype PerasQry xs a
+  = PerasQry
+  { runPerasQry ::
+      HF.Summary xs ->
+      Either HF.PastHorizonException (PerasEnabled a)
```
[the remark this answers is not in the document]
agustinmista: I agree. This was also raise during review for #1681, which introduces this type. 

Maybe we can open a low-prio issue to rename it?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: I have already added: 

```haskell
-- NOTE: note that the voting rules do not need to know _which_ candidate block
-- is being voted for, only whether it extends the block boosted by the most
-- recent certificate seen by the voter, which is provided to the rules via
-- the 'lcsCandidateBlockExtendsCert' field inside 'latestCertSeen'.
```

To the docstring of `PerasVotingView`. Would you like me to add a pointer to that docstring instead?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: Replace the newtypes with a GADT `PerasVotingRulesDecision`:

```haskell
data PerasVotingRulesDecision
  = Vote (Evidence True PerasVotingRule)
  | NoVote (Evidence False PerasVotingRule)
  deriving Show
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Pred.hs
agustinmista: Using CPS is the (IMO) standard way to introduce existential types. 

FTR: using 

```haskell
evalPred :: Pred tag -> Either (Evidence True tag) (Evidence False tag)
```
also works, but breaks in a more annoying way if we want to add a new constructor to `Evidence` in the future (because we either need to introduce a three-way either, or to bite the bullet an use CPS). Adding a new constructor when you have everything already written using CPS is cleaner, as we just get warnings about non-exhaustive case expressions.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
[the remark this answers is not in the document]
agustinmista: Opened https://github.com/tweag/cardano-peras/issues/189

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,398 @@
...
+          (currRoundNo :>: getPerasCertRound cert)
+            -- The certificate round is c⋅K rounds away from the current one
+            :/\: (currRoundNo `rmod` _K :==: getPerasCertRound cert `rmod` _K)
+        -- There is no certificate on chain ==> check if we are recovering
+        -- from an initial cooldown after having initially failed to
+        -- reach a quorum during bootstrapping.
+        --
+        -- NOTE: '_K - 1' here is treating the 'Origin' certificate as being
+        -- from round -1.
+        Origin -> currRoundNo `rmod` _K :==: _K - 1
```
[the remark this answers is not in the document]
agustinmista: There is no "initial" certificate being created for round -1. Here we are just "pretending" there is one to account for the `cert_genesis` defined in the [CIP](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0140/README.md#non-normative-overview-of-the-peras-protocol). 

This should allow the voting rules to bootstrap fresh nodes: they can start voting even if they haven't seen a certificate yet, but only if the round matches either origin, or some c*K cooldown periods after it (to account for initially failed voting rounds).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+                     :==: (getPerasCertRound (lcocCert cert) `rmod` _K)
+                 )
+        -- There is no certificate on chain ==> check if we are recovering
+        -- from an initial cooldown after having initially failed to
+        -- reach a quorum during bootstrapping.
+        --
+        -- NOTE: '_K - 1' here is treating the 'Origin' certificate as being
+        -- from round -1.
+        Origin ->
+          currRoundNo `rmod` _K :==: _K - 1
```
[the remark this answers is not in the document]
tbagrel1: Answer: we must have an offset of `c *_K` since the last _successful_ vote, which means a delay of `c * _K - 1` between a failed vote and a new attempt

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | Pure Peras voting rules
+--
+-- This module implements the Peras voting rules in a pure fasion, along with
```
dnadales:
```suggestion
-- This module implements the Peras voting rules in a pure fashion, along with
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+-- the necessary inpure machinery to retrieve their inputs. These rules are
```
dnadales:
```suggestion
-- the necessary impure machinery to retrieve their inputs. These rules are
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+  Voting helpers based on a hard fork summary
+-------------------------------------------------------------------------------}
+
+-- | Monad for querying Peras-related information from a hard fork summary.
+--
+-- This covers two possible error cases:
+-- 1. the hard fork summary does not cover the needed point in time
+--    (case @Left ...@), or
+-- 2. Peras is not enabled at the needed point in time
+--    (case @Right NoPerasEnabled@).
```
fraser-iohk: is this considered an actual error in this case?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+newtype PerasQry xs a
+  = PerasQry
+  { runPerasQry ::
+      HF.Summary xs ->
+      Either HF.PastHorizonException (PerasEnabled a)
+  }
```
bladyjoker: A couple of questions on `PerasQry xs a`. 

1. Why not return `Either PerasQryException a` and enumerate the error conditions

```haskell
data PerasQryException = PerasQryExceptionPastHorizon HF.PastHorizonException | PerasQryExceptionPerasDisabled

newtype PerasQry xs a
  = PerasQry
  { runPerasQry ::
      HF.Summary xs ->
      Either PerasQryException a
```

With that in mind, to me this sounds like it could be modeled simply as

```haskell
newtype PerasQry xs a = PeraseQuery {
  unPerasQry :: ReaderT (HF.Summary xs) (Either PerasQryException) a
} deriving newtype (Functor, Applicative, Monad)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+-- This type additionally carries the evidence for the decision taken.
+data PerasVotingRulesDecision
+  = Vote (Evidence True PerasVotingRule)
+  | NoVote (Evidence False PerasVotingRule)
+  deriving Show
+
+instance Explainable PerasVotingRulesDecision where
+  explain mode = \case
+    Vote (ETrue e) -> "Vote(" <> explain mode e <> ")"
+    NoVote (EFalse e) -> "NoVote(" <> explain mode e <> ")"
```
bladyjoker: At this point I don't know where this output will land...however, let's say I'm analyzing Peras related logs and I want to fiter/group/aggregate on Peras Voting Evidence. This is making my life much more difficult.

What I'm suggesting is, if this is to be written to a trace and potentially consumed downstream, then I'd want to see a machine readable version of these things (JSON) alongside their human readable version.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: Thanks, I was hoping you'd suggest this :smile:

The reason we initially introduced `PerasQry` was to keep the implementation of `mkPerasVotingView` clean from having to deal with nested errors, but didn't care too much for its resulting type because the client could still pattern match on all the nested variants to extract a result. 

That said, I also prefer the idea of using a well-established monadic stack (at the cost of having a new exception type and a bit of extra plumbing in the monad's derived operations). This should make the client code easier to understand.

I have now implemented your suggestion :smile_cat:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -0,0 +1,563 @@
...
+  deriving Show
+
+-- | Interface needed to evaluate the Peras voting rules
+--
+-- NOTE: the voting rules depend on the candidate block indirectly. This is
+-- reflected in the fact that the voting view does not contain the candidate
+-- block or its point, but only whether the candidate block extends the block
+-- boosted by the most recent certificate seen by the voter, which is provided
+-- to the rules via 'lcsCandidateBlockExtendsCert' inside 'latestCertSeen'.
+data PerasVotingView cert = PerasVotingView
```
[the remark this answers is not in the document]
agustinmista: I moved all the non-rule stuff into `O.C.Peras.Voting.View`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: I agree with you: structured logging should be the norm and not the exception.

> Check out https://cardano-node.cardano.intersectmbo.org/trace-dispatcher-2.11.1-inplace/Cardano-Logging.html#t:LogFormatting

Interesting. I couldn't find any mention of anything related to `LogFormatting` or `trace-dispatcher` in `ouroboros-consensus`, so I wonder if other components here are implementing any specialized `forMachine`/`forHuman` rendering of their traces.

In the case of `Pred PerasVotingRule`, we can also choose to trace both the human-readable explanantion + its `show` representation (which I think is what most traces currently do? Just `show` stuff?)

bladyjoker: Absolutely, trace-dispatcher is something we are considering using directly, but it's not yet adopted in the codebase. You definitely don't need to implement this, but bare in mind the API we eventually want to hook into.

PR #1788, opened 2025-12-01
[Peras 12] Store previous epoch nonce in PraosState

agustinmista:
This PR extends the existing Praos state with an extra field storing the nonce from the previous epoch. This nonce is updated by 'tickChainDepState' at every new epoch boundary. This is needed to validate Peras certificates when they appear in blocks.

Note that the CBOR serialization had to be adapted to include this new field, making this a **breaking change**.

For more context, see https://github.com/tweag/cardano-peras/issues/150

### DIFF
--- a/ouroboros-consensus-protocol/changelog.d/20251222_093851_agustin.mista_store_previous_epoch_nonce.md
+++ b/ouroboros-consensus-protocol/changelog.d/20251222_093851_agustin.mista_store_previous_epoch_nonce.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
+
+- Store previous epoch nonce in PraosState and adapt its serialization format.
+


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs
```diff
@@ -423,11 +428,13 @@ instance PraosCrypto c => ConsensusProtocol (Praos c) where
...
   --
   -- If we are not in a new epoch, then nothing happens. If we are in a new
-  -- epoch, we do two things:
+  -- epoch, we do three things:
   -- - Update the epoch nonce to the combination of the candidate nonce and the
   --   nonce derived from the last block of the previous epoch.
-  -- - Update the "last block of previous epoch" nonce to the nonce derived from
-  --   the last applied block.
+  -- - Store the current epoch nonce as the "previous epoch" nonce. This is
+  --   needed by Peras to be able to validate slightly old certificates.
```
tbagrel1: "slightly old" is not very precise, is there an issue or part of the design document we could refer to ?

tbagrel1 (approved):
Code looks good to me!

As stated previously (when we merged it into `peras-staging`), I don't know if there is any way to review the binary files?

tbagrel1:
Needs to update the CDDL spec, as suggested during the weekly consensus meeting of 14/01/2026

bladyjoker (approved):
Thanks mate!

bladyjoker:
> @bladyjoker Hey, I've just updated the CDDL spec [`praos.cddl`](https://github.com/IntersectMBO/ouroboros-consensus/blob/43173d3344a0292856e4a378851ecbd9dc62b7e3/ouroboros-consensus-cardano/cddl/disk/ledger/praos.cddl) in [43173d3](https://github.com/IntersectMBO/ouroboros-consensus/commit/43173d3344a0292856e4a378851ecbd9dc62b7e3), but I'm wondering if I should also update [`tpraos.cddl`](https://github.com/IntersectMBO/ouroboros-consensus/blob/43173d3344a0292856e4a378851ecbd9dc62b7e3/ouroboros-consensus-cardano/cddl/disk/ledger/tpraos.cddl)?
> 
> I don't really understand what is the relationship between these two files. Thanks!

Good question! In Consensus you already realize we abstract over types of blocks, perhaps you missed that [type of blocks](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus-cardano/Ouroboros-Consensus-Cardano-Block.html#t:CardanoEras) includes ledger/crypto/consensus information.

TPraos is an older variant that was used before Praos

https://github.com/IntersectMBO/ouroboros-consensus/blob/main/docs/website/contents/explanations/index.md

> A core design principle in the implementation of these components is the abstraction from specific ledger and protocol implementations. The aim is to decouple the consensus protocol from the ledger and to support multiple consensus algorithms and ledgers for improved adaptability and maintainability. This design allows different ledgers (like the Byron or Shelley ledgers) and different Ouroboros protocol instances (like Praos or TPraos) to be integrated into the abstract consensus framework. 

I don't think you need to update `tpraos.cddl` which is a local storage concern. But pinging @jasagredo to confirm.

PR #1812, opened 2025-12-16
More consistent traces between immdb-server and cardano-node

ch1bo:
Aligns `MsgLeiosBlock` and `MsgBlock` traces between patched `immdb-server` and `cardano-node` such that both have similar contents and at least contain block hashes.

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Json/SendRecv.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Json/SendRecv.hs
@@ -16,7 +16,7 @@ data Direction = Recv | Send
 data SendRecvEvent muxtm tm cnt = MkSendRecvEvent {
         at           :: tm
         -- ^ when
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Json/SendRecv.hs
```diff
@@ -15,19 +14,19 @@ data Direction = Recv | Send
...
-      , mux_at :: muxtm
+      , mux_at       :: Maybe muxtm
         -- ^ For 'Recv', when these bytes left the demuxer.
         --
         -- For 'Send', when the mini protocol began trying to put these bytes
         -- into the muxer.
-      , direction :: Direction
-      , msg :: String
+      , direction    :: Direction
+      , msg          :: Aeson.Value
```
bladyjoker: Not sure why this isn't a type var `msg` with `Aeson.(To|From)Object msg` instances? But not really important at this point.

bladyjoker (approved):
LGTM

PR #1815, opened 2025-12-18
Add more accurate tracing to LSM

jasagredo:


jasagredo:
This is mostly sequential, if not entirely, but this lets us disambiguate whether a deadlock happens in LSM or in Consensus code.

PR #1818, opened 2025-12-24
Use the memoized hash, instead of recomputing it from original bytes

lehins:
# Description

Prior implementation was suboptimal. We already provide memoized hash of the block header. With these changes not only memoized version of the hash will be used, but also it will not result in two redundant memcopies just for the sake of hash computation (MemoBytes uses `ShortByteString` underneath, so in order to compute the hash through CBOR it needs to be first converted to a lazy `ByteString` and then again converted to a strict `ByteString`, which can later be passed over FFI for hash computation)

Also note that:


bladyjoker (approved):
Hi @lehins, thanks for the PR!

PR #1820, opened 2025-12-29
Cardano-config

jasagredo:
# Description

this PR might resolve.

Also note that:


cardano-config/cardano-config.cabal
```diff
@@ -0,0 +1,82 @@
...
+
+common warnings
+  ghc-options:
+    -Wall
+    -Wunused-packages
+
+library
+  import: warnings
+  exposed-modules:
+    Cardano.Node.Configuration
```
carbolymer: Maybe
```suggestion
    Cardano.Configuration
```
?

`Cardano.Node...` would suggest that the modules are placed in `cardano-node` package.

cardano-config/src/Cardano/Node/Configuration/CLIArgs.hs
```diff
@@ -0,0 +1,361 @@
...
+  , byronSigningKey :: Maybe FilePath
+  , shelleyKESKey :: Maybe FilePath
+  , shelleyVRFKey :: Maybe FilePath
+  , shelleyOperationalCertificate :: Maybe FilePath
+  , bulkCredentialsFile :: Maybe FilePath
+  }
+  deriving Show
+
+-- | The CLI arguments, parsed with 'parseCLIArgs'
+data CLIArgs = CLIArgs
```
carbolymer:
```suggestion
data CliArgs = CliArgs
```
Same remark as for the module name

cardano-config/src/Cardano/Node/Configuration/Basics.hs
```diff
@@ -0,0 +1,66 @@
...
+import Data.Default
+import Data.String (IsString)
+import GHC.Generics (Generic)
+import GHC.TypeLits (Symbol)
+import qualified System.FilePath as F
+import Prelude hiding (FilePath)
+import qualified Prelude as P
+
+-- | A filepath annotated with the intended use for it
+newtype FilePath (s :: Symbol) = FilePath {getFilePath :: P.FilePath}
```
carbolymer: The current `FilePath` conflicts with `System.FilePath.FilePath`. Using a different name would make it a bit unambiguous. What do you think about using a shorter name `File`? 
```suggestion
newtype File (s :: Symbol) = File {getFile :: P.FilePath}
```

cardano-config/src/Cardano/Node/Configuration/Basics.hs
```diff
@@ -0,0 +1,66 @@
...
+(</>) :: FilePath a -> FilePath b -> FilePath c
+FilePath a </> FilePath b = FilePath (a F.</> b)
+
+-- | A FilePath that requires a mount point
+newtype RelativeFilePath (s :: Symbol)
+  = RelativeFilePath {relativeFilePath :: FilePath s}
+  deriving (Generic, Show)
+
+anchorRelativePath :: FilePath a -> RelativeFilePath b -> FilePath b
+anchorRelativePath fp1 (RelativeFilePath fp2) = fp1 </> fp2
```
carbolymer: There's a footgun here if someone puts an absolute path into `fp2`: `fp1` will be ignored https://hackage-content.haskell.org/package/filepath-1.5.4.0/docs/System-FilePath.html#v:-60--47--62-

I don't know if there's an easy way to protect against, which is portable. Checking for relative paths requires `IO` access afair. But a haddock with a warning would be nice. 😉

cardano-config/src/Cardano/Node/Configuration/Basics.hs
```diff
@@ -0,0 +1,66 @@
...
+          _ -> fail "Not text"
+      )
+      v
+      <|> Override <$> parseJSON v
+
+--------------------------------------------------------------------------------
+
+-- | If the key is missing, use the default value
+(.:=) :: (FromJSON a, Default a) => Object -> Key -> Parser a
+a .:= b = a .:? b .!= def
```
carbolymer: I think adding a precedence would be nice.
```suggestion
a .:= b = a .:? b .!= def
infixr 8 .:=
```

cardano-config/src/Cardano/Node/Configuration/File.hs
```diff
@@ -0,0 +1,135 @@
...
+          fn <- v' .: fromString sectionName
+          pure $
+            SubFile $ do
+              exists <- doesFileExist $ root </> fn
+              if exists
+                then
+                  parseEither p
+                    . either (\err -> error $ "Invalid JSON: " <> show err) id
+                    . decodeEither'
+                    <$> BS.readFile (root </> fn)
```
carbolymer: So this expression can be:
- an IO exception
- a bottom from `error` wrapped in `Left` (in `IO`)
- a parser failure as `Left` (in `IO`)
- a successful parse as `Right` (in `IO`)



What's the advantage of having three different levels of errors here? Why not just work here in `IO a`?

All those different kinds of errors make this function just harder to reason about without much benefit IMO.

Shouldn't the second and third type of errors be short-circuited to the topmost parser instead i.e. this one  `Parser (SubFileOrValue a)`?

cardano-config/src/Cardano/Node/Configuration/File/Consensus.hs
```diff
@@ -0,0 +1,55 @@
...
+  | GenesisMode GenesisConfigFlags
+  deriving (Generic, Show)
+
+instance FromJSON ConsensusConfiguration where
+  parseJSON =
+    withObject "ConsensusMode" $ \v -> do
+      v .:? "ConsensusMode" >>= \case
+        Just "GenesisMode" -> GenesisMode <$> parseJSON (Object v)
+        Just "PraosMode" -> pure PraosMode
+        Nothing -> pure def
```
carbolymer: I am not convinced that putting default values in JSON instances is a good idea. It conflates defaulting with deserialisation. I think those should be separated. Defaulting can be implemented like that: https://chrispenner.ca/posts/hkd-options#setting-default-values

carbolymer (commented):
Good effort Javier!

cardano-config/src/Cardano/Node/Configuration.hs
```diff
@@ -0,0 +1,131 @@
...
+  , hostAddr :: Maybe IPv4
+  , hostIPv6Addr :: Maybe IPv6
+  , port :: Maybe PortNumber
+  , tracerSocket :: Maybe CLI.TracerConnection
+  , shutdownIPC :: Maybe Fd
+  , shutdownOnTarget :: Maybe CLI.ShutdownOn
+  }
+
+maybeFirst :: Maybe a -> Maybe a -> Maybe a
+maybeFirst a b = getFirst $ ((<>) `on` First) a b
```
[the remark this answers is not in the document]
jasagredo: Good one!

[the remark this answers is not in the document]
jasagredo: Done

cardano-config/src/Cardano/Node/Configuration/Basics.hs
jasagredo: Done

cardano-config/src/Cardano/Node/Configuration/File/Consensus.hs
jasagredo: Good point. I have added some intermediate `Maybe`s to signal that some things still expect a default, and only when resolving the configuration at the end I pull in the default for those.

crocodile-dentist:
I think it's a logical design, good work!

jasagredo:
Obsoleted by #2080

PR #1824, opened 2026-01-09
Introduce TxHash in Mempool

### LINKED ISSUE #1444: Add TxHash to MempoolSnapshot
The Mempool should provide the hash of the full transaction in addition to the `TxId` (ie the hash of just the transaction's body) and the size. This data will enable more optimizations in the centralized TxSubmission logic.

bladyjoker:
# Description

See #1444 

`TxHash` identifies a transaction uniquely up to its contents (aux, witnesses etc). That's what makes it different to `TxId` which identifies a transaction up to its ledger effect (in Ledger that means the same `TxBody` and in Plutus that means the same `TxInfo`).

See [prop_tx_id_valid_effect_is_same](https://github.com/IntersectMBO/ouroboros-consensus/pull/1824/changes/466932601c8d1fd41d70c9f2a9e33c3e1754f358#diff-d92b4a547e2fb345037f07749db629d4ccd6bfa378af03d2c32b308ce1b2ba6cR525) and [prop_tx_hash_effect_is_same](https://github.com/IntersectMBO/ouroboros-consensus/pull/1824/changes/466932601c8d1fd41d70c9f2a9e33c3e1754f358#diff-d92b4a547e2fb345037f07749db629d4ccd6bfa378af03d2c32b308ce1b2ba6cR547) for their exact properties and how they differ.

In a nutshell, we anticipate we might want to diffuse transactions identified via `TxHash` for Leios as we want to make sure only the validated transactions are diffused. Since `TxId` obtained from node-A and node-B can have a different witness set and one might fail while the other doesn't.

DONE
- [x] Add `GenTxHash` following the same pattern as `GenTxId` and wire the types and concrete implementations
- [x] Align coding styles among different implementations (byron, shelley, test, hfc)
- [ ] Test properties of `MempoolSnapshot` and `LedgerSupportsMempool`
  - [x] Specify the properties
  - [ ] Wire them in a concrete testsuite
- [ ] !!! Decide how to proceed given there's no immediate users of this feature and it's unclear there would be.
- [ ] Optimization: Pass around the serialized transaction to spare us from unnecessarily re-serializing.


jasagredo:
Closed (it will remain in the closed list)

PR #1826, opened 2026-01-13
Expose reason for a switch in ChainSel tracer

jasagredo:
While the performance team was looking into https://github.com/IntersectMBO/cardano-node/issues/6014, it became apparent that it is unreasonable to ask the tracing system to infer why a switch has happened, and instead Consensus could provide the reason.

This PR adds a field to `AddedToCurrentChain` and `SwitchedToAFork` that shows the reason for the switch.

### DIFF
--- a/scripts/ci/check-changelogs.sh
+++ b/scripts/ci/check-changelogs.sh
@@ -35,9 +35,9 @@ else
     for p in "${tracked_libs[@]}"; do
         printf "\t- %s\n" "$p"
         if ! git diff --quiet --name-only "origin/${BASE_REF}" -- "$p/***.hs"; then
-            if ! git diff --quiet --name-only --diff-filter=A "origin/${BASE_REF}" -- "$(echo $p | cut -d/ -f1)/changelog.d" ; then
+            if ! git diff --quiet --name-only --diff-filter=AMR "origin/${BASE_REF}" -- "$(echo $p | cut -d/ -f1)/changelog.d" ; then
                 printf "\t\tNew fragments found. OK.\n"
-                git diff --name-only --diff-filter=A "origin/${BASE_REF}" -- "$p/changelog.d" | sed 's/^/\t\t- /g'
+                git diff --name-only --diff-filter=AMR "origin/${BASE_REF}" -- "$p/changelog.d" | sed 's/^/\t\t- /g'
             else
                 printf "\t\tNo new fragments found, but code changed. Please push a fragment or add the \"no changelog\" label to the PR. The diff follows:\n"
                 git --no-pager -c color.diff=always diff "origin/${BASE_REF}" -- "$p/***.hs" | sed 's/^/diff> /g'
@@ -50,9 +50,9 @@ else
     for p in "${packages[@]}"; do
         printf "\t- %s\n" "$p"
         if ! git diff --quiet --name-only "origin/${BASE_REF}" -- "$p/***.cabal"; then
-            if ! git diff --quiet --name-only --diff-filter=A "origin/${BASE_REF}" -- "$(echo $p | cut -d/ -f1)/changelog.d" ; then
+            if ! git diff --quiet --name-only --diff-filter=AMR "origin/${BASE_REF}" -- "$(echo $p | cut -d/ -f1)/changelog.d" ; then
                 printf "\t\tNew fragments found. OK.\n"
-                git diff --name-only --diff-filter=A "origin/${BASE_REF}" -- "$p/changelog.d" | sed 's/^/\t\t- /g'
+                git diff --name-only --diff-filter=AMR "origin/${BASE_REF}" -- "$p/changelog.d" | sed 's/^/\t\t- /g'
             else
                 printf "\t\tNo new fragments found, but code changed. Please push a fragment or add the \"no changelog\" label to the PR. The diff follows:\n"
                 git --no-pager -c color.diff=always diff "origin/${BASE_REF}" -- "$p/***.cabal" | sed 's/^/diff> /g'


bladyjoker (commented):
Heyoo! So one question for you, I'm looking at `ReasonForSwitchByTiebreaker` instance and think to myself: should we reverse engineer the underlying switching decision?

Did you consider changing, for example:

```haskell
comparePraos ::
  VRFTiebreakerFlavor ->
  PraosTiebreakerView c ->
  PraosTiebreakerView c ->
  Ordering
```

into 

```haskell
comparePraos ::
  VRFTiebreakerFlavor ->
  PraosTiebreakerView c ->
  PraosTiebreakerView c ->
  (Ordering, Maybe ReasonForTiebreakerSwitch)
```

If that makes sense? Then you just carry the decision around and trace it when needed? It's computed once basically.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs
bladyjoker: Just realized this is an ideal use case for the [Predicate DSL](https://github.com/IntersectMBO/ouroboros-consensus/blob/b781aa0f1174d2bc498f788c9add130377d4ae10/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Pred.hs). It's not yet merged, but would be noooice to have this formulated with that such that we gain 'explanations' for free.

dnadales: I was thinking the same 👍 Perhaps we could add a note linking to that Peras PR, so that we might consider using the DSL in the future.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
dnadales: This bit should be updated as well.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -47,83 +53,143 @@ data AcrossEraTiebreaker :: Type -> Type -> Type where
...
+-- result type. The result type indicates whether we expect something that
+-- changes between eras or some constant output.
+data AcrossEraMode cfg (f :: [Type] -> Type) where
+  AcrossEraCompare :: AcrossEraMode Proxy (ConstOutput Ordering)
+  AcrossEraPreferCandidate :: AcrossEraMode WrapChainOrderConfig SwitchOutput
+
+newtype OneEraReasonForSwitch xs = OneEraReasonForSwitch {getOneEraReasonForSwitch :: NS WrapReasonForSwitch xs}
+
+-- | Abstract the result of chain selection so we can handle both
+-- invariant results (Ordering) and covariant results (ShouldSwitch).
```
bladyjoker: I'm sure this comment makes sense, but perhaps next to the instance definitions it would enligthen the reader more.

This is a very generic class that relates to more than chain selection results right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs
```diff
@@ -47,83 +53,143 @@ data AcrossEraTiebreaker :: Type -> Type -> Type where
...
+class AcrossEraOutput (f :: [Type] -> Type) where
+  type SingleEraOutput f (x :: Type) :: Type
+
+  liftHead :: SingleEraOutput f x -> f (x ': xs)
+  liftTail :: f xs -> f (x ': xs)
```
bladyjoker: TODO: I'm not sure how to think about this? Come back when you read the instances

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -257,7 +266,23 @@ class Ord sv => ChainOrder sv where
     sv ->
     -- | Tip of the candidate
     sv ->
-    Bool
+    ShouldSwitch (ReasonForSwitch sv)
+
+data ShouldSwitch reason = ShouldNotSwitch Ordering | ShouldSwitch reason
```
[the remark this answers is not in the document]
jasagredo: I don't think is should be necessary. If you are not switching, the lifetime of the candidate ends there, there is nothing to report because you won't even consider the other candidate.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -267,7 +292,15 @@ newtype SimpleChainOrder sv = SimpleChainOrder sv
 instance Ord sv => ChainOrder (SimpleChainOrder sv) where
   type ChainOrderConfig (SimpleChainOrder sv) = ()
 
-  preferCandidate _cfg ours cand = ours < cand
+  -- All interesting protocols (PBFT, TPraos, Praos, HardForkProtocol) use a
+  -- proper TiebreakerView, so this is only for tests with Bft (or other testing
+  -- protocols), therefore we don't really elaborate the reasons.
+  type ReasonForSwitch (SimpleChainOrder sv) = ()
```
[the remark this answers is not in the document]
jasagredo: I don't think this is worth it for the little impact it will make (only testing unreal protocols)

dnadales (approved):

PR #1839, opened 2026-01-26
[Peras 16] Implement committee selection scheme model

agustinmista:
This PR provides a model implementation of the weighted Fait Accompli committee selection scheme that will be used for Peras, where:

* Some "persistent" voters are selected deterministically based solely on their stake (at the beginning of every epoch), and
* Some residual "non-persistent" voters randomly select themselves privately via local sortition (at the beginning of every Peras round).   

The main goal is to use this model as an oracle for conformance testing of the real implementation in the future, where the deterministic parts could be matched 1:1 to the model, whereas the non-deterministic local sortition might have to be subject to some statistical regression analysis to validate its output distribution (this is yet to be discussed).

Finally, we provide some general QuickCheck properties to verify the validity of the deterministic parts of the algorithm.


tbagrel1 (commented):
I gave a quick look at the Model implementation, and left a few comments.

Very nice work! Just sometimes I'm having a bit of trouble to follow deferred declaration/abstractions.

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Committee/Model.hs
```diff
@@ -0,0 +1,612 @@
...
+instance Show VoterId where
+  show (VoterId i) = "#" <> show i
+
+-- ** Ledger and voting stakes
+
+-- | Voter stake type
+data StakeOrigin = Ledger | Voting
+  deriving Show
+
+data StakeType = Absolute | AbsoluteCumulated | Normalized
```
nbacquey: The distinction between `Absolute` and `AbsoluteCumulated` is not very clear to me

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Committee/Model.hs
```diff
@@ -0,0 +1,642 @@
...
+  -- | Selected persistent and non-persistent seats with their corresponding
+  -- voting stakes
+  ( StakeDistr 'Voting 'Persistent
+  , StakeDistr 'Voting 'Residual
+  )
+weightedFaitAccompliWith fallback globalNumSeats stakeDistr
+  | globalNumSeats <= 0 =
+      error "Number of seats must be positive"
+  | Map.size stakeDistr == 0 =
+      error "Stake distribution cannot be empty"
```
nbacquey: Even if the property tests don't generate such cases, we should check that the stake distribution sums to non-zero as well.

nbacquey (approved):
Otherwise lgtm

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Committee/Model.hs
```diff
@@ -0,0 +1,642 @@
...
+  -- | Non-persistent stake distribution candidated (not necessarily normalized)
+  StakeDistr 'Ledger st ->
+  -- | Selected non-persistent seats with their correspoinding voting stakes
+  StakeDistr 'Voting st
+localSortition rng0 numSeats stakeDistr =
+  -- TODO: what happens if stake distribution is of size 0, or numSeats is 0?
+
+  Map.fromList $
+    sampleNonPersistentVoters $
+      stakeDistrToDecreasingStakes stakeDistr
```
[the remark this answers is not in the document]
agustinmista: Good catch, it's not. 

Previously we had a `newtype` wrapper over `StakeDistr` so it was just convenient to serialize the map using the existing `stakeDistrToIncreasingStakes`, but we can now just use `Map.toList`.

ouroboros-consensus/changelog.d/20260211_132811_agustin.mista_committee_selection.md
jasagredo: This file has to be moved to `<root>/changelog.d`

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Committee/Model.hs
```diff
@@ -0,0 +1,754 @@
...
+
+    let judgments =
+          fmap
+            ( \(voterIndex, voterStake, seatIndex, stakeAccum) ->
+                ( voterIndex
+                , voterStake
+                , seatIndex
+                , stakeAccum
+                , -- We still have remaining seats to allocate
+                  (seatIndex < numSeats)
```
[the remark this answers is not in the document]
agustinmista: I made some tweaks in the threshold function to actively avoid underflows if we ask for more seats than nodes in the stake distribution. With that change in place, this conjunction is no longer needed, and we can just call the threshold function here to get the same threshold.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS.hs
dnadales: Shouldn't these tests be placed under the "Peras" group where other Peras tests live?

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Model.hs
```diff
@@ -0,0 +1,793 @@
...
+  ( IsStake (Stake Ledger st)
+  , IsStake (Stake Weight st)
+  , Integral (NumSeats st)
+  ) =>
+  RandomGen rng =>
+  -- | Random number generator
+  rng ->
+  -- | Expected number of non-persistent seats
+  NumSeats st ->
+  -- | Non-persistent stake distribution candidated (not necessarily normalized)
```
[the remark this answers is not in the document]
agustinmista: Yes, good catch!

PR #1847, opened 2026-01-28
consensus: add MempoolRejectionDetails to TraceMempoolRejectedTx

nfrisby:
This simplifies challenges that arose when trying to add counters in `cardano-node` to count defensive mempool timeouts.

### DIFF
--- a/ouroboros-consensus/changelog.d/20260128_110959_nick.frisby_fix_defensive_mempool_tracer.md
+++ b/ouroboros-consensus/changelog.d/20260128_110959_nick.frisby_fix_defensive_mempool_tracer.md
@@ -1,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Breaking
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -550,6 +556,8 @@ data TraceEventMempool blk
       (GenTx blk)
       -- | The reason for rejecting the transaction.
       (ApplyTxErr blk)
```
jasagredo: I think there is a duplication here isn't there? If a transaction fails on the ledger we will have an error here and then `MempoolRejectedByLedger`, but if a transction timeouts we will have a `Conway.ConwayMempoolFailure "MempoolTxTooSlow (n secs)"` and then `MempoolRejectedByTimeoutSoft (n secs)` no?

Couldn't the `ApplyTxErr` be pushed inside the `MempoolRejectedByLedger` constructor?

nfrisby: (I guess my claims about LocalTxSubmission are technically irrelevant since this PR is only about the tracer. But the tracer is extremely similar to the value returned by `Mempool.addTx`, so I conflate them somewhat when thinking about it. And I think there's good reason for them to remain similar.)

PR #1859, opened 2026-02-04
Leios prototype: Get rid of EbID

ch1bo:
This PR gets rid of the (likely premature) optimization that emerged through the prototypes of october/november/december 2025 and cleans things up.

Most of the time this means using `EbHash` where we had an `EbId`. Except for the cases where we had a hack in place to derive a slot from an `EbId` -> those result in usages of `LeiosPoint` now.

I tested this manually by running this branch in the `ouroboros-leios` repo's `demo/2025-11` (see also https://github.com/input-output-hk/ouroboros-leios/pull/753)

~~FIXME: This seems to be only fetching blocks/txs correctly, but not offering it (when integrated into the cardano-node). Need to look into this further..~~ Nevermind, was behaving just as expected with this commit https://github.com/input-output-hk/ouroboros-leios/commit/c43eba8c456060fc5605758d72cb0970bab58fd3. So the refactor appears to be sound (from only the very high-level view of the end-to-end scenario).

### DIFF
--- a/ouroboros-consensus/app/leiosdemo202510.hs
+++ b/ouroboros-consensus/app/leiosdemo202510.hs
@@ -44,6 +44,7 @@ import qualified Data.Vector as V
 import Data.Word (Word16, Word32, Word64, Word8)
 import qualified Database.SQLite3.Direct as DB
 import GHC.Generics (Generic)
+import qualified LeiosDemoDb
 import qualified Numeric
 import System.Directory (doesFileExist)
 import System.Environment (getArgs)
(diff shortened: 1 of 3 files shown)


ch1bo:
The 2025-11 demo was flaky and it turned out that the refactor did not change that (not yet). The root cause was database errors due to "database locked" and this was not properly traced/visible from grafana. See https://github.com/IntersectMBO/ouroboros-consensus/commit/24b2a67782fd39ee05dea12a4d866228d86765fb

PR #1864, opened 2026-02-06
[Peras 17] Store round number of latest Peras certificate included on chain

agustinmista:
The main goal of this PR is to extend the ChainDB API and implementation with a new method that retrieves the latest (as in _highest_) round number of a Peras certificate included in a block in our preferred chain:

 ```haskell
getLatestPerasCertOnChainRound :: STM m (Maybe PerasRoundNo)
```

This is needed by an upcoming voting thread to evaluate the Peras voting rules at the beginning of each Peras round (modulo being also part of the voting committee). 

Worth mentioning, the reason this method only retrieves round numbers and not entire certificates is twofold:
* after certificates are validated, we no longer care about their contents other than their corresponding round numbers (which is ultimately only needed by one of the (sub-) voting rules), and
* round numbers are non-parametric (as opposed to certificates being parameterized by `blk`), substantially simplifying the implementation of this PR.
Nonetheless, it should be straightforward to extend this changes if in the future we find out that we need to extract more information from the latest certificate on chain.

The changes are divided into a couple of (hopefully digestible) atomic commits and I would suggest reviewing them in order:

- 1) Extending `BlockSupportsPeras` with a primitive to (possibly) extract certificates from blocks[^1].

- 2) Extend the Shelley ledger state with a field to cache the latest round number seen in a block. This value is updated whenever a new block is applied to the ledger.

- 3) Introducing the `LedgerSupportsPeras` type class to give access to this field from an arbitrary ledger state (notably needed by the hard fork ledger state), and propagating the constraint everywhere.

- 4) Extending the ChainDB API with the aforementioned `getLatestPerasCertOnChainRound` method, and providing an implementation that returns the latest round number between the ones found in the ledger state of both the volatile and immutable tips[^2].

- 5) (and vi.) Extending the `TestBlock` used by the storage tests to replicate the changes made by (ii) and (iii) to the Shelley ledger state.

- 7) Extending the ChainDB QSM model and tests to exercise the new method in the ChainDB API[^3].

Closes https://github.com/tweag/cardano-peras/issues/198

[^1]: this currently has a degenerate instance that always returns `Nothing`, but this will notably will not be the case after integrating the [changes](https://github.com/IntersectMBO/cardano-ledger/pull/5439) made to the Ledger layer to (possibly) include Peras certificates in Dijkstra block bodies.

[^2]: this might seem overly conservative at first but, as far as we can tell, there is no guarantee that a newer certificate (as in, one included in a younger block) will necessarily point to a higher Peras round number. This means that, the latest round number could be in a certificate included in an immutable block, even if there are newer certificates included in blocks in the volatile suffix. So, we always end up reaching for the maximum of all the available options.

[^3]: this notably required disabling the generation of the new command (as well as the inclusion of certificates in test blocks for good measure) whenever a test runs with `LoEEnabled`. Supporting both simultaneously is outside of the scope of this PR (and of pre-alpha Peras as whole).

### DIFF
--- a/changelog.d/20260212_154504_agustin.mista_store_latest_cert_on_chain.md
+++ b/changelog.d/20260212_154504_agustin.mista_store_latest_cert_on_chain.md
@@ -1,0 +1,35 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Add `getPerasCertInBlock` method to `BlockSupportsPeras` typeclass.
+- Extend Shelley ledger state to keep track of the latest Peras certificate round number seen on chain.
+- Add `LedgerSupportsPeras` typeclass.
+- Add instance `LedgerSupportsPeras` for the extended Shelley ledger state.
+- Add instance `LedgerSupportsPeras` for `HardForkBlock`.
+- Add trivial empty `LedgerSupportsPeras` for other blocks that do not support Peras.
+- Add `getLatestPerasCertOnChainRound` method to `ChainDB` API.
+
+### Breaking
+
+- Add `LedgerSupportsPeras` constraint to `SingleEraBlock`.
+- Update serialization of Shelley ledger state to encode/decode the new field.
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->

### CODE REVIEW

File: ouroboros-consensus-cardano/changelog.d/20260212_154510_agustin.mista_store_latest_cert_on_chain.md [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
```
jasagredo: Remove

ouroboros-consensus-diffusion/changelog.d/20260212_154514_agustin.mista_store_latest_cert_on_chain.md
```diff
@@ -0,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Patch
+
+- Add trivial `LedgerSupportsPeras` for `BlockA` and `BlockB`.
```
jasagredo: This is part of the testing library and does not need to be reflected in the changelogs.

ouroboros-consensus/changelog.d/20260212_154504_agustin.mista_store_latest_cert_on_chain.md
```diff
@@ -0,0 +1,30 @@
...
+- A bullet item for the Patch category.
+
+-->
+
+### Non-Breaking
+
+- Add `getPerasCertInBlock` method to `BlockSupportsPeras` typeclass.
+- Add `LedgerSupportsPeras` typeclass.
+- Add instance `LedgerSupportsPeras` for `HardForkBlock`.
+- Extend block body and ledger state for `TestBlock` to support containing and extracting Peras certificate round numbers.
```
jasagredo: This belongs to tests and does not need to be in the changelog.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Peras.hs
```diff
@@ -0,0 +1,19 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.HardFork.Combinator.Peras () where
```
jasagredo: Same question, do we really need to put this in a separate module?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -825,6 +864,13 @@ decodeShelleyTransition = do
   shelleyAfterVoting <- CBOR.decodeWord32
   return ShelleyTransitionInfo{shelleyAfterVoting}
 
+encodeShelleyLatestPerasCertRound :: PerasRoundNo -> Encoding
+encodeShelleyLatestPerasCertRound roundNo =
+  CBOR.encodeWord64 (unPerasRoundNo roundNo)
+
+decodeShelleyLatestPerasCertRound :: Decoder s PerasRoundNo
+decodeShelleyLatestPerasCertRound = PerasRoundNo <$> CBOR.decodeWord64
```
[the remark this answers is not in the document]
agustinmista: Yep. I will move this as `FromCBOR`/`ToCBOR` instances next to where `PerasRoundNo` is defined.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -834,14 +880,22 @@ encodeShelleyLedgerState
...
         , toCBOR shelleyLedgerState
         , encodeShelleyTransition shelleyLedgerTransition
         ]
+          <> [ encodeShelleyLatestPerasCertRound roundNo
+             | SJust roundNo <- [shelleyLedgerLatestPerasCertRound]
+             ]
+   where
+    len
+      | SNothing <- shelleyLedgerLatestPerasCertRound = 3
+      | otherwise = 4
```
bladyjoker: Aha I see the problem, this is way too subtle for this level. Let's try and make this super obvious and mechanical.

Can I suggest we encode this as 

```
encodeVersion serialisationFormatVersion2 $
      mconcat $
        [ encodeWithOrigin encodeShelleyTip shelleyLedgerTip,
          , toCBOR shelleyLedgerState
          , encodeShelleyTransition shelleyLedgerTransition
          , toCBOR shelleyLedgerLatestPerasCertRound -- simply a Maybe
         ]
```

ouroboros-consensus/src/unstable-consensus-testlib/Test/Ouroboros/Storage/TestBlock.hs
```diff
@@ -613,7 +621,27 @@ instance ApplyBlock (LedgerState TestBlock) TestBlock where
...
+            TestLedger
+              (Chain.blockPoint tb)
+              (BlockHash (blockHash tb))
+              ( let
+                  -- NOTE: this bypasses the degenerate global implementation of
+                  -- 'BlockSupportsPeras.getPerasCertInBlock' for 'TestBlock',
+                  -- which currently always returns 'Nothing'.
+                  --
+                  -- TODO: refactor this to use 'getPerasCertInBlock' after the
+                  -- HFC plumbing for 'BlockSupportsPeras' is in place.
```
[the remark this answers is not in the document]
agustinmista: I see how this can be confusing, by "HFC plumbing" I'm referring to the fact that we still need to split the existing (degenerate):

```haskell
instance StandartHash blk => BlockSupportsPeras blk
```

Into separate instances for different concrete blocks (including testing ones). This TODO should be read as:

> Replace this with `getPerasCertInBlock` once `TestBlock` has a proper `BlockSupportsPeras` implementation.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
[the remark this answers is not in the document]
agustinmista: An initial CDDL diff for the fixed-length encoding can be found in fc05a1c79.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Peras.hs
agustinmista: Fixed in d761e4d1b

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
```diff
@@ -0,0 +1,18 @@
...
+module Ouroboros.Consensus.Ledger.SupportsPeras
+  ( LedgerSupportsPeras (..)
+  )
+where
+
+import Ouroboros.Consensus.Block.SupportsPeras (PerasRoundNo)
+import Ouroboros.Consensus.Ledger.Abstract (LedgerState)
+
+-- | Extract Peras information stored in the ledger state
+class LedgerSupportsPeras blk where
```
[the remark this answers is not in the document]
agustinmista: Given that it seems that Leios and Peras will share the committee selection implementation (with possibly a different crypto scheme), at least that one interface will likely become a superclass of `BlockSupportsPeras`.

I personally prefer smaller compositional interfaces because it helps me better understand which methods are connected directly to each capability, but I don't have a very strong opinion either way.

dnadales (approved):

PR #1869, opened 2026-02-10
Leios prototype: trace db exception

ch1bo:
The previous `die` call was not showing up properly in the observability stack. This now gives at least the chance to be visible as a big fat error :x:

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoException.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoException.hs
@@ -1,11 +1,27 @@
-module LeiosDemoException (LeiosDbException (..)) where
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module LeiosDemoException where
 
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -125,16 +132,17 @@ msgLeiosBlockRequest ::
...
-          loop (i + 1) rest
-    loop 0 items
+    traceException tracer TraceLeiosPeerDbException $ do
+      -- get the EB items using new db
+      items <- leiosDbLookupEbBody db pointEbHash
+      let loop !i [] = pure i
+          loop !i ((txHash, txBytesSize) : rest) = do
+            MV.write buf i (txHash, txBytesSize)
+            loop (i + 1) rest
+      loop 0 items
```
bladyjoker: Smtn like this?
```haskell
forM_ (zip [0..] items) $ \(i, (txHash, txBytesSize)) ->
  MV.write buf i (txHash, txBytesSize)
```

PR #1872, opened 2026-02-10
Fix `chunksBetween`, which was always either empty or singleton

vreuter:
This change fixes a bug in the `chunksBetween` function in the `Ouroboros.Consensus.Storage.ImmutableDB.Chunks.Internal`, whereby the result was always either empty (if the bounds of the range were unequal) or a single-item list (if the bounds were equal).


dnadales:
Thank you @vreuter!

PR #1877, opened 2026-02-12
WIP: Use ContT instead of ResourceRegistry for handles

jasagredo:
# Description

this PR might resolve.

Also note that:

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/ContT.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/ContT.hs
@@ -8,6 +8,7 @@ module Ouroboros.Consensus.Util.ContT
   , lift
   , liftContT
   , withReadAccessCont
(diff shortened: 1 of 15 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/ContT.hs
```diff
@@ -0,0 +1,102 @@
...
+  (st -> ContT a m a) ->
+  ContT r m a
+withReadAccessCont lock action = lift $
+  withReadAccess lock $
+    \st -> runContT (action st) return
+
+-- | Allocates a resource and returns:
+-- 1. The Resource
+-- 2. A Safe (Idempotent) Cleanup Action (to save for later)
+-- 3. A "Disarm" Action (STM) to cancel the automatic local cleanup
```
jasagredo: Should the cleanup be provided to the caller only when disarm is run? i.e. should the type be:
```
ContT r m (res, STM m (m ()))
```

PR #1879, opened 2026-02-13
[CTC 1] Generalize point schedule tests to parametric `blk`

ninioArtillero:
Node-vs-Environment tests (a.k.a. point schedule tests, represented by `GenesisTest`) are currently hardcoded to use `TestBlock`. As part of the [approved Conformance Testing of Consensus (CTC) proposal](https://github.com/tweag/cardano-conformance-testing-of-consensus), we are tasked with exposing these tests to the wider ecosystem and therefore must support real Cardano blocks.

This PR generalizes the relevant types to make the tests block-agnostic. To accomplish this:

1. The existing code is parameterized over an arbitrary block type, subject to the necessary constraints.
2. New code implementing `deforestBlockTree` is introduced to generalize property test computations that previously depended on the `TestHash` implementation. This function constructs a map enabling *prefix* lookups by `HeaderHash`, where a prefix is the `AnchoredFragment` from genesis to the block corresponding to the given hash at its tip.
3. A new type class, `HasPointScheduleTestParams`, is introduced to construct the `ProtocolInfo` and other pieces of state required for a test run.
4. A new type class, `IssueTestBlock`, defining the interface for producing blocks in tests, is introduced.

Care has been taken to document the previously implicit `TestHash`-related logic (based on the use of ad hoc fork numbers) and to include comprehensive property tests for the *deforestation* code under `ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree/Tests.hs`.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Setup.hs
@@ -2,15 +2,19 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Test.Consensus.Genesis.Setup
   ( module Test.Consensus.Genesis.Setup.GenChains
+  , castHeaderHash
   , forAllGenesisTest
+  , honestImmutableTip
   , runGenesisTest
   , runGenesisTest'
+  , selectedHonestChain
   ) where
 
 import Control.Exception (throw)
@@ -20,7 +24,12 @@ import Control.Monad.Class.MonadAsync
 import Control.Monad.IOSim (IOSim, runSimStrictShutdown)
 import Control.Tracer (debugTracer, traceWith)
 import Data.Maybe (mapMaybe)
-import Ouroboros.Consensus.Block.Abstract (ConvertRawHash, Header)
+import Ouroboros.Consensus.Block.Abstract
+  ( ChainHash (..)
+  , ConvertRawHash
+  , GetHeader
+  , Header
+  )
 import Ouroboros.Consensus.Block.SupportsDiffusionPipelining
   ( BlockSupportsDiffusionPipelining
   )
@@ -40,10 +49,12 @@ import Ouroboros.Consensus.Storage.LedgerDB.API
   )
 import Ouroboros.Consensus.Util.Condense
 import Ouroboros.Consensus.Util.IOLike (Exception, fromException)
+import qualified Ouroboros.Network.AnchoredFragment as AF
 import Ouroboros.Network.Driver.Limits
   ( ProtocolLimitFailure (ExceededTimeLimit)
   )
 import Ouroboros.Network.Util.ShowProxy
+import Test.Consensus.BlockTree (onTrunk)
 import Test.Consensus.Genesis.Setup.Classifiers
   ( Classifiers (..)
   , ResultClassifiers (..)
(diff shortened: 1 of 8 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
```diff
@@ -120,8 +126,15 @@ rollbackSchedule n blockTree =
...
+--
+-- PRECONDITION: The hash must correspond to a block in the 'BlockTree'
+hashOnTrunk :: (AF.HasHeader blk, Eq blk) => BlockTree blk -> ChainHash (Header blk) -> Bool
+hashOnTrunk _ GenesisHash = True
+hashOnTrunk bt (BlockHash hash) = do
+  case M.lookup hash (deforestBlockTree bt) of
+    -- Returning 'False' here would be correct in a general setting,
+    -- but as we expect inputs resulting from the simulation (which necessarily
+    -- are on the block tree) we short circuit here instead.
+    Nothing -> error "hashOnTrunk: the selected chain head must be on the BlockTree"
```
isovector: I think we should probably just return `False`; nothing says the hash needs to be on this particular block tree!

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Trace.hs
```diff
@@ -1,9 +1,11 @@
 {-# LANGUAGE ExistentialQuantification #-}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE NumericUnderscores #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE FlexibleContexts          #-}
```
isovector: bad formatting?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/Tests/Rollback.hs
ninioArtillero: The only thing that says it are the call sites to this function. There, the hash is picked from a `BlockTree` block. As this function is not even exported, I figured it might be better to set an /alarm/ here in case _something_ at those call sites went terribly wrong.

To be honest I had a hard time deciding, as going with `False` seemed the best thing just by looking at the types. What I mean is that I have no strong preference for this error. Thinking about it now, it seems that such bad behavior would be found elsewhere. So I can agree going with the `False`.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree/Tests.hs
```diff
@@ -0,0 +1,176 @@
...
+genTestBlockTree = fmap gtBlockTree . genChains
+
+genTestAnchoredFragment :: QC.Gen (AF.AnchoredFragment TestBlock)
+genTestAnchoredFragment = fmap btTrunk $ genTestBlockTree (pure 0)
+
+tests :: TestTree
+tests =
+  let branchFactor = pure 4
+   in adjustQuickCheckTests (* 100) $
+        adjustOption (\(QuickCheckMaxSize n) -> QuickCheckMaxSize (n `div` 10)) $
```
jasagredo: Why these adjustments?

ouroboros-consensus-diffusion/changelog.d/20260210_112021_isovector_testblocks.md
```diff
@@ -22,3 +23,7 @@ For top level release notes, leave all the headers commented out.
 - A bullet item for the Breaking category.
 
 -->
+
+### Non-Breaking
+
+- Generalized all conformance testing machinery to operate over arbitrary block types
```
jasagredo: No need to push a changelog fragment as this is test code for now

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree.hs
```diff
@@ -229,3 +259,79 @@ prettyBlockTree blockTree =
...
+--
+-- nonemptyPrefixesOf builds the list of prefixes of an `AF.AnchoredFragment` with at
+-- least one non-anchor entry. The name is a little confusing because the way we
+-- usually think of cons-lists these would be suffixes:
+--
+--              A :> B1     A :> B1 :> B2     A :> B1 :> B2 :> B3
+--        A :> B1 :> B2 :> B3 :> B4     A :> B1 :> B2 :> B3 :> B4 :> B5
+--
+-- However this is consistent with `Ouroboros.Network.AnchoredSeq.isPrefixOf`.
+nonemptyPrefixesOf ::
```
jasagredo: This could be a haddock, no? (i.e. change from `--` to `-- |`)
```suggestion
-- | An `AF.AnchoredFragment` is a list where the last element (the anchor) is a ghost.
-- Here they represent the partial ancestry of a block, where the anchor is either
-- `Genesis` (start of the chain, not itself an actual block) or the hash of a block.
-- Say we have blocks B1 through B5 (each succeeded by the next) and anchor A. You
-- can think of the chain as growing **from left to right** like this:
--
-- >                    A :> B1 :> B2 :> B3 :> B4 :> B5
--
-- nonemptyPrefixesOf builds the list of prefixes of an `AF.AnchoredFragment` with at
-- least one non-anchor entry. The name is a little confusing because the way we
-- usually think of cons-lists these would be suffixes:
--
-- >            A :> B1     A :> B1 :> B2     A :> B1 :> B2 :> B3
-- >      A :> B1 :> B2 :> B3 :> B4     A :> B1 :> B2 :> B3 :> B4 :> B5
--
-- However this is consistent with `Ouroboros.Network.AnchoredSeq.isPrefixOf`.
nonemptyPrefixesOf ::
```

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/BlockTree/Tests.hs
nbloomf: There isn't a good reason; at one point during development there was a performance regression and these parameters were able to generate interesting test cases without being too slow. The regression was fixed so we can use the default parameters instead.

ninioArtillero: Confirmed tests run fine with default parameters. Reverted 🙂

PR #1888, opened 2026-02-18
Leios prototype: Store txs of forged EBs and diffuse them

ch1bo:
Based on #1874, this writes the transaction closure of a forged (and to be announced) EB into the prototype leios db.

Next, the store will notify downstream nodes of the newly acquired tx closure by responding a `LeiosBlockTxsOffer` to any pending notification requests.

This will have the fetching logic of downstream nodes pull the tx closure, store the bytes retrieved (which are now real transactions!) and serves them further downstream.

I needed to use a cbor-in-cbor encoding to skip decoding into proper `Tx era` just for serving them downstream. Any other ideas about that bit?


bladyjoker (approved):
Ship it!

PR #1891, opened 2026-02-20
Remove usages of ResourceRegistry on Forkers

jasagredo:
# Description

Forkers have been a source of headaches for a very long time already, and all due to the limitation of the Resource registry that enforces registries to only be used from known threads. This has lead to a situation where we had hierarchies of registries just to satisfy that rule and even we had to extend the resource-registry API to allow transferring between registries which is unsafe.

The general pattern followed now is:
1. Bracket the acquisition of the resource
2. Can't bracket the resource usage 
   1. `allocateTemp` the resource such that there is a temporary registry holding the value
   2. Run the action with `runWithTempRegistry`
   3. Before exiting the `runWithTempRegistry` scope, the resource is stored somewhere else that is kept track by an upper registry or bracket.
3. The only violation of this pattern is in the Mempool and LocalStateQuery where we do use a resource registry to keep the forker, which is read-only.

Sadly, the change is so fundamental that there is no meaningful way of splitting the changes into reviewable commits that compile on their own.

Proposed review order (recommended to disable whitespace-diff in Github diff view):

1. Ouroboros.Consensus.Storage.ChainDB.API
2. Ouroboros.Consensus.MiniProtocol.LocalStateQuery.Server + Ouroboros.Consensus.Network.NodeToClient
3. Ouroboros.Consensus.Mempool
4. Ouroboros.Consensus.Storage.ChainDB.ChainSel
5. Ouroboros.Consensus.Storage.ChainDB.Impl.Types + Ouroboros.Consensus.Storage.ChainDB.Impl
6. Ouroboros.Consensus.Storage.LedgerDB.API (both the API changes and the initialization)
7. Ouroboros.Consensus.Storage.LedgerDB
8. Ouroboros.Consensus.Storage.LedgerDB.Forker
9. Ouroboros.Consensus.NodeKernel
10. Ouroboros.Consensus.Storage.LedgerDB.V2.LedgerSeq
11. Ouroboros.Consensus.Storage.LedgerDB.V2.Forker
12. Ouroboros.Consensus.Storage.LedgerDB.V2.Backend
13. Ouroboros.Consensus.Storage.LedgerDB.V2.InMemory
14. Ouroboros.Consensus.Storage.LedgerDB.V2.LSM
15. Ouroboros.Consensus.Storage.LedgerDB.V2    
16. Ouroboros.Consensus.Storage.LedgerDB.V1.Forker
17. Ouroboros.Consensus.Storage.LedgerDB.V1.Snapshots
18. Ouroboros.Consensus.Storage.LedgerDB.V1

The rest of the changes are just adapting tests or adapting the cardano-tools to use the new access pattern.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -521,8 +520,8 @@ type ResolveBlock m blk = RealPoint blk -> m blk
 -------------------------------------------------------------------------------}
 
 -- | When validating a sequence of blocks, these are the possible outcomes.
-data ValidateResult m l blk
-  = ValidateSuccessful (Forker m l)
+data ValidateResult l blk
+  = ValidateSuccessful
```
jasagredo: For reviewers: by running this code as a continuation, we don't need to pass back the forker, which lead to issues for when to dealloc it and demanded resource registries.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2/InMemory.hs
```diff
@@ -79,185 +75,99 @@ import Prelude hiding (read)
...
+  , StandardHash l
+  , GetTip l
   ) =>
   Tracer m LedgerDBV2Trace ->
+  -- | FileSystem in order to take snapshots
   SomeHasFS m ->
-  (ResourceKey m, StrictTVar m (LedgerTablesHandleState l)) ->
+  -- | The tables
+  LedgerTables l ValuesMK ->
   m (LedgerTablesHandle m l)
```
jasagredo: For reviewers: Note this implementation is monadic ONLY for tracing. All the methods are pure.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -189,46 +195,48 @@ mkInternals h snapManager =
...
+              let st' =
+                    tickThenReapply
+                      (ledgerDbCfgComputeLedgerEvents (ldbCfg env))
+                      (ledgerDbCfg $ ldbCfg env)
+                      blk
+                      (st `withLedgerTables` tables)
+              forkerPush frk st' >> Monad.join (atomically (forkerCommit frk))
+              pruneLedgerSeq env
+          )
+    , closeLedgerDB = implCloseDB h
```
jasagredo: For reviewers: this might look intimidating, but consider this is only for tests and db-analyser.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -412,51 +422,25 @@ data LedgerDBEnv m l blk = LedgerDBEnv
...
-  -- be released by each one of the client's registries. This means
-  -- that for example ChainSelection will, upon closing its registry,
-  -- release its forker and any resources associated.
-  --
-  -- Upon closing the LedgerDB we will overwrite this variable such
-  -- that existing forkers can only be closed, as closing doesn't
-  -- involve accessing this map (other than possibly removing the
-  -- forker from it if the map still exists).
-  --
-  -- As the LedgerDB should outlive any clients, this is fine.
```
jasagredo: For reviewers: The LedgerDB used to track the handles but this is unnecessary, so this map and all the associated logic is gone

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -81,44 +82,55 @@ mkInitDb ::
...
+                ( \case
+                    -- captured above
+                    Left failure -> throwIO $ SnapshotExc failure
+                    Right (st, p) ->
+                      let lseq = LedgerSeq (AS.Empty st)
+                       in pure ((lseq, p), lseq)
+                )
+            )
+          $ runExceptT
+            ( openStateRefFromSnapshot
```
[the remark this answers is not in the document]
jasagredo: As this is done in the alloc step of an `allocateTemp`, could we not use the internal `WithTempRegistry`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V1/Forker.hs
```diff
@@ -110,25 +110,43 @@ closeForkerEnv ForkerEnv{foeBackingStoreValueHandle} = do
...
+        Right bsvh -> do
+          putMVar foeBackingStoreValueHandle (Right bsvh)
+          pure bsvh
+        Left (l, bs) -> do
+          bsvh <-
+            bracketOnError
+              (bsValueHandle bs)
+              bsvhClose
+              ( \bsvh -> do
+                  atomically $ writeTVar tv (Just bsvh)
```
jasagredo:
```suggestion
                  atomically $ putMVar tv (Right bsvh)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -281,13 +284,37 @@ getPastLedger ::
...
+    (\_ -> unsafeGetReadOnlyForkerAtPoint cdb tgt)
+    (either (const $ pure ()) LedgerDB.roforkerClose)
+
+unsafeGetReadOnlyForkerAtPoint ::
+  IOLike m =>
+  ChainDbEnv m blk ->
   Target (Point blk) ->
   m (Either LedgerDB.GetForkerError (LedgerDB.ReadOnlyForker' m blk))
-getReadOnlyForkerAtPoint CDB{..} = LedgerDB.getReadOnlyForker cdbLedgerDB
+unsafeGetReadOnlyForkerAtPoint CDB{..} = LedgerDB.unsafeGetReadOnlyForker cdbLedgerDB
```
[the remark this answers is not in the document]
nfrisby: I'm just a few GitHub Comments in so far, but I'm immediately wondering if most of these Comments should be preserved as Haddock comments so they outlive this PR.

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
```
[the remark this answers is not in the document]
nfrisby: Or just `flip (,)` in this particular example

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -143,9 +143,9 @@ mkHandlers NodeKernelArgs{cfg, tracers} NodeKernel{getChainDB, getMempool} =
         localTxSubmissionServer
           (Node.localTxSubmissionServerTracer tracers)
           getMempool
-    , hStateQueryServer =
-        localStateQueryServer (ExtLedgerCfg cfg)
-          . ChainDB.getReadOnlyForkerAtPoint getChainDB
+    , hStateQueryServer = \reg ->
```
nfrisby: Please fully eta-expand this.

I suspect that either no eta or full-eta are significantly less surprising than partial eta-expansion.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -266,40 +264,36 @@ initMempoolEnv ::
...
+  MempoolLedgerDBView st meFrk <- atomically $ getCurrentLedgerState ledgerInterface
+  (rk, eFrk) <- meFrk
+  case eFrk of
+    -- This should happen very rarely, if between getting the state and getting
+    -- the forker, the ledgerdb has changed. We just retry here.
+    Left{} -> do
+      _ <- release rk
+      initMempoolEnv ledgerInterface cfg capacityOverride mbTimeoutConfig tracer
+    Right frk -> do
+      frkMVar <- newMVar (frk, rk)
```
nfrisby: unexpected swap of the argument order that was established so far in the API.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Init.hs
```diff
@@ -71,14 +73,14 @@ forkSyncStateOnTipPointChange menv =
...
+  action :: Point blk -> m ()
   action _a =
     void $ implSyncWithLedger menv
 
   -- Using the tip ('Point') allows for quicker equality checks
-  getCurrentTip :: STM m (MempoolLedgerDBView m blk)
+  getCurrentTip :: STM m (Point blk)
   getCurrentTip =
-    getCurrentLedgerState (mpEnvLedger menv) (mpEnvRegistry menv)
+    ledgerTipPoint . mldViewState <$> getCurrentLedgerState (mpEnvLedger menv)
```
nfrisby: First thought: now that it's very clear we're only using it for the tip `Point`, hopefully don't even need to be opening a Forker just to throw it away.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -23,7 +23,9 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.Query
   , getPastLedger
   , getPerasWeightSnapshot
   , getPerasCertSnapshot
-  , getReadOnlyForkerAtPoint
+  , unsafeGetReadOnlyForkerAtPoint
```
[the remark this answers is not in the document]
nfrisby: Now that I'm seeing the repeat pattern in test code, maybe we should instead be exporting a general version of `allocInRegistryReadOnlyForkerAtPoint` that's parameterized by how to open the forker.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -920,33 +906,32 @@ switchTo CDB{..} weights triggerPt vChainDiff reason = do
...
           followerHandles <- Map.elems <$> readTVar cdbFollowers
           -- The suffix of @curChain@ that we are going to orphan by
           -- adopting @chainDiff@.
           let oldSuffix = AF.anchorNewest (getRollback chainDiff) curChain
           forM_ followerHandles $ \hdl -> fhSwitchFork hdl oldSuffix
 
-        return (curChain, newChain, events, prevTentativeHeader, newLedger)
+        return (curChain, newChain, events, prevTentativeHeader, newLedger, closeOrphanedStates)
+
+  closeOrphanedStates
```
nfrisby: It'll be helpful to have a comment here explaining why it's OK for an exception to prevent this from being called. For example: ChainSel doesn't catch any exceptions, so the node is coming down anyway. That'll close the UTxO/Ledger HD backend's session, which means it doesn't matter if any Forkers are unclosed.

Or if it's not, then we need to fix it up. (Maybe delay the `forkerCommit` into its own `atomically` so that exceptions can be masked during it?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -240,24 +241,25 @@ data LedgerDB m l blk = LedgerDB
       l ~ ExtLedgerState blk =>
       STM m (HeaderStateHistory blk)
   -- ^ Get the header state history for all ledger states in the LedgerDB.
-  , getForkerAtTarget ::
-      ResourceRegistry m ->
+  , unsafeGetForkerAtTarget ::
```
nfrisby: Maybe we should say `leaky*` instead of `unsafe*`. Just so it's more obvious what the risk is.

.... or is `open*` enough, considering `openFile` is the fundamental example?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -531,83 +522,102 @@ initialize
...
-          initDb
-          (Point Origin)
-          dbIface
+
+      tvldb <-
+        allocateTemp
+          (newTVarIO =<< initFromGenesis)
+          (\tv -> readTVarIO tv >>= initCloseDB >> pure True)
+          (\_ _ -> True)
+
```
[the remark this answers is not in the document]
nfrisby: Let's float this pattern out as a combinator, if we can't get rid of the temporary registry altogether. Even just local in this module is fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -629,33 +638,37 @@ replayStartingWith ::
...
+    Right (_, v) -> pure v
  where
   push ::
     blk ->
     (db, Word64) ->
     m (db, Word64)
-  push blk (!db, !replayed) = do
-    !db' <- initReapplyBlock cfg blk db
+  push blk !(!db, !replayed) = do
+    db' <- runWithTempRegistry $ do
```
nfrisby: Calling this for every block at least seems excessive. But I don't have a strong argument either way.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB.hs
```diff
@@ -129,26 +131,32 @@ openDBInternal ::
...
-      lgrConfig
-      stream
-      replayGoal
-      initDb
-      snapManager
-      lgrStartSnapshot
-  (ledgerDb, internal) <- mkLedgerDb initDb db
+  lift $ createDirectoryIfMissing fs True (mkFsPath [])
+  (db, replayCounter) <-
+    runInnerWithTempRegistry
```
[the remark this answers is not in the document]
nfrisby: Same question as in the other module: why does it matter whether we close the UTxO HD backend handles if there's an exception during initialization? Am I right it doesn't mater for InMem and LSM? What about LMDB?

PR #1893, opened 2026-02-24
[Peras 18] Implement generic weighted Fait-Accompli with Local Sortition

agustinmista:
This PR implements the initial version of a generic weighted Fait-Accompli with Local Sortition (wFA^LS) committee selection scheme. To allow for both Peras and Leios to reuse the same implementation, the code is abstracted in the following ways:

* `ElectionId` type family: central to committee selection, it allows the client to provide their own different election identifier. Peras will use round numbers, whereas Leios will use something isomorphic to a slot number.
* `CryptoSupportsVoteSignature`: crypto interface needed to sign/verify vote signatures using a public/private key pair.
* `CryptoSupportsVRF`: crypto interface needed to compute/validate VRF outputs using a public/private key pair. This is needed by the Local Sortition fallback scheme.
* `CryptoSupportsWFALS`: Wrapper on top of `CryptoSupportsVoteSignature` and `CryptoSupportsVRF` that allows us to derive both their key pairs from a single key pair. This could be used to have potentially different crypto schemes for the vote signing and VRF evaluation routines derived from the same "composite" key pair.
* `VoteSupportsWFALS`: interface needed to extract the bits relevant to committee selection and signature verification from a client vote (i.e., a concrete Peras or Leios vote).

In addition, this PR also provides an initial instantiation of all these type classes for Peras using BLS crypto (the best approximation for the final one we have for now).

## Other notes

* While this PR only provides wFA^LS, the design should be modular enough to also allow one to reuse some of the interfaces (e.g., `ElectionId`, and `CryptoSupportsVoteSigning`) for other committee selection schemes (e.g. wFA^IID).
* Conformance tests against the Rust implementation will be added after [Peras 16](https://github.com/IntersectMBO/ouroboros-consensus/pull/1839) lands into `main`.
* Because epoch boundaries are a bit special, `O.C.Committee.AcrossEpochs` defines a small wrapper to keep the committee selections from both the current and previous epochs at hand. This is needed to validate votes cast at the end of the previous epoch arriving at the beginning of the current one.
* Because any instance of a weighted-Fait-Accompli-based committee selection for a given epoch would need to aggregate the input stake distribution in the same way, we have extracted this into the `ExtCumulativeStakeDistr` data type. This would allow both Peras and Leios to share part of the heavy work needed at the beginning of every new epoch.
* This PR *does not* include any other potential optimization beyond the one described above.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
@@ -31,7 +31,7 @@
 module Ouroboros.Consensus.Committee.WFALS
   ( -- * Voting committee membership
     MembershipType (..)
-  , MemberhipProof
+  , MembershipProof
   , CommitteeMember (..)
 
     -- * Committee votes
@@ -96,28 +96,28 @@ data MembershipType
   deriving Show
 
 -- | Proof of committee membership for a given voter
-type MemberhipProof :: Type -> MembershipType -> Type
-data MemberhipProof crypto membership where
-  PersistentMemberProof ::
+type MembershipProof :: Type -> MembershipType -> Type
+data MembershipProof crypto membership where
+  PersistentMembershipProof ::
     SeatIndex ->
     VoteSignature crypto ->
-    MemberhipProof crypto Persistent
-  NonPersistentMemberProof ::
+    MembershipProof crypto Persistent
+  NonPersistentMembershipProof ::
     LocalSortitionNumSeats ->
     VRFOutput crypto ->
     VoteSignature crypto ->
-    MemberhipProof crypto NonPersistent
+    MembershipProof crypto NonPersistent
 
 -- | Committee members (i.e., no longer candidates)
 type CommitteeMember :: Type -> Type
 data CommitteeMember crypto
   = -- | A persistent member of the voting committee
     PersistentCommitteeMember
-      (MemberhipProof crypto Persistent)
+      (MembershipProof crypto Persistent)
       LedgerStake
   | -- | A (realized) non-persistent member of the voting committee
     NonPersistentCommitteeMember
-      (MemberhipProof crypto NonPersistent)
+      (MembershipProof crypto NonPersistent)
       LedgerStake
 
 -- * Committee votes
@@ -266,7 +266,7 @@ mkCommitteeSelection nonce totalSeats stakeDistr = do
 data CommitteeSelectionError
   = -- | A voter ID is missing from the committee selection
     MissingPoolId PoolId
-  | -- | A voter claims to be a persistent member of the committe, but it's not
+  | -- | A voter claims to be a persistent member of the committee, but it's not
     NotAPersistentMember SeatIndex
   | -- | A voter claims to be a non-persistent member of the committe, but it's not
     NotANonPersistentMember PoolId
@@ -297,14 +297,14 @@ committeeMemberWeight ::
 committeeMemberWeight selection = \case
   -- Persistent members have their voting power equal to their stake
   PersistentCommitteeMember
-    (PersistentMemberProof _seatIndex _sig)
+    (PersistentMembershipProof _seatIndex _sig)
     (LedgerStake stake) ->
       VoteWeight stake
   -- Non-persistent members have their voting power proportional to their
   -- number of seats granted by local sortition and their stake (normalized
   -- by the total non-persistent stake)
   NonPersistentCommitteeMember
-    (NonPersistentMemberProof numSeats _sig _vrfOutput)
+    (NonPersistentMembershipProof numSeats _sig _vrfOutput)
     (LedgerStake stake) ->
       VoteWeight $
         fromIntegral (unLocalSortitionNumSeats numSeats)
@@ -411,7 +411,7 @@ checkShouldVote ourId ourPrivateKey electionId message selection
             pure $
               Just $
                 PersistentCommitteeMember
-                  (PersistentMemberProof seatIndex sig)
+                  (PersistentMembershipProof seatIndex sig)
                   ourStake
           False -> do
             let vrfContext =
@@ -430,7 +430,7 @@ checkShouldVote ourId ourPrivateKey electionId message selection
                 pure $
                   Just $
                     NonPersistentCommitteeMember
-                      (NonPersistentMemberProof numSeats vrfOutput sig)
+                      (NonPersistentMembershipProof numSeats vrfOutput sig)
                       ourStake
   | otherwise =
       Left (MissingPoolId ourId)
@@ -457,7 +457,7 @@ verifyVote vote selection =
           pure $
             Just $
               PersistentCommitteeMember
-                (PersistentMemberProof seatIndex sig)
+                (PersistentMembershipProof seatIndex sig)
                 voterStake
       | otherwise -> do
           Left (NotAPersistentMember seatIndex)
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+--   VoterId 2 -> 15
+--   VoterId 3 -> 10
+--   VoterId 4 -> 20
+--   VoterId 5 -> 5
+-- @
+--
+--  We would have the following cummulative stake distribution:
+--
+-- @
+--   Map.fromList
```
perturbing: This is awesome and a good optimization to share data between peras and leios 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+--     , (VoterId 3, SeatIndex 3, LedgerStake 10, CumulativeStake 15)
+--     , (VoterId 5, SeatIndex 4, LedgerStake 5,  CumulativeStake 5)
+--     ]
+-- @
+--
+-- NOTE: this wrapper exists to allow us to share the same cummulative stake
+-- distribution across multiple committee selection instances derived from the
+-- same underlying stake distribution (e.g. Leios and Peras voting committees
+-- for the same epoch).
+newtype CumulativeStakeDistr
```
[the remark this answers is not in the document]
agustinmista: I had already renamed this to `ExtCumulativeStakeDistr` to follow the convention used by other "extended" data types in consensus.

I'm not very attached to this name though, so feel free to insist and I will change it haha.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Types.hs
```diff
@@ -0,0 +1,198 @@
...
+  deriving (Show, Eq, Ord)
+
+-- * Committee size
+
+-- | Target committee size
+newtype CommitteeSize = CommitteeSize {unCommitteeSize :: Word64}
+  deriving (Show, Eq, Ord)
+
+-- | Persistent committee size
+newtype PersistentCommitteeSize = PersistentCommitteeSize
```
tbagrel1: Why not using a similar type-level label as for `Stake`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,121 @@
...
+  (CumulativeStake cumulativeStake)
+    | cumulativeStake <= 0 =
+        False -- Avoid division by zero in the left-hand side of the inequality
+    | voterSeat > totalSeats =
+        False -- Avoid underflow in the right-hand side of the inequality
+    | otherwise =
+        ( (1 - (voterStake / cumulativeStake))
+            ^ (2 :: Integer)
+        )
+          < ( toRational (totalSeats - voterSeat)
```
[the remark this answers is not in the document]
perturbing: I have a flaked version of it [here](https://github.com/input-output-hk/ouroboros-leios/tree/perturbing/flake-crypto-bench). 

Note that the implementation [here](https://github.com/cardano-scaling/leios-wfa-ls-demo), can be checked against the rust one via [these](https://github.com/cardano-scaling/leios-wfa-ls-demo/issues/1#issuecomment-3952671373) steps.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,138 @@
...
+  forall c.
+  CryptoSupportsVRF c =>
+  -- | Expected number of non-persistent voters in the committee
+  NonPersistentCommitteeSize ->
+  -- | Total stake of non-persistent voters
+  TotalNonPersistentStake ->
+  -- | Stake of the voter
+  Stake Ledger ->
+  -- | VRF output from the participant
+  VRFOutput c ->
```
tbagrel1: If we require `CryptoSupportsVRF c`, I would expect us to take a  `VRFPoolContext c` and `VRFElectionInput c` and compute the output ourselves. Or we could take a normalized rational direcly, and not depend on `CryptoSupportsVRF c` at all

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Types.hs
```diff
@@ -0,0 +1,184 @@
...
+
+-- | Non-persistent committee size
+newtype NonPersistentCommitteeSize = NonPersistentCommitteeSize
+  { unNonPersistentCommitteeSize :: Word64
+  }
+  deriving (Show, Eq, Ord)
+
+-- * Cumulative stake distributions
+
+-- | Extended cumulative stake distribution for a given role.
```
tbagrel1: "for a given role" I think this is some leftover/outdated description :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,116 @@
...
+-- seats according to the weighted Fait-Accompli scheme.
+--
+-- This function returns:
+--   * number of persistent seats granted via the weighted Fait-Accompli scheme
+--   * number of non-persistent seats expected to vote via local sortition
+--   * total persistent stake
+--   * total non-persistent stake
+--
+-- TODO: this can be more efficient if we compute the values directly on the array.
+weightedFaitAccompliSplitSeats ::
```
tbagrel1: Do you think this function should state and check the invariant that we must have at least the same number of non-zero-stake candidates as the committee size?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
agustinmista: I think the cleanest here was to refactor the function to take the normalized VRF output directly and drop the `CryptoSupportsVRF` constraint entirely.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Types.hs
agustinmista: Good catch. Thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,481 @@
...
+  ( CryptoSupportsWFALS c
+  , VoteSupportsWFALS c vote
+  ) =>
+  vote ->
+  CommitteeSelection c ->
+  Either CommitteeSelectionError (Maybe (CommitteeMember c))
+verifyVote vote selection =
+  getVoteView @c vote $ \case
+    PersistentVote seatIndex electionId message sig
+      | seatIndexWithinBounds seatIndex selection -> do
```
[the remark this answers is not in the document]
agustinmista: Yes, good catch! 

Likewise, we also need to check in the `NonPersistentVote` variant that, after retrieving the seat index of someone who claims to be a non-persistent voter, that seat index does _not_ correspond to a persistent member:

```haskell
NonPersistentVote poolId electionId message vrfOutput sig
  | Just seatIndex <- Map.lookup poolId (candidateSeats selection)
  , not (isPersistentMember seatIndex selection) ->
      ...
```

agustinmista:
Thanks @dnadales for the review! 

I think I addressed all your feedback now (last 5 commits in this PR, to be squashed) :crossed_fingers:

agustinmista:
Superseded by #1974 and #1975

PR #1938, opened 2026-03-24
[Peras 26] Concrete certs and votes using BLS crypto

### LINKED ISSUE #128: Use actual cryptography for votes/certs
At some point, we need to use actual cryptography for votes and certificates (see [here](https://tweag.github.io/cardano-peras/peras-design.pdf#appendix.B) for details).
Concretely (this ticket should be broken further down):

 - [ ] Committee selection, via [Fait Accompli](https://eprint.iacr.org/2023/1273.pdf) (either $\text{wFA}^{\text{IID}}$ or $\text{wFA}^{\text{LS}}$ or potentially one of the even more aggresive schemes currently in development at IOR).
 - [ ] Multisignature primitive, either [BLS](https://en.wikipedia.org/wiki/BLS_digital_signature) or [Pixel](https://eprint.iacr.org/2019/514) depending on whether we need forward secrecy (see #88).
 
   The IOG cryptography team has already started working on BLS Haskell bindings: https://github.com/IntersectMBO/cardano-base/pull/548

### LINKED ISSUE #212: Template Vote and Cert Cryptography

agustinmista:
This PR implements concrete Peras votes and certificates using BLS signatures, including all necessary cryptography infrastructure and integration with the existing voting committee abstractions.

- Added compact `ByteString`-based bitmap implementation to identify voters in a certificate
- Defined concrete `PerasVote` and `PerasCert` types using BLS signatures to carry validation and eligibility proofs
- Defined `PerasCrypto` scheme and implemented all the necessary crypto infrastructure to use it with both `WFALS` and `EveryoneVotes`
- Implemented `VotingCommitteeSupportsPeras` type class with conversions between concrete Peras types and the abstract `WFALS`/`EveryoneVotes` voting committee types
- Added property tests covering bitmap operations, serialization roundtrips, and conversion roundtrips (both total and partial)

### DIFF
--- a/changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
+++ b/changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
@@ -1,0 +1,28 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- Introduce `Ouroboros.Consensus.Util.Bitmap` providing `ByteString`-based compact bitmaps.
+- Define `PerasBLSCrypto` scheme with support for all the voting committee superclasses.
+- Define concrete `PerasVote` and `PerasCert` types using BLS signatures.
+- Define `PerasVoteCompatibleWithVotingCommittee` and `PerasCertCompatibleWithVotingCommittee` type classes with conversions between concrete Peras types and their abstract voting committee counterparts.
+- Instantiate `VotingCommitteeSupportsPeras` for both `WFALS` and `EveryoneVotes`.
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
@@ -1,0 +1,218 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Concrete Peras certificate types using BLS signatures.
+--
+-- NOTE: this module is meant to be imported qualified.
+--
+-- NOTE: the validation performed during serialization is minimal, and does not
+-- cover any of additional semantic and cryptographic checks that must be
+-- performed on the certificate later on.
+module Ouroboros.Consensus.Peras.Cert.V1
+  ( PerasCert (..)
+  , PerasCertVoters (..)
+  ) where
+
+import Cardano.Binary
+  ( FromCBOR (..)
+  , ToCBOR (..)
+  , decodeListLenOf
+  , encodeListLen
+  )
+import Control.Monad (when)
+import Control.Monad.Error.Class (MonadError (..))
+import Data.Containers.NonEmpty (HasNonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map.NonEmpty as NEMap
+import Data.Map.Strict (Map)
+import Data.Maybe (catMaybes)
+import Data.Word (Word16)
+import Ouroboros.Consensus.Block.SupportsPeras
+  ( PerasBoostedBlock
+  , PerasRoundNo
+  , PerasSeatIndex (..)
+  )
+import Ouroboros.Consensus.Committee.Crypto
+  ( CryptoSupportsAggregateVoteSigning (..)
+  )
+import Ouroboros.Consensus.Peras.Crypto.BLS
+  ( PerasBLSCrypto
+  , VRFOutput
+  )
+import Ouroboros.Consensus.Peras.Vote.V1 (PerasVoteEligibilityProof (..))
+import Ouroboros.Consensus.Util.Bitmap (Bitmap)
+import qualified Ouroboros.Consensus.Util.Bitmap as Bitmap
+
+-- | Concrete Peras certificates using BLS signatures
+data PerasCert
+  = PerasCert
+  { pcRoundNo :: !PerasRoundNo
+  -- ^ Election identifier
+  , pcBoostedBlock :: !PerasBoostedBlock
+  -- ^ Certificate message, i.e., the hash of the block being boosted
+  , pcVoters :: !PerasCertVoters
+  -- ^ Voters who contributed to this certificate
+  , pcSignature :: !(AggregateVoteSignature PerasBLSCrypto)
+  -- ^ Aggregate BLS signature on the hash of the election identifier and
+  -- the certificate message
+  }
+  deriving (Show, Eq)
+
+instance FromCBOR PerasCert where
+  fromCBOR = do
+    decodeListLenOf 4
+    pcRoundNo <- fromCBOR
+    pcBoostedBlock <- fromCBOR
+    pcVoters <- fromCBOR
+    pcSignature <- fromCBOR
+    pure
+      PerasCert
+        { pcRoundNo
+        , pcBoostedBlock
+        , pcVoters
+        , pcSignature
+        }
+
+instance ToCBOR PerasCert where
+  toCBOR cert =
+    encodeListLen 4
+      <> toCBOR (pcRoundNo cert)
+      <> toCBOR (pcBoostedBlock cert)
+      <> toCBOR (pcVoters cert)
+      <> toCBOR (pcSignature cert)
+
+-- | Voters contained in a certificate with their appropriate eligibility proof
+newtype PerasCertVoters
+  = PerasCertVoters
+  { unPerasCertVoters ::
+      NE (Map PerasSeatIndex PerasVoteEligibilityProof)
+  }
+  deriving (Eq, Show)
+
+instance FromCBOR PerasCertVoters where
+  fromCBOR = do
+    decodeListLenOf 2
+    votersBitmap <- fromCBOR
+    nonPersistentSigs <- fromCBOR
+
+    either fail pure
+      . fromCompactRepr
+      $ CompactPerasCertVoters
+        { votersBitmap
+        , nonPersistentSigs
+        }
+
+instance ToCBOR PerasCertVoters where
+  toCBOR voters =
+    encodeListLen 2
+      <> toCBOR votersBitmap
+      <> toCBOR nonPersistentSigs
+   where
+    CompactPerasCertVoters
+      { votersBitmap
+      , nonPersistentSigs
+      } =
+        toCompactRepr voters
+
+-- | Compact representation of the voters in a Peras certificate.
+--
+-- This compact representation consists of a bitmap of voter seat indices and a
+-- list of non-persistent eligibility proofs (VRF outputs). In this setup, the
+-- last @np@ indices in the bitmap that are flipped to 1 correspond to
+-- non-persistent voters, where @np@ is the length of the list of non-persistent
+-- eligibility proofs. The remaining flipped indices in the bitmap correspond
+-- to persistent voters.
+--
+-- @
+--   fromCompactRepr
+--      CompactPerasCertVoters {
+--        votersBitmap = <01101011>,
+--        nonPersistentSigs = [np1, np2, np3]
+--      }
+--   ==
+--   PerasCertVoters {
+--     1 => persistent
+--     2 => persistent
+--     4 => non-persistent(np1)
+--     6 => non-persistent(np2)
+--     7 => non-persistent(np3)
+--   }
+-- @
+data CompactPerasCertVoters
+  = CompactPerasCertVoters
+  { votersBitmap :: !(Bitmap Word16)
+  , nonPersistentSigs :: ![VRFOutput PerasBLSCrypto]
+  }
+  deriving (Eq, Show)
+
+-- | Decode 'PerasCertVoters' from their compact representation.
+--
+-- See 'CompactPerasCertVoters' for the encoding scheme used here.
+fromCompactRepr ::
+  CompactPerasCertVoters ->
+  Either String PerasCertVoters
+fromCompactRepr
+  CompactPerasCertVoters
+    { votersBitmap
+    , nonPersistentSigs
+    } = do
+    let voterSeatIndices =
+          PerasSeatIndex <$> Bitmap.toIndices votersBitmap
+
+    when (null voterSeatIndices) $
+      throwError "Invalid Peras certificate: empty voters bitmap"
+
+    when (length nonPersistentSigs > length voterSeatIndices) $
+      throwError $
+        unlines
+          [ "Invalid Peras certificate:"
+              <> " more non-persistent voter eligibility proofs were provided"
+              <> " than the number of voters in the certificate"
+          , " * number of voters: "
+              <> show (length voterSeatIndices)
+          , " * number of proofs: "
+              <> show (length nonPersistentSigs)
+          ]
+
+    let numPersistentVoters =
+          length voterSeatIndices - length nonPersistentSigs
+    let persistentProofs =
+          take numPersistentVoters (repeat PersistentPerasVoteEligibilityProof)
+    let nonPersistentProofs =
+          fmap NonPersistentPerasVoteEligibilityProof nonPersistentSigs
+    let voters =
+          NEMap.fromAscList
+            . NonEmpty.fromList
+            . zip voterSeatIndices
+            $ persistentProofs <> nonPersistentProofs
+
+    pure (PerasCertVoters voters)
+
+-- | Encode 'PerasCertVoters' into their compact representation.
+--
+-- See 'CompactPerasCertVoters' for the encoding scheme used here.
+toCompactRepr ::
+  PerasCertVoters ->
+  CompactPerasCertVoters
+toCompactRepr (PerasCertVoters voters) =
+  CompactPerasCertVoters
+    { votersBitmap
+    , nonPersistentSigs
+    }
+ where
+  logicalUpperBound =
+    unPerasSeatIndex (fst (NEMap.findMax voters))
+  votersByAscSeatIndex =
+    NonEmpty.toList (NEMap.toAscList voters)
+  votersSeatIndices =
+    fmap (unPerasSeatIndex . fst) votersByAscSeatIndex
+  votersBitmap =
+    Bitmap.fromIndices logicalUpperBound votersSeatIndices
+  nonPersistentSigs =
+    catMaybes (fmap getNonPersistentSig votersByAscSeatIndex)
+  getNonPersistentSig = \case
+    (_, PersistentPerasVoteEligibilityProof) -> Nothing
+    (_, NonPersistentPerasVoteEligibilityProof p) -> Just p
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
@@ -1,0 +1,283 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | BLS-based crypto scheme used in Peras voting committees
+module Ouroboros.Consensus.Peras.Crypto.BLS
+  ( PerasBLSCrypto
+  , ElectionId
+  , VoteCandidate
+  , PerasPrivateKey (..)
+  , PerasPublicKey (..)
+  , VoteSignature (..)
+  , VRFElectionInput (..)
+  , VRFOutput (..)
+  , AggregateVoteVerificationKey
+  , AggregateVoteSignature
+
+    -- * For testing purposes
+  , PerasBLSCryptoAggregateVoteVerificationKey (..)
+  , PerasBLSCryptoAggregateVoteSignature (..)
+  ) where
+
+import Cardano.Binary (FromCBOR, ToCBOR (..))
+import Cardano.Crypto.DSIGN (BLS12381MinSigDSIGN, DSIGNAlgorithm (..))
+import Cardano.Crypto.Hash (Hash)
+import qualified Cardano.Crypto.Hash as Hash
+import Cardano.Ledger.BaseTypes (Nonce (..), SlotNo (..))
+import Cardano.Ledger.Binary (runByteBuilder)
+import Cardano.Ledger.Hashes (HASH)
+import qualified Data.ByteString.Builder as BS
+import qualified Data.ByteString.Builder.Extra as BS
+import qualified Data.ByteString.Short as BS
+import Ouroboros.Consensus.Block.RealPoint
+  ( bytes32RealPointHash
+  , bytes32RealPointSlot
+  )
+import Ouroboros.Consensus.Block.SupportsPeras
+  ( PerasBoostedBlock (..)
+  , PerasRoundNo (..)
+  )
+import Ouroboros.Consensus.Committee.Crypto
+  ( CryptoSupportsAggregateVoteSigning (..)
+  , CryptoSupportsBatchVRFVerification (..)
+  , CryptoSupportsVRF (..)
+  , CryptoSupportsVoteSigning (..)
+  , ElectionId
+  , PrivateKey
+  , PublicKey
+  , VRFPoolContext (..)
+  , VoteCandidate
+  )
+import Ouroboros.Consensus.Committee.Crypto.BLS (KeyRole (..))
+import qualified Ouroboros.Consensus.Committee.Crypto.BLS as BLS
+
+-- | BLS-based crypto scheme used in Peras voting committees
+data PerasBLSCrypto
+
+type instance ElectionId PerasBLSCrypto = PerasRoundNo
+type instance VoteCandidate PerasBLSCrypto = PerasBoostedBlock
+
+-- | Private key of a Peras committee member
+data PerasPrivateKey
+  = PerasPrivateKey
+  { perasVoteSignKey :: BLS.PrivateKey SIGN
+  , perasVRFSignKey :: BLS.PrivateKey VRF
+  }
+
+type instance PrivateKey PerasBLSCrypto = PerasPrivateKey
+
+-- | Public key of a Peras committee member
+data PerasPublicKey
+  = PerasPublicKey
+  { perasVoteVerKey :: BLS.PublicKey SIGN
+  , perasVRFVerKey :: BLS.PublicKey VRF
+  }
+
+type instance PublicKey PerasBLSCrypto = PerasPublicKey
+
+-- | Hash the message of a Peras vote
+--
+-- NOTE: this is inspired by the implementation used by the Praos VRF check in
+-- 'Ouroboros.Consensus.Protocol.Praos.VRF.mkInputVRF'.
+hashVoteSignature ::
+  ElectionId PerasBLSCrypto ->
+  VoteCandidate PerasBLSCrypto ->
+  Hash HASH (SigDSIGN BLS12381MinSigDSIGN)
+hashVoteSignature roundNo boostedBlock =
+  Hash.castHash
+    . Hash.hashWith id
+    . runByteBuilder (8 + 8 + 32)
+    $ roundNoBytes
+      <> boostedBlockSlotBytes
+      <> boostedBlockHashBytes
+ where
+  roundNoBytes =
+    BS.word64BE
+      . unPerasRoundNo
+      $ roundNo
+  boostedBlockSlotBytes =
+    BS.word64BE
+      . unSlotNo
+      . bytes32RealPointSlot
+      . unPerasBoostedBlock
+      $ boostedBlock
+  boostedBlockHashBytes =
+    BS.byteStringCopy
+      . BS.fromShort
+      . bytes32RealPointHash
+      . unPerasBoostedBlock
+      $ boostedBlock
+
+-- | Hash the input for the VRF used in Peras elections
+--
+-- NOTE: this is inspired by the implementation used by the Praos VRF check in
+-- 'Ouroboros.Consensus.Protocol.Praos.VRF.mkInputVRF'.
+hashVRFInput ::
+  ElectionId PerasBLSCrypto ->
+  Nonce ->
+  Hash HASH (SigDSIGN BLS12381MinSigDSIGN)
+hashVRFInput roundNo epochNonce =
+  Hash.castHash
+    . Hash.hashWith id
+    . runByteBuilder (8 + 32)
+    $ roundNoBytes <> epochNonceBytes
+ where
+  roundNoBytes =
+    BS.word64BE (unPerasRoundNo roundNo)
+  epochNonceBytes =
+    case epochNonce of
+      NeutralNonce -> mempty
+      Nonce h -> BS.byteStringCopy (Hash.hashToBytes h)
+
+-- * Crypto instances
+
+instance CryptoSupportsVoteSigning PerasBLSCrypto where
+  type VoteSigningKey PerasBLSCrypto = BLS.PrivateKey SIGN
+  type VoteVerificationKey PerasBLSCrypto = BLS.PublicKey SIGN
+
+  newtype VoteSignature PerasBLSCrypto
+    = PerasBLSCryptoVoteSignature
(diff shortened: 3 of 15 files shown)


changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,27 @@
...
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+### Non-Breaking
+
+- Introduce Ouroboros.Consensus.Util.Bitmap providing ByteString-based compact bitmaps.
```
tbagrel1:
```suggestion
- Introduce `Ouroboros.Consensus.Util.Bitmap` providing `ByteString`-based compact bitmaps.
```

changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,27 @@
...
+- Define concrete PerasVote and PerasCert types using BLS signatures.
```
tbagrel1:
```suggestion
- Define concrete `PerasVote` and `PerasCert` types using BLS signatures.
```

changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,27 @@
...
+- Define PerasCrypto scheme with support for all the voting committee superclasses.
```
tbagrel1:
```suggestion
- Define `PerasCrypto` scheme with support for all the voting committee superclasses.
```

changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,27 @@
...
+- Define VotingCommitteeSupportsPeras type class with conversions between concrete Peras types and their abstract counterparts.
```
tbagrel1:
```suggestion
- Define `VotingCommitteeSupportsPeras` type class with conversions between concrete Peras types and their abstract counterparts.
```

changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,27 @@
...
+- Instantiate VotingCommitteeSupportsPeras for both WFALS and EveryoneVotes.
```
tbagrel1:
```suggestion
- Instantiate `VotingCommitteeSupportsPeras` for both WFALS and EveryoneVotes.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,324 @@
...
+data PerasConversionError
+  = EveryoneVotesButFoundNonPersistentVoterInVote SeatIndex
+  | EveryoneVotesButFoundNonPersistentVotersInCert (NE [SeatIndex])
+  | SeatIndexOverflowError Word64
+  | BLSError String
+  deriving stock (Eq, Show)
+
+-- | Voting committees supported by Peras, providing conversion between concrete
+-- Peras votes and certificates and their abstract counterparts.
+class CommitteeSupportsPeras committee where
```
tbagrel1: We may rename it to something like `CommitteeCompatibleWithConcretePerasTypes`. Not pretty, but it gives a very clear idea of what this class is doing.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,324 @@
...
+-- | Errors that can occur when converting between Peras and committee types
+data PerasConversionError
+  = EveryoneVotesButFoundNonPersistentVoterInVote SeatIndex
+  | EveryoneVotesButFoundNonPersistentVotersInCert (NE [SeatIndex])
+  | SeatIndexOverflowError Word64
+  | BLSError String
+  deriving stock (Eq, Show)
+
+-- | Voting committees supported by Peras, providing conversion between concrete
+-- Peras votes and certificates and their abstract counterparts.
```
tbagrel1: Problem is, the concrete Peras votes and certs are supposed to be a type family depending on `blk` (so the definition can evolve with eras).

From my understanding, to instantiate the class below, you need to know how to convert from the commitee abstract types to _all_ of the Peras era-specific concrete types?

tbagrel1 (commented):
I could only review the first two files before the weekly. I will continue reviewing the rest later that day (from a different computer so I need to push this review now)

changelog.d/20260421_170000_agustin.mista_concrete_certs_and_votes.md
```diff
@@ -0,0 +1,28 @@
...
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- Introduce `Ouroboros.Consensus.Util.Bitmap` providing `ByteString`-based compact bitmaps.
+- Define concrete `PerasVote` and `PerasCert` types using BLS signatures.
+- Define `PerasCrypto` scheme with support for all the voting committee superclasses.
+- Define `VotingCommitteeSupportsPeras` type class with conversions between concrete Peras types and their abstract counterparts.
```
tbagrel1: Probably this point needs to be updated with the recent changes

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+    pcBoostedBlock <- fromCBOR
+    pcVoters <- fromCBOR
+    pcSignature <- fromCBOR
+    pure
+      PerasCert
+        { pcRoundNo
+        , pcBoostedBlock
+        , pcVoters
+        , pcSignature
+        }
```
tbagrel1, line 77: I'm wondering if it's worth writing these instances manually, or deriving them so they get automatically updated if we add/remove a field (provided there is a way to derive them automatically)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+  toCBOR voters =
+    encodeListLen 2
+      <> toCBOR votersBitmap
+      <> toCBOR nonPersistentProofs
+   where
+    (votersBitmap, nonPersistentProofs) =
+      either (\err -> error ("Internal encoding error: " <> err)) id $
+        reflectVoters voters
+
+reifyVoters ::
```
tbagrel1: Probably deserves some doc :)

tbagrel1: We must also that that at this point, we haven't checked the validity of voters

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+        [ "Invalid Peras certificate:"
+            <> " more non-persistent voter eligibility proofs where provided"
+            <> " than the number of voters in the certificate"
+        , " * number of voters: "
+            <> show (length voterSeatIndices)
+        , " * number of proofs: "
+            <> show (length nonPersistentSignatures)
+        ]
+
+  let numPersistentVoters =
```
tbagrel1: We can probably remove the many subsequent `let`s for just a single one?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+        fmap NonPersistentPerasVoteEligibilityProof nonPersistentSignatures
+  let voters =
+        NEMap.fromAscList
+          . NonEmpty.fromList
+          . zip voterSeatIndices
+          $ persistentProofs <> nonPersistentProofs
+
+  pure (PerasCertVoters voters)
+
+reflectVoters ::
```
tbagrel1: I'm not fully convinced by the `reify`/`reflect` terminology for voters. Something like `toNetworkRepr`/`fromNetworkRepr` or  `toCompactRepr`/`fromCompactRepr` might be slightly more descriptive

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+  PerasCertVoters ->
+  Either String (Bitmap Word16, [VRFOutput PerasBLSCrypto])
+reflectVoters (PerasCertVoters voters) = do
+  let logicalUpperBound =
```
tbagrel1: Why `logical`?
Also we can probably remove the successive `let`s here too

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+        unPerasSeatIndex (fst (NEMap.findMax voters))
+  let votersByAscSeatIndex =
+        NonEmpty.toList (NEMap.toAscList voters)
+  let votersSeatIndices =
+        fmap (unPerasSeatIndex . fst) votersByAscSeatIndex
+
+  when (length votersSeatIndices > fromIntegral logicalUpperBound + 1) $
```
tbagrel1: I think this check cannot fail at all. Seat index is a positive integral type, and there can't be more than $n$ elements in a map whose max key is $n-1$.
The error message implies it would be a faulty cert, while I think it would rather denote of a faulty implem on our side.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,179 @@
...
+        , " * voters bitmap size: "
+            <> show @Int (fromIntegral logicalUpperBound + 1)
+        ]
+
+  let votersBitmap =
+        Bitmap.fromIndices
+          logicalUpperBound
+          votersSeatIndices
+  let nonPersistentProofs =
+        catMaybes $ flip fmap votersByAscSeatIndex $ \(_, proof) ->
```
tbagrel1: nitpicking, but maybe we can rewrite that in a slightly more obvious way

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,354 @@
...
+data PerasPublicKey
+  = PerasPublicKey
+  { perasVoteVerKey :: BLS.PublicKey SIGN
+  , perasVRFVerKey :: BLS.PublicKey VRF
+  }
+
+type instance PublicKey PerasBLSCrypto = PerasPublicKey
+
+-- | Hash the message of a Peras vote
+hashVoteSignature ::
```
tbagrel1, line 88: We should document how we come up with this implem. A link to the source/reference would be enough I think

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,354 @@
...
+
+-- | Hash the input for the VRF used in Peras elections
+hashVRFInput ::
+  ElectionId PerasBLSCrypto ->
+  Nonce ->
+  Hash HASH (SigDSIGN BLS12381MinSigDSIGN)
+hashVRFInput roundNo epochNonce =
+  Hash.castHash
+    . Hash.hashWith id
+    . runByteBuilder (8 + 32)
```
tbagrel1, line 128: Is it ok to `runByteBuilder` with size (8 + 32) even though we may not have bytes for the epoch nonce?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,354 @@
...
+
+toPerasBLSVoteSignature ::
+  VoteSignature PerasBLSCrypto ->
+  PerasBLSVoteSignature
+toPerasBLSVoteSignature = coerce
+
+fromPerasBLSVoteSignature ::
+  PerasBLSVoteSignature ->
+  VoteSignature PerasBLSCrypto
+fromPerasBLSVoteSignature = coerce
```
tbagrel1: Not sure what is the role of this function here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,354 @@
...
+    roundNo
+    boostedBlock
+    (PerasBLSCryptoVoteSignature sig) =
+      BLS.verifyWithRole @SIGN
+        pk
+        (hashVoteSignature roundNo boostedBlock)
+        sig
+
+instance CryptoSupportsVRF PerasBLSCrypto where
+  type VRFSigningKey PerasBLSCrypto = BLS.PrivateKey VRF
```
tbagrel1, line 173: I don't understand why we aren't using our newly defined `PerasBLSCryptoVoteSignature` wrappers here. What are they used for then?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,354 @@
...
+        pure $ PerasBLSCryptoVRFOutput sig
+
+  normalizeVRFOutput (PerasBLSCryptoVRFOutput sig) =
+    BLS.toNormalizedVRFOutput sig
+
+-- * Support for aggregate signatures and VRF outputs
+
+instance CryptoSupportsAggregateVoteSigning PerasBLSCrypto where
+  -- NOTE: for now, we individually collect the verification keys of all the
+  -- voters who contributed to the certificate being verified.
```
tbagrel1: Why are we reimplementing some naive-like aggregation logic, given that BLS supports real/efficient aggregation? I think this comment should be expanded

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,295 @@
...
+
+-- | Errors that can occur when converting between Peras and committee types
+data PerasConversionError
+  = EveryoneVotesButFoundNonPersistentVoterInVote SeatIndex
+  | EveryoneVotesButFoundNonPersistentVotersInCert (NE [SeatIndex])
+  | SeatIndexOverflowError Word64
+  | CryptoError String
+  deriving stock (Eq, Show)
+
+-- | Voting committees a concrete Peras vote can be used with
```
tbagrel1: I would like to rephrase this a bit, same below

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,295 @@
...
+class PerasVoteSupportsVotingCommittee vote committee where
+  toPerasVote ::
+    Committee.Vote (PerasVoteCrypto vote) committee ->
```
tbagrel1: Maybe we could make the crypto argument explicit, with fun deps
I think it would be more consistent with what we did previously in Peras 21-25

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,295 @@
...
+-- | Errors that can occur when converting between Peras and committee types
+data PerasConversionError
+  = EveryoneVotesButFoundNonPersistentVoterInVote SeatIndex
+  | EveryoneVotesButFoundNonPersistentVotersInCert (NE [SeatIndex])
+  | SeatIndexOverflowError Word64
+  | CryptoError String
+  deriving stock (Eq, Show)
+
+-- | Voting committees a concrete Peras vote can be used with
+class PerasVoteSupportsVotingCommittee vote committee where
```
tbagrel1: Still not convinced by the name. I would probably go for something like `ConcretePerasVoteCompatibleWithCommittee`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,295 @@
...
+  toPerasCert ::
+    Committee.Cert (PerasCertCrypto cert) committee ->
+    Either PerasConversionError cert
+  fromPerasCert ::
+    cert ->
+    Either PerasConversionError (Committee.Cert (PerasCertCrypto cert) committee)
+
+-- 'V1.PerasVote's are compatible with 'WFALS' as long as we make sure to avoid
+-- overflowing their `Word16` seat index.
+instance PerasVoteSupportsVotingCommittee V1.PerasVote WFALS where
```
tbagrel1: Shoudln't these instances be in the V1.hs files instead?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,295 @@
...
+          Just (NonEmpty.fromList nonPersistentSeats)
+
+-- * Helpers
+
+-- | Convert a Peras seat index to a committee seat index.
+fromPerasSeatIndex ::
+  PerasSeatIndex ->
+  SeatIndex
+fromPerasSeatIndex (PerasSeatIndex seatIndex) =
+  SeatIndex (fromIntegral @Word16 @Word64 seatIndex)
```
tbagrel1, line 247: Not directly related to this PR, but I'm wondering if we should use a non-bounded integral type for the abstract committee selection side, instead of Word64 (if we wanna emphasize on the abstract characteristic)

tbagrel1 (commented):
This looks really solid! Nice work Agustin.

I think there may still be a bit of room for improvement about what is concrete/abstract, why each wrapper is necessary/where it is used.
I would be in favor of giving them longer but more descriptive names, to avoid having several types with almost the same names in different modules.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: I would be against automation here. IIUC, the order of the fields in the data type definition could affect the order in the expected CBOR layout --- most likely not not what we want.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: I usually put one `let` per binding* because i) it's more diff-friendly (e.g. nothing breaks if you move the top binding and leave the rest unchanged), and ii) I find having one very distinctively highlighted `let` keyword next to each binding much easier to scan through, especially with indentation set to only 2 spaces.

[*] notably not possible with mutually-recursive bindings, as those need to be type-checked "at the same type" (same SCC), and thus need to be under the same `let`.

I wish the formatter had a stronger opinion though, but I guess it can't because of my note above about mutually-recursive bindings.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: I agree. I changed this to `(to|from)CompactRepr` now :smile:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: This is the terminology used in `O.C.Util.Bitmap`.

FTR: the difference between logical and real upper bound is the remainder from having bytestrings aligned on `Word8`s:

```
indices = [ 1, 3, 7, 10, 12 ]
==>

fromIndices indices 
= 
                        v logical upper bound = 12
Bitmap 12 <01010001|00101???> 
                           ^ real upper bound = 15
```

In _our_ representation, we want the logical upper bound of the bitmap to coincide with the max seat index to save some space, but a different use case could instead use a larger logical upper bound to e.g. have a fixed sized bitmap regardless of the data it contains.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: You're right! We shouldn't need this check at all.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: I gave the `(Bitmap Wor16, [VRFOutput PerasBLSCrypto])` a proper name and attached some documentation/examples about the encoding to its type.

> We must also that that at this point, we haven't checked the validity of voters

Yes, it's better if serialization is not at all concerned about validation beyond basic syntactic properties. I will add a disclaimer to the decoder.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: I split the mapped function into a separate binding and removed the `flip`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: This is very similar to how it's done for the Praos VRF:

image

And after tracking down `runByteBuilder`, it seems that the number passed is not the size, but a _size hint_:

image

So we should be fine, I think!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: Added a link to the reference code from `O.C.Protocol.Praos.VRF.mkInputVRF`.

I also added this to `hashVRFInput` and the corresponding test versions of these functions in `TestCrypto` (to be backported to Peras 25)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: This will be simplified by removing the Semigroup requirement on the aggregation step

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: I think I would prefer to keep this in a single file to keep the entire marhsalling implementation in one place.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: Word64 is probably good enough for now :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: No longer needed after the refactor on aggregatable types

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: No longer relevant after the refactor on aggregatable types

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: I turned `crypto` into a fundep from the concrete vote/cert type being instantiated.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: I renamed this to `Peras(Vote|Cert)CompatibleWithVotingCommittee`

The rationale: I don't think we would ever need to disambiguate Peras votes/cert w.r.t. being abstract or concrete. To the contrary: in my head, a Peras vote/cert is always concrete, and only when turned into a `Committee.Vote/Cert crypto committee` is that we consider them abstract, at which point, there's nothing Peras specific in the pertinent (generic) codebase.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: See my other comment

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -0,0 +1,283 @@
...
+
+  getVRFSigningKey _ =
+    perasVRFSignKey
+
+  getVRFVerificationKey _ =
+    perasVRFVerKey
+
+  mkVRFElectionInput roundNo epochNonce =
+    PerasBLSCryptoVRFElectionInput $
+      hashVRFInput epochNonce roundNo
```
geo2a: The type of `mkVRFElectionInput` is `Nonce -> ElectionId crypto -> VRFElectionInput crypto`, so the arguments need to be renamed:

```suggestion
  mkVRFElectionInput epochNonce roundNo =
    PerasBLSCryptoVRFElectionInput $
      hashVRFInput roundNo epochNonce
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,303 @@
...
+  | SeatIndexOverflowError Word64
+  | CryptoError String
+  deriving stock (Eq, Show)
+
+-- | Conversion between (concrete) Peras votes and (abstract) committee votes.
+--
+-- NOTE: the functional dependency @vote -> crypto@ explicitly ties each
+-- concrete Peras vote type to a specific crypto scheme.
+class
+  PerasVoteCompatibleWithVotingCommiittee vote crypto committee
```
geo2a:
```suggestion
  PerasVoteCompatibleWithVotingCommittee vote crypto committee
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,303 @@
...
+    vote ->
+    Either PerasConversionError (Committee.Vote crypto committee)
+
+-- | Conversion between (concrete) Peras certificates and (abstract) committee
+-- certificates.
+--
+-- NOTE: the functional dependency @cert -> crypto@ explicitly ties each
+-- concrete Peras certificate type to a specific crypto scheme.
+class
+  PerasCertCompatibleWithVotingCommiittee cert crypto committee
```
geo2a:
```suggestion
  PerasCertCompatibleWithVotingCommittee cert crypto committee
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,218 @@
...
+  }
+  deriving (Eq, Show)
+
+instance FromCBOR PerasCertVoters where
+  fromCBOR = do
+    decodeListLenOf 2
+    votersBitmap <- fromCBOR
+    nonPersistentSigs <- fromCBOR
+
+    either (\err -> fail ("Invalid Peras certificate: " <> err)) pure
```
geo2a: `fromCompactRepr` already prepends `"Invalid Peras certificate: "`, so these error messages will read `"Invalid Peras certificate: Invalid Peras certificate: ..."`. I suggest to drop the prefix here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,218 @@
...
+          PerasSeatIndex <$> Bitmap.toIndices votersBitmap
+
+    when (null voterSeatIndices) $
+      throwError "Invalid Peras certificate: empty voters bitmap"
+
+    when (length nonPersistentSigs > length voterSeatIndices) $
+      throwError $
+        unlines
+          [ "Invalid Peras certificate:"
+              <> " more non-persistent voter eligibility proofs where provided"
```
geo2a:
```suggestion
              <> " more non-persistent voter eligibility proofs were provided"
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -0,0 +1,303 @@
...
+      let voters' = fromPerasCertVoters voters
+      pure $
+        WFALSCert
+          electionId
+          candidate
+          voters'
+          sig
+
+-- 'V1.PerasVote's are compatible with 'EveryoneVotes' as long as we make sure
+-- to only accept votes with persistent elegibility proofs (in addition to
```
geo2a:
```suggestion
-- to only accept votes with persistent eligibility proofs (in addition to
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -0,0 +1,218 @@
...
+--     2 => persistent
+--     4 => non-persistent(np1)
+--     6 => non-persistent(np2)
+--     7 => non-persistent(np3)
+--   }
+-- @
+data CompactPerasCertVoters
+  = CompactPerasCertVoters
+  { votersBitmap :: !(Bitmap Word16)
+  , nonPersistentSigs :: !([VRFOutput PerasBLSCrypto])
```
geo2a:
```suggestion
  , nonPersistentSigs :: ![VRFOutput PerasBLSCrypto]
```

geo2a (approved):
The bulk of the changes look good to me, but I have to admit that I lack context to provide deep feedback. I have some superficial comments related to naming and typos.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: Good catch!

jasagredo (commented):
Nothing jumped at me except for the V1.hs module renaming

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/Bitmap.hs
ch1bo, line 1: @agustinmista Just saw this being merged by accident and thought I should let you know that there is now also a `BitField` implemented as part of `cardano-crypto-leios`: https://github.com/IntersectMBO/cardano-base/blob/master/cardano-crypto-leios/src/Cardano/Crypto/Leios.hs

I made it re-usable at first, but then doubled down on doing just the bare minimum needed for Leios certs. Happy to try find a common denominator :)

agustinmista: Hey @ch1bo!

Nice. This Bitmap implementation is actually based on the one you guys previously had for Leios, so I'm happy to see we went full circle :sweat_smile:

PR #1963, opened 2026-04-07
Leios: Prototype voting

### LINKED ISSUE #790: Prototype Voting
## Why
Within our prototyping effort, we also want to "trace the path" of voting and uncover any problems on it. While #774 deliberately takes the short-cut of treating any EB as certified, thus not needing voting. We want to close that gap and see vote traffic affect the prototype networks already.

Furthermore, we are currently also discussing the committee selection scheme #856. Within that, we are going to need evidence about the vote traffic load #805 from simulations, but ideally can also validate the simulation behavior with a more "real" implementation. Lastly, adding voting to the prototype will test drive the network protocol and message structures.

We want to have this self-contained and not require #776 for this to work.

## What
- Out of scope: key registration and committee selection
  - Re-use existing key material (hackily re-interpret Ed25519 keys as BLS key material)
  - All pools are voters ("everyone votes" scheme in https://github.com/cardano-scaling/leios-wfa-ls-demo/issues/2 or https://github.com/IntersectMBO/ouroboros-consensus/pull/1975)
- Create, validate and propagate votes
  - A basic voting thread issues votes on EBs that are not too old (between 3 * `L_hdr` and the following `L_vote`)
  - Out of scope: Equivocation detection and EB validation
  - Votes are kept in memory (no persistence)
  - Votes are sent to downstream nodes (if they requested more votes)
  - Votes are individually validated before relayed
- Collect votes and aggregate a certificate
  - Trace all received votes and progress towards threshold (hard-coded)
  - Once threshold is reached, create a certificate and keep it in memory
  - Out of scope: use the certificate in block production (maybe easy after #774)

## How
- Drive development using `ThreadNet` property tests
- ~~Add vote messages to LeiosNotify/LeiosFetch mini-protocols~~
- Sketch `LeiosVotes` protocol as speculatively discussed in https://github.com/cardano-foundation/CIPs/pull/1167
  - [ ] https://github.com/cardano-scaling/cardano-blueprint/pull/65
- Create a basic vote forge loop
  - Maybe be optimistically driven by completing a tx closure
  - Trace EB candidates and when voted (don't spam)
- Implement a simple, but deduplicating vote fetching logic
  - Using an in-memory a `LeiosVotingState`
- Store certs along side votes in the `LeiosVotingState`

ch1bo:
First PR to add voting capabilities to the `leios-prototype`:

- Naiive N^2 vote diffusion via `LeiosNotify`
- A new `VoteState` that allows to `addVote` and `subscribeVotes`
- A voting thread that votes on completed EB closures
  - Missing: EB too old to vote check
  - Missing: EB validation
- Mocked voting key registration (re-uses KES key bytes)
- No vote validation (yet)
- Threadnet properties that drove the implementation so far
- Fixed re-trigger of EB notifications for already completed closures

Roughly the components that were implemented in this PR:

Leios voting (Current)(1)

**NOTE**: The failing test with seed `(SMGen 15462600834920737326 9156690725779514573,20)` is actually a lucky one that conflates two `VoterId` from the hacked KES key bytes. It will drive the follow-up PR that explores commitee selection paths.


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -377,15 +387,35 @@ mkHandlers
                       let !offers2' = Set.insert ebHash offers2
                       pure (offers1, offers2')
                     void $ MVar.tryPutMVar getLeiosReady ()
+                  MsgLeiosVotes vs -> do
+                    -- TODO: reduce this trace a bit to the essential
+                    traceWith tracer $ MkTraceLeiosPeer $ "MsgLeiosVotes " <> show vs
+                    mapM_ addVote vs
+                    forM_ vs $ \MkLeiosVote{electionId, ebHash, voterId} ->
+                      traceWith kernelTracer $
+                        TraceLeiosVoteAcquired
```
bladyjoker: Let's just give it a Vote?, you already have it in `vs`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -377,15 +387,35 @@ mkHandlers
                       let !offers2' = Set.insert ebHash offers2
                       pure (offers1, offers2')
                     void $ MVar.tryPutMVar getLeiosReady ()
+                  MsgLeiosVotes vs -> do
+                    -- TODO: reduce this trace a bit to the essential
+                    traceWith tracer $ MkTraceLeiosPeer $ "MsgLeiosVotes " <> show vs
+                    mapM_ addVote vs
```
bladyjoker, line 392: perf nit; Note it might be useful to have `addVotes` rather such that the backend can choose to implement this more efficiently.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -377,15 +387,35 @@ mkHandlers
...
+
+          LeiosVoteSubscription{getNextVote} <- subscribeVotes leiosVoteState
+          let processVote = do
+                vote <- getNextVote
+                -- FIXME: Yields each vote separately
+                pure $ MsgLeiosVotes [vote]
+
+          pure . leiosNotifyServerPeer $
+            atomically $
+              processEbNotification <|> processVote
```
bladyjoker, line 413: This is nice!

However, be careful here! We don't want to have a heavy "processor" to avoid contention problems. Rather you could say `atomically $ getNextVote <|> getNextEb` and continue in a non-atomic context (if you're not updating the "state" that needs to remain consistent).

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -448,6 +456,48 @@ initNodeKernel
...
+                      MkLeiosVote
+                        { electionId = point.pointSlotNo
+                        , voterId = me
+                        , ebHash = point.pointEbHash
+                        , voteSignature = True
+                        }
+                -- Store vote in memory and notify downstream peers
+                addVote vote
+                traceWith tracer TraceLeiosVoted{point, voter = me}
+                traceWith tracer TraceLeiosVoteAcquired{point, voter = me}
```
bladyjoker: Let's just use `vote` here?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -364,40 +368,33 @@ sqlQueryCompletedEbByPoint db ebPoint =
...
-  \    txBytesSize INTEGER NOT NULL\n\
-  \  );\n\
-  \"
+  unlines
+    [ "CREATE TABLE ebs ("
+    , "  ebSlot INTEGER NOT NULL,"
+    , "  ebHashBytes BLOB NOT NULL,"
+    , "  ebBytesSize INTEGER NOT NULL,"
+    , -- NULL = body not downloaded, >0 = txs missing, 0 = just completed, <0 = notified
+      "  missingTxCount INTEGER,"
```
bladyjoker, line 380: I'm thinking to reduce contention

```sql

CREATE TABLE ebStatuses (
  ebSlot INTEGER NOT NULL,
  ebHash BLOB NOT NULL,
  missingTxCount INTEGER,
  notified BOOL
)
```

This way we can add a new EB into `ebs` without waiting on transactions that are quite intensely trying to update the missing transaction count.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -311,7 +323,45 @@ prettyLeiosOutstanding x =
...
+demoLeiosFetchStaticEnv =
+  MkLeiosFetchStaticEnv
+    { maxRequestedBytesSize = 50 * million
+    , maxRequestedBytesSizePerPeer = 5 * million
+    , maxRequestBytesSize = 500 * thousand
+    , maxRequestsPerEb = 2
+    , maxRequestsPerTx = 2
+    , maxLeiosNotifyIngressQueue = 1 * millionBase2
+    , maxLeiosFetchIngressQueue = 50 * millionBase2
+    }
```
bladyjoker, line 355: Let me know if you want me to add this to cardano.json config.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -436,76 +488,78 @@ decodeLeiosEb = do
...
- where
-  hashOverhead = 1 + 1 -- bytestring major byte + a length = 32
-  minSizeOverhead = 1 + 1 -- int major byte + a value at low as 55
+-- | A vote in the Leios protocol.
+data LeiosVote = MkLeiosVote
+  { electionId :: ElectionId
+  , voterId :: VoterId
+  , ebHash :: EbHash
+  , voteSignature :: VoteSignature
+  }
```
bladyjoker, line 504: As we spoke, I'd rather see

```haskell

-- Why `Mk` prefix?
data LeiosVote = LeiosVote {
  votePoint :: LeiosPoint,
  voterId :: VoterId,
  voteSignature :: VoteSignature
}
```

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
```diff
@@ -0,0 +1,49 @@
...
+  seenVotes <- atomically $ newTVar Set.empty
+  pure
+    LeiosVoteState
+      { addVote = \vote -> atomically $ do
+          seen <- readTVar seenVotes
+          if Set.member vote seen
+            then pure ()
+            else do
+              writeTVar seenVotes $! Set.insert vote seen
+              writeTChan votesChan vote
```
bladyjoker, line 42: Because I'm seeing this in several places, use `alterF` "generaelized set mutation"

```haskell
import Control.Concurrent.STM

-- | Inserts an element and returns 'True' if it was newly added.
insertUnique :: (Ord a) => a -> Set.Set a -> (Bool, Set.Set a)
insertUnique vote = Set.alterF (\exists -> (not exists, True)) vote

-- Your updated addVote:
addVote = \vote -> atomically $ do
    seen <- readTVar seenVotes
    let (inserted, nextSeen) = insertUnique vote seen
    
    if inserted
        then do
            writeTVar seenVotes nextSeen
            writeTChan votesChan vote
        else pure ()
```

bladyjoker:
```haskell
addVotes = \newVotes -> atomically $ do
    seen <- readTVar seenVotes
    
    -- 1. Find elements in newVotes that aren't in seen
    let added = Set.difference newVotes seen
    
    -- 2. If there are new items, update the state and the channel
    if Set.null added
        then pure ()
        else do
            -- 3. Efficiently merge the two sets
            let nextSeen = Set.union seen added
            writeTVar seenVotes $! nextSeen
            
            -- 4. Stream the new items to the channel
            mapM_ (writeTChan votesChan) added -- WOULD EVEN NOTIFY A `Set LeiosVote` rather than one by one.
```

something that works on batches

bladyjoker (approved):

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: Ah, so another resolution table like `ebTxs`? It would be keyed by `ebHashBytes` only though (slot does not matter for completion tracking)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: > Why `Mk` prefix?

It was used in the demo code in this module when @nfrisby wrote it and I usually stick to consistency (at least on module level). I might have seen it in other places on the codebase too? Are there naming / code conventions for `ouroboros-consensus`?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
ch1bo: I think about the VoteState a bit like the Mempool and have been inspired by that API (specifically the `addTx`). Also, I don't think we will keep the 1:1 mapping of `network message -> validate,store,notify -> network message` of this naiive implementation anyways (where the batched API would be _required_ to avoid buffering)

.. I think I don't want to be too smart about it yet. Is that okay?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
ch1bo: I agree with the code for the batched API, but not so much with the alterF. The Set.member is more clear to me.

See the other comment why I would stick with the singular API for now.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
ch1bo: Addressed in https://github.com/input-output-hk/ouroboros-consensus/commit/9cad3f79398039b8ee2e2f631f8bf8cb7fdd09a0

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
ch1bo: Addressed in https://github.com/input-output-hk/ouroboros-consensus/commit/9cad3f79398039b8ee2e2f631f8bf8cb7fdd09a0

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: I just re-ordered these, not sure what you mean.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: Addressed in https://github.com/input-output-hk/ouroboros-consensus/commit/9cad3f79398039b8ee2e2f631f8bf8cb7fdd09a0

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
ch1bo: That's true and I have not thought about it! But as you can see, both are basically an `fmap` over a STM read... so we are fine here.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: Let's do this in another PR if it becomes an issue.

PR #1966, opened 2026-04-08
[Peras 3.75] Overhaul of Peras{Cert,Vote}DB and wiring in ChainDB

### LINKED ISSUE #124: Choose "realistic" boosted blocks in tests
Currently, the function for generating random certificates for test purposes is too broad: we select arbitrary blocks when in practice Peras has strong conditions on which blocks can get certificates.

This isn't a problem so far, because we end up testing stronger proprieties that what will happen in Peras. This may become a problem later, if we choose to rely on specific Peras properties to optimize our code, cf https://github.com/IntersectMBO/ouroboros-consensus/pull/1613

tbagrel1:
This PR is a prerequisite for Peras 4 revival, and introduces a bunch of changes to PerasCertDB and PerasVoteDB  that improve consistency in the API design and testing.

It also plugs PerasVoteDB in ChainDB, and update the associated code + tests.


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/RethrowPolicy.hs
```diff
@@ -97,6 +98,12 @@ consensusRethrowPolicy pb =
     -- for some blocks we used to have but got GCed. This means the
     -- peer is on a chain that forks off more than @k@ blocks away.
     <> mkRethrowPolicy (\_ctx (_ :: BlockFetchServerException) -> distantPeer)
+    <> mkRethrowPolicy
+      ( \_ctx (e :: PerasVoteDbError blk) ->
+          case e of
+            MultipleWinnersInRound{} -> ourBug -- TODO: should we instead shutdown the node?
+            ForgingCertError{} -> ourBug
+      )
```
agustinmista, line 108: This is probably something we would want to try out in a testnet.

Maybe we can create an issue and link it here?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -76,3 +97,107 @@ newtype PerasCertTicketNo = PerasCertTicketNo Word64
...
 zeroPerasCertTicketNo :: PerasCertTicketNo
 zeroPerasCertTicketNo = PerasCertTicketNo 0
+
+data AddPerasCertResult
+  = AddedPerasCertToDB
+  | PerasCertAlreadyInDB
+  deriving stock (Generic, Eq, Ord, Show)
+  deriving anyclass NoThunks
+
+-- * Invariants
```
agustinmista, line 108: Same caveat as with the PerasVoteDB: these invariants require a non-trivial amount of work to be validated in a test suite against random inputs, so, given that we already have state-machine tests, we will focus on that in a subsequent project (Peras -> mainnet).

@tbagrel1 do we have an issue for this somewhere (it could include both the vote and cert DB invariant tests).

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/PerasCertDB/StateMachine.hs
```diff
@@ -124,15 +123,22 @@ instance StateModel Model where
...
-    _ -> model.open
+    action ->
+      model.open && case action of
+        -- Do not add equivocating certificates.
+        AddCert cert -> all p model.certs
+         where
+          p cert' = getPerasCertRound cert /= getPerasCertRound cert' || cert == cert'
+        GetWeightSnapshot -> True
+        GetLatestCertSeen -> True
+        GarbageCollect _roundNo -> True
```
agustinmista: Can we link back the rationale for this? I remember it took some time to figure out a good argument for this, and I think the question might pop up again.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -34,34 +66,23 @@ data PerasCertDB m blk = PerasCertDB
...
+  -- ^ Garbage-collect state strictly older than the given slot number.
+  -- The STM transaction clears the relevant state from the in-memory index, and
+  -- the resulting 'm' action performs tracing and might perform side-effects in
+  -- implementations with on-disk storage.
   --
-  -- FIXME(peras): having seen a certificate is a precondition to start voting
-  -- in every round except for the first one (at origin). As a consequence, only
-  -- caught-up nodes can actively participate in the Peras protocol for now.
-  , garbageCollect :: SlotNo -> m ()
-  -- ^ Garbage-collect state older than the given slot number.
```
agustinmista: I'd like to hear a bit more about this change in the API.

Also, the updated haddock still says "Garbage-collect state strictly older than the given [[[slot]]] number.".

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -95,11 +116,21 @@ makePerasCertPoolWriterFromChainDB ::
...
-        addPerasCerts systemTime (ChainDB.addPerasCertAsync chainDB) certs
+    , opwAddObjects = \certs ->
+        processCerts
+          systemTime
+          -- We do not want to block the writer thread on waiting for ChainSel
+          -- side-effects to complete, so we use the async version of adding
+          -- certs to the ChainDB and ignore the returned promise.
+          -- The async action is still launched and executed behind the scenes
+          -- even though we drop the promise.
+          (ChainDB.getPerasCertIds chainDB)
```
agustinmista: Thanks for adding this comment!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -109,16 +140,28 @@ deriving instance Show PerasCertInboundException
...
-  (StandardHash blk, MonadSTM m) =>
+-- | Process a batch of inbound Peras certificates received from a peer.
+--
+-- Certificates whose round number is already present in the database (as
+-- determined by @alreadyInDbSTM@) are silently skipped. The remaining
+-- certificates are validated; if /any/ certificate in the batch fails
+-- validation, the entire batch is rejected by throwing a
+-- 'PerasCertInboundException'. Otherwise, each valid certificate is
+-- timestamped with the current wall-clock time and added to the database via
+-- @addCert@.
```
agustinmista: > if /any/ certificate in the batch fails validation, the entire batch is rejected by throwing a 'PerasCertInboundException'.

Is this exception being handled anywhere? I think in practice it would mean something that `theyBuggyOrEvil`, so we immediately disconnect from the sender.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -23,62 +25,108 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
...
-makePerasVotePoolWriterFromVoteDB distrVar systemTime PerasVoteDB{..} =
+makePerasVotePoolWriterFromVoteDB systemTime perasVoteDB =
+  ObjectPoolWriter
+    { opwObjectId = getPerasVoteId
+    , opwAddObjects = \votes ->
+        processVotes
+          systemTime
+          (PerasVoteDB.getVoteIds perasVoteDB)
+          (validatePerasVote mkPerasParams (PerasVoteStakeDistr mempty)) -- TODO replace when actual plumbing is in place
+          (void . join . atomically . PerasVoteDB.addVote perasVoteDB)
```
agustinmista, line 112: Would it make sense to also put a disclaimer here about using `void`? I.e., that we don't care about the `AddPerasVoteResult` for now.

In the future, we might need inspect this value and push any generated certificate into the CertDB. Not sure if we want to do that here or elsewhere, but I thought maybe we can link to an issue here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -402,9 +408,32 @@ garbageCollectBlocks CDB{..} slotNo = do
...
+garbageCollectPeras CDB{..} slotNo = do
+  ledger <- atomically $ ledgerState <$> LedgerDB.getVolatileTip cdbLedgerDB
+  let summary = hardForkSummary (configLedger cdbTopLevelConfig) ledger
+  case History.runQuery (History.slotToPerasRoundNo slotNo) summary of
+    Right (History.PerasEnabled (roundNo, _, _)) -> do
+      join . atomically $ PerasCertDB.garbageCollect cdbPerasCertDB roundNo
+      join . atomically $ PerasVoteDB.garbageCollect cdbPerasVoteDB roundNo
+    _ ->
+      -- Peras not enabled or query failed; skip Peras GC
+      pure ()
```
agustinmista: I'm not sure that delegating the responsibility of running the query to the ChainDB is the right approach.

I think it would rather leave that to up the caller, and let `garbageCollectPeras` run the inner GCs unconditionally. 

Which takes me to my second point: who is (or will be) the caller of this function?

agustinmista (commented):
Looking good! I have a couple of comments and concerns, but nothing big.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/RethrowPolicy.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/216

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
tbagrel1: From my understanding, any exception raised inside a mini-protocol thread will bubble up until it reaches the `withPeer` function (from `ouroboros-network`) that is a bracket function. In which case, the exception will make us implicitly disconnect from the faulty/adversarial peer.

A consequence is that exceptions in mini-protocol are typically not used to indicate recoverable failures from our side/logic; unlike how they might be used in the ChainDB for example.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
tbagrel1: Issue created: -- (currently not enforced, see https://github.com/tweag/cardano-peras/issues/217)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
tbagrel1: Indeed, I'll revert back the garbage-collection logic to use slotNo as we agreed on in our meeting :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -35,36 +38,60 @@ import Ouroboros.Consensus.Util.IOLike
...
-                fmap
-                  (vpcCert . forgetArrivalTime)
-                  (takeAscMap (fromIntegral limit) certsAfterLastKnown)
-        pure $
-          if Map.null certsAfterLastKnown
-            then Nothing
-            else Just loadCertsAfterLastKnown
+        certsAfterLastKnownNoLimit <- getCertsAfterSTM lastKnown
+        let certsAfterLastKnown = takeAscMap (fromIntegral limit) certsAfterLastKnownNoLimit
+        if Map.null certsAfterLastKnown
```
jasagredo: Do we need to compute the AscMap before checking if it is null? if the limit is not null and `certsAfterLastKnownNoLimit` is not null this will be not null.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -35,36 +38,60 @@ import Ouroboros.Consensus.Util.IOLike
...
+          then pure Nothing
+          else pure . Just $ do
+            loadedCertsAfterLastKnown <-
+              zip (Map.keys certsAfterLastKnown) <$> sequence (Map.elems certsAfterLastKnown)
+            pure $
+              Map.fromList
+                [ (ticketNo, vpcCert $ forgetArrivalTime cert)
+                | (ticketNo, cert) <- loadedCertsAfterLastKnown
+                ]
```
jasagredo: This looks like a very complicated (and repetitive) way of doing:
```haskell
traverseBind :: (Monad m) => (a -> m b) -> Map k (m a) -> m (Map k b)                                                                                       
traverseBind f = traverse (>>= f) 
```
no?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -95,11 +122,21 @@ makePerasCertPoolWriterFromChainDB ::
...
-    , opwAddObjects = \certs -> do
-        addPerasCerts systemTime (ChainDB.addPerasCertAsync chainDB) certs
+    , opwAddObjects = \certs ->
+        processCerts
+          systemTime
+          -- We do not want to block the writer thread on waiting for ChainSel
+          -- side-effects to complete, so we use the async version of adding
+          -- certs to the ChainDB and ignore the returned promise.
+          -- The async action is still launched and executed behind the scenes
+          -- even though we drop the promise.
```
jasagredo, line 131: I think this should be moved 2 lines below?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -23,62 +25,132 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
...
+-- Writers
+-------------------------------------------------------------------------------
+
+-- | Create a pool writer directly from a 'PerasVoteDB'.
+-- In particular, the result of 'addVote' is ignored, so any produced cert will
+-- have to be handled manually by another mean.
+-- This function is mostly meant for tests against the 'PerasVoteDB' in
+-- isolation; for actual production use, see
+-- 'makePerasVotePoolWriterFromChainDB' which creates a pool writer from the
+-- 'ChainDB' and thus properly handles the produced certs.
```
jasagredo, line 91: Weird line length?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -23,62 +25,132 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
...
 makePerasVotePoolWriterFromVoteDB ::
   (StandardHash blk, IOLike m) =>
-  StrictTVar m PerasVoteStakeDistr ->
   SystemTime m ->
+  -- | This is needed for validating votes (since its during the validation of
```
jasagredo:
```suggestion
  -- | This is needed for validating votes (since it's during the validation of
```
or "it is"

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -23,62 +25,132 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
...
+          (void . join . atomically . PerasVoteDB.addVote perasVoteDB)
+          votes
+    , opwHasObject = do
+        voteIds <- PerasVoteDB.getVoteIds perasVoteDB
+        pure $ \voteId -> Set.member voteId voteIds
+    }
+
+-- | Create a pool writer from the 'ChainDB'. This properly handles the produced
+-- certs by letting the ChainDB take care of them
+-- (see 'ChainDB.addPerasVoteAsync').
```
jasagredo: Weird line length

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -23,62 +25,132 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
...
-              throw (PerasVoteValidationError errs)
-    , opwHasObject = (flip Set.member) <$> getVoteIds
+    , opwAddObjects = \votes ->
+        processVotes
+          systemTime
+          -- We do not want to block the writer thread on waiting for ChainSel
+          -- side-effects to complete, so we use the async version of adding
+          -- votes to the ChainDB and ignore the returned promise.
+          -- The async action (if any) is still launched and executed behind the
+          -- scenes even though we drop the promise.
```
jasagredo: Shouldn't this be moved some lines below, next to `addPerasVoteAsync`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -88,15 +160,36 @@ deriving instance Show PerasVoteInboundException
...
+-- | Process a batch of inbound Peras votes received from a peer.
+--
+-- Votes whose ID is already present in the database (as determined by
+-- @alreadyInDbSTM@) are silently skipped. The remaining votes are validated;
+-- if /any/ vote in the batch fails validation, the entire batch is rejected
+-- by throwing a 'PerasVoteInboundException' (which should make us disconnect
+-- from the distant peer, see 'withPeer' bracket function from
+-- `ouroboros-network`).
+-- Otherwise, each valid vote is timestamped with the current wall-clock time
+-- and added to the database via @addVote@.
```
jasagredo: Weird line length

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -88,15 +160,36 @@ deriving instance Show PerasVoteInboundException
...
+  now <- systemTimeCurrent systemTime
+  case partitionEithers validationResults of
+    -- All votes are valid => add them to the pool
+    ([], validatedVotes) ->
+      mapM_
+        (addVote . WithArrivalTime now)
+        validatedVotes
+    -- Some votes are invalid => reject the whole batch
+    (errs, _) ->
+      throw (PerasVoteValidationError errs)
```
jasagredo, line 192: I wonder if we want to still validate all of them. Perhaps as soon as we have one error we can return:
```
(err:_, _) -> 
  throw (PerasVoteValidationError err)
```
by laziness this should stop as soon as it finds one error

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -110,8 +112,9 @@ launchBgTasks cdb@CDB{..} replayed = do
   gcSchedule <- newGcSchedule
   !gcThread <-
     launch "ChainDB.gcBlocksScheduleRunner" $
```
jasagredo: Should we take this opportunity to rename this thread?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -307,6 +309,24 @@ addPerasCertAsync ::
 addPerasCertAsync CDB{cdbTracer, cdbChainSelQueue} =
   addPerasCertToQueue (TraceAddPerasCertEvent >$< cdbTracer) cdbChainSelQueue
 
+-- | Add a Peras vote to the VoteDB contained in the ChainDB, and if this
+-- results in a new cert being generated, add that cert /asynchronously/ to
+-- the ChainDB as well.
+addPerasVoteAsync ::
```
jasagredo: At first I thought this meant "do not wait for peras vote addition", but vote addition is Sync and then the cert addition is async. I think we should rename this function to something that conveys that

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -307,6 +309,24 @@ addPerasCertAsync ::
...
+  WithArrivalTime (ValidatedPerasVote blk) ->
+  m (Maybe (AddPerasCertPromise m))
+addPerasVoteAsync cdb@CDB{cdbPerasVoteDB} vote = do
+  addVoteRes <- join . atomically . addVote cdbPerasVoteDB $ vote
+  case addVoteRes of
+    AddedPerasVoteAndGeneratedNewCert cert -> do
+      let certTime = getArrivalTime vote
+      Just <$> addPerasCertAsync cdb (WithArrivalTime (certTime) cert)
+    -- No cert to add, so we return a promise that resolves immediately.
+    _ -> pure Nothing
```
jasagredo, line 327: We don't return a promise, we return a Nothing (the comment is wrong)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -34,38 +66,24 @@ data PerasCertDB m blk = PerasCertDB
...
-  , getCertSnapshot :: STM m (PerasCertSnapshot blk)
-  , getLatestCertSeen :: STM m (Maybe (WithArrivalTime (ValidatedPerasCert blk)))
-  -- ^ Get the certificate with the highest round number that has been added to
-  -- the db since it has been opened. This certificate is not affected by garbage
-  -- collection, but it's forgotten when the db is closed.
+  , getLatestCertSeen ::
+      STM m (Maybe (WithArrivalTime (ValidatedPerasCert blk)))
+  , -- FIXME(peras): having seen a certificate is a precondition to start voting
+    -- in every round except for the first one (at origin). As a consequence, only
+    -- caught-up nodes can actively participate in the Peras protocol for now.
```
jasagredo: I don't see what this comment relates to. It seems it belonged to `getLatestCertSeen` but even then I don't see how this general constraint of peras has to be in this particular field?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -80,3 +98,109 @@ newtype PerasCertTicketNo = PerasCertTicketNo Word64
...
 zeroPerasCertTicketNo :: PerasCertTicketNo
 zeroPerasCertTicketNo = PerasCertTicketNo 0
+
+data AddPerasCertResult
+  = AddedPerasCertToDB
+  | PerasCertAlreadyInDB
+  deriving stock (Generic, Eq, Ord, Show)
+  deriving anyclass NoThunks
+
+-- * Invariants
```
jasagredo, line 108: Before Peras we never had properties next to impls, but I guess it is fine.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -18,7 +18,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
-{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-orphans -Wno-unused-local-binds #-}
```
jasagredo, line 21: Why?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -55,307 +129,149 @@ defaultArgs =
...
+  SlotNo ->
+  STM m (m ())
+implGarbageCollect PerasCertDbEnv{pcdbTracer, pcdbState} slotNo = do
   -- No need to update the 'Fingerprint' as we only remove certificates that do
   -- not matter for comparing interesting chains.
-  atomically $ modifyTVar pcdbVolatileState (fmap gc)
+  modifyTVar pcdbState (fmap gc)
+  pure $ do
+    traceWith pcdbTracer (GarbageCollected slotNo)
+    return ()
```
jasagredo: Redundant `do` and `return ()`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
tbagrel1: We don't really need to bind here; but I agree that using `traverse` is a much better idea than what I was previously doing
I went for:
```hs
            traverse
              (\loadCertAction -> (vpcCert . forgetArrivalTime) <$> loadCertAction)
              certsAfterLastKnown
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
tbagrel1: Yes indeed!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
tbagrel1: Collecting them all has been requested by @dnadales and @bladyjoker [here](https://github.com/IntersectMBO/ouroboros-consensus/pull/1768#discussion_r2747873186) :)

tbagrel1: N.B: this seemed to be based on the assumption that it's cheap to validate votes, which might not be the case in practice

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
tbagrel1: This has been directly requested by @bladyjoker for other modules, see [this](https://github.com/IntersectMBO/ouroboros-consensus/pull/1768/changes#r2816376594)

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
tbagrel1: For consistency/exhaustiveness, and ease-of-tweaking in the future, I've defined all possible favours of block generators inside `genBlkPair`. Not all of them may be used currently, but it's quite practical to be able to change them on the fly even if it creates a bit of dead code atm.

If you really want to, I may comment out the unused ones.

tbagrel1:
Thanks @jasagredo for the review, I've just addressed most of your comments ;)

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
jasagredo: Or just prefix the bindings with an underscore

tbagrel1:
@jasagredo I've addressed the local-bind warning thing, and also added a commit [fc74147](https://github.com/IntersectMBO/ouroboros-consensus/pull/1966/commits/fc7414750e91881b8464deef775f759f9764e1f0) to adapt after rebasing on latest `main`.

Is there anything else needed before we can get this approved? Many thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
jasagredo: Let's add a comment mentioning what you said above, for future reference, and keep going

tbagrel1: Done in https://github.com/IntersectMBO/ouroboros-consensus/pull/1966/changes/8d4cdcb0d7d54d944f8cc7e70206b5ec73b88571#diff-7ef0fa8fc64496495f0835f5db0a598fe584a9b558a7d6e16e7a59c4a2e8b993R192

jasagredo (approved):
Thanks for addressing the comments!

PR #1968, opened 2026-04-09
Re-use implementations of commands in ChainDB/Unit.hs

geo2a:
This PR refactors the infrastructure that allows recording sequences of testing commands as explicit test scenarios.

The `SupportsUnitTest` type class is a shallow embedding of the `Cmd` data type from the `Test.Ouroboros.Storage.ChainDB.StateMachine` module. It exists to allow running these commands with both the SUT and the model semantics separately, but only write them once.

The PR makes the instances of `SupportsUnitTest` less awkward by re-using the code that is already written in `Test.Ouroboros.Storage.ChainDB.StateMachine`. Specifically, we call `run` (for the SUT) and `runPure` (for the model) directly on the command, instead of in-lining the corresponding cases from this function. This greatly reduce code duplication and the danger of accidentally a semantic mismatch in the unit tests.

I was motivated to re-factor this infrastructure as I found the unit-tests very useful while debugging test failures in https://github.com/IntersectMBO/ouroboros-consensus/pull/1575.

### DIFF
--- a/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
+++ b/ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
@@ -38,16 +38,11 @@ import Ouroboros.Consensus.Block.RealPoint
   )
 import Ouroboros.Consensus.Config
   ( TopLevelConfig
-  , configSecurityParam
   )
 import Ouroboros.Consensus.Config.SecurityParam (SecurityParam (..))
 import Ouroboros.Consensus.Ledger.Basics
 import Ouroboros.Consensus.Ledger.Extended (ExtLedgerState)
-import Ouroboros.Consensus.Ledger.SupportsProtocol
-  ( LedgerSupportsProtocol
-  )
 import qualified Ouroboros.Consensus.Storage.ChainDB.API as API
-import qualified Ouroboros.Consensus.Storage.ChainDB.API.Types.InvalidBlockPunishment as API
 import Ouroboros.Consensus.Storage.ChainDB.Impl (TraceEvent)
 import Ouroboros.Consensus.Storage.ChainDB.Impl.Args
 import Ouroboros.Consensus.Storage.Common
@@ -65,9 +60,7 @@ import Test.Ouroboros.Storage.ChainDB.StateMachine
   ( AllComponents
   , ChainDBEnv (..)
   , ChainDBState (..)
-  , ShouldGarbageCollect (..)
   , TestConstraints
-  , allComponents
   , close
   , mkTestCfg
   , open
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
```diff
@@ -434,7 +428,9 @@ class SupportsUnitTest m where
           (Maybe (Point (Block m)))
       )
 
-  persistBlks :: ShouldGarbageCollect -> m ()
+  persistBlks :: m ()
```
geo2a, line 430: I've split the `persistBlks` and `persistBlksThenGC` to more closely match the `Cmd` data type.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
```diff
@@ -477,55 +473,68 @@ runModel ::
...
   type IteratorId (ModelM blk) = Model.IteratorId
   type Block (ModelM blk) = blk
 
-  newFollower = withModelContext $ \model _ ->
-    Model.newFollower model
+  newFollower = do
+    result <- runModelCmd (SM.NewFollower API.SelectedChain)
+    case result of
+      SM.Flr fid -> pure fid
+      _ -> failWith "newFollower: unexpected result"
```
jasagredo: I wonder if we could show what we get as a result here, just so that if an error happens we can find it quickly

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
```diff
@@ -622,23 +631,30 @@ withTestChainDbEnv topLevelConfig chunkInfo extLedgerState cont =
...
-    SystemM $ lift $ lift $ do
-      internal <- internal <$> readTVarIO (varDB env)
-      SM.persistBlks shouldGarbageCollect internal
+  persistBlks =
+    void $ runCmd SM.PersistBlks
+
+  persistBlksThenGC =
+    void $ runCmd SM.PersistBlksThenGC
 
   newFollower = do
```
jasagredo, line 659: Shouldn't we also use `runCmd` for other commands below?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
geo2a: Good idea, addressed

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Unit.hs
geo2a: Now sure what happened here, I've added the missing changes now, thanks!

jasagredo (approved):

PR #1970, opened 2026-04-09
Add a benchmark about LeiosDemoDb

### LINKED ISSUE #794: Benchmark SQLite performance using criterion (or similar)
### Why

We want to be able to measure the performance of SQLite-based operations used by the Leios prototype in order to identify database-access-related bottlenecks.

### What

Add a `leios-benchmark` benchmark component to `ouroboros-consensus`, which should run various operations using a `LeiosDbHandle`. The benchmark should use the `criterion` library and should generate randomized data sets to populate the database before running insert / query / filter operations and measuring their performance.

### How

- [ ] Add a `leios-benchmark` cabal component using `criterion`

ch1bo:
This uses a more consistent generation of quantities and resembles the concurrent access similar to the production code.

This is currently quite barebones, but gives us a good first numbers of a reasonably respresentative scenario:

```
λ cabal bench leios-db-bench
Build profile: -w ghc-9.6.6 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-0.27.0.0 (bench:leios-db-bench) (ephemeral targets)
Preprocessing benchmark 'leios-db-bench' for ouroboros-consensus-0.27.0.0...
Building benchmark 'leios-db-bench' for ouroboros-consensus-0.27.0.0...
Running 1 benchmarks...
Benchmark leios-db-bench: RUNNING...
LeiosDemoDb concurrent benchmark

Database setup:
  EBs pre-populated : 500
  TXs per EB        : 200
  Total TXs         : 100000

Concurrent workload per iteration:
  Active reader (×1): 10 rounds of filterMissingEbBodies (200+200) + filterMissingTxs (500+500)
  Writers       (×2): 20 insertEbPoint/insertEbBody/insertTxs each
  Readers       (×8): 30 lookupEbBody + 10 batchRetrieveTxs each

Runs: 1 warmup + 5 timed

Inserting EBs: 50 100 150 200 250 300 350 400 450 500 done
  run 1/5: 3.830633993 s
  run 2/5: 3.696195071 s
  run 3/5: 3.675190096 s
  run 4/5: 3.789160281 s
  run 5/5: 3.890996196 s
  => min=3.675190096 s  avg=3.7764351274 s  max=3.890996196 s
Benchmark leios-db-bench: FINISH
```

### DIFF
--- a/ouroboros-consensus/bench/leios-db-bench/Main.hs
+++ b/ouroboros-consensus/bench/leios-db-bench/Main.hs
@@ -16,7 +16,7 @@
 --   10 'leiosDbBatchRetrieveTxs' calls cycling through the pre-populated EBs.
 --
 -- All data is deterministic (no QuickCheck generators), so runs are stable and
--- comparable before\/after lock-removal refactors.
+-- comparable across refactors.
 --
 -- Usage:
 --
@@ -29,8 +29,8 @@ import Cardano.Slotting.Slot (SlotNo (..))
 import Control.Concurrent.Async (async, mapConcurrently_, wait)
 import Control.Monad (forM, forM_, replicateM_, when)
 import Control.Monad.Class.MonadTime.SI (diffTime, getMonotonicTime)
-import Data.Bits (shiftR, (.&.))
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as BS8
 import Data.IORef (IORef, atomicModifyIORef', newIORef)
 import Data.Time.Clock (DiffTime)
 import qualified Data.Vector as V
@@ -208,7 +208,11 @@ insertOneEb :: LeiosDbHandle IO -> Int -> IO ()
 insertOneEb db ebIdx = do
   let point = genPoint ebIdx
       eb = genEb ebIdx
-      txs = [(genTxHash ebIdx txIdx, fixedTxBytes) | txIdx <- [0 .. txsPerEb - 1]]
+      txs =
+        [ (h, genTx h)
+        | txIdx <- [0 .. txsPerEb - 1]
+        , let h = genTxHash ebIdx txIdx
+        ]
   leiosDbInsertEbPoint db point (leiosEbBytesSize eb)
   leiosDbInsertEbBody db point eb
   _ <- leiosDbInsertTxs db txs
@@ -216,33 +220,15 @@ insertOneEb db ebIdx = do
 
 -- * Deterministic data generation
 
--- | 'EbHash' from a 32-bit index: 4 index bytes followed by 28 zeros.
-genEbHash :: Int -> EbHash
-genEbHash i =
-  MkEbHash $
-    BS.pack
-      [ fromIntegral (i .&. 0xFF)
-      , fromIntegral ((i `shiftR` 8) .&. 0xFF)
-      , fromIntegral ((i `shiftR` 16) .&. 0xFF)
-      , fromIntegral ((i `shiftR` 24) .&. 0xFF)
-      ]
-      <> BS.replicate 28 0
-
--- | 'TxHash' from an EB index + TX offset: 4 bytes followed by 28 ones.
-genTxHash :: Int -> Int -> TxHash
-genTxHash ebIdx txIdx =
-  MkTxHash $
-    BS.pack
-      [ fromIntegral (ebIdx .&. 0xFF)
-      , fromIntegral ((ebIdx `shiftR` 8) .&. 0xFF)
-      , fromIntegral (txIdx .&. 0xFF)
-      , fromIntegral ((txIdx `shiftR` 8) .&. 0xFF)
-      ]
-      <> BS.replicate 28 1
-
 -- | 'LeiosPoint' from an index (SlotNo = index).
 genPoint :: Int -> LeiosPoint
 genPoint i = MkLeiosPoint (SlotNo $ fromIntegral i) (genEbHash i)
+
+-- | 'EbHash' from an index: \"ebHash:<index>\" padded to 32 bytes with zeros.
+genEbHash :: Int -> EbHash
+genEbHash i =
+  let tag = BS8.pack ("ebHash:" <> show i)
+   in MkEbHash $ BS.take 32 (tag <> BS.replicate 32 0)
 
 -- | 'LeiosEb' with 'txsPerEb' transactions (200 bytes each).
 genEb :: Int -> LeiosEb
@@ -251,6 +237,16 @@ genEb ebIdx =
     V.fromList
       [(genTxHash ebIdx txIdx, 200 :: BytesSize) | txIdx <- [0 .. txsPerEb - 1]]
 
--- | Fixed TX payload: 16 KiB zeros (realistic worst-case size).
-fixedTxBytes :: BS.ByteString
-fixedTxBytes = BS.replicate 16_384 0
+-- | 'TxHash' from an EB index + TX offset: \"txHash:<ebIdx>:<txIdx>\" padded
+-- to 32 bytes with zeros.
+--
+-- NOTE: This is taking an EB index as it always generates the worst case of
+-- fully disjunct transaction closures between EBs.
+genTxHash :: Int -> Int -> TxHash
+genTxHash ebIdx txIdx =
+  let tag = BS8.pack ("txHash:" <> show ebIdx <> ":" <> show txIdx)
+   in MkTxHash $ BS.take 32 (tag <> BS.replicate 32 0)
+
+-- | Generate a TX payload: the TX hash bytes padded with zeros to 16 KiB.
+genTx :: TxHash -> BS.ByteString
+genTx (MkTxHash h) = h <> BS.replicate (16_384 - BS.length h) 0
--- a/ouroboros-consensus/bench/leios-db-bench/Main.hs
+++ b/ouroboros-consensus/bench/leios-db-bench/Main.hs
@@ -6,14 +6,15 @@
 --
 -- The following three roles run concurrently against the same SQLite handle:
 --
--- * __Active reader__ (fetch logic loop): 10 rounds of
+-- * __Fetch logic__ (1 thread): 10 rounds of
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+genTxHash :: Int -> Int -> TxHash
+genTxHash ebIdx txIdx =
+  MkTxHash $
+    BS.pack
+      [ fromIntegral (ebIdx .&. 0xFF)
+      , fromIntegral ((ebIdx `shiftR` 8) .&. 0xFF)
+      , fromIntegral (txIdx .&. 0xFF)
+      , fromIntegral ((txIdx `shiftR` 8) .&. 0xFF)
+      ]
+      <> BS.replicate 28 1
```
bladyjoker: What's going on here?

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+
+-- | Concurrent benchmark for 'LeiosDemoDb' mirroring production access patterns.
```
bladyjoker, line 5: Great elevator pitch!

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+txsPerEb :: Int
+txsPerEb = 200
+
+-- | Number of serving-reader threads (represents downstream peers).
+numServingThreads :: Int
+numServingThreads = 8
+
+-- | Timed repetitions (plus one warmup).
+numRuns :: Int
+numRuns = 5
```
bladyjoker, line 97: I suggest just stuffing this in LeiosDbBenchOpts and optparse-applicative the CLI.

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+--
+-- * Writers (2 threads): each inserts 20 fresh EBs with full TX payloads.
+--   Indices are unique per iteration via 'beWriterIdx'.
+--
+-- * Serving readers (8 threads): each does 30 @lookupEbBody@ + 10
+--   @batchRetrieveTxs@ (sampled offsets) cycling through pre-populated EBs.
+benchConcurrentAll :: BenchEnv -> IO ()
+benchConcurrentAll BenchEnv{beDb = db, bePoints = points, beWriterIdx = writerIdxRef} = do
+  -- Allocate 40 fresh EB indices for the writers in this iteration.
+  startIdx <- atomicModifyIORef' writerIdxRef (\n -> (n + 40, n))
```
bladyjoker: Why are we working in IORef and not just good old STM?

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+  writerIdx <- newIORef numPrePopulatedEbs
+  pure $ BenchEnv db points writerIdx
+
+-- * Timing
+
+-- | Warm up once, then time 'numRuns' repetitions, printing each result and a
+-- final min\/avg\/max summary.
+runBench :: IO () -> IO ()
+runBench action = do
+  action -- warmup (not printed)
```
bladyjoker, line 176: Not sure why we're warming up?

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+--   Indices are unique per iteration via 'beWriterIdx'.
+--
+-- * Serving readers (8 threads): each does 30 @lookupEbBody@ + 10
+--   @batchRetrieveTxs@ (sampled offsets) cycling through pre-populated EBs.
+benchConcurrentAll :: BenchEnv -> IO ()
+benchConcurrentAll BenchEnv{beDb = db, bePoints = points, beWriterIdx = writerIdxRef} = do
+  -- Allocate 40 fresh EB indices for the writers in this iteration.
+  startIdx <- atomicModifyIORef' writerIdxRef (\n -> (n + 40, n))
+  let writerRange1 = [startIdx .. startIdx + 19]
+      writerRange2 = [startIdx + 20 .. startIdx + 39]
```
bladyjoker: Why are writers split?

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+--
+-- * Active reader (1 thread): 10 rounds of @filterMissingEbBodies@ on
+--   200 present + 200 absent points, and @filterMissingTxs@ on 500 present +
+--   500 absent hashes.
+--
+-- * Writers (2 threads): each inserts 20 fresh EBs with full TX payloads.
+--   Indices are unique per iteration via 'beWriterIdx'.
+--
+-- * Serving readers (8 threads): each does 30 @lookupEbBody@ + 10
+--   @batchRetrieveTxs@ (sampled offsets) cycling through pre-populated EBs.
```
bladyjoker: I'd extract the 3 threads active reader, writers and serving into their own function to make it super explicit what their role is and to make `benchConcurrentAll` (maybe rename to `benchScenario`) more obvious.

I'd rename the threads into

active reader -> missingWorkTrackerThread
writer -> ebInsertionThread
serving -> ebServingThread

ouroboros-consensus/bench/leios-db-bench/Main.hs
bladyjoker: I think I get it, you need a deterministic source of data and you use an Int index to manage that. Makes sense!

I generally use a lazy list approach `genEb :: [LeiosEb]` which is the same.

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+  MkEbHash $
+    BS.pack
+      [ fromIntegral (i .&. 0xFF)
+      , fromIntegral ((i `shiftR` 8) .&. 0xFF)
+      , fromIntegral ((i `shiftR` 16) .&. 0xFF)
+      , fromIntegral ((i `shiftR` 24) .&. 0xFF)
+      ]
+      <> BS.replicate 28 0
+
+-- | 'TxHash' from an EB index + TX offset: 4 bytes followed by 28 ones.
```
bladyjoker: I think this forms a very special scenario where no transactions are shared between EBs. I'd note that as it's an important deet.

ouroboros-consensus/bench/leios-db-bench/Main.hs
```diff
@@ -0,0 +1,256 @@
...
+  s = realToFrac t :: Double
+
+-- * DB helpers
+
+-- | Insert one complete EB (point + body + all TXs) by index.
+insertOneEb :: LeiosDbHandle IO -> Int -> IO ()
+insertOneEb db ebIdx = do
+  let point = genPoint ebIdx
+      eb = genEb ebIdx
+      txs = [(genTxHash ebIdx txIdx, fixedTxBytes) | txIdx <- [0 .. txsPerEb - 1]]
```
bladyjoker: How about we have `genTx :: TxHash -> ByteString` and then we just repeat the TxHash bytes in the Tx? Just so in the case we look into `bench.db` we can spot the difference between txs.

bladyjoker (approved):
I like it!

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: I like infinite lists. Should probably switch to it here too!

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: Suggest = follow-up PR that you create? :)

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: Only reach for what you need?

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: @nfrisby Did speak about that a bit [here](https://docs.google.com/presentation/d/1IMftAxKZ1gmTZ9J8gF9iumpfTgmcTKAkQeykSxvaTcQ/edit?slide=id.g3d3dd2562f3_10_558#slide=id.g3d3dd2562f3_10_558) and I figured that the :genie: had a good idea here with warming up, so I kept it.

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: Good idea, I even encoded the indices into it now.
image

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: Refactored and renamed things.. not exactly as you said, but to what I prefer. Hope you like it. https://github.com/input-output-hk/ouroboros-consensus/commit/d34475b54a387f058a18e082b058f784f73ee44a

ouroboros-consensus/bench/leios-db-bench/Main.hs
ch1bo: To not trip the unique constraint (which we deliberately not made an `INSERT OR IGNORE` yet, see here for an outdated comment about one situation https://github.com/input-output-hk/ouroboros-consensus/blob/d34475b54a387f058a18e082b058f784f73ee44a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs#L571)

PR #1973, opened 2026-04-13
[Peras 21] Remove out-of-spec WFALS test data

agustinmista:
This PR removes out-of-spec test data generated by the weighted Fait-Accompli prototype written in rust. This prototype accepts target committee sizes beyond the number of pools with strictly positive stake, whereas both our model and implementation will turn such cases into an error.

To keep the conformance testing consistent, rust_results.json now only contains the results up to a target committee size of 2052, corresponding to the number of pools with strictly positive stake in the example weight distribution.


dnadales (commented):
It could be worth adding a comment or assertion somewhere (e.g., in Conformance.hs or alongside the JSON file) documenting why the data stops at 2052, so future contributors regenerating test vectors from the Rust prototype know to apply the same cutoff.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Conformance.hs
```diff
@@ -46,7 +46,12 @@ instance FromJSON RustResult where
       <*> obj .: "persistent"
       <*> obj .: "nonpersistent"
 
--- | Embedded Rust implementation results
+-- | Embedded Rust implementation results.
+--
+-- These are the results of the Rust implementation when applied to
+-- 'exampleStakeDistr' for every valid target committee size, i.e., from 1 to
+-- the maximum number of pools with strictly positive stake, which, in the case
+-- of 'exampleStakeDistr', corresponds to 2052 pools.
```
agustinmista, line 54: @dnadales I added this disclaimer. Hope it helps!

dnadales (approved):

PR #1974, opened 2026-04-13
[Peras 22] Generic voting committee API

### LINKED ISSUE #195: Implement committee selection schemes

### LINKED ISSUE #210: Voting Committee selection logic

### LINKED ISSUE #212: Template Vote and Cert Cryptography

agustinmista:
This PR implements a generic interface that can be used to instantiate multiple voting committee schemes using multiple crypto backends for vote and certificate signing and validation.

The main `CryptoSupportsVotingCommittee` type class allows implementations to define abstract votes and certificate types, check when a node should vote, and forge and verify certificates and votes.

In addition, it provides a wrapper that allows us to keep track of two
simultaneous voting committees, the one corresponding to the current epoch, and the one from the previous one. This is needed to validate votes and certificates arriving slightly late at the beginning of a new epoch.

### DIFF
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -115,6 +115,10 @@ library
     Ouroboros.Consensus.BlockchainTime.WallClock.Simple
     Ouroboros.Consensus.BlockchainTime.WallClock.Types
     Ouroboros.Consensus.BlockchainTime.WallClock.Util
+    Ouroboros.Consensus.Committee.AcrossEpochs
+    Ouroboros.Consensus.Committee.Class
+    Ouroboros.Consensus.Committee.Crypto
+    Ouroboros.Consensus.Committee.Types
     Ouroboros.Consensus.Config
     Ouroboros.Consensus.Config.SecurityParam
     Ouroboros.Consensus.Config.SupportsNode
@@ -372,6 +376,7 @@ library
     monoid-subclasses,
     mtl,
     multiset ^>=0.3,
+    nonempty-containers,
     nothunks ^>=0.2 || ^>=0.3,
     ouroboros-network:{api, api-tests-lib, protocols} ^>=1.1,
     primitive,
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
@@ -1,0 +1,188 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Generic interface used by implementations of voting committees.
+module Ouroboros.Consensus.Committee.Class
+  ( -- * Voting committee interface
+    CryptoSupportsVotingCommittee (..)
+
+    -- * Votes with same target
+  , VotesWithSameTarget
+  , getElectionIdFromVotes
+  , getVoteCandidateFromVotes
+  , getRawVotes
+  , VotesWithSameTargetError (..)
+  , ensureSameTarget
+  ) where
+
+import Data.Containers.NonEmpty (HasNonEmpty (..))
+import Data.Either (partitionEithers)
+import Data.Kind (Type)
+import Data.List.NonEmpty (NonEmpty (..))
+import Ouroboros.Consensus.Committee.Crypto
+  ( CryptoSupportsVoteSigning
+  , ElectionId
+  , PrivateKey
+  , VoteCandidate
+  )
+import Ouroboros.Consensus.Committee.Types (PoolId, VoteWeight)
+
+-- * Voting committee interface
+
+-- | Interface for voting committee schemes.
+--
+-- This class is parametrized by the crypto primitives and the committee
+-- selection data structure. Instances define how to check whether a party
+-- should vote and how to compute the voting weight of a committee member.
+class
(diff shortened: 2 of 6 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -0,0 +1,351 @@
...
+    -- * Vote signing interface
+  , CryptoSupportsVoteSigning (..)
+  , CryptoSupportsAggregateVoteSigning (..)
+
+    -- ** Trivial aggregate vote signature verification helpers
+  , TrivialAggregateVoteVerificationKey (..)
+  , TrivialAggregateVoteSignature (..)
+  , trivialLiftVoteVerificationKey
+  , trivialLiftVoteSignature
+  , trvialVerifyAggregateVoteSignature
```
dnadales:
```suggestion
  , trivialVerifyAggregateVoteSignature
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -0,0 +1,351 @@
...
+  -- | Key used for signing votes
+  type VoteSigningKey crypto :: Type
+
+  -- | Key used for verifying votes
+  type VoteVerificationKey crypto :: Type
+
+  -- | Cryptographic signature of a vote
+  data VoteSignature crypto :: Type
+
+  -- | Derive a signing key from a voting committe private key
```
dnadales: We should committe → committee everywhere 👍

ouroboros-consensus.cabal
```diff
@@ -372,6 +376,7 @@ library
     monoid-subclasses,
     mtl,
     multiset ^>=0.3,
+    nonempty-containers,
```
dnadales, line 379: Shouldn't we version bound this?

dnadales, line 379: Can't we use `NonEmpety` from `base`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -0,0 +1,188 @@
...
+data VotesWithSameTargetError crypto committee
+  = EmptyVotes
+  | TargetMismatch
+      -- First vote and the rest of the votes that match its target
+      (NE [Vote crypto committee])
+      -- Votes that do not match the target of the first vote
+      (NE [Vote crypto committee])
+
+-- | Check that a list of votes all target the same election and candidate
+ensureSameTarget ::
```
dnadales, line 158: Ignore if this will be done later on, but shouldn't we add some test for this function?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -0,0 +1,351 @@
...
+
+trivialLiftVRFOutput ::
+  Proxy crypto ->
+  VRFOutput crypto ->
+  TrivialAggregateVRFOutput crypto
+trivialLiftVRFOutput _ =
+  TrivialAggregateVRFOutput
+    . NonEmpty.singleton
+
+trivialVerifyAggregateVRFOutput ::
```
dnadales, line 323: Shouldn't we add tests for this function?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
agustinmista: Thanks for pointing it out. I added a property to test this function in [Peras 25](https://github.com/IntersectMBO/ouroboros-consensus/blob/01757d11b78a6feffaabcbb2aab0f0c4e8b6d060/ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Class.hs)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
agustinmista: I have now also added property tests for these trivial verification functions in [Peras 25](https://github.com/IntersectMBO/ouroboros-consensus/blob/peras/voting-committee-tests/ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Crypto.hs)

ouroboros-consensus.cabal
agustinmista: We are using not only `NotEmpty` lists (a.k.a. `NE [a]` via [HasNonEmpty](https://hackage-content.haskell.org/package/nonempty-containers-0.3.5.0/docs/Data-Containers-NonEmpty.html#t:HasNonEmpty)), but also non-empty maps (`NE (Map k v)`), and non-empty sets (`NE (Set a)`).

agustinmista: I'd be happy to add a version bound. Is there a concrete policy for this? E.g. `^>=<major>.<minor>`?

dnadales: `^>=<major>.<minor>` should work :+1:

dnadales (approved):

ouroboros-consensus.cabal
jasagredo: Usually we do components until the rest are zero, so for `1.0.0.0` you would do `^>=1`, but for `1.0.1.0` you would do `^>=1.0.1`

PR #1975, opened 2026-04-13
[Peras 23] wFA^LS and EveryoneVotes voting committee implementations

### LINKED ISSUE #195: Implement committee selection schemes

### LINKED ISSUE #210: Voting Committee selection logic

### LINKED ISSUE #212: Template Vote and Cert Cryptography

agustinmista:
This PR implements two concrete voting committee strategies:

1. Weighted Fait-Accompli with Local Sortition (`WFALS`), and
2. A simpler one where every pool with positive stake can vote (`EveryoneVotes`)

While Peras will use `WFALS` at the beginning, `EveryoneVotes` serves as a baseline for future benchmarks, as well as a starting point for Leios shall they end up opting for a simpler approach.

NOTES: 

* `WFALS` depends on two separate sub-componenents:
  + `WFA`: the pure weighted Fait-Accompli seat split algorithm, which relies on an "extended stake distribution" `ExtWFAStakeDistr` (a sorted array of pools with cumulative stake). The latter can be precomputed at the beginning of an epoch and shared across multiple voting committees (e.g. shared between Leios and Peras).
  + `LS`: the Local Sortition check for non-persistent voters, mapping their VRF outputs into seats sampled through a Poisson distribution. 
  
* For simplicity, `EveryoneVotes` reuses the same `ExtWFAStakeDistr` as WFALS. This way, if Leios is instantiated with `EveryoneVotes`, it could still use the same precomputed data as Peras.

* Both voting committee implementations defined their own abstract vote and certificate type and it's up to the clients to marshal between these types and their concrete ones (possibly using the same underlying representation and serialization routines)

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
@@ -1,0 +1,330 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | A simple voting committee where pools with positive stake can vote.
+module Ouroboros.Consensus.Committee.EveryoneVotes
+  ( -- * Voting committee interface
+    EveryoneVotes
+  , VotingCommittee -- VotingCommittee internals are not exported
+  , VotingCommitteeInput (..)
+  , VotingCommitteeError (..)
+  , EligibilityWitness (..)
+  , Vote (..)
+  , Cert (..)
+
+    -- * Metrics about the voting committee composition
+  , candidateSeats
+  , numActiveVoters
+  ) where
+
+import Cardano.Ledger.BaseTypes (NonZero)
+import Cardano.Ledger.BaseTypes.NonZero (NonZero (..), nonZero)
+import Control.Monad.Zip (MonadZip (..))
+import qualified Data.Array as Array
+import Data.Bifunctor (Bifunctor (..))
+import Data.Containers.NonEmpty (HasNonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.Proxy (Proxy (..))
+import Data.Set (Set)
+import qualified Data.Set.NonEmpty as NESet
+import Ouroboros.Consensus.Committee.Class
+  ( CryptoSupportsVotingCommittee (..)
+  , UniqueVotesWithSameTarget
+  , getElectionIdFromVotes
+  , getRawVotes
+  , getVoteCandidateFromVotes
+  )
+import Ouroboros.Consensus.Committee.Crypto
+  ( CryptoSupportsAggregateVoteSigning (..)
+  , CryptoSupportsVoteSigning (..)
+  , ElectionId
+  , PrivateKey
+  , PublicKey
+  , VoteCandidate
+  )
+import Ouroboros.Consensus.Committee.Types
+  ( LedgerStake (..)
+  , PoolId
+  , VoteWeight (..)
+  )
+import Ouroboros.Consensus.Committee.WFA
+  ( ExtWFAStakeDistr (..)
+  , NumPoolsWithPositiveStake (..)
+  , SeatIndex
+  , WFAError
+  , getCandidateIfSeatWithinBounds
+  , unsafeGetCandidateInSeat
+  )
+
+-- | Tag for a simple voting committee where pools with positive stake can vote.
+data EveryoneVotes
+
+instance
+  CryptoSupportsAggregateVoteSigning crypto =>
+  CryptoSupportsVotingCommittee crypto EveryoneVotes
+  where
+  data VotingCommittee crypto EveryoneVotes
+    = EveryoneVotesVotingCommittee
+    { -- Preaccumulated stake distribution used to compute committee composition
+      extWFAStakeDistr :: !(ExtWFAStakeDistr (PublicKey crypto))
+    , -- Index of a given candidate in the cumulative stake distribution
+      candidateSeats :: !(Map PoolId SeatIndex)
+    , -- Number of active voters (i.e., those with non-zero stake)
+      numActiveVoters :: !NumPoolsWithPositiveStake
+    }
+
+  data VotingCommitteeInput crypto EveryoneVotes
+    = EveryoneVotesVotingCommitteeInput
+        -- Extended cumulative stake distribution of the potential voters
+        !(ExtWFAStakeDistr (PublicKey crypto))
+
+  data VotingCommitteeError crypto EveryoneVotes
+    = -- An error occurred during the computation of the committee selection
+      WFAError WFAError
+    | -- Pool ID is missing from the voting committee
+      MissingPoolId PoolId
+    | -- Seat index is out of bounds for the voting committee
+      MissingSeatIndex SeatIndex
+    | -- Pool has no stake and thus is not entitled to vote
+      PoolHasNoStake SeatIndex
+    | -- The vote signature is invalid
+      InvalidVoteSignature String
+    | -- The certificate signature is invalid
+      InvalidCertSignature String
+    | -- We triggered an unexpected cryptographic error
+      CryptoError String
+    deriving (Show, Eq)
+
+  data EligibilityWitness crypto EveryoneVotes
+    = EveryoneVotesMember
+        !SeatIndex
+        !(NonZero LedgerStake)
+
+  data Vote crypto EveryoneVotes
+    = EveryoneVotesVote
+        !SeatIndex
+        !(ElectionId crypto)
+        !(VoteCandidate crypto)
+        !(VoteSignature crypto)
+
+  data Cert crypto EveryoneVotes
+    = EveryoneVotesCert
+        !(ElectionId crypto)
+        !(VoteCandidate crypto)
+        !(NE (Set SeatIndex))
+        !(AggregateVoteSignature crypto)
+
+  mkVotingCommittee = mkEveryoneVotesVotingCommittee
+  checkShouldVote = implCheckShouldVote
+  forgeVote = implForgeVote
+  verifyVote = implVerifyVote
+  eligiblePartyVoteWeight = implEligiblePartyVoteWeight
+  forgeCert = implForgeCert
+  verifyCert = implVerifyCert
+
+-- | Construct a 'EveryoneVotesVotingCommittee' for a given epoch
+mkEveryoneVotesVotingCommittee ::
+  VotingCommitteeInput crypto EveryoneVotes ->
+  Either
+    (VotingCommitteeError crypto EveryoneVotes)
+    (VotingCommittee crypto EveryoneVotes)
+mkEveryoneVotesVotingCommittee
+  ( EveryoneVotesVotingCommitteeInput
+      stakeDistr
+    ) = do
+    let seats =
+          Map.fromList
+            . fmap (\(seatIndex, (poolId, _, _, _)) -> (poolId, seatIndex))
+            . Array.assocs
+            . unExtWFAStakeDistr
+            $ stakeDistr
+
+    pure $
+      EveryoneVotesVotingCommittee
+        { extWFAStakeDistr = stakeDistr
+        , candidateSeats = seats
+        , numActiveVoters = numPoolsWithPositiveStake stakeDistr
+        }
+
+-- | Check whether we should vote in a given election
+implCheckShouldVote ::
+  forall crypto.
+  VotingCommittee crypto EveryoneVotes ->
+  PoolId ->
+  PrivateKey crypto ->
+  ElectionId crypto ->
+  Either
+    (VotingCommitteeError crypto EveryoneVotes)
+    (Maybe (EligibilityWitness crypto EveryoneVotes))
+implCheckShouldVote committee ourId _ourPrivateKey _electionId
+  | Just seatIndex <- Map.lookup ourId (candidateSeats committee) = do
+      let (_, _, ourStake, _) =
+            unsafeGetCandidateInSeat seatIndex (extWFAStakeDistr committee)
+      case nonZero ourStake of
+        Nothing ->
+          Right $
+            Nothing
+        Just nonZeroOurStake ->
+          Right $
+            Just $
+              EveryoneVotesMember
+                seatIndex
+                nonZeroOurStake
+  | otherwise =
+      Left (MissingPoolId ourId)
+
+-- | Forge a vote for a given election and candidate
+implForgeVote ::
+  forall crypto.
+  CryptoSupportsVoteSigning crypto =>
+  EligibilityWitness crypto EveryoneVotes ->
+  PrivateKey crypto ->
+  ElectionId crypto ->
+  VoteCandidate crypto ->
+  Vote crypto EveryoneVotes
+implForgeVote member ourPrivateKey electionId candidate =
+  EveryoneVotesVote seatIndex electionId candidate sig
+ where
+  EveryoneVotesMember seatIndex _ =
+    member
+  ourVoteSigningKey =
+    getVoteSigningKey (Proxy @crypto) ourPrivateKey
+  sig =
+    signVote ourVoteSigningKey electionId candidate
+
+-- | Verify a vote cast by a committee member in a given election
+implVerifyVote ::
+  forall crypto.
+  CryptoSupportsVoteSigning crypto =>
+  VotingCommittee crypto EveryoneVotes ->
+  Vote crypto EveryoneVotes ->
+  Either
+    (VotingCommitteeError crypto EveryoneVotes)
+    (EligibilityWitness crypto EveryoneVotes)
+implVerifyVote committee = \case
+  EveryoneVotesVote seatIndex electionId candidate sig
+    | Just (_, voterPublicKey, voterStake, _) <-
+        getCandidateIfSeatWithinBounds seatIndex (extWFAStakeDistr committee) -> do
+        let voterVerificationKey =
+              getVoteVerificationKey (Proxy @crypto) voterPublicKey
+        bimap InvalidVoteSignature id $ do
+          verifyVoteSignature
+            voterVerificationKey
+            electionId
+            candidate
+            sig
+        case nonZero voterStake of
+          Nothing ->
+            Left (PoolHasNoStake seatIndex)
+          Just nonZeroVoterStake ->
+            pure $
+              EveryoneVotesMember
+                seatIndex
+                nonZeroVoterStake
+    | otherwise ->
+        Left (MissingSeatIndex seatIndex)
+
+-- | Compute the voting power of an eligible committee member.
+--
+-- In this simple voting committee, the vote weight of a member is equal to
+-- their ledger stake, as long as it is positive.
+implEligiblePartyVoteWeight ::
+  VotingCommittee crypto EveryoneVotes ->
+  EligibilityWitness crypto EveryoneVotes ->
+  VoteWeight
+implEligiblePartyVoteWeight _committee member =
+  VoteWeight (unLedgerStake (unNonZero voterStake))
+ where
+  EveryoneVotesMember _ voterStake = member
+
+-- | Forge a certificate attesting the winner of a given election
+implForgeCert ::
+  forall crypto.
+  CryptoSupportsAggregateVoteSigning crypto =>
+  UniqueVotesWithSameTarget crypto EveryoneVotes ->
+  Either
+    (VotingCommitteeError crypto EveryoneVotes)
+    (Cert crypto EveryoneVotes)
+implForgeCert votes = do
+  aggSig <-
+    bimap CryptoError id $ do
+      aggregateVoteSignatures
+        (Proxy @crypto)
+        voteSignatures
+  pure $
+    EveryoneVotesCert
+      (getElectionIdFromVotes votes)
+      (getVoteCandidateFromVotes votes)
+      (NESet.fromList voters)
+      aggSig
+ where
+  (voters, voteSignatures) =
+    munzip $ flip fmap votesInAscendingSeatIndexOrder $ \case
+      EveryoneVotesVote seatIndex _ _ sig ->
+        ( seatIndex
+        , sig
+        )
+
+  -- Make sure we have votes in ascending seat index order, which is something
+  -- 'VotesWithSameTarget' cannot guarantee by itself, since seat indices are
+  -- an implementation detail of this voting committee scheme.
+  votesInAscendingSeatIndexOrder =
+    flip NonEmpty.sortWith (getRawVotes votes) $ \case
+      EveryoneVotesVote seatIndex _ _ _ -> seatIndex
+
+-- | Verify a certificate attesting the winner of a given election
+implVerifyCert ::
+  forall crypto.
+  CryptoSupportsAggregateVoteSigning crypto =>
+  VotingCommittee crypto EveryoneVotes ->
+  Cert crypto EveryoneVotes ->
+  Either
+    (VotingCommitteeError crypto EveryoneVotes)
+    (NE [EligibilityWitness crypto EveryoneVotes])
+implVerifyCert committee = \case
+  EveryoneVotesCert electionId candidate voters aggSig -> do
+    -- Traverse the list of voters in ascending seat index order, collecting:
+    -- 1. their membership status
+    -- 2. their vote verification keys (to verify the aggregate vote signature)
+    (members, voteVerificationKeys) <-
+      fmap munzip . flip traverse (NESet.toAscList voters) $ \case
+        seatIndex
+          | Just (_, voterPublicKey, voterStake, _) <-
+              getCandidateIfSeatWithinBounds seatIndex (extWFAStakeDistr committee) -> do
+              let voterVerificationKey =
+                    getVoteVerificationKey (Proxy @crypto) voterPublicKey
+              case nonZero voterStake of
+                Nothing ->
+                  Left (PoolHasNoStake seatIndex)
+                Just nonZeroVoterStake ->
+                  pure
+                    ( EveryoneVotesMember
+                        seatIndex
+                        nonZeroVoterStake
+                    , voterVerificationKey
+                    )
+          | otherwise ->
+              Left (MissingSeatIndex seatIndex)
+    -- Verify aggregate signature
+    aggVerificationKey <-
+      bimap CryptoError id $ do
+        aggregateVoteVerificationKeys
+          (Proxy @crypto)
+          voteVerificationKeys
+    bimap InvalidCertSignature id $
+      verifyAggregateVoteSignature
+        (Proxy @crypto)
+        aggVerificationKey
+        electionId
+        candidate
+        aggSig
+
+    -- Return the list of voters attesting the election winner
+    pure members
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
@@ -1,0 +1,177 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Local sortition used by non-persistent members of the voting committee.
+--
+-- Implements the @LS@ component of the @wFA^LS@ scheme from the Fait-Accompli
+-- Committee Selection paper: (https://eprint.iacr.org/2023/1273.pdf, §2.3).
+-- See also:
+--  https://github.com/input-output-hk/ouroboros-leios/blob/c5658913221a7f58063bc4f82efaec0900e53dab/post-cip/weighted-fait-accompli.pdf
+module Ouroboros.Consensus.Committee.LS
+  ( -- * Local sortition check
+    LocalSortitionNumSeats (..)
+  , localSortitionNumSeats
+  ) where
+
+import Cardano.Ledger.BaseTypes (FixedPoint, HasZero)
+import Data.Maybe (fromMaybe)
+import Data.Word (Word64)
+import Ouroboros.Consensus.Committee.Crypto (NormalizedVRFOutput (..))
+import Ouroboros.Consensus.Committee.Types (Cumulative (..), LedgerStake (..))
+import Ouroboros.Consensus.Committee.WFA
+  ( NonPersistentCommitteeSize (..)
+  , TotalNonPersistentStake (..)
+  )
+
+-- * Local sortition check
+
+-- | Number of non-persistent seats granted by local sortition to a voter
+newtype LocalSortitionNumSeats = LocalSortitionNumSeats
+  { unLocalSortitionNumSeats :: Word64
+  }
+  deriving stock (Show, Eq, Ord)
+  deriving newtype (Num, HasZero)
+
+-- | Compute how many non-persistent seats can be granted by local sortition to
+-- a voter given their normalized VRF output and stake
+localSortitionNumSeats ::
+  -- | Expected number of non-persistent voters in the committee
+  NonPersistentCommitteeSize ->
+  -- | Total stake of non-persistent voters
+  TotalNonPersistentStake ->
+  -- | Stake of the voter
+  LedgerStake ->
+  -- | Normalized VRF output from the participant
+  NormalizedVRFOutput ->
+  LocalSortitionNumSeats
+localSortitionNumSeats
+  (NonPersistentCommitteeSize numNonPersistentVoters)
+  (TotalNonPersistentStake (Cumulative (LedgerStake totalNonPersistentStake)))
+  (LedgerStake voterStake)
+  (NormalizedVRFOutput normalizedVRFOutput)
+    -- None of the non-persistent voters have any stake => nobody gets a seat.
+    -- NOTE: this check avoids the expensive computation below and also prevents
+    -- division by zero when computing @orders@.
+    | voterStake <= 0 = LocalSortitionNumSeats 0
+    -- If the voter has stake close to zero, the conversion from 'Rational' to
+    -- 'FixedPoint' for 'lambda' might underflow to zero, which would cause the
+    -- "orders" computation below to divide by zero.
+    | lambda <= 0 = LocalSortitionNumSeats 0
+    -- This voter might be entitled to some seats => run the local sortition.
+    | otherwise = LocalSortitionNumSeats (fromIntegral expectedSeats)
+   where
+    -- Expected number of seats granted by local sortition
+    lambda :: FixedPoint
+    lambda =
+      fromRational $
+        fromIntegral numNonPersistentVoters
+          * voterStake
+          / totalNonPersistentStake
+
+    -- Compute the "orders" of the Poisson distribution with parameter lambda,
+    -- which are used as thresholds to determine how many seats we get based on
+    -- the normalized VRF output
+    orders :: [FixedPoint]
+    orders =
+      (fromRational normalizedVRFOutput / lambda)
+        : zipWith
+          (\k prev -> k * prev / lambda)
+          [2 ..]
+          orders
+
+    -- Estimate how many seats we get by comparing the normalized VRF output
+    -- against the thresholds defined by the orders.
+    --
+    -- TODO(peras): evaluate whether the limit used below (3) makes sense in
+    -- this context. One possible starting point would be to understand why
+    -- @checkLeaderNatValue@ (in Ledger) also uses 3 as its own limit when
+    -- computing slot leadership proofs.
+    --
+    -- Tracked by this issue:
+    -- https://github.com/tweag/cardano-peras/issues/234
+    expectedSeats :: Int
+    expectedSeats =
+      fromMaybe 0 $
+        taylorExpCmpFirstNonLower
+          3
+          orders
+          (-lambda)
+
+-------------------------------------------------------------------------------
+-- Helpers vendored from:
+-- https://github.com/cardano-scaling/leios-wfa-ls-demo/blob/7bbd846d9765191ca83b58477dc1596f64ac80fd/leios-wfa-ls-demo/lib/Cardano/Leios/NonIntegral.hs#L227
+--
+-- TODO: merge these into @Cardano.Ledger.NonIntegral@ in @cardano-ledger@
+
+data Step a
+  = Stop
+  | -- Here we have `Below n err acc divisor`
+    Below Int a a a
+
+-- Returns the index of the first element that is NOT certainly BELOW.
+-- It evaluates cmps left-to-right, reusing the Taylor-expansion state
+-- (acc/err/divisor/n) across elements so we don't redo work.
+--
+-- Behavior:
+--   * If cmp_i is proven ABOVE -> return i
+--   * If max iterations reached while testing cmp_i -> return i
+--   * If every element is proven BELOW -> returns Nothing
+--
+-- IMPORTANT: boundX must be e^{|x|} for correct error bounds (see taylorExpCmp).
+taylorExpCmpFirstNonLower ::
+  forall a.
+  RealFrac a =>
+  -- | boundX = e^{|x|} for correct error estimation
+  a ->
+  -- | list of cmp thresholds (checked in order)
+  [a] ->
+  -- | x in e^x
+  a ->
+  Maybe Int
+taylorExpCmpFirstNonLower boundX cmps x =
+  goList 1000 0 x 1 1 0 cmps
+ where
+  -- Traverse the list of cmps, advancing the Taylor state as needed while
+  -- checking if the current cmp is ABOVE or BELOW. If ABOVE, return the index.
+  goList ::
+    Int -> -- maxN
+    Int -> -- n
+    a -> -- err
+    a -> -- acc
+    a -> -- divisor
+    Int -> -- current index
+    [a] -> -- remaining cmps
+    Maybe Int
+  goList _ _ _ _ _ _ [] = Nothing
+  goList maxN n err acc divisor i (cmp : rest) =
+    case decideOne maxN n err acc divisor cmp of
+      Stop ->
+        Just i
+      Below n' err' acc' divisor' ->
+        goList maxN n' err' acc' divisor' (i + 1) rest
+
+  -- Decide current cmp by advancing the shared Taylor state as needed.
+  -- If BELOW is established, returns the *advanced* state to continue with.
+  -- If ABOVE is established or maxN reached, returns Stop.
+  decideOne ::
+    Int -> -- maxN
+    Int -> -- n
+    a -> -- err
+    a -> -- acc
+    a -> -- divisor
+    a -> -- cmp
+    Step a
+  decideOne maxN n err acc divisor cmp
+    | maxN == n = Stop
+    | cmp >= acc' + errorTerm = Stop
+    | cmp < acc' - errorTerm = Below (n + 1) err' acc' divisor'
+    | otherwise = decideOne maxN (n + 1) err' acc' divisor' cmp
+   where
+    divisor' = divisor + 1
+    nextX = err
+    acc' = acc + nextX
+    err' = (err * x) / divisor'
+    errorTerm = abs (err' * boundX)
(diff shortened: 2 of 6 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+-- distribution across multiple committee selection instances derived from the
+-- same underlying stake distribution (e.g. Leios and Peras voting committees
+-- for the same epoch).
+data ExtWFAStakeDistr a
+  = ExtWFAStakeDistr
+  { unExtWFAStakeDistr ::
+      Array
+        SeatIndex
+        ( PoolId -- Voter ID of this voter
+        , a -- Extra payload associated to this voter
```
dnadales, line 351: Could you give an example here (eg `pubkey` I guess?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+        -- NOTE: because the check should behave monotonically, we can skip it
+        -- entirely after the first non-persistent voter is found.
+        checkPersistentSeatThreshold
+          && isAbovePersistentSeatThreshold
+            totalSeats
+            currSeatIndex
+            (LedgerStake voterStake)
+            cumulativeStake
+
+-- | Evaluate whether a voter with its give stake and relatile position in the
```
dnadales:
```suggestion
-- | Evaluate whether a voter with the given stake and relative position in the
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+  forall crypto.
+  CryptoSupportsVRF crypto =>
+  VRFPoolContext crypto ->
+  ElectionId crypto ->
+  VotingCommittee crypto WFALS ->
+  Either
+    (VotingCommitteeError crypto WFALS)
+    (VRFOutput crypto)
+checkVRFOutput context electionId committee =
+  bimap InvalidVoteSignature id $ do
```
dnadales: `evalVRF` errors come from the VRF layer (invalid proof, signing-key issues), not from vote-signature verification — reporting them as `InvalidVoteSignature` is misleading and also leaves the purpose-built `LocalSortitionError String` constructor (declared at L164) as dead code.

Suggest swapping to `bimap LocalSortitionError id`. Propagates to both call sites (`implCheckShouldVote` L281 and `implVerifyVote` L361) and makes the existing error constructor actually reachable.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+implForgeCert ::
+  forall crypto.
+  CryptoSupportsAggregateVoteSigning crypto =>
+  VotesWithSameTarget crypto WFALS ->
+  Cert crypto WFALS
+implForgeCert votes = do
+  WFALSCert
+    (getElectionIdFromVotes votes)
+    (getVoteCandidateFromVotes votes)
+    (NEMap.fromAscList voters)
```
dnadales: `NEMap.fromAscList voters` silently deduplicates entries with the same `SeatIndex`, while `sconcat voteSignatures` on the line below keeps every signature. If any two elements of `voters` share a seat index, the resulting cert has fewer seat entries than signatures; downstream, `trivialVerifyAggregateVoteSignature` (`Committee/Crypto.hs` L181–183) rejects it with an opaque "number of keys and signatures do not match" string.

The root cause sits at the class level: `VotesWithSameTarget` (`Committee/Class.hs` L119–123) guarantees shared election and candidate but makes no claim about voter distinctness, and `ensureSameTarget` does not enforce any. Byzantine voters can legitimately produce same-seat-index votes with different signatures / VRF outputs, so this is not purely a caller-hygiene issue.

Three options; worth choosing one:
- (a) Strengthen `VotesWithSameTarget` to carry a distinctness guarantee on a per-instance voter-identity key (enforced in `ensureSameTarget`).
- (b) Change `forgeCert`'s return to `Either (VotingCommitteeError …) (Cert …)` so duplicates can be rejected with a dedicated error constructor.
- (c) Document at the class level that `forgeCert` callers must pre-deduplicate — band-aid, weakest of the three.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -0,0 +1,371 @@
...
+implForgeCert ::
+  forall crypto.
+  CryptoSupportsAggregateVoteSigning crypto =>
+  VotesWithSameTarget crypto EveryoneVotes ->
+  Cert crypto EveryoneVotes
+implForgeCert votes = do
+  EveryoneVotesCert
+    (getElectionIdFromVotes votes)
+    (getVoteCandidateFromVotes votes)
+    (NESet.fromList voters)
```
dnadales: Same class-level concern as noted on [`WFALS.hs`](https://github.com/IntersectMBO/ouroboros-consensus/pull/1975#discussion_r3131282442) — `NESet.fromList` here silently dedupes seat indices while `sconcat` below retains every signature. The fix is structural, at `VotesWithSameTarget` in `Committee/Class.hs`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,167 @@
...
+  (NonPersistentCommitteeSize numNonPersistentVoters)
+  (TotalNonPersistentStake (Cumulative (LedgerStake totalNonPersistentStake)))
+  (LedgerStake voterStake)
+  (NormalizedVRFOutput normalizedVRFOutput)
+    -- None of the non-persistent voters have any stake => nobody gets a seat.
+    -- NOTE: this check also exists to prevent division by zero below.
+    | totalNonPersistentStake <= 0 = LocalSortitionNumSeats 0
+    -- This voter has no stake (but some others do) => it does not get any seat.
+    -- NOTE: this is an optimization to avoid the expensive computation below.
+    | voterStake <= 0 = LocalSortitionNumSeats 0
```
dnadales, line 58: This guard protects against the integer-zero-stake case, but not against `lambda` **rounding to zero** after the `Rational → FixedPoint` conversion on L62–65. Because `voterStake` and `totalNonPersistentStake` are `Rational` (via `LedgerStake`), a voter with positive but very small stake relative to the total can produce a `Rational λ` below `FixedPoint`'s precision (`E34` in `cardano-ledger-core`), which then rounds to exactly zero. The subsequent division `normalizedVRFOutput / lambda` at L72 (and the `zipWith` below it) would then divide by zero.

On Cardano mainnet this is currently not reachable — `voterStake/totalNonPersistentStake ≥ 1 / (3.6 × 10^16)` ≈ `3 × 10^-17`, safely above `10^-34` — but the code does not document that dependency. If `LedgerStake`'s representation ever changes, or if tests synthesize extreme distributions, the guard is silently gone.

Suggested: add an explicit post-conversion guard alongside the existing two:
```haskell
| lambda <= 0 = LocalSortitionNumSeats 0
```
and a short comment next to it explaining the FixedPoint-precision dependency.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,167 @@
...
+    --
+    -- TODO(peras): evaluate whether the limit used below (3) makes sense in
+    -- this context. One possible starting point would be to understand why
+    -- @checkLeaderNatValue@ (in Ledger) also uses 3 as its own limit when
+    -- computing slot leadership proofs.
+    expectedSeats :: Int
+    expectedSeats =
+      fromMaybe 0 $
+        taylorExpCmpFirstNonLower
+          3
```
dnadales, line 99: This hardcoded `3` is suspicious beyond the TODO comment. The helper's own docstring (L113) says `boundX must be e^{|x|}` for correct error estimation. Here `x = -lambda`, so the contract requires `boundX ≥ e^λ`.

`λ = numNonPersistentVoters × voterStake / totalNonPersistentStake` is not bounded by any small constant — it scales with a voter's stake share, so a pool holding ~3× the average non-persistent stake share already gives `λ ≈ 3`, and `e^λ ≈ 20`. `3` would then violate the helper's precondition.

The reference to `checkLeaderNatValue` in ledger is misleading: there, `x` is bounded by the active-slot probability `f` (≈ 5%), so `|x| ≤ 0.05 ≪ ln 3` and `3` is a safe over-approximation. Here `λ` has no equivalent small bound, so the same constant may no longer be correct.

Suggested resolutions, in increasing order of rigor:
- Derive an application-specific upper bound on `λ` from committee parameters and pass that; OR
- Cap `λ` at a known-safe value and document the cap; OR
- Compute the required bound dynamically at call time.

At minimum, the TODO should not outlive this module — whatever `boundX` ends up being, it must have a derivation in the comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+      TotalNonPersistentStake (Cumulative (LedgerStake nonPersistentStake)) =
+        totalNonPersistentStake committee
+
+-- | Forge a certificate attesting the winner of a given election
+implForgeCert ::
+  forall crypto.
+  CryptoSupportsAggregateVoteSigning crypto =>
+  VotesWithSameTarget crypto WFALS ->
+  Cert crypto WFALS
+implForgeCert votes = do
```
dnadales, line 446: Worth noting that the certificate path (`forgeCert` → `verifyCert`) remains untested even with PR #1977 landing alongside this one. PR #1977's test suite covers:
- vote-level round trip (`checkShouldVote` → `forgeVote` → `verifyVote`),
- fake-vote rejection,
- WFA invariants via a Haskell model + Rust conformance,
- generic helpers (`VotesWithSameTarget`, aggregate verification).

But `forgeCert`/`verifyCert` only has mock implementations in `Test/Consensus/Committee/Class.hs`:
```haskell
forgeCert _ = MockCert
verifyCert _ _ = Right (MockWitness :| [])
```

The certificate is the protocol artifact that actually reaches the wire, and exercises logic not covered by vote-level tests — aggregate signature construction, VRF-output threading for non-persistent voters in `implVerifyCert` (L536–551), ordering/merge via `NEMap.fromAscList`/`sconcat`, and the dedup hazard raised on the nearby L431 thread. Worth adding at least one round-trip property (`forgeVote × many → forgeCert → verifyCert == Right witnesses`) to #1977, ideally with a generator that can produce duplicate seat indices so the dedup concern becomes observable.

Same gap applies to `EveryoneVotes.implForgeCert`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -0,0 +1,371 @@
...
+  data VotingCommittee crypto EveryoneVotes
+    = EveryoneVotesVotingCommittee
+    { -- Preaccumulated stake distribution used to compute committee composition
+      extWFAStakeDistr :: !(ExtWFAStakeDistr (PublicKey crypto))
+    , -- Index of a given candidate in the cumulative stake distribution
+      candidateSeats :: !(Map PoolId SeatIndex)
+    , -- Number of active voters (i.e., those with non-zero stake)
+      numActiveVoters :: !NumPoolsWithPositiveStake
+    , --  Epoch nonce of the epoch where this committee selection takes place
+      epochNonce :: !Nonce
```
dnadales: `epochNonce` is declared and populated (from `VotingCommitteeInput` at L171) but never read anywhere in this module. In `WFALS` the nonce is consumed by `mkVRFElectionInput` inside `checkVRFOutput`; `EveryoneVotes` has no VRF path, so nothing reads it.

Either:
- Drop the field (and the `Nonce` parameter in `EveryoneVotesVotingCommitteeInput` at L92–93) if it was copy-pasted from `WFALS`, or
- Add a short comment explaining why it is carried around unused (e.g., "kept for API symmetry with `WFALS`" or "anticipated future use for X").

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -0,0 +1,371 @@
...
+    (VotingCommitteeError crypto EveryoneVotes)
+    (Maybe (EligibilityWitness crypto EveryoneVotes))
+implCheckShouldVote committee ourId _ourPrivateKey _electionId
+  | Just seatIndex <- Map.lookup ourId (candidateSeats committee) =
+      assert (seatIndexWithinBounds seatIndex (extWFAStakeDistr committee)) $ do
+        let (_, _, ourStake, _) =
+              getCandidateInSeat seatIndex (extWFAStakeDistr committee)
+        case nonZero ourStake of
+          Nothing ->
+            Left (PoolHasNoStake seatIndex)
```
dnadales: Returning `Left` here treats a zero-stake pool as an error, but "my own pool has no stake this epoch" is a legitimate ledger state, not a protocol violation. The analogous branch in `WFALS.implCheckShouldVote` (L288–290) returns `Right Nothing` — "not eligible this election". That asymmetry will force any consumer that handles both instances uniformly to translate, and may log spurious errors for harmless conditions.

On the verify side (L239, L314) `PoolHasNoStake` as an error is correct — someone else's vote from a zero-stake pool is a protocol violation. Only the `checkShouldVote` direction is affected.

Suggested:
```haskell
case nonZero ourStake of
  Nothing -> pure Nothing
  Just nz -> pure (Just (EveryoneVotesMember seatIndex nz))
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,167 @@
...
+          orders
+          (-lambda)
+
+-------------------------------------------------------------------------------
+-- Helpers vendored from:
+-- https://github.com/cardano-scaling/leios-wfa-ls-demo/blob/7bbd846d9765191ca83b58477dc1596f64ac80fd/leios-wfa-ls-demo/lib/Cardano/Leios/NonIntegral.hs#L227
```
dnadales, line 105: This is consensus-critical code vendored as a copy. Two things worth locking in before production:

**(1) Golden-output test.** A small table of `(lambda, normalizedVRFOutput, expectedSeats)` triples pinned via `testCase` would catch behavioural drift from any future edit to this helper — whether our local edit or a sync from upstream. Without it, a subtle numerical change can pass code review and silently cause a consensus split between nodes on different versions.

**(2) Invariants comment.** The block comment at L104–113 describes *what* the helper returns but not *what the Taylor-expansion state must satisfy* (error-bound correctness, the relationship between `acc`, `err`, and `divisor` after each iteration, why the `maxN=1000` cap is safe, etc.). Spelling those out makes it possible to sanity-check edits without re-deriving the algorithm.

The outstanding TODO to upstream into `Cardano.Ledger.NonIntegral` is also worth tracking as a release-blocker issue rather than a silent source-level TODO — once upstream lands, the copy here must be removed in the same release to avoid a window where different nodes can run different versions of this code.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+-- but (under some hypothetical parameterization) only the one with seat index 2
+-- can be granted a persistent seat according to the weighted Fait-Accompli
+-- scheme. Then, the job of this tiebreaker is to ensure that the seat index 2
+-- is fairly distributed among the pools with the same stake.
+--
+-- One possible implementation of this tiebreaker is to sort the pools with the
+-- same stake according to the hash of the epoch nonce and the pool ID. This way
+-- the tiebreaker would be deterministic and resilient to manipulation since an
+-- adversary would not be able to predict the epoch nonce in advance.
+newtype WFATiebreaker
```
dnadales, line 272: The docstring above describes a clear security contract ("deterministic and resilient to manipulation … adversary would not be able to predict the epoch nonce in advance"), but nothing enforces it. The newtype wraps an arbitrary `PoolId -> PoolId -> Ordering`, so a caller can pass:
- `WFATiebreaker compare` — deterministic and total, but predictable (no nonce dependency).
- A function returning `EQ` unconditionally — sort becomes stable-but-arbitrary.

With all callers internal to consensus code, the risk is moderated but non-zero — chiefly "someone reaches for a test-grade tiebreaker when wiring production" (the test suite in #1977 already uses `WFATiebreaker compare`).

Two cheap mitigations worth considering:

**(1) Provide the canonical tiebreaker in this module.** Keep the newtype as-is (preserves flexibility for tests and alternate schemes) but add:
```haskell
-- | The standard tiebreaker: pools with equal stake are ordered by
-- @hash(nonce, poolId)@. Deterministic and adversary-resistant because
-- the epoch nonce cannot be predicted before finalization.
--
-- Production consumers should use this unless they have specific
-- requirements. Tests and experimental schemes can still construct
-- @WFATiebreaker@ directly.
standardWFATiebreaker :: Nonce -> WFATiebreaker
standardWFATiebreaker nonce = WFATiebreaker $ \a b ->
  compare (hashWithSerialiser (nonce, a)) (hashWithSerialiser (nonce, b))
```
Gives Peras/Leios consumers "one obvious thing to use" and makes "why did not you use the standard one?" a natural review question.

**(2) Plug a small test gap.** PR #1977's `prop_stakeDistrDecreasingStakes` covers sort order but does not specifically exercise the tied-stake branch where the tiebreaker actually runs. A property test that generates pools with equal stakes and checks the output order matches a synthetic tiebreaker would pin the behavior of `mkExtWFAStakeDistr` against the tiebreaker contract.

Neither addresses adversary-resistance of a concrete tiebreaker (that is a crypto-primitive property, not a unit-test property), but together they reduce the accidental-misuse surface without giving up the newtype's flexibility.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,167 @@
...
+localSortitionNumSeats
+  (NonPersistentCommitteeSize numNonPersistentVoters)
+  (TotalNonPersistentStake (Cumulative (LedgerStake totalNonPersistentStake)))
+  (LedgerStake voterStake)
+  (NormalizedVRFOutput normalizedVRFOutput)
+    -- None of the non-persistent voters have any stake => nobody gets a seat.
+    -- NOTE: this check also exists to prevent division by zero below.
+    | totalNonPersistentStake <= 0 = LocalSortitionNumSeats 0
+    -- This voter has no stake (but some others do) => it does not get any seat.
+    -- NOTE: this is an optimization to avoid the expensive computation below.
```
dnadales: This guard is **required for correctness**, not an optimization. Without it, `lambda` at L62–65 computes to `0`, and then L72 (`fromRational normalizedVRFOutput / lambda`) divides by zero.

Contrast with the guard just above at L51, which correctly documents itself as defensive:
```haskell
-- NOTE: this check also exists to prevent division by zero below.
| totalNonPersistentStake <= 0 = ...
```

Suggested: align the comment here, e.g. `"required to avoid division by zero at L72 — voterStake = 0 gives lambda = 0"`. Otherwise a future reader may remove the guard during a cleanup pass.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+  forall crypto.
+  CryptoSupportsVRF crypto =>
+  VotingCommittee crypto WFALS ->
+  PoolId ->
+  PrivateKey crypto ->
+  ElectionId crypto ->
+  Either
+    (VotingCommitteeError crypto WFALS)
+    (Maybe (EligibilityWitness crypto WFALS))
+implCheckShouldVote committee ourId ourPrivateKey electionId
```
dnadales, line 262: The persistent-vs-non-persistent case-split is duplicated across three functions in this module: `implCheckShouldVote` (here), `implVerifyVote` (L334–379), and inside the `traverse` in `implVerifyCert` (L474–533). All three do: bounds check → `getCandidateInSeat` → branch on `isPersistentMember` → build the appropriate witness/keys.

Consequence: bug fixes have to be applied in three places (e.g., the VRF-error-constructor fix from the L597 thread lives in the shared helper `checkVRFOutput`, but the call sites still duplicate the classify-and-branch structure), and invariants about persistent/non-persistent seat ranges are re-asserted three times instead of enforced once.

Rough shape of an extracted classifier:

```haskell
data MemberKind crypto
  = Persistent    LedgerStake (VoteVerificationKey crypto)
  | NonPersistent LedgerStake (VoteVerificationKey crypto) (VRFVerificationKey crypto)

classifyMember
  :: VotingCommittee crypto WFALS
  -> SeatIndex
  -> Either (VotingCommitteeError crypto WFALS) (MemberKind crypto)
```

Each of the three impls becomes a short post-classification step. The bounds check, `getCandidateInSeat`, and `isPersistentMember` live in `classifyMember` exactly once.

Not blocking — purely a refactor suggestion. Defer if you prefer to land this PR as-is and clean up once the Peras wiring arrives.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+            )
+          ,
+            ( poolId
+            , poolPublicKey
+            , LedgerStake poolStake
+            , Cumulative (LedgerStake stakeAccR')
+            )
+          )
+
+-- | Retrieve the candidate information associated to a given seat index.
```
dnadales: This is a partial function — `Array.!` bottoms with `"array index out of range"` if the documented PRECONDITION is violated. All current call sites in WFALS and EveryoneVotes guard with `seatIndexWithinBounds` first, but nothing in the type system forces them to. The `assert` used at call sites (e.g., `WFALS.hs` L266) is `Control.Exception.assert` — compiled out with `-O`, so in release builds the bounds check is purely convention.

If a future caller forgets the guard (new Peras wiring, refactor, additional code path), the node crashes with an unhelpful `Array.!: undefined array element` in the middle of a consensus-critical path.

Suggested: make the function total. Two reasonable options:

**(a) Return `Maybe`:**
```haskell
getCandidateInSeat :: SeatIndex -> ExtWFAStakeDistr a
                   -> Maybe (PoolId, a, LedgerStake, Cumulative LedgerStake)
getCandidateInSeat i d
  | seatIndexWithinBounds i d = Just ((Array.!) (unExtWFAStakeDistr d) i)
  | otherwise                 = Nothing
```
Callers `maybe`-propagate to an existing error constructor (e.g., `MissingSeatIndex`).

**(b) Return `Either SomeError`** — more uniform with the rest of the committee API if a shared error type exists; commits this module to a particular error type (currently there is not one at the `WFA` level).

Either way, the call-site `assert` blocks become unnecessary — bounds-checking is encoded in the return type. Author's call which of (a) or (b) fits best with the envisioned API shape.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -0,0 +1,371 @@
...
+            . List.mapAccumL accumVotersWithPositiveStake 0
+            . Array.assocs
+            . unExtWFAStakeDistr
+            $ stakeDistr
+
+    pure $
+      EveryoneVotesVotingCommittee
+        { extWFAStakeDistr = stakeDistr
+        , candidateSeats = seats
+        , numActiveVoters = poolsWithPositiveStake
```
dnadales: `ExtWFAStakeDistr` already carries `numPoolsWithPositiveStake` (`WFA.hs` L311–316), precomputed during `mkExtWFAStakeDistr`. Recomputing it here via `mapAccumL` undoes the "precompute once, reuse across instances" pattern called out in the PR description.

Suggested: drop the `accumVotersWithPositiveStake` helper entirely and read the precomputed value:
```haskell
let seats =
      Map.fromList
        . fmap (\(seatIndex, (poolId, _, _, _)) -> (poolId, seatIndex))
        . Array.assocs
        . unExtWFAStakeDistr
        $ stakeDistr

pure $
  EveryoneVotesVotingCommittee
    { extWFAStakeDistr = stakeDistr
    , candidateSeats = seats
    , numActiveVoters = numPoolsWithPositiveStake stakeDistr
    , epochNonce = nonce
    }
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+        (seatIndex, Just vrfOutput)
+          | seatIndexWithinBounds seatIndex (extWFAStakeDistr committee)
+          , not (isPersistentMember seatIndex committee) -> do
+              let (_, voterPublicKey, voterStake, _) =
+                    getCandidateInSeat seatIndex (extWFAStakeDistr committee)
+              let voterVoteVerificationKey =
+                    getVoteVerificationKey (Proxy @crypto) voterPublicKey
+              let voterVRFVerificationKey =
+                    getVRFVerificationKey (Proxy @crypto) voterPublicKey
+              let numSeats =
```
dnadales, line 523: Perf observation, not a bug. Each `implVerifyCert` call runs this Taylor expansion **once per non-persistent voter** in the cert. For a committee of ~500 non-persistent voters that is roughly 5k–10k inner iterations on `FixedPoint` per cert — microseconds per core in practice, but not free.

Caching within this module is awkward:
- Keying on `(SeatIndex, VRFOutput)` would need mutable state on `VotingCommittee`, since VRF outputs arrive with each cert. That breaks the pure-data shape and grows unbounded per epoch.
- Within a single `implVerifyCert` call, each voter has a unique `(SeatIndex, VRFOutput)` — no redundancy to memoize.

The real caching lever is **at the cert-storage layer** (`PerasCertDB` after #1966): once a cert is verified, subsequent verification attempts on the same cert should short-circuit to O(1). That is architectural and outside this PR's scope.

Suggested action: benchmark `implVerifyCert` against realistic committee sizes during the Peras integration before drawing conclusions. If verification latency matters, the fix likely lives in the cert-store, not here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+      nonce
+      totalSeats
+      stakeDistr
+    ) = do
+    ( numPersistentVoters
+      , numNonPersistentVoters
+      , persistentStake
+      , nonPersistentStake
+      ) <-
+      bimap WFAError id $
```
dnadales, line 228: Style nit (applies throughout both files): `bimap f id` is `first f` from `Data.Bifunctor`. Seven occurrences:

- `WFALS.hs`: L230, L579, L597, L618, L640
- `EveryoneVotes.hs`: L346, L365

Suggest swapping to `first` everywhere — shorter and more idiomatic. Not blocking; purely consistency.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+--
+-- PDF: https://eprint.iacr.org/2023/1273.pdf
+--
+-- For this, we combine the deterministic portion of the weighted Fait-Accompli
+-- scheme (defined in @Ouroboros.Consensus.Committee.WFA@) with local sortition
+-- (defined in @Ouroboros.Consensus.Committee.LS@) as a fallback scheme.
+module Ouroboros.Consensus.Committee.WFALS
+  ( -- * Voting committee interface
+    WFALS
+  , VotingCommittee -- opaque, only the metrics below are exported
```
dnadales: Minor docs nit (same pattern in `EveryoneVotes.hs` L13): "`opaque, only the metrics below are exported`" attached to `VotingCommittee` can read to a scanner as "this entire section is opaque", but `VotingCommitteeInput (..)`, `VotingCommitteeError (..)`, `EligibilityWitness (..)`, `Vote (..)`, and `Cert (..)` on the same list are all fully exposed.

Suggested rephrasing to make the scope clear, e.g.:
```haskell
  , VotingCommittee -- constructor hidden; use the record accessors below as read-only metrics
```
or promote the opacity note to a Haddock-level section header above the export list.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+  | notEnoughPoolsWithPositiveStake =
+      Left
+        ( NotEnoughPoolsWithPositiveStake
+            totalSeats
+            (numPoolsWithPositiveStake extWFAStakeDistr)
+        )
+  | otherwise =
+      -- We should have /at most/ as many persistent voters as the total
+      -- committee size, but not more.
+      assert (numPersistentVoters <= unTargetCommitteeSize totalSeats) $
```
dnadales, line 129: `Control.Exception.assert` is compiled out with `-O` (default for release builds), so this invariant is not enforced at runtime — it serves as documentation only. If it ever fails (arithmetic bug, refactor breaking the monotonicity shortcut), the result is silently wrong.

Good news: PR #1977 already adds `prop_weightedFaitAccompliPersistentSeats_persistentSeatsRespectNumSeats` (`WFALS/Model/Tests.hs` L56–57) which checks exactly this invariant on the model, and the model is pinned to the real implementation via `modelConformsToRealImplementation`. So the property is effectively enforced through the test suite, even though the runtime `assert` is a no-op in release builds.

Worth a one-line pointer comment here so a future reader of release-build crash forensics does not assume the invariant is unenforced:

```haskell
-- Runtime-asserted here for documentation; CI enforces this via
-- prop_weightedFaitAccompliPersistentSeats_persistentSeatsRespectNumSeats
-- (Test.Consensus.Committee.WFALS.Model.Tests).
assert (numPersistentVoters <= unTargetCommitteeSize totalSeats) $ ...
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
dnadales: Comment typos collected across both files:

- `WFA.hs` L186: "with its **give** stake and **relatile** position" → `given`, `relative`
- `WFALS.hs` L162: "committe" → `committee`
- `WFALS.hs` L383: "NOTE: theres is" → `there is`

One-line fixes, grouped here for visibility.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -0,0 +1,661 @@
...
+  ) where
+
+import Cardano.Ledger.BaseTypes (NonZero (..), Nonce, nonZero)
+import Control.Exception (assert)
+import Control.Monad (void)
+import Control.Monad.Zip (MonadZip (..))
+import qualified Data.Array as Array
+import Data.Bifunctor (Bifunctor (..))
+import Data.Containers.NonEmpty (HasNonEmpty (..))
+import Data.Data (Proxy (..))
```
dnadales, line 54: Import polish, bundled:

**1. `WFALS.hs` L55** — `import Data.Data (Proxy (..))`: `Data.Data` re-exports `Proxy` for historical reasons, but the canonical location is `Data.Proxy`. `EveryoneVotes.hs` L36 already uses `Data.Proxy` in this PR, so switching here aligns the two modules.

**2. `EveryoneVotes.hs` L25–26** — `NonZero` is imported twice:
```haskell
import Cardano.Ledger.BaseTypes (HasZero (..), NonZero, Nonce)
import Cardano.Ledger.BaseTypes.NonZero (NonZero (..), nonZero)
```
Harmless but easy to trim: drop `NonZero` from the first line (the second brings it in with constructors already).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+ where
+  stakeDistrArray =
+    listArray
+      ( SeatIndex 0
+      , SeatIndex (fromIntegral (Map.size pools) - 1)
+      )
+      cumulativeStakeAndPools
+
+  ((_totalStake, numPoolsWithPositiveStakeAcc), cumulativeStakeAndPools) =
+    List.mapAccumR
```
dnadales, line 394: `mapAccumR` (right-to-left) is load-bearing here: seat 0's cumulative stake ends up equal to the total, and the last seat's cumulative equals its own stake. Flipping to `mapAccumL` would silently invert the cumulative semantics, corrupting `isAbovePersistentSeatThreshold`.

The direction is encoded in a single character (`R` vs `L`) which is easy to skim past. A brief comment at the call site would help:

```haskell
-- mapAccumR (right-to-left) so seat 0's cumulative = total stake
-- and the last seat's cumulative = its own stake.
```

**Follow-up worth considering:** PR #1977's model already uses a phantom-typed `Cumulative` phase (`WFALS/Model.hs` L88, L451 — `Stake Ledger Cumulative`) that mechanically distinguishes left- from right-cumulative. Mirroring this on the real `Cumulative a` type in `Committee/Types.hs` would make the direction a compile-time guarantee rather than a comment-level contract. Larger refactor, but aligns real and model implementations.

**Test gap:** PR #1977 tests the direction on the model, and the real impl is pinned to the model via `modelConformsToRealImplementation` — so direction regressions are caught transitively. A direct property on the real implementation (`cumulative[0] == sum of all stakes`, `cumulative[lastSeat] == stake at last seat`) would make the invariant explicit at the implementation side, not just indirect via conformance. Small addition to `WFALS/Tests.hs`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -0,0 +1,403 @@
...
+-- | Evaluate whether a voter with its give stake and relatile position in the
+-- stake distribution can be granted a persistent seat in the voting committee.
+isAbovePersistentSeatThreshold ::
+  -- | Total committee size (persistent + non-persistent)
+  TargetCommitteeSize ->
+  -- | Current voter seat index
+  SeatIndex ->
+  -- | Current voter stake
+  LedgerStake ->
+  -- | Cumulated stake of voters with smaller or equal stake than the current one
```
dnadales: Docstring precision nit: "Cumulated stake of voters with smaller or equal stake than the current one" is accurate **when there are no ties**. With ties, the tiebreaker splits equal-stake pools across seat indices, and the cumulative at seat `i` only includes the *subset* of equal-stake pools that come after seat `i` in tiebreaker order (plus the current voter) — not all equal-stake pools.

The mathematically clean characterization is just seat-index-based:

```haskell
-- | Cumulated stake of voters at the current seat index or later in sorted order.
--   Equivalently: total stake of voters with stake ≤ this voter's, with ties
--   resolved by the tiebreaker.
Cumulative LedgerStake ->
```

Matters only at ties, but a reader building a mental model from the current wording may underestimate how the tiebreaker affects "equal stake" accounting.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
```diff
@@ -0,0 +1,167 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Local sortition used by non-persistent members of the voting committee
```
dnadales: This module's header is a single line ("Local sortition used by non-persistent members of the voting committee") with no reference to the paper or CIP-0164. Compare with `WFALS.hs` L13–29, which cites the Fait-Accompli paper with DOI and eprint link.

Since `LS.hs` implements non-trivial numerical machinery (Poisson thresholding, Taylor expansion), a reader trying to verify the implementation against a spec currently has to go hunting. Two lines would close the gap:

```haskell
-- | Local sortition for non-persistent voters.
--
-- Implements the @LS@ component of the wFA^LS scheme from the
-- Fait-Accompli Committee Selection paper
-- (https://eprint.iacr.org/2023/1273.pdf, §<section>).
--
-- Also specified in CIP-0164:
-- https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md
```

(Author should supply the correct paper section — I have not cross-referenced.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
tbagrel1: Indeed it isn't needed in EveryoneVotes. I'll remove it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
tbagrel1: Indeed 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
tbagrel1: Agree, we should return `Right Nothing` in this case

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
tbagrel1: I'll let @agustinmista handle this one, since he was the one involved in these maths :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
tbagrel1: Actually there hasn't been a final decision yet on the committee selection scheme that will be adopted for Peras.

The points mentioned here are probably worth considering in the second phase of the project, once a committee selection choice has been solidified. I opened https://github.com/tweag/cardano-peras/issues/231 to track this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
tbagrel1: I don't think this is worth mentioning in a comment. We use `assert` in many other places of the code to help catch potentially invalid implementation, with no such comment attached. I'll mark as resolved, feel free to reopen if you think we should really modify this one.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
tbagrel1: The docstring above clearly ties the guarantees ("deterministic and resilient to manipulation") to the _one possible_ implementation of it we mention.
It is actually desirable to pass a simpler tiebreaker e.g. for testing purposes.

So I don't think we need to enforce anything here (and also, enforcing fairness is not a trivial task).

That being said, I agree that it could be good for this module to also provide the reference tiebreaker using the epoch nonce. Let me add it

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
tbagrel1: I'm not I understand what you would like to see here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
tbagrel1: I don't think we should modify the implementation to conform to the model here. I don't think the cumulativeStake is ever updated elsewhere than here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: I tend to disagree, `first` is slightly more equivocal. Feel free to reopen if you want us to really change it :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: Depending on the use-case, there is also a third case where a _candidate_ is not a member at all. Sometimes we should treat this as an error, sometimes it's fine.

Now that `getCandidateInSeat` and `seatWithinBounds` have been merged, I think the structure is fine as-is. Feel free to reopen if you feel strongly about it :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: Benchmarking and optimisation is targeted for a future stage of the project. We don't know yet which part of vote verification will dominate the CPU costs. I'll close it for now, since we already have tasks tracking that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: `LocalSortitionError` was indeed redundant with `InvalidVoterEligibilityProof` (which has a clearer name), so I removed the former.

In `implCheckShouldVote`, I changed the error kind to `CryptoError`, since computing our own VRF output should only fail if crypto goes wrong.
However in `implVerifyVote`, we should keep the `InvalidVoterEligibilityProof` error.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: Good catch! These tests will be included as part of [Peras 25](https://github.com/IntersectMBO/ouroboros-consensus/pull/1977). Closing for now

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
tbagrel1: I went for option (a), and clarified the name to better reflect all invariants it holds

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
tbagrel1: Solved directly in VotesWithSameTarget :)

tbagrel1 (commented):
Hi @dnadales,
I should have answered and/or fixed most of the review comments in the chain of commits [here](https://github.com/IntersectMBO/ouroboros-consensus/pull/2015)

I didn't push them directly to the current branch to prevent breaking the PR tower on top of this one, as requested by @agustinmista

@agustinmista I'll let you have a look, fix what is needed, and backport these changes in the original branch. Thansk! Let me know if you have questions!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/LS.hs
agustinmista: This is something we probably want to do while working on Peras-to-mainnet, and probably something to also discuss with @perturbing if Leios ever decides to go back to wFA^LS.

I opened https://github.com/tweag/cardano-peras/issues/234 to keep track of this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
dnadales: Sorry, I meant an example of the extra payload associated to the voter (I don't know if `pubkey` is one such example).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
dnadales: Fair enough.

dnadales (approved):
Thanks for the round of fixes.

One follow-up worth surfacing: with `UniqueVotesWithSameTarget` from #2020 closing the type-level dedup hazard, the remaining caller contract — `cmpVotes` must order `Eq` on shared `SeatIndex` for both `WFALS` and `EveryoneVotes` `Vote` constructors — will need to be wired correctly at the Peras integration site. Worth adding this to a production-readiness tracker (new or existing).

dnadales (approved):
Thanks for the round of fixes.

PR #1976, opened 2026-04-13
[Peras 24] BLS crypto backend for voting committees

### LINKED ISSUE #195: Implement committee selection schemes

### LINKED ISSUE #212: Template Vote and Cert Cryptography

agustinmista:
This PR:

1. Implements crypto utilities needed instantiate concrete voting committee implementations with BLS crypto, and
2. a `TestCrypto` scheme defined around these utilities, with:
    + instances for all the type classes relevant to committee selection, and
    + property tests for round-trips and resilience to swap attacks

NOTES:

* #1977 instantiates both `WFALS` and `EveryoneVotes` with `TestCrypto` to define property tests.

### DIFF
--- a/changelog.d/20260413_101837_agustin.mista_bls_crypto_for_voting_committees.md
+++ b/changelog.d/20260413_101837_agustin.mista_bls_crypto_for_voting_committees.md
@@ -1,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+### Non-Breaking
+
+- Implemented BLS-based crypto helpers to instantiate voting committee implementations.
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto/BLS.hs
```diff
@@ -0,0 +1,425 @@
...
+  rawSerialiseVerKeyDSIGN . unPublicKey
+
+coercePublicKey ::
+  forall r2 r1.
+  PublicKey r1 ->
+  PublicKey r2
+coercePublicKey = coerce
+
+-- | BLS signature type, parameterized by key role
+type Signature :: KeyRole -> Type
```
ch1bo, line 160: Why does a Signature have a `KeyRole`? What are these key roles needed for and what is the difference to this `KeyScope` type?

I asked as similar question to @perturbing on his PoC repo a couple week back, but can't recall what his response was exactly, but I do remember "actually, not really needed".

agustinmista: For once, having the KeyRole as a phantom type here allows us to avoid mixing signatures used to sign votes, and signatures used as VRF outputs. Non-persistent votes have one of each, and it could be an easy mistake to make. 

In addition, certain operations are also only well-defined (at least in spirit) for `Signature VRF`, like extracting a natural number from them.

This module is based on @perturbing's PoC, so I'll leave it up to him to tell if I'm completely mistaken about this.

perturbing: I removed them from the PoC here https://github.com/cardano-scaling/leios-wfa-ls-demo/commit/3f29334ad05c0dcaa2fe5f1cece48a2c3d1876a3

agustinmista: Hey @perturbing! 

> I removed them from the PoC here https://github.com/cardano-scaling/leios-wfa-ls-demo/commit/3f29334ad05c0dcaa2fe5f1cece48a2c3d1876a3

Is this only applicable to Leios (because you no longer care about VRF outputs and only use BLS signatures for vote signing), or would you say it's also applicable to Peras? (I don't think so, but I might be wrong).

perturbing: The private key for the VRF and voting is the same, so it does not need the role. That is why I removed it; this also applies to peras.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto/BLS.hs
```diff
@@ -0,0 +1,425 @@
...
+
+-- | Key roles
+type data KeyRole
+  = -- | Key role for signing votes
+    SIGN
+  | -- | Key role for local sortition in elections
+    VRF
+  | -- | Key role for Proof of Possession
+    POP
+
```
dnadales, line 82: I wonder if `ProofOfPossession` would be clearer here? `SIGN` and `VRF` read immediately, but `POP` seems to require BLS-specific context — especially since the value-level type a few lines below already spells it out as `ProofOfPossession`.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/TestCrypto.hs
```diff
@@ -0,0 +1,552 @@
...
+                         input
+                         vrfOutputs
+                     ) of
+                  Left err ->
+                    counterexample
+                      ("Aggregate VRF verification failed: " <> err)
+                      $ property False
+                  Right () ->
+                    property True
+
```
dnadales, line 419: Tiny typo: `verifaction` → `verification`.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/TestCrypto.hs
```diff
@@ -0,0 +1,552 @@
...
+                counterexample
+                  ("VRF evaluation failed: " <> err)
+                  $ property False
+              Right vrfOutputs -> do
+                forAll (swapTwoElements vrfOutputs) $ \(i, j, swappedVRFOutputs) -> do
+                  case ( batchVerifyVRFOutputs @TestCrypto
+                           vrfVerificationKeys
+                           input
+                           swappedVRFOutputs
+                       ) of
```
dnadales, line 449: Tiny typo: `module` → `modulo`.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/TestCrypto.hs
```diff
@@ -0,0 +1,552 @@
...
+
+instance CryptoSupportsAggregateVoteSigning TestCrypto where
+  -- NOTE: in a real implementation, we would want to wrap these into newtypes
+  type AggregateVoteVerificationKey TestCrypto = VoteVerificationKey TestCrypto
+  type AggregateVoteSignature TestCrypto = VoteSignature TestCrypto
+
+  aggregateVoteVerificationKeys _ pks =
+    BLS.aggregatePublicKeys @SIGN pks
+
+  aggregateVoteSignatures _ sigs = do
```
dnadales, line 169: Aren't these `coercePrivateKey @VRF` / `coercePublicKey @VRF` calls no-ops here? `sk` is already `BLS.PrivateKey VRF` and `pk` is already `BLS.PublicKey VRF`, so the coercion goes from `VRF` to `VRF`. I verified it compiles without them. I wonder if we should drop them to avoid suggesting a real role conversion is happening.

dnadales (commented):
No objections on my side, other than the comments I've made. 

There's an ongoing discussion between @ch1bo, @agustinmista, and @perturbing about whether the role-parameterized key types are needed given that the underlying BLS key is the same for SIGN and VRF. I'll defer to the outcome of that thread.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/TestCrypto.hs
agustinmista: Good catch!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto/BLS.hs
agustinmista: We can change it, just not to `ProofOfPossession`, because it would clash with the actual type carrying the proof:

```haskell
-- | BLS proof of possession type
newtype ProofOfPossession = ProofOfPossession
  { unProofOfPossession :: PossessionProofDSIGN BLS12381MinSigDSIGN
  }
  deriving stock (Eq, Show)
  deriving newtype (FromCBOR, ToCBOR)
```

Personally, I think `POP` is fine as long as the acronym is documented in the constructor's haddock (which it is, also just confirmed that the LSP picks it up correctly).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto/BLS.hs
agustinmista: > The private key for the VRF and voting is the same, so it does not need the role.

Our initial implementation doesn't make this assumption, allowing for the vote-signing and VRF-computing keys to be different. I think this is for the better until we have a final decision on the crypto scheme for Peras, but I would be happy to simplify it later on if we end up using the very same key for both in mainnet.

In addition, I mentioned in my [initial reply](https://github.com/IntersectMBO/ouroboros-consensus/pull/1976#discussion_r3072401028) that _do_ we want `Signature SIGN` and `Signature VRF` to be separate types for now to avoid bugs, so we can't entirely get rid of `KeyRole` for now.

For now, I opened https://github.com/tweag/cardano-peras/issues/245 to keep track of this until we start working on the `peras-to-mainnet` phase.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto/BLS.hs
jasagredo: Just happened to see this comment, and wanted to mention this: https://hackage.haskell.org/package/sop-core-0.5.0.2/docs/Data-SOP-NP.html#t:POP

PR #1977, opened 2026-04-13
[Peras 25] Conformance and property tests for voting committee implementations

### LINKED ISSUE #195: Implement committee selection schemes

### LINKED ISSUE #210: Voting Committee selection logic

agustinmista:
This PR provides:

* Property tests for generic committee helpers (`VotesWithSameTarget`)
* Conformance tests for `WFALS` between the real implementation and:
  + the (Haskell) model, against randomly generated stake distributions,
  + the (Rust) prototype, against its precomputed outputs for the example stake distribution taken from mainnet.
* Property tests for both `WFALS` and `EveryoneVotes`, asserting that:
  + votes cast by eligible voters are successfully verifiable under the same voting committee, and
  + fake (illegal) votes are successfully rejected.

NOTES:

* The first two commits reorganize a bit the test suite (originally only containing WFALS tests), to share definitions and make room for `EveryoneVotes` tests.

### DIFF
--- a/ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Utils.hs
+++ b/ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Utils.hs
@@ -1,0 +1,202 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+-- | Utility functions for the voting committee tests.
+module Test.Consensus.Committee.Utils
+  ( -- * General utilities
+    mkPoolId
+  , unfairWFATiebreaker
+
+    -- * QuickCheck generators
+  , genEpochNonce
+  , genPositiveStake
+  , genPools
+
+    -- * Property helpers
+  , eqWithShowCmp
+  , onError
+  , mkBucket
+
+    -- * Tabulation helpers
+  , tabulateNumPools
+  , tabulatePoolStake
+  ) where
+
+import qualified Cardano.Crypto.DSIGN.Class as SL
+import qualified Cardano.Crypto.Seed as SL
+import Cardano.Ledger.BaseTypes (Nonce (..), mkNonceFromNumber)
+import qualified Cardano.Ledger.Core as SL
+import qualified Cardano.Ledger.Keys as SL
+import Data.Map (Map)
+import qualified Data.Map.Strict as Map
+import Data.String (IsString (..))
+import Ouroboros.Consensus.Committee.Types (LedgerStake (..), PoolId (..))
+import Ouroboros.Consensus.Committee.WFA (WFATiebreaker (..))
+import Test.QuickCheck
+  ( Arbitrary (..)
+  , Gen
+  , Property
+  , choose
+  , counterexample
+  , elements
+  , frequency
+  , tabulate
+  , vectorOf
+  )
+import Test.Util.QuickCheck (geometric)
+
+-- * General utilities
+
+-- | Create a pool ID from an arbitrary string of any length.
+--
+-- NOTE: we are assuming that this function preserves uniqueness.
+mkPoolId :: String -> PoolId
+mkPoolId str =
+  PoolId
+    . SL.hashKey
+    . SL.VKey
+    . SL.deriveVerKeyDSIGN
+    . SL.genKeyDSIGN
+    . SL.mkSeedFromBytes
+    . fromString
+    $ paddedStr
+ where
+  paddedStr
+    | length str >= neededBytes = take neededBytes str
+    | otherwise = str <> replicate (neededBytes - length str) '0'
+
+  neededBytes = 32
+
+-- | An unfair tie-breaker that compares pool IDs lexicographically.
+unfairWFATiebreaker :: WFATiebreaker
+unfairWFATiebreaker =
+  WFATiebreaker compare
+
+-- * QuickCheck generators
+
+-- | Generate a random nonce for testing purposes.
+genEpochNonce :: Gen Nonce
+genEpochNonce =
+  frequency
+    [ (1, pure NeutralNonce)
+    , (9, mkNonceFromNumber <$> arbitrary)
+    ]
+
+-- | Generate a positive stake value using a geometric distribution.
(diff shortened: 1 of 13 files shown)


ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Tests.hs
```diff
@@ -0,0 +1,552 @@
...
+      unNonPersistentCommitteeSize (nonPersistentCommitteeSize committee)
+
+    persistentRange =
+      (0, numPersistentSeats - 1)
+
+    nonPersistentRange =
+      (numPersistentSeats, numPersistentSeats + numNonPersistentSeats - 1)
+
+    -- This witness looks like a persistent member, but its seat index lies
+    -- outside the persistent members range (case 1).
```
dnadales, line 378: I wonder if we could hit an inverted `choose` range here when `numNonPersistentSeats == 0`? In that case `nonPersistentRange` becomes `(n, n - 1)` on `Word64`, and `genFakePersistentMemberWitness` calls `choose` on it — `randomR` does not guarantee anything when `lo > hi`.

Unlike the EveryoneVotes generator, which guards `genPoolWithZeroStakeWitness` behind a frequency-weight check, both WFALS alternatives always have weight 1.

What about something like the EveryoneVotes pattern — gate each alternative on whether its range is non-empty? Something like:

```haskell
frequency
  [ (if numNonPersistentSeats > 0 then 1 else 0, genFakePersistentMemberWitness)
  , (if numPersistentSeats > 0 then 1 else 0, genFakeNonPersistentMemberWitness)
  ]
```

(The second guard covers `persistentRange` being empty if `numPersistentSeats == 0`, though that may not happen in practice.)

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Tests.hs
```diff
@@ -0,0 +1,552 @@
...
+                            witness
+                    case verifyVote committee vote of
+                      Left err ->
+                        counterexample
+                          ("vote verification failed: " <> show err)
+                          $ property False
+                      Right witness' ->
+                        counterexample
+                          "vote verificateion mismatch"
+                          $ eqWithShowCmp
```
dnadales, line 153: Tiny typo: `"verificateion"` → `"verification"`?

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS.hs
```diff
@@ -1,36 +1,278 @@
...
-            (fromIntegral targetCommitteeSize)
-            (Map.map Model.rationalToStake stakeDistr)
-     in ( Map.size persistentSeats
+  -- NOTE: this is not a good tiebreaker for real-world use, but it is
+  -- sufficient here because the actual order of pools with the same stake does
+  -- not matter when computing the persistent vs. non-persistent seat /counts/.
+  tiebreaker =
+    compare
+
+  mkStakeDistr =
```
dnadales, line 68: Tiny typo: `"strake"` → `"stake"`?

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Utils.hs
```diff
@@ -0,0 +1,203 @@
...
+-- | Generate a positive stake value using a geometric distribution.
+genPositiveStake :: Gen LedgerStake
+genPositiveStake =
+  LedgerStake
+    . toRational
+    . (+ 1)
+    <$> geometric 0.25
+
+-- | Generate an alphanumeric string of length 8.
+-- | Generate a non-empty map of pools with crypto keys and varying stakes.
```
dnadales, line 93: Should we drop this first haddock line? It looks like it describes the internal `alphaNumString` helper rather than `genPools` — GHC will silently discard it since the next `-- |` wins.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Tests.hs
```diff
@@ -0,0 +1,552 @@
...
+                            , "but got these " <> show (length witnessesFromCert) <> " witnesses:"
+                            , intercalate "\n" (fmap showWitness witnessesFromCert)
+                            ]
+                        )
+                        $ cmpWitnesses originalWitnesses witnessesFromCert
+
+-- * WFALS helpers
+
+castVote ::
+  PrivateKey TestCrypto ->
```
dnadales: I notice `castVote` reimplements `forgeVote` / `implForgeVote` from scratch rather than calling it. Would it make sense to use `forgeVote` from the class directly? That way we would also exercise the class dispatch, and if `forgeVote` were ever refactored these tests would catch regressions.

(Not blocking — the implementations are identical today.)

dnadales (commented):
Than you for adding the tests! Only one substantial concern about the use of `choose` and some minor observations.

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/Utils.hs
agustinmista: Good catch! :sweat_smile:

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Tests.hs
agustinmista: Great idea! I don't remember exactly why I had the `castVote` helpers originally, but it no longer makes sense to have them. I refactored the code to call `forgeVote` from the class directly :)

ouroboros-consensus/test/consensus-test/Test/Consensus/Committee/WFALS/Tests.hs
agustinmista: Good idea! 

Even if we don't hit the corner case in practice (which I don't think we do because it would make the property fail), having the guards is good for documentation :+1:

dnadales (approved):
Thanks for incorporating the changes 👍

PR #1978, opened 2026-04-13
[Leios prototype] Migrate EB announcement to Praos header

### LINKED ISSUE #837: [Leios Prototype] Migrate announcement and certifies flag from the Ledger Block into the Praos Header
In https://github.com/IntersectMBO/ouroboros-consensus/pull/1921 we implemented EB inclusion.

For convenience we carry the [EB Announcement and the Certification flag in the Ledger Block](https://github.com/IntersectMBO/cardano-ledger/blob/7e1fb14c916e58709aa36c4b4fe0265ba537e0f0/libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs#L123) itself, while the design expects it to be part of the [Praos Header](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs#L75).

Expected changes (I think):
- [x]  Move the  [EB Announcement and the Certification flag in the Ledger Block](https://github.com/IntersectMBO/cardano-ledger/blob/7e1fb14c916e58709aa36c4b4fe0265ba537e0f0/libs/cardano-ledger-core/src/Cardano/Ledger/Block.hs#L123) into the  [Praos Header](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs#L75).
- [x] Move [Leios state bits](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L291) from [ShelleyLedgerState](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L342) into [PraosState](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L272).
- [x] Move the [applyTick](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L305) and [applyBlock](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L325) to [tickChanDepState](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L431) and [reupdateChainDepState](https://github.com/IntersectMBO/ouroboros-consensus/blob/5c62210fac956a3ecc5bd6fe4b2453e755d8f8e9/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L492) respectively.

bladyjoker:
Resolves https://github.com/input-output-hk/ouroboros-leios/issues/837

DONE
  - [x] Move the EB announcement from the Ledger Block to the Praos Header
     - [x] Add `EbAnnouncement` type that holds an `EbHash` and its size!
     - [x] Remove the `certifies` field
  - [x] Move the LeiosState from the Ledger to the ChainDepState
    - [x] Adds 'cummulative' EB size tracking in the state (TODO: Grafana this)
  - [ ] ~~Move the Certificate from the Ledger into Consensus~~
      - Ugh for now I'm keeping it in the Ledger because it's easier, albeit not ideal
   - [x] Update to Ledger with backward compatible Block codec
   - [x] Introduces `ForgeBlockArgs` for my own sanity
   - [x] FIX: [EB only diffused if announced in forge](https://github.com/IntersectMBO/ouroboros-consensus/pull/1978/commits/85aafd4146617fea79f825d989bf1ca5b9dfb6b6)
 
NOTES
 - `ResolveLeiosBlock` machinery is a hack that works! We're blindly guessing that `blk` might contain something that can be resolved like a LeiosCertificate and it returns a `blk` that hopefully has a fully resolved `blk` that can be applied. Imo a morally correct approach would to have `data RankingBlock blk = LedgerRb blk | CertRb LeiosCertificate` which we can use to make such distinction and manage the resolution process here and elsewhere in the abstract code base where that distinction matters (like the ForgeLoop)
- `applyBlock` currently does `resolveLeiosBlock` and then `tickAndReapply` which is not incorrect as in the case of there being a Leios Cert, we need to base changes from the associated EB on top of the `prev` Ledger State (I think that means we shouldn't tick?)
  - see https://github.com/input-output-hk/ouroboros-leios/issues/857
- Apropos `tx-centrifuge` it doesn't seem like there's any way around breaking the API `BlockFetch`. For the current pragmatic solutions we hide the `Certificate` behind the `blk` and therefore `BlockFetch` will happily return you a Certificate which will break `tx-centrifuge`. That being said, `ChainSync` is also necessarily having a breaking change since we added new fields to the PraosHeader.


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -700,6 +700,16 @@ forkBlockForging IS{..} blockForging =
     -- At this point we have established that we are indeed slot leader
     trace $ TraceNodeIsLeader currentSlot
 
+    -- Leios: decide if we're forging an CertRb or a TxsRb
+    forgeType <-
+      lift $
+        leiosDecideForgeType blockForging $
+          LeiosDecideForgeTypeArgs
+            { ldftaChainDepState = headerStateChainDep (headerState unticked)
```
ch1bo, line 708: `ldfta` looks like your cat walked around on your keyboard 🐱

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -700,6 +700,16 @@ forkBlockForging IS{..} blockForging =
     -- At this point we have established that we are indeed slot leader
     trace $ TraceNodeIsLeader currentSlot
 
+    -- Leios: decide if we're forging an CertRb or a TxsRb
+    forgeType <-
+      lift $
+        leiosDecideForgeType blockForging $
```
ch1bo, line 706: Why is this function needed? The decision what to forge lies within `forgeBlock` already and having this additional abstract function which is `const $ return ForgeTxsRb` for most instances seems avoiadable.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -220,8 +234,10 @@ instance Crypto crypto => DecCBOR (HeaderBody crypto) where
         <! From
         <! From
         <! From
+        <! From
         <! mapCoder unCBORGroup From
         <! From
+        <! From
```
ch1bo, line 240: Must: prototype a backwards compatible decoder

We want to use previously recorded Praos blocks in the `demo/burst` scenario. The wire format of praos headers changed in a non-backwards compatible way.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -89,10 +95,14 @@ data HeaderBody crypto = HeaderBody
   -- ^ Size of the block body
   , hbBodyHash :: !(Hash.Hash HASH EraIndependentBlockBody)
   -- ^ Hash of block body
+  , hbBodyType :: !BodyType
+  -- ^ Type of the block body
```
ch1bo, line 99: Should: not need this

What is it good for? The CIP also does not specify it

ch1bo (changes requested):

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
bladyjoker: Removed, and introduced EbAnnouncement with Hash and Size

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
bladyjoker: Done.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
bladyjoker: Keeping it like this for now because otherwise I need to carry some additional information `ForgeBlockArgs`.

bladyjoker: Added a NOTE though

PR #1979, opened 2026-04-13
Leios: per-thread connections to LeiosDemoDb

### LINKED ISSUE #794: Benchmark SQLite performance using criterion (or similar)

ch1bo:
Change the LeiosDemoDb interface to have a per-thread `LeiosDbConnection`.

This allows us to use sqlite's multi-thread primitives (e.g. `SQLOpenFullMutex`) and improves the benchmark runtime by at least 2x.

No WAL and `SQLOpenFullMutex`:
```
  run 1/5: 1.51683722 s
  run 2/5: 1.542545014 s
  run 3/5: 1.55612177 s
  run 4/5: 1.588956144 s
  run 5/5: 1.650447848 s
  => min=1.51683722 s  avg=1.5709815992 s  max=1.650447848 s
```

`pragma journal_mode = WAL` and `pragma synchronous = normal` and `SQLOpenFullMutex`:
```
  run 1/5: 1.05603214 s
  run 2/5: 1.135455939 s
  run 3/5: 1.120406063 s
  run 4/5: 1.144128676 s
  run 5/5: 1.194659677 s
  => min=1.05603214 s  avg=1.130136499 s  max=1.194659677 s
```

`pragma journal_mode = WAL` only:
```
  run 1/5: 1.464840392 s
  run 2/5: 2.16140798 s
  run 3/5: 1.407903973 s
  run 4/5: 1.368493564 s
  run 5/5: 1.405035783 s
  => min=1.368493564 s  avg=1.5615363384 s  max=2.16140798 
```

`pragma journal_mode = WAL` and `pragma synchronous = normal` only:
```
  run 1/5: 1.032053432 s
  run 2/5: 1.078641398 s
  run 3/5: 1.146966806 s
  run 4/5: 1.175950854 s
  run 5/5: 1.234927115 s
  => min=1.032053432 s  avg=1.133707921 s  max=1.234927115 s
```


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -0,0 +1,98 @@
...
+
+data LeiosDbHandle m = LeiosDbHandle
+  { subscribeEbNotifications :: HasCallStack => m (StrictTChan m LeiosNotification)
+  -- ^ Subscribe to new EBs and EBTxs being stored by the LeiosDB. This will
+  -- only inform about new additions, starting from when this function was
+  -- called.
+  -- TODO: make return type more descriptive (e.g. Subscription { getNext :: STM m LeiosNotification })
+  , open :: m (LeiosDbConnection m)
+  -- ^ Open a new connection to the LeiosDb.
+  }
```
bladyjoker, line 41: It's a bit dubious to have a difference between a Handle and a Connection, I'd rather remove this in favor of just `openInMemoryLeiosDb` and `openSqliteLeiosDb`.

Not sure why `subscribeEbNotifications` is not in LeiosDbConnection?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
bladyjoker, line 1: Thanks for splitting this into modules ❤️

bladyjoker (approved):
Lovely! Just don't forget to clean up for the CI, we need to be aware of testing failures moving forwards.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
ch1bo: >  I'd rather remove this in favor of just openInMemoryLeiosDb and openSqliteLeiosDb.

`LeiosDbHandle` turned into a bit of a facade, I agree, but a common handle is needed for exactly the subscription / notification mechanism.

> Not sure why subscribeEbNotifications is not in LeiosDbConnection?

We could provide the subscription also off the connection, but importantly it must be shared across all handles. Otherwise the notifications would not be seen across threads.

ch1bo: I'm also not particularly happy about the `Connection` abstraction, but it should be more semantical. Also a distinction of read and write operations would allow us, in the case of sqlite, improve performance even further.

Something like `EBStoreReader` and `EBStoreWriter` comes to mind..

ch1bo: In fact, that subscription mechanism could be more focused too and connect the Leios Fetch logic with exactly the only two consumers: leios notify servers and vote logic.

PR #1971, opened 2026-04-10
Test EB inclusion

dnadales:
Closes https://github.com/input-output-hk/ouroboros-leios/issues/836

PR #1988, opened 2026-04-14
Add cumulative tx bytes to ledger state

dnadales:
## Summary

- Add `sllsCumulativeTxBytes :: Word64` to `ShelleyLedgerLeiosState`, incremented on each block application by summing individual `sizeTxF` values per transaction.
- Handles both `BodyInline` and resolved `BodyCertificate` blocks via explicit pattern match (avoids the partial `bodyTxs`).
- ThreadNet test independently resolves and folds the chain, comparing the computed sum against the ledger state accumulator.
- Remove hardcoded `withMaxSuccess 1` from all Leios ThreadNet tests so `--quickcheck-tests=N` works from the command line.

Part of: [ouroboros-leios#845](https://github.com/input-output-hk/ouroboros-leios/issues/845#issuecomment-4236763900)

Stacked on top of: https://github.com/IntersectMBO/ouroboros-consensus/pull/1971
Blocked by: https://github.com/IntersectMBO/ouroboros-consensus/pull/1996


bladyjoker (commented):
Nice!

Two wishes:

1. Move the `sllsCumulativeTxBytes` directly under `ShelleyLedgerState` as I just removed the Leios bits. It also makes sense, there's nothing Leios specific about this.

2. Can we use `Ouroboros.Consensus.Mempool.TxSeq:TxSeqMeasure` sooner rather than later?
  - It's a bit more work to provide CBOR codecs for and wire types, BUT
  - We have the JSON encoders for it,
  - We have tracing events that contain it `ForgedBlock`
  - We have Grafana dashboards that show it
 
Example:

```haskell
TxSeqMeasure
    { mCount = 384
    , mMinTicket = TicketNo
        { unTicketNo = 1269 }
    , mMaxTicket = TicketNo
        { unTicketNo = 1652 }
    , mSize = ConwayMeasure
        { alonzoMeasure = AlonzoMeasure
            { byteSize = IgnoringOverflow
                { unIgnoringOverflow = ByteSize32
                    { unByteSize32 = 89088 }
                }
            , exUnits = ExUnits'
                { exUnitsMem' = 0
                , exUnitsSteps' = 0
                }
            }
        , refScriptsSize = IgnoringOverflow
            { unIgnoringOverflow = ByteSize32
                { unByteSize32 = 0 }
            }
        }
    }
```

dnadales:
Thank you for checking the PR @bladyjoker 

> Nice!
> 
> Two wishes:
> 
> 1. Move the `sllsCumulativeTxBytes` directly under `ShelleyLedgerState` as I just removed the Leios bits. It also makes sense, there's nothing Leios specific about this.

Makes sense! 

> 2. Can we use `Ouroboros.Consensus.Mempool.TxSeq:TxSeqMeasure` sooner rather than later?
> 

For wish 2, I looked into replacing the plain `Word64` with a richer `CumulativeTxMeasure` type (tx count, byte size, execution units mem/steps). `TxSeqMeasure ConwayMeasure` can't be used directly — `ByteSize32` is `Word32` with silent overflow (~11 days at mainnet rates). 

I have an implementation ready (stashed locally) that works for tx count and byte size, but there's a constraint issue with execution units: `totExUnits` requires `AlonzoEraTxWits era`, which is not implied by `ShelleyCompatible proto era`. Since `applyHelper` (where the accumulation happens) is era-polymorphic and gets instantiated for all eras during the hard fork sequence (Shelley → ... → Conway), we can't add the `AlonzoEraTxWits` constraint without breaking pre-Alonzo eras.

So I'd say let's leave this as is for now.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -330,6 +333,16 @@ applyBlockShelleyLedgerLeiosState blk leiosSt =
         MkLeiosPoint (blockSlot blk) . fromLedgerEbHash <$> SL.blockMayAnnouncedEb (shelleyBlockRaw blk)
     }
 
+blockTxBytes ::
```
bladyjoker, line 336: blockBytesSize maybe?

bladyjoker (approved):
This is cool!

bladyjoker:
Thanks for taking the time to investigate this

> > 2. Can we use `Ouroboros.Consensus.Mempool.TxSeq:TxSeqMeasure` sooner rather than later?
> 
> For wish 2, I looked into replacing the plain `Word64` with a richer `CumulativeTxMeasure` type (tx count, byte size, execution units mem/steps). `TxSeqMeasure ConwayMeasure` can't be used directly — `ByteSize32` is `Word32` with silent overflow (~11 days at mainnet rates).
> 
> I have an implementation ready (stashed locally) that works for tx count and byte size, but there's a constraint issue with execution units: `totExUnits` requires `AlonzoEraTxWits era`, which is not implied by `ShelleyCompatible proto era`. Since `applyHelper` (where the accumulation happens) is era-polymorphic and gets instantiated for all eras during the hard fork sequence (Shelley → ... → Conway), we can't add the `AlonzoEraTxWits` constraint without breaking pre-Alonzo eras.
> 
> So I'd say let's leave this as is for now.

If it makes sense to mark this down in an issue that we can pick up later?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
ch1bo: I like `TxBytes`. It's about the transaction bytes, not about anything else in the block (if there is anything else)

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -140,10 +143,9 @@ tests :: TestTree
 tests =
   testGroup
     "Leios ThreadNet"
-    [ testProperty "EB production" $
-        withMaxSuccess 1 prop_leios_blocksProduced
-    , testProperty "EB certificate inclusion" $
-        withMaxSuccess 1 prop_leios_ebCertificateInclusion
+    [ testProperty "EB production" prop_leios_blocksProduced
```
ch1bo: This will take ages (at least on my computers). One run is `4s` for me on this and I could not bear to wait until the end of `100` runs. The difference between the 100 runs will only be the `Seed`. Do you think that's worth 5-10min longer test runs? 

Also, do we need distinct runs for each property, or can we have one run (or a few) and assert all these things about it after?

ch1bo (approved):
Finally, something seemed to be simple in the consensus codebase :D

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
dnadales: Good point. The things is that with one run only we were not triggering the prototype errors. I can adjust the runs so that we run 10 by default, and maybe group the properties into one, so that we save time there.

dnadales: Done!

dnadales:
This is now blocked by the proposed fix in https://github.com/IntersectMBO/ouroboros-consensus/pull/1996 Without it, CI will most likely fail.

PR #1989, opened 2026-04-15
[Leios Prototype] Cleanup

bladyjoker:
DONE
- [x] Just use `nix fmt` to run the formatters
- [x] Removes obsolete leios-mvd
- [ ] Filters the Haskell source to avoid unnecessary rebuilds


ch1bo (approved):
Do we have a CONTRIBUTING.md that could instruct contributors to know about this?

PR #1993, opened 2026-04-16
Replace set-git-rev binary patching with compile-time CPP flag

angerman:
## Summary

- Replace the `set-git-rev` binary patching approach with a compile-time GHC
  preprocessor flag (`-D__GIT_REV__`), following the same pattern used by
  `cardano-cli`
- Remove `haskellBuildUtils` and `set-git-rev` from `nix/tools.nix`, and the
  `setGitRevs` wrapper from `nix/ci.nix`
- Rewrite `Cardano.Tools.GitRev` to use CPP instead of Template Haskell,
  dropping the `cardano-git-rev` and `githash` dependencies

## Motivation

The `set-git-rev` approach modifies executable binaries after building to embed
the git revision hash. On aarch64-darwin, this breaks Mach-O code signatures
because the binary contents change without re-signing. While a workaround
existed (re-signing after patching), the compile-time approach is simpler, more
correct, and avoids the problem entirely.

## How it works

In `nix/haskell.nix`, when `inputs.self.rev` is available (clean git tree in
CI), the flag `--ghc-option=-D__GIT_REV__="<commit-hash>"` is passed via
`configureFlags`. The Haskell source uses CPP to read this define:

```haskell
gitRevStr :: String
#ifdef __GIT_REV__
gitRevStr = __GIT_REV__
#else
gitRevStr = ""
#endif
```

When built from a dirty tree (local development), `__GIT_REV__` is not defined
and `gitRev` returns `"unavailable (git info missing at build time)"`.

## Test plan

- [x] `nix build .#hydraJobs.aarch64-darwin.native.haskell96.exesNoAsserts.db-analyser` succeeds
- [x] `/usr/bin/codesign -vvv` confirms valid Mach-O signature
- [x] `db-analyser --help` shows the git rev footer
- [ ] CI builds on x86_64-linux, x86_64-darwin, aarch64-darwin
- [ ] Verify git rev appears in `db-analyser --help` when built from a clean commit in CI


angerman:
I believe #1990 to be the more pragmatic fix for this.

PR #1996, opened 2026-04-18
Fix error "impossible!" in Leios Prototype's fetch decision logic

### LINKED ISSUE #845: Demonstrate 200 TxkB/s on local devnet with prototype
## Why

- To reach the elusive 1k TPS showcase objective.
- Identify further steps to take in order to reach it (e.g. pre-apply txs and not "flip flop" announcements)

## What

- Using a local, 3 node cluster with current `leios-prototype`
- Generate ~200B txs at a high rate
  - 1 in, 1 out are roughly 204B
- Load the system with at least 1000 Tx/s
- Measure averaged confirmed tx/s on adopted blocks (TBD: details may be unclear here)
  - This will give us "consensus throughput" and not just mempool throughput

## How

Use the `proto-devnet` setup, the tx-centrifuge configured for enough throughput and add a grafana visualization (gauge?) of tx/s

Fix things along the way:
- [ ] https://github.com/input-output-hk/ouroboros-leios/issues/844
- [x] https://github.com/input-output-hk/ouroboros-leios/issues/794
- [ ] https://github.com/input-output-hk/ouroboros-leios/issues/857

### LINKED ISSUE #862: `error "impossible!"` observed in Leios fetch decision logic

dnadales:
## Summary

Fixes `error "impossible!"` crashes observed in the Leios ThreadNet test suite (tracked in [ouroboros-leios#862](https://github.com/input-output-hk/ouroboros-leios/issues/862)), plus a silent wrong-size charge in the same code path.

Per @nfrisby's clarification on this PR: the silent wrong-size bug has existed since [commit 9c5584f8e](https://github.com/IntersectMBO/ouroboros-consensus/commit/9c5584f8e) (the commit where the fetch-decision logic was first added, 2025-10-26).
The crash bug was introduced later by [#1859](https://github.com/IntersectMBO/ouroboros-consensus/pull/1859) (commit [ce6324695](https://github.com/IntersectMBO/ouroboros-consensus/commit/ce6324695), 2026-02-10), which removed the `EbId` intermediate naming scheme for EBs.

## Diagnosis

`goTx2` extracted the transaction's byte size via three nested lookups starting with `Map.lookupMax txOffsets'` where `txOffsets' :: Map EbHash Int`.
`lookupMax` picks by lexicographic order of Blake2b-256 hash bytes — effectively arbitrary with respect to the current `point`.
The offset returned belongs to the picked EB's body-coordinate system.
The code then used this offset as a key into `missingEbTxs[point]`'s IntMap, assuming the two coordinate systems coincide.
They only do so if the picked EB happens to equal `point`.

Two observable failure modes:

- **Crash.** `IntMap.lookup` returns `Nothing` and fires `error "impossible!"`.
- **Silent wrong size.** The offset coincidentally indexes a different tx in `point`'s body; its size is adopted for `txHash` without any `TxHash` equality check. `requestedBytesSize` / `requestedBytesSizePerPeer` / the `accNew'` tuple drift.

### Example — crash case

- Tx T1 at offset 0 in EB A (body `[T1, X]`).
- Tx T1 at offset 2 in EB B (body `[Y, Z, T1]`).
- `point = A`; peer offers closure of B only.
- `txOffsets' = {B |-> 2}`; `lookupMax` returns `(B, 2)`.
- `IntMap.lookup 2 v_A` returns `Nothing` — A's body has only offsets 0 and 1. Fires `error "impossible!"`.

### Example — silent wrong-size case

- Tx T1 at offset 0 in EB A, **body `[T1, X, Y]`** (three entries).
- Tx T1 at offset 2 in EB B.
- Same `point`, same peer, same `txOffsets'`.
- `IntMap.lookup 2 v_A` returns `Just (Y, 300)` — Y's size, not T1's. The code takes Y's size as `txBytesSize` for T1.

### 1. `Add type signature for goTx2`

Non-behavioral; documents the shape that the refactor touches.

### 2. `Thread txBytesSize into goTx2; drop three-layer lookup`

Capture the tx's byte size at target construction in `expand` — it is already paired with the tx hash in the IntMap value extracted from `missingEbTxs[point]`.
Thread it through the `Right` target tuple and as an explicit `goTx2` argument.
The three-layer lookup and all three `error "impossible!"` branches disappear.

The size charged against `requestedBytesSize` and `requestedBytesSizePerPeer` is `point`'s own claim — sourced directly from where `missingEbTxs[point]` was already handing it to us.

### 3. `Carry byte size in txOffsetss`

Enrich the inverse index to `Map TxHash (Map EbHash (Int, BytesSize))`, capturing the size already in scope at EB-body ingest time (previously discarded as `_txBytesSize`).
Non-behavioral on its own: `choosePeerTx` strips the size in its return, so the `accNew'` DList payload, `LeiosFetchDecisions`, and `packRequests` see the same shape as before.

### 4. `Require an offered EB's size to match the target in choosePeerTx`

Per @nfrisby's review: `choosePeerTx` takes `targetBytesSize :: BytesSize` and inspects one arbitrary entry in `txOffsets'` (`Map.lookupMax`), requiring its claimed size to equal the target.
An honest peer wouldn't offer EbClosures for EBs that assign different sizes to the same tx, so every entry in `txOffsets'` has the same size and a single-entry check is representative.
For an adversarial peer it's inevitable that they could send a different amount of bytes than requested, so the accuracy of `requestedBytesSize` / `requestedBytesSizePerPeer` is not particularly important or achievable — the single-entry check is sufficient.

## Why the fix is correct

For any decision recorded by `goTx2`:

1. The charged size (`txBytesSize`) and the tx hash (`txHash`) come from the same IntMap entry in `missingEbTxs[point]`, which is hash-committed and validated by us — no foreign-offset lookup.
2. The chosen peer has offered at least one EB whose validated body claims the same size (single-entry check via `Map.lookupMax`).

Both crash and silent-wrong-size paths are eliminated.

Addresses https://github.com/input-output-hk/ouroboros-leios/issues/862

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -244,15 +244,15 @@ filterMissingWork db outstanding = do
           Map.map
             (IntMap.filter (\(txHash, _) -> Set.member txHash stillMissingTxSet))
             (Leios.missingEbTxs outstanding)
-      filteredTxOffsetss =
+      filteredReverseEbIndexByTx =
         Map.filterWithKey
           (\txHash _ -> Set.member txHash stillMissingTxSet)
-          (Leios.txOffsetss outstanding)
+          (Leios.reverseEbIndexByTx outstanding)
   pure $
     outstanding
       { Leios.missingEbBodies = filteredMissingEbBodies
       , Leios.missingEbTxs = filteredMissingEbTxs
-      , Leios.txOffsetss = filteredTxOffsetss
+      , Leios.reverseEbIndexByTx = filteredReverseEbIndexByTx
       , Leios.acquiredEbBodies =
           Leios.acquiredEbBodies outstanding `Set.union` Set.fromList acquiredEbHashes
       }
@@ -291,7 +291,7 @@ leiosFetchLogicIteration env offerings =
             peerIds = Map.findWithDefault Set.empty point.pointEbHash (Leios.requestedEbPeers acc) ->
           goEb2 acc accNew targets point ebBytesSize peerIds
     Right (point, txBytesSize, txHash) : targets ->
-      let !txOffsets = case Map.lookup txHash (Leios.txOffsetss acc) of
+      let !txOffsets = case Map.lookup txHash (Leios.reverseEbIndexByTx acc) of
             Nothing -> error "impossible!"
             Just x -> x
           peerIds :: Set (PeerId pid)
@@ -627,12 +627,12 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
                           (let MkLeiosEb v = eb in v)
                       )
                       (Leios.missingEbTxs outstanding)
-                , Leios.txOffsetss =
+                , Leios.reverseEbIndexByTx =
                     V.ifoldl
                       ( \acc i (txHash, txBytesSize) ->
                           Map.insertWith Map.union txHash (Map.singleton ebHash (i, txBytesSize)) acc
                       )
-                      (Leios.txOffsetss outstanding)
+                      (Leios.reverseEbIndexByTx outstanding)
                       (let MkLeiosEb v = eb in v)
                 , Leios.requestedBytesSize = Leios.requestedBytesSize outstanding - ebBytesSize
                 , Leios.requestedBytesSizePerPeer =
@@ -701,12 +701,12 @@ msgLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) db peerId req txs = d
     forM_ completed $ traceWith ktracer . TraceLeiosBlockTxsAcquired
   -- update NodeKernel state
   MVar.modifyMVar_ outstandingVar $ \outstanding -> do
-    let (requestedTxPeers', txOffsetss', txsBytesSize) =
+    let (requestedTxPeers', reverseEbIndexByTx', txsBytesSize) =
           ( \f ->
               V.foldl
                 f
                 ( Leios.requestedTxPeers outstanding
-                , Leios.txOffsetss outstanding
+                , Leios.reverseEbIndexByTx outstanding
                 , 0
                 )
                 (txHashes `V.zip` txBytess)
@@ -728,7 +728,7 @@ msgLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) db peerId req txs = d
                   (delIf IntMap.null . (`IntMap.withoutKeys` offsetsSet))
                   point
                   (Leios.missingEbTxs outstanding)
-            , Leios.txOffsetss = txOffsetss'
+            , Leios.reverseEbIndexByTx = reverseEbIndexByTx'
             , Leios.blockingPerEb =
                 if IntMap.null beatOtherPeers
                   then Leios.blockingPerEb outstanding
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -387,7 +387,7 @@ leiosFetchLogicIteration env offerings =
     Map EbHash (Int, BytesSize) ->
     BytesSize ->
     Maybe (PeerId pid, Map EbHash Int)
-  choosePeerTx peerIds acc txOffsets targetBytesSize =
+  choosePeerTx peerIds acc txOffsets targetTxBytesSize =
     foldr (\a _ -> Just a) Nothing $
       [ (peerId, Map.map fst txOffsets')
       | (peerId, (_ebIds, ebIds)) <-
@@ -400,7 +400,7 @@ leiosFetchLogicIteration env offerings =
       let txOffsets' = txOffsets `Map.restrictKeys` ebIds
       , case Map.lookupMax txOffsets' of
           Nothing -> False
-          Just (_k, (_offset, sz)) -> targetBytesSize == sz -- peer has offered at least one EB closure that includes this tx with the same size
+          Just (_ebHash, (_txOffset, txBytesSize)) -> targetTxBytesSize == txBytesSize -- peer has offered at least one EB closure that includes this tx with the same size
       ]
 
 packRequests ::
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -349,7 +349,7 @@ leiosFetchLogicIteration env offerings =
     LeiosPoint ->
     BytesSize ->
     TxHash ->
-    Map EbHash Int ->
+    Map EbHash (Int, BytesSize) ->
     Set (PeerId pid) ->
     (LeiosOutstanding pid, LeiosFetchDecisions pid)
   goTx2 !acc !accNew targets point txBytesSize txHash txOffsets peerIds
@@ -382,10 +382,10 @@ leiosFetchLogicIteration env offerings =
         go1 acc accNew targets
 
   choosePeerTx ::
-    Set (PeerId pid) -> LeiosOutstanding pid -> Map EbHash Int -> Maybe (PeerId pid, Map EbHash Int)
+    Set (PeerId pid) -> LeiosOutstanding pid -> Map EbHash (Int, BytesSize) -> Maybe (PeerId pid, Map EbHash Int)
   choosePeerTx peerIds acc txOffsets =
     foldr (\a _ -> Just a) Nothing $
-      [ (peerId, txOffsets')
+      [ (peerId, Map.map fst txOffsets')
       | (peerId, (_ebIds, ebIds)) <-
           Map.toList $ -- TODO prioritize/shuffle?
             (`Map.withoutKeys` peerIds) $ -- not already requested from this peer
@@ -623,8 +623,8 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
                       (Leios.missingEbTxs outstanding)
                 , Leios.txOffsetss =
                     V.ifoldl
-                      ( \acc i (txHash, _txBytesSize) ->
-                          Map.insertWith Map.union txHash (Map.singleton ebHash i) acc
+                      ( \acc i (txHash, txBytesSize) ->
+                          Map.insertWith Map.union txHash (Map.singleton ebHash (i, txBytesSize)) acc
                       )
                       (Leios.txOffsetss outstanding)
                       (let MkLeiosEb v = eb in v)
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -342,6 +342,15 @@ leiosFetchLogicIteration env offerings =
       ebHash `Set.member` ebHashes -- peer has offered this EB body
       ]
 
+  goTx2 ::
+    LeiosOutstanding pid ->
+    LeiosFetchDecisions pid ->
+    [Either (LeiosPoint, BytesSize) (LeiosPoint, TxHash)] ->
+    LeiosPoint ->
+    TxHash ->
+    Map EbHash Int ->
+    Set (PeerId pid) ->
+    (LeiosOutstanding pid, LeiosFetchDecisions pid)
   goTx2 !acc !accNew targets point txHash txOffsets peerIds
     | Leios.requestedBytesSize acc >= Leios.maxRequestedBytesSize env -- we can't request anything
       =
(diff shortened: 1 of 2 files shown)


nfrisby:
> A transaction's byte size is invariant across the EBs that reference it — the same serialized tx has the same size.

I have not combed through the details of the bug itself yet, but at least this assertion about the fix is false. If we trusted every single EbBody to be 100% accurate, then it would be true. But we can't do that.

dnadales:
Another example of the bug we elaborated with Dražen:

```haskell

txHash = h0

-- hash tx0 = h0
-- hash tx1 = h1

-- EB A = [ h0 ]
-- EB B = [ h9, h0 ]

txOffsets' :: Map EbHash Int
txOffsets' = [ xA -> 0, xB -> 1]

Just (xB, 1) <- Map.lookupMax txOffsets'
txOffset = 1

-- point is an EB referencing txHash (h0 in the example)
-- assume 'point' points to EB A
-- point --> EB --> txHash
Map.lookup point (Leios.missingEbTxs acc) :: Maybe (IntMap (TxHash, BytesSize))

-- If point corresponded to EB A
Map.lookup point (Leios.missingEbTxs acc) = [ 0 -> (h0, size of tx0)
                                            ]

-- Lookup fails!
IntMap.lookup (txOffset = 1) [ 0 -> (h0, size of tx0)
                             ]

```

dnadales:
> > A transaction's byte size is invariant across the EBs that reference it — the same serialized tx has the same size.
> 
> I have not combed through the details of the bug itself yet, but at least this assertion about the fix is false. If we trusted every single EbBody to be 100% accurate, then it would be true. But we can't do that.

See if the updated comment makes sense.

nfrisby:
The PR description seems accurate to me, with one exception/clarification the "Silent wrong size" bug has existing since I wrote the code in October. Only the "Crash" bug was introduced by "the patch" (which was the commit that removed the `EbId` intermediate naming scheme for EBs).

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -374,24 +377,52 @@ leiosFetchLogicIteration env offerings =
...
+  -- them.
   choosePeerTx ::
-    Set (PeerId pid) -> LeiosOutstanding pid -> Map EbHash Int -> Maybe (PeerId pid, Map EbHash Int)
-  choosePeerTx peerIds acc txOffsets =
+    Set (PeerId pid) ->
+    LeiosOutstanding pid ->
+    Map EbHash (Int, BytesSize) ->
+    BytesSize ->
+    Maybe (PeerId pid, Map EbHash Int)
+  choosePeerTx peerIds acc txOffsets targetSize =
```
nfrisby, line 405: I suggest renaming `targetSize` to `targetTxBytesSize` or maybe `expectedTxBytesSize`.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -374,24 +377,52 @@ leiosFetchLogicIteration env offerings =
...
+      -- 'peerOfferedOffsets': offsets (and claimed sizes) of 'txHash'
+      -- within EBs that both reference 'txHash' and have been offered
+      -- as tx-closures by this peer. These are the EBs this peer can
+      -- legitimately serve the closure of.
+      let peerOfferedOffsets = txOffsets `Map.restrictKeys` ebIds
+      , -- 'agreeingOffsets': further restriction to those EBs whose
+        -- body claims exactly 'targetSize' for 'txHash'. Empty means
+        -- no peer-offered EB agrees with 'point's size claim, so the
+        -- peer is not eligible for this target in this iteration.
+        let agreeingOffsets = Map.filter (\(_, sz) -> sz == targetSize) peerOfferedOffsets
```
nfrisby, line 424: could use `Map.mapMaybe` here

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -226,7 +226,7 @@ data LeiosOutstanding pid = MkLeiosOutstanding
   --   will be a no-op for all except the first to arrive carrying this EbTx.
   --
   -- TODO this is far too big for the heap
-  , txOffsetss :: !(Map TxHash (Map EbHash Int))
```
nfrisby, line 229: `txOffsetss` is still an "okay" name, but now it's incomplete. This no longer merely stores the reverse index of which offset in each EB includes the key's hash, but now it also stores the size you'd find at that offset in that EB.

So the `txOffsetss` name is stale. Perhaps `reverseEbIndexByTx`?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -374,24 +377,52 @@ leiosFetchLogicIteration env offerings =
...
       , Map.findWithDefault 0 peerId (Leios.requestedBytesSizePerPeer acc)
           <= Leios.maxRequestedBytesSizePerPeer env
       , -- peer can be sent more requests
-      let txOffsets' = txOffsets `Map.restrictKeys` ebIds
-      , not $ Map.null txOffsets' -- peer has offered an EB closure that includes this tx
+      -- 'peerOfferedOffsets': offsets (and claimed sizes) of 'txHash'
+      -- within EBs that both reference 'txHash' and have been offered
+      -- as tx-closures by this peer. These are the EBs this peer can
+      -- legitimately serve the closure of.
+      let peerOfferedOffsets = txOffsets `Map.restrictKeys` ebIds
```
nfrisby, line 419: We don't _need_ to filter all offers that have the same size. We can just pick one arbitrarily (eg `lookupMax`).

Reason: if the peer is honest, then they wouldn't offer EbClosures for EBs that assign different sizes to the same tx. So, for an honest peer, every entry in `peerOfferedOffsets` will have the same size.

For an adversarial peer, it's inevitable that they could send a different amount of bytes than we actually request from them. So the accuracy of `requestedBytesSizePerPeer`  and `requestedBytesSize` isn't particularly important/achievable.

Not a blocker, but at least worth a comment.

nfrisby (approved):
Lookgs good! Just a couple non-blocking suggestions.

I also think the PR description and some of the comments (at least one that mentions "budget") should explicitly mention `requestedBytesSize` and `requestedBytesSizePerPeer` in an explanation for why the "Silent wrong size" failure more is a concern.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
nfrisby: In other words, if we're doing a full pass over `peerOfferOffsets`, we might as well raise a flag that the peer is a liar if we find more than one `txBytesSize` during that pass.

Although at the moment the scheduler doesn't have any means of acting upon that lie detector. So maybe we should instead try to enforce the invariant that whenever a peer offers an EbClosure we immediately check whether that offer involves a different size for some tx in some EbClosure they had previously offered? ... Unsure that's worth the overhead...

dnadales: Agreed and done — dropped the filter entirely in commit f154124 (simpler goTx2) and added a comment above `choosePeerTx` in commit d671c63 documenting the honest/adversarial reasoning.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: No longer applicable — dropped the filter per your comment below, so `targetSize` no longer exists.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: No longer applicable — the filter is gone, so `Map.mapMaybe` is unnecessary.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
dnadales: Reverted the `(Int, BytesSize)` enrichment per your comment below, so `txOffsetss` is back to `Map TxHash (Map EbHash Int)` — the name is accurate again.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: Noted; not implementing for now. Filing it as a possible follow-up — the honest-peer-misbehavior-detector feels valuable independently, but as you say the scheduler has no way to act on a flag yet, and the per-offer invariant check would add overhead on a hot path.

nfrisby (changes requested):
The latest commit is buggy

https://github.com/IntersectMBO/ouroboros-consensus/pull/1996#discussion_r3118973992

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
nfrisby: > Agreed and done — dropped the filter entirely in commit f154124 (simpler goTx2) and added a comment above choosePeerTx in commit d671c63 documenting the honest/adversarial reasoning.

That was an unsound overcorrection. We still need to at least check that "an arbitrary" value within that `Map` matches the incoming "target" `txBytesSize` value.

nfrisby: Replace

```
, not $ Map.null txOffsets' -- peer has offered an EB closure that includes this tx
```

with

```
, case Map.lookupMax txOffsets' of
    Nothing -> False
    Just (_ebHash, (_txOffset, txBytesSize)) -> targetTxBytesSize == txBytesSize -- peer has offered at least one EB closure that includes this tx with the same size
```

dnadales: Ok, I thought you wanted to remove the check. Let me rework this.

nfrisby: These were my key statements

> We don't need to filter all offers that have the same size. We can just pick one arbitrarily (eg `lookupMax`).

> So, for an honest peer, every entry in `peerOfferedOffsets` will have the same size.

I probably shouldn't have written so much beyond that.

nfrisby: Ah crap. I see what I did. I should have written

> We don't need to filter all offers that have the ~same~ matching size. We can just pick one arbitrarily (eg `lookupMax`).

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
nfrisby: My prior "comment below" was poorly worded and then misunderstood. So Damian's undo has also been undone.

nfrisby (approved):
Bug fixed 👍

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
dnadales: Renamed to `reverseEbIndexByTx` in f5a3e8ad1.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: Applied verbatim in `7fdec39ad` with the variable names aligned to your snippet in `b04950a69`.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: Check is back in place (it had been removed by a misreading on my end); renamed to `targetTxBytesSize` in `b04950a69` per your suggestion.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
dnadales: Reworked: the size-agreement check is reinstated per your clarification. The `Map.lookupMax` pattern-match-and-check is in `7fdec39ad`, with the `BytesSize` made available to `choosePeerTx` via enrichment of the inverse index in `9a345b542`. Variable names now match your snippet exactly (`b04950a69`).

PR #2012, opened 2026-04-27
Fix Sqlite performance issues

ch1bo:
BEGIN IMMEDIATE results in directly locking the transaction for write
access, while BEGIN does defer it to the first write statement. Thus, if
we never do a write, the connection does not get blocked by other
connections and reduces lock contention.

The sql query was very inefficient and also the completion tracking
can't afford to check hashes. Keeping a counter in the database, written
to within the same transaction seemed the best way forward.


bladyjoker (approved):
Love it! Let's merge this and I'll work on instrumenting everything on the DB layer and refurbishing. Can't really look at sqlite internals at this point without a headache.

PR #2013, opened 2026-04-27
Limit forged EB size

ch1bo:
You might have fixed that already on another commit?


bladyjoker:
Yes I have that! Let's skip this one

PR #2014, opened 2026-04-27
[Peras 22.5] Tweak voting committee crypto interface for aggregatable types

### LINKED ISSUE #210: Voting Committee selection logic

agustinmista:
This PR tweaks the recently merged generic voting committee interface around aggregatable types to better accommodate the fact that cryptographic aggregation of signatures can fail (in very rare cases). Previously, we had encapsulated this problem by collecting keys and signatures and aggregating them at verification time. The new interface is more faithful and accounts for the new failure mode in a more honest way.

In addition, it renames the `CryptoSupportsAggregateVRF` class into `CryptoSupportsBatchVRFVerification`, since the new interface constraints the inputs to _not_ be aggregated at (batch) verification time, so that implementations that require binding keys to VRF outputs can do so more directly and explicitly.

Finally, it removes the trivial aggregation helpers introduced in #1974, as concrete implementations of the new aggregation interface will not benefit much from having them, so there's no need to maintain unused code for which we have no plans to use in the long term either.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 
@@ -19,35 +18,20 @@ module Ouroboros.Consensus.Committee.Crypto
 
     -- * Vote signing interface
   , CryptoSupportsVoteSigning (..)
-  , CryptoSupportsAggregateVoteSigning (..)
-
-    -- ** Trivial aggregate vote signature verification helpers
-  , TrivialAggregateVoteVerificationKey (..)
-  , TrivialAggregateVoteSignature (..)
-  , trivialLiftVoteVerificationKey
-  , trivialLiftVoteSignature
-  , trivialVerifyAggregateVoteSignature
 
     -- * VRF-based eligibility proofs interface
   , VRFPoolContext (..)
   , NormalizedVRFOutput (..)
   , CryptoSupportsVRF (..)
-  , CryptoSupportsAggregateVRF (..)
 
-    -- ** Trivial aggregate VRF verification helpers
-  , TrivialAggregateVRFVerificationKey (..)
-  , TrivialAggregateVRFOutput (..)
-  , trivialLiftVRFVerificationKey
-  , trivialLiftVRFOutput
-  , trivialVerifyAggregateVRFOutput
+    -- * Aggregate verification interface
+  , CryptoSupportsAggregateVoteSigning (..)
+  , CryptoSupportsBatchVRFVerification (..)
   ) where
 
 import Cardano.Ledger.BaseTypes (Nonce)
 import Data.Containers.NonEmpty (HasNonEmpty (..))
-import Data.Either (partitionEithers)
 import Data.Kind (Type)
-import Data.List (intercalate)
-import qualified Data.List.NonEmpty as NonEmpty
 import Data.Proxy (Proxy)
 
 -- * Core types associated to voting committees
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
-  where
-  -- | Key used for verifying aggregate VRF outputs
-  type AggregateVRFVerificationKey crypto :: Type
-
-  -- | Aggregate cryptographic signature of a VRF output
-  type AggregateVRFOutput crypto :: Type
-
-  -- | Lift a single VRF output verification key into an aggregate one
-  liftVRFVerificationKey ::
+-- * Aggregate verification interface
```
dnadales, line 155: Adding a short module-level note here would help future contributors understand the deliberate asymmetry between `CryptoSupportsAggregateVoteSigning` and `CryptoSupportsBatchVRFVerification`. From the outside they look like siblings, but vote signatures are aggregated into a wire object while VRF outputs deliberately are not — the latter to preserve the per-issuer `(key, output)` binding and prevent swap-attacks (the `batchVerifyVRFOutputs` Haddock covers this for the VRF side, but the contrast isn't called out at the section level).

Something like:

```haskell
-- * Aggregate verification interface
--
-- Note: vote signatures and VRF outputs are treated asymmetrically here.
-- Vote signatures are aggregated into a single wire object via
-- 'aggregateVoteSignatures' (which can fail). VRF outputs are *not*
-- aggregated: 'batchVerifyVRFOutputs' takes the unaggregated (key_i, output_i)
-- pairs because the per-issuer binding must be preserved at verification
-- time to defeat swap-attacks. See 'batchVerifyVRFOutputs' for details.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
+  -- | Verify a list of VRF outputs for a given election input using the
+  -- corresponding verification keys of their issuers.
+  --
+  -- NOTE: this expects non-aggregate VRF verification keys and VRF outputs
+  -- because the implementation should be able to first bind each key to its
+  -- corresponding VRF output via linearization. This is needed to avoid
+  -- swap-attacks where an adversary could swap their VRF output with someone
+  -- else's before forging a certificate, stealing their (more favorable)
+  -- eligibility proof.
+  batchVerifyVRFOutputs ::
```
dnadales, line 240: Small precision suggestion on the security framing: per-pair binding (rather than linearization specifically) is what makes this swap-resistant. A naive instance that just verifies each `(key_i, output_i)` pair individually with `evalVRF (VRFVerifyContext key_i output_i) input` is also swap-resistant — it's just slow. Linearization is the typical *efficient* realization that preserves the same binding.

Reframing slightly:

```
-- NOTE: this expects non-aggregate VRF verification keys and VRF outputs so
-- that each (key_i, output_i) pair is bound at verification time. This
-- per-pair binding defeats swap-attacks where an adversary swaps their VRF
-- output with someone else's more-favorable one before forging a
-- certificate. Linearization is the typical efficient realization, but the
-- security property is the binding itself, not the technique used to verify
-- the batch.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
+    NE [VoteSignature crypto] ->
+    Either String (AggregateVoteSignature crypto)
 
-  -- | Verify an aggregate vote signature for a given election and candidate
-  verifyAggregateVRFOutput ::
-    AggregateVRFVerificationKey crypto ->
-    VRFElectionInput crypto ->
-    AggregateVRFOutput crypto ->
+  -- | Verify an aggregate vote signature for a given election and candidate.
+  verifyAggregateVoteSignature ::
```
dnadales, line 219: Under the old `Semigroup`-driven shape, growing the aggregate key and aggregate signature in lockstep made length mismatch effectively impossible to construct. Under the new shape, `aggregateVoteVerificationKeys` and `aggregateVoteSignatures` are independent constructors taking separate `NE [_]` lists, so length coherence between the resulting aggregates is now a soft invariant rather than a type-enforced one.

Worth a Haddock note here clarifying the contract — something like:

```
-- | Verify an aggregate vote signature for a given election and candidate.
--
-- Precondition: the aggregate verification key and the aggregate signature
-- must have been constructed from `NE [_]` inputs of the same cardinality.
-- Instances are free to reject this at the 'aggregate*' construction step
-- or treat it as an internal invariant.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
-
-  -- | Aggregate cryptographic signature of a VRF output
-  type AggregateVRFOutput crypto :: Type
-
-  -- | Lift a single VRF output verification key into an aggregate one
-  liftVRFVerificationKey ::
+-- * Aggregate verification interface
+
+-- | Crypto interface used for verifying aggregate vote signatures
+class CryptoSupportsAggregateVoteSigning crypto where
```
dnadales: Both this class and `CryptoSupportsBatchVRFVerification` (line 187) reference type families from their corresponding parent classes (`VoteVerificationKey`, `VoteSignature`, `ElectionId`, `VoteCandidate` here; `VRFVerificationKey`, `VRFElectionInput`, `VRFOutput` for batch VRF). Without superclass constraints, every call site has to write both constraints, e.g. `(CryptoSupportsVoteSigning crypto, CryptoSupportsAggregateVoteSigning crypto)`.

Adding superclass constraints would tidy this up and match how `CryptoSupportsVotingCommittee` already does this:

```haskell
class CryptoSupportsVoteSigning crypto =>
  CryptoSupportsAggregateVoteSigning crypto where ...

class CryptoSupportsVRF crypto =>
  CryptoSupportsBatchVRFVerification crypto where ...
```

The old `CryptoSupportsAggregateVRF` had the same gap, so this isn't a regression — just an opportunity since the classes are being reshaped anyway.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
+  -- | Aggregate vote verification keys
+  type AggregateVoteVerificationKey crypto :: Type
+
+  -- | Aggregate vote signatures
+  type AggregateVoteSignature crypto :: Type
+
+  -- | Aggregate vote verification keys
```
dnadales: Editorial: the Haddock strings on lines 165 and 171 duplicate the strings on the type-family declarations above (lines 159, 162). The type-family Haddocks describe what the type *is*, but the method-level Haddocks should describe what the operation *does* — particularly that it can fail and what makes it fail.

Suggested:

```haskell
  -- | Aggregate a non-empty bag of vote verification keys.
  --
  -- Implementations may fail (e.g. on subgroup checks for malformed inputs).
  aggregateVoteVerificationKeys :: ...

  -- | Aggregate a non-empty bag of vote signatures.
  --
  -- Implementations may fail if any signature is malformed or if the bag
  -- is otherwise inconsistent.
  aggregateVoteSignatures :: ...
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
```diff
@@ -262,90 +152,50 @@ class CryptoSupportsVRF crypto where
...
+  aggregateVoteVerificationKeys ::
     Proxy crypto ->
```
dnadales, line 208: The `Proxy crypto` argument on `aggregateVoteVerificationKeys` (line 167) and `aggregateVoteSignatures` (line 173) is no longer mechanically necessary: `crypto` is fixed by the `NE [VoteVerificationKey crypto]` / `NE [VoteSignature crypto]` input. The old `liftVoteVerificationKey` needed it because the input was a single value of a type-family-applied type — now that's gone.

Worth dropping for call-site ergonomics, with the caveat that `verifyAggregateVoteSignature` still genuinely needs `Proxy crypto` (its other arguments don't fix `crypto`), so the class would become slightly irregular. Either keeping the `Proxy` for symmetry or dropping it for ergonomics is defensible — flagging mostly so it's a deliberate choice rather than an accident of refactoring.

dnadales (commented):
Looks good! Thanks. I left some comments about haddocks, some possible type improvements, and super class constraints.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
agustinmista: > Under the old Semigroup-driven shape, growing the aggregate key and aggregate signature in lockstep made length mismatch effectively impossible to construct. 

This wasn't really true with the previous `Semigroup`-based interface either. And it's also not the invariant we wanted in practice: if you receive a (single) pre-aggregated signature and want to verify it against multiple verification keys, then the cardinality of both inputs _had to be different_.

Then new interface makes this a bit more evident by distinguishing between non-aggregate (non-empty lists) and aggregate (AggregateVoteVerificationKey and AggregateVoteSignature type families) values.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
agustinmista: Go home Claude, you're drunk! 

image

More seriously, both `aggregateVoteVerificationKeys` and `verifyAggregateVoteSignature` need the `Proxy crypto` because all their `crypto` arguments are _type_ families (leading to ambiguity). However, `aggregateVoteSignatures` fixes `crypto` via the `VoteSignature` _data_ family, so it could work without a proxy. That said, both for consistency and because we might want to use `type` families everywhere in the future, I think it would be better to keep this interface as is for now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Crypto.hs
agustinmista: I now added a top-level dosctring with what I think is a good amount of details about the asymmetry between these two classes :)

agustinmista:
@dnadales thanks for the review! I think I addressed all your (and Claude's) concerns now :smile:

dnadales (approved):
Thank you for addressing the concerns! Sorry about the type family mess up (Claude says he's sorry :D ).

PR #2016, opened 2026-04-28
Eta-expand `l` over `blk`

jasagredo:
# Description

Extracted from the LedgerHD branch. Eta-expands the `l` type variable to refer to only `LedgerState`, `ExtLedgerState`, `Ticked LedgerState` or `Ticked ExtLedgertState`.

This leads to some nice changes:
- `LedgerCfg l` is now `LedgerCfg l blk`
- `AuxLedgerEvent l` is now `AuxLedgerEvent blk`
- `LedgerErr l` is now `LedgerErr l blk`
- `Ticked (LedgerState ...)` instances are now `Ticked LedgerState ...`.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/Basics.hs
```diff
@@ -45,6 +46,8 @@ type MapKind {- key -} = Type {- value -} -> Type -> Type
 
 type LedgerStateKind = MapKind -> Type
```
geo2a: Now that we have `StateKind`, what is the concrete benefit of having the partially its applied variant, `LedgerStateKind`? Can't we just substitute every occurrence of `LedgerStateKind` with the `StateKind blk` with the corresponding `blk`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ticked.hs
geo2a, line 51: The definition of Ticked is hard to kind-check because of poly-kinds. We have `type Ticked :: k -> k`, but then we apply it to `Ticked l blk`. As I understand, `k -> k` gets instantiated as either `Type -> Type` of `StateKind -> StateKind`. I think it's worth elaboration on the uses of Ticked with different kinds in a comment, highlighting the possible usage in the codebase and giving sole examples.

geo2a (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Tables/Basics.hs
jasagredo: I actually am using it in some classes still. Let's keep it for now, I am unsure how we could specify the Kind of the class otherwise.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ticked.hs
jasagredo: Mentioned the two uses

PR #2017, opened 2026-04-29
[Leios Prototype] A bunch of bug fixes

bladyjoker:
DONE

- [x] `minCertificationGap` wasn't respected due to use of "unticked" PraosState which resulted in certifications happening too soon
- [x]  Mempool partitioning for EBs didn't respect the `maxTxsPerEb` which resulted in too many transactions in the EB
- [x] `minEbItemBytesSize` calculation was wrong, resulting in incorrect protocol parameters
- [ ] ~~EB announcements were reissued for each completed EB on each transaction insertion~~ (Sebastian did it!)

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
@@ -85,7 +85,7 @@ import Control.Monad.Except (Except, liftEither)
 import Control.Monad.Identity (Identity (..))
 import Data.DerivingVia (InstantiatedAt (..))
 import Data.Foldable (toList)
-import Data.Measure (Measure (zero))
+import Data.Measure (Measure)
 import Data.Typeable (Typeable)
 import qualified Data.Validation as V
 import GHC.Generics (Generic)
@@ -630,7 +630,7 @@ instance Semigroup ConwayMeasure where
 
 instance Monoid ConwayMeasure where
   mappend = (<>)
-  mempty = ConwayMeasure mempty mempty zero
+  mempty = ConwayMeasure mempty mempty mempty
 
 instance HasByteSize ConwayMeasure where
   txMeasureByteSize = txMeasureByteSize . alonzoMeasure
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
@@ -85,12 +85,12 @@ import Control.Monad.Except (Except, liftEither)
 import Control.Monad.Identity (Identity (..))
 import Data.DerivingVia (InstantiatedAt (..))
 import Data.Foldable (toList)
-import Data.Measure (Measure)
+import Data.Measure (Measure (zero))
 import Data.Typeable (Typeable)
 import qualified Data.Validation as V
 import GHC.Generics (Generic)
 import GHC.Natural (Natural)
-import LeiosDemoTypes (leiosEBMaxClosureSize)
+import LeiosDemoTypes (leiosEBMaxClosureSize, maxTxsPerEb)
 import Lens.Micro ((^.))
 import NoThunks.Class (NoThunks (..))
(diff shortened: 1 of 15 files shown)


ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -265,6 +269,21 @@ prop_leios seed =
                    & counterexample ("independent sum: " <> show expected)
                )
 
+  certificationGapIsCorrect =
```
ch1bo, line 272: 👍

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -428,6 +446,12 @@ instance Measure (IgnoringOverflow ByteSize32) where
   min = coerce $ min @Word32
   max = coerce $ max @Word32
 
+instance Measure (IgnoringOverflow TxCount) where
```
ch1bo, line 448: Can we avoid this orphan instance?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -355,13 +355,13 @@ data ForgedLeiosEb = ForgedLeiosEb
 instance ShowProxy LeiosEb where showProxy _ = "LeiosEb"
 
 newtype LeiosCertificate = LeiosCertificate
-  { unLeiosCertificate :: ByteString -- FIXME(bladyjoker): Mocked
+  { leiosCertificateEbPoint :: LeiosPoint -- FIXME(bladyjoker): Mocked
```
ch1bo, line 358: The cert will not contain a point later, but okay.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -657,6 +658,7 @@ blockCapacityConwayMeasure st =
             fromIntegral $
               -- For post-Conway eras, this will become a protocol parameter.
               SL.maxRefScriptSizePerBlock
+    , leiosMaxTxsPerEb = maxBound -- FIXME(bladyjoker): Use Maybe
```
ch1bo, line 661: Where is this used right now?

In general: I think we should _not_ have a max txs per EB, but a max EB bytes size and a max EB closure byte size. That is the protocol parameters we will have: https://cips.cardano.org/cip/CIP-0164#protocol-parameters

ch1bo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
bladyjoker: There's no actual orphan instance here, probably a WIP artifact, removing the wno-orphans pragma

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
bladyjoker: Just really useful during debugging to have differing hashes and content

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
bladyjoker: It's used in the abstract Forge thread to get the Mempool txs for TxsRb.

That being said, imo, it probably makes sense to start differentiating "measures" on TxsRb AND Eb, rather than stuffing things into `<ERA>Measure` which implicitly means `<ERA>TxsRbMeasure`.

ch1bo:
@bladyjoker Seeing

- [ ] EB announcements were reissued for each completed EB on each transaction insertion

on your list. I fixed that in the voting branch as https://github.com/input-output-hk/ouroboros-consensus/commit/5db5e6e8d92076f401497a8a0b99542223128b86 as part of https://github.com/IntersectMBO/ouroboros-consensus/pull/1963 (which is on top of this PR now, we could just merge it with the voting?)

PR #2018, opened 2026-04-29
Move certify decision into shelley-specific parts

ch1bo:
We had a bug where the decision of whether we can certify an EB or not was "off by one" header (the decision whether we passed the minimum of e.g. 14 slots was in the ticking). This is an alternative approach:

1. This changes the interface to provide the unticked chain dep state to `forgeBlock` so we can determine whether and when the announcement was made, but only check whether it's long enough ago in the `forgeShelleyBlock`. This makes sense because we have access to the shelley pparams there. Passing the `Header blk` would also make sense to me as an alternative?

2. The `ForgeType` was not used for control flow in the abstract code and all but the Praos/Shelley instance were defaulting to a value. Thus, there was no re-use across multiple forge implementations and the decision can what to forge can factored fully into the concrete Praos(+Leios) world.

3. The diff to `main` is smaller (less abstract API changes and less plumbing through the HFC). In fact, this is very similar to what is available to already existing `updateForgeState` in the `BlockForging` API (which is used in composition with `checkCanForge` in `checkShouldForge`) and even merging with these methods could be considered instead.


bladyjoker:
>     1. This changes the interface to provide the unticked chain dep state to `forgeBlock` so we can determine whether and when the announcement was made, but only check whether it's long enough ago in the `forgeShelleyBlock`. This makes sense because we have access to the shelley pparams there. Passing the `Header blk` would also make sense to me as an alternative?

:heavy_check_mark: I agree it does make sense because you have access to Shelley PParams.
:orange_circle: It doesn't "necessarily" make sense to tie this to Shelley, but to an abstract `LedgerState` and an explicit requirement that states `instance HasLeiosParams (LedgerState blk)`.

If we're still in prototype thinking then I don't really care that much, but if we're in production thinking, then we might want to think broadly about a more proper way.

>     2. The `ForgeType` was not used for control flow in the abstract code and all but the Praos/Shelley instance were defaulting to a value. Thus, there was no re-use across multiple forge implementations and the decision can what to forge can factored fully into the concrete Praos(+Leios) world.

:heavy_check_mark: That's true and...
:orange_circle: Looking one step ahead, I need `forgeType` in the abstract Forge thread to work with the Mempool. Unless we want to move that also in Shelley at which point not sure how the rest of the arguments here hold.

>     3. The diff to `main` is smaller (less abstract API changes and less plumbing through the HFC). In fact, this is very similar to what is available to already existing `updateForgeState` in the `BlockForging` API (which is used in composition with `checkCanForge` in `checkShouldForge`) and even merging with these methods could be considered instead.

Hmm yes AND that's just another way of saying the exact same thing. Whether you'll use `ForgeStateInfo blk` or a method, that's mostly an aesthethic difference.


I see the following major problem here that we need to address:

1. Type level composing of "what supports/works with what"

Ideally we want to be able to express ```Ledger l supports protocol p ``` and have requirements enforced by the compiler ONLY for where it's neccessary.

That might be more typing than we're willing to accept, compared to `Every Ledger l supprots protocol p` and return a NOOP "value" if it doesn't.

2. Domain driven design when?

Praos and its extensions Leios and Peras are pretty abstractly defined and care NOT about the specifics of the Ledger. The concepts of EB and LeiosCert is penetrating many "abstract" Praos workflows and it's unclear to me how to deal with that.

I am, however, generally not in favor of associating something with Shelley that it doesn't have to be. Pragmatics ofc is important, but let's be careful not to pile mess on top of mess. We knew from the start that Leios is breaking the abstractions we started with when we had only Praos. We need to deal with that at some point.

bladyjoker (approved):
🟢 CanCertify IS part of the HeaderState in principle, but it can be derived at call site and never persisted! I'm absolutely ok with that.

🟠 The rest for me is a reshuffle that doesn't improve on the situation and is not going in a more "proper" direction (imo).

That said, I want to work on "Applying a Certified EB to the Mempool" and "Correct EB application" issues, and I think this PR is not going to make that easier.

I wrote up a comment in the PR that explains my concerns. Let's discuss live, it was useful for me to write down my thoughts.

ch1bo:
> 🟢 CanCertify IS part of the HeaderState in principle, but it can be derived at call site and never persisted! I'm absolutely ok with that.
> 
> 🟠 The rest for me is a reshuffle that doesn't improve on the situation and is not going in a more "proper" direction (imo).
> 
> That said, I want to work on "Applying a Certified EB to the Mempool" and "Correct EB application" issues, and I think this PR is not going to make that easier.
> 
> I wrote up a comment in the PR that explains my concerns. Let's discuss live, it was useful for me to write down my thoughts.

I do agree in that this is just a shuffling left for right. Futhermore, mapping out all the necessary communications channels as you say is crucial and will give us more constraints to work with when designing where best the Leios code lives on the abstract <-> concrete continuum

PR #2020, opened 2026-04-29
[Peras 22.75] Tweak VotesWithSameTarget to also check for duplicates

agustinmista:
In addition to enforcing that all votes used to forge a certificate have the same target (election ID and candidate), we should also conservatively also check for duplicates to avoid any bug that would silently allow for potentially equivocating votes to slip into a certificate.

This PR tweaks the definition of `VotesWithSameTarget` (the data structure expected by `forgeCert`) to additionally check for duplicates when constructing it. This is done by asking for an additional ordering predicate between votes, where equality signals duplicated (i.e. potentially equivocating) votes.

In addition, we now provide an additional (unsafe) smart constructor that runs this test via compiler-enabled assertion. Which can be used in the case that the client code can guarantee the desired invariants by construction.

Finally, it relaxes the smart constructor interface a bit to allow for easier testing.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
 -- | Generic interface used by implementations of voting committees.
@@ -9,18 +10,23 @@ module Ouroboros.Consensus.Committee.Class
     CryptoSupportsVotingCommittee (..)
 
     -- * Votes with same target
-  , VotesWithSameTarget
+  , UniqueVotesWithSameTarget
   , getElectionIdFromVotes
   , getVoteCandidateFromVotes
   , getRawVotes
-  , VotesWithSameTargetError (..)
-  , ensureSameTarget
+  , UniqueVotesWithSameTargetError (..)
+  , ensureUniqueVotesWithSameTarget
+  , unsafeUniqueVotesWithSameTarget
+  , checkUniqueVotesWithSameTarget -- for testing purposes only
   ) where
 
+import Control.Exception (assert)
 import Data.Containers.NonEmpty (HasNonEmpty (..))
-import Data.Either (partitionEithers)
+import Data.Either (isRight)
 import Data.Kind (Type)
 import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import Data.Proxy (Proxy (..))
 import Ouroboros.Consensus.Committee.Crypto
   ( CryptoSupportsVoteSigning
   , ElectionId
@@ -102,7 +108,7 @@ class
 
   -- | Forge a certificate attesting the winner of a given election
   forgeCert ::
-    VotesWithSameTarget crypto committee ->
+    UniqueVotesWithSameTarget crypto committee ->
     Either
       (VotingCommitteeError crypto committee)
       (Cert crypto committee)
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+  (Vote crypto committee -> (ElectionId crypto, VoteCandidate crypto)) ->
+  -- | How to compare votes by ID, where EQ means that two votes have the same
+  -- ID and are either total duplicates, or are equivocating (i.e., they same
+  -- ID but a different target)
+  (Vote crypto committee -> Vote crypto committee -> Ordering) ->
+  -- | Collection of votes to check
+  NE [Vote crypto committee] ->
+  UniqueVotesWithSameTarget crypto committee
+unsafeUniqueVotesWithSameTarget getTarget cmpVotes votes =
+  assert
```
dnadales, line 229: With assertions stripped, this constructor produces a `UniqueVotesWithSameTarget` *without* checking the invariant — exactly the scenario the type is meant to rule out. In a Peras context, that means a buggy caller could silently land equivocating votes in a certificate in production, defeating the purpose of this PR.

A few options:

1. Always run the check, panic on failure (no flag dependency):
    ```haskell
    unsafeUniqueVotesWithSameTarget getTarget cmpVotes votes =
      case checkUniqueVotesWithSameTarget (Proxy @crypto) getTarget cmpVotes votes of
        Right () -> UniqueVotesWithSameTarget electionId candidate votes
        Left e   -> error $ "unsafeUniqueVotesWithSameTarget: " <> show e
    ```
2. Drop the unsafe constructor entirely; force callers through the `Either`-returning one.

If callers truly can guarantee the invariant by construction (per the PR description), option 1 is essentially free and removes the build-flag footgun. If they can't, option 1 is necessary.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -9,18 +11,23 @@ module Ouroboros.Consensus.Committee.Class
...
+  , UniqueVotesWithSameTarget
   , getElectionIdFromVotes
   , getVoteCandidateFromVotes
   , getRawVotes
-  , VotesWithSameTargetError (..)
-  , ensureSameTarget
+  , UniqueVotesWithSameTargetError (..)
+  , ensureUniqueVotesWithSameTarget
+  , unsafeUniqueVotesWithSameTarget
+  , checkUniqueVotesWithSameTarget -- for testing purposes only
```
dnadales, line 20: Are we going to add tests for these? In particular:

1. Well-formed input (same target, distinct under `cmpVotes`) → `Right`.
2. Duplicate detection (any two votes equal under `cmpVotes`) → `Left DuplicateVotes`, payload contains the cluster.
3. Target mismatch → `Left TargetMismatch`, payload contains all offending votes.
4. Both kinds of bad input simultaneously → `Left DuplicateVotes` (codifying the algorithm's priority — see separate comment).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+      -- First vote we are comparing against
+      vote
       -- Votes that do not match the target of the first vote
-      (NE [Vote crypto committee])
+      (NE [vote])
 
--- | Check that a list of votes all target the same election and candidate
-ensureSameTarget ::
+-- | Check that a non-empty list of votes all target the same election and
+-- candidate and there are have no duplicates
```
dnadales: Two small things on this docstring:

1. Typo: "there are have no duplicates" → "there are no duplicates".
2. The algorithm reports `DuplicateVotes` in preference to `TargetMismatch` — for an input like `[a, a, b]` where `b`'s target differs, the caller only sees the duplicate, not the mismatch. That's a defensible choice, but it's load-bearing enough to mention here so future readers don't have to read the `go` loop to figure out which error wins.

Suggested:

```haskell
-- | Check that a non-empty list of votes all target the same election and
-- candidate and that there are no duplicates. Duplicates are reported in
-- preference to target mismatches.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+getRawVotes (UniqueVotesWithSameTarget _ _ votes) =
   votes
 
 -- | Errors when votes do not all target the same election and candidate
-data VotesWithSameTargetError crypto committee
-  = EmptyVotes
+data UniqueVotesWithSameTargetError vote
+  = DuplicateVotes
+      -- First vote that occurs more than once
+      (NE [vote])
```
dnadales, line 162: The `cmpVotes` parameter docstring (a bit further down) explains that `EQ` covers two semantically distinct cases — "total duplicates" (same id, same target) and "equivocating" votes (same id, different target). Both end up reported through this constructor, but the comment here only describes the first case.

Worth tightening, since this constructor is the load-bearing piece of the equivocation defense:

```haskell
| DuplicateVotes
    -- A cluster of votes equal under the supplied ordering, i.e.,
    -- either true duplicates or equivocating votes (same id, different
    -- target).
    (NE [vote])
```

If you want to surface equivocation distinctly from harmless duplication in the future, this would also be the natural place to split the constructor — but no need to do that now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+    NonEmpty.sortBy cmpVotes votes
+
+  -- Check that all votes have the same target and there are no duplicates. The
+  -- first argument is the last vote we have checked, passed sequentially to the
+  -- next step to check against duplicates (the input must be sorted for this).
+  go _ [] =
+    Right ()
+  go v (v' : vs)
+    | isDuplicateOf v v' = do
+        let duplicateVotes = v :| (v' : filter (isDuplicateOf v) vs)
```
dnadales: Since `vs` is the tail of a list sorted by `cmpVotes`, all votes equal to `v` under that order are contiguous at its head. So `takeWhile (isDuplicateOf v) vs` produces the same result as `filter` and short-circuits at the first non-duplicate. Equivalent in correctness, faster on long inputs, and more accurately documents intent.

```diff
-        let duplicateVotes = v :| (v' : filter (isDuplicateOf v) vs)
+        let duplicateVotes = v :| (v' : takeWhile (isDuplicateOf v) vs)
```

(The companion `filter doesNotMatchFirstVoteTarget` two lines below cannot be simplified the same way — mismatches aren't contiguous.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
-data VotesWithSameTargetError crypto committee
-  = EmptyVotes
+data UniqueVotesWithSameTargetError vote
+  = DuplicateVotes
+      -- First vote that occurs more than once
+      (NE [vote])
   | TargetMismatch
-      -- First vote and the rest of the votes that match its target
-      (NE [Vote crypto committee])
+      -- First vote we are comparing against
```
dnadales: A caller reading this will assume "first vote" means the first vote they passed in, but `firstVote` here is taken from the input *after* `NonEmpty.sortBy cmpVotes`, so it's "first under the supplied ordering". For diagnostics it's still actionable (`getTarget firstVote` recovers the canonical target), but the wording can mislead.

```diff
  | TargetMismatch
-      -- First vote we are comparing against
+      -- First vote (under the supplied ordering) whose target is treated
+      -- as the canonical target for the comparison
       vote
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+ensureUniqueVotesWithSameTarget ::
+  forall crypto committee.
   ( Eq (ElectionId crypto)
   , Eq (VoteCandidate crypto)
   ) =>
+  -- | How to project the target from an abstract vote
   (Vote crypto committee -> (ElectionId crypto, VoteCandidate crypto)) ->
-  [Vote crypto committee] ->
+  -- | How to compare votes by ID, where EQ means that two votes have the same
+  -- ID and are either total duplicates, or are equivocating (i.e., they same
```
dnadales: Typo: "they same ID" → "they have the same ID". Same paragraph appears verbatim at lines 214 and 246, all three need the fix.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+    )
+    $ UniqueVotesWithSameTarget
+      electionId
+      candidate
+      (firstVote :| nextVotes)
+ where
+  firstVote :| nextVotes = votes
+  (electionId, candidate) = getTarget firstVote
+
+checkUniqueVotesWithSameTarget ::
```
dnadales, line 255: This function has the most subtle contract in the module — caller-supplied total ordering, sort-and-scan algorithm, mutually-exclusive errors with priority — and it's the only exported function with no Haddock. Worth a short doc-comment for future readers, e.g.:

```haskell
-- | Validate that a non-empty collection of votes is well-formed for
-- certificate forging: all votes target the same election and candidate
-- (per @getTarget@), and no two votes are equal under @cmpVotes@.
--
-- @cmpVotes@ must be a total order. Equality (@EQ@) is treated as evidence
-- of a duplicate or equivocating vote and is reported via 'DuplicateVotes'
-- in preference to any 'TargetMismatch'.
--
-- Exposed for testing; production code should use 'ensureUniqueVotesWithSameTarget'
-- or 'unsafeUniqueVotesWithSameTarget'.
checkUniqueVotesWithSameTarget :: …
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -1,6 +1,8 @@
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
```
dnadales: The new `forall crypto committee.` quantifiers in `ensureUniqueVotesWithSameTarget` / `unsafeUniqueVotesWithSameTarget` are rank-1 — `ScopedTypeVariables` alone enables them. I don't see a rank-2-or-higher type anywhere in the module; this pragma can be dropped to avoid signalling "this module uses higher-rank types" to readers.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
```diff
@@ -117,74 +124,158 @@ class
...
+        )
+    )
+    $ checkUniqueVotesWithSameTarget
+      (Proxy @crypto)
+      getTarget
+      cmpVotes
+      votes
+ where
+  firstVote :| nextVotes = votes
+  (electionId, candidate) = getTarget firstVote
```
dnadales, line 206: Optional cleanup: this `where` block plus the destructuring on the line above is repeated verbatim in `unsafeUniqueVotesWithSameTarget` (line 235–236). A small private helper would deduplicate:

```haskell
mkUniqueVotes
  :: (Vote crypto committee -> (ElectionId crypto, VoteCandidate crypto))
  -> NE [Vote crypto committee]
  -> UniqueVotesWithSameTarget crypto committee
mkUniqueVotes getTarget votes@(firstVote :| _) =
  let (electionId, candidate) = getTarget firstVote
   in UniqueVotesWithSameTarget electionId candidate votes
```

Saves ~6 lines total. Not worth a round-trip on its own, but if you're touching this for the other comments anyway.

dnadales (commented):
Looks good! I like the added checks 👌 Thanks!

I left some comments, in particular I'm concerned about the `unsafeUniqueVotesWithSameTarget` when asserts are disabled (which is the case in production).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
agustinmista: > If callers truly can guarantee the invariant by construction (per the PR description)

This is exactly why we are introducing two versions of the same constructor: the `PerasVoteDB` (from which we will collect the votes used to forge certificates from) should already satisfy all the invariants being checked here, so we ultimately want to avoid running these checks altogether in production (while still having them as assertions to catch bugs during testing).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
agustinmista: Yes, we already had tests for the old implementation in [Peras 25](https://github.com/IntersectMBO/ouroboros-consensus/pull/1977), which I already adapted to work with the new one :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
agustinmista: I think I would rather avoid doing this. It would be too easy to end up exposing this _trully unsafe_ constructor by mistake when tweaking export lists (which is something I observed LLMs being particularly eager to).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/Class.hs
dnadales: > we ultimately want to avoid running these checks altogether in production

Makes sense if `PerasVoteDB` enforces the invariant on its side — fully agreed on the two-versions setup.
I wonder if we should still spell out the contract here, though, since `assert` becomes a no-op under `-O` and a future caller skimming the type might not notice.
Something like a `WARNING` line on the docstring saying callers MUST guarantee the invariant because no check runs in production builds.

We can leave the `PerasVoteDB`-side property test out of this PR — added it to our production-readiness tracker.

agustinmista: Sounds good! I added [this warning](https://github.com/IntersectMBO/ouroboros-consensus/compare/21655232a0aadb4cdbf57b348794794bbc94a0e8..663a8cb54c27e01fd3e89cdcf571918102a2e5d1) to the haddock :crossed_fingers:

dnadales (approved):

PR #2023, opened 2026-05-01
Allow LSQ clients to connect during initial LedgerDB replay (EarlyN2C)

Cmdv:
# Allow LSQ clients to connect during initial LedgerDB replay (opt-in)

Today, cardano-node binds its local n2c socket only after the LedgerDB finishes its initial replay (typically 10–30 minutes from genesis). Any local client — notably `cardano-db-sync` — blocks on a missing socket until then. This PR adds an opt-in `EarlyN2C` flag on `ChainDbSpecificArgs` that:

- binds the local socket as soon as the ImmutableDB is ready,
- runs the LedgerDB replay on a background thread, and
- serves a *synthesised* read-only forker for LSQ Acquire requests while replay is in progress, so history-only queries (e.g. `GetInterpreter`) succeed.

The default (`EarlyN2CDisabled`) preserves the existing single-phase open and late-bind behaviour byte-for-byte. The opt-in (`EarlyN2CEnabled`) is enabled via the node config; cardano-node supplies the synthesis function (typically a closure over `pInfoInitLedger`).

```
EarlyN2CDisabled (default — unchanged)
  open ImmutableDB ─► replay LedgerDB ─► socket bind ─► clients
                     └─── synchronous ──┘                arrive

EarlyN2CEnabled (opt-in)
  open ImmutableDB ─► socket bind ─► clients arrive
                                      │
                   replay LedgerDB    ▼
                   (background)      Acquire ─► synthetic forker
                       │                        (history-only)
                       │
                       └─► ready ────────────► real forker
```

In `config.json` the operator opts in with:

```json
"EarlyN2C": "Enabled"
```

## Synthetic forker scope

Only `roforkerGetLedgerState` is implemented. The synthesised state suffices for `GetInterpreter` via `hardForkSummary`. Queries that touch the ledger tables (`GetUTxO*` etc.) hit the `error`-stubbed methods — we fail loudly rather than return empty data to a wallet.

## Wire format

Unchanged. `LedgerNotReady` is mapped to the existing `AcquireFailurePointTooOld` constructor; clients that already retry  on that failure (incl. older `cardano-db-sync`) work without changes.

## Tests

Two new properties in
`Test.Consensus.MiniProtocol.LocalStateQuery.Server`:

- `acquireFailsDuringLedgerReplay` checks that `LedgerNotReady` is reported as `AcquireFailurePointTooOld` regardless of `Target`.
- `acquireSucceedsAfterLedgerReady` checks the cross-state transition in a single LSQ session.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
@@ -32,7 +32,8 @@ import Ouroboros.Consensus.Ledger.Extended
 import Ouroboros.Consensus.Ledger.SupportsProtocol
 import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.Storage.ChainDB.API
-  ( GetLoEFragment
+  ( EarlyN2C (..)
+  , GetLoEFragment
   , LoE (LoEDisabled)
   )
 import Ouroboros.Consensus.Storage.ChainDB.Impl.Types
@@ -98,6 +99,16 @@ data ChainDbSpecificArgs f m blk = ChainDbSpecificArgs
   -- current LoE fragment.
   , cdbsSnapshotDelayRNG :: HKD f StdGen
   -- ^ RNG used to randomly determine the duration of snapshot delays
+  , cdbsEarlyN2C :: EarlyN2C
+  -- ^ See 'EarlyN2C'. Defaults to
+  -- 'EarlyN2CDisabled'.
+  , cdbsSynthesiseLedger ::
+      Maybe (Point blk -> Maybe (ExtLedgerState blk EmptyMK))
+  -- ^ Used when 'EarlyN2C' is enabled to build a read-only
+  -- ledger view at the immutable tip. Evaluated once at 'openDB' time
+  -- and cached on the 'ChainDbEnv'. If 'Nothing' (default) or if the
+  -- function returns 'Nothing' for the tip, LSQ Acquire during replay
+  -- fails with 'AcquireFailurePointTooOld'.
   }
 
 -- | Default arguments
@@ -133,6 +144,8 @@ defaultSpecificArgs =
     , cdbsTopLevelConfig = noDefault
     , cdbsLoE = pure LoEDisabled
     , cdbsSnapshotDelayRNG = noDefault
+    , cdbsEarlyN2C = EarlyN2CDisabled
+    , cdbsSynthesiseLedger = Nothing
     }
 
 -- | Default arguments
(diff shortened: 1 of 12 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
```diff
@@ -96,8 +94,16 @@ data ChainDbSpecificArgs f m blk = ChainDbSpecificArgs
...
+  , cdbsEarlyN2C :: EarlyN2C
+  -- ^ See 'EarlyN2C'. Defaults to
+  -- 'EarlyN2CDisabled'.
+  , cdbsSynthesiseLedger ::
+      Maybe (Point blk -> Maybe (ExtLedgerState blk EmptyMK))
+  -- ^ Used when 'EarlyN2C' is enabled to build a read-only
+  -- ledger view at the immutable tip. Evaluated once at 'openDB' time
+  -- and cached on the 'ChainDbEnv'. If 'Nothing' (default) or if the
+  -- function returns 'Nothing' for the tip, LSQ Acquire during replay
+  -- fails with 'AcquireFailurePointTooOld'.
```
jasagredo, line 111: It is still unclear to me what would I put here. How do I construct an ExtLedgerState from a Point?

jasagredo (commented):
There architectural change seems reasonable but I see that the TMVars hit many places and now we will have to run some STM to check every time. Furthermore, we now have some implicit dependencies that lead to PointTooOld in LSQ, when we actually would want to say something else.

Would it be possible to do something like this:
```haskell
data ChainDB m blk = ... -- same type as before

data PartiallyInitialisedChainDB m blk = ... -- the fields that would be tmvars are not even present.
```

Then the things that get started before completing the initialization get an `Either (PartiallyInitialisedChainDB m blK) (ChainDB m blk)` and we start them with the former and then stop them and start again with the latter. And the things that expect a fully intialised ChainDB will get a `ChainDB m blk` directly, and we can avoid those awaits and errors.

-------------------------------------------------------------------

As I was writing the above, the sentence:

> we start them with the former and then stop them and start again with the latter

made me think. How would we tell the client that the new instance is up? Perhaps it would need to change the protocol with a new message `LedgerReady` that tells the client to close the connection and start it again to connect to the new LSQ.

But maybe all this is just a needless complication. 

------------------------------------------------------------------

If we split the threads/services in Pre-ledger and Post-ledger, the Pre-ledger ones could expect a `TMVar (Either (PartiallyInitialisedChainDB m blK) (ChainDB m blk))` such that those are the ones that have to do the STM extra step, and the Post-ledger services expect a real ChainDB with no TMVars. In general the Post-ledger services are time-critical (ChainSel, forging), and the Pre-ledger services are not time-critical.

Cmdv:
Hi sorry I've been doing a lot of work locally for a new version of dbsync, this feature does work for me locally but wanted to double check your comments above were still valid then I can look a making changes and clearer the intent for you? :)

PR #2027, opened 2026-05-05
[Peras 27] Split degenerate BlockSupportsPeras instance

### LINKED ISSUE #213: Interface design & HFC plumbing for Peras

agustinmista:
This PR has the main goal of removing the existing degenerate `forall blk. BlockSupportsPeras blk` instance and replace it with separate, independent ones for the several block types in the codebase that need it.

To aid with this process, we have made a couple of preliminary changes:

* Moving base Peras types into a separate `O.C.Peras.Types` module to keep `O.C.Block.SupportsPeras` as focused on the `blk`-dependent API as possible.
* Consolidating all the `HasPeras(Vote|Cert)X` classes into two, `IsPerasVote` and `IsPerasCert`, grouping all the necessary non-crypto-related projections needed to operate on Peras votes and certificates.
* Removing the `PerasCfg blk` associated type, in favor of keeping things simple and using the monomorphic `PerasParams` for now. These will likely need a major overhaul in the future when on-chain governance of these parameters becomes necessary.
* Introducing dedicated modules for `MockPeras(Vote|Cert)`s, containing the non-crypto mocked Peras votes and certificates that were previously hardcoded everywhere. These will still be used for certain (mostly test) block types.
* Provide an "empty" default `BlockSupportsPeras` instance defined in terms of `Void`, so Peras votes, certs, and errors are uninstantiable and can be discharged via `absurd`.

With this in place, the remaining changes mostly encompass the plumbing needed after removing the existing `BlockSupportsPeras` instance and replacing it with dedicated ones. Concretely:

* `ByronBlock` uses the "emtpy" instance
* `ShelleyBlock proto era` uses a mocked instance (to be replaced with a concrete one using real type for `era ~ Dijkstra` and empty ones for every other `era.` (@tbagrel1 is working on this on a separate PR)
* `HardForkBlock xs` uses a mocked instance, but will be replaced with a compositional one defined around the two previous ones (also WIP) 

On the testing side:

* `BlockA` and `BlockB` both use the "empty" instance
* `DualByronBlock` uses the "empty" instance
* `O.C.Mock.Ledger.Block.SimpleBlock c ext` uses the "empty" instance
* `O.C.Storage.TestBlock.TestBlock` uses a mocked instance (now with a properly instantiated `getPerasCertInBlock`)
* `Test.Util.TestBlock.TestBlockWith ptype` uses a mocked instance (without any support for certificates in blocks, as it's not needed for any of the existing tests)

Finally, some general notes (notice that some of these are already covered by new issues linked in the comments below):

* There is a substantial number of instances involving type families that had to be derived using standalone deriving statements now to keep GHC happy.
* `BlockSupportsPeras` is still missing a `forgePerasVote` method. This will become evident soon when trying to write a `blk`-generic voting thread (https://github.com/tweag/cardano-peras/issues/248)
* Some of the new `BlockSupportsPeras` instances are defined via orphans in dedicated modules. This is because:
  +  `SerialiseNodeToNodeConstraints blk` now requires `SerialiseNodeToNode (PerasVote blk)` and `SerialiseNodeToNode (PerasCert blk)` which are only in scope after `PerasVote blk` and `PerasCert blk` are established when defining `BlockSupportsPeras blk`
  + Some `SerialiseNodeToNodeConstraints` instances are already defined as orphans in their own dedicated modules, forcing us to create a new module or to define `BlockSupportsPeras blk` in the same one as `SerialiseNodeToNodeConstraints blk` (confusing!). We might be able to rearrange constraints to avoid this, but that would probably mean adding `SerialiseNodeToNode (Peras(Vote|Cert) blk)` as superclasses to a type-class higher-up in the chain.
  + We favored tight constraints over "umbrella" ones as much as possible to avoid having to implement unnecessary/unrelated interfaces for test blocks as much as possible.
* Mocked votes and certs are trivially validatable without errors (technically, with uninstantiable errors if being pedantic). For this to work, `MockPerasVote` carries its own stake. Please note this is not the case for real votes, as their stake is retrieved from the stake distribution after validation.
* Some `Peras...` types will become redundant and get removed in a later PR when we start integrating components.

### DIFF
--- a/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
+++ b/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Empty Peras support for Byron.
+--
+-- NOTE: this module exists solely because the orphan module
+-- 'Ouroboros.Consensus.Byron.Node.Serialisation' needs this instance, but
+-- defining it there would be too confusing.
+module Ouroboros.Consensus.Byron.Node.Peras () where
+
+import Ouroboros.Consensus.Block.SupportsPeras (BlockSupportsPeras)
+import Ouroboros.Consensus.Byron.Ledger.Block (ByronBlock)
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: Byron does not support Peras, so we can use the empty instance here.
+instance BlockSupportsPeras ByronBlock
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
@@ -1,0 +1,55 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Mocked Peras support for Shelley.
+--
+-- NOTE: this module exists solely because the orphan module
+-- 'Ouroboros.Consensus.Shelley.Node.Serialisation' needs this instance, but
+-- defining it there would be too confusing.
+module Ouroboros.Consensus.Shelley.Node.Peras () where
+
+import Ouroboros.Consensus.Block.SupportsPeras
+  ( BlockSupportsPeras (..)
+  , VoidPerasError
+  )
+import Ouroboros.Consensus.Peras.Cert.Mock
+  ( MockPerasCert (..)
+  , forgeMockPerasCert
+  , validateMockPerasCert
+  )
+import Ouroboros.Consensus.Peras.Vote.Mock
+  ( MockPerasVote (..)
+  , validateMockPerasVote
+  )
+import Ouroboros.Consensus.Shelley.Ledger.Block
+  ( ShelleyBlock
+  , ShelleyCompatible
+  )
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
+
+-- TODO: replace this with a concrete implementation using 'Peras.Vote.V1' and
+-- 'Peras.Cert.V1', either forall eras or for @forall era. era >= DijkstraEra@.
+
+instance
+  ShelleyCompatible proto era =>
+  BlockSupportsPeras (ShelleyBlock proto era)
+  where
+  type PerasVote (ShelleyBlock proto era) = MockPerasVote (ShelleyBlock proto era)
+  type PerasCert (ShelleyBlock proto era) = MockPerasCert (ShelleyBlock proto era)
+  type PerasError (ShelleyBlock proto era) = VoidPerasError (ShelleyBlock proto era)
+
+  validatePerasVote = validateMockPerasVote
+  validatePerasCert = validateMockPerasCert
+  forgePerasCert = forgeMockPerasCert
+
+  -- TODO: extract actual Peras certificates from blocks
+  getPerasCertInBlock _ = Nothing
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -12,6 +13,7 @@
 -- defining it there would be too confusing.
 module Ouroboros.Consensus.Shelley.Node.Peras () where
 
+import Cardano.Ledger.Api
 import Ouroboros.Consensus.Block.SupportsPeras
   ( BlockSupportsPeras (..)
   , VoidPerasError
@@ -34,18 +36,26 @@ import Ouroboros.Consensus.Shelley.Ledger.Block
   BlockSupportsPeras
 -------------------------------------------------------------------------------}
 
+-- Peras support starts with DijkstraEra, so earlier eras use the default void
+-- implementation.
+
+instance ShelleyCompatible proto ShelleyEra => BlockSupportsPeras (ShelleyBlock proto ShelleyEra)
+instance ShelleyCompatible proto AllegraEra => BlockSupportsPeras (ShelleyBlock proto AllegraEra)
+instance ShelleyCompatible proto MaryEra => BlockSupportsPeras (ShelleyBlock proto MaryEra)
+instance ShelleyCompatible proto AlonzoEra => BlockSupportsPeras (ShelleyBlock proto AlonzoEra)
+instance ShelleyCompatible proto BabbageEra => BlockSupportsPeras (ShelleyBlock proto BabbageEra)
+instance ShelleyCompatible proto ConwayEra => BlockSupportsPeras (ShelleyBlock proto ConwayEra)
+
 -- NOTE: this is a mocked up implementation without crypto!
-
 -- TODO: replace this with a concrete implementation using 'Peras.Vote.V1' and
--- 'Peras.Cert.V1', either forall eras or for @forall era. era >= DijkstraEra@.
-
+-- 'Peras.Cert.V1' for era >= DijkstraEra.
 instance
-  ShelleyCompatible proto era =>
-  BlockSupportsPeras (ShelleyBlock proto era)
+  ShelleyCompatible proto DijkstraEra =>
+  BlockSupportsPeras (ShelleyBlock proto DijkstraEra)
   where
-  type PerasVote (ShelleyBlock proto era) = MockPerasVote (ShelleyBlock proto era)
-  type PerasCert (ShelleyBlock proto era) = MockPerasCert (ShelleyBlock proto era)
-  type PerasError (ShelleyBlock proto era) = VoidPerasError (ShelleyBlock proto era)
+  type PerasVote (ShelleyBlock proto DijkstraEra) = MockPerasVote (ShelleyBlock proto DijkstraEra)
+  type PerasCert (ShelleyBlock proto DijkstraEra) = MockPerasCert (ShelleyBlock proto DijkstraEra)
+  type PerasError (ShelleyBlock proto DijkstraEra) = VoidPerasError (ShelleyBlock proto DijkstraEra)
 
   validatePerasVote = validateMockPerasVote
   validatePerasCert = validateMockPerasCert
--- a/ouroboros-consensus-cardano/src/unstable-byron-testlib/Ouroboros/Consensus/ByronDual/Node/Peras.hs
+++ b/ouroboros-consensus-cardano/src/unstable-byron-testlib/Ouroboros/Consensus/ByronDual/Node/Peras.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | Empty Peras support for DualByron.
+--
+-- NOTE: this module exists solely because the orphan module
+-- 'Ouroboros.Consensus.ByronDual.Node.Serialisation' needs this instance, but
+-- defining it there would be too confusing.
+module Ouroboros.Consensus.ByronDual.Node.Peras () where
+
+import Ouroboros.Consensus.Block.SupportsPeras (BlockSupportsPeras)
+import Ouroboros.Consensus.ByronDual.Ledger (DualByronBlock)
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: DualByron does not support Peras, so we can use the empty instance here.
+instance BlockSupportsPeras DualByronBlock
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -679,6 +679,7 @@ test-suite consensus-test
     Test.Consensus.Peras.Util
     Test.Consensus.Peras.Voting.Rules
     Test.Consensus.Peras.WeightSnapshot
+    Test.Consensus.Util.Bitmap
     Test.Consensus.Util.MonadSTM.NormalForm
     Test.Consensus.Util.Pred
     Test.Consensus.Util.Versioned
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -676,6 +676,7 @@ test-suite consensus-test
     Test.Consensus.MiniProtocol.ObjectDiffusion.PerasVote.Smoke
     Test.Consensus.MiniProtocol.ObjectDiffusion.Smoke
     Test.Consensus.Peras.Cert.Inclusion
+    Test.Consensus.Peras.Serialisation
     Test.Consensus.Peras.Util
     Test.Consensus.Peras.Voting.Rules
     Test.Consensus.Peras.WeightSnapshot
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -320,6 +320,7 @@ library
     Ouroboros.Consensus.Util.AnchoredSeq
     Ouroboros.Consensus.Util.Args
     Ouroboros.Consensus.Util.Assert
+    Ouroboros.Consensus.Util.Bitmap
     Ouroboros.Consensus.Util.CallStack
     Ouroboros.Consensus.Util.CBOR
     Ouroboros.Consensus.Util.Condense
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -236,12 +236,12 @@ library
     Ouroboros.Consensus.Node.Run
     Ouroboros.Consensus.Node.Serialisation
     Ouroboros.Consensus.NodeId
-    Ouroboros.Consensus.Peras.Cert
     Ouroboros.Consensus.Peras.Cert.Inclusion
+    Ouroboros.Consensus.Peras.Cert.V1
     Ouroboros.Consensus.Peras.Params
     Ouroboros.Consensus.Peras.SelectView
-    Ouroboros.Consensus.Peras.Vote
     Ouroboros.Consensus.Peras.Vote.Aggregation
+    Ouroboros.Consensus.Peras.Vote.V1
     Ouroboros.Consensus.Peras.Voting.Rules
     Ouroboros.Consensus.Peras.Voting.View
     Ouroboros.Consensus.Peras.Weight
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -678,6 +678,7 @@ test-suite consensus-test
     Test.Consensus.Peras.Cert.Inclusion
     Test.Consensus.Peras.Serialisation
     Test.Consensus.Peras.Util
+    Test.Consensus.Peras.Voting.Committee
     Test.Consensus.Peras.Voting.Rules
     Test.Consensus.Peras.WeightSnapshot
     Test.Consensus.Util.Bitmap
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -243,6 +243,7 @@ library
     Ouroboros.Consensus.Peras.SelectView
     Ouroboros.Consensus.Peras.Vote.Aggregation
     Ouroboros.Consensus.Peras.Vote.V1
+    Ouroboros.Consensus.Peras.Voting.Committee
     Ouroboros.Consensus.Peras.Voting.Rules
     Ouroboros.Consensus.Peras.Voting.View
     Ouroboros.Consensus.Peras.Weight
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -238,6 +238,7 @@ library
     Ouroboros.Consensus.NodeId
     Ouroboros.Consensus.Peras.Cert.Inclusion
     Ouroboros.Consensus.Peras.Cert.V1
+    Ouroboros.Consensus.Peras.Crypto.BLS
     Ouroboros.Consensus.Peras.Params
     Ouroboros.Consensus.Peras.SelectView
     Ouroboros.Consensus.Peras.Vote.Aggregation
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -241,6 +241,7 @@ library
     Ouroboros.Consensus.Peras.Crypto.BLS
     Ouroboros.Consensus.Peras.Params
     Ouroboros.Consensus.Peras.SelectView
+    Ouroboros.Consensus.Peras.Types
     Ouroboros.Consensus.Peras.Vote.Aggregation
     Ouroboros.Consensus.Peras.Vote.V1
     Ouroboros.Consensus.Peras.Voting.Committee
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -676,6 +676,7 @@ test-suite consensus-test
     Test.Consensus.MiniProtocol.ObjectDiffusion.PerasVote.Smoke
     Test.Consensus.MiniProtocol.ObjectDiffusion.Smoke
     Test.Consensus.Peras.Cert.Inclusion
+    Test.Consensus.Peras.Util
     Test.Consensus.Peras.Voting.Rules
     Test.Consensus.Peras.WeightSnapshot
     Test.Consensus.Util.MonadSTM.NormalForm
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -246,7 +246,7 @@ library
     Ouroboros.Consensus.Peras.Vote.Aggregation
     Ouroboros.Consensus.Peras.Vote.Mock
     Ouroboros.Consensus.Peras.Vote.V1
-    Ouroboros.Consensus.Peras.Voting.Committee
+    Ouroboros.Consensus.Peras.Voting.Adapter
     Ouroboros.Consensus.Peras.Voting.Rules
     Ouroboros.Consensus.Peras.Voting.View
     Ouroboros.Consensus.Peras.Weight
@@ -682,7 +682,7 @@ test-suite consensus-test
     Test.Consensus.Peras.Cert.Inclusion
     Test.Consensus.Peras.Serialisation
     Test.Consensus.Peras.Util
-    Test.Consensus.Peras.Voting.Committee
+    Test.Consensus.Peras.Voting.Adapter
     Test.Consensus.Peras.Voting.Rules
     Test.Consensus.Peras.WeightSnapshot
     Test.Consensus.Util.Bitmap
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -237,12 +237,14 @@ library
     Ouroboros.Consensus.Node.Serialisation
     Ouroboros.Consensus.NodeId
     Ouroboros.Consensus.Peras.Cert.Inclusion
+    Ouroboros.Consensus.Peras.Cert.Mock
     Ouroboros.Consensus.Peras.Cert.V1
     Ouroboros.Consensus.Peras.Crypto.BLS
     Ouroboros.Consensus.Peras.Params
     Ouroboros.Consensus.Peras.SelectView
     Ouroboros.Consensus.Peras.Types
     Ouroboros.Consensus.Peras.Vote.Aggregation
+    Ouroboros.Consensus.Peras.Vote.Mock
     Ouroboros.Consensus.Peras.Vote.V1
     Ouroboros.Consensus.Peras.Voting.Committee
     Ouroboros.Consensus.Peras.Voting.Rules
@@ -580,6 +582,7 @@ library unstable-mock-block
     Ouroboros.Consensus.Mock.Node.Abstract
     Ouroboros.Consensus.Mock.Node.BFT
     Ouroboros.Consensus.Mock.Node.PBFT
+    Ouroboros.Consensus.Mock.Node.Peras
     Ouroboros.Consensus.Mock.Node.Praos
     Ouroboros.Consensus.Mock.Node.PraosRule
     Ouroboros.Consensus.Mock.Node.Serialisation
@@ -1315,6 +1318,7 @@ library cardano
     Ouroboros.Consensus.Byron.Ledger.PBFT
     Ouroboros.Consensus.Byron.Ledger.Serialisation
     Ouroboros.Consensus.Byron.Node
+    Ouroboros.Consensus.Byron.Node.Peras
     Ouroboros.Consensus.Byron.Node.Serialisation
     Ouroboros.Consensus.Byron.Protocol
     Ouroboros.Consensus.Cardano
@@ -1346,6 +1350,7 @@ library cardano
     Ouroboros.Consensus.Shelley.Node
     Ouroboros.Consensus.Shelley.Node.Common
     Ouroboros.Consensus.Shelley.Node.DiffusionPipelining
+    Ouroboros.Consensus.Shelley.Node.Peras
     Ouroboros.Consensus.Shelley.Node.Praos
     Ouroboros.Consensus.Shelley.Node.Serialisation
     Ouroboros.Consensus.Shelley.Node.TPraos
@@ -1443,6 +1448,7 @@ library unstable-byron-testlib
   exposed-modules:
     Ouroboros.Consensus.ByronDual.Ledger
     Ouroboros.Consensus.ByronDual.Node
+    Ouroboros.Consensus.ByronDual.Node.Peras
     Ouroboros.Consensus.ByronDual.Node.Serialisation
     Test.Consensus.Byron.Examples
     Test.Consensus.Byron.Generators
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -19,7 +19,7 @@ module Ouroboros.Consensus.Block.SupportsPeras
   , ValidatedPerasVotesWithQuorum
     ( vpvqTarget
     , vpvqVotes
-    , vpvqPerasCfg
+    , vpvqPerasParams
     )
   , votesReachQuorum
   , HasPerasCertRound (..)
@@ -80,8 +80,8 @@ data ValidatedPerasVotesWithQuorum blk = ValidatedPerasVotesWithQuorum
   -- ^ The target that all the votes are for
   , vpvqVotes :: !(NonEmpty (ValidatedPerasVote blk))
   -- ^ The votes that reached quorum for the given target
-  , vpvqPerasCfg :: !(PerasCfg blk)
-  -- ^ The Peras configuration used to validate that the votes reach quorum
+  , vpvqPerasParams :: !PerasParams
+  -- ^ The Peras parameters used to validate that the votes reach quorum
   }
   deriving stock (Show, Eq, Generic)
   deriving anyclass NoThunks
@@ -93,10 +93,10 @@ data ValidatedPerasVotesWithQuorum blk = ValidatedPerasVotesWithQuorum
 -- It returns 'Nothing' if either of these conditions is not met.
 votesReachQuorum ::
   StandardHash blk =>
-  PerasCfg blk ->
+  PerasParams ->
   [ValidatedPerasVote blk] ->
   Maybe (ValidatedPerasVotesWithQuorum blk)
-votesReachQuorum cfg votes =
+votesReachQuorum params votes =
   case votes of
     -- We need at least one vote to determine who these votes are for, so we
     -- can't vacuously reach a quorum, even if the quorum threshold is 0.
@@ -113,26 +113,19 @@ votesReachQuorum cfg votes =
             ValidatedPerasVotesWithQuorum
               { vpvqTarget = getPerasVoteTarget v0
               , vpvqVotes = v0 :| vs
-              , vpvqPerasCfg = cfg
+              , vpvqPerasParams = params
               }
  where
   totalVoteStake =
     mconcat (vpvVoteStake <$> votes)
   votesHaveEnoughStake =
-    stakeAboveThreshold cfg totalVoteStake
+    stakeAboveThreshold params totalVoteStake
   allVotesMatchTarget target =
     all ((== (getPerasVoteTarget target)) . getPerasVoteTarget)
 
 -- * BlockSupportsPeras class
 
-class
-  ( Show (PerasCfg blk)
-  , NoThunks (PerasCert blk)
-  ) =>
-  BlockSupportsPeras blk
-  where
-  type PerasCfg blk
-
+class BlockSupportsPeras blk where
   data PerasCert blk
 
   data PerasVote blk
@@ -142,18 +135,18 @@ class
   data PerasForgeErr blk
 
   validatePerasCert ::
-    PerasCfg blk ->
+    PerasParams ->
     PerasCert blk ->
     Either (PerasValidationErr blk) (ValidatedPerasCert blk)
 
   validatePerasVote ::
-    PerasCfg blk ->
+    PerasParams ->
     PerasVoteStakeDistr ->
     PerasVote blk ->
     Either (PerasValidationErr blk) (ValidatedPerasVote blk)
 
   forgePerasCert ::
-    PerasCfg blk ->
+    PerasParams ->
     ValidatedPerasVotesWithQuorum blk ->
     Either (PerasForgeErr blk) (ValidatedPerasCert blk)
 
@@ -167,9 +160,7 @@ class
 
 -- TODO: degenerate instance for all blks to get things to compile
 -- see https://github.com/tweag/cardano-peras/issues/73
-instance StandardHash blk => BlockSupportsPeras blk where
-  type PerasCfg blk = PerasParams
-
+instance BlockSupportsPeras blk where
   data PerasCert blk = PerasCert
     { pcCertRound :: PerasRoundNo
     , pcCertBoostedBlock :: Point blk
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
@@ -12,18 +11,7 @@
 {-# LANGUAGE UndecidableInstances #-}
 
 module Ouroboros.Consensus.Block.SupportsPeras
-  ( PerasRoundNo (..)
-  , onPerasRoundNo
-  , PerasBoostedBlock (..)
-  , PerasSeatIndex (..)
-  , PerasVoteId (..)
-  , PerasVoteTarget (..)
-  , PerasVoterId (..)
-  , PerasVoteStake (..)
-  , stakeAboveThreshold
-  , PerasVoteStakeDistr (..)
-  , lookupPerasVoteStake
-  , BlockSupportsPeras (..)
+  ( BlockSupportsPeras (..)
   , PerasCert (..)
   , PerasVote (..)
   , ValidatedPerasCert (..)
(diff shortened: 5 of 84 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
+  , Typeable (PerasVote blk)
+  , Typeable (PerasCert blk)
+  , IsPerasVote (PerasVote blk) blk
+  , IsPerasCert (PerasCert blk) blk
+  , Show (PerasVote blk)
+  , Show (PerasCert blk)
+  , Eq (PerasVote blk)
+  , Eq (PerasCert blk)
+  , NoThunks (PerasVote blk)
+  , NoThunks (PerasCert blk)
```
tbagrel1, line 67: I suggested we could move the vote-related and cert-related constraints to the `IsPerasVote` and `IsPerasCert` classes. If there is a reason why this isn't desirable, we should probably document it :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -262,319 +193,68 @@ votesReachQuorum cfg votes =
...
--- * BlockSupportsPeras class
--------------------------------------------------------------------------------}
-
-class
-  ( Show (PerasCfg blk)
-  , NoThunks (PerasCert blk)
-  ) =>
-  BlockSupportsPeras blk
-  where
-  type PerasCfg blk
```
tbagrel1, line 284: Mention #wontfix-for-testnet in the PR description

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
-  deriving newtype (Eq, Ord, Num, Fractional, NoThunks, Serialise)
-  deriving stock Generic
-  deriving Show via Quiet PerasVoteStake
-  deriving Semigroup via Sum Rational
-  deriving Monoid via Sum Rational
+  -- | Extract a Peras certificate optionally stored in a block.
+  --
+  -- Returns 'Nothing' if the block does not contain a Peras certificate, or
+  -- if the block is from an era that does not support Peras certificates.
+  getPerasCertInBlock ::
```
tbagrel1, line 125: Mention #wontfix-for-testnet

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
-      (perasQuorumStakeThresholdSafetyMargin params)
+-- * Error types
 
-newtype PerasVoteStakeDistr = PerasVoteStakeDistr
-  { unPerasVoteStakeDistr :: Map PerasVoterId PerasVoteStake
-  }
-  deriving newtype NoThunks
+-- TODO: enrich with actual error types
+-- see https://github.com/tweag/cardano-peras/issues/120
+data PerasValidationErr blk
```
tbagrel1: Is there a point in having a concrete data here?
I don't think we will be able to populate it with concrete variants and it will just be a carrier for a string payload.

I think it may be simpler to have a per-blk `PerasErr` type family (just one instead of distinguishing validation / forging), so we could reuse almost immediately the committee selection data structure and preserve its precision/shape at no cost.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
   deriving anyclass NoThunks
 
-data PerasVoteId blk = PerasVoteId
-  { pviRoundNo :: !PerasRoundNo
-  , pviVoterId :: !PerasVoterId
-  }
-  deriving stock (Show, Eq, Ord, Generic)
+-- TODO: enrich with actual error types
+-- see https://github.com/tweag/cardano-peras/issues/120
+data PerasForgeErr blk
```
tbagrel1: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
 
--- ** Validated types
+data ValidatedPerasVote blk
+  = ValidatedPerasVote
+  { vpvVote :: !(PerasVote blk)
+  , vpvVoteStake :: !PerasVoteStake
+  }
 
-data ValidatedPerasCert blk = ValidatedPerasCert
+deriving instance Show (PerasVote blk) => Show (ValidatedPerasVote blk)
```
tbagrel1, line 177: Not sure that the `deriving instance` are doing anything different than the `deriving` clause embedded in the data definition?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
+deriving instance Eq (PerasCert blk) => Eq (ValidatedPerasCert blk)
+deriving instance Ord (PerasCert blk) => Ord (ValidatedPerasCert blk)
+deriving instance NoThunks (PerasCert blk) => NoThunks (ValidatedPerasCert blk)
+deriving instance Generic (ValidatedPerasCert blk)
 
 -- | A collection of validated Peras votes that:
 -- 1. are all for the same target, and
 -- 2. have total stake above the quorum threshold for a given 'PerasCfg'.
-data ValidatedPerasVotesWithQuorum blk = ValidatedPerasVotesWithQuorum
+data ValidatedPerasVotesWithQuorum blk
```
tbagrel1, line 198: We should probably ensure uniqueness too (although I guess this will be automatically done when we merge committee stuff with this)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Fragment/Diff.hs
```diff
@@ -35,6 +35,7 @@ module Ouroboros.Consensus.Fragment.Diff
 import Data.Word (Word64)
 import GHC.Stack (HasCallStack)
 import Ouroboros.Consensus.Block
+import Ouroboros.Consensus.Peras.Params (PerasWeight)
```
tbagrel1: Not very happy about this. Can't we reexport `PerasWeight` from `Ouroboros.Consensus.Peras.Weight` (or move its definition to this file)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -57,9 +58,18 @@ import Ouroboros.Consensus.HardFork.Combinator.State.Types
...
+  , validateMockPerasCert
+  )
+import Ouroboros.Consensus.Peras.Vote.Mock
+  ( MockPerasVote (..)
+  , validateMockPerasVote
+  )
 import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.TypeFamilyWrappers
-import Ouroboros.Consensus.Util (ShowProxy)
+import Ouroboros.Consensus.Util (ShowProxy (..))
```
tbagrel1: Probably can remove this diff

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -5,247 +5,178 @@
...
+  , Show (PerasCert blk)
+  , Eq (PerasVote blk)
+  , Eq (PerasCert blk)
+  , NoThunks (PerasVote blk)
+  , NoThunks (PerasCert blk)
+  ) =>
+  BlockSupportsPeras blk
+  where
+  type PerasVote blk = (vote :: Type) | vote -> blk
+  type PerasCert blk = (cert :: Type) | cert -> blk
```
tbagrel1, line 79: We should tell somewhere why `forgePerasVote` is not part of this class with its friends :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -258,3 +268,25 @@ instance CanHardFork xs => LedgerSupportsPeras (HardForkBlock xs) where
...
       . State.tip
       . hardForkLedgerStatePerEra
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
+
+instance
```
tbagrel1, line 285: Is it the right place to define the HFC instance for `BlockSupportsPeras`? I think in our previous attempt, we had a dedicated file.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -258,3 +268,25 @@ instance CanHardFork xs => LedgerSupportsPeras (HardForkBlock xs) where
...
+  ( StandardHash (HardForkBlock xs)
+  , CanHardFork xs
+  ) =>
+  BlockSupportsPeras (HardForkBlock xs)
```
tbagrel1, line 289: We need a `TODO: use concrete vote V1 wrapped in a HFC wrapper (to allow future binary evolutions)`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/EraParams.hs
```diff
@@ -39,6 +39,7 @@ import NoThunks.Class (NoThunks)
 import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types
 import Ouroboros.Consensus.Config.SecurityParam
+import Ouroboros.Consensus.Peras.Params (PerasRoundLength (..))
```
tbagrel1: This looks a bit fishy; where did this import come from (before the change)? Maybe a reexport is missing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
```diff
@@ -68,6 +68,8 @@ import Ouroboros.Consensus.BlockchainTime.WallClock.Types
 import Ouroboros.Consensus.HardFork.History.EraParams
 import Ouroboros.Consensus.HardFork.History.Summary
 import Ouroboros.Consensus.HardFork.History.Util
+import Ouroboros.Consensus.Peras.Params (PerasRoundLength (..))
```
tbagrel1: This looks a bit fishy; where did this import come from (before the change)? Maybe a reexport is missing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Summary.hs
```diff
@@ -73,6 +73,8 @@ import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types
 import Ouroboros.Consensus.HardFork.History.EraParams
 import Ouroboros.Consensus.HardFork.History.Util
+import Ouroboros.Consensus.Peras.Params (PerasRoundLength (..))
```
tbagrel1: This looks a bit fishy; where did this import come from (before the change)? Maybe a reexport is missing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Util.hs
```diff
@@ -13,6 +13,7 @@ import Control.Exception (assert)
 import Data.Word
 import GHC.Stack
 import Ouroboros.Consensus.Block
+import Ouroboros.Consensus.Peras.Types (PerasRoundNo (..))
```
tbagrel1: This looks a bit fishy; where did this import come from (before the change)? Maybe a reexport is missing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
```diff
@@ -5,8 +5,8 @@ module Ouroboros.Consensus.Ledger.SupportsPeras
   )
 where
 
-import Ouroboros.Consensus.Block.SupportsPeras (PerasRoundNo)
 import Ouroboros.Consensus.Ledger.Abstract (LedgerState)
+import Ouroboros.Consensus.Peras.Types (PerasRoundNo)
```
tbagrel1: This looks a bit fishy; where did this import come from (before the change)? Maybe a reexport is missing?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -44,7 +47,9 @@ takeAscMap n = Map.fromDistinctAscList . take n . Map.toAscList
 
 -- | Internal helper: create a pool reader from a @getCertsAfter@ function.
 makePerasCertPoolReader ::
-  IOLike m =>
+  ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 49: I would prefer if we used uniformly `BlockSupportsPeras blk` constraint instead of `IsPerasCert ...` in these functions

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -65,15 +70,19 @@ makePerasCertPoolReader getCertsAfterSTM =
     }
 
 makePerasCertPoolReaderFromCertDB ::
-  IOLike m =>
+  ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 72: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -65,15 +70,19 @@ makePerasCertPoolReader getCertsAfterSTM =
...
   PerasCertDB m blk ->
   ObjectPoolReader PerasRoundNo (PerasCert blk) PerasCertTicketNo m
 makePerasCertPoolReaderFromCertDB perasCertDB =
   makePerasCertPoolReader
     (PerasCertDB.getCertsAfter perasCertDB)
 
 makePerasCertPoolReaderFromChainDB ::
-  IOLike m =>
+  ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 82: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -154,7 +167,9 @@ instance Exception PerasCertInboundException
 -- each valid certificate is timestamped with the current wall-clock time and
 -- added to the database via @addCert@.
 processCerts ::
-  MonadSTM m =>
+  ( MonadSTM m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 172: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -45,7 +48,9 @@ takeAscMap n = Map.fromDistinctAscList . take n . Map.toAscList
 
 -- | Internal helper: create a pool reader from a @getVotesAfter@ function.
 makePerasVotePoolReader ::
-  IOLike m =>
+  ( IOLike m
+  , IsPerasVote (PerasVote blk) blk
```
tbagrel1, line 50: Same remark about contraints consistency as in the sibling file ObjectPool/PerasCert.hs

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/PerasCert.hs
```diff
@@ -14,6 +14,7 @@ import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Outbound
+import Ouroboros.Consensus.Peras.Types (PerasRoundNo)
```
tbagrel1: Since it's the PerasCert ID, it should probably be reexported by the CertDB API?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/PerasVote.hs
```diff
@@ -14,6 +14,7 @@ import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Outbound
+import Ouroboros.Consensus.Peras.Types (PerasVoteId)
```
tbagrel1: Should probably be reexported by the VoteDB API?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -120,6 +123,8 @@ class
   , ShowProxy (Header blk)
   , ShowProxy (BlockQuery blk)
   , ShowProxy (TxId (GenTx blk))
+  , ShowProxy (PerasVote blk)
+  , ShowProxy (PerasCert blk)
```
tbagrel1, line 126: What about adding these constraints in `BlockSupportsPeras` or `IsPerasCert`/`IsPerasVote` ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Mock.hs
```diff
@@ -0,0 +1,153 @@
...
+    pure
+      MockPerasCert
+        { mockCertRound
+        , mockCertBlock
+        }
+
+-- | Helper to write 'BlockSupportsPeras.validatePerasCert'.
+--
+-- TODO: perform actual validation against all possible 'PerasValidationErr's.
+-- See https://github.com/tweag/cardano-peras/issues/120
```
tbagrel1: Not sure we need validation for the mock cert?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Mock.hs
```diff
@@ -0,0 +1,153 @@
...
+      { vpcCert = coerce mockCert
+      , vpcCertBoost = perasWeight params
+      }
+ where
+  mockCert :: MockPerasCert blk
+  mockCert = coerce cert
+
+-- | Helper to write 'BlockSupportsPeras.forgePerasCert'.
+--
+-- TODO: perform actual validation against all possible 'PerasForgeErr's.
```
tbagrel1: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
```diff
@@ -109,27 +121,85 @@ data PerasRoundVoteState blk = PerasRoundVoteState
   { prvsRoundNo :: !PerasRoundNo
   , prvsState :: !(Either (NoQuorum blk) (Quorum blk))
   }
-  deriving stock (Generic, Eq, Show)
-  deriving anyclass NoThunks
 
-instance HasPerasVoteRound (PerasRoundVoteState blk) where
-  getPerasVoteRound = prvsRoundNo
+deriving instance
```
tbagrel1, line 116: Genuinely curious, are these separate deriving clauses doing something different that the one attached to the data declaration?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Mock.hs
```diff
@@ -0,0 +1,148 @@
...
+      MockPerasVote
+        { mockVoteRound
+        , mockVoteBlock
+        , mockVoteVoterId
+        }
+
+-- | Helper to write 'BlockSupportsPeras.validatePerasVote'.
+--
+-- TODO: perform actual validation against all possible 'PerasValidationErr's.
+-- See https://github.com/tweag/cardano-peras/issues/120
```
tbagrel1: Same here, not sure we need real verification?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
```diff
@@ -1,7 +1,6 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
```
tbagrel1, line 6: What about renaming this file CommitteeAdapter or something, to indicate more clearly that it is about interfacing Committee with Peras-related concerns?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -37,18 +37,14 @@ import Ouroboros.Consensus.Block (WithOrigin (..))
...
-  , onPerasRoundNo
-  )
+import Ouroboros.Consensus.Block.SupportsPeras (IsPerasCert (..))
 import Ouroboros.Consensus.Peras.Params
   ( PerasCertArrivalThreshold (..)
   , PerasCooldownRounds (..)
   , PerasIgnoranceRounds (..)
   , PerasParams (..)
   )
+import Ouroboros.Consensus.Peras.Types (PerasRoundNo (..), onPerasRoundNo)
```
tbagrel1: Again, I'm wondering if they should be reexported by `SupportsPeras` since they are used basically everywhere in that module

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
```diff
@@ -296,5 +296,5 @@ mkPerasVotingView
       -- Check whether the boosted block is within the volatile fragment leading
       -- to the candidate block.
       AF.withinFragmentBounds
-        (castPoint (getPerasCertBoostedBlock cert))
+        (castPoint (getPerasCertBlock cert))
```
tbagrel1, line 295: FTR, I think the previous name of this method was fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/SelectView.hs
```diff
@@ -23,6 +23,7 @@ module Ouroboros.Consensus.Peras.SelectView
 
 import Data.Function (on)
 import Ouroboros.Consensus.Block
+import Ouroboros.Consensus.Peras.Params (PerasWeight (..))
```
tbagrel1: ditto reexport

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
```diff
@@ -0,0 +1,211 @@
...
+
+-- | Lift a binary operation on 'Word64' to 'PerasRoundNo'
+onPerasRoundNo ::
+  (Word64 -> Word64 -> Word64) ->
+  (PerasRoundNo -> PerasRoundNo -> PerasRoundNo)
+onPerasRoundNo = coerce
+
+-- ** Boosted blocks
+
+-- | The slot number and 32-byte hash of the block being voted for
```
tbagrel1: We should document that this is intended to be the block being voted for in real eras + HFC, not in tests

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
```diff
@@ -0,0 +1,211 @@
...
+  , pvtBlock :: !(Point blk)
+  }
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving anyclass NoThunks
+
+-- | The identifier of a vote in a Peras election
+data PerasVoteId blk
+  = PerasVoteId
+  { pviRoundNo :: !PerasRoundNo
+  , pviVoterId :: !PerasVoterId
```
tbagrel1, line 125: If we keep `PerasVoteId` concrete with such a definition, then a `Peras.Vote.V1` won't actually carry its own ID (since it will use `PerasSeatIndex` instead for voter identity). We should clarify/document this (maybe with a TODO)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
```diff
@@ -0,0 +1,211 @@
...
+-- have at the moment of the concrete type for a relative vote stake that can be
+-- compared to the quorum threshold value (also currently a 'Rational').
+newtype PerasVoteStake
+  = PerasVoteStake
+  { unPerasVoteStake :: Rational
+  }
+  deriving newtype (Eq, Ord, Num, Fractional, NoThunks, Serialise)
+  deriving stock Generic
+  deriving Show via Quiet PerasVoteStake
+  deriving Semigroup via Sum Rational
```
tbagrel1, line 159: This comment is pretty outdated; we should mark `PerasVoteStake` as `TODO: delete` probably

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
```diff
@@ -0,0 +1,211 @@
...
+  }
+  deriving newtype (NoThunks, FromCBOR, ToCBOR)
+  deriving stock (Eq, Ord, Generic)
+  deriving Show via Quiet PerasVoterId
+
+instance Serialise PerasVoterId where
+  encode = toCBOR . unPerasVoterId
+  decode = PerasVoterId <$> fromCBOR
+
+-- | Voting stake distribution for a Peras election
```
tbagrel1, line 199: Same here, we should mention this will be deleted soon

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -91,6 +91,7 @@ import System.Random
 launchBgTasks ::
   forall m blk.
   ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 94: `BlockSupportsPeras` looks more consistent here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -571,6 +572,7 @@ dumpGcSchedule (GcSchedule varQueue) = toList <$> readTVar varQueue
 -- ChainDB.
 addBlockRunner ::
   ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 575: ditto

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -346,6 +351,7 @@ triggerChainSelectionAsync CDB{cdbTracer, cdbChainSelQueue} =
 chainSelSync ::
   forall m blk.
   ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
```
tbagrel1, line 352: `BlockSupportsPeras` sounds more fit here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -793,9 +799,11 @@ data TraceEvent blk
 
 deriving instance
   ( Show (Header blk)
+  , Show (TraceAddBlockEvent blk)
+  , Show (PerasVote blk)
```
tbagrel1, line 798: `BlockSupportsPeras` sounds better here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -194,7 +204,9 @@ prop_addCertLatestCertSeenMonotonic db cert =
 
 -- | 'getLatestCertSeen' is not affected by garbage collection.
 prop_garbageCollectPreservesLatestCertSeen ::
-  (MonadSTM m, StandardHash blk) =>
+  ( MonadSTM m
+  , Eq (PerasCert blk)
```
tbagrel1, line 206: For consistency, I would use `IsPerasCert` (if it gets `Eq` as a superclass constraint as I suggested)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -201,14 +223,17 @@ implAddCert PerasCertDbEnv{pcdbTracer, pcdbState} cert = do
     pure addPerasCertRes
 
 implGetWeightSnapshot ::
-  (IOLike m, StandardHash blk) =>
+  ( IOLike m
+  , StandardHash blk
```
tbagrel1, line 226: Why `StandardHash` is needed here?

ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ObjectDiffusion/PerasVote/Smoke.hs
```diff
@@ -139,7 +156,7 @@ prop_smoke =
                 stakeDistr =
                   PerasVoteStakeDistr $
```
tbagrel1: Note that if we use mock votes here, we won't have any actual validation, so we don't need to build the effective `VoteStakeDistr` out of the generated vote list, we can use an empty one instead

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Cert/Inclusion.hs
```diff
@@ -236,9 +235,12 @@ data TestCert
   }
```
tbagrel1, line 235: Wouldn't it be possible to reuse `MockCert` here instead of defining yet another one cert type?

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
```diff
@@ -256,9 +255,12 @@ data TestCert
   }
```
tbagrel1, line 256: Same here, can't we reuse MockCert or a similar existing mock cert structure?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -93,6 +93,7 @@ import Control.Monad (unless)
 import Control.Monad.Except (runExcept)
```
tbagrel1, line 93: I'm not sure exactly what happens in this module; the deal with `Coercible (MockPerasCert blk) (PerasCert blk)` seems far from obvious to me.

If we intend to mean that `blk` should be a mock/test blk for which `PerasCert blk` has to be `MockPerasCert blk`, I think we should document it (and probably use type eq `PerasCert blk ~ MockPerasCert blk` if possible instead of `Coercible`)

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -91,6 +91,7 @@ import Data.Bifunctor
 import qualified Data.Bifunctor.TH as TH
```
tbagrel1, line 91: ditto `Coercible` -> type eq instead + doc

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/PerasVoteDB/Model.hs
```diff
@@ -14,6 +17,7 @@ module Test.Ouroboros.Storage.PerasVoteDB.Model
   ) where
```
tbagrel1, line 18: ditto Coercible -> type eq instead + doc

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
```diff
@@ -0,0 +1,40 @@
...
+  ( MockPerasVote (..)
+  , validateMockPerasVote
+  )
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
+
```
tbagrel1: Need a `TODO: replace with uninstantiable types`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
```diff
@@ -0,0 +1,48 @@
...
+  ( ShelleyBlock
+  , ShelleyCompatible
+  )
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
+
```
tbagrel1, line 41: Again, we need to replace this with concrete `Peras.Vote.V1` for >= Djisktra eras, or at least add a TODO that says we need to do it :)

ouroboros-consensus-cardano/src/unstable-byron-testlib/Ouroboros/Consensus/ByronDual/Node/Peras.hs
```diff
@@ -0,0 +1,40 @@
...
+import Ouroboros.Consensus.Peras.Vote.Mock
+  ( MockPerasVote (..)
+  , validateMockPerasVote
+  )
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
```
tbagrel1: Need a TODO: replace with uninstantiable types

ouroboros-consensus.cabal
```diff
@@ -332,8 +335,8 @@ library
     Ouroboros.Consensus.Util.Enclose
     Ouroboros.Consensus.Util.FileLock
     Ouroboros.Consensus.Util.HList
-    Ouroboros.Consensus.Util.IndexedMemPack
     Ouroboros.Consensus.Util.IOLike
+    Ouroboros.Consensus.Util.IndexedMemPack
```
tbagrel1, line 338: This diff is fishy :)

tbagrel1 (commented):
This was a long review ahah

General comments (when I didn't find an appropriate location to pin them):

- We still have `PerasVoterId` clashing with `SeatIndex`. Same with `Point blk` and `PerasBoostedBlock`

- ShelleyBlock and HFC block are using Mock Vote/Cert despite real vote and certs being available (`Peras.{Vote,Cert}.V1`)

- It also could be worth generating a list of modified files with a mention for each saying if we just updated imports/replaced constraints `StandardHash -> BlockSupportsPeras`, or if we did something meaninful in them. It could be useful for future reviewers

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Embedded inline clauses don't like dealing with type families, unfortunately.

Everything that now has an standalone `deriving` clause is the result of GHC not being able to deduce the needed context and asking us to to it ourselved (not fun).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Yep. I explicitly avoided making functional changes as much as possible in this PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
agustinmista: Yeah, I remember. I put it it here for now because we already have the `LedgerSupportsPeras` instance here, and IIRC, the consensus maintainers prefer if we can avoid adding new* orphan instances.

[*] Unless they are needed, which is the case for the new ones in this PR --- those we can't go much around unless we put them in the `Serialization` module (confusing), which also contains orphans that depend on `BlockSupportsPeras`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
agustinmista: Good idea. I added:

```haskell
-- NOTE: when replacing this with a real votes and certificates, we need to make
-- sure that their binary representation would be compatible with the one the
-- HFC would produce if it were in charge of dispatching them. Concretely, this
-- means adding an envelope around the actual votes and certificates indicating
-- which era they belong to. This is to allow for the possibility of having the
-- HFC dispatch different types of votes and certificates in the future.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: I tend to disagree for two reasons:

* This would mean that, to test this function in isolation, we would need to implement the full `BlockSupportsPeras` for any potential ad-hoc test block used by the test suite (we have a couple of those), when in reality we only care about them being structurally compatible to a Peras vote/cert.

* Using the smallest constraint needed by a given context acts as self-documentation. Some functions only need to extract information from a vote/cert (like this one), while others actually require a validating/forging them (and need `BlockSupportsPeras`).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Mock.hs
agustinmista: We might still want to validate non-crypto properties, no?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
agustinmista: See [this comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/2027#discussion_r3246775472)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Committee.hs
agustinmista: That's a good idea, but let's do it in a different PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
agustinmista: This one is a tad more consistent w.r.t. the fields used in votes and certs, IMO.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: TODO: mention this in the PR description

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: #needs-a-todo(blk-dependent Peras params)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: #needs-a-todo(incomplete BlockSupportsPeras interface)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: #needs-a-todo(decide between tight vs. loose constraints)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
agustinmista: #needs-a-todo(remove redundant types when integrating voting and aggregation)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
agustinmista: #needs-a-todo(remove redundant types when integrating voting and aggregation)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
agustinmista: #needs-a-todo(remove redundant types when integrating voting and aggregation)

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/Model.hs
agustinmista: Good catch, there's no need for Coercible now that we have moved away from `newtype instance`s

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
agustinmista: I implemented a default BlockSupportsPeras instance that uses impossible (Void) types for votes and certs, and can be derived automatically for blocks without any kind of Peras support :)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
agustinmista: I added a TODO, also:

#needs-a-todo(use real Peras types for Shelley and HFC)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/119

ouroboros-consensus-cardano/src/unstable-byron-testlib/Ouroboros/Consensus/ByronDual/Node/Peras.hs
agustinmista: See [this comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/2027#discussion_r3247771597), this is now the default behavior of `instance BlockSupportsPeras Foo`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/246

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/248

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
tbagrel1: Not sure we need a todo here; but rather direct input from the consensus team :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/249

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/249

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
tbagrel1: See https://github.com/tweag/cardano-peras/issues/249

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
tbagrel1: So far I updated the bullet list in https://github.com/tweag/cardano-peras/issues/73
I'll let @qnikst decide if we should create subtask with their dedicated issue

qnikst: @tbagrel1 item is ok for now, this way we will not lose it. We can spli to a dedicated issue in case if it can be done in later milestone or by another person.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: I consolidated these two types into a single injective `PerasError blk` type family.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Node/Run.hs
agustinmista: For now, this follows the existing convention, but we could get rid of these constraints entirely if we move them to `BlockSupportsPeras` or the `IsPerasXXX` superclasses. I would wait for the consensus team to chime in here as well.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Mock.hs
agustinmista: We decided to skip validation entirely for mocked types, and instead use an uninstantiable error type to make the validation functions "morally total".

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
agustinmista: It's needed all the way down to `addToPerasWeightSnapshot`, which needs to be able to use `Ord (Point blk)`.

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Cert/Inclusion.hs
agustinmista: This is a good idea, but I would do it in a different PR to avoid cluttering this one with too many changes.

#needs-a-todo(reuse MockCert in voting and cert inclusion rules tests)

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/Rules.hs
agustinmista: [ditto](https://github.com/IntersectMBO/ouroboros-consensus/pull/2027#discussion_r3258621653)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Ok, I took another stab at this, and I think I have a decent argument to avoid doing this for now:

* Contrary to what I expected, we end up needing _more_ constraints than before. E.g. `StandardHash blk` becomes necessary on every call site of a vote/cert projection, regardless of whether we need to `Show`, `Eq`, etc. a vote/cert's `Point blk`.

* We additionally need the same constraints for `PerasError blk`, but we don't have a dedicated class to add these as superclasses to, creating an inconsistency.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -1,251 +1,240 @@
...
-  deriving newtype (Enum, Eq, Ord, Num, Bounded, NoThunks, Serialise, NFData, ToCBOR, FromCBOR)
+-- * BlockSupportsPeras class
 
-instance Condense PerasRoundNo where
-  condense = show . unPerasRoundNo
+class
+  ( StandardHash blk
+  , Typeable blk
+  , Typeable (PerasVote blk)
+  , Typeable (PerasCert blk)
```
tbagrel1, line 58: We lack `Typeable (PerasErr blk)`. May be worth adding/replacing it with `Exception (PerasErr blk)` which includes `Typeable (PerasErr blk)` as a superclass constraint.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -258,3 +268,33 @@ instance CanHardFork xs => LedgerSupportsPeras (HardForkBlock xs) where
...
       . State.tip
       . hardForkLedgerStatePerEra
+
+{-------------------------------------------------------------------------------
+  BlockSupportsPeras
+-------------------------------------------------------------------------------}
+
+-- NOTE: this is a mocked up implementation without crypto!
+
+-- NOTE: when replacing this with a real votes and certificates, we need to make
```
tbagrel1:
```suggestion
-- TODO: when replacing this with a real votes and certificates, we need to make
```

tbagrel1 (commented):
Looks good to me!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Good catch. I'm adding `Typeable` for now, as I'm not sure if these errors will be thrown as exceptions, returned purely in an `Either`, or wrapped in some other carrier type.

We can replace it with `Exception` later on if we find it more accurate.

agustinmista:
Closing in favor of https://github.com/IntersectMBO/ouroboros-consensus/pull/2067

PR #2028, opened 2026-05-05
Simplify UTxO-HD design

jasagredo:
# Description

This PR deletes several indirections:
- `TrackingMK` is gone as it was no longer used
- The indirection of `LedgerTables` is now gone. Instead we use `Values blk`, `Keys blk`, `Diffs blk`.
- The barbies-like setup for ledger tables is gone.
- Most combinators in Tables.Combinators are simplified.
- LedgerDB V1 is disabled

Tests are yet to be fixed


jasagredo:
Discarded in favor of #2049

PR #2029, opened 2026-05-05
[Peras] Improve ChainDB's `addPeras{Vote/Cert}` methods and fix corresponding ChainDB statemachine tests

### LINKED ISSUE #121: Implementation of garbage collection of certs in ChainDB q-s-m test
This isn't very important (as we already test garbage collection in the PerasCertDB state machine test), but it could still catch if we forget to run garbage collection at all. This issue is just here as a reminder for the moment.

### LINKED ISSUE #124: Choose "realistic" boosted blocks in tests

tbagrel1:
# Description

@jasagredo identified that ChainDB statemachine tests, updated to support full Peras VoteDB/CertDB in [Peras 3.75](https://github.com/IntersectMBO/ouroboros-consensus/pull/1966/changes/fd19c52bc140868d5c629d6473db78dab5a858ff), were sometimes running into the normally-impossible `MultipleWinnersInRound` exception.

This was a very intricate bug to track down, but ultimately we identified that the model `ChainDB` would not run Peras-related garbage collection, while the SUT would; and that was affecting how a vote reusing an old-but-normally-GCed ID, but with a different boosted block, would be rejected/accepted by the model/SUT, eventually leading to `MultipleWinnersInRound` exception in the SUT. So this PR updates the model to run the same GC behavior as the SUT.

In the process, we also identified that the SUT would conditionnally discard some certs who are considered too old (for chain selection purpose), in a non-visible/non-explicit way, while the model would not. So we made this ignoring behavior explicit, and updated the model to have the same behavior as the SUT, in order to prevent future annoying debugging in statemachine tests, even though it seemed not to create visible issues for now.

Finally, to prevent running into other `MultipleWinnersInRound` exceptions in the future, the generator for `AddPerasVote` commands in ChainDB's statemachine tests has been improved and documented, and now limits the total stake distributed in a given round to never reach 2 x quorum.

The only breaking changes are ChainDB's API methods `addPerasVoteSync`, `addPerasVoteWithAsyncCertHandling`, `addPerasCertSync`, `addPerasCertAsync` who get updated return types.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
@@ -33,6 +33,7 @@ import Control.Exception (assert)
 import Control.Monad (forM_, join, void, when)
 import Control.Monad.Except ()
 import Control.Monad.Trans.Class (lift)
+import Control.Monad.Trans.Except (ExceptT, runExceptT, throwE)
 import Control.Monad.Trans.State.Strict
 import Control.Tracer (Tracer, nullTracer, traceWith)
 import Data.Bifunctor (first)
@@ -70,6 +71,7 @@ import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.Storage.ChainDB.API
   ( AddBlockPromise (..)
   , AddBlockResult (..)
+  , AddPerasCertChainSelOutcome (..)
   , AddPerasCertPromise (..)
   , BlockComponent (..)
   , ChainType (..)
@@ -97,7 +99,6 @@ import Ouroboros.Consensus.Storage.VolatileDB (VolatileDB)
 import qualified Ouroboros.Consensus.Storage.VolatileDB as VolatileDB
 import Ouroboros.Consensus.Util
 import Ouroboros.Consensus.Util.AnchoredFragment
-import Ouroboros.Consensus.Util.EarlyExit (exitEarly, withEarlyExit_)
 import Ouroboros.Consensus.Util.Enclose (encloseWith)
 import Ouroboros.Consensus.Util.IOLike
 import Ouroboros.Consensus.Util.STM (WithFingerprint (..))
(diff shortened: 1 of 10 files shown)


tbagrel1:
FTR, [this commit](https://github.com/IntersectMBO/ouroboros-consensus/commit/0981bae8e6072bb11e95dd03fb079e94f961309e) (pinned here to prevent garbage collection) introduced improved `ToExpr` instances for Peras stuff to enable easier debugging. We may want to revive this for a future PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -537,6 +543,14 @@ chainSelSync cdb@CDB{..} (ChainSelAddPerasCert cert varProcessed) = do
   boostedBlock :: Point blk
   boostedBlock = getPerasCertBoostedBlock cert
 
+  -- \| Run a block that can exit early with a result value.
+  withEarlyExit :: ExceptT a (Electric m) a -> Electric m a
+  withEarlyExit = fmap (either id id) . runExceptT
+
+  -- \| Exit early with the given result.
+  exitEarly :: a -> ExceptT a (Electric m) b
+  exitEarly = throwE
```
jasagredo: It is confusing that you use `withEarlyExit` as there is a type exactly with that name used in block forging for example

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -1112,7 +1116,7 @@ generator loe genBlock genPerasBlock m@Model{..} =
               LoEDisabled ->
                 -- We must reduce the probability of Peras events to trigger on
                 -- an empty DB since there are no interesting blocks to vote for
-                if empty then 20 else 435
+                if empty then 20 else 500
```
jasagredo, line 1119: Any particular reason for this?

jasagredo (approved):
Minor suggestions, overall looks good

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
tbagrel1: Initially this piece of code was importing the `withEarlyExit` you mention, but it was too limited for our use-case, so I had to implement slightly more powerful helper functions.

What would you suggest we do here?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
tbagrel1: Yes, with all the changes made in this PR, to see as many votes created through vote aggregation as certs received, we had to tweak the number slightly.

If we don't, the comment above this code is no longer fully exact :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
jasagredo: I see. I assume this is because you want "early exit with a value" and not just "early exit".

Something like `EarlyExitId`?

PR #2030, opened 2026-05-05
Remove LedgerDB V1 and the LMDB backing store

jasagredo:
# Description

Retire the V1 LedgerDB and LMDB dependency along with the now-unreachable plumbing in the API, the DBAnalyser, and the snapshot-converter.

The change is structured as four commits so review can follow the layers:

  1. **Remove LedgerDB V1 implementation** — drop the `V1.*` module tree (DbChangelog, DiffSeq, Forker, Lock, Snapshots, BackingStore + LMDB/InMemory backends) and its wiring into `openDB`, `LedgerDbBackendArgs`, `FlavorImplSpecificTrace`, and
  `SeqDiffMK`.
  2. **Remove remaining V1 LedgerDB references** — V1-only DBAnalyser/snapshot-converter options, the V1 BackingStore Lockstep/Mock and DbChangelog/DiffSeq/LMDB test suites, the now-degenerate `SerializeTablesHint` type family, unused cabal dependencies,
   and stale V1 notes in the Mempool and API haddocks.
  3. **Remove `tryFlush` from the LedgerDB API** — `tryFlush` only ever did work in V1; in V2 it was a no-op. V2's snapshot work is now run in a forked thread so it doesn't block the caller.
  4. **Drop the second monad parameter from `SnapshotManager`** — V1 took snapshots in `ReadLocked m`, which is why `SnapshotManager` carried two monads (`m` and `n`). With V1 gone, `n` collapses into `m`.

 ## Key changes

  - Deleted modules: `Ouroboros.Consensus.Storage.LedgerDB.V1` and all its submodules (`V1.Args`, `V1.BackingStore`, `V1.BackingStore.API`, `V1.BackingStore.Impl.InMemory`, `V1.DbChangelog`, `V1.DiffSeq`, `V1.Forker`, `V1.Lock`, `V1.Snapshots`).
  - Deleted public sublibrary `ouroboros-consensus:lmdb` (which exposed `V1.BackingStore.Impl.LMDB`, `.Bridge`, `.Status`).
  - Removed `SeqDiffMK` and the `SerializeTablesHint` type family.
  - Removed the `tryFlush` field of `LedgerDB`.
  - Removed the V1 alternatives `LedgerDbBackendArgsV1` and `FlavorImplSpecificTraceV1`; both types are now `newtype`s with only the V2 constructor.
  - Removed the db-analyser `--lmdb` flag and the `V1LMDB` constructor of `LedgerDBBackend`.
  - Removed the snapshot-converter `--monitor-lmdb-snapshots-in` / `--input-lmdb` / `--output-lmdb` flags and the `LMDB` constructor of `StandaloneFormat`.

## Type changes

  - `SnapshotManager m n blk st` → `SnapshotManager m blk st`. `initialize`, `snapshotsMapM_`, `destroySnapshots`, and `trimSnapshots` lose the `n` parameter accordingly.
  - `SerializeTablesWithHint` methods (`encodeTablesWithHint` / `decodeTablesWithHint`) and their `default*` / `trivial*` helpers now take `l blk EmptyMK` directly instead of `SerializeTablesHint l (LedgerTables blk ValuesMK)`.
  - `openDB` and `Cardano.Tools.DBAnalyser.Run.openLedgerDB` no longer require the `LedgerDbSerialiseConstraints blk` constraint.


bladyjoker (approved):
🥳 

Let me know if you need me to look at something specific, all this seems pretty non controversial.

bladyjoker:
You might want to try integrate this change with cardano-node? There's configuration options there that need retiring right?

PR #2032, opened 2026-05-06
[Leios prototype] LeiosDb InMemory improvements

bladyjoker:
DONE
- [x] Fix contention by splitting the state into multiple TVars
- [x] Fix laziness issues that causes all nodes to critically error out due to timeouts

At https://github.com/IntersectMBO/ouroboros-consensus/pull/2032/commits/3ff2960e3abe2a0ac082bc58848461291eaea54f we have are still reaching the breaking point at which the `leiosDbInsertTxs` takes a very long time and causes general timeouts to be reported.

image

Along with a divergence in Mempools that can sometimes recover

image

The deeper cause is that the EB storage takes increasingly longer times (>1s) that eventually leads to nodes dying.


bladyjoker:
Instrumenting the forge thread, this is looking weird

image

Seems like it's MOSTLY In `partition-mempool``

image

And both the "partitioning logic" and the "snapshotting" are taking an absurd amount of time

image

PR #2033, opened 2026-05-07
[Peras 25.5] Introduce Bytes32RealPoint

agustinmista:
This PR introduces a specialization of `RealPoint` for blocks where the header hash is always 32 bytes long.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
@@ -1,8 +1,11 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Ouroboros.Consensus.Block.Abstract
@@ -89,7 +92,9 @@ import Data.ByteString.Short (ShortByteString)
 import qualified Data.ByteString.Short as Short
 import Data.Kind (Type)
 import Data.Maybe (isJust)
+import Data.Proxy (Proxy (..))
 import Data.Word (Word32, Word64)
+import GHC.TypeNats (KnownNat, Nat, natVal)
 import NoThunks.Class (NoThunks)
 import Ouroboros.Consensus.Block.EBB
 import Ouroboros.Network.Block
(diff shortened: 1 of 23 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs
```diff
@@ -120,3 +133,53 @@ castRealPoint ::
...
+fromBytes32RealPoint (Bytes32RealPoint s h) =
+  RealPoint s (coerce h)
+
+toBytes32RealPoint ::
+  forall blk.
+  Coercible (HeaderHash blk) ShortByteString =>
+  RealPoint blk ->
+  Bytes32RealPoint
+toBytes32RealPoint (RealPoint s h) =
+  assert (ByteString.length (coerce h) == 32) $
```
jasagredo: I am not sure this is how it should be. I think this is unsafe as assertions are not checked in production

agustinmista: I would have expected this to be an invariant for any `blk` such that:
```haskell
type instance HeaderHash blk = ShortByteString
```

But if that's not the case I guess we will need to deal with the partial conversion more explicitly. 

Would you prefer this to be discharged with a `Maybe` or with an `error`?

tbagrel1: IIRC:
- For any `blk` of a concrete era (`OneEra x`); or the `HardForkBlock xs`, `HeaderHash blk` is semantically isomorphic to a 32-byte bytestring, but it can take slightly more compact forms (e.g. 4 `Word64` zipped together for some eras)
- For other `blk`s (e.g. `TestBlock`), we can't make any assumption on the shape of `HeaderHash blk`

agustinmista: Thanks @tbagrel1 for the extra context.

FTR: we still need to decide how to discharge errors when coercing from a (potentially) non-32bytes-long `ShortByteString` into `Bytes32RealPoint`, as this is not captured at the type level.

What we need to address concretely here is whether we:

1. Capture this with a `Maybe` and let the client code (likely the HFC?) handle it.
2. Assume it can't happen and put an `error` for the impossible case.
3. Assume it can't happen and wrap the expression with an assertion.
4. Something else?

Then (1) and (2) will come with some minor performance penalty (length check and optionally wrapping on `Maybe`) that we can avoid if we already know that `forall blk. Coercible (HeaderHash blk) ShortByteString` implies `len(header_hash) == 32` in any _production_ context where this function could be invoked. On the other hand (3) should have no performance penalty in production, while still being able to detect issues in testing.

I'm OK with any of these approaches, but I would like to hear back from @jasagredo before making any change :)

tbagrel1: @jasagredo

Hey we just discussed this with @agustinmista, and here's a more refined context/question:

1. We will only convert from `(Real)Point blk` to `Bytes32RealPoint` in contexts in which it's morally safe to do it; that is, for `blk` of a concrete era or the HardForkBlock. So ideally the conversion functions should probably NOT return a `Maybe` (to avoid having to deal with impossible cases at call site). Do you agree?
2. Is it ok to assume that `OneEra blk` or `AllEra blks` (For `HardForkBlock blks`) are the right classes that carry the moral invariant that a header hash carries 32 bytes of information? If not, what would be the right class?
3. If the previous one is correct, would it be ok to create a conversion class from `(Real)Point blk` to `Bytes32RealPoint` and add it as a superclass constraint for `OneEra blk` ? This way I think it would be easier to implement the converter for the different concrete shape of 32-bytes headerhash (e.g. Bytestring, 4 `Word64`s, etc)

agustinmista:
Hey @jasagredo! :hand: 

Do you think you could take a look at the remaining comment on this PR? This is currently blocking [Peras 26](https://github.com/IntersectMBO/ouroboros-consensus/pull/1938) from being merged :sweat_smile: 

Thanks in advance! 

CC @dnadales in case you have time to chime in instead.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs
dnadales: I'd feel more comfortable if Javier would chime in. Re 1, It seems that to drop the `Maybe` from the equation we need some type-level assurance, don't we? 

Re 2, I don't have the that is the case, do you know why you think that'd be the case? I mean, the header hash for a HFC block is defined in terms of the header hash of the eras it contains, and in this case what happens if an era is added with a different size :thinking: 

Re 3 the conversion class makes sense. Sorry but I lack some more context on why addign it as a superclass constraint for `OneEra` would make sense. HFC test blocks can't satisfy it, so won't this bite us?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs
```diff
@@ -120,3 +134,54 @@ castRealPoint ::
...
+    ]
+
+decodeBytes32RealPoint :: forall s. Decoder s Bytes32RealPoint
+decodeBytes32RealPoint = do
+  enforceSize "Bytes32RealPoint" 2
+  s <- decode
+  h <- decode
+  case ByteString.length h of
+    32 -> pure (Bytes32RealPoint s h)
+    len -> fail $ "decodeBytes32RealPoint: expected 32 bytes, got " <> show len
```
dnadales, line 171: Should we add a couple of unit tests here while the type is fresh? In particular the `len -> fail` branch and an `encode`/`decode` round-trip, since those are the paths the sibling (#1938) round-trip doesn't reach (it only ever feeds valid 32-byte hashes). If we go the witness-class route, the `to`/`from` round-trip becomes a nice property test too.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs
```diff
@@ -120,3 +134,54 @@ castRealPoint ::
...
+data Bytes32RealPoint = Bytes32RealPoint !SlotNo !ShortByteString
+  deriving (Show, Eq, Generic, NoThunks)
+
+bytes32RealPointSlot :: Bytes32RealPoint -> SlotNo
+bytes32RealPointSlot (Bytes32RealPoint s _) = s
+
+bytes32RealPointHash :: Bytes32RealPoint -> ShortByteString
+bytes32RealPointHash (Bytes32RealPoint _ h) = h
+
+encodeBytes32RealPoint :: Bytes32RealPoint -> Encoding
```
dnadales, line 156: Since #1938 bakes this encoding into the Peras cert/vote format, is the bare 2-tuple the shape we want long-term? With no version tag, adding a field later is a breaking wire change. Worth a comment noting the format is intended to be stable, or is the envelope in #1938 already handling versioning?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs
jasagredo: I think I would:
1. Create a class that defines the conversion from the era hash to bytes32.
2. That method can be called by RealPoint -> Bytes32RealPoint. (i.e. I don't think the class should only deal with hashes).
3. Add it as a constraint to `SingleEraBlock blk`.
4. By using `hc* proxySingle` functions, you can invoke this constraint when you deal with particular eras. Then it should be as easy as to convert the summand to `K Bytes32RealPoint` and then use `hcollapse` to get the era-less Bytes32RealPoint

This way if an era doesn't have a 32Bytes hash, we will need to re-do this mechanism which is bad, but infinitely less bad than find out with a runtime crash that we forgot about this.

tbagrel1:
After discussing it with @dnadales and @jasagredo during Consensus weekly on 2026-06-29, we decided to:
- improve `ConvertRawHash` class to have an associated type `HashSize :: Nat`
- change the `fromRawHash` and `fromShortRawHash` methods to enforce the announced size, and thus return a `Maybe`
- introduce `unsafeFromRawHash` / `unsafeFromShortRawHash` for backwards compatility for call sites in which we don't have an easy way to handle failure
- Make it explicit in documentation that the hash size is now enforced by default, except if one opts out through the unsafe functions

With these changes, the Peras-related conversion functions from/to `Bytes32RealPoint` / `HeaderHash` can use constraints `(ConvertRawHash blk, HashSize blk ~ 32)` and become total, without need for asserts. This way we don't have to introduce yet another conversion class.

@dnadales and @jasagredo please amend this text if you see something inaccurate, thanks!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -230,32 +235,68 @@ headerPoint = castPoint . blockPoint
...
+  --
+  -- PRECONDITION: the input's size must match 'hashSize'.
+  unsafeFromRawHash :: proxy blk -> Strict.ByteString -> HeaderHash blk
+  unsafeFromRawHash p = unsafeFromShortRawHash p . Short.toShort
+
+  -- | Construct the hash from a raw hash /without/ enforcing the size.
+  --
+  -- PRECONDITION: the input's size must match 'hashSize'.
+  unsafeFromShortRawHash :: proxy blk -> ShortByteString -> HeaderHash blk
+  unsafeFromShortRawHash p = unsafeFromRawHash p . Short.fromShort
```
jasagredo, line 291: This is a change in API and as such, it has to be reflected in the changelog

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
```diff
@@ -1,10 +1,17 @@
...
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
 
-module Ouroboros.Consensus.HardFork.Combinator.Abstract.CanHardFork (CanHardFork (..)) where
+module Ouroboros.Consensus.HardFork.Combinator.Abstract.CanHardFork
+  ( CanHardFork (..)
+  , EqualHashSizeOfHead
+  , HashSizeOfHead
```
jasagredo: Do you need to expose these?

jasagredo: Seems like you only might need `HashSizeOfHead` for the HF instance?

jasagredo (changes requested):

jasagredo (approved):

PR #2037, opened 2026-05-11
Add Running Genesis Mode HOWTO

dnadales:
## Summary

- Add an operational HOWTO for running Genesis-mode nodes, covering configuration, networking, small-testnet tuning, stall debugging, and tracer setup.

## Motivation

The existing Genesis documentation (`genesis_design.md`) describes the design for implementers.
There was no documentation for people actually running, configuring, or debugging Genesis-mode nodes.

This gap became apparent while investigating sync stalls on the dijkstra and leios testnets (3 relays each).
The root cause was that `MinBigLedgerPeersForTrustedState` (default 5) prevents the HAA from being satisfied when the network has fewer than 5 big ledger peers — a fact not documented anywhere.
Lowering `SyncTargetNumberOfActiveBigLedgerPeers` alone is not sufficient; the HAA gate must also be lowered.

The new doc distills the debugging knowledge acquired during that investigation into reusable guidance.

### DIFF
--- a/docs/website/contents/howtos/running_genesis_mode.md
+++ b/docs/website/contents/howtos/running_genesis_mode.md
@@ -3,43 +3,51 @@
 ## When to use Genesis mode
 
 Genesis mode lets a syncing node defend against long-range attacks without trusting recent ledger state for peer evaluation.
-It adds defensive components (GSM, LoE, GDD, CSJ, LoP, DBF) on top of existing ChainSync/BlockFetch.
+It adds defensive components — LoE, GDD, CSJ, LoP, and Devoted BlockFetch (DBF) — on top of existing ChainSync/BlockFetch, coordinated by the Genesis State Machine (GSM).
+The GSM itself runs in both modes: it tracks whether the node is caught up and disables the Genesis-only components once it is.
 See the [Genesis design doc](../references/miscellaneous/genesis_design.md) for the full component description.
 
 Use Genesis mode on **relay nodes and syncing nodes** that connect to the public network.
 
-Use Praos mode on **block-producing nodes** that connect only to trusted local relays with `useLedgerAfterSlot: -1`.
-A forger with `GenesisMode` + trusted `localRoots` + `useLedgerAfterSlot: -1` will behave like Praos mode in practice (the peer selection targets only differ when `(GenesisMode, TooOld)`), but there is no benefit over Praos mode in that configuration.
+Use Praos mode on **block-producing nodes** that connect only to trusted local relays.
+Setting `useLedgerAfterSlot: -1` disables ledger peers, so the node uses only its configured roots (`-1` turns ledger peers off; `0` would mean "always use the ledger").
+A forger with `GenesisMode` + trusted `localRoots` + ledger peers disabled will behave like Praos mode in practice — the peer-selection target basis differs only in the `(GenesisMode, TooOld)` state — but there is no benefit over Praos mode in that configuration.
 
 ## Required configuration
 
-The following `config.json` fields are prerequisites for Genesis mode:
+Genesis mode is turned on by one `config.json` field:
 
 ```json
-"ConsensusMode": "GenesisMode",
-"UseTraceDispatcher": true,
-"TurnOnLogging": true
+"ConsensusMode": "GenesisMode"
 ```
 
 `ConsensusMode` activates the Genesis defenses.
-Praos-mode nodes do not emit GSM, CSJ, or GDD events at all — the tracers exist but are never written to.
+The GSM runs in both modes, so a Praos-mode node still emits GSM events.
+CSJ and GDD are Genesis-only defenses: a Praos-mode node never writes to those tracers.
 
-The node also requires a peer snapshot file (`peer-snapshot.json`) listed in the topology, which provides the initial set of peers for ledger peer selection.
+Two other `config.json` fields only take effect in Genesis mode: `MinBigLedgerPeersForTrustedState` (see [Small testnets](#small-testnets)) and `LowLevelGenesisOptions`.
+
+The topology may also list a big-ledger-peer snapshot via the optional `peerSnapshotFile` field (a file path).
+It seeds ledger peer selection with an initial set of big ledger peers.
+It is not required: in Genesis mode with ledger peers enabled, a node without one only logs a recommendation to supply it.
 
 ## Host networking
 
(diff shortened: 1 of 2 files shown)


docs/website/contents/howtos/running_genesis_mode.md
```diff
@@ -0,0 +1,178 @@
...
+A forger with `GenesisMode` + trusted `localRoots` + `useLedgerAfterSlot: -1` will behave like Praos mode in practice (the peer selection targets only differ when `(GenesisMode, TooOld)`), but there is no benefit over Praos mode in that configuration.
+
+## Required configuration
+
+The following `config.json` fields are prerequisites for Genesis mode:
+
+```json
+"ConsensusMode": "GenesisMode",
+"UseTraceDispatcher": true,
+"TurnOnLogging": true
```
coot: Last two are generic tracing options, rather than something `Genesis` related, using `UseTraceDispatcher` is probably not even necessary, e.g. one can run the new tracing system in a mode which traces to `stdout` rather than through the `TraceDispatcher`.

docs/website/contents/howtos/running_genesis_mode.md
```diff
@@ -0,0 +1,178 @@
...
+The node also requires a peer snapshot file (`peer-snapshot.json`) listed in the topology, which provides the initial set of peers for ledger peer selection.
+
+## Host networking
+
+The node binds listening sockets and creates outbound connections based on `HostIPv4Addr` and `HostIPv6Addr` in `config.json`:
+
+```json
+"HostIPv4Addr": "0.0.0.0",
+"HostIPv6Addr": "::"
+```
```
coot: I never used these config options (are you sure they exist?), there are CLI options for them in the node.

crocodile-dentist:
lgtm, and indeed there is documentation on configuring and running in `GenesisMode` on the [dev portal](https://developers.cardano.org/docs/get-started/infrastructure/node/topology/), in particular the note about `MinBigLedgerPeersForTrustedState` is mentioned.

docs/website/contents/howtos/running_genesis_mode.md
dnadales: Right — these are generic tracing options, not Genesis prerequisites, so they've been dropped from the required-config section.
Checking against `master`, the logging switch was in fact removed in [289dee332](https://github.com/IntersectMBO/cardano-node/commit/289dee332f20b97daaf19c019c1aaf4de114ab90) (2026-05-18): `UseTraceDispatcher` and `TurnOnLogging` are no longer parsed and the new tracing system runs unconditionally, so nothing needs enabling.
The tracer section now just notes that and points out the switch is gone on current nodes.
Fixed in d003d27ee.

docs/website/contents/howtos/running_genesis_mode.md
dnadales: Good catch — those config.json keys don't exist.
The host addresses come only from the `--host-addr` / `--host-ipv6-addr` CLI options (`Cardano/Node/Parsers.hs`); there's no `HostIPv4Addr`/`HostIPv6Addr` in the config parser, and `SocketConfig` has no `FromJSON` instance.
The section now uses the CLI options instead.
The IPv6 guidance was also corrected: the node dials literal IPv6 snapshot peers regardless of `--host-ipv6-addr` (the local address only affects socket binding and the DNS lookup family), so omitting it doesn't make the node skip them — the fix is host routing or dropping the IPv6 peers from the snapshot.
Fixed in d003d27ee.

docs/website/contents/howtos/running_genesis_mode.md
```diff
@@ -0,0 +1,196 @@
...
+
+Genesis mode lets a syncing node defend against long-range attacks without trusting recent ledger state for peer evaluation.
+It adds defensive components — LoE, GDD, CSJ, LoP, and Devoted BlockFetch (DBF) — on top of existing ChainSync/BlockFetch, coordinated by the Genesis State Machine (GSM).
+The GSM itself runs in both modes: it tracks whether the node is caught up and disables the Genesis-only components once it is.
+See the [Genesis design doc](../references/miscellaneous/genesis_design.md) for the full component description.
+
+Use Genesis mode on **relay nodes and syncing nodes** that connect to the public network.
+
+Use Praos mode on **block-producing nodes** that connect only to trusted local relays.
+Setting `useLedgerAfterSlot: -1` disables ledger peers, so the node uses only its configured roots (`-1` turns ledger peers off; `0` would mean "always use the ledger").
```
jasagredo, line 13: Unsure where this config is? I don't think it is in the configuration file. Perhaps in the topology?

docs/website/contents/howtos/running_genesis_mode.md
```diff
@@ -0,0 +1,196 @@
...
+
+```json
+"MinBigLedgerPeersForTrustedState":            1,
+"SyncTargetNumberOfRootPeers":                 0,
+"SyncTargetNumberOfKnownPeers":                1,
+"SyncTargetNumberOfEstablishedPeers":           1,
+"SyncTargetNumberOfActivePeers":                1,
+"SyncTargetNumberOfKnownBigLedgerPeers":        1,
+"SyncTargetNumberOfEstablishedBigLedgerPeers":  1,
+"SyncTargetNumberOfActiveBigLedgerPeers":       1
```
jasagredo, line 73: These will not be top-level fields anymore after `cardano-config`

docs/website/contents/howtos/running_genesis_mode.md
```diff
@@ -0,0 +1,196 @@
...
+`Consensus.GSM.InitializedInPreSyncing` → `PreSyncingToSyncing` → eventually `EnterCaughtUp`.
+Steady `Consensus.CSJ.SentJumpInstruction`.
+Periodic `Consensus.GDD.TraceGDDEvent`.
+`BlockFetch.Decision.PeersFetch` with one peer fetching.
+ChainDB tip extending.
+
+## Tracer configuration
+
+The overrides below use cardano-node's new tracing system, which is always on — no separate flag enables it.
+(Older nodes gated it behind `UseTraceDispatcher`/`TurnOnLogging`, both generic and defaulting to on; current cardano-node has removed that switch.)
```
jasagredo, line 147: Being the default, we probably can omit this info

jasagredo (commented):
I think this is in general useful and describes possible causes that would have taken me quite some time to figure out. 

The only remark is that this might not belong to Consensus, but instead to the node wiki, dev portal, or some such.

PR #2038, opened 2026-05-12
Peras: Add asserts in the implForgeCert

### LINKED ISSUE #240: UniqueVotesWithSameTarget caller contract: cmpVotes must be seat-index-aware
## Problem

PR #2020 made `UniqueVotesWithSameTarget` a properly opaque carrier whose construction checks for duplicates, replacing the silent-dedup hazard at the type level (option (a) of the original `forgeCert` dedup concern).
The remaining concern is at the caller side: uniqueness is parameterized by the `cmpVotes :: Vote crypto committee -> Vote crypto committee -> Ordering` function the caller supplies, so the duplicate-rejection guarantee is only as strong as that comparator.

For `WFALS` and `EveryoneVotes` the underlying hazard is two votes sharing a `SeatIndex` but differing in signature / VRF output:

- `WFALS.implForgeCert` (`WFALS.hs:456`) calls `NEMap.fromAscList voters` while signatures are aggregated separately.
  If two `voters` entries share a `SeatIndex`, the cert's seat-index map has fewer entries than the signature aggregate, and downstream `verifyCert` fails with the opaque "number of keys and signatures do not match" error.
- `EveryoneVotes.implForgeCert` (`EveryoneVotes.hs:265`) has the same shape via `NESet.fromList voters`.

The Peras integration site that builds `UniqueVotesWithSameTarget` from inbound votes must therefore choose a `cmpVotes` that returns `EQ` for any two votes that would collide at the seat-index level (or, more conservatively, on underlying voter identity).

## Proposed fix

- Locate the call site that builds `UniqueVotesWithSameTarget` for the Peras voting committee.
- Confirm `cmpVotes` orders `EQ` on shared `SeatIndex` for both persistent and non-persistent variants of `WFALS` and `EveryoneVotes` `Vote` constructors.
- Add a property test that generates votes with seeded seat-index collisions and asserts `ensureUniqueVotesWithSameTarget` returns `Left (DuplicateVotes …)` for them.

## References

- [ouroboros-consensus PR #1975 reviews r3131282442 and r3131290517](https://github.com/IntersectMBO/ouroboros-consensus/pull/1975#discussion_r3131282442)
- Closing of option (a) in [PR #2020](https://github.com/IntersectMBO/ouroboros-consensus/pull/2020)

qnikst:
There is a chance that incompatible ordering functions may be passed to the `verifyCert`. And we need to catch this problem sooner and prevent from building incorrect cert with mismatched amount of votes, so we add a check that the size of the deduplicated voters and certs is the same.

See https://github.com/tweag/cardano-peras/issues/240 for the deeper discussion.

# Description

this PR might resolve.

Also note that:

# WARNING

To update your feature branch if it's stale, please rebase it manually on top of `main`. Don't update your feature branch by merging `main` into it. Your pull request will not pass CI if you do.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
@@ -23,6 +23,7 @@ module Ouroboros.Consensus.Committee.EveryoneVotes
 
 import Cardano.Ledger.BaseTypes (NonZero)
 import Cardano.Ledger.BaseTypes.NonZero (NonZero (..), nonZero)
+import Control.Exception (assert)
 import Control.Monad.Zip (MonadZip (..))
 import qualified Data.Array as Array
 import Data.Bifunctor (Bifunctor (..))
@@ -251,19 +252,29 @@ implForgeCert ::
   Either
     (VotingCommitteeError crypto EveryoneVotes)
     (Cert crypto EveryoneVotes)
-implForgeCert votes = do
-  aggSig <-
-    bimap CryptoError id $ do
-      aggregateVoteSignatures
-        (Proxy @crypto)
-        voteSignatures
-  pure $
-    EveryoneVotesCert
-      (getElectionIdFromVotes votes)
-      (getVoteCandidateFromVotes votes)
-      (NESet.fromList voters)
-      aggSig
+implForgeCert votes =
+  assert allUniqueVoterSeats $ do
+    aggSig <-
+      bimap CryptoError id $ do
+        aggregateVoteSignatures
+          (Proxy @crypto)
+          voteSignatures
+
+    pure $
+      EveryoneVotesCert
+        (getElectionIdFromVotes votes)
+        (getVoteCandidateFromVotes votes)
+        voterMap
+        aggSig
  where
+  -- We want to fail fast in case if the number of votes is not the same
+  -- as voters representation in EveryoneVotes. This assumption may be
+  -- violated in case if implementation uses incorrect ordering function
+  -- in the `ensureUniqueVotesWithSameTarget`
+  allUniqueVoterSeats =
+    NESet.size voterMap == length voters
+      && NESet.size voterMap == length voteSignatures
+  voterMap = NESet.fromList voters
   (voters, voteSignatures) =
     munzip $ flip fmap votesInAscendingSeatIndexOrder $ \case
       EveryoneVotesVote seatIndex _ _ sig ->
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
@@ -41,6 +41,7 @@ module Ouroboros.Consensus.Committee.WFALS
   ) where
 
 import Cardano.Ledger.BaseTypes (NonZero (..), Nonce, nonZero)
+import Control.Exception (assert)
 import Control.Monad (void)
 import Control.Monad.Zip (MonadZip (..))
 import qualified Data.Array as Array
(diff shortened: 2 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -444,11 +450,20 @@ implForgeCert votes = do
       aggregateVoteSignatures
         (Proxy @crypto)
         voteSignatures
+  let voterSeats = NEMap.fromAscList voters
   pure $
     WFALSCert
       (getElectionIdFromVotes votes)
       (getVoteCandidateFromVotes votes)
-      (NEMap.fromAscList voters)
+      -- We want to fail fast in case if the number of votes is the same
```
tbagrel1: typo, probably you meant "is **not** the same"?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -444,11 +450,20 @@ implForgeCert votes = do
...
+      -- as voters representation in WFALSCert. This assumption may be
+      -- violated in case if implementation uses incorrect ordering function.
```
tbagrel1: not directly obvious that we are talking about `ensureUniqueVotesWithSameTarget`, we should probably be explicit

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -444,11 +450,20 @@ implForgeCert votes = do
       aggregateVoteSignatures
         (Proxy @crypto)
         voteSignatures
+  let voterSeats = NEMap.fromAscList voters
```
tbagrel1: voterSeats is actually a list of tuples, not a list of seat indices, so I would prefer a different name (ditto below). Something like `voterMap` would be fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -257,11 +259,17 @@ implForgeCert votes = do
       aggregateVoteSignatures
         (Proxy @crypto)
         voteSignatures
+  let votersSeats = NESet.fromList voters
   pure $
     EveryoneVotesCert
       (getElectionIdFromVotes votes)
       (getVoteCandidateFromVotes votes)
-      (NESet.fromList voters)
+      ( assert
```
tbagrel1: I would like to see the same comment here as the one below :)

tbagrel1 (approved):
I've left a couple of comments (mostly nitpicking), feel free to undraft after addressing them :)

qnikst:
Thanks @tbagrel1 for the review. I've addressed all the comments and undrafted PR

agustinmista:
I agree with @tbagrel1's review. In addition I would:

* move the `voterSeats` binding to the `where` block, and raise the assertion to the top (i.e., the first thing the function does instead of being buried inside the constructor). Something like:

```haskell
implForgeCert votes = 
  assert allUniqueVoterSeats $ do
    ...
    pure $
      WFALSCert
        (getElectionIdFromVotes votes)
        (getVoteCandidateFromVotes votes)
        ascVotersMap
        aggSig
 where
  -- We want to fail fast in case if the ...
  allUniqueVoterSeats =
     NEMap.size ascVotersMap == length voters
      && NEMap.size ascVotersMap == length voteSignatures
  ascVotersMap = 
    NEMap.fromAscList voters
  (voters, voteSignatures) =
    ...
```

qnikst:
I've addressed comments by @agustinmista, now PR should be fine for the final review

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/EveryoneVotes.hs
```diff
@@ -251,19 +252,28 @@ implForgeCert ::
...
+
+      pure $
+        EveryoneVotesCert
+          (getElectionIdFromVotes votes)
+          (getVoteCandidateFromVotes votes)
+          voterMap
+          aggSig
  where
+  -- We want to fail fast in case if the number of votes is not the same
+  -- as voters representation in WFALSCert. This assumption may be
```
dnadales: I wonder if "WFALSCert" here should be "EveryoneVotesCert"? Looks like it survived the copy-paste from `WFALS.hs`.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
```diff
@@ -438,19 +439,27 @@ implForgeCert ::
...
+          voterMap
+          aggSig
  where
+  -- We want to fail fast in case if the number of votes is not the same
+  -- as voters representation in WFALSCert. This assumption may be
+  -- violated in case if implementation uses incorrect ordering function
+  -- in the `ensureUniqueVotesWithSameTarget`
+  allUniqueVoterSeats = NEMap.size voterMap == length voters
+        && NEMap.size voterMap == length voteSignatures
+  voterMap = NEMap.fromAscList voters
```
dnadales, line 463: Should we align the continuation here with the rest of the `where` block? Something like:

```haskell
  allUniqueVoterSeats =
    NEMap.size voterMap == length voters
      && NEMap.size voterMap == length voteSignatures
```

dnadales (approved):
LGTM, thanks for adding these assertions!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFALS.hs
agustinmista: Yes, this would otherwise fail the linting job.

@qnikst I forgot to mention, but if you haven't yet setup your editor to format on save, you can always do:

```console
./scripts/ci/run-fourmolu.sh 
```

To reformat the repo :smile:

qnikst: Yes, I've did that don't know why I have not applied it after the last round.

qnikst:
Thanks, @dnadales for the review I've addressed and rebased my PR on top of the main.

qnikst:
@dnadales seems some workflows needs approval here

PR #2039, opened 2026-05-12
Leios: Vote signing and committee selection

### LINKED ISSUE #790: Prototype Voting

ch1bo:
Refs https://github.com/input-output-hk/ouroboros-leios/issues/790.

Builds on the initial Leios voting prototype (#1963) by adding committee selection and real signature creation/validation, replacing the placeholder voting from #1963. This is also built on the Leios prototype remake (still in flight as #2041).

- **Committee selection (per EB) from the stake distribution.** A `HasLeiosVoting` decides whether (and as which `VoterId`) the local node votes on an EB, derived from the Shelley `PoolDistr`. For now, this only runs in the latest era and uses a basic selection scheme.
- **Real BLS signatures.** Switched from placeholder/fake signing to actual BLS signatures, using a more recent `cardano-crypto-class` (now available after the remake #2041)
- **Voting key material.** BLS keys are still derived, but from the cold verification key hash (pool id) instead. This allows us to skip the key registration, but is obviously still unsafe.
- **Tests.** `LeiosVoteState` unit tests cover the new validation API; ThreadNet expectations remain green.

TODO

- [x] ThreadNet property tests pass (votes diffuse and are validated)
- [x] `LeiosVoteState` unit tests pass
- [x] Integrate golden `LeiosVote` test into existing machinery


ch1bo:
@kderme I'm hesitant in fixing most of the warnings/build errors of upstream packages because it would make the diff bigger for your rebase of `leios-prototype`, which would automatically include bumps to the latest `cardano-crypto-class` .. what do you think?

ch1bo:
Rebased this on top of #2041 instead of the bumps of `cardano-crypto-class` (previous PR description).

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/GenCDDLs.hs
```diff
@@ -111,7 +115,7 @@ setupCDDLCEnv = do
   let local_paths =
         map
           (localDataDir F.</>)
-          ["ouroboros-consensus-cardano/cddl"] -- Directories with other cddls that we import should go here
+          ["cardano-blueprint/src/codecs"] -- Directories with other cddls that we import should go here
```
ch1bo, line 118: @jasagredo I have explored here how we could move the `base.cddl` to the `cardano-blueprint`. I'm not yet asking for this to happen, but it seemed quite okay and the preparation of the leios notify mini protocol `messages.cddl` with `base.cddl` via `cddlc` was working well within your framework here. WDYT?

jasagredo: 1. How do you get the cardano-blueprint directory? Is it a git submodule?

2. I find it worrying that the base.cddl which belonged only to consensus now seems to be the base.cddl of the `codecs` folder. I think network and others probably has their own separate base.

3. I am also worried that Leios stuff has made it into the blueprints, as my understanding is that Leios is still experimental.

4. I am also worried to move our CDDLs to a separate repository where we will have to go through some _process_ to update them, without knowing what that process would be.

I think it is much more reasonable to have the blueprints get the CDDLs from us at least until the process for improving them and the responsibilities of the blueprints are clear.

I also think these kind of changes would deserve their own PR.

ch1bo: > 1. How do you get the cardano-blueprint directory? Is it a git submodule?

Yes, it already exists on main as a submodule.

> 2. I find it worrying that the base.cddl which belonged only to consensus now seems to be the base.cddl of the `codecs` folder. I think network and others probably has their own separate base.

You had put the base.cddl originally into the blueprint there. What's the problem with this?

> 3. I am also worried that Leios stuff has made it into the blueprints, as my understanding is that Leios is still experimental.

Yes, the pieces consistent with CIP-164 are reflected on the blueprint and marked as proposed. The commit here is just a branch, much like the prototype branch from this PR.

> 4. I am also worried to move our CDDLs to a separate repository where we will have to go through some _process_ to update them, without knowing what that process would be.

There is not much process. We can make the blueprint exactly the way we want it. What do you want to know / worries you?

> I think it is much more reasonable to have the blueprints get the CDDLs from us at least until the process for improving them and the responsibilities of the blueprints are clear.

Who is "us"? I, as a proposer and implementor of Leios, would like to see definitions of Leios protocol be easily sharable and directly used in all Leios-enabled Cardano software. It sitting inside the repository of one implementation is not ideal for that purpose.
 
> I also think these kind of changes would deserve their own PR.

Note that I'm not intending to merge this to `main` - the `leios-prototype` never will, but only small & focused PRs. This is just an exploration within the prototype to also find out what kind of integration would be possible - and to get feedback and early discussions like this one :slightly_smiling_face:

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -929,17 +935,17 @@ protocolInfoCardano paramsCardano
...
-                PraosCredentialsUnsound _ sk -> sk
-                PraosCredentialsAgent{} -> error "can't derive topLevelConfigVotingKey from agent"
-          Just
-            . rawSerialiseUnsoundPureSignKeyKES
-            . assumeUnsound
-            . praosCanBeLeaderCredentialsSource
-            . shelleyLeaderCredentialsCanBeLeader
+      , topLevelConfigCheckpoints =
+          cardanoCheckpoints
+      , -- FIXME: REMOVE THIS. Interprets cold key hashes as signing keys
```
perturbing: Very nice ;)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -113,6 +132,13 @@ instance ShowProxy LeiosPoint where showProxy _ = "LeiosPoint"
 instance Show LeiosPoint where
   show = prettyLeiosPoint
 
+instance SignableRepresentation LeiosPoint where
+  getSignableRepresentation point =
+    toStrictByteString $
+      -- REVIEW: Flat concatenation expected as what is signed?
```
perturbing, line 140: This is what I did in the wFA^LS demo.

What is the impact of this decision? Another option is some CBOR structure, e.g., a list of the elements?

Note that with the BLS signature scheme, we can sign arbitrary-sized data, as it hashes the data to the curve. So I think either way is fine.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
```diff
@@ -13,33 +15,53 @@ import Control.Concurrent.Class.MonadSTM.Strict
...
       { addVote = \vote -> atomically $ do
           seen <- readTVar seenVotes
           if Set.member vote seen
-            then pure ()
+            then pure AlreadyKnown
             else do
-              writeTVar seenVotes $! Set.insert vote seen
-              writeTChan votesChan vote
+              -- TODO: disallow votes from different epoch (than the committee is).
+              -- Could use slot numbers or put epoch into votes to distinguish?
```
perturbing, line 55: A vote already contains a slot number (or at least the `VoteId`/`election_id`), which maps onto an epoch, so adding an epoch to a vote is not needed.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: Yeah, I just kept the note because it's not concretely specified on the CIP. Thanks for confirming the flat concat is just fine.

PR #2040, opened 2026-05-13
Leios: late-join support

dnadales:
## Summary

Closes: https://github.com/input-output-hk/ouroboros-leios/issues/890

- Add `prop_leios_late_join` test: 4 nodes, node 3 joins at a random slot.
  Demonstrates the `resolveLeiosBlock` crash when a late node encounters a CertRB referencing an EB it never saw.
- Add `hbMayCertifiedEb :: Maybe LeiosPoint` to the Praos `HeaderBody` so CertRBs carry the certified EB point in the header (length-switching CBOR 10/11/12).
- Filter pending CertRBs from ChainSel: CertRBs whose EB closure hasn't arrived are recorded in `cdbPendingEBs` and made invisible to chain selection (both successor enumeration and predecessor tracing via `lookupBlockInfo`).

Late-join test passes 100 runs (no crash).
Still WIP — CertRBs are permanently excluded on the late node (no re-trigger yet).

Remaining steps:
- Chain-consistency assertion (all nodes converge to same chain)
- Re-trigger ChainSel when EB closure arrives
- Fetch mechanism for missing EB closures

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -56,7 +56,7 @@ import Data.Functor ((<&>))
 import Data.Hashable (Hashable)
 import Data.List.NonEmpty (NonEmpty)
 import qualified Data.List.NonEmpty as NE
-import Data.Maybe (fromMaybe, isJust, mapMaybe)
+import Data.Maybe (isJust, mapMaybe)
 import Data.Proxy
 import qualified Data.Text as Text
 import Data.Void (Void)
@@ -434,24 +434,10 @@ initNodeKernel
             when (new == old) retry
             writeTVar lastAppliedPendingEBs new
             pure (Map.difference new old, Map.difference old new)
-          MVar.modifyMVar_ getLeiosOutstanding $ \outstanding -> do
-            let bodies0 = Leios.missingEbBodies outstanding
-                -- Add entries from new pending CertRBs; never overwrite an
-                -- existing non-zero (offer-supplied) size.
-                bodies1 =
-                  Map.foldlWithKey'
-                    (\acc point _ -> Map.alter (Just . fromMaybe 0) point acc)
-                    bodies0
-                    added
-                -- Drop entries we no longer need; keep offer-driven entries
-                -- (non-zero size) untouched.
-                bodies2 =
-                  Map.foldlWithKey'
-                    (\acc point _ ->
-                      Map.update (\sz -> if sz == 0 then Nothing else Just sz) point acc)
-                    bodies1
-                    removed
-            pure outstanding{Leios.missingEbBodies = bodies2}
+          MVar.modifyMVar_ getLeiosOutstanding $
+            pure
+              . Leios.applyPendingRemoved (Map.keys removed)
+              . Leios.applyPendingAdded (Map.keys added)
           void $ MVar.tryPutMVar getLeiosReady ()
 
     void $
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -471,6 +471,13 @@ initNodeKernel
           -- ChainSel's "is the closure present?" check and its insert into
           -- cdbPendingEBs. The sweep covers that race and any other
           -- missed-notification scenarios (e.g. subscription startup gap).
+          --
+          -- An inline recheck in ChainSel.hs (the add-block path that
+          -- inserts into cdbPendingEBs) closes the immediate race window
+          -- so callers don't have to wait for the next sweep tick; this
+          -- sweep is the load-bearing fix and also handles cases the
+          -- inline recheck can't see (e.g. closure arrival just after a
+          -- prior pending-insert).
           pending <- atomically $ ChainDB.getPendingCertRBs chainDB
           forM_ (Map.toList pending) $ \(leiosPoint, certRBHash) -> do
             mayComplete <- LeiosDb.leiosDbQueryCompletedEbByPoint leiosConn leiosPoint
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -464,6 +464,18 @@ initNodeKernel
           traceWith tracer $ MkTraceLeiosKernel "leiosFetchLogic: wait for leios ready"
           () <- MVar.takeMVar getLeiosReady
           iterationStart <- getMonotonicTime
+          -- Sweep cdbPendingEBs for CertRBs whose EB closure has since
+          -- become available locally, and re-enqueue them for ChainSel.
+          -- ebCompletionRunner already does this when a notification fires,
+          -- but it can miss the window if the closure completes between
+          -- ChainSel's "is the closure present?" check and its insert into
+          -- cdbPendingEBs. The sweep covers that race and any other
+          -- missed-notification scenarios (e.g. subscription startup gap).
+          pending <- atomically $ ChainDB.getPendingCertRBs chainDB
+          forM_ (Map.toList pending) $ \(leiosPoint, certRBHash) -> do
+            mayComplete <- LeiosDb.leiosDbQueryCompletedEbByPoint leiosConn leiosPoint
+            when (isJust mayComplete) $
+              void $ ChainDB.addReprocessBlock chainDB certRBHash
           leiosPeersVars <- MVar.readMVar getLeiosPeersVars
           offerings <- mapM (MVar.readMVar . Leios.offerings) leiosPeersVars
           -- Per-peer certified EBs derived from ChainSync candidate fragments,
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
@@ -469,29 +469,44 @@ initNodeKernel
               void $ ChainDB.addReprocessBlock chainDB leiosPoint certRBHash
           leiosPeersVars <- MVar.readMVar getLeiosPeersVars
           offerings <- mapM (MVar.readMVar . Leios.offerings) leiosPeersVars
-          -- Per-peer certified EBs derived from ChainSync candidate fragments,
-          -- used as a fallback peer source for fetching EB closures that no
-          -- peer has explicitly offered.
+          -- Per-peer set of EB hashes the peer has certified on its
+          -- selected chain. Used as a fallback peer source in
+          -- 'choosePeerEb' / 'choosePeerTx' when no peer has actively
+          -- offered the EB body or closure.
           --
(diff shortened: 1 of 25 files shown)


ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -99,6 +100,10 @@ data HeaderBody crypto = HeaderBody
   -- ^ protocol version
   , hbMayEbAnnouncement :: Maybe EbAnnouncement
   -- ^ Leios EB announcement
+  , hbMayCertifiedEb :: Maybe LeiosPoint
```
ch1bo, line 103: What is this used for (right now)? In the CIP we speculatively put it (as a single bit) as it may help synchronizing nodes to know the size of what they request.

dnadales: Currently it tells `ChainSel` "this ranking block certifies EB X" and this is what we use to check whether the EB closure is available and, if not, defer chain selection until it arrives (concretely we use the `LeiosPoint` to look in the `LeiosDB` if the EB closure is available).

The [ranking-blocks](https://github.com/input-output-hk/ouroboros-leios/blob/3d7c38c88a502ee6987f382854d04080f3be8aef/docs/cddl/diffs/linear/ranking-blocks.md#L19) spec mentions the addition of a `hash32` but not a `(slotNo, EbHash)`, so I don't know if we want to change `hbMayCertifiedEb` accordingly.

ch1bo: I see it now, we need it to have a proof that the node that gave us the header actually has the EB. Of course, this means that we must also check whether there is a cert in the body if the header claims it.

bladyjoker: I'd propose to follow what I implemented but removed because it wasn't used...

https://github.com/IntersectMBO/ouroboros-consensus/pull/1978/changes/b7f6f6c1415310e89dd247552b92e4825867cec8

I introduced a notion of a `BodyType` which is imo better naming for what its purpose is (Header is associated with Bodies of many types) into the PraosHeader.

https://github.com/IntersectMBO/ouroboros-consensus/blob/b7f6f6c1415310e89dd247552b92e4825867cec8/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs#L98

https://github.com/IntersectMBO/ouroboros-consensus/blob/b7f6f6c1415310e89dd247552b92e4825867cec8/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs#L272


Combined with LeiosState that's part of PraosState

https://github.com/IntersectMBO/ouroboros-consensus/blob/b7f6f6c1415310e89dd247552b92e4825867cec8/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L292

https://github.com/IntersectMBO/ouroboros-consensus/blob/b7f6f6c1415310e89dd247552b92e4825867cec8/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L350-L353

One gets the answer to what's being certified...ie. just look into `leiosStatePreviousAnnouncement` and if the Header is associated with a LeiosCertificate BodyType.

bladyjoker (changes requested):

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
dnadales: It seems there's a contradiction between CIP and CDDL:

* CIP-0164 (prose + appendix CDDL): https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md
* Prose §"Ranking Blocks" — `certified_eb` as single bit:
  https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#L695-L697
* Appendix B CDDL — `? certified_eb : bool`:
  https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#L2903
* Linear ranking-blocks CDDL diff (`? certified_eb : hash32`):
  https://github.com/input-output-hk/ouroboros-leios/blob/main/docs/cddl/diffs/linear/ranking-blocks.md#L41

Before deciding on this we should agree which spec we should follow.

nfrisby: After a bit of git archeology, I see that Will wrote `ranking-blocks.md` as part of the process of writing the CIP. So I think that's a "showing your work" file. Perhaps we could just add a comment to its `certified_eb` field saying the CIP now uses a bit for this.

dnadales:
Action points after discussing this PR with @nfrisby (to be tackled in no particular order):

1. Keep the Leios fetch logic as it was. Assume that if a peer offers us a `CertRB` then we can consider this an implicit offer for the certified EB (both the body and the closure).
2. Try to get rid of `cdbPendingEBs`. We could keep a `announcementsMap :: hash RB -> hash EB` as part of the `VolDB` functionality. We still require in `ChainSel` the set of dowloaded EB closures, which the Leios fetch logic should take care of tracking in an in-memory data structure. This results in a better separation of concerns. We don't want to change the on disk format to avoid interfering with Mithril so the `announcementsMap` has to live in-memory and be reconstructed from scratch on node-startup.
    - NOTE: The other option is to **not** to filter the candidates based on EB closure availability and let block application fail dynamically (which might require changing the types involved in block application). In this path, we need to check EB closure availability before setting the tentative header (otherwise the assumption that (RB) header offer implies EB closure availability breaks).
    2.1 We need to re-trigger chain selection when an EB-closure is downloaded. We should do this on the announcing EB, because there's a 1-1 mapping between announcing RB and EB. Thus for a given  we'd only have to call selection once and chain selection would take care of sorting out the candidate chains that include the certifying RBs.
4. Change `hbMayCertifiedEb` to a single bit to comply with the CIP, and consider using the `PraosState` (`ChainDepState`) to get the certified EB.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
ch1bo: I dropped the superseded CDDL diffs to avoid further confusion https://github.com/input-output-hk/ouroboros-leios/commit/5f06c19db4ff798d8b0ff365ec40825503e1c181

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -298,6 +311,36 @@ emptyLeiosOutstanding =
     , blockingPerEb = Map.empty
     }
 
+-- | Insert pending-CertRB entries into 'missingEbBodies' with size 0,
+-- without clobbering any existing (offer-supplied) non-zero size. Size 0
+-- is the sentinel for "we don't know the body size because the entry was
+-- driven by ChainSel via 'cdbPendingEBs' rather than by an offer".
```
bladyjoker, line 317: EbAnnouncement contains a Size, and can be directly taken from the `praosState -> leiosState`

```
data EbAnnouncement = EbAnnouncement
  { ebAnnouncementHash :: EbHash
  , ebAnnouncementSize :: BytesSize
  }

data LeiosState = LeiosState
  { leiosStatePreviousAnnouncement :: Maybe EbAnnouncement
  , leiosStateCumulativeEbAnnouncementSize :: BytesSize
  }

data PraosState = PraosState
  { praosStateLastSlot :: !(WithOrigin SlotNo)
  , praosStateOCertCounters :: !(Map (KeyHash 'BlockIssuer) Word64)
  -- ^ Operation Certificate counters
  , praosStateEvolvingNonce :: !Nonce
  -- ^ Evolving nonce
  , praosStateCandidateNonce :: !Nonce
  -- ^ Candidate nonce
  , praosStateEpochNonce :: !Nonce
  -- ^ Epoch nonce
  , praosStateLabNonce :: !Nonce
  -- ^ Nonce constructed from the hash of the previous block
  , praosStateLastEpochBlockNonce :: !Nonce
  -- ^ Nonce corresponding to the LAB nonce of the last block of the previous
  -- epoch
  , praosStateLeios :: !LeiosState
  }
```

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -128,6 +130,17 @@ decodeLeiosPoint = do
...
+  encCBOR (MkLeiosPoint ebSlot ebHash) =
+    LedgerBinary.encodeListLen 2
+      <> encCBOR ebSlot
+      <> encCBOR ebHash
+
+instance DecCBOR LeiosPoint where
+  decCBOR = do
+    LedgerBinary.decodeListLenOf 2
+    MkLeiosPoint <$> decCBOR <*> decCBOR
+
```
bladyjoker, line 143: Above you have encodeLeiosPoint and decodeLeiosPoint. Make sure to keep the implementation singular and reuse.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -561,7 +589,11 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
   let MkLeiosPoint _ebSlot ebHash = point
   do
     let ebBytesSize' = leiosEbBytesSize eb
-    when (ebBytesSize' /= ebBytesSize) $ do
+    -- A 0 expected size signals a ChainSel-driven request (see
+    -- 'pendingEbReconciler'): we don't know the EB body size up front because
+    -- 'hbMayCertifiedEb' only carries the 'LeiosPoint'. The hash check below
+    -- is the authoritative integrity check.
+    when (ebBytesSize /= 0 && ebBytesSize' /= ebBytesSize) $ do
```
bladyjoker, line 596: See my comment about EbAnnouncement, contains both hash and size and is available in the PraosState

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -427,8 +448,65 @@ initNodeKernel
...
+              pure $
+                Set.fromList
+                  -- Singleton-list generator with a refutable pattern:
+                  -- each header that does not certify an EB is silently
+                  -- skipped.
+                  [ pointEbHash
+                  | hwt <- AF.toOldestFirst (csCandidate state)
+                  , Just MkLeiosPoint{pointEbHash} <-
+                      [LedgerDB.certifiedEbFromHeader (hwtHeader hwt)]
+                  ]
```
bladyjoker, line 500: Can we make this work entirely on state? We're threading through information via header that is only because of how we choose to implement things internally.

I'm sure that as we pick candidates we build up PraosState for each fragment/point, otherwise it isn't a valid candidate. I'd like us to find how to get access to the Praos state for each point and collect "certified"  EbAnnouncements  from there.

bladyjoker (commented):
I think I have a better understanding of what's going on here.

I have to reiterate on changing the approach to use the Praos State and avoid traversing fragments and cherry picking meaning from them that is already maintained in the Praos State (the LeiosState within).

I will take a look tomorrow to see at which point we have computed the Praos State for each point on the fragment and collect all Certified EBs that are not available in the LeiosDB

PR #2046, opened 2026-05-19
Updated ouroboros-network

coot:


cabal.project
jasagredo: You need to update this comment to signal where this SRP is pointing to

coot: The comment still stands, since the commit is after `main-peras-5202-merged` tagged commit :smile:

jasagredo: Right, even if it is still true, i would prefer if you update it to mention a more recent point in the history of the repo.

In any case, I am approving so it can be merged as is.

jasagredo (approved):

cabal.project
coot: I changed it to:
```
-- ouroboros-network dependency after introducing `bracketKeepAlive` (PR#5371)
-- and supporting `contra-tracer-0.2.1` (PR#5368).
```

PR #2049, opened 2026-05-25
LedgerHD: Per era storage and support for multiple tables

jasagredo:
# UTxO‑HD v3: Handle‑based ledger refactor

## Summary

This PR replaces the `mk :: MapKind` parameter on `LedgerState` with a handle vocabulary owned by the LedgerDB backend, redesigns the Mempool around per‑tx local data, narrows the ChainDB/Mempool `LedgerInterface` to fix a long‑standing race, splits the storage‑backend layer into per‑era pieces, and ports all consumers (HFC, Byron, Shelley, Cardano, Dual, diffusion, mini‑protocols, db‑analyser, db‑synthesizer, snapshot‑converter, every testlib + test‑suite + bench).

Scope: ~230 files, +8.3k / −12.5k. Net code reduction because the `Tables.*` / `IndexedMemPack` / V2 internals modules go away.

## Why

- **`mk :: MapKind` was a phantom across most of the codebase.** Only the on‑disk UTxO actually needs differential representations; threading `EmptyMK` / `ValuesMK` / `DiffMK` through every signature pushed complexity into ledgers that have no UTxO at all (Byron, mock, byron‑spec).
- **The `getCurrentLedgerState :: STM m (StateHandle …)` field invited use‑after‑close** between the mempool and ChainDB. Narrowing to a tip‑only STM read plus a bracketed duplicate eliminates the class of bug.
- **Storage backend coupling.** The LSM backend used to live inside the cardano sublib; targets that don't want lsm‑tree (e.g. wasm cross‑compile) couldn't avoid it. The umbrella+sublib split fixes that.

## Proposed review order

Each commit is shaped to stand on its own; layers below are dependency‑ordered. Within a layer, commits compile independently of later layers but reference earlier ones.

### Library — foundations and consumers

1. **`Foundations: handle-based ledger vocabulary`** — the load‑bearing change.
   Drops `mk` from `LedgerState`; introduces `LedgerTablesHandle`, `StateHandle`/`TickedStateHandle`, `ExtStateHandle`/`TickedExtStateHandle`, the `Handle l m blk` injective family, and `LedgerTablesFactory m blk`. Block application/ticking becomes `Handle l m blk -> m (Handle (Ticked l) m blk)`. Deletes `Tables.*` (except `Tables.Diff`) and `IndexedMemPack`. Most other modules touched here are mechanical mk‑strips on signatures.

2. **`Mempool: per-tx local data + accumulator redesign`** — replaces `MempoolCache` with two block‑indexed data families `TxLocalData blk` and `MempoolAcc blk`. `prepareTx` is the sole monadic entry; `applyTx`/`reapplyTx` become pure. HFC `applyHelper` no longer uses `undefined` / `error "Impossible!"` / `unsafeCoerce`. Shelley `txMeasureConway` now sees the real per‑tx UTxO entries.

3. **`HFC: handle-based, no mk parameter`** — per‑era `StateHandle`s carried inside `HardForkStateHandle`; `HFLedgerTablesFactory` declared on `CanHardFork`; ticking/translation own a single in‑flight handle end‑to‑end. `HeaderStateHistory.fromChain` no longer leaks intermediate handles.

4. **`LedgerDB: Handle-centric API`** — new `openHandleAtTarget` is the core read primitive; `openReadOnlyHandle` is a thin wrapper (no Forker TVars/tracer/lock allocated for read‑only access, no stray `ForkerOpen` trace). `Forker` is now used only by chain selection. Deletes standalone `V2/Forker.hs` and `V2/InMemory.hs`; their content folds back into `V2/Backend.hs` and `V2/LedgerSeq.hs`. Mempool race fix lives here too.

5. **`ChainDB: adapt to handles; narrow LedgerInterface`** — new `getCurrentLedgerRef` field on the `ChainDB` record. `LedgerInterface` splits into `getCurrentLedgerTip :: STM m (Point blk)` (caller can't deref) and `withCurrentLedgerStateDup :: (StateHandle m blk -> m a) -> m a` (bracketed duplicate under the LedgerDB read lock). Everything else is mechanical.

6. **`Mini-protocols and node init: adapt to handles`** — `ProtocolInfo b` → `ProtocolInfo m b`, `pInfoInitLedger` becomes a `LedgerTablesFactory m b -> m (ExtStateHandle m b)` continuation matching `lgrGenesis`. Drops `CanUpgradeLedgerTables` superclass on `RunNode`.

7. **`Diffusion: adapt to handles`** — mechanical: `EmptyMK` strip, `ProtocolInfo m b`, `withReadOnlyForkerAtPoint` → `withReadOnlyHandleAtPoint`, mempool‑snapshot callsite drops `roforkerReadTables`.

8. **`Byron: adapt ledger to handles`** — Byron has no on‑disk tables; trivial `LedgerTablesFactory = ()` and no‑op `LedgerTablesHandle`. `ByronHFC` adapts to the new HFC plumbing.

9. **`Shelley: adapt ledger / queries / TPraos / HFC to handles`** — UTxO behind a `LedgerTablesHandle`; in‑memory part is the ticked Shelley state minus UTxO. Adds missing `NoThunks (Ticked LedgerState …)` deriving needed by `MempoolAcc`.

10. **`Cardano: rewrite CanHardFork; drop obsolete Ledger and QueryHF`** — `Cardano.Ledger` and `Cardano.QueryHF` are deleted outright (their role was wiring `mk`‑aware injection through translation; with handles owning tables and HFC queries living in `HardFork.Combinator.Ledger.Query`, neither file does anything). `CanHardFork` rebuilt on `HFLedgerTablesFactory` + the new translation API.

11. **`Dual ledger: adapt to handles`** — paired `StateHandle`/`TickedStateHandle` holding a handle from each side; ticking and block application route through both.

12. **`Backends: split into umbrella + InMemory + LSM-as-its-own-sublibrary`** — `Ouroboros.Consensus.Backends` is now a tiny umbrella exposing `inMemoryBackendArgs` plus shared `loadSnapshot`/`mkSnapshotManager` helpers (both take a `SnapshotBackend` tag so the per‑backend snapshot path uses the right marker). `Backends.InMemory` lives in the cardano sublib (depends on Shelley UTxO). The `ouroboros-consensus:lsm` sublib is repurposed to expose just `Backends.LSM` with a real cursor‑based UTxO reader; the previously fully‑commented 853‑line `V2.LSM` is deleted. Snapshot writer/reader path mismatch (`utxo` vs `tables`) is unified.

13. **`Snapshot streaming: move to a per-era API`** — the block‑level `StreamingBackend` was carrying `((), ())` placeholders. Replaced with per‑era `SnapshotYielder`/`SnapshotSinker` value‑level records (new `Shelley.Ledger.SnapshotStream`); each backend exports its own constructor. CBOR over `<snap>/utxo` for in‑memory; BlockIO + cursor‑paged yield for LSM, with a `utxoSize` sidecar.

### Tests — layered to track the dep graph

14. **T1 — `Consensus testlib: adapt to handles and no-MK ledger`** — foundation. `Test.Util.TestBlock` + `Test.Ouroboros.Storage.TestBlock` ported with trivial `LedgerTablesHandle = ()` / `StateHandle = newtype (LedgerState blk)`. `Test.Util.ChainDB.MinimalChainDbArgs` grows `mcdbBackendArgs` and takes `mcdbInitLedger` as a continuation. Deletes `Test.LedgerTables` (MK‑laws only) and `Test.Util.LedgerStateOnlyTables` (no users).

15. **T2 — `Single-hop testlibs: adapt to handles and no-MK ledger`** — byronspec / mock‑block / mempool‑test‑utils / diffusion‑testlib / tutorials. Mock keeps a UTxO (the field was dropped from `LedgerState` but lives inside `MockState`). `ThreadNetworkArgs` grows `tnaLedgerTablesFactory`. `TestConfigMB` exposes the same. EBB‑forging path in `customForgeBlock` is stubbed with `error` at this point — restored in step 22 below.

16. **T3 — `Era testlibs: adapt to handles, no-MK ledger, and new HFC translation`** — mock / byron / shelley / cardano testlibs. `Test.ThreadNet.Infra.ShelleyBasedHardFork` is the headline rewrite (mirrors mainline `Cardano.CanHardFork`). Drops `BlockSupportsHFLedgerQuery`/`HasCanonicalTxIn`/`HasHardForkTxOut`/`SerializeTablesWithHint`/`IndexedMemPack` instances (~150 lines). `Test.Consensus.Cardano.Translation` deleted; `Test.ThreadNet.TxGen.Cardano.migrateUTxO` reads the unticked current era (the chain‑tick path is no longer available pure‑side).

17. **T4 — `Tools / snapshot-conversion: adapt to handles and no-MK ledger`** — db‑analyser, db‑synthesizer, snapshot‑converter. `Forking` and `Analysis` flip `withReadOnlyForkerAtPoint` → `…Handle…`. `checkNoThunksEvery` and `traceLedgerProcessing` keep `tickThenApply` (not `tickThenReapply`) — both want to surface validation failures, not skip them.

18. **T5 (partial) — `Test-suites T5 (partial): per-era + benches + checkNoThunksEvery fix`** — per‑era test‑suites + `ChainSync-client-bench` + the analyser semantics note.

19. **`ThreadNet testlib: TestConfigMB.nodeInfo monadic; finish shelley-test`** + **`ThreadNet testlib: TestConfigMB.ledgerTablesFactory monadic`** — HFC tests need an `MkHandle m` that requires a sim‑fs allocation. Pure tests still pass `pure ()`.

20. **`cardano-test`**, **`consensus-diffusion-test`**, **`consensus-test`** ports — three commits, one per test‑suite. `consensus-test` adapts mempool tests to the per‑tx workflow and new `LedgerInterface`; the QSM model (`Test.Consensus.Mempool.StateMachine`) is dropped here and restored in step 28 below.

21. **`ThreadNet testlib: supply mcdbBackendArgs for the in-tree LedgerDB`** + **`consensus-{diffusion-,}test: plumb mcdbBackendArgs through remaining sites`** — strict field on `MinimalChainDbArgs`; missing it builds (`-Wmissing-fields`) but crashes at runtime on the first LedgerDB op. New `Test.Util.ChainDB.testBackendArgs` helper. Bit us twice.

22. **`ThreadNet testlib: port the EBB-forging path in customForgeBlock`** — restore the EBB JIT‑forge path stubbed in step 15. Acquires a read‑only `ExtStateHandle` at the volatile tip via `ChainDB.withReadOnlyHandleAtPoint`, ticks/applies/re‑ticks. Six byron‑test failures + one shelley‑test variant go green.

23. **`byron-testlib Examples: reapply (ValidateNone), not full validation`** — the synthetic example block references a UTxO not in the empty initial state. Pre‑T3 used `reapplyLedgerBlock`; restoring that semantics fixes a golden‑construction crash.

24. **`mempool-bench: port TestBlock to no-MK + per-tx mempool API`** — UTxO pulled directly into `PayloadDependentState`; mempool routed through the per‑tx workflow.

25. **storage‑test cluster (5 commits)** — `Model/Unit/StateMachine` ported via a new `PureExtApplyBlock` helper class that simulates the monadic `tickThenApply` via `runSim` for trivial‑handle blocks. `LedgerDB V2: emit TookSnapshot` revives `ChainDB.LedgerSnapshots`. `LedgerDB.StateMachine` rebuilt around the new backend abstraction (no V1/V2 split, no LSM `Salt`); a new test backend serialises via `Serialise`. Leak‑check accepts `0` for trivial‑tables backends (no real handles ever allocated).

26. **`ShelleyBasedHardFork: bump era-1 maxMajorProtVer to era-2's`** — the two‑era HFC test stack called `protocolInfoTPraosShelleyBased` once per era with each era's own `protVer`, giving era‑1 a `maxMajorProtVer` of `protVer1`. Once the era‑1 PPUPDATE that targets `protVer2` is confirmed, the chain's current protocol version becomes `protVer2` *before* the HFC trigger fires at the next epoch boundary; era‑1 successor blocks in that window then trip Shelley's CHAIN rule with `ObsoleteNodeCHAIN (Version 2) (Version 1)`. Mainline `protocolInfoCardano` sidesteps this by sharing the FINAL era's `maxMajorProtVer` across every era; this commit does the same by patching era‑1's `tpraosMaxMajorPV` after `protocolInfoTPraosShelleyBased` returns.

27. **`Backends: strip UTxOs from the snapshot`** — corollary cleanup once snapshot streaming is per‑era. Snapshot writer/reader no longer carry the per‑backend UTxO blob alongside the per‑era stream.

28. **`Consensus-test: re-enable mempool QSM`** — restores `Test.Consensus.Mempool.StateMachine` (907 lines re‑added) after the model was deferred in step 20.

### Post-port polish

29. **`Formatting`** — fourmolu pass over the touched files.

30. **`Cleanup warnings`** — drops unused imports surfaced by the new module shapes (`ShelleyBasedHardFork.hs` sheds ~25 imports that were only there for the deleted `BlockSupportsHFLedgerQuery` / `HasCanonicalTxIn` / `HasHardForkTxOut` / `SerializeTablesWithHint` / `IndexedMemPack` instances), removes `Ouroboros.Consensus.Util` helpers that are no longer referenced, and tidies dead constraints in `ouroboros-consensus.cabal`.

31. **`Cleanup redundant constraints`** — bundles `NoThunks [Ticked]StateHandle` onto the `BlockSupportsLedgerHD` class so downstream signatures don't have to repeat them; drops a long tail of constraints made redundant by the new class header. Also nudges a couple of testlib instances to match.

## Test status

- `cabal build all` is clean.
- `cabal test all` is green: `mempool-bench`, `mock-test` (38/38), `byron-test` (98/98), `shelley-test`, `consensus-test`, `consensus-diffusion-test`, `cardano-test`, `storage-test` (62/62), `consensus-infra-test`, `protocol-test`, `diffusion-infra-test`, `tools-test`.

## Lost coverage (catalogued)

Coverage removed because the API it exercised no longer exists. Most has equivalent coverage elsewhere; flagged here so reviewers can decide:

- **MK‑law property tests** (`prop_hasledgertables_laws`, `prop_stowable_laws`) — the underlying classes are gone; no analogue to restore. Diff laws still live in `Test.Consensus.Ledger.Tables.Diff`.
- **`LedgerTables` golden CBOR tests** — no standalone `LedgerTables` data type; the on‑disk tables format is now per‑era inside the snapshot writer with its own `SnapshotMetadata` versioning.
- **`MempoolLedgerDBView` range‑query path** — the new `LedgerInterface` has no range query; the mocked mempool no longer simulates it.
- **HFC `SerializeTablesWithHint`/`IndexedMemPack`/`HasCanonicalTxIn`/`HasHardForkTxOut`/`BlockSupportsHFLedgerQuery` instances** (two‑era ShelleyBased test harness) — the classes are removed from the library; per‑era snapshot tests cover the equivalent path.
- **Shelley `testGenTxs` body** — was already gated behind `if True then pure []` (#2680); collapsed to a flat `pure []`. Restoring needs a port to the per‑tx workflow.
- **Cardano `migrateUTxO` chain‑tick** — now reads the unticked current era. Migration txs are generated one slot later than before; functionally still correct.

## Open follow‑ups before merge

None blocking. See "Lost coverage (catalogued)" above for behavioural deltas that are deliberate and need a reviewer's eye.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -289,9 +284,13 @@ doAddTx mpEnv caller wti tx = do
       (TestingAddTx _, Left _) -> pure Nothing
       (TestingAddTx _, Right x) -> pure $ Just x
 
-pureTryAddTx ::
+-- | Try to add a transaction to the mempool. Note: despite the historical
+-- @pure@ prefix, this is no longer pure — it reads the ticked state
+-- handle from 'mpEnvTickedHandle' and threads it through validation.
```
geo2a, line 289:
```suggestion
-- | Try to add a transaction to the mempool. 
-- Reads the ticked state
-- handle from 'mpEnvTickedHandle' and threads it through validation.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -426,82 +421,74 @@ implRemoveTxsEvenIfValid ::
...
-            pureRemoveTxs
-              capacityOverride
-              cfg
-              (isSlotNo is)
-              (isLedgerState is `withLedgerTables` emptyLedgerTables)
-              tbs
-              (isLastTicketNo is)
-              toKeep
-              toRemove
+      ts <- readMVar tickedHandleMVar
```
geo2a, line 432: Can we give `ts` a more descriptive name? Perhaps:
```suggestion
      tickedLedgerState <- readMVar tickedHandleMVar
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -551,85 +539,79 @@ implSyncWithLedger projectResult mpEnv =
       --
       -- Just for performance reasons, we will avoid re-validating the mempool
       -- if the state didn't change.
-      withTMVarAnd istate (const $ getCurrentLedgerState ldgrInterface) $
-        \is (MempoolLedgerDBView ls meFrk) -> do
-          let (slot, ls') = tickLedgerState cfg $ ForgeInUnknownSlot ls
-          if pointHash (isTip is) == castHash (getTipHash ls) && isSlotNo is == slot
+      withTMVar istate $ \is ->
```
geo2a, line 542: Nice, I'm very happy that `implSyncWithLedger` is gone!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -551,85 +539,79 @@ implSyncWithLedger projectResult mpEnv =
...
+        withCurrentLedgerStateDup ldgrInterface $ \st0 -> do
+          (slot, ls') <- tickLedgerState cfg $ ForgeInUnknownSlot st0
+          if pointHash (isTip is) == getTipHash (tickedState ls')
```
geo2a, line 545: I think we may be checking the hash of a different ledger state. On main compared the mempools tip hash with `getTipHash ls` (unticked), but now we compare it with `getTipHash (tickedState ls')`, which is the ticket state. Is this intentional and, if so, why?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -551,85 +539,79 @@ implSyncWithLedger projectResult mpEnv =
...
-          capacityOverride
-          lcfg
-          slot
-          lstate
-          values
-          (isLastTicketNo istate)
-          (TxSeq.toList $ isTxs istate)
-      mTrace =
+  m (InternalState blk, Maybe (TraceEventMempool blk))
+doSyncWithLedger capacityOverride lcfg slot ts istate = do
```
geo2a, line 599:
```suggestion
doSyncWithLedger capacityOverride lcfg slot tickedStateHandle istate = do
```
again, I suggest a more descriptive name.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -60,48 +63,37 @@ import NoThunks.Class
...
-deriving instance
-  ( NoThunks (Validated (GenTx blk))
-  , NoThunks (TxIn blk)
-  , NoThunks (TxOut blk)
-  ) =>
-  NoThunks (ValidatedTxWithDiffs blk)
+-- | What we store in the mempool sequence /per/ transaction: the
+-- validated tx and its block-specific 'TxLocalData' (typically the
+-- UTxO entries the tx's inputs reference, materialised from disk).
+type TxRecord blk = (Validated (GenTx blk), TxLocalData blk)
```
geo2a, line 87: Why not use a record type?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
geo2a, line 497:
```suggestion
-- | See 'Ouroboros.Consensus.Mempool.API.testSyncWithLedger' and 'Ouroboros.Consensus.Mempool.Init.forkSyncStateOnTipPointChange'.
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -191,100 +96,135 @@ data ExceededRollback = ExceededRollback
...
+-- code acquire a 'Handle' directly — via 'openHandleAtTarget' (and the
+-- 'openReadOnlyHandle' wrapper around it) or 'getVolatileTipRef' — and
+-- never touch a 'Forker'. Those entry points call
+-- 'openStateHandleAtTarget' without allocating a 'Forker' at all. The
+-- 'Forker' abstraction is reserved for chain selection, the one caller
+-- that actually needs commit. If a future caller is added that holds a
+-- 'Forker' but should be read-only, consider reintroducing a separate
+-- @ReadOnlyForker@ type so the type system can enforce that
+-- 'forkerCommit' is unavailable on that path.
+data Forker m blk = Forker
```
geo2a, line 120: The haddock of the new `Forker` type needs a re-work. It should explain what forkers are for. Also, the record fields need individual haddocs too.

geo2a, line 120: What does `foe` stand for? Perhaps use the `forker` prefix?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Snapshots.hs
```diff
@@ -269,7 +270,7 @@ data SnapshotManager m blk st = SnapshotManager
       st ->
       -- \^ The state needed for taking the snapshot:
       -- - In V1: this will be the DbChangelog and the Backing store
```
geo2a, line 274:
```suggestion
```
V1 does not exist anymore, right?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -228,27 +233,27 @@ data ChainDB m blk = ChainDB
   , getHeaderStateHistory :: STM m (HeaderStateHistory blk)
   -- ^ Get a 'HeaderStateHistory' populated with the 'HeaderState's of the
   -- last @k@ blocks of the current chain.
-  , allocInRegistryReadOnlyForkerAtPoint ::
+  , allocInRegistryReadOnlyHandleAtPoint ::
       Target (Point blk) ->
       ResourceRegistry m ->
-      m (Either GetForkerError (ResourceKey m, ReadOnlyForker' m blk))
+      m (Either GetForkerError (ResourceKey m, Handle ExtLedgerState m blk))
   -- ^ Allocate a read only forker at the given point in the given resource
```
geo2a, line 240: The haddocks on this record need updating, they mention forkers, but the actual function return handles. For this one specifically, I think read-only forkers do not even exist anymore.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -539,48 +373,194 @@ data instance Ticked LedgerState (ShelleyBlock proto era) mk = TickedShelleyLedg
...
 untickedShelleyLedgerTipPoint = shelleyTipToPoint . untickedShelleyLedgerTip
 
 type instance AuxLedgerEvent (ShelleyBlock proto era) = ShelleyLedgerEvent era
 
+type instance LedgerTablesHandle m (ShelleyBlock proto era) = TablesHandle m era
+
+slUtxoL :: Lens' (SL.NewEpochState era) (SL.UTxO era)
+slUtxoL = SL.nesEsL . SL.esLStateL . SL.lsUTxOStateL . SL.utxoL
+
+data MkHandle m = MkHandle
```
bladyjoker, line 396: Could I beg you to move everything storage related from this point on to Storage module? It's already too big and module structure helps us navigate what's what

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -177,27 +171,59 @@ fixedBlockBodyOverhead = 1024
...
+
+-- | Mempool accumulator for Shelley: the chain-ticked ledger state
+-- evolved by prior mempool tx applies (carrying their non-UTxO
+-- changes — governance, fees, …), plus the combined UTxO 'Diff' of
+-- every mempool tx applied so far.
+--
+-- The UTxO inside @saState@ is kept empty: the live UTxO view is
+-- reconstructed at validation time by applying @saDiff@ to each tx's
+-- 'TxLocalData' for the tx's input keys (see 'applyShelleyTx').
+data instance MempoolAcc (ShelleyBlock proto era) = ShelleyMempoolAcc
```
bladyjoker, line 190: Why wouldn't this be called MempoolState?

bladyjoker (commented):
Drive by comments x)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
jasagredo: You mean a Storage directory under Shelley, or to Ouroboros.Consensus.Storage?

We can do the former, but the latter is essentially the opposite of what this PR is about.

lsm/Ouroboros/Consensus/Backends/LSM.hs
```diff
@@ -0,0 +1,703 @@
...
+            writeUTxOSizeFile fs (snapshotToUTxOSizeFilePath ds) $ fromIntegral utxoSize
+            pure (Nothing, UTxOHDLSMSnapshot)
+      , castHandle = \st' ->
+          encloseTimedWith (TraceLedgerTablesHandleDuplicate >$< tracer) (LSM.duplicate table)
+            >>= newLSMTablesHandle tracer shfs st' utxoSize
+      , injectValues = \nes ->
+          encloseTimedWith (TraceLedgerTablesHandlePush >$< tracer) $ do
+            let (SL.UTxO utxos, nes') = nes & slUtxoL <<.~ SL.UTxO Map.empty
+            table' <- LSM.duplicate table
+            mapM_ (go table') . chunks 1000 . Map.toList $ utxos
```
dnadales, line 303: This global constant is buried quite deep in the definitions. Perhaps we should give it a name and place it before this definition.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -281,48 +271,46 @@ initMempoolEnv ::
...
-          , mpEnvStateVar = isVar
-          , mpEnvAddTxsRemoteFifo = addTxRemoteFifo
-          , mpEnvAddTxsAllFifo = addTxAllFifo
-          , mpEnvTracer = tracer
-          , mpEnvCapacityOverride = capacityOverride
-          , mpEnvTimeoutConfig = mbTimeoutConfig
-          }
+  (slot, ts) <- withCurrentLedgerStateDup ledgerInterface $ \st ->
+    tickLedgerState cfg (ForgeInUnknownSlot st)
+  tsMVar <- newMVar ts
```
dnadales, line 276: Should the ticked handle we cache in `mpEnvTickedHandle` be a `duplicateTicked` of `ts`/`ls'` rather than the handle itself? I might be misreading the lifetimes, so let me walk through what I think I'm seeing.

`withCurrentLedgerStateDup` brackets its handle: it opens a fresh duplicate at `VolatileTip` and `closeExt`s it once the continuation returns. Inside the continuation, `tickLedgerState` ticks that handle, and Shelley's `applyChainTickLedgerResult` carries the *same* `TablesHandle` into the `TickedStateHandle` (it doesn't duplicate). So the `ts` we store here (and the `ls'` stored in `implSyncWithLedger`) seems to share the very table the bracket closes on the way out.

For the in-memory backend that's harmless: `closeHandle` is a no-op and `readTxOuts` reads the NES captured in the closure. But for a backend whose handles own real resources (LSM: `closeHandle = LSM.closeTable`, `readTxOuts = LSM.lookups`), the cached handle's table would already be closed by the time the first `tryAddTx`/`prepareTx` reads it, and the first `implSyncWithLedger` would `closeTicked` it a second time. That looks like the use-after-close class this PR set out to remove, coming back for any backend that isn't the no-op one.

Am I missing a duplicate somewhere on this path? If not, would something like `duplicateTicked` before we cache the handle work, so the mempool owns one with its own lifetime that the bracket's close can't reach into?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -670,60 +644,74 @@ applyHelper ::
...
+      pure $
+        f
+          globals
+          tickedShelleyLedgerState'
+          ( let b = shelleyBlockRaw blk
+                h' = mkHeaderView (SL.blockHeader b)
+             in SL.Block h' (SL.blockBody b)
+          )
+
+  (st'', h') <- lift $ duplWithDiffs h tickedShelleyLedgerState' st'
```
dnadales, line 689: Not closing `h` here looks right for the common case: on a normal tick `h` is the LedgerSeq's current handle (Shelley's `applyChainTickLedgerResult` aliases it into the ticked handle), so the seq still owns it and `prune` closes it later. But at an era boundary I think `h` is a different beast, and I wonder if it ends up leaked.

`extendToSlot`/`howExtend` translate the previous era's handle into a *fresh* one via `castHandle` (which on LSM is `LSM.duplicate`), and that fresh handle is what becomes this tick's `h`. `Past` keeps no handle, and nothing else seems to own that translated handle, so once `duplWithDiffs` supersedes it with `h'` (also a fresh duplicate) the translated one looks dropped still-open.

For the in-memory backend this is invisible (`closeHandle` is a no-op). For LSM it would be one `LSM.Table` leaked per era crossing: the bare `castHandle` wrappers (Allegra→Mary and friends) never close their result, and even the Shelley→Allegra wrapper, which carefully closes its cast *intermediate* via `applyDiff`, still hands its final translated handle through here. Bounded (a handful over a full sync), so not urgent. It is the same intermediate-handle hazard that `HeaderStateHistory.fromChain` guards against by closing each `prev`.

Am I right that the boundary-tick handle has no owner, or is something upstream closing it that I've missed? If it really is unowned, where would we want the close to live, in the extend/translation path once the block is applied, or here?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Update.hs
```diff
@@ -551,85 +539,79 @@ implSyncWithLedger projectResult mpEnv =
...
-                          tbs
-                          is
-                  whenJust mTrace (traceWith trcr)
-                  pure (Just (projectResult is'), is')
+              modifyMVar_
+                tickedHandleMVar
+                ( \tsOld -> do
+                    closeTicked tsOld
+                    pure ls'
+                )
```
dnadales, line 557: Should we close the old handle outside the `modifyMVar_`, so a throw from `closeTicked` can't leave the MVar holding a half-closed handle?

As written, if `closeTicked tsOld` throws partway through, `modifyMVar_` runs its `onException` and puts the original `tsOld` back, dropping `ls'`. A partial close is reachable: the Dual handle closes `mainTSH` then `auxTSH` in sequence, and LSM's `closeHandle` is a real `LSM.closeTable` that can fail. The MVar would then hold a partially-closed `tsOld` that the next `tryAddTx`/sync touches (read-after-close, or a double-close on the next swap), and `ls'` leaks.

On the in-memory backend this can't fire (`closeHandle` is `pure ()`), so it's latent rather than live, but it lands on the same resource-bearing backends as the lifetime question on the cache above.

Would something like putting `ls'` into the MVar first and closing the previous handle afterwards (or bracketing the close) work, so the MVar always holds a fully-open handle even if the close fails?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/LedgerDB/StateMachine.hs
```diff
@@ -726,4 +691,4 @@ checkNoLeakedHandles (Environment _ testInternals _ _ _ getNumOpenHandles _) = d
           <> ", but the traces indicate that there are now these handles "
           <> show actual
       )
-      (actual == expected)
+      (actual == expected || actual == NumOpenHandles 0)
```
dnadales, line 694: Does this leak check have teeth as written? Two things make me doubt it can catch a real handle leak.

The quantity it compares against is `1 + maxRollback`, i.e. the number of LedgerSeq slots, not the number of backend tables allocated. And the only backend the suite runs is the trivial `TestStateHandle` (`:177`), which emits no `Create`/`Close` traces, so `actual` is always `NumOpenHandles 0` and the `|| actual == NumOpenHandles 0` clause makes the assertion pass regardless. The comment just above is candid that the invariant is vacuous for this backend.

The flip side is that this is exactly the check that would catch a handle that outlives its owner, which is the kind of thing I was wondering about on the mempool cache and the era-boundary translation. Would it be worth a variant that runs against a resource-bearing backend (in-memory or LSM, where the lifecycle traces actually fire) and asserts `actual == expected` without the `== 0` escape, so the handle lifecycle gets real coverage?

lsm/Ouroboros/Consensus/Backends/LSM.hs
```diff
@@ -0,0 +1,703 @@
...
+          encloseTimedWith (TraceLedgerTablesHandleClose >$< tracer) (LSM.closeTable table)
+      , getStatsHandle = Statistics $ fromIntegral utxoSize
+      , takeHandleSnapshot = \ds ->
+          encloseTimedWith (lsmBackendTrace LSMSnap >$< tracer) $ do
+            LSM.saveSnapshot
+              (fromString (snapshotToDirName ds))
+              (LSM.SnapshotLabel $ Text.pack $ "UTxO table")
+              table
+            writeUTxOSizeFile fs (snapshotToUTxOSizeFilePath ds) $ fromIntegral utxoSize
+            pure (Nothing, UTxOHDLSMSnapshot)
```
dnadales, line 295: Should the snapshot's two artifacts be written so a failure can't leave a half-written snapshot behind? `LSM.saveSnapshot` and `writeUTxOSizeFile` run as two bare statements here (and again in `lsmSnapshotSinker`), and the `utxoSize` sidecar is mandatory on load: `mkLSMFromSnapshot` rejects its absence as `BackendCorruptedData`, and `msz` also seeds the running size threaded through `implApplyDiff`. So a crash between the two leaves an LSM table snapshot with no sidecar that won't load. The InMemory backend writes its single `utxo` file in one `withFile … MustBeNew`, so it has no such window.

The broken snapshot is at least rejected rather than silently loaded, so this isn't corruption. The part that nags me is cleanup: `deleteSnapshotIfTemporary` only touches the snapshots fs, not the LSM session path where this snapshot lives, and the retake dedup guard in `mkSnapshotManager` checks the snapshots fs too, so a retake at the same slot could hit the `MustBeNew` sidecar (or re-`saveSnapshot` the same name) again.

Would it be worth bracketing the pair so a failed take removes the LSM snapshot, or treating a missing sidecar on load as "discard and retake" rather than a hard corruption error?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Backends.hs
```diff
@@ -0,0 +1,257 @@
...
+                Telescope.tip $
+                  getHardForkState ns
+          extLedgerSt :: HardForkState (StateHandle m) (CardanoEras c)
+          extLedgerSt = hmap (snd . unComp) ns
+      let computedCRC = maybe checksumAsRead (crcOfConcat checksumAsRead) mbCrcTables
+      Monad.when (computedCRC /= snapshotChecksum snapshotMeta) $
+        throwError $
+          InitFailureRead $
+            ReadSnapshotDataCorruption
+      pure (ExtStateHandle (HardForkStateHandle extLedgerSt mkH) hs, pt)
```
dnadales, line 186: Should a corrupt LSM snapshot fall back to an older one the way a corrupt in-memory snapshot does? The checksum we record and verify here covers the pure state plus, for in-memory, the `utxo` bytes (`mbCrcTables = Just …`); for LSM `takeHandleSnapshot` returns `Nothing`, so this only covers the state file. The on-disk UTxO is then left to lsm-tree's own checksums, which `openTableFromSnapshot` raises as a thrown `SnapshotCorruptedError`.

`initFromSnapshot` is `runExceptT …`, which turns `throwError` into `Left` but does not catch a thrown exception, and `tryNewestFirst` only falls back on a `Left`. So an in-memory corruption surfaces as `ReadSnapshotDataCorruption` (a `Left`) and we try the next-oldest snapshot, whereas an LSM corruption looks like it escapes as an exception past that fallback.

I believe this predates the PR (main's LSM backend also returns `Nothing` here), so not necessarily something to change in this PR. But is the degraded-startup behaviour meant to differ between backends, or should the LSM read path also surface corruption as a `SnapshotFailure` so the same fallback kicks in?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Backends.hs
```diff
@@ -0,0 +1,257 @@
...
+                    :* sf
+                    :* sf
+                    :* Nil
+            )
+            (Telescope.tip $ getHardForkState $ hardForkStateHandlePerEra $ unExtStateHandle st)
+    writeSnapshotMetadata shfs ds $
+      SnapshotMetadata
+        { snapshotBackend = bknd
+        , snapshotChecksum = maybe crc1 (crcOfConcat crc1) crc2
+        , snapshotTablesCodecVersion = TablesCodecVersion1
```
dnadales, line 256: We record `snapshotTablesCodecVersion = TablesCodecVersion1` here, but I couldn't find anywhere on the read path that consults it: `loadSnapshot` / `fromSnapshot` pick the decoder purely from the era, never from this field. If it's meant as a forward-compat hook (so a future `TablesCodecVersion2` layout could be decoded differently), should the decode path branch on it, or at least reject a version it doesn't recognise? As written the field round-trips through the metadata but doesn't gate anything, so a future bump would still be fed to the v1 decoder.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -259,7 +244,10 @@ chainDBLedgerInterface chainDB =
 -- different operations.
 data MempoolEnv m blk = MempoolEnv
   { mpEnvLedger :: LedgerInterface m blk
-  , mpEnvForker :: StrictMVar m (ReadOnlyForker m LedgerState blk)
+  , mpEnvTickedHandle :: StrictMVar m (TickedStateHandle m blk)
```
dnadales, line 247: Might it be worth a note here that the safety of this cached handle leans entirely on the `istate` `TMVar`? Every reader (`tryAddTx` → `prepareTx`) and the swapper (`implSyncWithLedger`'s `modifyMVar_` + `closeTicked`) currently runs under `withTMVar istate`, so a `prepareTx` disk read can't race with this handle being closed and replaced. That holds today, but it's an implicit coupling between two separate primitives (this `StrictMVar` and the `istate` `TMVar`): a future reader that grabs `mpEnvTickedHandle` without also holding `istate` would get a use-after-`closeTicked`. A short comment on the field (or an assertion) stating that it must only be read/swapped while holding `istate` would make that contract explicit.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/API.hs
```diff
@@ -267,47 +252,63 @@ import System.FS.CRC
...
-      m (Either GetForkerError (Forker m l blk))
+      m (Either GetForkerError (Forker m blk))
   -- ^ Acquire a 'Forker' at the requested point. If a ledger state associated
   -- with the requested point does not exist in the LedgerDB, it will return a
   -- 'GetForkerError'.
   --
   -- Note this will allocate resources; see the "'Forker' management
   -- in the running node" comment above.
+  --
+  -- Use 'openHandleAtTarget' for read-only access — it avoids allocating a
```
dnadales, line 293: While we're near it: the Forker-management note higher up in this module (around lines 89-146) still describes the mempool path in terms of a `MempoolLedgerDBView` whose "Forker part is used directly only in `initMempoolEnv` … and `implSyncWithLedger`", closed via `modifyMVar_`. With this PR the mempool goes through `withCurrentLedgerStateDup` + `mpEnvTickedHandle`, and there's no `MempoolLedgerDBView`/Forker on that path anymore, so that paragraph reads stale. Could we refresh it to match the handle-based flow (this `openHandleAtTarget` being the read primitive)?

dnadales (commented):
Let's discuss this during a call. I post some questions that came out of the session review assisted by Claude.

PR #2056, opened 2026-06-02
Switch Windows cross compilation to use GHC 9.12

jasagredo:
GHC 9.6.7 can't compile Plutus >1.61. We use 9.12 only for the Hydra cross compilation.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -46,9 +46,6 @@ if os (windows)
 
 constraints:
   tasty <1.5.4,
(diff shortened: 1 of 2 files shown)


angerman (approved):
Yes, I think this is the way. I really hope @lehins and @luite find the compiler issue with ledger and ghc96+.

@zeme-wana, @zliu41 (windows with 9.12; rest with 9.6 fyi).

geo2a (approved):

PR #2058, opened 2026-06-03
Leios: hot fix CertRB staging area + emergency fetch + resolve CertRBs in immutable chain

### LINKED ISSUE #890: Prototype: CertRB staging area and catch-up
## Why

```
```

### Why closures go missing

## What

## Scope

Also:

Deferred:

## References

- Key prototype files:

ch1bo:
Fixes the crash in `resolveLeiosBlock` (*"FIXME(bladyjoker): … Announced EB is being certified by the current chain but it's not available?"*) that hits late-joining or briefly-disconnected nodes when a CertRB arrives before its certified EB closure has been streamed locally.

This follows the design idea of a "staging area" as outlined in https://github.com/input-output-hk/ouroboros-leios/issues/890

### Notes on the design

- **Cert carries `(LeiosPoint, BytesSize)` payload.** Forge encodes; receiver decodes. No predecessor lookup, no chain-dep state at the gate. Requires the paired `cardano-ledger` change (`LeiosCert` becomes `newtype LeiosCert { leiosCertPayload :: ByteString }`).
- **Gate at `addBlockAsync`** (`wrapChainDbViewForLeiosStaging` in `NodeKernel.hs`). When a CertRB arrives with a missing closure, the gate parks `(blk, size, peers-from-ChainSync-candidates)` in `LeiosStagingArea` instead of admitting to ChainSel. New `checkLeiosBlockResolvable` class method does the decode + LeiosDb lookup.
- **Gate blocks the BlockFetch client thread** until the staged entry releases (closure arrived and drain admitted the block via the unwrapped `addBlockAsync`). The previous "lie via `blockWrittenToDisk = pure True`" approach plus the `getIsFetched` widening did not suppress BlockFetch's redecision; the resulting refetch spin was the dominant retainer of iosim `SimTrace` state and made `prop_leios_late_join` OOM at `numSlots=200`. Cost: head-of-line blocking on this peer's BlockFetch pipeline for the closure-fetch duration; LeiosFetch runs on a separate channel so progress is not deadlocked. Caveat tied to open point #4 below.
- **Fetch loop synthesises augmented inputs each tick** from `stagedSnapshot` — adds synthetic `missingEbBodies` entries and per-peer `offerings` (both halves so `choosePeerEb` and `choosePeerTx` both see the staged peers). Synth is gated on `not (Map.null stagedNow)` and the body-add is short-circuited against `acquiredEbBodies` to avoid the per-tick re-add+filter round-trip. The per-tick `leiosDbQueryFetchWork` (whose own API doc declared it startup-only) is gone; tx-closure synth is dropped — once the body arrives, the natural `msgLeiosBlock` handler populates `missingEbTxs`. Restart recovery of partial tx closures (the case the per-tick DB query was working around) is deferred to a follow-up that primes `LeiosOutstanding` from the DB once at MVar creation.
- **Drain releases on `AcquiredEbTxs` only** (not `AcquiredEb`) — `leiosDbQueryCompletedEbByPoint` only returns `Just` once the full closure is in.

### Orthogonal fix included: EB resolution on immutable-DB replay

Commit fbaa872 is a separate, narrowly-scoped fix shipped here for convenience. The immutable-DB replay path (`replayStartingWith` → V1/V2 `reapplyBlock`) was bypassing `resolveLeiosBlock` entirely. On every restart, CertRBs in the immutable range were re-applied with their empty wire body, leaving the post-replay ledger state missing every EB-tx output. The first volatile block to spend one of those outputs failed with `BadInputsUTxO`. Fix: thread `LeiosDbConnection` into `reapplyBlock` (V1: new `reapplyThenPushLeios` wrapper; V2: in-place) and call `resolveLeiosBlock` before `tickThenReapply`. Verified on `proto-devnet` (k=5 restart) and on the Leios testnet (~530k immutable blocks replayed cleanly, no `BadInputsUTxO`). This subsumes the always-failing-restart half of open point #2 below; the closure-goes-missing edge case is unchanged.

### Validation

- [x] Threadnet test `prop_leios_late_join` (cherry-picked from `origin/leios-late-join-v2`) passes at `numSlots=200`. Existing `prop_leios` still passes.
- [x] iosim perf: `prop_leios_late_join` @ `numSlots=200`, 1 iter, `+RTS -M4G -s`: ~52 s wall, **134 MB peak residency**, 95.6 % productivity (was 119 s, OOM @ 4 GB, 49.6 % productivity before blocking the BlockFetch client). For comparison, `prop_leios` @ same scope: 49 s, 88 MB peak. Cardano simple convergence baseline (numSlots ≈266): 7 s, 278 MB.
- [x] Manually ran `ouroboros-leios/demo/proto-devnet`, killed and restarted one of the nodes: node 3 missed 3 EB certs while down → on restart logged 3 distinct `CertRBStaged` → Phase 2 fetched the EBs from peers → 3 `CertRBReleased` → all three nodes converged on the same tip.
- [x] Leios testnet catchup: caught up to current tip from genesis, then restarted; `OpenedDB` reported `immtip=535588`, `tip=544039` (~530k blocks replayed via the patched path, 0 `InvalidBlock` / `BadInputsUTxO`). Runnable setup at `ouroboros-leios/testnet/` https://github.com/input-output-hk/ouroboros-leios/pull/930

### Open points (not fixed by this PR)

1. **Gate is best-effort for fork blocks.** The wrapper looks up the parent on the current chain *and* the active ChainSync candidate fragments. That covers chain-extension (catch-up) and BlockFetch out-of-order delivery, but not a fork whose parent is unknown to all live candidates. In that case the lookup returns `Nothing` ⇒ we admit ⇒ if ChainSel later switches to that fork and the CertRB's closure isn't local, `resolveLeiosBlock` crashes as before. Proper fix in the long run: make ChainSel itself aware of missing-cert-closure state and resolve at adoption time. The staging area is a deliberate stop-gap; the more surgical ChainSel-level design subsumes it.
2. **InitChainSel bypasses the staging gate.** The orthogonal fix above makes immutable-replay apply CertRBs correctly when the LeiosDb closure is present (the common case on every restart). What's still uncovered: if the closure later goes missing (LeiosDb wipe, crash between ChainSel adopt and LeiosDb flush, file corruption), restart's InitChainSel hits the original error because the wrap sits on `addBlockAsync` and is skipped at startup. Same long-run fix subsumes this — gating in the apply path / via ChainSel.
3. **Fetch-loop bootstrap from DB.** The synth pipeline now relies entirely on the in-memory `LeiosOutstanding` plus the staging snapshot. The case that's lost is restart recovery when a previous run had received an EB body but not its full tx closure: the natural `msgLeiosBlock` handler won't re-fire on restart, so `missingEbTxs` stays empty for that EB and the closure never finishes. Follow-up: prime `LeiosOutstanding` from `LeiosDb.leiosDbQueryFetchWork` once at MVar creation in `initNodeKernel` — exactly the use the API doc reserves it for. Sketch in this branch's `leios-fetch-loop-sketch.md` (local).
4. **No GC of staged entries** whose peer-set has emptied (all peers disconnected before EB arrived). TODO in `LeiosStagingArea`. With the new BlockFetch-blocking semantics this also leaks parked client threads in that scenario — the GC pass needs to wake parked threads with a `FailedToAddBlock` verdict before evicting an entry.

### Traces

- `Consensus.LeiosKernel.TraceLeiosKernel` with `kind:"CertRBStaged"` (`blockPoint`, `ebHash`, `ebSlot`, `knownPeers`) and `kind:"CertRBReleased"` (`ebHash`, `ebSlot`).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -55,12 +55,13 @@ source-repository-package
     cardano-diffusion
     network-mux
 
--- Points to ouroboros-ledger/ch1bo/leios-staging-area
+
+-- Points to ouroboros-ledger/leios-prototype
 source-repository-package
   type: git
   location: https://github.com/IntersectMBO/cardano-ledger
-  tag: dff221b6915bd694f1b0a740d010b13e436bbe06
-  --sha256: sha256-QAk7lH9cKNxWxCMtwMGjvX2bW5tdT+cAdWkibLy6VOE=
+  tag: c47305fcf47bd77437b837d0dfb9cb4181bfbc77
+  --sha256: sha256-U9luV5wuSytDg4zbXZnyQo29Jn4fZ10GKI8Q/QTrYrU=
   subdir:
     libs/cardano-ledger-core
     eras/dijkstra/impl
--- a/cabal.project
+++ b/cabal.project
@@ -55,12 +55,12 @@ source-repository-package
     cardano-diffusion
     network-mux
 
--- Points to ouroboros-ledger/leios-prototype-remake
+-- Points to ouroboros-ledger/ch1bo/leios-staging-area
 source-repository-package
   type: git
   location: https://github.com/IntersectMBO/cardano-ledger
-  tag: c47305fcf47bd77437b837d0dfb9cb4181bfbc77
-  --sha256: 1db2xc2gs44g5035srqzgqkbv3a2yacmvnwchd1jnjrfkibnxnak
+  tag: dff221b6915bd694f1b0a740d010b13e436bbe06
+  --sha256: sha256-QAk7lH9cKNxWxCMtwMGjvX2bW5tdT+cAdWkibLy6VOE=
   subdir:
     libs/cardano-ledger-core
     eras/dijkstra/impl
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Block.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Block.hs
@@ -1541,10 +1541,9 @@ instance
       fmap BlockDijkstra <$> resolveLeiosBlockHdr db prevAnn dijkstraBlk
     _ -> pure Nothing
 
-  checkLeiosBlockResolvable db blk = case blk of
-    BlockDijkstra dijkstraBlk ->
-      checkLeiosBlockResolvable db dijkstraBlk
-    _ -> pure Nothing
+  blockHasLeiosCert blk = case blk of
+    BlockDijkstra dijkstraBlk -> blockHasLeiosCert dijkstraBlk
+    _ -> False
 
   headerLeiosAnnouncement hdr = case hdr of
     HeaderDijkstra dHdr -> headerLeiosAnnouncement dHdr
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Block.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Block.hs
@@ -1541,6 +1541,11 @@ instance
       fmap BlockDijkstra <$> resolveLeiosBlockHdr db prevAnn dijkstraBlk
     _ -> pure Nothing
 
+  checkLeiosBlockResolvable db blk = case blk of
+    BlockDijkstra dijkstraBlk ->
+      checkLeiosBlockResolvable db dijkstraBlk
+    _ -> pure Nothing
+
   headerLeiosAnnouncement hdr = case hdr of
     HeaderDijkstra dHdr -> headerLeiosAnnouncement dHdr
     _ -> Nothing
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
@@ -38,7 +38,7 @@ import LeiosDemoTypes
   , ForgedLeiosEb (..)
   , LeiosPoint (..)
   , TraceLeiosKernel (..)
-  , encodeLeiosCertPoint
+  , encodeLeiosCertInfo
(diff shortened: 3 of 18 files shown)


dermetfan:
:information_source: FYI This repeatedly ran [out of memory](https://devxci.monitoring.aws.iohkdev.io/explore?schemaVersion=1&panes=%7B%22h28%22%3A%7B%22datasource%22%3A%22loki%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Binstance%3D%7E%5C%22ci.*%5C%22%7D+%7C%3D+%5C%22Out+of+memory%5C%22%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22loki%22%7D%2C%22editorMode%22%3A%22code%22%2C%22direction%22%3A%22backward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-12h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1) in [CI](https://ci.iog.io/build/13170848/nixlog/1) recently.

ch1bo:
> ℹ️ FYI This repeatedly ran [out of memory](https://devxci.monitoring.aws.iohkdev.io/explore?schemaVersion=1&panes=%7B%22h28%22%3A%7B%22datasource%22%3A%22loki%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22%7Binstance%3D%7E%5C%22ci.*%5C%22%7D+%7C%3D+%5C%22Out+of+memory%5C%22%22%2C%22queryType%22%3A%22range%22%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22loki%22%7D%2C%22editorMode%22%3A%22code%22%2C%22direction%22%3A%22backward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-12h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1) in [CI](https://ci.iog.io/build/13170848/nixlog/1) recently.

Should be fixed now

ch1bo:
Merged this now to consolidate branches. A follow-up PR by @dnadales will likely address the short-comings of this way to fix it.

PR #2059, opened 2026-06-03
Enable searching the documentation website

geo2a:
Adds local, build-time search via @easyops-cn/docusaurus-search-local. The search index is generated as static assets at build time, so it needs no external service.

Search is only available in production builds (yarn build && yarn serve), not in yarn start.

Peek 2026-06-03 11-26


jasagredo (approved):

dnadales (approved):
Thanks!

PR #2064, opened 2026-06-10
LSM: Rework snapshot-converter into a command tree

jasagredo:
Rework the snapshot-converter into a command tree:

  - daemon:  watch a node's snapshot directory and convert each exported LSM
             snapshot into a Mem snapshot as it is produced
  - convert: one-shot conversion between an exported LSM snapshot and a Mem
             snapshot, in either direction
  - lsm export / lsm import: move snapshots out of / into an offline LSM
             database

Conversions operate only on standalone (exported) LSM snapshots and Mem
snapshots, never on a live database.

db-analyser surfaces the LSM snapshot export directory via a new 'mkLSMConfig'
method on 'HasProtocolInfo' (read from the node config's
'LedgerDB.LSMExportPath'); it keeps using a random salt, as it builds a
throwaway ledger database whose exported snapshots record their own salt.

### DIFF
--- a/ouroboros-consensus-cardano/app/snapshot-converter.hs
+++ b/ouroboros-consensus-cardano/app/snapshot-converter.hs
@@ -1,245 +1,338 @@
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
-#if __GLASGOW_HASKELL__ >= 912
-{-# LANGUAGE MultilineStrings #-}
-#endif
 
 module Main (main) where
 
 import Cardano.Crypto.Init (cryptoInit)
 import Cardano.Tools.DBAnalyser.HasAnalysis (mkProtocolInfo)
-import Control.Concurrent
+import Control.Concurrent (threadDelay)
+import Control.Exception (SomeException, displayException, try)
 import Control.Monad (forever, void)
-import Control.Monad.Except
-import DBAnalyser.Parsers
+import Control.Monad.Except (runExceptT)
+import DBAnalyser.Parsers (CardanoBlockArgs, parseCardanoArgs)
 import qualified Data.List as L
-#if __GLASGOW_HASKELL__ < 912
-import qualified Data.Text as T
-#endif
 import Main.Utf8
 import Options.Applicative
-import Options.Applicative.Help (Doc)
-#if __GLASGOW_HASKELL__ < 912
-import Options.Applicative.Help.Pretty (Pretty (pretty))
-#endif
 import Ouroboros.Consensus.Cardano.SnapshotConversion
 import Ouroboros.Consensus.Storage.LedgerDB.Snapshots
+import Ouroboros.Consensus.Storage.LedgerDB.V2.LSM
+  ( lsmDbExportSnapshot
+  , lsmDbImportSnapshot
+  )
 import System.Exit
 import System.FSNotify
-import System.FilePath
+import System.FilePath (splitDirectories, splitFileName, (</>))
 
-data Config
-  = -- | Run in daemon mode
-    DaemonConfig
-      -- | Where the input snapshot will be in
-      SnapshotsDirectoryWithFormat
-      -- | Where to put the converted snapshots
-      SnapshotsDirectory
-  | -- | Run in normal mode
-    NoDaemonConfig
-      -- | Where and in which format the input snapshot is in
-      Snapshot'
-      -- | Where and in which format the output snapshot must be in
-      Snapshot'
+{-------------------------------------------------------------------------------
+  Commands
+-------------------------------------------------------------------------------}
 
--- | Helper for parsing a directory that contains both the snapshots directory
--- and the particular snapshot.
-data Snapshot'
-  = StandaloneSnapshot' FilePath StandaloneFormat
-  | LSMSnapshot' FilePath LSMDatabaseFilePath
+-- | The various ways in which the tool can be invoked.
+--
+-- 'Daemon' and 'Convert' operate purely on /standalone (exported) LSM
+-- snapshots/ and Mem snapshots; they never touch a live LSM database, and so
+-- carry the 'CardanoBlockArgs' needed to decode ledger states.
+--
+-- 'LsmExport' and 'LsmImport' operate directly on a (offline) LSM database; no
+-- ledger decoding is involved, hence no 'CardanoBlockArgs'.
+data Command
+  = Daemon DaemonOpts CardanoBlockArgs
+  | Convert ConvertOpts CardanoBlockArgs
+  | LsmExport LsmDbOpts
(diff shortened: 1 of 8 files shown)


bladyjoker (commented):
Lots of comments on types for the CLI commands/opts. Should we rather push this information to the description of the options and commands as displayed to the CLI users?

ouroboros-consensus-cardano/app/snapshot-converter.hs
bladyjoker, line 1: Thinking that we don't need to document types since they seem to be documented in CLI commands/options.

geo2a (approved):
I've built the branch locally and explored the cli options, but I did not try running the new `snapshot-converter`. The code LGTM!

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
+data DaemonOpts = DaemonOpts
+  { daemonMonitorIn :: FilePath
+  -- ^ The directory holding the @state@/@meta@ files of the snapshots produced
+  -- by the node (watched for completed snapshots).
+  , daemonExportedPath :: FilePath
+  -- ^ The directory into which the node exports its LSM snapshots. The exported
+  -- snapshot for a snapshot named @N@ is expected at @<this>/N@.
+  , daemonOutputIn :: FilePath
+  -- ^ The directory into which to write the converted Mem snapshots.
+  }
```
bladyjoker: Why not be more descriptive in the names? 

```haskell
data DaemonOpts = DaemonOpts
  { daemonMonitorMetaDir :: FilePath
  -- ^ The directory holding the @state@/@meta@ files of the snapshots produced
  -- by the node (watched for completed snapshots).
  , daemonLsmSnapshotsExportDir :: FilePath
  -- ^ The directory into which the node exports its LSM snapshots. The exported
  -- snapshot for a snapshot named @N@ is expected at @<this>/N@.
  , daemonMemSnapshotDir :: FilePath
  -- ^ The directory into which to write the converted Mem snapshots.
  }
```

Additionally, using the suffix Path is less descriptive than saying File or Dir. In your naming scheme it's a bit inconsistent, you use `In` without  `Path` and but `daemonExportedPath`.

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
+  -- @state@/@meta@ files).
+  , convertImportFrom :: Maybe FilePath
+  -- ^ If set, the input is an exported LSM snapshot whose tables live at
+  -- @<this>/<input snapshot name>@; otherwise the input is a Mem snapshot.
+  , convertOut :: FilePath
+  -- ^ The output snapshot (a directory named after the slot).
+  , convertExportTo :: Maybe FilePath
+  -- ^ If set, the output is an exported LSM snapshot whose tables are written to
+  -- @<this>/<output snapshot name>@; otherwise the output is a Mem snapshot.
+  }
```
bladyjoker, line 73: Similar here

```haskell
data ConvertOpts = ConvertOpts
  { convertSnapshotInDir :: FilePath
  -- ^ The input snapshot (a directory named after the slot, holding at least the
  -- @state@/@meta@ files).
  , convertImportInDir :: Maybe FilePath
  -- ^ If set, the input is an exported LSM snapshot whose tables live at
  -- @<this>/<input snapshot name>@; otherwise the input is a Mem snapshot.
  , convertSnapshotOutDir :: FilePath
  -- ^ The output snapshot (a directory named after the slot).
  , convertExportOutDir :: Maybe FilePath
  -- ^ If set, the output is an exported LSM snapshot whose tables are written to
  -- @<this>/<output snapshot name>@; otherwise the output is a Mem snapshot.
  }
```

Also inconsistency, using `From` and `To` and `In` and `Out`

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
+
+-- | Options for the @lsm export@/@lsm import@ commands.
+data LsmDbOpts = LsmDbOpts
+  { lsmDb :: FilePath
```
bladyjoker: lsmDbDir

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
+  -- ^ The LSM database (session) directory.
+  , lsmDir :: FilePath
```
bladyjoker: lsmRootDir?

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
+  -- ^ The export-to (for @export@) or import-from (for @import@) root directory.
+  -- The exported snapshot named @N@ lives at @<this>/N@.
+  , lsmSnap :: String
```
bladyjoker: lsmSnapName?

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
 main = withStdTerminalHandles $ do
   cryptoInit
-  (conf, args) <- getCommandLineConfig
+  cmd <- execParser opts
+  case cmd of
+    Daemon o args -> runDaemon o args
+    Convert o args -> runConvert o args
+    LsmExport o ->
+      runLsmCmd (lsmSnap o) $
+        lsmDbExportSnapshot (lsmDb o) (lsmSnap o) (lsmDir o </> lsmSnap o)
```
bladyjoker: duplicating `lsmSnap o` call, consider handing the lsmSnap once and calculating the directory within lsmDbExportSnapshot and lsmDbImportSnapshot

ouroboros-consensus-cardano/app/snapshot-converter.hs
```diff
@@ -1,245 +1,340 @@
...
                 _ -> pure ()
+            _ -> pure ()
+        )
+    forever $ threadDelay 1000000
+
+-- | Validate the snapshot name and run an LSM database operation, reporting the
+-- result.
+runLsmCmd :: String -> IO () -> IO ()
+runLsmCmd name act =
+  case snapshotFromPath name of
```
bladyjoker, line 166: I see `snapshotFromPath` used many times and the errors are handled differently. 

I suggest you find a top level place to handle what essentially is argument parsing/validation and pass it around to the call that need it.

bladyjoker (approved):
Mostly nits, I would prefer to have more consistency in naming things and refactor such that same validations, like the one from `snapshotFromName`, is done once and shared across the call sites.

ouroboros-consensus-cardano/app/snapshot-converter.hs
jasagredo: Done

ouroboros-consensus-cardano/app/snapshot-converter.hs
jasagredo: Done

PR #2066, opened 2026-06-10
Introduce voting thread

### LINKED ISSUE #168: Implement and register a voting thread
The goal of this ticket is to implement and register a voting thread that keeps track of:

- [ ] When and whether to vote in the current round (both w.r.t. the voting rules, the voting committe)
- [ ] Votes for the current round
- [ ] Adding votes to the PerasVoteDB (depends on #156)

nbacquey:


adithyaov (commented):
LGTM

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -404,10 +414,14 @@ initNodeKernel
...
 
-    whenPerasEnabled $
+    whenPerasEnabled $ do
       void $
         forkLinkedWatcher registry "NodeKernel.perasCertInclusion" $
           perasCertInclusionController st
+      void $
+        forkLinkedWatcher registry "NodeKernel.voteMinting" $
+          knownSlotWatcher btime $
+            \currentSlot -> withEarlyExit_ $ voteMintingController systemTime st currentSlot
```
adithyaov: Should we eta-reduce?
Explicit usage helps in ease of reading though.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -95,6 +96,7 @@ data Tracers' remotePeer localPeer blk f = Tracers
       f (TraceLabelPeer remotePeer (TracePerasVoteDiffusionOutbound blk))
   , perasCertInclusionTracer ::
       f (TracePerasCertInclusionEvent blk)
+  , perasVotingLogicTracer :: f (TracePerasVotingEvent blk)
```
tbagrel1: Note sure about the naming; I would use `voteForgingThreadTracer` or `forgeVoteThreadTracer`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -404,10 +414,14 @@ initNodeKernel
           txChannelsVar
           sharedTxStateVar
 
-    whenPerasEnabled $
+    whenPerasEnabled $ do
       void $
         forkLinkedWatcher registry "NodeKernel.perasCertInclusion" $
           perasCertInclusionController st
+      void $
+        forkLinkedWatcher registry "NodeKernel.voteMinting" $
```
tbagrel1: Should be renamed `voteForging(Thread)` or `forgeVoteThread` for consistency

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
tbagrel1: No need to eta-reduce here.

BTW, same here `voteMintingController` should be renamed.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+    Left err -> do
+      trace $ TraceCantReadEnv err
+      exitEarly
+    Right poolId -> pure poolId
+
+  privateKey <- case readPerasPrivateKeyFromEnv (Proxy @blk) of
+    Left err -> do
+      trace $ TraceCantReadEnv err
+      exitEarly
+    Right privateKey -> pure privateKey
```
tbagrel1: As said yesterday, obtaining these info should probably be done outside of the voting controller.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
       trace :: TracePerasCertInclusionEvent blk -> WithEarlyExit m ()
       trace ev = lift $ traceWith (perasCertInclusionTracer tracers) ev
 
+voteMintingController ::
```
tbagrel1: Needs to be renamed

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+  , HasHardForkHistory blk, BlockSupportsPeras blk
+  , LedgerStateHeaderStateSupportsPerasVoting blk
+  ) =>
+  SystemTime m ->
+  InternalState m remotePeer localPeer blk ->
+  SlotNo ->
+  WithEarlyExit m ()
+voteMintingController systemTime IS{chainDB, tracers} slotNo = do
+
+  poolId <- case unsafePerasPoolIdFromEnv of
```
tbagrel1: Should be replaced by `readPerasPoolIdFromEnv`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+    Left err -> do
+      trace $ TraceCantReadEnv err
+      exitEarly
+    Right privateKey -> pure privateKey
+
+  (voteM :: Maybe (ValidatedPerasVote blk), traceEvents :: [TracePerasVotingEvent blk]) <- lift $ atomically $ do
+
+    -- Resolve Peras round number
+    roundInfo <- Time.resolveSlotToPerasRoundInfoWithHandle
+      (ChainDB.getTimeResolutionContextHandle chainDB)
```
tbagrel1: Ideally you can get all the handles outside, and pass them as parameters to the voting controller. It would clean up the code

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+      exitEarly
+    Right poolId -> pure poolId
+
+  privateKey <- case readPerasPrivateKeyFromEnv (Proxy @blk) of
+    Left err -> do
+      trace $ TraceCantReadEnv err
+      exitEarly
+    Right privateKey -> pure privateKey
+
+  (voteM :: Maybe (ValidatedPerasVote blk), traceEvents :: [TracePerasVotingEvent blk]) <- lift $ atomically $ do
```
tbagrel1: Using Writer monad (for logging, as suggested by @adithyaov) + some sort of early exit (through the monadic aspect of `Maybe`) could help to make it more readable.

Feel free to try a couple different designs and see what is better.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+              poolId
+              privateKey
+              roundNo
+              candidateBlock
+            case voteM of
+              Nothing -> pure (Nothing, [traceVotingRule, TraceNotElectedInRound roundNo])
+              Just vote -> pure (Just vote, [traceVotingRule])
+
+  traverse_ trace traceEvents
+  case voteM of
```
tbagrel1: usually we name this `mVote` or `mbVote`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+    Nothing -> exitEarly
+    Just vote -> do
+      (now :: RelativeTime) <- lift $ systemTimeCurrent systemTime
```
tbagrel1: We tend not to use type annotations on the same line as a let/bind in the codebase.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+            -- Mint the vote, if allowed
+            (voteM :: Maybe (ValidatedPerasVote blk)) <- forgePerasVoteIfEligibleInContext
+              (ChainDB.getPerasEpochContextResolverHandle chainDB)
+              poolId
+              privateKey
+              roundNo
+              candidateBlock
+            case voteM of
+              Nothing -> pure (Nothing, [traceVotingRule, TraceNotElectedInRound roundNo])
+              Just vote -> pure (Just vote, [traceVotingRule])
```
tbagrel1: This also needs a tracing event, saying we successfully forged a vote

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+
+  traverse_ trace traceEvents
+  case voteM of
+    Nothing -> exitEarly
+    Just vote -> do
+      (now :: RelativeTime) <- lift $ systemTimeCurrent systemTime
+      -- Add vote to DB
+      -- TODO: process new return types, possibly log, once the following is merged:
+      -- https://github.com/IntersectMBO/ouroboros-consensus/pull/2029
+      lift $ ChainDB.addPerasVoteSync chainDB $ WithArrivalTime now vote
```
tbagrel1: We could also have a tracing event saying the vote was added to the DB

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+        votingDecision <- isPerasVotingAllowedInContext
+          (ChainDB.getPerasVotingViewHandle chainDB)
+          roundNo
+        let traceVotingRule = TraceVotingRuleEvent votingDecision
+        case votingDecision of
+          NoVote _ -> pure (Nothing, [traceVotingRule])
+          Vote _ candidateBlock -> do
+
+            -- Mint the vote, if allowed
+            (voteM :: Maybe (ValidatedPerasVote blk)) <- forgePerasVoteIfEligibleInContext
```
tbagrel1: We tend not to use type annotations on the same line as a let/bind in the codebase.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +507,75 @@ initNodeKernel
...
+      exitEarly
+    Right privateKey -> pure privateKey
+
+  (voteM :: Maybe (ValidatedPerasVote blk), traceEvents :: [TracePerasVotingEvent blk]) <- lift $ atomically $ do
+
+    -- Resolve Peras round number
+    roundInfo <- Time.resolveSlotToPerasRoundInfoWithHandle
+      (ChainDB.getTimeResolutionContextHandle chainDB)
+      slotNo
+    let roundNo = Time.stpriPerasRoundNo roundInfo
```
tbagrel1, line 510: Do we need the qualified name?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE TypeOperators #-}
```
tbagrel1: Probably not needed?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS/Unsafe.hs
```diff
@@ -34,6 +37,20 @@ import System.IO.Unsafe (unsafePerformIO)
 keyScope :: BLS.KeyScope
 keyScope = "TESTNET"
 
+-- | Read a private key from the environment variable 'PERAS_POOL_ID'
+unsafePerasPoolIdFromEnv :: Either String PoolId
```
tbagrel1: I don't think you need this function, given the addition of `readPerasPoolIdFromEnv` in `BlockSupportsPeras`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Trace.hs
```diff
@@ -0,0 +1,13 @@
...
+import Ouroboros.Consensus.Peras.Voting.Rules (PerasVotingRulesDecision)
+import Ouroboros.Consensus.Block (PerasRoundNo)
+import Data.Word (Word64)
+
+data TracePerasVotingEvent blk
+  = TraceVotingRuleEvent (PerasVotingRulesDecision blk)
+  | TraceNoVoteAfterFirstSlotInRound PerasRoundNo Word64 -- slot index
+  | TraceNotElectedInRound PerasRoundNo
+  -- | TODO: Maybe get rid of that when we no longer read crypto data from env variables
+  | TraceCantReadEnv String
```
tbagrel1: Usually we prefix the constructor names with the type of the trace event (so here `TracePerasVoting...`)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Trace.hs
```diff
@@ -0,0 +1,13 @@
+module Ouroboros.Consensus.Peras.Voting.Trace (TracePerasVotingEvent (..)) where
+
+import Ouroboros.Consensus.Peras.Voting.Rules (PerasVotingRulesDecision)
+import Ouroboros.Consensus.Block (PerasRoundNo)
+import Data.Word (Word64)
+
+data TracePerasVotingEvent blk
+  = TraceVotingRuleEvent (PerasVotingRulesDecision blk)
+  | TraceNoVoteAfterFirstSlotInRound PerasRoundNo Word64 -- slot index
+  | TraceNotElectedInRound PerasRoundNo
```
tbagrel1: `Elected` is slightly confusing, `Eligible` would be more consistent

tbagrel1 (commented):
Left a couple of comments, according to what we discussed yesterday :)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
nbacquey: I'd like to include `peras` in the name, for consistency with similar tracers

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I added `WriterT` somewhere, along with `WithEarlyExit` form the outer scope, which is `MaybeT` in a trench coat. I like it a bit better, even if there are sightings of `lift . lift` here and there...

`capabilities` is one of the fancy stuff we don't yet use, I didn't want to add it to the stack.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: Putting it outside of the voting controller would mess a bit with the tracing. As this is not the definitive version of the code, I settled with a TODO for when we get the proper values from the context.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I think it makes the code clearer to explicitly state that the three functions belong to the same module and ought to work together.
With all due respects to the work involved in creating the `Time` module, I think the identifiers there are quite scary and should not be living unqualified in the rest of the code base :D

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Trace.hs
nbacquey: Replaced with `NotAVoter`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I'm not sure what you mean exactly. Should we gather all 4 handles and pass them to the voting thread, instead of just passing the ChainDB? I think it would clutter the signature more that it would clean the code

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
tbagrel1: For consistency, I would prefer if the type name could match the new field name, i.e. `TracePerasVoteForgingEvent`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
tbagrel1: The parent context has access to the tracers inside `st`.
I think the PoolId can't change at all (hash of the public cold key), so it should be obtained outside of the voting controller.

I guess the private key may change over time, so I'm fine with it being inside the controller.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
tbagrel1: Agree indeed; since we have the ChainDB at hand anyway, we can get them just when we need them.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -493,6 +506,85 @@ initNodeKernel
...
+      exitEarly
+    Right poolId -> pure poolId
+
+  privateKey <- case readPerasPrivateKeyFromEnv (Proxy @blk) of
+    Left err -> do
+      trace $ TracePerasVotingCantReadEnv err
+      exitEarly
+    Right privateKey -> pure privateKey
+
+  (mVote, traceEvents :: [TracePerasVotingEvent blk]) <- atomically $ runWriterT $ do
```
tbagrel1: Could we add a comment just above saying that we run all STM computations in a WriterT monad to have proper logging while keeping the 3 actions in a same atomic block?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
tbagrel1: FTR I'm not against removing the scary prefix; but Is it ok to have 6-8 modules instead of 1 for time conversions? If it is, I can ask claude to do this for us.

tbagrel1:
Looks much better now, thanks @adithyaov  @nbacquey !

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I quite like it as it is, with a qualified import. What's bothering you about it exactly?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: If I remember correctly our conversation with @nfrisby, it is a design choice for `initNodeKernel` to be unaware of who the pool is, and to access identity as low as possible.
I'd rather leave it there until we decide on a proper way to access the pool's credentials

tbagrel1: What I mean is the parent function that calls the controller every second should probably resolve the identity of the pool, and pass it down to the controller.

But it's not a hill I will die on ^^

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -95,6 +96,7 @@ data Tracers' remotePeer localPeer blk f = Tracers
       f (TraceLabelPeer remotePeer (TracePerasVoteDiffusionOutbound blk))
   , perasCertInclusionTracer ::
       f TracePerasCertInclusionEvent
+  , perasVoteForgingTracer :: f (TracePerasVotingEvent blk)
```
tbagrel1: We probably want the type to be `TracePerasVoteForgingEvent`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,89 @@ initNodeKernel
...
+    -- Forge the vote, if allowed
+    mVote <-
+      dyel $
+        forgePerasVoteIfEligibleInContext
+          (ChainDB.getPerasEpochContextResolverHandle chainDB)
+          poolId
+          privateKey
+          roundNo
+          candidateBlock
+    when (isNothing mVote) $ lift $ tell $ [TracePerasVotingNotAVoterInRound roundNo]
```
tbagrel1: I would maybe use a `case` here, to get consistency with the rest of the do block:
```suggestion
    case mVote of
      Nothing -> do
        lift $ tell $ [TracePerasVotingNotAVoterInRound roundNo]
        hoistMaybe Nothing
      Just vote -> pure vote
```

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,89 @@ initNodeKernel
...
+    hoistMaybe mVote
+
+  traverse_ trace traceEvents
+  now <- lift $ systemTimeCurrent systemTime
```
tbagrel1: I would probably extract the vote out of the maybe; before trying to add the arrival time to it, so we exit as early as possible

tbagrel1 (approved):
Left a couple of nitpicking comments, feel free to ignore them

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I already use `when` for another `Maybe`, at line 509.
I prefer that approach, which I couldn't use later on at 520, because the type there is a bit more complex.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -426,6 +446,12 @@ initNodeKernel
         , getTxMempoolSem = txMempoolSem
         }
    where
+    -- Start a thread conditionally when the PerasFlag is provided and the
+    -- current block type supports Peras.
+    whenPerasEnabled =
+      when $
+        PerasFlag `member` featureFlags
+          && blockDoesReallySupportsPeras (Proxy @blk)
     blockForgingController ::
```
agustinmista, line 456: NIT: add newline

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,90 @@ initNodeKernel
...
+          privateKey
+          roundNo
+          candidateBlock
+    when (isNothing mVote) $ lift $ tell $ [TracePerasVotingNotAVoterInRound roundNo]
+    hoistMaybe mVote
+
+  traverse_ trace traceEvents
+  vote <- maybe exitEarly pure mVote
+  now <- lift $ systemTimeCurrent systemTime
+  let tickedVote = WithArrivalTime now vote
```
agustinmista: The `addArrivalTime` helper does exactly this. You probably need a `lift` somewhere, though.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,90 @@ initNodeKernel
...
+    Left err -> do
+      trace $ TracePerasVotingCantReadEnv err
+      exitEarly
+    Right privateKey -> pure privateKey
+
+  -- We run all 3 STM computations in a WriterT monad so that we can have proper logging,
+  -- while keeping everything in the same transaction. We also use MaybeT because there is
+  -- a natural abort/continue logic within the transaction. Unfortunately, we can't leverage
+  -- the outer WithEarlyExit monad, because we _always_ want to get the trace.
+  (mVote, traceEvents :: [TracePerasVoteForgingEvent blk]) <- lift $ atomically $ runWriterT $ runMaybeT $ do
```
agustinmista, line 499: I always get confused by this: is this the right stack order to always get the trace regardless of the `Maybe` result? Or should it be `runMaybeT . runWriterT` instead?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,90 @@ initNodeKernel
...
+      lift $ tell [TracePerasVotingNoVoteAfterFirstSlotInRound roundNo slotInRound]
+      hoistMaybe Nothing
+
+    -- Do the voting rules state that we should vote?
+    votingDecision <-
+      dyel $
+        isPerasVotingAllowedInContext
+          (ChainDB.getPerasVotingViewHandle chainDB)
+          roundNo
+    lift $ tell [TracePerasVotingRuleEvent votingDecision]
```
agustinmista: I think you _should_ be able to just `tell` --- `MaybeT` [is](https://hackage-content.haskell.org/package/mtl-2.3.2/docs/Control-Monad-Writer-Class.html) `MonadWriter`, too.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,90 @@ initNodeKernel
...
+      dyel $
+        Time.resolveSlotToPerasRoundInfoWithHandle
+          (ChainDB.getTimeResolutionContextHandle chainDB)
+          slotNo
+    let roundNo = Time.stpriPerasRoundNo roundInfo
+        slotInRound = Time.stpriSlotsSpentInPerasRound roundInfo
+
+    when (not $ Time.stpriIsFirstSlotOfPerasRound roundInfo) $ do
+      lift $ tell [TracePerasVotingNoVoteAfterFirstSlotInRound roundNo slotInRound]
+      hoistMaybe Nothing
```
agustinmista, line 515: Is it correct to say that we only vote at the first slot of a new round?

If so, I think we can be a bit clever and not do time resolution on every slot, and instead keep track of the next slot that we should do time resolution again (i.e. at the beginning of the next round). That way we could reduce the amount of calls to the time resolver by ~90x.

This assumes that the length of a round won't change within a round, which I think is fair. 

CC @tbagrel1 what do you think?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -439,6 +465,90 @@ initNodeKernel
...
+  -- Add vote and potential cert to the DB
+  (addVoteResult, mAddCertChainSelOutcome) <- lift $ ChainDB.addPerasVoteSync chainDB tickedVote
+  trace $ TracePerasVotingAddVoteResult addVoteResult
+  traverse_ (trace . TracePerasVotingAddCertChainSelOutcome) mAddCertChainSelOutcome
+ where
+  trace :: TracePerasVoteForgingEvent blk -> WithEarlyExit m ()
+  trace = lift . traceWith (perasVoteForgingTracer tracers)
+
+  -- Do you even lift, bro?
+  dyel = lift . lift
```
agustinmista, line 553: I like it, but now sure it will fly with the real reviewers 😸

agustinmista (commented):
Good job! Some minor comments I'd like to discuss with you.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: Indeed it is :+1: 
> [MonadWriter](https://hackage-content.haskell.org/package/mtl-2.3.2/docs/Control-Monad-Writer-Class.html#t:MonadWriter) w m => [MonadWriter](https://hackage-content.haskell.org/package/mtl-2.3.2/docs/Control-Monad-Writer-Class.html#t:MonadWriter) w ([MaybeT](https://hackage-content.haskell.org/package/transformers-0.6.1.0/docs/Control-Monad-Trans-Maybe.html#t:MaybeT) m)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I think the order here is correct. You can check with this snippet, which I shared [here](https://play.haskell.org/saved/VJArDQ6T):

```haskell
import Control.Monad (when)
import Control.Monad.Trans.Class (lift)
import Control.Monad.Trans.Maybe (hoistMaybe, runMaybeT)
import Control.Monad.Writer (runWriterT, tell)

main :: IO ()
main = do

  -- Transformers order is runWriterT . runMaybeT
  
  -- Condition is true, computation gets interrupted with Nothing
  result <- runWriterT $ runMaybeT $ do
    tell "a"
    cond <- lift $ lift true
    when cond $ do
      tell "b"
      hoistMaybe $ Nothing
    tell "c"
    hoistMaybe $ Just ()
  print result

  -- Condition is false, computation ends normally with Just ()
  result <- runWriterT $ runMaybeT $ do
    tell "a"
    cond <- lift $ lift false
    when cond $ do
      tell "b"
      hoistMaybe $ Nothing
    tell "c"
    hoistMaybe $ Just ()
  print result

  -- Transformers order is runMaybeT . runWriterT

  -- Condition is true, computation gets interrupted and logging is lost
  result <- runMaybeT $ runWriterT $ do
    tell "a"
    cond <- lift $ lift true
    when cond $ do
      tell "b"
      lift $ hoistMaybe $ Nothing
    tell "c"
    lift $ hoistMaybe $ Just ()
  print result
  
  -- Condition is false, computation ends normally and logging is preserved
  result <- runMaybeT $ runWriterT $ do
    tell "a"
    cond <- lift $ lift false
    when cond $ do
      tell "b"
      lift $ hoistMaybe $ Nothing
    tell "c"
    lift $ hoistMaybe $ Just ()
  print result
  
true :: IO Bool
true = pure True

false :: IO Bool
false = pure False
```
The output is:
```
(Nothing,"ab")
(Just (),"ac")
Nothing
Just ((),"ac")
```
It also shows in the return type: with `runWriterT . runMaybeT`, the return type is `(Maybe (), String)`, while with `runMaybeT . runWriterT`, the return type is `Maybe ((), String)`, meaning that if the computation aborts with `Nothing`, then the logs get lost

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: > Is it correct to say that we only vote at the first slot of a new round?

It is correct. In theory, we could also compute which slots are going to be voting slots at the start of each epoch (if I'm not mistaken, protocol parameters can only change at epoch boundaries).

However, this means introducing an additional bit of state to manage, and I'm not sure the potential pitfalls are worth the gain: one call to the time resolver per second _might_ be alright?

My opinion would be to leave a TODO, to be resolved when we have proper benchmarks for Peras

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nbacquey: I actually had a bit of trouble thinking how to name this helper :grin:
I'm open for suggestions

But also, think of the memes!

nbacquey:
I cleaned the commit history a little bit. Feel free to merge if you think all your concerns have been assuaged

nbacquey:
Has been merged manually

PR #2073, opened 2026-06-12
Leios late join v3

dnadales:
# Leios: late-join catch-up for CertRBs with a missing EB closure

## Summary

A late joiner, or a node that reconnects after losing connectivity, can receive a ranking block carrying a Leios certificate (a CertRB) whose endorsement-block (EB) closure it does not hold.
Applying it needs the closure, so the node crashes.
This PR holds such a CertRB back from chain selection until the closure is fetched, infers the fetch from the peer's CertRB header, and adopts the block once the closure arrives.

Base branch: `leios-prototype`.
Addresses input-output-hk/ouroboros-leios#890.

## The problem

CertRBs arriving over BlockFetch are written to the volatile DB and selected at once, which runs ledger application through `resolveLeiosBlock`.
That needs the EB closure in the local LeiosDb; when it is missing the node crashes with "Announced EB is being certified by the current chain but it's not available?".

The LeiosFetch protocol only fetches EBs a peer has explicitly offered.
A late joiner was never offered the closure it is missing, so it cannot discover it.
The result: a new node cannot sync, and a node that falls behind cannot recover without copying another node's data directory.

## Approach

The issue proposes two phases.
Phase 1 buffers a CertRB outside the volatile DB until its closure arrives, then promotes it.
Phase 2 is a fetch mechanism: ChainSync inference (Option A) or batched range requests (Option B).

We keep the two-phase shape but change Phase 1: the CertRB is written to the volatile DB as usual and held back from chain selection while its closure is missing, with no separate buffer.
For Phase 2 we take Option A: a peer's CertRB header is an implicit offer of the EB closure it certifies, which drives the existing LeiosFetch machinery.
Decisions 1 and 4 justify these.

The new code is a late-join kernel: an AddBlock listener, the `LeiosLateJoinState` bookkeeping, a trigger worker, an AcquiredEbTxs subscriber, a GC pruner, and a startup walker.
It sits alongside the existing `ChainDB`, `ChainSync client`, `LeiosFetch`, and `LeiosDb`.

The catch-up loop:
A CertRB arrives and `ChainDB` triggers the listener.
The listener writes the block into the blocked set.
`ChainDB` reads that set during chain selection and skips the block.
The `ChainSync client` triggers an implicit closure offer for the announced EB.
`LeiosFetch` writes the acquired closure to `LeiosDb`.
The subscriber reads the acquisition notification and writes a pending trigger.
The trigger worker triggers reconsideration in `ChainDB`, which re-runs selection and adopts the block.

The startup walker covers restart: at boot it replays the volatile-DB headers through the listener, so a node that restarts with blocked CertRBs on disk rebuilds the bookkeeping before selection runs.

### 1. Hold CertRBs back from chain selection, not in a separate buffer

The CertRB follows the normal path into the volatile DB.
A chain-selection filter keeps it out of selection while its closure is missing; it is reconsidered once the closure is present.

Why: a volatile-DB write does not apply the ledger; selection does, and that is where the missing-closure crash happens.
So a CertRB can sit in the volatile DB unselected at no cost.
The block bytes live once, in the volatile DB, which already addresses and garbage-collects them; the kernel keeps only the header hash in a blocked set.
The filter mirrors the existing one for invalid blocks.

Rejected: a separate staging area, the issue's literal Phase 1.
It is a different store, not a copy, so the cost is not duplication.
The cost is that it intercepts BlockFetch before insertion, adds a promote step, and needs its own size bound, eviction, and restart handling.
The chosen approach takes all of that from the volatile DB and adds only the blocked-set bookkeeping, pruned at the GC frontier (decision 7).

### 2. Add generic, Leios-agnostic primitives to the ChainDB

Four additions, none mentioning Leios.
`registerHeaderListener` runs a caller-supplied STM action for every block added.
A blocks-to-ignore source excludes header hashes from chain selection.
`reconsiderBlockAsync` re-enqueues a previously-ignored block.
`getLastGcSlot` exposes the GC frontier.

Why: the ChainDB never learns what a CertRB or a closure is.
All Leios knowledge stays in the Leios module, and the dependency points one way, Leios to ChainDB.
The primitives serve any "defer this block until a side condition holds" need.

Rejected: teaching the ChainDB about CertRBs and closures, which couples block storage to the Leios feature and the LeiosDb.

### 3. Keep the bookkeeping in the Leios kernel, fed by a synchronous STM listener

`LeiosLateJoinState` owns the maps tracking which CertRBs are blocked, which announcements have been seen, and which blocks await reconsideration.
The header listener fires inside the AddBlock step, in the same STM transaction that commits the block.

Why: in one transaction, an observer woken by the block-written signal sees the block on disk and the bookkeeping update together.
There is no window where one is visible without the other.

Rejected: an asynchronous listener, which would expose a window where a block is on disk but not yet recorded as blocked, and a concurrent chain selection could act on it.

### 4. Fetch the closure via implicit offers (Option A), not range requests (Option B)

When ChainSync rolls a peer's CertRB header forward, the node records an EB-closure offer for the EB the parent announces, then lets LeiosFetch fetch it.

Why: it reuses LeiosFetch and the existing peer offer-tracking.
The header already says which EB the CertRB certifies, so no new message is needed.

Rejected: Option B, batched `MsgLeiosBlockRangeRequest`.
It is still a stub and a larger protocol change than the workshop scope needs.
The issue asked for one fetch option; this is it.

### 5. Source the EB byte size from a new `headerEbAnnouncement` method

`headerEbAnnouncement`, added to `ResolveLeiosBlock`, returns the announced EB hash with its byte size; the implicit offer uses both.

Why: the existing `headerLeiosAnnouncement` returns only the hash.
The offer needs the size, because on EB-body receipt the node checks received size against offered size by exact equality and aborts on a mismatch.

Safety: on the honest forge path the announced size equals the forged EB body's size by construction, and the body is addressed and hash-checked on receipt.
So received and forged bodies match byte for byte and the sizes are equal; the implicit offer cannot cause a size-mismatch abort on the honest path.

Rejected: widening `headerLeiosAnnouncement` to carry the size, which changes the accessor every caller uses for a value only the offer path needs.

### 6. Add a header `isCertRB` flag with a canonical 12-element CBOR encoding

The Praos header carries a flag marking it a CertRB, in a fixed 12-element CBOR array.

Why: the listener and the filter classify a block as a CertRB from its header alone, before the body is resolved.

Rejected: deriving CertRB-ness by resolving the body, which defeats the point: the problem is a body whose closure is not yet available.

Consequence: the header encoding changes, so this PR updates the golden serialisation files.
The CDDL spec needs the matching 12-element update upstream; until it lands the CDDL-compliance test fails (see gaps).

### 7. Bound the bookkeeping at the GC frontier

A pruning loop drops bookkeeping for blocks whose slot is below the ChainDB's last GC slot, read through `getLastGcSlot`.
The reverse index of orphaned CertRBs holds only entries waiting for their announcer, which is transient.

Why: the maps cannot grow without bound on the common path; once the volatile DB drops a block, its bookkeeping goes too.

Limit: this bounds the honest case.
Full bounding against a flood of adversarial CertRBs is out of scope and deferred, matching the issue.

Rejected: unbounded maps, which would leak for the process lifetime.

### 8. Use one wiring path for the threadnet harness and production

The boot glue that installs the hooks and forks the background workers lives in shared helpers, called by both the threadnet node and the production node boot.

Why: the two paths cannot drift, since they set the same fields and fork the same workers through one definition.
A green threadnet result transfers to production by construction.

Coverage: the production boot path (`runWith`) is not exercised by any in-repo test; the threadnet tests drive the harness path.
The production wiring is verified by compilation and by sharing the helpers, not by a test that runs it (see gaps).

## What the change verifies

`prop_leios_late_join` (ThreadNet) starts one node late, its join slot capped at a quarter of the run, and checks that it catches up.
It asserts every node agrees on its chain prefix up to a cutoff at three-quarters of the run, well past the join slot.
The check is not vacuous: the compared prefixes are real and several blocks long.
Without the fetch path the late joiner's prefix is strictly shorter and the assertion fails for the right reason; with it, the prefixes agree.

`prop_leios` and the consensus storage, mock, Shelley, Byron, and Cardano suites were run and pass, except for the two items in the gaps section.

## Scope and known gaps

Deferred, matching the issue:

- Bulk sync from genesis.
- Adversarial CertRB handling.
- Full resource bounding (see decision 7).

Known gaps:

- The CDDL spec is not yet updated for the 12-element header.
  The CDDL-compliance test fails until that update lands upstream.
- `prop_leios` has a pre-existing flake in the voting-diffusion layer ("created votes not diffused").
  It is unrelated to late join and reproduces on the base branch with the same seed.
- The production `runWith` boot path is not exercised by any in-repo test (see decision 8).
- The full `required` CI build was not run locally; verification was per-component `-Werror` builds plus the targeted suites above.

## Issue

Addresses input-output-hk/ouroboros-leios#890: Phase 1 as chain-selection deferral, Phase 2 as Option A.
The deferred items above remain open.


dnadales:
# Leios CertRB late-join: staging area (#2058) vs a bookkeeping alternative 

Both approaches fix the same crash.
A late-joining or briefly-disconnected node calls `resolveLeiosBlock` and errors (Shelley `Ledger.hs`) when a CertRB arrives before its certified EB closure is local.

-   **Staging area**: PR #2058, merged into `leios-prototype`. The PR describes itself as a stop-gap.
-   **Bookkeeping**: a proposed alternative (the `leios-late-join-v3` branch).

If you read PR #2058, read the merged code, not the description.
The description claims the cert carries a `(LeiosPoint, BytesSize)` payload decoded by a `checkLeiosBlockResolvable` method.
Neither shipped.
The merged gate reads the parent header's `headerLeiosAnnouncement`.
Both approaches learn the EB point from the header.


<a id="org3cc18d5"></a>

# What they share

-   Both learn the EB point from the parent header announcement, scanning the current chain and the ChainSync candidate fragments. Neither reads the cert body or the Praos chain-dep state.
-   Both release the block only on `AcquiredEbTxs` (the full closure), not on `AcquiredEb` (the body alone).
-   Both fetch the closure through the existing LeiosFetch machinery, feeding both the body (`choosePeerEb`) and the tx (`choosePeerTx`) halves.


<a id="org3cb0a3b"></a>

# The core difference

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Axis</th>
<th scope="col" class="org-left">Staging area (#2058)</th>
<th scope="col" class="org-left">Bookkeeping</th>
</tr>
</thead>
<tbody>
<tr>
<td class="org-left">Layer</td>
<td class="org-left">Above ChainDB: wraps BlockFetch's ChainDbView</td>
<td class="org-left">Inside ChainDB selection</td>
</tr>

<tr>
<td class="org-left">Where a blocked CertRB sits</td>
<td class="org-left">Parked in a TVar; never enters the VolatileDB or ChainSel</td>
<td class="org-left">In the VolatileDB; reaches ChainSel but is filtered from selection</td>
</tr>

<tr>
<td class="org-left">What blocks</td>
<td class="org-left">The BlockFetch client thread (an STM <code>retry</code>)</td>
<td class="org-left">Nothing; the CertRB is skipped at selection, reconsidered later</td>
</tr>

<tr>
<td class="org-left">Fetch trigger</td>
<td class="org-left">Per-tick synthesis from the staged snapshot</td>
<td class="org-left">A one-shot implicit offer when the header arrives</td>
</tr>

<tr>
<td class="org-left">Release</td>
<td class="org-left">A drain thread re-submits the block via <code>addBlockAsync</code></td>
<td class="org-left">A subscriber enqueues a reconsider request</td>
</tr>

<tr>
<td class="org-left">ChainDB API change</td>
<td class="org-left">None</td>
<td class="org-left">Six additions (a header listener, a reconsider call, a blocks-to-ignore source, a last-GC-slot query, two boot hooks)</td>
</tr>
</tbody>
</table>

The difference is layering.
The staging area parks the block above ChainDB and stalls a fetch thread.
The bookkeeping approach leaves the block in place and filters it out of selection, stalling nothing.


<a id="org212ff68"></a>

# Staging-area limitations, and whether the bookkeeping approach closes them

Each row is a known limitation of the staging area: its four open points and the two costs of the STM-retry gate.

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Staging-area limitation</th>
<th scope="col" class="org-left">Closed?</th>
<th scope="col" class="org-left">How</th>
</tr>
</thead>
<tbody>
<tr>
<td class="org-left">A fork CertRB with an unknown parent is admitted blindly and crashes if that fork is later selected</td>
<td class="org-left">Yes</td>
<td class="org-left">The filter prunes the CertRB and its whole successor subtree at selection. It is never validated without the closure.</td>
</tr>

<tr>
<td class="org-left">Startup selection bypasses the gate, so a closure-missing CertRB still crashes on restart</td>
<td class="org-left">Yes</td>
<td class="org-left">A boot hook seeds the filter before startup selection runs, covering the case the staging area leaves open.</td>
</tr>

<tr>
<td class="org-left">The gate stalls the peer's whole BlockFetch pipeline for the closure-fetch duration</td>
<td class="org-left">Yes</td>
<td class="org-left">No fetch thread is blocked; the CertRB is not selected until it is reconsidered.</td>
</tr>

<tr>
<td class="org-left">If every offering peer disconnects, the parked BlockFetch thread leaks forever</td>
<td class="org-left">Yes</td>
<td class="org-left">Nothing is parked, so no thread can leak.</td>
</tr>

<tr>
<td class="org-left">Held bookkeeping is never garbage-collected</td>
<td class="org-left">Partial</td>
<td class="org-left">A pruner bounds the large per-block cache; the two smaller CertRB maps are still unbounded.</td>
</tr>

<tr>
<td class="org-left">Restart loses recovery of a body received without its full tx closure</td>
<td class="org-left">No</td>
<td class="org-left">Both rely on in-memory offers, and both defer this.</td>
</tr>

<tr>
<td class="org-left">Fetching the closure for a CertRB whose parent is off the candidate fragment is not guaranteed</td>
<td class="org-left">No</td>
<td class="org-left">Both are best-effort, relying on other peers' offers. The bookkeeping approach still avoids the crash; the staging area would crash.</td>
</tr>
</tbody>
</table>

The bookkeeping approach closes the staging area's two crashes and removes both costs of the gate.
The fork-crash fix holds for honestly-produced blocks: a CertRB whose parent never arrives is unreachable from the immutable anchor, so chain selection never builds a candidate for it.
It is not strictly safer, though.
Gating on the header bit opens a crash the staging area avoids (next section).
It only partly closes the GC gap, and defers the restart-recovery and off-fragment-fetch cases.


<a id="org1507062"></a>

# Where the staging area is the safer one

-   **Header bit vs body cert.** The bookkeeping approach decides hold-back from the header bit (`hbIsCertRB`), but `resolveLeiosBlock` crashes off the body cert. A block whose header bit is clear but whose body carries a cert is never held back, so it crashes at apply. The staging area reads the body (`blockHasLeiosCert`), so it avoids this. The crash is reachable only from a Byzantine producer, since the header fields are KES-signed and a relay cannot forge them. Neither approach enforces header/body consistency.
-   **Lost-wakeup release.** Both have a lost-wakeup race, and the staging area's is worse. Its gate queries the closure and stages the block in separate transactions, so a notification in that window parks the client thread forever even though the closure is present, with no fallback. The bookkeeping approach self-heals: its blocked-set read subtracts the completed-closure set every time, so the next selection unblocks the CertRB.

The bookkeeping approach also carries fixable gaps of its own: the two CertRB maps grow unbounded, and the implicit-offer path has robustness bugs (an over-broad exception catch, a 10 ms busy-wait, a non-atomic two-field write).
Net: it trades the staging area's parking, thread-blocking, and two crashes for a larger ChainDB API surface plus these gaps.


<a id="orga95f2ab"></a>

# Blind spots in both

Neither approach has an adversarial threat model; the design issue (input-output-hk/ouroboros-leios#890) deferred adversarial and equivocating CertRBs.

-   **Unfetchable EB.** Neither admits CertRBs under any limit. A CertRB whose closure no honest peer holds costs the staging area a parked thread, a staged entry, and recurring per-tick work; it costs the bookkeeping approach one map entry. The staging area is the worse target.
-   **Wrong-size or wrong-hash body.** Shared code: on receipt the EB size and hash are checked by exact equality, and the fetch thread errors on mismatch before the peer is cleaned up. A relay serving a malformed body kills the LeiosFetch client thread. Both are affected equally.
-   **Re-org.** If the chain switches away from the fork holding the blocked CertRB, the staging area leaks the entry and the parked thread. The bookkeeping approach's reconsider safely does nothing on a garbage-collected block.
-   **Scaling.** For N stacked CertRBs, the staging area does O(N × peers × fragment) work per tick and parks N threads; the bookkeeping approach does O(N²) across the unblock burst. Both are bounded and self-terminating.
-   **LoE interaction (bookkeeping only), verified safe.** The LoE trim's intersection precondition cannot be broken by the filter, because the LoE fragment and every candidate share the current chain's anchor. A CertRB blocked by both the filter and the LoE is selected once both clear, because both reprocess paths re-read both gates.
-   **Restart (bookkeeping only).** A restart does not strand a held-back CertRB. The boot walk re-holds it without recording an offer, but because the CertRB stays off the selected chain it sits below the ChainSync intersection, so on reconnect the peer re-sends its header, the offer is recorded again, and the closure is fetched. This needs a peer that still serves the header. The staging area has no equivalent re-trigger.


<a id="org78bfaef"></a>

# Sources

PR #2058 (merged) and design issue input-output-hk/ouroboros-leios#890.
Code references name functions in the consensus repository.

PR #2074, opened 2026-06-15
Leios: EB Fetching and staging area fixes

### LINKED ISSUE #890: Prototype: CertRB staging area and catch-up

ch1bo:
Resolves three issues we found with the current prototype related to EB fetching and staging:

- Offered EBs with size 0 are ignored and not override honest, non-0 size offers
- Resolve announced EB hash using all chain sync candidate fragments instead of current chain only
- Garbage collection of staging area to avoid forever stalled EBs (e.g. when we lost the original source peer)
- Resolution of multiple CertRBs that were blocked on the same EbHash
- Fixes a size-match condition that was vetoing unrelated offers in the fetching logic.

Fixes are unit tested where possible and some additional tracing in the fetching logic should help us pin down any remaining problems. Not fixed this round:

- First non-0 size of `MsgLeiosBlockOffer` is trusted -> should use validated EB announcements for that instead (see FIXMEs) https://github.com/input-output-hk/ouroboros-leios/issues/772
- The fetching logic implementation is quite error-prone with it's many `MVar`s, refactoring it to a `STM` with `TVar` or less cross-referencing data types is probably a good idea. CC @nfrisby


nfrisby (commented):
The comments I'm reading here seem reasonable. Two things stand out to me:

- It seems like implementing the EbAnnouncement messaging would eliminate some necessary cruft in its absence (eg the MsgLeiosBlockoffer size field being untrustworthy).

- I'm not clear on the current status of the StagingArea. Damian's https://github.com/IntersectMBO/ouroboros-consensus/pull/2073#issuecomment-4694135613 is a nice comparison. When I read that, it sounds like the StagingArea is a temporary workaround. Is that the correct takeaway? (Inspired by this review---which I started this morning---I'm trying to steer Claude through what I _think_ should replace the StagingArea. I'm time boxing this to today and maybe a little in the morning---I need to focus on retrofitting the LeiosFetch logic instead.)

    - One additional takeaway: my intention for the StagingArea replacement---and I'm guessing the StagingArea too---would also benefit from EbAnnouncements, so maybe we should move that up in the timeline. (And true voting logic depends on it too, eg.)

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -56,90 +120,324 @@ data StagedCertRB peer blk = MkStagedCertRB
...
+          -- deliberately don't consult 'getCurrentChain' here: the
+          -- locally selected chain is incidental — if the block
+          -- extends our chain, the peer that delivered it tracks the
+          -- same chain via ChainSync and the parent is in its
+          -- candidate fragment too.
+          mHeader <- atomically $ do
+            candidates <- candidateFragments varChainSyncHandles
+            pure $ findParentHeader (Block.blockPrevHash blk) candidates
+          case mHeader of
+            Nothing -> error "LeiosStagingArea was asked to resolve a block for which we don't have a header chain"
```
dnadales, line 269: Should this be a graceful `FailedToAddBlock` rather than `error`? The comment argues the parent is on a candidate "by construction", but we read the candidate snapshot here at a different moment than when BlockFetch delivered the block. If a rollback or a peer disconnect drops the parent from every candidate in between, don't we crash the client thread on a path that used to just blind-admit and let ChainSel decide?

There also seems to be a related tension with `test_gateIgnoresCurrentChain`: it puts the parent only on the current chain (no candidate) and expects `SuccesfullyAddedBlock`, but with empty candidates we land in this `Nothing` branch and `error` instead. I wonder if the test is encoding the old blind-admit contract while the code now throws. Something like returning `FailedToAddBlock` here (or restoring the admit) would line the two back up.

ch1bo:
@nfrisby It's a bit more complicated now with these fixes in place and after we see both approaches and how they would attach to the architecture (there are likely more possible takes).

> It seems like implementing the EbAnnouncement messaging would eliminate some necessary cruft in its absence

Agreed, this is something we need to do independently.

> I'm not clear on the current status of the StagingArea.

The initial PR was a only intended as a stop-gap solution that we know works most likely and have something quick to unblock the testnet. This PR now closes some of the short-comings and the otherwise good analysis in https://github.com/IntersectMBO/ouroboros-consensus/pull/2073#issuecomment-4694135613 is not the complete picture with these in place.

I modeled both in C4 and could also write about them in the leios-design document (I plan to write-up at least one design there).
- This is the staging area design: https://s.icepanel.io/HOv1KVrObHqYxm/BbR4
- This is the late joining design: https://s.icepanel.io/HOv1KVrObHqYxm/fHBa

My current opinion:
- I like that the fetching logic knows about the high priority set of EBs/txs that are missing and need to be fetched. "Synthesizing" implicit offers by modifying the fetch logic state is not so nice (neither from chain sync clients in `late-join` nor from the staging area)
- I like the clean "slotting in" of the staging area between block fetch clients and the volatile db via the existing `ChainDbView` interface. No API changes make it feel right.
- I don't like the "stateful" lookup of what to EB hash to fetch of walking parent headers or inspecting chain dep states (but this is true for both approaches)
- I don't like that we can only bail once we have a CertRB without closure persisted (volatileDB not consistent with leiosDB). Is this even avoidable, or could it always happen? A more integrated storage system, where transactions span praos _and_ leios blocks would help but seems orthogonal to the topic here.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -56,90 +120,324 @@ data StagedCertRB peer blk = MkStagedCertRB
...
+      peers <- atomically $ peersThatKnowBlock varChainSyncHandles blk
+      traceWith
+        tracer
+        TraceLeiosCertRBStaged
+          { stagedBlockPoint = show (Block.blockPoint blk)
+          , stagedEbPoint = point
+          , stagedKnownPeers = Set.size peers
+          }
+      outcomeVar <- newEmptyTMVarIO
+      atomically (stage tv point size peers blk outcomeVar)
```
dnadales, line 319: Aren't the completion check (`leiosDbQueryCompletedEbByPoint` above) and this `stage` insert in two separate transactions? If the closure completes in between, the drain handles the `AcquiredEbTxs` against a map that has no entry yet, drops it, and then we insert and park on a `takeTMVar` nobody will fill, so we wait for the 120s GC and only get freed if every peer also drops the block.

I wonder if we could fold the completion re-check into the same STM transaction as `stage` (re-query inside `stage`, admit instead of staging if it completed), or have the drain re-query the DB on an empty-map hit. Could the test suite cover the "closure lands between check and stage" ordering too? The current `test_drainRelease` always stages before completing.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
```diff
@@ -182,8 +183,15 @@ imInsertEbBody stateVar notificationChan point eb = do
...
+          -- hash, so a second insertion at the same hash is a no-op.
+          imEbBodies =
+            Map.insertWith (\_ old -> old) point.pointEbHash entries (imEbBodies s)
+        , -- Mark this point as downloaded. Other points referencing
+          -- the same EB hash are unaffected; each needs its own
+          -- 'leiosDbInsertEbBody' to be considered complete.
+          imEbBodiesDownloaded =
+            Set.insert point (imEbBodiesDownloaded s)
         }
     writeTChan notificationChan $ AcquiredEb point (leiosEbBytesSize eb)
```
dnadales, line 196: Since `AcquiredEbTxs` only ever fires from `imInsertTxs`, do we have a gap when a point's body is inserted *after* its txs are already present (e.g. a sibling point with the same EB hash completed earlier)? In that case no new tx batch arrives, so `imInsertTxs` never runs for this point and the staged CertRB waiting on it never gets released.

Should `imInsertEbBody` itself check whether the closure is already complete at body-insert time and fire `AcquiredEbTxs` for the point if so? `test_multiSlotRelease` inserts both bodies before the txs, so I don't think it would catch this ordering.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -56,90 +120,324 @@ data StagedCertRB peer blk = MkStagedCertRB
...
+    void $ forkLinkedThread registry "LeiosStagingArea.drain" (drainLoop tv notifications)
+    void $ forkLinkedThread registry "LeiosStagingArea.gc" (gcLoop tv)
+    pure
+      LeiosStagingArea
+        { stagedSnapshot = stagedSnapshotSTM tv
+        , wrappedChainDbView = wrapView tv
+        }
+   where
+    -- Pure helpers
+    mergeEntry new old =
```
dnadales, line 197: The comment on `stagedOutcomes` reasons about two peers delivering the *same* CertRB, but could two *distinct* CertRBs share one announced-EB point? The key here is the parent's announcement (`headerLeiosAnnouncement`, slot + EB hash), and `getIsFetched` dedups by block point, so two competing children of one announcer (equivocation, or a short fork where both branches carry the cert for the same EB) both reach the gate and merge under this key.

When that happens `mergeEntry` keeps `old`'s block, so the drain admits only that one; both parked callers still wake with `StagedReleased` and each reports `SuccesfullyAddedBlock` for its own block, including the second one ChainDB never saw. `gcOne` (`:379-384`) has a sibling of this: it checks `peersThatKnowBlock` against the *snapshot* block but `removeEntry` deletes whatever block now sits at that point, so a released-then-re-staged different block can be evicted on the old block's peer set.

I think the consequence is bounded today: the `AddBlockPromise` is `void`ed by the caller (`mkAddFetchedBlock_`), so the false verdict is never read, and the dropped block gets re-fetched and admitted on a later pass. So this reads more as a latent smell than a live bug. Still, a parked caller is woken with a verdict decided for a *different* block than the one it staged, and that only stays harmless while the promise is discarded. Should the entry key on (or validate against) the CertRB's own block point, so each caller's verdict matches the block it staged?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -1,52 +1,116 @@
...
 
--- | A single staged CertRB.
+-- | Why a parked stage caller is being woken up.
+data StagedOutcome
+  = -- | The drain admitted the block via the unwrapped
+    -- 'addBlockAsync'; the caller should report
+    -- 'SuccesfullyAddedBlock'.
+    StagedReleased
+  | -- | GC gave up: no live peer's ChainSync candidate contains the
+    -- staged CertRB anymore, so the closure has become unreachable.
```
dnadales, line 104: Does "demote peer take over" actually happen on this path? The `FailedToAddBlock` we return here is `void`ed by the caller (`mkAddFetchedBlock_`), so I think the verdict is never read, and eviction ends up a silent drop that only re-fetch recovers, no peer demotion. Maybe the comment should say that.

Separately, since `gcOne` evicts on a single empty `peersThatKnowBlock` observation with no grace, could a transient candidate drop (rollback, brief KeepAlive gap) evict a still-valid entry whose closure is still coming? Would requiring a couple of consecutive empty ticks before eviction make this less twitchy?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -56,90 +120,324 @@ data StagedCertRB peer blk = MkStagedCertRB
...
+          { stagedBlockPoint = show (Block.blockPoint blk)
+          , stagedEbPoint = point
+          , stagedKnownPeers = Set.size peers
+          }
+      outcomeVar <- newEmptyTMVarIO
+      atomically (stage tv point size peers blk outcomeVar)
+      -- TODO: Should not be necessary if fetching logic would
+      -- retry on the staging area query.
+      _ <- MVar.tryPutMVar readyMVar ()
+      outcome <- atomically $ takeTMVar outcomeVar
```
dnadales, line 323: If this client thread gets async-killed while parked on `takeTMVar` (peer teardown, mux timeout), the entry stays in the map with a dead outcome var. It's usually reaped later by the drain or GC, but if neither fires (the same windows as the staging/notification races flagged above) it lingers. Would wrapping the park in a `bracket` that deregisters the outcome var on exception make the cleanup unconditional?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
```diff
@@ -255,21 +273,21 @@ imQueryFetchWork stateVar = atomically $ do
   state <- readTVar stateVar
   let missingEbBodies =
         Map.fromList
-          [ (MkLeiosPoint (SlotNo (fromIntegral slot)) ebHash, ebBytesSize)
-          | (slot, (ebHash, ebBytesSize)) <- IntMap.toAscList (imEbPoints state)
-          , not $ Map.member ebHash (imEbBodies state)
+          [ (point, ebBytesSize)
+          | (point, ebBytesSize) <- Map.toList (imEbPoints state)
+          , not (Set.member point (imEbBodiesDownloaded state))
```
dnadales, line 278: Since a point's completeness is judged by EB hash elsewhere (`imQueryCompletedEbByPoint`), a second point that shares an already-downloaded hash still shows up here as a missing body and gets refetched, even though the content is already present. Not harmful (the insert is a no-op), but should this also skip points whose hash is already in `imEbBodies` to avoid the redundant fetch?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
```diff
@@ -56,90 +120,324 @@ data StagedCertRB peer blk = MkStagedCertRB
...
+
+    isStagedBlockSTM tv = do
+      staged <- readTVar tv
+      let points =
+            Set.fromList
+              [Block.blockPoint (stagedBlock (stagedRB e)) | e <- Map.elems staged]
+      pure (`Set.member` points)
+
+    stage tv point size peers blk outcomeVar =
+      modifyTVar tv $
```
dnadales, line 219: Should this be `modifyTVar'`? The map is created with `uncheckedNewTVarM` (no NoThunks check), and `mergeEntry` adds `Set.union` / `++` thunks on each re-stage. The strict fields in `StagedCertRB` force the unions when an entry is forced, so I don't think this is an actual leak, but a strict update plus forcing the merged value would keep the map thunk-free by construction. Minor.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
```diff
@@ -54,21 +54,28 @@ import Ouroboros.Consensus.Util.IOLike
...
-  , imAnnouncedCompletedEbs :: !(Set EbHash)
-  -- ^ EBs for which we have already broadcast an 'AcquiredEbTxs'
-  -- notification. The completion predicate stays true once an EB is
-  -- complete, so any later batch containing one of its txs would
-  -- re-trigger it without this guard; downstream consumers
-  -- (e.g. 'runLeiosVoting') treat the re-notification as fatal
-  -- ('AlreadyKnown' from 'addVote').
+  -- ^ EB tx-hash list, keyed by content hash. Shared across every
+  -- point that references the same EB.
+  , imCompletedEbs :: !(Set LeiosPoint)
```
dnadales, line 72: A question about the new per-point bookkeeping versus how completion is consumed: this PR keys completion per `LeiosPoint` (slot + hash), but `imQueryCompletedEbByPoint` (InMemory.hs:294-309) and the apply-time `resolveLeiosBlock` (Shelley `Ledger.hs:1029-1036`, which builds the point from `praosStateLastSlot` + the announcement hash) both resolve by hash only. Neither is touched by this PR, so this is just a consistency question, not a regression: now that the DB distinguishes points sharing a hash, is the hash-only resolve still the intended granularity, or should it eventually align with the per-point view?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: We deliberately put this here today with @bladyjoker and the test you spotted was a false positive of this condition (IMO).

> If a rollback or a peer disconnect drops the parent from every candidate in between

Can this happen? We are using the `ChainSyncClientHandleCollection` to find candidates. Do we know when headers get pruned from there?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: Good catch. This indeed could be a race condition. The `leiosDbQueryCompletedEbByPoint` is not an `STM` here because it is a DB query to sqlite in production (I recall you added a cache to make it so). An alternative could be to not drain while we do this checking. Is this needed before we can ship these fixes or a follow-up?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
ch1bo: I agree that the behavior (for both InMemory and SQLite) should be to resolve the closure immediately on body insert if we happen to have all txs. I hoped we have a test for that, but you seemingly checked those. Needs fixing or should this be a follow-up?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: This sounds plausible and changing the data structure to key on (praos) block header hash would be more correct and require us to search through the map on removal, but this should be fine. As you say this should be handled today -> follow-up issue?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: > could a transient candidate drop (rollback, brief KeepAlive gap) evict a still-valid entry

This depends on what the ChainSync logic does with such situations, no? We are merely working by the shared chain sync client state to determine whether some peer still has the block (header) we need.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: > by the drain or GC, but if neither fires

The GC will clean it up, not enough?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
ch1bo: I can't really follow.

> hash-only resolve still the intended granularity

Which resolution do you mean here?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: > Do we know when headers get pruned from there?

I went digging through the ChainSync client. Headers leave the candidate fragments in `ChainSyncClientHandleCollection` in three ways:

- **Forward progress (anchor side):** as our selection advances, the candidate stays anchored ~`k` behind the tip, so only *old* headers (≳`k` back) fall off. A recent header is safe here.
- **Rollback (tip side):** `MsgRollBackward` truncates the candidate via `AF.rollback`, dropping every header newer than the rollback point (`Client.hs:2074`, `setCandidate` at `:1615`). This drops *recent* headers, so a reorg that reaches back far enough can take out a CertRB's immediate parent.
- **Whole handle removed:** `deleteHandle` / `cschcRemoveHandle` runs on any exit of a peer's ChainSync client (`Client.hs:434`), so a disconnect, an exception, or a rollback-past-intersection termination wipes that peer's candidate entirely.

The header we need here is the CertRB's immediate parent (the announcer), one block behind the candidate tip. So forward progress never prunes it, which is the good news: in steady state the `error` won't fire. It only becomes reachable when a near-tip reorg rolls back through the parent, or the delivering peer disconnects, in the window between the fetch decision and `addBlockAsync` (BlockFetch and ChainSync are separate threads, and we re-read candidates at delivery, `:265-267`). It gets likelier with few peers, since fewer other candidates might still carry the parent.

So I think "by construction" holds on the happy path but not under reorg/disconnect. Since this used to blind-admit and let ChainSel decide, would returning `FailedToAddBlock` here be safer than `error`, so a recoverable race doesn't take down the client thread?

On the test: agreed it's a false positive of the scenario, but as written it asserts `SuccesfullyAddedBlock` for empty candidates, which now lands in this `Nothing` branch and throws, so it would fail at runtime (the build-and-test jobs were still pending). Whichever way we go on `error` vs admit, should we reconcile `test_gateIgnoresCurrentChain` with it (assert the new behaviour, or drop it)?

On whether this blocks the testnet: the trigger (a near-tip reorg or a peer disconnect in the fetch-to-deliver window) is normal on a small testnet, so it might be better to fix it before the testnet rather than defer, especially as the fix looks cheap (return `FailedToAddBlock` and let BlockFetch retry from a fresh candidate).

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: It's not a hard blocker, but we should prioritize fixing this, since when the bug hits we're looking at a full 120s stall. No idea what that would do to the network.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
dnadales: Tricky one. It's a different mechanism than the DB-keying bug the PR fixes (that part works), but it shows up in the same scenario, two CertRBs sharing an EbHash, with the same symptom: one of them never resolves. So even with per-point keying, the notification path can still leave a staged CertRB stuck. That's why I'd lean fixing it here rather than a follow-up.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: Agreed on a follow-up. The consequence is bounded today (the `AddBlockPromise` is `void`ed, so the false success isn't read, and the dropped block gets re-fetched and admitted later), and it only bites under equivocation or competing forks, so it doesn't need to block this PR.

One detail for whoever picks it up: keying by block point means the EB point is no longer the map key, so the entry needs to carry it (a `stagedEbPoint` field), and the drain scans for all entries matching the `AcquiredEbTxs` point and releases each. Nice side effect: that same change also removes the `gcOne` stale-snapshot mismatch, since the key then uniquely identifies the block whose peers we check.

Shall I open the issue?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: Yes, it does, and the answer is that ChainSync transiently drops the block on exactly the two cases you mention: a rollback truncates the candidate (`AF.rollback`), so the block leaves until the peer rolls forward again (seconds), and a disconnect removes the handle entirely until reconnect. So `peersThatKnowBlock` can briefly go empty for a still-valid block, and since `gcOne` evicts on a single observation it could act on that transient. But it's minor: a 120s tick rarely lands in that window, and re-fetch recovers the block anyway. So I'm happy to leave the eviction logic as is.

The one thing I'd still fix is the comment, separately: it says eviction lets BlockFetch "demote peer", but the `FailedToAddBlock` we return is `void`ed by the caller (`mkAddFetchedBlock_`), so the verdict is never read and no demotion happens. Eviction is really a silent drop that only re-fetch recovers. Could we reword it to match?

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: You're right, in practice GC or the drain reaps it: the entry is removed either when the closure arrives (drain) or when all peers drop the block (GC). The only case it lingers is if the closure never completes *and* some peer keeps the block on its candidate forever, and even then it's just one stale map entry, not a thread leak. So I'm fine leaving it. A `bracket` around the park would make the cleanup unconditional if we ever want it, but it's not worth it here.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
dnadales: Sorry, that was unclear. I meant the EB-closure lookups, `imQueryCompletedEbByPoint` (used by the staging gate) and the apply-time `resolveLeiosBlock`, which both look up the closure by EB hash, while the new bookkeeping (`imCompletedEbs`, `imEbBodiesDownloaded`) is per point. But thinking it through, hash-only is correct there: the closure is content-addressed, so once a hash is complete any point with that hash resolves, and the per-point tracking is only needed for the completion notifications. So there's no real inconsistency, ignore this one.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: @bladyjoker what do you think? You wanted the error of the "should not happen" case, which seems quite likely if candidates get pruned.

Re: the test suite.. it's already gone on the diff. But we can bring back a test that stimulates such chain sync behavior. The mentioned test being about the "current chain" was capturing this more by accident.

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: 120s is much less than forever.. which is the current behavior 😅

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo: How about a PR or commit on this branch? You seem to have a clear idea of what needs changing how exactly

nfrisby:
FYI @dnadales @ch1bo Claude and I pair-programmed what I had had in mind for the VolDB-based solution that replaces the StagingArea. You can find it here: https://github.com/IntersectMBO/ouroboros-consensus/pull/2076 (read PR description and the first PR comment for context, including what I think are the high-level comparisons to your recent PRs on this topic).

nfrisby (approved):
I'm approving, based on a conversation I just had with Sebastian.

- Having scanned this PR's diff, I didn't see anything that would degrade the current testnet.

- I did not review the comments on this PR from Damian and Sebastian. However, those comments are localized to the `LeiosStagingArea.hs` file.

- I just today proposed https://github.com/IntersectMBO/ouroboros-consensus/pull/2076 as a wholesale and long-term replacement for LeiosStagingArea. But I think we want to ship this PR---to improve the current testnet UX---without waiting on 2076 to pass review.

So I'm approving this PR for merge, in that context.

bladyjoker (approved):
We had a call about this PR yday, imo let's proceed with this!

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
dnadales: Sure, after the db-synth fixes.

ch1bo:
Synced the testnet with this, slowly, but it worked and I only needed to restart the node 2 times :shrug: .. let's merge this and look at https://github.com/IntersectMBO/ouroboros-consensus/pull/2076 next.

PR #2076, opened 2026-06-17
Out of order RB and EBs in ChainSel+VolDB

### LINKED ISSUE #890: Prototype: CertRB staging area and catch-up

nfrisby:
Main ideas:

- If a CertRB arrives before the EB closure that is certifies, add it to the VolDB, but don't let ChainSel properly process it.

- Whenever an EB closure arrives, ChainSel should reprocess CertRBs that were skipped because that EB was absent.

- Add the `has_Leios_cert` bit to the headers.

    - When `MsgRollForward` arrives in the ChainSync client with that bit set, interpret it (ie update the LeiosFetch state) as if that peer had just sent both `MsgLeiosBlockOffer` and `MsgLeiosBlockTxsOffer` for the EB that that header claims to certify.

    - When checking if a block matches its header, require the the bit is set in the header if and only if the body contains a cert and then also require zero txs if there's a cert.

---

I think the only desideratum that this PR loses is informing LeiosFetch that it should prioritize that EB because some CertRBs are stuck waiting for it. We could add that, but it seems more like an optimization, so it can at least be a separate PR. (Edit: A subsequent realization: that optimization might be effectively crucial in order for this to work well for _syncing_ nodes. Otherwise, we'll always be downloading the historical EBs within our forecast-range in freshest-first order, which is the opposite order a syncing node needs.)

Upsides:

- To prevent ChainSel from processing stuck CertRBs, I enriched the VolDB index to know whether each RB depends on its predecessor's EB. Thankfully, that didn't change any on-disk file formats, since the VolDB index is recreated from scratch on node start-up.

- The interpretation of `MsgRollForward` as EB "offers" will smoothly allow a syncing node to fetch EBs certified on the historical chain.

- The tentative header (aka "block diffusion pipelining") does The Right Thing automatically.

There are a few loose-ends still, but nothing inappropriate for this `leios-prototype` branch.


nfrisby:
Here's the PR's overview document, rendered: https://github.com/IntersectMBO/ouroboros-consensus/blob/nfrisby/leios-prototype-out-of-order-RB-EB/plan-1-for-out-of-order-RB-EB.md

The basic idea is that the VolDB is already a place where RBs wait when they arrive before their dependencies---this was @dnadales 's immediate push back on the StagingArea idea from the initial Leios design work, and I find it very convincing. So that's the main differentiator from @ch1bo's StagingArea impl (which was merged to unblock the early testnet UX).

The key differentiator from Damian's recent PRs is that I slightly enriched the VolDB index and then leveraged it as an existing component instead of adding additional/separately-maintained/ephemeral state to ChainSel. This does require _also_ adding the "announcer index" (a map from EbHash to the Points of RBs that announced that EB), but we're going to need that anyway.

For now the announcer index lives in and is sourced from the VolDB, to avoid scope-creep within this PR. But once `MsgLeiosBlockAnnouncement` is no longer degenerate, that seems like the more natural place to source that index. And I _think_ we'll find that voting also needs access to the announce index. *However*, once both ChainSel and LeiosVoting depend on the announcer index, I still think it'll be the case that we only need to know the announcers of an EB _if that announcer is in our VolDB_ (plus the imm tip as an extreme case). So maybe the announcer index impl in this PR is ultimately fine for the long-term (if we also add the simple corresponding `Maybe` to the ImmDB API or maybe access the imm tip's announcement via its HeaderState available in the LedgerDB?).

----

OK, that's my hand-written summary of the high-level ideas and concerns. You can find far more details in the `plan-1-for-out-of-order-RB-EB.md` document that I co-wrote with Claude. That logorrhea is potential impenetrable for mere humans, but it will at least hopefully help bootstrap your LLM agent.

nfrisby:
For what it's worth:

- I pushed a `cardano-node` commit that redirects its build (both cabal.project and flake) to use this PR's tip. See the `nfrisby/leios-prototype-out-of-order-RB-EB` branch there: https://github.com/IntersectMBO/cardano-node/compare/leios-prototype...nfrisby/leios-prototype-out-of-order-RB-EB

- The enables the following.

```
[nfrisby@tweag-nfrisby1:~/cardano-haskell/ouroboros-leios]$ git lg -1
f652e300b - (HEAD -> nfrisby/leios-prototype-out-of-order-RB-EB, origin/main, origin/HEAD) Improve release preparation workflow (6 days ago) <Sebastian Nagel>

[nfrisby@tweag-nfrisby1:~/cardano-haskell/ouroboros-leios]$ git diff
diff --git a/demo/proto-devnet/config/genesis/shelley-genesis.json b/demo/proto-devnet/config/genesis/shelley-genesis.json
index 448738d7d..41d5c84af 100644
--- a/demo/proto-devnet/config/genesis/shelley-genesis.json
+++ b/demo/proto-devnet/config/genesis/shelley-genesis.json
@@ -4,7 +4,7 @@
     "networkMagic": 164,
     "slotLength": 1,
     "activeSlotsCoeff": 5.0e-2,
-    "securityParam": 2160,
+    "securityParam": 6,
     "epochLength": 432000,
     "slotsPerKESPeriod": 129600,
     "maxKESEvolutions": 60,

```

(Edit: I'm pretty sure that genesis.json tweak didn't achieve anything: the final `db/immutable` has empty files :shrug: )

(Edit: Ah ha, I forgot you have to do it in the byron genesis.json file as well. If I do it in both, then I see the expected behavior: the `immutable/chunk` files grow quickly during the run.)

Then I ran the following

```
$ nix develop .#dev-demo-proto-devnet --accept-flake-config \
    --override-input cardano-node-leios github:IntersectMBO/cardano-node/1cb4650f81617e532f53cc2ed7efa7c6db65e9b9
$ cardano-node --version
cardano-node 11.0.1.164 - linux-x86_64 - ghc-9.6
git rev 1cb4650f81617e532f53cc2ed7efa7c6db65e9b9
$ (cd demo/proto-devnet; TC=0 XRAY=0 ./run.sh)
```

That put me in the TUI. I then joined a ~2hr meeting.

When I came back, the nodes were on BlockNo ~300 or something.

I used the TUI to stop and then start one, and it went fine.

Then I stopped one, deleted its on-disk files with `rm -rf tmp-devnet/node3/db tmp-devnet/node3/leios.db` and started it again. Eventually it recovered, and I saw they were in agreement again:

```
$ for n in 1 2 3; do CARDANO_NODE_SOCKET_PATH=tmp-devnet/node$n/node.socket     cardano-cli query tip --testnet-magic 164; done   # matching tip hashes
{
    "block": 340,
    "epoch": 0,
    "era": "Dijkstra",
    "hash": "2e7ba616bcb81820aed75c5c61afc9ce7f910a0468f24be02bddaecef1bd10b1",
    "slot": 7284,
    "slotInEpoch": 7284,
    "slotsToEpochEnd": 424716,
    "syncProgress": "100.00"
}{
    "block": 340,
    "epoch": 0,
    "era": "Dijkstra",
    "hash": "2e7ba616bcb81820aed75c5c61afc9ce7f910a0468f24be02bddaecef1bd10b1",
    "slot": 7284,
    "slotInEpoch": 7284,
    "slotsToEpochEnd": 424716,
    "syncProgress": "100.00"
}{
    "block": 340,
    "epoch": 0,
    "era": "Dijkstra",
    "hash": "2e7ba616bcb81820aed75c5c61afc9ce7f910a0468f24be02bddaecef1bd10b1",
    "slot": 7284,
    "slotInEpoch": 7284,
    "slotsToEpochEnd": 424716,
    "syncProgress": "100.00"
}
```

So that's pretty encouraging.

Edit: Once I noticed that all the data was in the VolDB, I suddely realized I had no confirmation that my tweak the cardano build worked. I had Claude superficially parse the VolDB data file checking how many fields each header therein had. The 12-field headers are produced by only my patched consensus, so this confirms it was at least running my path.

```
  Header-body CBOR field counts across all three nodes' VolDBs:

  ┌───────┬────────┬──────────┬──────────┬──────────┐
  │ node  │ blocks │ 10-field │ 11-field │ 12-field │
  ├───────┼────────┼──────────┼──────────┼──────────┤
  │ node1 │    367 │       31 │        0 │      335 │
  ├───────┼────────┼──────────┼──────────┼──────────┤
  │ node2 │    362 │       29 │        0 │      332 │
  ├───────┼────────┼──────────┼──────────┼──────────┤
  │ node3 │    346 │       29 │        0 │      316 │
  └───────┴────────┴──────────┴──────────┴──────────┘
```

Edit 2: I did it again, based on Sebastian's comment below. I took Node3 down, then watched ObserveTip until I say the mempool capacity plummet by several hundred kilobytes, waited for it to happen again, then brought Node3 backup, and soon all 3 nodes were on the same tip again.

ch1bo:
@nfrisby I also did the following:

- Start `proto-devnet`
- Have it produce a few EBs
- Stop one node (don't delete anything)
- Make sure one EB was announced and certified which the stopped node could have not seen
- Start the node again and see if it catches up

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Abstract.hs
```diff
@@ -130,6 +138,13 @@ class
     ShelleyProtocolHeader proto ->
     Except (EnvelopeCheckError proto) ()
 
+-- | The 'pHeaderLeiosContainsCert' for protocols/headers without Leios support:
+-- a header that cannot carry a Leios certificate never claims to. Instances
+-- that have no Leios support should define
+-- @'pHeaderLeiosContainsCert' = 'default_pHeaderLeiosContainsCert'@ explicitly.
+default_pHeaderLeiosContainsCert :: ShelleyProtocolHeader proto -> Bool
+default_pHeaderLeiosContainsCert = const False
```
ch1bo, line 146: Why not use this directly on the type class as a default implementation for the method?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -670,44 +698,85 @@ addReprocessLoEBlocks tracer ChainSelQueue{varChainSelQueue} = do
...
     Nothing -> do
+      -- The queue is empty: the runner is starved, even if it has an EB arrival
+      -- to react to (that carries no new chain work).
       startStarvationMeasure
-      msg <- atomically $ readTBQueue queue
-      terminateStarvationMeasure msg
-      pure msg
+      result <-
+        atomically $
+          (queuedMessage <$> readTBQueue queue) `orElse` ebArrival
```
ch1bo: This looks like a thread that drains a queue and then triggers chain sel accordingly.

Why is it special cased on Leios specifics? Couldn't we instead yield a message into the queue (upon `ebArrival`)?

I suppose this would get rid of some plumbing here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -614,44 +615,68 @@ addBlockRunner ::
...
+                          (FailedToAddBlock "Failed to add block synchronously")
+                      pure ()
+                    Left (ChainSelAddPerasCert _cert varProcessed) ->
+                      void $ tryPutTMVar varProcessed ()
+                  closeChainSelQueue cdbChainSelQueue
+              )
+              ( \(work, lastSeen') -> do
+                  case work of
+                    Right ebs ->
+                      reprocessLeiosEbs cdb ebs
```
ch1bo: Should: make this a message type and interpret it here like the rest

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -174,6 +176,14 @@ data VolatileDbArgs f m blk = VolatileDbArgs
...
   -- ^ Should the parser for the VolatileDB fail when it encounters a
   -- corrupt/invalid block?
+  , volBlockHasLeiosCert :: blk -> Bool
+  -- ^ Does a block carry a Leios certificate? Supplied by the caller (e.g.
+  -- ChainDB, via 'blockHasLeiosCert') so the VolatileDB stays agnostic of the
+  -- 'ResolveLeiosBlock' class. Defaults to @const False@ (see 'defaultArgs').
+  , volBlockAnnouncedEb :: blk -> StrictMaybe LeiosPoint
+  -- ^ The EB a block announces, if any. Supplied by the caller (via
+  -- 'headerLeiosAnnouncement'). Defaults to @const Nothing@. See the
+  -- TODO(EbAnnouncement) note on 'biLeiosAnnouncedEb'.
```
ch1bo: Should: Just use `blockHasLeiosCert` and `headerLeiosAnnouncement` from `ResolveLeiosBlock` (the latter you also introduce in this PR)

We can use the existing type class and get these from adding a `ResolveLeiosBlock blk` constraint. Less plumbing and I didn't see "value level" need of these functions (e.g. monkey patching)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/VolatileDB/API.hs
```diff
@@ -158,6 +169,27 @@ data BlockInfo blk = BlockInfo
...
+  -- ^ The endorser block this block announces, if any. Note this is /not/ a
+  -- bijection: many RBs may announce the same EB.
+  --
+  -- TODO(EbAnnouncement): unlike 'biHasLeiosCert', the announcing-RB →
+  -- announced-EB relation does not really belong in the VolatileDB. It is
+  -- authoritatively carried by the RB header (which 'MsgLeiosBlockAnnouncement'
+  -- is meant to diffuse). We track it here for now to drive the announcer
+  -- index; once the in-memory EbAnnouncement data flow exists, this field (and
+  -- the VolatileDB's announcer index) should be relocated there. It is kept
+  -- separate from 'biHasLeiosCert' precisely so it can be ripped out cleanly.
```
ch1bo, line 192: 👍

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -756,22 +756,6 @@ data TraceLeiosKernel
...
-    -- contained the block (they're treated as implicit offerers of the
-    -- EB by the fetch loop).
-    TraceLeiosCertRBStaged
-      { stagedBlockPoint :: String
-      , stagedEbPoint :: LeiosPoint
-      , stagedKnownPeers :: Int
-      }
-  | -- | A staged CertRB has been released back into ChainSel because
-    -- the EB closure (body + txs) is now locally available.
-    TraceLeiosCertRBReleased {releasedEbPoint :: LeiosPoint}
```
ch1bo, line 808: Do we have similar traces to find when a CertRB was ignored / reconsidered in ChainSel?

plan-1-for-out-of-order-RB-EB.md
ch1bo, line 1: TLDR, PR description should suffice

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -686,9 +769,23 @@ constructPreferableCandidates ::
...
-      <$> (ignoreInvalidSuc p invalid <$> VolatileDB.filterByPredecessor cdbVolatileDB)
-      <*> (ignoreInvalid p invalid <$> VolatileDB.getBlockInfo cdbVolatileDB)
+    -- Hide RBs whose Leios certificate references an EB whose closure has not
+    -- yet been acquired (issue #890's successor): they are parked in the
+    -- VolatileDB until the EB arrives, both as candidate tips (via
+    -- 'ignoreUnacquiredCertRB' on the lookup) and as successors (via
+    -- 'ignoreUnacquiredCertRBSuc'). This is a /transient/ exclusion (the EB may
+    -- arrive later and make the block selectable) — unlike the permanent
+    -- 'ignoreInvalid'/'ignoreInvalidSuc' exclusion of invalid blocks.
+    acquiredEbs <- forgetFingerprint <$> runGetAcquiredLeiosEbs cdbGetAcquiredLeiosEbs
```
ch1bo: Instead of needing to know which EBs we have (which will grow indefinitely), shouldn't we instead filter out EBs we don't have?

This is much more bounded operation (no matter if IO or STM)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -316,6 +328,77 @@ triggerChainSelectionAsync ::
 triggerChainSelectionAsync CDB{cdbTracer, cdbChainSelQueue} =
   addReprocessLoEBlocks (TraceAddBlockEvent >$< cdbTracer) cdbChainSelQueue
 
+-- | Reprocess the RBs that were parked because they certify an endorser block
+-- (EB) whose transaction closure had not yet been acquired, now that the
+-- closures of the given EBs have been acquired.
+--
+-- This is the Leios analogue of 'ChainSelReprocessLoEBlocks': just as a block
```
ch1bo: Nice, so there is a precedent to follow. What does LoE stand for?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -38,6 +40,15 @@ data LeiosDbHandle m = LeiosDbHandle
...
   , open :: m (LeiosDbConnection m)
   -- ^ Open a new connection to the LeiosDb.
+  , leiosDbGetAcquiredEbClosures :: HasCallStack => STM m (WithFingerprint (Set EbHash))
+  -- ^ The set of endorser blocks (by hash) whose tx closure is currently
+  -- available locally, tagged with a 'Fingerprint' that is bumped whenever a
+  -- new EB becomes available. ChainSel reads this (via
+  -- 'cdbGetAcquiredLeiosEbs') to avoid selecting an RB that certifies an EB
+  -- whose closure has not yet been acquired. An 'STM' action so the read is
+  -- observed transactionally and always reflects the current contents (both
+  -- acquisitions and, once closure GC exists, evictions); the fingerprint lets
```
ch1bo: See my comment above: while this will GC never certified EBs, the set will still grow indefinitely. Instead of keeping such a growing index in memory, I'd prefer to filter out candidates using the ground truth (from disk in the case of sqlite, potentially paged in via its own indexing).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -782,6 +782,15 @@ data ConfigEnv m blk = ConfigEnv
   , chainDbView :: ChainDbView m blk
   , getDiffusionPipeliningSupport ::
       DiffusionPipeliningSupport
+  , leiosCertRbCallback ::
+      Header blk -> ChainDepState (BlockProtocol blk) -> m ()
```
ch1bo, line 786: Not too bad and could be packed up quite similar like these future and historicity checks?

In the haddock, ideally, we'd not point to the praos specific accessors, but the `blk` polymorphic variants on the type class (currently called `ResolveLeiosBlock`)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -617,22 +506,6 @@ initNodeKernel
...
-        let drainTr = leiosKernelTracer tracers
-        chan <- subscribeEbNotifications leiosDB
-        runStagingAreaDrain leiosCertRbStaging chan $ \point blk -> do
-          traceWith drainTr TraceLeiosCertRBReleased{releasedEbPoint = point}
-          _ <-
-            ChainDB.addBlockAsync
-              chainDB
-              InvalidBlockPunishment.noPunishment
-              blk
-          pure ()
```
ch1bo: I'm sorry for the rebase conflicts you will get from this, but your commits were less confidently named so I merged the staging area fixes and this will conflict at least here :)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -559,6 +567,56 @@ mkHandlers
...
+            -- they are registered (ChainSync and LeiosNotify race at connection
+            -- startup, so we wait rather than drop the offer). This does not
+            -- block indefinitely: it short-circuits ('Nothing') if the
+            -- ChainSync client is being torn down (its control message is
+            -- 'Terminate'), which is exactly when the peer's LeiosNotify client
+            -- has (or, once peers are removed from the map on disconnect, will
+            -- have) unregistered. So we neither drop the offer while the peer is
+            -- merely still initialising, nor hang once it is already gone.
+            mbPeerVars <- atomically $ do
+              peersVars <- LazySTM.readTVar getLeiosPeersVars
```
ch1bo: Should: move this function next to the fetching logic

It manipulates the sate of the fetching logic "component" and has less to do with the chain sync clients than with the fetch logic.

ch1bo (commented):
I like it!

I think it's fine to not have the distinction between synthesized and real offers right now - it was not leveraged in the other design and maybe YAGNI. The ChainSel changes are not too bad, but I think the plumbing can be made even more lean.

My biggest concern is the EBHash cache and my comment of instead filtering out incomplete closures might change the communication paths. What do you think?

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
```diff
@@ -113,6 +113,14 @@ data HeaderBody crypto = HeaderBody
   -- ^ Optional Leios endorser-block announcement (Dijkstra-only;
   -- 'SNothing' on earlier eras). Placed on the Praos header for
   -- early-diffusion of EB references before the body arrives.
+  , hbLeiosContainsCert :: !Bool
```
bladyjoker, line 116: Generally I'm not in favor of the current way we model the header and the body as elaborated in https://github.com/input-output-hk/ouroboros-leios/issues/932.

`hbLeiosContainsCert` is a bolted on flag whereas it could and imo should be under an informative `hbBodyInfo` type that models what we expect Body to be.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Abstract.hs
nfrisby: Merely because I prefer to be explicit rather than implicit in the codebase. If I'm looking at an instance, I want to see everything there is to consider without having to cross-reference/remember everything about the class.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
nfrisby: Ah, I will try to reintroduce the tracers this PR is currently removing.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
nfrisby: Great point. Unfortunately, both sets could grow indefinitely, due to adversarial EBs being withheld. I'll think on this today---there will be some leverageable fact about the immutable tip, I think.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
nfrisby: It's from the Ouroboros Genesis implementation: "Limit on Eagerness".

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
nfrisby: I wasn't expecting this PR to forestall the StagingArea fixes.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -38,6 +40,15 @@ data LeiosDbHandle m = LeiosDbHandle
   -- TODO: make return type more descriptive (e.g. Subscription { getNext :: STM m LeiosEbNotification })
   , open :: m (LeiosDbConnection m)
   -- ^ Open a new connection to the LeiosDb.
+  , leiosDbGetAcquiredEbClosures :: HasCallStack => STM m (WithFingerprint (Set EbHash))
```
bladyjoker: Why use Fingerprint when we have established to use notify/subscribe? Why not just subscribe to completed AcquiredEbTxs plus `leiosDbGetCompletedEbIds`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
nfrisby: The semantics of `leiosDbGetAcquiredEbClosures :: STM m (Set EbHash)` should be:

> the set of EBs that are available and announced by an RB no older than the immutable tip.

That is a bounded set, and it's all the EBs that ChainSel might ever need.

- Recall that, for a syncing node, the immutable tip is very old.

- Recall that the node does apply some RBs from the immutable chain when it starts up. But those RBs can only be immutable now if we've selected them previously (or if the SPO did some shenanigans with the node's on-disk state), which means their EB is definitely available (because we don't GC EBs that are certified by an immutable block).

The current comment says:

> An 'STM' action so the read is observed transactionally and always reflects the current contents (both acquisitions and, once closure GC exists, evictions).

"once closure GC exists" is the key phrase: the prototype doesn't yet trim out EBs that we acquired but didn't end up on the immutable chain. Once the prototype is discarding those when appropriate (ie when they're announced by an RB older than the immutable tip), then it should also be updating this `leiosDbGetAcquiredEbClosures` variable upon GC.

plan-1-for-out-of-order-RB-EB.md
nfrisby: Yeah, this is primarily an artifact of how I was working with Claude. I thought it might be interesting for reviewers, but I have definitely been leaning towards annihilating it before merge.

ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Header.hs
nfrisby: As long as we get the CDDL right, potentially needing to refactor the haskell data types later doesn't worry me.

So far, all the Leios fields we've added are indeed orthogonal (edit: with themselves): all four combinations of SNothing/SJust + True/False are validation combinations.

You 932 Issue is specifically about the fact that `hbLeiosContainsCert` is (in some sense) redundant with `hbBodySize`, yeah? So my diff here adds a "junk" value to this type where `hbLeiosContainsCert = True` and `hbBodySize /=` this epoch's cert size.

Generally, I'm eager to avoid junk types. However, this particular rule doesn't seem to me like something the codec should be responsible for enforcing. And we definitely have a natural place to enforce it: header validation.
Morever, it's still a very open case whether we would eventually want to permit txs to also be in a CertRB (semantically _after_ the cert).

So: keeping the header data type simple (just a product of flat values), not breaking existing code (which is able to read the size off the header, when it's trusted---eg in a data node), keeping CDDL/codecs "simple", and not painting ourselves into a corner (ie requiring more codec changes in the future if we do allow some txs in a CertRB) all suggest to me that `hbBodySize` + `hbLeiosContainsCert` seems good.

nfrisby: I replied similarly on 932. Thanks for linking to it from here.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
nfrisby: Thanks for catching that. I'll close this thread an reply to your comment above.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
nfrisby: I'll reply to this in your message above https://github.com/IntersectMBO/ouroboros-consensus/pull/2076#discussion_r3440849575

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
nfrisby: I'll reply to this in Sebastian's message above: https://github.com/IntersectMBO/ouroboros-consensus/pull/2076#discussion_r3440849575

nfrisby: So please look for a response there. But be advised: I don't know what you mean by `leiosDbGetCompletedEbIds` (I'm assuming you meant adding that method to LeiosDB).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
nfrisby: I hope to collect any further discussion here into your above message: https://github.com/IntersectMBO/ouroboros-consensus/pull/2076#discussion_r3440849575

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
nfrisby: Nice yeah. Will be included in my next push.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: You and Drazen both pounced on this, and it was related to your observation about unbounded size. I replied about the size here https://github.com/IntersectMBO/ouroboros-consensus/pull/2076#discussion_r3441396099, but I've since spent the day refactoring towards the impl both you and Drazen were expecting (with consideration of the bounded size).

### The ChainDB must control LeiosDB promotion/garbage collection.

The LeiosDB does not know whether a particular EB is certified _on the current selection_. Therefore, it cannot know which EBs need to be promoted to immutable storage. The ChainDB owns the current selection, and therefore it must be the component to decided which EBs to immutalise.

Similarly, the ChainDB owns the immutable tip. So either it calls LeiosDB when the immutable tip advances, or else something monitors the immutable tip via the ChainDB API and then notifies LeiosDB when it changes.

Why have the middle man? Hence this PR has the ChainDB call LeiosDB methods when appropriate for GC / immutalization.

I did that _in this PR_ because it establishes context for the details more directly relevant to this PR. But to avoid scope-creep, I left those LeiosDB methods as stubs: it still not doing any GC. But ChainDB is already telling it when to do the GC-related work.

### Should the ChainDB or the LeiosDB own the acquiredLeiosEbs set?

Sebastian regularly expresses a preference for using the LeiosDB as a single source of truth, where possible. The option most directly compatible with that would be for every ChainSel invocation to query the LeiosDB for the set of EBs that are available. There are two ways:

- LeiosDB answers that query by querying the on-disk data.
- The LeiosDB incrementally maintains an in-memory cache of this query (updated whenever an EB closure arrives). --- That cache is a second source of truth, but at least its localized.

I have been and still am very concerned about the latency required to go to disk for LeiosDB queries. ChainSel is in particular a place we don't want to add unnecessary latency. This is the primary reason I prefer an in-memory TVar for this ChainSel-query rather than a SQL query.

A secondary reason is race conditions. It'd be problematic if the SQL query informed ChainSel that some EB is acquired (and so its CertRBs can now be processed) only for that EB to be garbage collected before ChainSel actually fetched it. We could maybe add synchronization to prevent that, but that'd make LeiosDB GC and ChainSel mutex, which seems undesirable: neither should be starved---especially not ChainSel. The existing ChainDB precedent for this is to delete some data from the in-memory index that ChainSel reads (ie VolDB) at least 60s before that data is actually deleted from disk. We can adopt that same solution only if we maintain the index separately from the on-disk store---and once the index exists separate of the backing store, why shouldn't it live in-memory?

(ChainSel needs to know the whole set, not just new additions, because if an EB arrives before a CertRB, ChainSel needs to be able to handle that too.)

-----

Therefore, this PR adds a TVar for monitoring the acquiredLeiosEbs set. I had originally let LeiosDB own it (as in the second bullet above), but here's where the size concern comes into play. This TVar needs to be garbage collected and _well-before_ the blocks it indexes get garbage collected. As discussed above, that requires some synchronization of LeiosDB and ChainDB.

The TVar either lives in ChainDB or LeiosDB. Neither is obviously much tidier than the other:

- If LeiosDB owns it, the timing requires another hook for ChainDB to call, and we need an adhoc newtype wrapper to workaround QuantifiedConstraints not handling `io-sim`'s `STM` family :/.

- If ChainDB owns it, then we need an new background thread to subscribe to LeiosDB's new EB arrivals and insert into the TVar when its new.

I opted for ChainDB owning it in the commits I'm about to push. The new thread can also enqueue a message for into the existing addBlockRunner queue to wake-up ChainSel when an EB arrives. Which is what you were both expecting to see.

The key advantage of having a new message in the existing queue is: we get natural FIFO ordering of EB arrivals and the other messages in that queue, instead of my original PR having to favor one over the other (or do weird things like alternate, etc... FIFO is best).

---

If LeiosDB owned the TVar, then we wouldn't need a ChainDB thread to subscribe to EB arrivals; the LeiosDB hook itself could just update the TVar. But I think we'd _still_ need something to enqueue message into the addBlockRunner queue, so that we get FIFO interleaving and _fair_ back pressure among all the queue writers.

Speaking of back pressure, it is the case that this queue might be full, meaning the new thread will block. That's OK, though. The `TChan` will accumulate messages without blocking LeiosDB. That sounds like a space leak, but Leios only downloads a bounded number of EBs per fixed interval, so this TChan is actually bounded. If ChainSel is so overwhelmed that even that rate is too great and this TChan uses too much memory, then our node has been defeated---ChainSel that slow already violates Praos requirements.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
nfrisby: Agreed, thanks.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Abstract.hs
ch1bo: Okay. I like explicitness .. so we should follow this approach then for all methods?

plan-1-for-out-of-order-RB-EB.md
ch1bo: So I just gave you confirmation that I, as a reviewer, did not find it interesting enough to try to read it :D

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
ch1bo: Okay, if it's only about the "volatile EBs" then it should be fine to do it this way around. Let's continue the discussion in the other thread.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/Abstract.hs
nfrisby: I try to when I see it. Did I miss one in this PR? I'll re-review.

(I added methods to ResolveLeiosBlock with defaults, but that's merely to follow precedent of that class---I don't want _this PR_ to refactor the existing methods and I don't want a single class to have mixed idiom. Let's just fix this up when the eventual ResolveLeiosBlock refactor happens (as its own PR))

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
nfrisby: Thinking it through this morning: I don't think we can easily add these tracers. And that makes sense: part of the reason to replace the StagingArea with the VolDB is because that lets us reuse a more fundamental component whose existing control/data flow is sufficient for our needs after just a couple tweaks: track hasLeiosCert bit in VolDB index and let ChainSel read the mutable set of which EBs are locally available. It is unsurprising that the existing ChainSel/VolDB tracers don't provided a specific/thorough tracing as a bespoke component they can supplant.

We could enrich the ChainSel/VolDB tracers, but that would require adding _state_ that ChainSel/VolDB don't already maintain: whether some RB has _ever_ participated in any "plausible" ChainSel candidate. The only existing tracer that's very similar is `StoreButDontChange p`: when an RB first arrives and is not _immediately_ participating in plausible candidate. What's missing is that ChainSel/VolDB doesn't track the RBs that emitted `StoreButDontChange p`, so it can't subsequently emit _another_ tracer when they're either used (ie `CertRBReleased`) or die before ever being used (ie `CertRBEvicted`).

The increment I think we could make here would be to add some extra info to/alongside `StoreButDontChange p`, to distinguish the reasons why the RB didn't immediately have any plausible candidates (1 its predecessor chain is missing some RBs, 2 it's missing the EB it confirms, or 3 the chain it's on isn't longer than our selection) (or more than one of those three at a time).

At least then, a stateful analysis of the node log could see `StoreButDontChange` with the additional info possibly indicating that's due to a missing EB (which is `CertRbStaged`, if so). And then it could track subsequent tracers to see when that EB was acquired (which would roughly be `CertRBReleased`). If it's monitoring tracers that indicate the immtip, then it could also infer `CertRBEvicted` for that CertRB if it hadn't already seen the EB arrive.

---

I wrote the above as a summary after investigating each event and writing the below:

- `TraceLeiosCertRBStaged`: whenever ChainSel attempts to process a block but finds no eligible candidates for it, it emits `traceWith addBlockTracer $ StoreButDontChange p`. There's two reasons for that to happen: it's on a chain that isn't longer than the current selection or the RB can't be reached from the immutable tip (eg it arrived before some of its predecessors). This PR is implementing "staging" by categorizing it into that second category. Perhaps we could enrich the `StoreButDontChange` tracer with an additional flag. Alternatively, we could complicate the `constructPreferableCandidates` logic to detect when it _would_ have had a candidate were it not for EBs. Or we could leave it alone and know that on the testnet,`StoreButDontChange p` strongly suggests a "parked RB" on the Leios testnet when `p` is known to be a CertRB. Not explicit enough?

- `TraceLeiosCertRBReleased`: As explained for `TraceLeiosCertRBStaged`, ChainSel emits the generic `StoreButDontChange` for any RB that doesn't immediately yield any candidates. But it doesn't signal anything interesting _the first time_ that RB later becomes activated. Doing so would require additional state keeping track of which RBs were `StoreButDontChange`d when they arrived and then were later able to participate in some candidates. Seems undesirable for ChainSel itself to do that tracking when it doesn't actually need to (ie only for more convenient tracing).

- `TraceLeiosCertRBEvicted`: Similar to ``TraceLeiosCertRBReleased`, ChainSel itself doesn't keep the necessary state to know whether the CertRB was GC'd before it participated in any ChainSel candidates. The only indicator we currently have is: did this CertRB eventually end up in the ImmDB?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: > A secondary reason is race conditions. It'd be problematic if 

I'm having second thoughts on this. With or without the race condition, the code that actually validates the CertRB is going to have to have a branch that handles the situation where the EB isn't found.

- The StagingArea/this PR/etc merely ensures is that that branch won't be reached _before the EB arrives_.
- But that means if that branch _is reached_ then the EB was deleted. And (assuming no bugs), that means this code happened to hit the rare situation where the EB was present when this ChainSel invocation started but is now absent.
- If we were to _also_ make that impossible (eg via the hypothetical mutex between ChainSel and LeiosDB GC), then this branch could be an `error` call.
- But: `error` calls are a smell. So one alternative to be to let the branch actually be reachable, and consider the CertRB itself a deadend (but not necessarily invalid). It'd be similar to the `ValidateExceededRollBack` branch... which is currently an `error` but maybe could instead return a `ValidPrefix` (ending with the EB's announcer) instead of `FullyValid`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
nfrisby: Nice. Just pushed up a commit to do this.

ch1bo (approved):
Discussed the changes today and I'm fine with moving into this direction.

nfrisby:
I did a few more `demo/proto-devnet` tests and it worked except once. On that one failure, there was one EB that the node didn't fetch at all---I didn't figure out why and then it hasn't happened again.

My plan:

- fix up this commit history --- just squash it for the most part, but potentially find a nice couple commit boundaries (eg add header bit in its own, early commit)
- merge
- try again to repro the failure (I'm wondering if it's related to a race condition in LeiosDB/LeiosFetch logic, more so than this PR's content)

nfrisby:
I just now force pushed after having Claude rearrange the diff into 7 nicely sequenced commits that each builds. But I confirmed that `git diff` between the before-and-after state of the PR was ∅ (except for also removing the `plan-1...md` document, which was only present for the sake of reviewers).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -579,6 +579,29 @@ class ResolveLeiosBlock blk where
...
   headerLeiosAnnouncement _ = Nothing
 
+  -- | Whether this header records that the block's body carries a Leios
+  -- certificate (a CertRB; the @hbLeiosContainsCert@ bit). 'False' for
+  -- headers in eras without Leios support. Lets a peer recognise a CertRB
+  -- from its header alone — e.g. the ChainSync client uses it to synthesise
+  -- a LeiosFetch offer for the EB the CertRB certifies (announced by its
+  -- predecessor).
+  headerContainsLeiosCert :: Header blk -> Bool
+  headerContainsLeiosCert _ = False
```
nfrisby, line 589: As discussed here https://github.com/IntersectMBO/ouroboros-consensus/pull/2076/#discussion_r3450694952, I prefer to avoid type class method defaults---implicit meaning in instances trips people up.

I don't want my PR to undo the precedent for this specific class, so _these methods_ follow the precedent

PR #2077, opened 2026-06-17
Honor ExperimentalHardForksEnabled in db-synthesizer config parser

### LINKED ISSUE #2075: db-synthesizer forges 0 blocks in the Dijkstra era
## Summary

`db-synthesizer` produces 0 blocks when run against a chain in the Dijkstra era.
It does not fail; it reports `ForgeResult {resultForged = 0}` after processing a full epoch.
A live node started from the same genesis (single bootstrap pool) forges in Dijkstra without trouble.

## Environment

- Binary: `ouroboros-consensus-exe-db-synthesizer-3.0.1.0`.
- Built from the `cardano-node-leios` pin in `ouroboros-leios` `prototype-2026w24`, commit `intersectmbo/cardano-node@f924314225ee4a6f4130c399f6ae19098d54ba23`.
- Chain state: created in Conway, then forked to Dijkstra.

## Steps to reproduce

1. Take the attached `workbench` state (chain in the Dijkstra era).
2. Run:

   db-synthesizer \
     --config workbench/custom/rundir/node-config.json \
     --db workbench/custom/rundir/db \
     --bulk-credentials-file workbench/custom/bulk.creds.all.json \
     -a -e 1

## Expected

The synthesizer forges blocks for the slots where the configured forger is leader, the same way a live node does with this genesis.

## Actual

It forges nothing across the whole epoch:

    --> forger count: 1
    --> opening ChainDB on file system with mode: OpenAppend
    --> starting at: SlotNo 0
    --> epoch size: EpochSize 86400
    --> will process until: ForgeLimitEpoch 1
    --> forged and adopted 0 blocks; reached SlotNo 86400
    --> done; result: ForgeResult {resultForged = 0}

## Where to look

The fix belongs in this repo: the synthesizer source is `ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/`.
The `db-synthesizer` in `cardano-node` builds the same code through the pin above.

The per-slot loop in `runForge` exits early on the first failing step and discards the reason.
See `ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs:132`, where `isRight` drops the `Left` message from `goSlot`.
Surfacing that reason (a tracer or a print) should localize the cause to one of the early-exit points in `goSlot` (`Forging.hs:151`):

- `"no ledger view"` (`Forging.hs:179`): the Dijkstra ledger-view forecast fails.
- `"NoLeader"` (`Forging.hs:199`): the leadership check never elects the forger for Dijkstra.
- `"block not adopted"` (`Forging.hs:245`): the forged block is rejected by the ChainDB.

These are hypotheses, not confirmed; I have not run the reproduction yet.

One observation from the attached state: the immutable db is empty (`db/immutable/00000.chunk` is 0 bytes), and the forged Dijkstra chain sits in the volatile store (`db/volatile/blocks-0.dat`, `blocks-1.dat`).
The ledger dir holds no snapshot, so the synthesizer replays from genesis.
On open, the ChainDB tip resolves to `Origin`, which is why the run prints `starting at: SlotNo 0` even though volatile holds a chain.
Whether the ChainDB selects those volatile blocks on open is worth checking.

## Impact

Blocks the Leios public testnet.
Current workaround is a live node at 100x speed, which is slower than the synthesizer and loses some block forges.

## Attachments

`workbench.tgz` (321 KB), a self-contained repro. Tarball root is `workbench/`, so paths match the command above. Contents:

- `custom/rundir/node-config.json` (the `--config`).
- `custom/bulk.creds.all.json` (the `--bulk-credentials-file`).
- `custom/rundir/db/` (the `--db` ChainDB: empty `immutable`, the chain in `volatile`, no `ledger` snapshot).
- The five genesis files referenced by the config, by relative name, alongside it: `byron-`, `shelley-`, `alonzo-`, `conway-`, `dijkstra-genesis.json`.

dnadales:
`db-synthesizer` read the obsolete `TestEnableDevelopmentHardForkEras` key, which `cardano-node` renamed to `ExperimentalHardForksEnabled`. On a config that uses the current key, the development-eras flag defaulted to `False`. That capped the node's max protocol version at 11 while the same config scheduled Dijkstra (protocol version 12) at epoch 0, so every block was rejected as `ObsoleteNode` and db-synthesizer forged nothing.

Now we parse `ExperimentalHardForksEnabled`, keeping the old key as a deprecation guard that fails on inconsistent use, matching `cardano-node`'s POM.hs. Rename the vendored field `npcTestEnableDevelopmentHardForkEras` to `npcExperimentalHardForksEnabled`.

Fixes #2075

### DIFF
--- a/changelog.d/20260617_151209_damian.nadales_db_synth_experimental_hardforks.md
+++ b/changelog.d/20260617_151209_damian.nadales_db_synth_experimental_hardforks.md
@@ -1,0 +1,29 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+### Patch
+
+- `db-synthesizer`: honor the `ExperimentalHardForksEnabled` configuration key.
+  The parser previously read only the obsolete `TestEnableDevelopmentHardForkEras`
+  key (renamed in `cardano-node`), so configs using the current key left the
+  development-eras flag at its `False` default. That capped the node's max
+  protocol version below the era the same config scheduled, so every block was
+  rejected with `ObsoleteNode` and `db-synthesizer` forged nothing. The vendored
+  `NodeHardForkProtocolConfiguration` field is renamed to
+  `npcExperimentalHardForksEnabled` to match `cardano-node`.
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Node/Protocol/Cardano.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Node/Protocol/Cardano.hs
@@ -88,7 +88,7 @@ mkConsensusProtocolCardano
     }
   npcDijkstraProtocolConfig
   NodeHardForkProtocolConfiguration
-    { npcTestEnableDevelopmentHardForkEras
+    { npcExperimentalHardForksEnabled
     , -- During testing of the latest unreleased era, we conditionally
     -- declared that we knew about it. We do so only when a config option
     -- for testing development/unstable eras is used. This lets us include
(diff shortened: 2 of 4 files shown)


bladyjoker (approved):

ch1bo (approved):
Merge and ship it!

PR #2079, opened 2026-06-18
Remove the `mk` parameter from `LedgerState`

jasagredo:
TL;DR: there is no HF TxIn, so we can throw away all the HF machinery. Also we have era-indexed NS of tables, instead of Table of era-indexed NS of values.

UTxO-HD indexes `LedgerState` (and `Ticked`, `ExtLedgerState`, the hard-fork telescope, ...) by a `MapKind` to thread the on-disk table data around at the type level. The index turns out to earn its keep in very few places: mostly it's bookkeeping that every signature has to carry, plus the plumbing it drags in (the `Flip`/`unFlip` dance to use an `mk`-indexed state as an SOP functor, the `EmptyMK`/`ValuesMK`/`DiffMK` shuffling, the canonical key encoding in the HFC).

This branch drops the parameter. `LedgerState blk` goes back to being a plain `Type`. The table payloads become opaque, per-block associated types (`Keys`/`Values`/`Diff blk`) on a single class, `BlockSupportsUTxOHD`, threaded explicitly through the few operations that actually read or apply them instead of riding along inside the state.

A few things this buys us:

- `applyBlock` is pure again. It takes the values read for the block and the ticked state, and returns `(state, diff)`. Reading and applying are no longer entangled through a `*MK`-indexed state.
- The HFC stops needing a canonical key/value encoding across eras. `Keys`/`Values`/`Diff (HardForkBlock xs)` are just `NS Wrap… xs`, so the era tag rides on the payload and there's no `TxIn`/`TxOut (HardForkBlock xs)` at all. Storage and the LSM backend dispatch per-era off that tag.

Removed along the way:

- `Tables/{MapKind,Combinators,Utils}`, the `LedgerTables` newtype, and the `EmptyMK`/`ValuesMK`/`DiffMK`/`KeysMK` kinds.
- `HasLedgerTables`, `CanStowLedgerTables`, `CanUpgradeLedgerTables`, `SerializeTablesWithHint`, `LedgerTablesAreTrivial`.
- `CanonicalTxIn`, `HasHardForkTxOut`, and the era-sum `HardForkTxOut`.
- `IndexedMemPack`: the LSM backend uses plain, forwards-deserialisable `MemPack`, so era-indexed packing is no longer needed.
- The `unsafeCoerce`-based `coerceSet`/`coerceMapKeys`. Shelley keys are plain `SL.TxIn`; `BigEndianTxIn` is applied only at (de)serialisation, where the byte order has to agree with `Ord`.

Trade-off worth calling out: some of the totality the type index gave us is now a runtime check (a handful of `error "… impossible"` arms where an `NS` match can't fail by construction). In exchange the type surface across consensus, the HFC, storage and the era instances is a good deal smaller and flatter.


jasagredo:
Linting/fourmolu will presumably still fail because the integration branch is not yet rebased on top of #2091.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -355,25 +392,44 @@ instance MemPack BigEndianTxIn where
...
+
+  blockKeys = Core.neededTxInsForBlock . shelleyBlockRaw
+
+  -- One era ⇒ no translation. Replay the in-flight diffs in chain order; the
+  -- 'Diff' 'Monoid' composes them so later ones win.
+  forward diffs vals = Diff.applyDiff vals (mconcat diffs)
+
+  restrictValues keys vals = vals `Map.restrictKeys` keys
+
+  valuesSize = Map.size
```
bladyjoker, line 423: Minor naming suggestions, from the rhs hints :)

forward -> applyDiff?

restrictValues -> restrictKeys as it's a common Map operation?

valuesSize? why not keysSize? and define it on Keys rather than Values?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -559,41 +531,37 @@ instance ShelleyBasedEra era => IsLedger LedgerState (ShelleyBlock proto era) wh
...
-                  shelleyLedgerTransition
-          , tickedShelleyLedgerState = l'
-          , -- The UTxO set is only mutated by block/transaction execution and
-            -- era translations, that is why we put empty tables here.
-            tickedShelleyLedgerTables = emptyLedgerTables
-          , tickedShelleyLedgerLatestPerasCertRound =
-              shelleyLedgerLatestPerasCertRound
-          }
+      let (tickedNoUTxO, events) =
+            applyTickShim evs globals slotNo shelleyLedgerStateNoUTxO
```
bladyjoker, line 539: Is it time to rename this to shelleyLedgerEpochState or smtn? 

It misreads hard when part has the same name as the whole.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/EnvelopeChecks.hs
bladyjoker, line 62: BodySizeTooLarge rather?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToClient.hs
bladyjoker, line 393: well `>$<` to you too

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
bladyjoker, line 1: Why not call this module `LedgerCalls`? and strip the `Shim` suffix?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
```diff
@@ -0,0 +1,262 @@
...
+
+-- | Split a 'SL.NewEpochState' that /does/ carry a UTxO (genesis, or a snapshot
+-- with a populated field) into the UTxO-free state and the extracted entries.
+-- Unlike 'mkNewEpochStateNoUTxOs' this does not assert the field is empty; it
+-- is precisely the path that pulls a populated UTxO out into the backend.
+splitUTxO ::
+  Core.EraTxOut era =>
+  SL.NewEpochState era ->
+  (NewEpochStateNoUTxOs era, Map SL.TxIn (Core.TxOut era))
+splitUTxO nes =
```
bladyjoker, line 129: Isn't this `unstow`? `curry stowUTxO (unstowUTxO nes) == nes` and `unstowUTxO (stowUTxO (nes, utxos)) = (nes, utxos)`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
```diff
@@ -0,0 +1,262 @@
...
+-- | Extract the resulting UTxO from a NES produced by a ledger computation,
+-- clear the field (re-establishing the invariant), and compute the diff of the
+-- result against the values that were stowed in.
+clearUTxOWithDiff ::
+  Core.EraTxOut era =>
+  -- | The values that were read in (the pre-image).
+  Map SL.TxIn (Core.TxOut era) ->
+  SL.NewEpochState era ->
+  (NewEpochStateNoUTxOs era, Diff.Diff SL.TxIn (Core.TxOut era))
+clearUTxOWithDiff values nes =
```
bladyjoker, line 152: Why not

unstowUTxOWithDiff values = let (nes, utxo) = unstowUTxO in (nes, Diff.diff values utxo)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -671,9 +665,10 @@ forkBlockForging IS{..} (MkBlockForging blockForgingM) =
           -- At this point we have established that we are indeed slot leader
           trace blockForging $ TraceNodeIsLeader currentSlot
 
-          -- Tick the ledger state for the 'SlotNo' we're producing a block for
-          let tickedLedgerState :: Ticked LedgerState blk DiffMK
-              tickedLedgerState =
+          -- Tick the ledger state for the 'SlotNo' we're producing a block for.
+          -- Ticking also yields the tick diff, which the mempool snapshot
+          -- below needs to forward the read values to the ticked state.
+          let (tickedLedgerState, tickDiff) =
```
bladyjoker, line 676: Can you help me understand tickDiff and how it works with Mempool snapshotting?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -251,3 +292,158 @@ instance StandardHash blk => StandardHash (LedgerState blk)
...
+--   * Shelley carries the real instance: @Keys = Set TxIn@,
+--     @Values = Map TxIn TxOut@, @Diff = Diff TxIn TxOut@.
+--
+--   * Byron\/mock blocks have no on-disk tables, so their @Keys@\/@Values@\/
+--     @Diff@ are trivial (@Void@\/@()@).
+--
+--   * The hard-fork combinator uses era-tagged @NS@ payloads; 'forward'
+--     translates values across a rare era boundary using the per-era
+--     translations carried on the 'CanHardFork' class (no config needed).
+type BlockSupportsUTxOHD :: Type -> Constraint
```
bladyjoker, line 319: Why are we naming this `UTxOHD`? What happens when we add non UTxO values? Why not just BlockSupportsStorage or smtn?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -251,3 +292,158 @@ instance StandardHash blk => StandardHash (LedgerState blk)
...
+-- opaque @'Keys'@\/@'Values'@\/@'Diff'@ into individual @('TxIn', 'TxOut')@
+-- rows. The 'Ouroboros.Consensus.HardFork.Combinator.Basics.HardForkBlock' has
+-- /no/ instance: it answers such requests by dispatching to the current era
+-- (via @matchNS@, see the LSM backend's @BlockSupportsLSM@) and working there,
+-- so there is no single @TxIn@\/@TxOut@ at the @HardForkBlock@ level.
+-- The 'MemPack' codecs for the era's @TxIn@\/@TxOut@ are superclasses: the
+-- on-disk (LSM) backend serialises individual entries with them, and bundling
+-- them here means the hard-fork dispatch reaches them through @proxySingle@
+-- (via @SingleEraBlock@) with no extra per-era constraint.
+type SingleEraBlockSupportsUTxOHD :: Type -> Constraint
```
bladyjoker, line 389: Does this have to be here? Single, Era, TxIn, TxOut, sounds like we're misplacing this abstraction, perhaps this belongs in Shelley or HFC?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -251,3 +292,158 @@ instance StandardHash blk => StandardHash (LedgerState blk)
...
+  -- on-disk backend).
+  valuesToList :: Values blk -> [(TxIn blk, TxOut blk)]
+
+  -- | Rebuild the values from a list of entries (e.g. from a backend page /
+  -- point-lookup result).
+  valuesFromList :: [(TxIn blk, TxOut blk)] -> Values blk
+
+  -- | The diff as a plain list of per-key insert\/delete deltas (e.g. to batch
+  -- onto the on-disk backend).
+  diffToList :: Diff blk -> [(TxIn blk, Diff.Delta (TxOut blk))]
```
bladyjoker, line 417: This entire class is saying: Values and Keys are contains TxIn and TxOut, so let's be explicit about its purpose by just renaming things a bit if it makes sense?

```haskell
class  ValuesHasUTxOs blk
  where
  utxoRangeRead ::
    (Maybe (TxIn blk), Int) -> Values blk -> (Values blk, Maybe (TxIn blk))

  utxoKeys :: Keys blk -> [TxIn blk]

  utxosFromValues :: Values blk -> [(TxIn blk, TxOut blk)]

  utxosToValues :: [(TxIn blk, TxOut blk)] -> Values blk

  utxosFromDiff :: Diff blk -> [(TxIn blk, Diff.Delta (TxOut blk))]
```

In the future when you need to provide similar for Accounts or whatever, you can have `ValuesHasAccounts` and such? Although, I wonder if this entire thing should be evicted from the abstract area

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -251,3 +292,158 @@ instance StandardHash blk => StandardHash (LedgerState blk)
...
+  --
+  -- Takes the already-loaded 'LedgerState' as an era hint: there is no on-disk
+  -- era tag (see 'encodeValues'), so the hard-fork combinator uses the state's
+  -- current era to pick which @NS@ arm to decode into. Single-era instances
+  -- ignore it. The snapshot loads the state before the tables, so it is always
+  -- available at the call site.
+  decodeValues :: forall s. LedgerState blk -> Decoder s (Values blk)
+
+-- | The on-disk table operations that only /single-era/ blocks support, split
+-- out of 'BlockSupportsUTxOHD'.
```
bladyjoker, line 375: This feels off, let's chat about this more. Now the abstract layer speaks about eras?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
bladyjoker, line 1: I really struggle with the SingleEra classes here. Can you help me understand this better?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -43,30 +55,52 @@ module Ouroboros.Consensus.Ledger.Basics
...
+--
+-- This is era-stable across all Shelley-based eras. It is convenient per-block
+-- vocabulary for the Shelley\/dual instances that build their opaque 'Keys' out
+-- of it.
+type TxIn :: Type -> Type
+type family TxIn blk
+
+-- | Each block has a notion of a @TxOut@: the value of the on-disk UTxO table.
+type TxOut :: Type -> Type
+type family TxOut blk
```
bladyjoker, line 87: Why don't you place this in `SingleEraBlockSupportsUTxOHD` or how I dubbed it `ValuesHasUTxOs`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -251,3 +292,158 @@ instance StandardHash blk => StandardHash (LedgerState blk)
...
+  -- still at era @E@) first upgrades the values @E → E+1@ using the per-era
+  -- translations carried on the 'CanHardFork' class (@hardForkEraTranslation@),
+  -- so no config is needed here.
+  --
+  -- @'forward' [] = 'id'@ must hold genuinely so that, per block, applying an
+  -- empty diff never rebuilds the values.
+  --
+  -- Note: @blk@ occurs only under the non-injective families 'Diff'\/'Values',
+  -- so call sites disambiguate with @'forward' \@blk diffs vals@.
+  forward :: [Diff blk] -> Values blk -> Values blk
```
bladyjoker, line 350: Again here we're talking about `forwarding` through eras. At this point aren't we saying that the abstract layer is using "forward" to facilitate HFC? Did you consider removing this from here and handling that in concrete layer, I'd imagine HFC?

bladyjoker (commented):
First try!

Loving the removal of `mk` and what that looks like!

As I was reading the changes I left a lot of comments around naming, sorry about nagging, I find it important as it can facilitate building mental models.

I'm also struggling with `forward` and Era specific classes and workflows in the abstract `blk` realm. I want to dig deeper into that.

I realized my preferred way to review this was by components, focusing on the ones that I'm familiar with first. I'll continue with Mempool.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
bladyjoker: Javier explained on Slack:

Dražen Popović  [1:03 PM]
From the convo I head with in ledger ticking produces a state change that's observable to transaction only when epoch boundary is crossed.

Even so, I'm not aware that "values" can change in tick? Or at least utxos

Javier Sagredo  [1:12 PM]
Only in Era boundaries. And only in one of them (Shelley -> Allegra)
[1:17 PM]AVVMs were some accounts from the beginning of Cardano that no one claimed
[1:17 PM]So those funds were transferred to the treasury in Shelley -> Allegra boundary

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
bladyjoker: Well I read to much into the comment, `forward` is properly `blk` abstract and is just about applying changes to values.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
jasagredo: `restrictValues` and `valuesSize` will be the map operations only on Shelley, they don't _need_ to be on Maps.

`forward`, maybe we can rename that one, but I think forward is fine too.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Protocol/EnvelopeChecks.hs
jasagredo: Not sure what you mean here. This is probably an artifact of a needed rebase on f-f/prepare-11.1

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToClient.hs
jasagredo: Not sure what you mean here. This is probably an artifact of a needed rebase on f-f/prepare-11.1

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
jasagredo: The suffix `Shim` implies what we want to convey with this module. It is what we would want the Ledger to expose, we don't want to own it forever.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
jasagredo: Probably yes, but we are familiar with stow/unstow, others are not, so split sounds reasonable.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/LedgerCallShim.hs
jasagredo: Same reason, this name clearly conveys "clear" i.e. strip away

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: Well, UTxOHD is the name of the feature for now, until we have LedgerHD. `BlockSupportsStorage` would be much more confusing (Storage in what precise way?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: This belongs in the abstract layer to be able to refer to it from within the HFC code when we go into a particular era.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: I am not sure I like this. Precisely by saying Values we are hiding what values we are reading. They might be UTxOs now but might be more in the future. 

I am not entirely convinced on rejecting your idea, but I feel it doesn't hurt as is for now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: It always did, `SingleEraBlock` is in the abstract layer

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: Hmmm... I think the key point to understand them is that they will only be required by `SingleEraBlock` so that the HFC has them at hand, BUT the HF block does not implement them, thus it doesn't need to define TxIn/TxOut etc.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: I think this would be a good move, yes

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Basics.hs
jasagredo: Indeed, it is advancing through blocks, not necessarily eras

PR #2043, opened 2026-05-18
Node 11.1 integration

f-f:
# Description

Integration for release 11.1 - here we update to the latest `base`/`network`/`ledger`

PR #2082, opened 2026-06-18
Drastically reduce the cardano-api vendored modules in preparation for dependency inversion

jasagredo:
Subtle errors are lurking in these functions. This is preparatory work for cardano-api to get these definitions from Consensus instead of Consensus vendoring modules that get out of data.

Based on #2043


jasagredo:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2087

PR #2083, opened 2026-06-18
leios-prototype: prevent leaking LeiosFetch budget on disconnects

nfrisby:
We've seen a some forks on the testnet recently, and this PR fixes a bug that is a plausible cause.

- The testnet is seeing "spurious" disconnects (for other reasons)---unsurprising on a _testnet_.
- It's possible those disconnects are correlated with Leios fetch requests (since they are bursts of work, and the disconnects we've seen are from the Mempool hard timeout, which CPU starvation could cause).
- If so, then the bug fixed in this PR might explain the testnet forks.

The bug was: this testnet prototype evolved from a much narrower demo, in which disconnects were irrelevant. As such, it didn't handle them at all. In particular, if some requests were in-flight for the peer during the disconnect, then the LeiosFetch logic was never given the bad news that those requests' replies would never arrive. The unfortunate side-effect was that those outstanding requests still consumed some of the LeiosFetch budget. If that were to happen often enough, then the entire budget would be consumed by zombie requests, and so LeiosFetch would stop requesting anything. And that might lead to a testnet node not being able to follow the other testnet node's chain(s).

----

This PR updates the LeiosFetch logic (and related mini protocol spawning/tear down) to react properly when a peer disconnects. In particular, its outstanding requests (both assigned-but-not-yet-sent and also sent-but-not-yet-replied-to) are removed from the central state tracking outstanding requests.

The yak-shaving beyond that specific change is necessary to preventing write-conflicts:

- The `msgLeiosBlock` and `msgLeiosBlockTxs` handlers might race with the budget-flush, which could have resulted in them _also_ decreasing the budget when one of the replies arrived around the same time of the disconnect.

- The LeiosFetch logic snapshots the current peers and then makes its decisions and updates under a `modifyMVar outstandingVar ...`. Which means if a peer disconnects in between those two, there's a risk that the LeiosFetch logic would update `outstandingVar` by assigning new requests to the disconnected peer---which would still risk the same leak (just probably smaller and much less often).

The first we fix by making those handlers slightly smarter: skip the update to `outstanding` if the peer isn't still present in the current `outstanding`. The second we fix by re-snapshotting the live peer set inside of the `modifyMVar outstandingVar` that decides on new requests.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -435,10 +435,13 @@ mkHandlers
           let tracer = leiosPeerTracer peer
               kernelTracer = Node.leiosKernelTracer tracers
               LeiosVoteState{addVote} = leiosVoteState
-          MVar.modifyMVar_ getLeiosPeersVars $ \leiosPeersVars -> do
-            x <- Leios.newLeiosPeerVars
-            let !leiosPeersVars' = Map.insert (Leios.MkPeerId peer) x leiosPeersVars
-            pure leiosPeersVars'
+          -- Allocate this peer's vars once and keep them in scope; the message
+          -- handlers below can write to 'peerVars' directly. The map insertion
+          -- exists only to publish them to the LeiosFetch client and the fetch
+          -- decision logic.
+          peerVars <- Leios.newLeiosPeerVars
+          MVar.modifyMVar_ getLeiosPeersVars $
+            pure . Map.insert (Leios.MkPeerId peer) peerVars
           pure $
             leiosNotifyClientPeerPipelined
               ( atomically controlMessageSTM <&> \case
@@ -478,11 +481,6 @@ mkHandlers
                               { Leios.missingEbBodies =
                                   Map.insert point ebBytesSize (Leios.missingEbBodies outstanding)
                               }
-                    peerVars <- do
-                      peersVars <- MVar.readMVar getLeiosPeersVars
-                      case Map.lookup (Leios.MkPeerId peer) peersVars of
-                        Nothing -> error "impossible!"
-                        Just x -> pure x
                     MVar.modifyMVar_ (Leios.offerings peerVars) $ \(offers1, offers2) -> do
                       let !offers1' = Set.insert ebHash offers1
                       pure (offers1', offers2)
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -521,13 +509,23 @@ mkHandlers
...
-                  case Map.lookup (Leios.MkPeerId peer) leiosPeersVars of
-                    Just x -> pure $ Leios.requestsToSend x
-                    Nothing -> do
-                      threadDelay (0.010 :: DiffTime)
-                      loop
+                  terminating <- atomically $ (== Terminate) <$> controlMessageSTM
+                  if terminating
+                    then TVar.Unchecked.newTVarIO Seq.empty
+                    else do
+                      leiosPeersVars <- MVar.readMVar getLeiosPeersVars
```
ch1bo, line 542: How wonderfully composable would it be if `getLeiosPeersVars` were `STM`.

I think we'll go there another time though :)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -310,6 +310,17 @@ data LeiosOutstanding pid = MkLeiosOutstanding
   --
   -- * 'blockingPerEb' is only decremented when txs are actually inserted
   --   into the DB (via @MsgLeiosBlockTxs@ handling).
+  --
+  -- TODO: 'blockingPerEb' can go permanently stale for txs shared across EBs.
+  -- 'msgLeiosBlockTxs' only decrements the entry for the EB it was requesting;
+  -- a tx that also belongs to another EB B is then in the DB, so 'filterMissingWork'
+  -- drops it from B's missing set and B never fetches it itself -- so B's
+  -- 'blockingPerEb' is never decremented for that tx and stays > 0. This is
+  -- currently harmless only because nothing reads 'blockingPerEb' as a gate: the
```
ch1bo, line 320: I just checked and `blockingPerEb` is really only ever written to and the explanation here of it going stale/inconsistent with the EB is plausible too.

Should: just remove it

ch1bo (approved):
Gonna give this a spin manually, but looks good as a hotfix. Looking forward to a proper revamp of the whole logic :)

ch1bo:
Testnet sync was successful without restart, which makes me confident that this helped :+1:

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
nfrisby: Agreed --- just wanted to keep the "hotfix" narrow. And, as you said, the bigger LeiosFetch is incoming anway

PR #2086, opened 2026-06-19
Leios: vote and certify over the announcing RbHash

### LINKED ISSUE #941: Leios Certificates and Votes must be signing RbHash
Currently the Leios "point" is specified as 

https://github.com/IntersectMBO/ouroboros-consensus/blob/d941d06ef8bbb35fd369c09c1d9d76c20b561969/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs#L127-L129

which includes the EB hash and the SlotNo of the announcing Praos header (RB).

The requirement for Leios "point" is to convey what "change" is made on top of which "state". With that in mind, the "change" is the Endorser Block, however, is the SlotNo a sufficient representation of the "state"? Let's examine!

If we have two RBs on different forks that are issued on the same Slot and carry the same EB Announcement then we might use a Leios Certificate destined for one fork on another! Is this bad? Let's examine!

Imagine 2 forks X and Y.

X announces an EB 0xA that is a VALID ledger state extension at Slot 10 that it certifies at Slot 20
Y announces the same EB 0xA at Slot 10 but it's an INVALID ledger state extension, and then it certifies it with a certificate from fork X.

Why would nodes apply a Certificate that is INVALID? Because the Certificate represents the notion that a majority of Stake has downloaded and validated the referred EB...so we trust it to be valid and don't validate the transactions contained within.

🔴 With all this in mind that means that the fork Y will end up applying an Endorser Block INVALID wrt to its own selected Ledger State.

TODO:
  - [ ] Update the prototype to use the [Ranking Block Hash](https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network/api/Ouroboros-Network-Block.html#t:HeaderHash)
  - [ ] Update the [CDDLs](https://github.com/cardano-scaling/CIPs/blob/ch1bo/committee-selection/CIP-0164/README.md#votes-certificates-cddl)  to use the Ranking Block Hash
  - [ ] Can we make a test that assures that a Leios Certificate is unique per (StateId, ChangeId) tuple. 

## References

- https://dbdiagram.io/d/Leios-relational-view-6a2a75259340ecc065755feb

geo2a:
Fixes https://github.com/input-output-hk/ouroboros-leios/issues/941

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
@@ -32,7 +32,7 @@ import LeiosDemoDb
   ( leiosDbInsertEbBody
   , leiosDbInsertEbPoint
   , leiosDbInsertTxs
-  , leiosDbQueryCompletedEbByPoint
+  ( leiosDbQueryCompletedEbByHash
   )
 import LeiosDemoTypes
   ( EbAnnouncement (..)
@@ -180,7 +180,7 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
             -- TODO: Why exactly do we guard against this? Also, shouldn't we
             -- detect it the other way around: if we have a cert, but not
             -- downloaded it ourselves -> warning!
-            mClosure <- leiosDbQueryCompletedEbByPoint fbLeiosDb ebPoint
+            mClosure <- leiosDbQueryCompletedEbByHash fbLeiosDb (pointEbHash ebPoint)
             case mClosure of
               Nothing -> do
                 traceWith fbLeiosTracer $
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
@@ -29,20 +29,22 @@ import Data.ByteString.Short (fromShort)
 import qualified Data.Sequence.Strict as Seq
 import qualified Data.Typeable as Typeable
 import LeiosDemoDb
-  ( leiosDbInsertEbBody
-  , leiosDbInsertEbPoint
-  , leiosDbInsertTxs
   ( leiosDbQueryCompletedEbByHash
+  , LeiosDbConnection (..)
   )
 import LeiosDemoTypes
   ( EbAnnouncement (..)
   , ForgedLeiosEb (..)
   , LeiosPoint (..)
   , TraceLeiosKernel (..)
+  , RbHash (..)
+  , EbHash (..)
   , forgeLeiosEb
   , leiosEbBytesSize
   , minCertificationGap
+  , hashLeiosEb
   )
+import Data.ByteString.Short (fromShort)
 import LeiosVoteState (LeiosVoteState (queryCert))
 import Lens.Micro ((&), (.~))
 import Ouroboros.Consensus.Block
@@ -85,8 +87,8 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
     case Typeable.eqT @era @DijkstraEra of
       Just Refl -> decideLeios
       Nothing -> pure (Nothing, SNothing)
-  let body = mkBody mayLeiosCert
-      actualBodySize = SL.blockBodySize protocolVersion body
+  let rbBody = mkBody mayLeiosCert
+      actualRbBodySize = SL.blockBodySize protocolVersion rbBody
   hdr <-
     mkHeader @_ @(ProtoCrypto proto)
       hotKey
@@ -95,13 +97,13 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
       fbCurrentSlotNo
       fbCurrentBlockNo
       prevHash
-      (SL.hashBlockBody @era body)
-      actualBodySize
+      (SL.hashBlockBody @era rbBody)
+      actualRbBodySize
       protocolVersion
-      mayEbAnn
-  let blk = mkShelleyBlock $ SL.Block hdr body
-  case mayEbAnn of
-    Just ann -> do
+      (snd <$> mayEbAnn)
+  let blk = mkShelleyBlock $ SL.Block hdr rbBody
+  case fst <$> mayEbAnn of
+    Just forgedEb -> do
       let rbHashBytes =
             fromShort $
               toShortRawHash (Proxy @(ShelleyBlock proto era)) $
@@ -109,13 +111,14 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
           ebPoint =
             MkLeiosPoint
               { pointSlotNo = fbCurrentSlotNo
-              , pointEbHash = ebAnnouncementHash ann
+              , pointEbHash = hashLeiosEb . body $ forgedEb
               }
       traceWith fbLeiosTracer $
         TraceLeiosBlockAnnounced
           { announcingRbHashBytes = rbHashBytes
           , announcedEbPoint = ebPoint
           }
+      storeEb ebPoint (MkRbHash rbHashBytes) forgedEb
     Nothing -> pure ()
   return $
     assert (verifyBlockIntegrity (configSlotsPerKESPeriod $ configConsensus fbConfig) blk) $
@@ -154,13 +157,13 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
   -- certificate available in LeiosDb), suppressing this block's own EB
   -- announcement. Otherwise fall back to forging a new EB (when the
   -- mempool has txs for one). Mirrors the prototype's 'decideForgeType'.
-  decideLeios :: m (Maybe EbAnnouncement, StrictMaybe LeiosCert)
+  decideLeios :: m (Maybe (ForgedLeiosEb, EbAnnouncement), StrictMaybe LeiosCert)
   decideLeios = do
     cert <- decideCertify
     case cert of
       SJust _ -> pure (Nothing, cert)
       SNothing -> do
-        ann <- mkAndStoreEb
+        ann <- mkEb
         pure (ann, SNothing)
 
   decideCertify :: m (StrictMaybe LeiosCert)
@@ -188,12 +191,12 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
                     "EB not yet downloaded: " <> show ebPoint
                 pure SNothing
               Just _ -> do
-                -- Pull the assembled certificate from this node's
-                -- 'LeiosVoteState' — populated as votes accumulate and
-                -- cross 'minCertificationThreshold' (see
-                -- "LeiosVoteState"). Replaces the previous LeiosDb
-                -- placeholder cert (an empty bitfield + dummy sig).
-                mCert <- queryCert fbLeiosVoteState ebPoint
+                -- we get the hash of the previous block header because eb certification must
+                -- happen withing one block (this is the linear aspect of linear leios).
+                let announcingRbHash = case getTipHash fbCurrentTickedLedgerState of
+                      BlockHash h -> MkRbHash (toRawHash (Proxy @(ShelleyBlock proto era)) h)
+                      GenesisHash -> error "decideCertify: cannot certify on top of genesis"
+                mCert <- queryCert fbLeiosVoteState announcingRbHash
                 case mCert of
                   Nothing -> do
                     traceWith fbLeiosTracer $
@@ -212,12 +215,12 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
   -- announcement to embed in the header. An honest forger only emits an
   -- EB when it has txs to put in it; empty mempool ⇒ no EB ⇒ no
   -- announcement (matches the original prototype).
-  mkAndStoreEb :: m (Maybe EbAnnouncement)
-  mkAndStoreEb = case nonEmpty (fmap extractTx fbEbTxs) of
+  mkEb :: m (Maybe (ForgedLeiosEb, EbAnnouncement))
+  mkEb = case nonEmpty (fmap extractTx fbEbTxs) of
     Nothing -> pure Nothing
     Just ebTxs -> do
       let forgedEb = forgeLeiosEb fbCurrentSlotNo ebTxs
-          ebHash = pointEbHash (forgedEb.point)
+          ebHash = hashLeiosEb forgedEb.body
           ebSize = leiosEbBytesSize (forgedEb.body)
           ebAnn =
             EbAnnouncement
@@ -231,12 +234,17 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
           , ebMeasure = ByteSize32 ebSize
           , mempoolRestMeasure = ByteSize32 0
           }
-      leiosDbInsertEbPoint fbLeiosDb (forgedEb.point) ebSize
-      leiosDbInsertEbBody fbLeiosDb (forgedEb.point) (forgedEb.body)
-      void $ leiosDbInsertTxs fbLeiosDb (forgedEb.txClosure)
-      traceWith fbLeiosTracer $
-        TraceLeiosBlockStored
-          { slot = fbCurrentSlotNo
-          , eb = forgedEb.body
-          }
-      pure (Just ebAnn)
+
+      pure (Just (forgedEb, ebAnn))
+
+  storeEb :: LeiosPoint -> RbHash -> ForgedLeiosEb -> m ()
+  storeEb point announcingRbHash forgedEb = do
+    let ebSize = leiosEbBytesSize (forgedEb.body)
+    leiosDbInsertEbPoint fbLeiosDb point ebSize
+    leiosDbInsertEbBody fbLeiosDb point (forgedEb.body)
+    void $ leiosDbInsertTxs fbLeiosDb (Just announcingRbHash) (forgedEb.txClosure)
+    traceWith fbLeiosTracer $
+      TraceLeiosBlockStored
+        { slot = point.pointSlotNo
+        , eb = forgedEb.body
+        }
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -63,16 +63,28 @@ import Data.Proxy (Proxy (..))
 import Data.Sequence.Strict ((|>))
 import qualified Data.Set as Set
 import Data.Word (Word64)
-import LeiosDemoDb (LeiosDbConnection, newLeiosDBInMemoryWith, withLeiosDb)
+import LeiosDemoDb
+  ( LeiosDbConnection
+  , newLeiosDBInMemoryWith
+  , withLeiosDb
+  )
 import LeiosDemoTypes
-  ( LeiosPoint (..)
+  ( EbHash
+  , LeiosPoint (..)
   , LeiosVote (..)
+  , RbHash (..)
   , TraceLeiosKernel (..)
   , hashLeiosEb
   , minCertificationGap
   )
 import Lens.Micro ((%~), (^.))
-import Ouroboros.Consensus.Block (SlotNo (..), blockSlot)
+import Ouroboros.Consensus.Block
+  ( SlotNo (..)
+  , blockHash
+  , blockSlot
+  , getHeader
+  , toRawHash
+  )
 import Ouroboros.Consensus.Cardano
   ( CardanoBlock
   , Nonce (NeutralNonce)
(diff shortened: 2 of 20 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -525,8 +526,13 @@ applyBlock leiosDb evs cfg ap fo doResolveBlock = case ap of
...
                 -- TODO: make this less fatal
                 error "applyBlock: CertRB seen but no Leios committee for this era"
-            -- FIXME: This should not be about a LeiosPoint, but an RbHash
-            case verifyLeiosCert cm minCertificationThreshold announcedPoint cert of
+            let announcingRbHashValue = case announcingRbHash b of
+                  Just h -> h
+                  Nothing ->
+                    -- A CertRB always has a (non-genesis) announcing parent;
+                    -- the apply path shouldn't have produced one otherwise.
+                    error "applyBlock: cannot determine announcing RB hash for CertRB"
```
ch1bo, line 534: Should: replace the call to `protocolStateLeiosAnnouncement` above with this

The RbHash is the new announcement and all what we need to verify the cert. We might be able to drop `protocolStateLeiosAnnouncement` completely

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -741,7 +771,7 @@ data TraceLeiosKernel
     TraceLeiosBlockCertified {atSlot :: SlotNo, certifiedPoint :: LeiosPoint}
   | TraceLeiosVoted {vote :: LeiosVote, weight :: Weight}
   | TraceLeiosVoteAcquired {vote :: LeiosVote}
-  | TraceLeiosCertified {point :: LeiosPoint}
+  | TraceLeiosCertified {rbHash :: RbHash}
```
ch1bo, line 778: Could: keep `point` and give this a more descriptive name like `announcingRbHash`

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -508,9 +519,9 @@ mkHandlers
                   )
               processEbNotification =
                 readTChan chan >>= \case
-                  AcquiredEb point ebSize ->
-                    pure $ MsgLeiosBlockOffer point ebSize
-                  AcquiredEbTxs point ->
+                  AcquiredEb point announcingRbHash ebSize ->
+                    pure $ MsgLeiosBlockOffer point (Leios.rbHashBytes announcingRbHash) ebSize
```
ch1bo: Must: not change the N2N protocol because of this

Why would we need the offer to include the announcement?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -131,6 +132,32 @@ decodeEbHash = MkEbHash <$> CBOR.decodeBytes
...
+decodeRbHash :: Decoder s RbHash
+decodeRbHash = MkRbHash <$> CBOR.decodeBytes
+
+prettyRbHash :: RbHash -> String
+prettyRbHash (MkRbHash bytes) = BS8.unpack (BS16.encode bytes)
+
+instance SignableRepresentation RbHash where
+  getSignableRepresentation point =
+    toStrictByteString $
+      encodeRbHash point
```
ch1bo, line 161: Should: just use the hash bytes directly instead of CBOR

Not entirely sure what is the more common paradigm in the codebase? In any case, we should keep a note in the CIP / the vote CDDL about this.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -561,7 +594,7 @@ mkCommitteeEveryoneVotes inputs =
 
 -- | A vote in the Leios protocol.
 data LeiosVote = MkLeiosVote
-  { point :: LeiosPoint
+  { point :: RbHash
   -- ^ Point that gets signed. The slot also identifies the voting round.
```
ch1bo: Should: rename this field. It's not a point, but just a block header hash.

Also, there is no slot in this.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -582,36 +615,33 @@ instance ShowProxy LeiosVote where showProxy _ = "LeiosVote"
...
       , voterId
       , voteSignature
       }
 
 voteToObject :: LeiosVote -> Aeson.Object
 voteToObject MkLeiosVote{point, voterId} =
   mconcat
-    [ "slot" .= point.pointSlotNo
-    , "ebHash" .= prettyEbHash point.pointEbHash
+    [ "rbHash" .= prettyRbHash point
```
ch1bo: Note to myself: The frontend uses the `slot` and `ebHash` to visualize votes

We must update it

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -582,36 +615,33 @@ instance ShowProxy LeiosVote where showProxy _ = "LeiosVote"
...
     , "voterId" .= voterId.voterIndex
     ]
 
 -- | Create a vote for given 'LeiosPoint' and signing key.
```
ch1bo, line 640: Should: fix Haddocks

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -40,8 +41,8 @@ data LeiosDbHandle m = LeiosDbHandle
   }
 
 data LeiosEbNotification
-  = AcquiredEb LeiosPoint BytesSize
-  | AcquiredEbTxs LeiosPoint
+  = AcquiredEb LeiosPoint RbHash BytesSize
+  | AcquiredEbTxs LeiosPoint RbHash
```
ch1bo: Do we emit this now for each announcement (RbHash) that was completed when an EbTxs (closure) gets completed?

I'm asking because we would need to vote for each announcement (that is not equivocated).

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -80,6 +81,8 @@ data LeiosDbConnection m = LeiosDbConnection
   , leiosDbFilterMissingTxs :: HasCallStack => [TxHash] -> m [TxHash]
   -- ^ Batch filter: returns the subset of input TxHashes that we do NOT have.
   , leiosDbQueryCompletedEbByPoint :: HasCallStack => LeiosPoint -> m (Maybe [(TxHash, ByteString)])
+  , leiosDbRbOfEb :: HasCallStack => EbHash -> m RbHash
+  -- ^ For an EB, find the RB that announced it.
```
ch1bo: Must: store announcements differently

This is not a 1:1 relationship. Instead, we would have a n:m relationship between announcements (RB headers) and EBs.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -61,9 +62,9 @@ data LeiosDbConnection m = LeiosDbConnection
   -- - Missing TXs: TXs in ebTxs without entries in txs
   -- NOTE: This is O(n) and should only be used at startup for initialization.
   , -- NOTE: yields a LeiosOfferBlock notification
-    leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m ()
+    leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> RbHash -> LeiosEb -> m ()
```
ch1bo: Should: not need to change this

An EB body should actually only be indexed by `EbHash` and having the `Slot` via LeiosPoint here is already problematic.

ch1bo (changes requested):
Not fully through it, but I'm not convinced that we need that many changes.

The certificate verification and vote signing is correct AFAIS, but it's a big of a too big rats tail on the storage and definitely the N2N mini protocols should not change? Please argue for why if I'm wrong.

johnalotoski:
Hello, the consensus-test is consuming ~60 GB RAM and OOMing in CI.

Examples: [here](https://ci.iog.io/build/13419037/nixlog/1), [here](https://ci.iog.io/build/13419102/nixlog/1) and [here](https://ci.iog.io/build/13418972/nixlog/1).

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -54,18 +81,32 @@ runLeiosVoting tracer chainDB leiosDB voteState = \case
...
+      --   most recent header that announces this EB. If it is no
+      --   longer avaliable (we switched to a different fork), we do not vote.
+      (ls, mChainRbHash) <- atomically $ do
+        extLedger <- ChainDB.getCurrentLedger chainDB
+        frag <- ChainDB.getCurrentChain chainDB
+        -- Note: here we will traverse the currently selected volatile chain fragment,
+        --       i.e. at most k headers.
+        let announcingRbHash = findAnnouncingRbHash frag (pointEbHash announcingPoint)
+        pure (ledgerState extLedger, announcingRbHash)
+      let mAnnouncingRbHash = maybe mChainRbHash Just mNotifiedRbHash
```
ch1bo:
```suggestion
      let mAnnouncingRbHash = mNotifiedRbHash <|> mChainRbHash 
```

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -54,18 +81,32 @@ runLeiosVoting tracer chainDB leiosDB voteState = \case
...
+      -- * Otherwise, we've received an EB from a network peer, and we derive
+      --   the announcing RB hash from our selected chain: the
+      --   most recent header that announces this EB. If it is no
+      --   longer avaliable (we switched to a different fork), we do not vote.
+      (ls, mChainRbHash) <- atomically $ do
+        extLedger <- ChainDB.getCurrentLedger chainDB
+        frag <- ChainDB.getCurrentChain chainDB
+        -- Note: here we will traverse the currently selected volatile chain fragment,
+        --       i.e. at most k headers.
+        let announcingRbHash = findAnnouncingRbHash frag (pointEbHash announcingPoint)
```
ch1bo: Could: extend and use `protocolStateLeiosAnnouncement` here that starts from a `ChainDepState`

This allows us to just use the `extLedger` state and skips walking/inspecting headers and should reduce the constraints needed on this function.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -54,18 +81,32 @@ runLeiosVoting tracer chainDB leiosDB voteState = \case
     let getNext f =
           atomically (readTChan chan) >>= \case
             AcquiredEb{} -> pure ()
-            AcquiredEbTxs point -> f point
+            AcquiredEbTxs point mRbHash -> f point mRbHash
 
     -- Enter voting loop
-    forever $ getNext $ \point -> do
+    forever $ getNext $ \announcingPoint mNotifiedRbHash -> do
```
ch1bo: Could: not call this `announcingPoint`

This is includes the Leios EB hash that _was announced_. It was slightly confusing when reading this.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -180,20 +183,20 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
...
-                -- 'LeiosVoteState' — populated as votes accumulate and
-                -- cross 'minCertificationThreshold' (see
-                -- "LeiosVoteState"). Replaces the previous LeiosDb
-                -- placeholder cert (an empty bitfield + dummy sig).
-                mCert <- queryCert fbLeiosVoteState ebPoint
+                -- we get the hash of the previous block header because eb certification must
+                -- happen withing one block (this is the linear aspect of linear leios).
+                let announcingRbHash = case getTipHash fbCurrentTickedLedgerState of
+                      BlockHash h -> MkRbHash (toRawHash (Proxy @(ShelleyBlock proto era)) h)
+                      GenesisHash -> error "decideCertify: cannot certify on top of genesis"
```
ch1bo, line 198: Could: inspect the chain (state) only once in this function

This is correct, but feels a bit duplicate to the `fbChainDepState >>= protocolStateLeiosInfo` above.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -231,12 +234,18 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
...
-          }
-      pure (Just ebAnn)
+
+      pure (Just (forgedEb, ebAnn))
+
+  -- TODO(geo2a): 'RbHash' should go inside 'LeiosPoint'
+  storeEb :: LeiosPoint -> RbHash -> ForgedLeiosEb -> m ()
+  storeEb point announcingRbHash forgedEb = do
+    let ebSize = leiosEbBytesSize (forgedEb.body)
+    leiosDbInsertEbPoint fbLeiosDb point ebSize
```
ch1bo, line 243: Nitpick: I don't agree with the TODO text. Instead this `leiosDbInsertEbPoint` should actually become a `leiosDbInsertEbAnnouncement` that takes an `RbHash`, `SlotNo`, `EbHash` and an `EbSize` .. or all of this wrapped up in an announcement type.

Out of scope for this PR though

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -63,10 +68,12 @@ data LeiosDbConnection m = LeiosDbConnection
   , -- NOTE: yields a LeiosOfferBlock notification
     leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m ()
   , -- TODO: Take [LeiosTx] and hash on insert?
-    leiosDbInsertTxs :: HasCallStack => [(TxHash, ByteString)] -> m CompletedEbs
+    leiosDbInsertTxs :: HasCallStack => Maybe RbHash -> [(TxHash, ByteString)] -> m CompletedEbs
```
ch1bo, line 71: Should: not store the announcing RbHash (optionally!) with the txs

This is not the direction we'd want to go. Instead, `leiosDbInsertEbPoint` should become more and more the "store this announcement" API. And this would also need to have the `RbHash` (the root identity of our announcement). The database then would have a table of announcements (at which `RbHash`, which `EbHash` was announced), which we then should use upon inserting txs: If a closure of a specific `EbHash` is complete, create notifications for each announcement (if we really need to do it this way: alternatively, the calling code could resolve `EbHash` -> `RbHash`)

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -242,20 +254,46 @@ prop_leios seed =
...
+        )
+      | chain <- Map.elems nodeChains
+      , blk <- chain
+      , Just (annPoint, _) <- [headerLeiosAnnouncement (getHeader blk)]
+      ]
+
+  -- 'acquiredPoints' translated to the RB hashes voters sign, so it can be
+  -- compared directly against 'votedAnnouncingRbHashes'.
+  acquiredRbHashes :: Set.Set RbHash
+  acquiredRbHashes =
```
ch1bo, line 276: Could: call these announcements (acquired / voted)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -492,8 +492,8 @@ mkHandlers
                       -- a cert for the point.
                       case result of
                         Added _ (Just _) ->
-                          let Leios.MkLeiosVote{Leios.point = pt} = vote
-                           in traceWith kernelTracer TraceLeiosCertified{point = pt}
+                          let point = undefined -- TODO(geo2a): where to get the point from?? Used to come from the vote.
+                           in traceWith kernelTracer TraceLeiosCertified{point, rbHash = Leios.announcingRbHash vote}
```
ch1bo: Should: drop the point from the trace or add it to the `AddVoteResult`, as we _could_ keep track of which cert is about which announced EB in the votestate

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -41,7 +42,11 @@ data LeiosDbHandle m = LeiosDbHandle
 
 data LeiosEbNotification
   = AcquiredEb LeiosPoint BytesSize
-  | AcquiredEbTxs LeiosPoint
+  | -- | An EB's tx closure is complete. Carries the announcing ranking block's
+    -- hash when it is known locally at insertion time (a node knows it for an
+    -- EB it forged itself); 'Nothing' for EBs fetched from peers, in which
+    -- case the voter derives it from its selected chain.
+    AcquiredEbTxs LeiosPoint (Maybe RbHash)
```
ch1bo, line 49: Should: avoid the `Maybe` here

We should not accumulate/store txs for an EB that was not announced. Hence, we must always know the `RbHash`.

ouroboros-consensus/test/consensus-test/Test/LeiosDemoDb.hs
ch1bo, line 1: All these `Nothing` calls are a bit smelly (see my comment about the db API above)

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
ch1bo, line 1: Must: rebase to latest `ch1bo/leios-certifying` .. there is no staging area anymore

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
ch1bo, line 534: Should: enhance/use `protocolStateLeiosAnnouncement` to not inspect the header/state twice

This would avoid this error case

ch1bo (approved):
I think this is correct and much better now in terms of impact. Quite some re-use still possible. Good to be integrated into `ch1bo/leios-certifying` (after rebase)!

ouroboros-consensus/src/ouroboros-consensus/LeiosStagingArea.hs
geo2a: Done.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
geo2a: :+1:

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
geo2a: I'll rename to simply "point" for the lack of a better short name.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
geo2a: Yeah, I agree. I've removed this TODO, changing the DB interface is something I could work next.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
geo2a: I'm not actually storing the `RbHash` into the database. The implementations of `leiosDbInsertTxs` are the only source of the `AcquiredEbTxs` notification, which now has to carry an `RbHash` so that the voting thread can vote on it. I.e. the implementations of `leiosDbInsertTxs` don't actually write the rb hash to disk.

I agree that going forward we need to change `leiosDbInsertEbPoint` to be `leiosDbInsertEbAnnouncement` and create an announcement table.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
geo2a: I've dropped the point from the trace.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
geo2a: Having a `Maybe` here is necessary to limit the scope of this PR. This notification is issued in the implementations of `leiosDbInsertTxs`, which also now has to take an `Maybe RbHash`, since we only know the hash when we issue an EB ourselves, and we do not know it when we get it from the network via N2N. I've tried to augment the N2N logic in the previous iteration, but I agree that it is unnecessary. If we had an `EbHash -> RbHash` mapping in the DB (or stored `RbHash` in `LeiosPoint`), we would be able to eliminate this `Maybe`. Until then, I propose to keep the `Maybe` here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
geo2a: This requires me to change the `PraosState` type, probably by changing the payload of the `praosStateLeiosAnnouncement` filed, i.e. `EbAnnouncement`. This will cause some fallout, but not very much. Would you mind if I do it in a follow-up PR?

PR #2087, opened 2026-06-19
Eject db-synthesizer

jasagredo:
We keep all the funcionality but as we don't want to replicate cardano-api code, we delete the executable here and create it again downstream.


jasagredo:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2141

PR #2094, opened 2026-06-29
Optimistic Mempool

bladyjoker:
DONE
- [x] New tracing CacheHit/CacheMiss
- [x] Remove the check on matching slot number in getSnapshotFor


ch1bo:
This was only an experiment and can never be used as-is (it could lead to invalid blocks)

PR #2096, opened 2026-06-30
Change semantics of PerasRoundNo field in Bound

nbacquey:
This commit changes the semantics of the field of type `PerasRoundNo` in eras `Bound`. It used to be a `PerasEnabled PerasRoundNo`, holding no value when Peras was disabled, and holding the value of the Peras round corresponding to the bound when it was enabled.

There were two problems with that approach:
* There was a conflicting source of truth with the `PerasEnabled PerasRoundLength` value in `EraParams`. It was possible to see Peras as disabled in the parameters, but enabled in the bounds, or vice versa, which did not make sense.
* In the hypothetical case where Peras is enabled, then disabled in the next era, then re-enabled again in the following era, we would lose information of the current Peras round. We could not restart from 0 either, because the unicity of Peras round numbers is an implicit assumption of the protocol.

This commit solves both issues, by slightly tweaking the semantics of the field. When Peras is enabled (as specified by `EraParams`), it has the same signification as before. When Peras is disabled, it keeps its old value, thus representing what the Peras round number will be when Peras gets enabled again.

The new name of the field `boundNextPerasRound` reflects that new role.


geo2a (approved):
I think this does indeed look much cleaner!

nbacquey:
This has been manually merged in `peras/the-big-pr` already

PR #2099, opened 2026-07-01
[WIP Peras] Try implementing proper hfc dispatch

tbagrel1:
Remaining tasks:

- [x] Try getting rid of crypto superclass constraints in `BlockSupportsPeras` so we can avoid implement stupid crypto code for the HFC that will never be called (because BlockSupportsPeras will first dispatch to concrete era code and they call era-specific crypto, not the HFC one)
- [ ] ~Use the latest `strict-sop-core` so we can get rid of `All Top xs` constraints everywhere~ delegated to https://github.com/tweag/cardano-peras/issues/254


agustinmista:
Manually merged into `peras/the-big-pr`

PR #2100, opened 2026-07-01
Peras/fix tests

nbacquey:
This PR re-generates ledger-related golden files, and fixes issues in the block forging and vote forging threads


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
agustinmista, line 894: NIT: I think I would merge the new binding for `mResult` directly inside the one for `mbPerasCert` and see if we can take advantage of `fmap`/`>>=` for `Maybe`.

PR #2101, opened 2026-07-03
Fix resource management of leios db connections

ch1bo:
Should follow the TODOs right away next time!

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Diffusion.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Diffusion.hs
@@ -3,7 +3,6 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Diffusion.hs
```diff
@@ -127,14 +126,9 @@ run immDBDir sockAddr cfg getSlotDelay leiosDbFile leiosSchedule = withRegistry
             (leiosScheduler getSlotDelay leiosNotifyContext leiosSchedule)
         pure leiosNotifyContext
   let mkLeiosFetchContext = do
-        -- each LeiosFetch server calls this when it initializes
```
johnalotoski, line 130: A typo'd `--leios-db` will now fail silently with a created empty db?

johnalotoski (approved):

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ImmDBServer/Diffusion.hs
ch1bo: This was reminiscent when the schema was not initialized by the `newLeiosDBSQLite` and another call site would have initialized before this one here (likely the initial chain selection). Thus, removing this code does not change anything.

However, yes, if the path is mistyped between starts it would start a fresh leios db without much notice.

PR #2107, opened 2026-07-08
Fix genesis tests

nbacquey:
This PR fixes a bunch of issues in the Genesis tests, that were unearthed during Peras development:

- Shrinking the peer schedule could remove all peers, thus breaking several downstream assumptions
- Shrinking the adversarial schedule could break the assumption that `TipPoint >= HeaderPoint >= BlockPoint`, thus potentially leading to unexpected (and uninteresting, in the context of Genesis tests) disconnections.
- While generating chains, we would sometimes generate less than one block per epoch, thus leading to errors when the ledger would try to update directly from epoch `n` to epoch `n+2`

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking/Tests.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking/Tests.hs
@@ -4,8 +4,10 @@
 -- | Test properties of the shrinking functions
 module Test.Consensus.PointSchedule.Shrinking.Tests (tests) where
 
+import Cardano.Slotting.Slot (WithOrigin (..))
+import Control.Monad.Class.MonadTime.SI (Time (..))
 import Data.Foldable (toList)
-import Data.Map (keys)
+import Data.Map (elems, empty, keys, singleton)
 import Data.Maybe (mapMaybe)
 import Ouroboros.Consensus.Util (lastMaybe)
 import Test.Consensus.Genesis.Setup (genChains)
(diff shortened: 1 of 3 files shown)


ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking/Tests.hs
```diff
@@ -79,9 +88,23 @@ doesNotChangeFinalState original shrunk =
   lastBP :: PeerSchedule blk -> Maybe (SchedulePoint blk)
   lastBP sch = lastMaybe $ mapMaybe (\case (_, p@(ScheduleBlockPoint _)) -> Just p; _ -> Nothing) sch
 
-checkShrinkProperty ::
+-- | Checks that the shrunk schedule still has the properties TP >= HP and HP >= BP at any time
+preservesConsistency :: Ord blk => PeerSchedule blk -> PeerSchedule blk -> Bool
+preservesConsistency _original shrunk =
```
geo2a: Why keep `_origin` if it is unused?

geo2a (approved):

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PointSchedule/Shrinking/Tests.hs
nbacquey: I wanted the `checkShrinkProperty` function to have the same `Before -> After -> Bool` signature as the ones for honest peer schedule shrinking, so that it would be easier to extend them in the future. Maybe this was premature optimization on my part, I'll remove that.

PR #2114, opened 2026-07-09
Stop applying uncertified EBs from the ImmDB on startup (and increase LeiosNotify pipelining depth)

nfrisby:
Another bug I discovered during my time with the Leios testnet in the past couple days.


nfrisby:
I moments ago asked John Lotoski to try out a deployment of this PR's tip: https://input-output-rnd.slack.com/archives/C074AHSKJF7/p1783615689059509?thread_ts=1783609405.629099&cid=C074AHSKJF7

Two reasons:

- We should have a release that doesn't crash/build the wrong ledger state if there's an uncertified EB that's younger than the ledger snapshot file but on the immutable chain.
- The increased LeiosNotify pipelining depth will make vote diffusion more reliable.

bladyjoker (approved):

PR #2123, opened 2026-07-15
[Peras 27] Introduce runQueryEraIndexed as a generalization of runQuery

agustinmista:
This PR introduces `runQueryEraIndexed` as a generalization of `runQuery` that also returns the index of the era that a query was resolved in. In addition, it defines `runQuery` in terms of
`runQueryEraIndexed`.

Perhaps not too useful to mention this in the air without the rest of the code but, in any case, this will be used later to dispatch the creation of a Peras epoch context to the right era-based implementation.


nbacquey (approved):
lgtm

jasagredo (approved):

PR #2124, opened 2026-07-15
Update cardano-ledger dep to latest leios-prototype commit hash

koslambrou:
* Bump cardano-ledger source-repository-package tag
* Add `sppLeiosKey = SNothing` to StakePoolParams record constructions in Shelley and Cardano testlib
* Add .direnv to .gitignore


ch1bo (approved):
See my comment about calling this "leios key" on the upstream PR, but for the prototyping this is definitely fine.

Issue #2003, opened 2026-04-23
[BUG] - Eq instance for txid allocates memory


karknu:
*Internal* 

**Summary**
Profiling with cardano-ignite highlighted a performance problem tied to the Eq/Ord instance for txid.

This is the code that is causing performance problems:

```
oneEraGenTxIdRawHash :: CanHardFork xs => OneEraGenTxId xs -> ShortByteString
oneEraGenTxIdRawHash =
  hcollapse
    . hcmap proxySingle (K . toRawTxIdHash . unwrapGenTxId)
    . getOneEraGenTxId
```

Every node is expected to talk to 10s or 100s of down stream peers. All of them will present the node with mostly the same txs. To keep track of that and to acknowledge TXs already in our mempool the network layer maintains structures keyed on txid. This turns into a lot of extra allocations even today, and with Leios it will get a lot worse.

A work-around like this in ouroboros-network-api removed the issue, but it would be better if it is fixed in consensus.

```
-- | Raw byte representation of a transaction identifier.
newtype RawTxId = RawTxId ShortByteString
  deriving newtype (Eq, Ord, Show, NFData, NoThunks)

-- | Abstract over transaction identifiers,
--
-- Laws:
--
-- * If @getRawTxId x == getRawTxId y@ then @x == y@
--   (the raw bytes must uniquely identify the transaction)
class HasRawTxId txid where
  getRawTxId :: txid -> RawTxId 

```

jasagredo:
A first step towards resolving this:
- Network will define a class they can call to extract a `RawTxId`.
- Consensus will implement instances for the blocks, all defining `RawTxId ~ SL.TxId`.
- In particular there will be an instance for the HardForkBlock that Network will use to only convert once to this representation.

Doing this on top of `karknu/tx_undecision` as `js/karknu/tx_undecision`

dnadales:
# `Eq`/`Ord` for txid allocates: analysis and options

## The problem

`Eq` and `Ord` for a Cardano transaction id allocate a fresh 32-byte array on every comparison, **once per operand**.
The bytes are never needed.
The id already sits in memory as four unboxed machine words that have an allocation-free `Eq`.

### Code path

[...]

PR #2125, opened 2026-07-15
Make OneEraGenTxId Eq/Ord allocation-free (#2003)

dnadales:
Makes the hard-fork transaction id `Eq`/`Ord` comparisons allocation-free (#2003).

On `main`, the `Eq`/`Ord` instances for `OneEraGenTxId` (and hence for `TxId (GenTx (HardForkBlock xs))`) compared through `oneEraGenTxIdRawHash`, which serialised each operand's hash to a fresh 32-byte `ShortByteString`.
That allocated on every comparison, on a path the mempool and local queries exercise heavily.

## The change

The comparison now goes through two new `CanHardFork` methods, `hardForkEqGenTxId` and `hardForkCompareGenTxId`.
The `Eq`/`Ord (OneEraGenTxId xs)` instances unwrap the newtype and delegate to them, so each hard fork decides how to compare its transaction ids.
The methods have no default, so every `CanHardFork` instance must supply them.
This is a breaking change to the class.

The Cardano instance implements them without allocating, on every comparison:

- it reads each id's 32-byte hash as four 64-bit words (`Word64#`) and compares them in registers, instead of serialising to a `ShortByteString`;
- Shelley-based eras return the four words they already store, and Byron reads them out of its hash bytes in place;
- this covers same-era and cross-era pairs alike, for every era including Byron.

The other `CanHardFork` instances (the single-era instance and the two test hard forks) implement the methods with `rawHashNS`, the shared raw-hash reference.
Those still allocate, but they are not the mempool traffic that #2003 targets.

The observable semantics are unchanged: order by the hash bytes, ignoring the era.
`oneEraGenTxIdRawHash` is removed; its walk survives as `rawHashNS`, now exported so the property test uses it directly instead of a copy.

## Testing

`Test.Consensus.Cardano.TxId` (in `cardano-test`) checks that the instances agree with `rawHashNS` over every era pair, against hashes that differ at the first and last byte of each 64-bit word.
Because the Cardano comparison is no longer the reference itself, these cross-era cells now genuinely check the packed comparison against the raw-hash order.
The era-txid builders live in `unstable-cardano-testlib`, shared by the test and the benchmark.

## Micro-benchmark

`compare`/`==` across the 8 Cardano eras: one distinct-hash pair for every ordered era pair (64 cells), plus one same-hash pair per era (8 cells).
"Before" is `main` (raw hash on every comparison), "after" is this branch.
`compare` and `==` combined (`tasty-bench`, `+RTS -T`):

| class | alloc before (B) | alloc after (B) | time before (ns) | time after (ns) |
|-----------|-----------------:|----------------:|-----------------:|----------------:|
| same era  | 2676 | 80 | 408 | 75 |
| cross era | 2676 | 80 | 398 | 74 |

Representative `compare` cells (era index = position in the Cardano era list, 0 = Byron, 7 = Dijkstra):

| cell | alloc before (B) | alloc after (B) | time before (ns) | time after (ns) |
|------------------------------|-----------------:|----------------:|-----------------:|----------------:|
| `equal 0` (Byron)            | 2480 | 80 | 419 | 48  |
| `equal 7` (Dijkstra)         | 2800 | 80 | 428 | 116 |
| `0-vs-1` (Byron, Shelley)    | 2544 | 80 | 387 | 47  |
| `1-vs-2` (Shelley, Allegra)  | 2624 | 80 | 397 | 51  |
| `7-vs-0` (Dijkstra, Byron)   | 2640 | 80 | 399 | 78  |
| `7-vs-7` (Dijkstra)          | 2800 | 80 | 434 | 118 |

The 80 B after is `tasty-bench`'s fixed per-measurement floor (`Copied=0`), so the comparison itself allocates nothing, on every cell.
The generated STG confirms it: no heap object on either path, Byron included.
Comparisons run about 5x faster.

### DIFF
--- a/ouroboros-consensus-cardano/bench/txid-eq-ord-instances-bench/Main.hs
+++ b/ouroboros-consensus-cardano/bench/txid-eq-ord-instances-bench/Main.hs
@@ -1,0 +1,96 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeApplications #-}
+-- The NFData instance for 'OneEraGenTxId' below is a bench-local orphan.
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Main (main) where
+
+import Cardano.Protocol.Crypto (StandardCrypto)
+import Control.DeepSeq (NFData (..))
+import Data.ByteString.Short (ShortByteString)
+import qualified Data.ByteString.Short as SBS
+import Data.SOP (All, Proxy (..), lengthSList)
+import Data.Word (Word8)
+import Ouroboros.Consensus.Cardano.Block (CardanoEras)
+import Ouroboros.Consensus.Cardano.Node ()
+import Ouroboros.Consensus.HardFork.Combinator.AcrossEras (OneEraGenTxId)
+import Ouroboros.Consensus.Shelley.HFEras ()
+import Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+import Test.Consensus.Cardano.GenTxIdBuilders (BuildGenTxId, oneEraGenTxIds)
+import Test.Tasty.Bench (Benchmark, bench, bgroup, defaultMain, env, whnf)
+
+-- | 'OneEraGenTxId' has no 'NFData' instance, and a real one would need the
+-- raw-hash walk (or a copy of it) here, which we want to avoid. A deep force is
+-- not needed regardless: operands are built once and shared, and 'compare' and
+-- '==' re-walk each operand on every call, so their construction never enters
+-- the per-iteration figure (confirmed in the generated Core). WHNF suffices.
+instance NFData (OneEraGenTxId xs) where
+  rnf x = x `seq` ()
+
+main :: IO ()
+main =
+  defaultMain
+    [ -- 'env' builds the benchmark tree before running 'mkOperands', so the
+      -- operands are not available yet at that point. The irrefutable pattern
+      -- (@~@) avoids forcing the tuple while the tree is built; the operands are
+      -- touched only inside the 'bench' bodies, which run after 'mkOperands'.
+      env (mkOperands @(CardanoEras StandardCrypto)) $ \ ~(lhs, equalRhs, unequalRhs) ->
+        bgroup
+          "txid-eq-ord"
+          [ operationGroup "compare" compare lhs equalRhs unequalRhs
+          , operationGroup "==" (==) lhs equalRhs unequalRhs
+          ]
+    ]
+
+-- | One operation (@compare@ or @==@), laid out as @equal@ + @unequal@ cells
+-- over the era grid.
+operationGroup ::
+  String ->
+  (a -> a -> b) ->
+  [a] ->
+  [a] ->
+  [a] ->
+  Benchmark
+operationGroup name op lhs equalRhs unequalRhs =
+  bgroup
+    name
+    [ bgroup
+        "equal"
+        [ bench (show i) $ whnf (uncurry op) (lhs !! i, equalRhs !! i)
+        | i <- eras
+        ]
+    , bgroup
+        "unequal"
+        [ bench (show i ++ "-vs-" ++ show j) $ whnf (uncurry op) (lhs !! i, unequalRhs !! j)
+        | i <- eras
+        , j <- eras
+        ]
+    ]
+
+-- | Number of eras to benchmark, read from the era list type. It must come from
+-- the type, not the operand lists: the benchmark tree shape cannot depend on the
+-- runtime operands produced by 'env'.
+numEras :: Int
+numEras = lengthSList (Proxy @(CardanoEras StandardCrypto))
+
+eras :: [Int]
+eras = [0 .. numEras - 1]
+
+-- | A 32-byte hash, zero except for the given first byte. Two such hashes are
+-- equal iff that first byte matches.
+hashWithFirstByte :: Word8 -> ShortByteString
+hashWithFirstByte b = SBS.pack (b : replicate 31 0)
+
+-- | The three operand lists. @lhs@ and @equalRhs@ share a hash (first byte 0),
+-- so they compare equal; @unequalRhs@ has first byte 1, so it never does.
+mkOperands ::
+  All BuildGenTxId xs =>
+  IO ([OneEraGenTxId xs], [OneEraGenTxId xs], [OneEraGenTxId xs])
+mkOperands = pure (operandsFor 0, operandsFor 0, operandsFor 1)
+
+--                  lhs            equalRhs        unequalRhs
+
+-- | One operand per era (list index = era position) built from the given first
+-- byte.
+operandsFor :: All BuildGenTxId xs => Word8 -> [OneEraGenTxId xs]
+operandsFor b = oneEraGenTxIds (hashWithFirstByte b)
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -62,6 +62,9 @@ import Ouroboros.Consensus.Byron.ByronHFC ()
 import Ouroboros.Consensus.Byron.Ledger
 import Ouroboros.Consensus.Byron.Node ()
 import Ouroboros.Consensus.Cardano.Block
+import Ouroboros.Consensus.Cardano.CanHardFork.OptimizedTxIdComparison
+  ( compareCardanoGenTxId
+  )
 import Ouroboros.Consensus.Forecast
 import Ouroboros.Consensus.HardFork.Combinator
 import Ouroboros.Consensus.HardFork.Combinator.State.Types
@@ -230,6 +233,16 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
 
     fromTrivial :: TrivialTxMeasurePhase2 -> RefScriptSize
     fromTrivial TrivialTxMeasurePhase2 = mempty
+
+  -- Both ids are ordered by their txid hash, ignoring the era. Equality reuses
+  -- 'compare' rather than a separate path: 'hardForkEqGenTxId' is
+  -- 'compareCardanoGenTxId' returning 'EQ'. This allocates no heap: all three
+  -- 'Ordering' constructors take zero arguments, so they are never allocated on
+  -- the dynamic heap; every 'Ordering' pointer is to one of three statically
+  -- allocated addresses. 'compareCardanoGenTxId' keeps the comparison itself
+  -- allocation-free.
+  hardForkCompareGenTxId = compareCardanoGenTxId
+  hardForkEqGenTxId l r = compareCardanoGenTxId l r == EQ
 
 class TiebreakerView (BlockProtocol blk) ~ PraosTiebreakerView c => HasPraosTiebreakerView c blk
 instance TiebreakerView (BlockProtocol blk) ~ PraosTiebreakerView c => HasPraosTiebreakerView c blk
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork/OptimizedTxIdComparison.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork/OptimizedTxIdComparison.hs
@@ -1,0 +1,133 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+-- | Allocation-free cross-era comparison of Cardano transaction ids.
+--
+-- Both 'Ouroboros.Consensus.HardFork.Combinator.AcrossEras.OneEraGenTxId'
+-- instances order by the txid hash, ignoring the era. We read each id's 32-byte
+-- hash (Blake2b-256) as four big-endian 'Word64#' and compare those in
+-- registers. The words are unboxed, so no hash is boxed on the heap on any
+-- era's path:
+--
+--   * Shelley-based eras store the hash as 'PackedBytes32' (four words
+--     already); we unbox its fields.
+--   * Byron stores it as a 'ShortByteString'; we read four big-endian words
+--     out.
+--
+-- The extraction walks the era sum with a single dictionary used on every
+-- branch, so it is strict in the dictionary and allocates no per-level thunk.
+module Ouroboros.Consensus.Cardano.CanHardFork.OptimizedTxIdComparison
+  ( ToTxIdWords (..)
+  , compareCardanoGenTxId
+  ) where
+
+import Cardano.Crypto (abstractHashToShort)
+import qualified Cardano.Crypto.Hash as Hash
+import Cardano.Crypto.Hash.Class (PackedBytes (PackedBytes32))
+import Cardano.Crypto.PackedBytes (unpackBytes)
+import qualified Cardano.Ledger.Core as SL
+import qualified Cardano.Ledger.Shelley.API as SL
+import Data.Bits (unsafeShiftL, (.|.))
+import Data.ByteString.Short (ShortByteString)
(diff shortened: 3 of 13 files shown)


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -245,6 +258,122 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
     fromConway x = fromDijkstra $ DijkstraMeasure x
     fromDijkstra x = x
 
+  -- Both ids are compared by their txid hash, ignoring the era.
```
nfrisby: My initial though with this code is:

> It could probably be "simpler" but still cause the GHC optimizer to yield equivalent code to what we have here.

That would be more fragile, eg if a GHC version didn't opimitize the code as well, but you did add the benchmark to confirm no allocations---so we'd notice if that fragility struck.

On the other hand, this code as-is is not at all fragile, and it's not incomprehensible, merely fiddly.

Food for thought.

nfrisby: Also: since this code is so atypical, maybe we isolate it in its own module? `Ouroboros.Consensus.Cardano.CanHardFork.OptimizedTxIdComparison`?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -245,6 +258,122 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
...
+  -- \* Every branch of 'txIdWords' ('Z' and 'S') must use its
+  --   @All ToTxIdWords ys@ dictionary. Then GHC infers the function is
+  --   strict in the dictionary and projects the tail dictionary with a
+  --   strict 'case'. If some branch omitted it, the dictionary would be
+  --   passed lazily, and each recursive step would allocate a thunk for
+  --   the tail-dictionary projection (one per era descended).
+  --
+  -- \* The compared words are unboxed ('Word64#'), so no hash is boxed on
+  --   the heap.
+  hardForkEqGenTxId l r =
```
nfrisby: Another thought: maybe `hardForkEqGenTxId` can simply be defined as `hardForkCompareGenTxId l r == EQ` without us losing any of the performance gains?

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/GenTxIdBuilders.hs
```diff
@@ -0,0 +1,56 @@
...
+
+-- | Inject each leaf into the sum at its own era position. The strict 'NS' has
+-- no @apInjs_NP@, so we go through 'injections'.
+apInjs :: SListI xs => NP f xs -> [NS f xs]
+apInjs np = hcollapse (hap injections np)
+
+-- | One 'OneEraGenTxId' per era, list index equal to the era position, every
+-- entry carrying the given hash bytes.
+oneEraGenTxIds :: All BuildGenTxId xs => ShortByteString -> [OneEraGenTxId xs]
+oneEraGenTxIds bs = OneEraGenTxId <$> apInjs (buildNP bs)
```
jasagredo, line 56: I think this will not make it into a CAF but we maybe could do it, so that these projections are built once and exist for the duration of the program.

jasagredo (approved):

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/Consensus/Cardano/GenTxIdBuilders.hs
dnadales: As far as I can tell, turning `injections` into a CAF will require monomorphizing oneEraGenTxIds (to remove the `xs` parameter), and add `NOINLINE`. However, also as far as I understand :), `env` forces the list of txids when it sets up the benchmark (and forcing the spines of those lists requires running `injections` (off the measured loop)). So I don't think the added complexity is worth the trouble. What do you think?

ouroboros-consensus-cardano/bench/txid-eq-ord-instances-bench/Main.hs
```diff
@@ -0,0 +1,103 @@
...
+-- | Number of eras to benchmark, read from the era list type. It must come from
+-- the type, not the operand lists: the benchmark tree shape cannot depend on the
+-- runtime operands produced by 'env'.
+numEras :: Int
+numEras = lengthSList (Proxy @(CardanoEras StandardCrypto))
+
+eras :: [Int]
+eras = [0 .. numEras - 1]
+
+-- | The 32 bytes of a txid's raw hash: the role in the first byte, the rest
```
nfrisby: What does `role` means in these comments? It's not clear to me.

Perhaps there could be a comment near the top of the file (on `main`?) that explains the architecture of the benchmark and defines/exemplifies the jargon used in the rest of these comments.

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
```diff
@@ -312,6 +313,11 @@ instance
     (Z (WrapTxMeasure x)) -> translateTxMeasure x
     S (Z (WrapTxMeasure x)) -> x
 
+  -- Test-only hard fork: compare txids by their raw hash. No allocation-free
+  -- path here (that lives in the Cardano instance); the oracle is enough.
```
nfrisby: 1) "the oracle is enough" I don't understand this sentence.
2) I think a human would have written "for a test block, unoptimized comparisons is OK" and left it at that... or even no comment at all

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
```diff
@@ -93,6 +97,30 @@ class
...
+
+  -- | Order two transaction ids of @xs@. See 'hardForkEqGenTxId'.
+  hardForkCompareGenTxId ::
+    SOP.NS WrapGenTxId xs -> SOP.NS WrapGenTxId xs -> Ordering
+
+-- | The raw hash of an era sum, era ignored.
+--
+-- The reference comparison for transaction ids. Non-optimizing 'CanHardFork'
+-- instances implement 'hardForkEqGenTxId'\/'hardForkCompareGenTxId' by comparing
+-- this hash. It serialises each id via 'toRawTxIdHash', which allocates; the
```
nfrisby: could drop the "the Cardano instance overrides both methods with an allocation-free walk."

This is Claude over-commenting again.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
```diff
@@ -101,3 +129,6 @@ instance SingleEraBlock blk => CanHardFork '[blk] where
   hardForkInjectTxs = InPairs.mk1
 
   hardForkInjTxMeasure (SOP.Z (WrapTxMeasure x)) = x
+
+  hardForkEqGenTxId = (==) `on` rawHashNS
```
nfrisby, line 137: This pair _does_ probably deserve a comment: "no production code uses a single era".

And probably "TODO if some production code were to start using a single era HFC, they might eventually want to be able to opt-in to higher performance comparators"

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
```diff
@@ -165,18 +164,22 @@ instance Condense (OneEraHash xs) where
...
--- Note that this means that transactions in different eras can have equal
--- 'TxId's. This should only be the case when the transaction format is
--- backwards compatible from one era to the next.
+-- Two transactions in different eras can therefore have equal 'TxId's. This
+-- should only happen when the transaction format is backwards compatible from
+-- one era to the next.
+--
+-- The comparison itself lives in the 'CanHardFork' instance for @xs@, as
+-- 'hardForkEqGenTxId'\/'hardForkCompareGenTxId'. The Cardano instance implements
+-- them without allocating; other instances use 'rawHashNS'. We unwrap the
```
nfrisby: Can drop "We unwrap the newtype here and hand the bare 'NS' to the method."

ouroboros-consensus.cabal
```diff
@@ -925,6 +926,30 @@ benchmark PerasCertDB-bench
     tasty-bench,
     unstable-consensus-testlib,
 
+benchmark txid-eq-ord-instances-bench
+  import: common-bench
+  type: exitcode-stdio-1.0
+  hs-source-dirs: ouroboros-consensus-cardano/bench/txid-eq-ord-instances-bench
+  main-is: Main.hs
+  -- -fno-cse stops GHC merging the benchmark's two equal operand lists (built
```
nfrisby: Why would CSE spoil our benchmark?

Is our optimized comparison code short-circuiting on equal closure pointers or something?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -245,6 +248,14 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
     fromConway x = fromDijkstra $ DijkstraMeasure x
     fromDijkstra x = x
 
+  -- Both ids are ordered by their txid hash, ignoring the era. Equality is
```
nfrisby: > 'Ordering''s constructors are static

I suspect it would help to unpack that low-level detail to: "All of the 'Ordering' constructors take zero arguments, and so they're never allocated on the dynamic heap; every 'Ordering' pointer is to one of three static-allocated addresses".

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork/OptimizedTxIdComparison.hs
```diff
@@ -0,0 +1,136 @@
...
+import Ouroboros.Consensus.Ledger.SupportsMempool (GenTxId)
+import Ouroboros.Consensus.Shelley.Ledger
+import Ouroboros.Consensus.TypeFamilyWrappers (WrapGenTxId, unwrapGenTxId)
+
+-- | Order two Cardano transaction ids by their txid hash, ignoring the era.
+--
+-- 'txIdWords' reads each hash as four unboxed words and 'compareW64' orders
+-- them in registers. Two requirements keep this allocation-free:
+--
+-- * Every branch of 'txIdWords' ('Z' and 'S') must use its
```
nfrisby: Would `-fstrict-dicts` safely allow us to drop this comment?

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/TxId.hs
```diff
@@ -0,0 +1,91 @@
...
+-- | Byte positions for the probe hashes. A 32-byte hash is four 64-bit words.
+-- Each era's 'Ord' compares those words; the reference compares the raw bytes.
+-- The two agree only if each word holds its bytes in the order we expect.
+--
+-- For each position here, 'hashes' builds a hash whose bytes are all 0 except
+-- byte p, which is 1. The positions are the first and last byte of each of the
+-- four words, so if a word ordered its bytes the wrong way, two of these hashes
+-- would compare differently under the era's 'Ord' than under the reference.
+positions :: [Int]
+positions = [0, 7, 8, 15, 16, 23, 24, 31]
```
nfrisby: We can probably afford for this test to be simpler/less precise here: `[0 .. 31]`, yeah? One less mistake we might have made.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/TxId.hs
```diff
@@ -0,0 +1,91 @@
...
+import Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+import Test.Consensus.Cardano.GenTxIdBuilders (oneEraGenTxIds)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.HUnit (Assertion, assertEqual, testCase)
+
+tests :: TestTree
+tests =
+  testGroup
+    "TxIdEqOrd"
+    [ testCase "Eq/Ord agree with the raw-hash reference" $
```
nfrisby, line 57: Let's also have a QuickCheck property test that just checks `compare` and `==` on some random 32 bytes hashes.

nfrisby (changes requested):
I'm Requesting changes only for the QuickCheck property.

Otherwise, I asked for simplifications/clarifications in a few places, mostly comments.

----

edit: once the QuickCheck property exists, feel free to Dismiss my review

ouroboros-consensus-cardano/bench/txid-eq-ord-instances-bench/Main.hs
dnadales: Good point, I removed the use of the word `role` and renamed `roleBytes` to `hashWithFirstByte`, which seems clearer.

ouroboros-consensus-cardano/src/unstable-cardano-testlib/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
dnadales: Makes sense. I simplified the comment.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
dnadales: :+1:, simplified

ouroboros-consensus.cabal
dnadales: Thanks for checking :pray: It actually doesn't. I dropped this :+1:

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork/OptimizedTxIdComparison.hs
dnadales: Hmmm, maybe, but wouldn't we then need a module wide which will need its own clarifying comment?

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/TxId.hs
dnadales: Makes sense :+1:

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/TxId.hs
dnadales: Done!

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork/OptimizedTxIdComparison.hs
nfrisby: Yeah, it would still need a comment, but the `-fstrict-dicts` is a more robust mechanism than "ensure each dictionary is used in every branch".

... unless there's some _semantic_ reason why this particular code inherently must be using the dict in every branch. In which case, _that's_ what the comment should be explaining: not just the "obligation" but rather why that necessary property is certainly going to be satisfied by (any reasonable implementation of) this function.

nfrisby (approved):
New QuickCheck property test looks great! So switching from Request Changes to Approved! 👏

PR #2131, opened 2026-07-17
[Peras 30]  Tweak generic voting committee interface and implementations 

agustinmista:
This PR brings more groundwork needed in anticipation of the upcoming integration of Peras with the rest of the codebase:

- Refactors the generic voting-committee interface:
  + Moving the `VotingCommittee` data family out of the `CryptoSupportsVotingCommittee` class (simplifies future HFC instance).
  + Adding voteTarget / compareVotesById methods so vote projection and comparison are part of the class rather than passed around as functions

- Establishes a clear boundary between absolute and relative stakes: `VoteWeight` now represents a voter's stake relative to the total committee stake, enabling quorum detection against a fixed threshold (bounded between 0 and 1).

- Adds all necessary instances across committee types: `Serialise`, `NoThunks`, `Generic`, `Exception`, etc., including some orphan ones for `Array` and `Nonce`, among others.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
@@ -1,4 +1,10 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TypeApplications #-}
 
 -- | Deterministic portion of the Weighted Fait-Accompli committee selection scheme
 module Ouroboros.Consensus.Committee.WFA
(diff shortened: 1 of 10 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Committee/WFA.hs
```diff
@@ -356,8 +387,22 @@ data ExtWFAStakeDistr a
...
+  , totalStake :: TotalStake
+  -- ^ Total stake in the underlying stake distribution. This is also
+  -- precomputed at the beginning of each epoch to allow for quick
+  -- transformations between absolute and relative stakes.
   }
-  deriving Show
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NoThunks, Serialise)
+
+instance Serialise a => Serialise (Array SeatIndex a) where
```
jasagredo: I think we have avoided in many places relying on Serialise and instead use ToCBOR/FromCBOR directly. Why are we adding these instances?

agustinmista: Took a bit of time, but I removed all the `Serialise` instances in our integration branch and replace them with `FromCBOR`/`ToCBOR` ones --- I derived most via `newtype`, but some had to be manually written.

jasagredo (approved):

Issue #2116, opened 2026-07-10
OOM in Leios prototype testsuite


bladyjoker:
Observed by @johnalotoski in 
- https://github.com/IntersectMBO/ouroboros-consensus/pull/2085#issuecomment-4811247455
- https://github.com/IntersectMBO/ouroboros-consensus/pull/2113

geo2a:
I've disabled the OOMing test in https://github.com/IntersectMBO/ouroboros-consensus/pull/2129. I've opened a more specific issue to investigate the test and re-enable it once fixed: https://github.com/input-output-hk/ouroboros-leios/issues/991

PR #2133, opened 2026-07-17
Re-enable the mempool atomic QSM test

geo2a:
Fixes https://github.com/IntersectMBO/ouroboros-consensus/issues/2116.

This PR re-enables the test that used to run out of memory in CI. 

The suspected reason for this test to OOM is the way `quickcheck-state-machine` parallel testing explores the generated sequence of test steps. The memory required to explore the sequence of steps grows exponentially with the length of the sequence. On `main`, we use the default length of 100 steps, but apparently something Leios-related makes things more expensive.

I've restricted the length of the sequence of actions run by QSM to 40 steps, which locally seems to hover around 200MB max residency.

### DIFF
--- a/ouroboros-consensus.cabal
+++ b/ouroboros-consensus.cabal
@@ -773,7 +773,6 @@ test-suite consensus-test
     sop-extras,
     strict-sop-core,
     tasty,
-    tasty-expected-failure,
     tasty-hunit,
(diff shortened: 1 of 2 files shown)


ch1bo (approved):
I also did re-enable that in my high throughput mempool investigations

PR #2134, opened 2026-07-17
Do not translate transactions, instead encode and re-decode

jasagredo:
Translating a transaction means throwing away the decoded transaction and then re-decoding it in the new era. When matching transactions for the HF mempool, we now do the same but from the origin era to the target era directly, without computing it for the intermediate eras.

### DIFF
--- a/changelog.d/20260721_105834_javier.sagredo_no_tx_translate.md
+++ b/changelog.d/20260721_105834_javier.sagredo_no_tx_translate.md
@@ -1,0 +1,28 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+- Delete `hardForkInjectTxs` from `CanHardFork`.
+- Delete `TranslateEra` instances for `(GenTx :.: ShelleyBlock proto)` and `(WrapValidatedGenTx :.: ShelleyBlock proto)`.
+- Require `From|ToCBOR (GenTx blk)` in `SingleEraBlock`.
+- Delete contents of `InjectTxs.hs` to keep only `matchTx` and `rematchValidatedTxs`.
+- `TraceMempoolRemoveTxs` and `TraceMempoolManuallyRemovedTxs` now carry a `GenTx blk` instead of a `Validated (GenTx blk)`.
+
+
+### Non-Breaking
+
+- When injecting transactions from other eras into the current era in the HFC
+  mempool, encode the transaction in the origin era and decode it in the target
+  era instead of wastefully translate the transaction through all intermediate eras.
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
@@ -365,11 +365,15 @@ hardForkForgeBlock
         ([], hfs) ->
           hmap
             ( \case
-                Pair _ ApplyTxs{} -> error "Impossible!!"
+                Pair _ ApplyTxs{} ->
+                  error
+                    "Impossible! we have translated the txs to the current era, but they should already be in this era!"
                 Pair a (ReapplyTxs b) -> Pair a $ Comp $ map (\(x, (), ()) -> x) b
             )
             $ State.tip hfs
-        (_ : _, _) -> error "Impossible!!"
+        (_ : _, _) ->
+          error
+            "Impossible! some transactions were rejected as untranslatable by rematchValidatedTxs but all of them have been translated and applied just now."
 
     -- \| Unwraps all the layers needed for SOP and call 'forgeBlock'.
     forgeBlockOne ::
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -20,9 +21,6 @@ import Data.Maybe (fromMaybe)
 import Data.SOP (Top)
 import Data.SOP.BasicFunctors
 import Data.SOP.Constraint (All)
-import Data.SOP.Functors (Product2 (..))
-import Data.SOP.InPairs (InPairs)
-import qualified Data.SOP.InPairs as InPairs
 import Data.SOP.Index
 import qualified Data.SOP.Match as Match
 import Data.SOP.OptNP (NonEmptyOptNP, OptNP, ViewOptNP (..))
@@ -330,14 +328,13 @@ hardForkForgeBlock
         forgeBlockOne
         cfgs
         (OptNP.toNP blockForging)
-      $ injectValidatedTxs (map (getOneEraValidatedGenTx . getHardForkValidatedGenTx) txs)
       -- We know both NSs must be from the same era, because they were all
       -- produced from the same 'BlockForging'. Unfortunately, we can't enforce
       -- it statically.
       $ Match.mustMatchNS
         "IsLeader"
         (getOneEraIsLeader isLeader)
-        (State.tip ledgerState)
+      $ injectValidatedTxs ledgerState
    where
     cfgs = distribTopLevelConfig ei cfg
     ei =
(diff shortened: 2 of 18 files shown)


changelog.d/20260721_105834_javier.sagredo_no_tx_translate.md
```diff
@@ -0,0 +1,28 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+- Deleted `hardForkInjectTxs` from `CanHardFork`.
```
dnadales: Nit: we might want to use either past or present tense.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
```diff
@@ -351,48 +348,49 @@ hardForkForgeBlock
...
-        ([Match.Mismatch WrapValidatedGenTx f xs], NS (Product f ([] :.: WrapValidatedGenTx)) xs) ->
-        NS (Product f ([] :.: WrapValidatedGenTx)) xs
-      noMismatches ([], xs) = xs
-      noMismatches (_errs, _) = error "unexpected unmatchable transactions"
+    injectValidatedTxs st =
+      case rematchValidatedTxs getHardForkValidatedGenTx st $ map (\x -> (x, (), ())) txs of
+        ([], hfs) ->
+          hmap
+            ( \case
+                Pair _ ApplyTxs{} -> error "Impossible!!"
```
dnadales: I think this refers to the comment above (...    -- Otherwise there is a bug.), I don't know if we want to add a small comment here referring to the comment above this function.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
```diff
@@ -351,48 +348,49 @@ hardForkForgeBlock
...
+                Pair a (ReapplyTxs b) -> Pair a $ Comp $ map (\(x, (), ()) -> x) b
+            )
+            $ State.tip hfs
+        (_ : _, _) -> error "Impossible!!"
```
dnadales: This might also deserve some comments.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
```diff
@@ -351,48 +348,49 @@ hardForkForgeBlock
...
-      -- state, the Mempool maintains that invariant. That means they are
-      -- either from the same era, or can be injected into that era.
-      noMismatches ::
-        ([Match.Mismatch WrapValidatedGenTx f xs], NS (Product f ([] :.: WrapValidatedGenTx)) xs) ->
-        NS (Product f ([] :.: WrapValidatedGenTx)) xs
-      noMismatches ([], xs) = xs
-      noMismatches (_errs, _) = error "unexpected unmatchable transactions"
+    injectValidatedTxs st =
+      case rematchValidatedTxs getHardForkValidatedGenTx st $ map (\x -> (x, (), ())) txs of
+        ([], hfs) ->
```
dnadales, line 365: Would it make sense to replace `hfs` for the thing it denotes? (I mean a more descriptive variable name)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
```diff
@@ -215,28 +177,51 @@ instance
...
+              , FlipTickedLedgerState $ forgetLedgerTables st'
+              )
+       where
+        foldApplyTxs ::
+          [(GenTx blk, InputTxDiffs (HardForkBlock xs) wtd, extra)] ->
+          ( [Invalidated blk]
+          , [(Validated (GenTx blk), InputTxDiffs (HardForkBlock xs) wtd, extra)]
+          , TickedLedgerState blk ValuesMK
+          )
+        foldApplyTxs =
```
dnadales: Aren't we handing these back in reversed order here? `foldApplyTxs` prepends each survivor to `accV` and we never reverse it, whereas the `ReapplyTxs` branch gets its order restored for free by the default `reapplyTxs` (the `reverse val` in `SupportsMempool`). The note on `reapplyTxs` says the validated list has to come back in the order the txs were given, since it's later used to filter the `TxTicket`s, so I worry a dependent pair `[A, B]` (B spends A) comes back as `[B, A]` and B gets dropped on the next revalidation, or the forged block ends up invalid.

Could we reverse before returning? Something like:
```haskell
foldApplyTxs xs =
  let (accE, accV, st') = Foldable.foldl' step ([], [], st) xs
  in (reverse accE, reverse accV, st')
```
(only `accV` order is load-bearing; the `Invalidated` list order doesn't matter, so reversing it too is just for tidiness.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/InjectTxs.hs
```diff
@@ -1,190 +1,217 @@
...
+      Acc xs a b blk ->
+      [ATx xs a b xs'] ->
+      ( [(Validated (GenTx (HardForkBlock xs)), MismatchEraInfo xs)]
+      , TxsToApply a b blk
+      )
+    go' (rej, surv) [] = (L.reverse rej, toTxsToApply (L.reverse surv))
+    -- A serialised past-era transaction: upgrade it if it decodes in the tip
+    -- era, report it as untranslatable otherwise.
+    go' (rej, surv) ((Left bs, txEraNS, orig, a, b) : txs) =
+      case deserialiseFromBytes fromCBOR bs of
```
dnadales, line 164: Are we sure encode-then-decode matches the old `translateEra` at every boundary? They come apart when the tx's top-level CBOR arity changes. Concretely, for a Mary tx sitting in the mempool when the tip crosses into Alonzo:

- **Before:** `translateEra` rebuilt it as an Alonzo tx (component-wise, setting `isValid = True`), so it stayed in the mempool.
- **Now:** we encode it as Mary's 3-element record and decode with Alonzo's `fromCBOR`, which expects 4 elements (`decodeRecordNamed "AlonzoTx" (const 4)`, reading `isValid` in slot 3). That fails, so we hit the `Left _ -> reject` branch and drop the tx.

So the same Mary tx is translated under the old code and rejected under the new code. It's harmless today (Mary→Alonzo is behind us, and Conway→Dijkstra stays 4-element), so I'm not asking to change the approach. But could we record the assumption near here, that this relies on the origin and tip tx encodings sharing a top-level shape? Otherwise a future arity change would silently start dropping txs, and nothing tests for it right now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/InjectTxs.hs
```diff
@@ -1,190 +1,217 @@
...
+        Left _ -> go' ((orig, reject txEraNS) : rej, surv) txs
+        Right (_, t) -> go' (rej, (Right t, a, b) : surv) txs
+    -- A transaction already in the tip era: kept as-is.
+    go' (rej, surv) ((Right (Z v), _, _, a, b) : txs) =
+      go' (rej, (Left v, a, b) : surv) txs
+    -- A transaction from a later era: we try to downgrade it.
+    go' (rej, surv) ((Right (S tx), txEraNS, orig, a, b) : txs) =
```
dnadales, line 173: Should this branch reject the tx rather than try to downgrade it? The doc right above (lines 65-66) says a tx from a later era than the tip "is reported back as a `MismatchEraInfo`", and the old `matchPolyTx'` did exactly that: a later-era tx against a tip ledger was an unconditional era mismatch (`go _ (S x) (TZ f) = Left $ MR x f`). Here we instead encode the later-era tx and keep it whenever the bytes happen to decode under the earlier tip era.

Where this fires: a rollback across an era boundary, e.g. the mempool holds a Conway tx and the tip retracts to Babbage. The old code dropped it; now, if the Conway tx uses only Babbage-shaped fields, it decodes and we keep it, silently reinterpreting a Conway-authored tx as a Babbage one. It still has to pass tip-era validation, so I don't think it's unsafe, but it does contradict the comment just above and the previous behaviour.

Could we reject unconditionally in the `Right (S tx)` case to match the doc? Or, if the downgrade is intended, update the doc comment (and the `reject` helper's) so the two agree. There's also a related worry I haven't confirmed: reusing a later-era diff for such a downgraded tx looks like it could hit `error "anachrony"` in `ejectHardForkTxOut`, which would be a crash rather than a drop.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs
jasagredo: `hfs` = `HardForkState`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Mempool.hs
jasagredo: Nice one, thanks

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/InjectTxs.hs
jasagredo: I updated the comment. We do want to downgrade the future era txs if possible.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/InjectTxs.hs
jasagredo: Discussed it with Alexey and indeed this is a problem. He is going to implement a PR that provides us the property that decoding is translation, i.e. in AlonzoTx allow for a missing isValid flag that gets set to True.

jasagredo: The PR https://github.com/IntersectMBO/cardano-ledger/pull/5946

jasagredo: Updated the cardano-ledger pin

dnadales (approved):

PR #2136, opened 2026-07-20
[Peras 32] Define Peras IsPerasVote and IsPerasCert type classes

agustinmista:
This commit introduces the `IsPerasVote` and `IsPerasCert` type classes, which will allow us to project fields from different vote/cert container types (notably mocked and production ones).

NOTE: these type classes will replace the existing `HasPeras(Vote|Cert)X` ones in the future.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
@@ -1,0 +1,37 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | The 'IsPerasCert' projection/injection class.
+module Ouroboros.Consensus.Peras.Cert.Class
+  ( IsPerasCert (..)
+  ) where
+
+import Ouroboros.Consensus.Block.Abstract (Point)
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
```diff
@@ -0,0 +1,37 @@
...
+  ( BoostedBlock
+  , BoostedBlockCompatibleWithPoint (..)
+  , PerasRoundNo
+  )
+
+-- | Types that support being treated as Peras certificates
+class
+  BoostedBlockCompatibleWithPoint (BoostedBlock cert) blk =>
+  IsPerasCert cert blk
+    | cert -> blk
```
jasagredo, line 24: I got a brain sprain from this. Doesn't fourmolu accept `IsPerasCert cert blk | cert -> blk`?

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
agustinmista: I feel you. Sadly, fourmolu only accepts:


```haskell
class BoostedBlockCompatibleWithPoint (BoostedBlock cert) blk => IsPerasCert cert blk | cert -> blk where
```

or

```haskell
class
  BoostedBlockCompatibleWithPoint (BoostedBlock cert) blk =>
  IsPerasCert cert blk
    | cert -> blk
  where
```

The first one is way too long. The second one, while being fugly, at least fits within a reasonable width 🤷‍♂️

PR #2139, opened 2026-07-20
Hotfix: memoize bytes in ShelleyBlock

ch1bo:
Ledger's EncCBOR for SL.Block re-serializes the body structure and can produce bytes that differ from the block's original encoding (e.g. the outer body list re-emitted as definite when the producer used indefinite framing). Once a haskell relay re-encodes on disk-write, the on-disk body no longer matches the header's committed hbBodySize / hbBodyHash, and validation on reload/serve fails. Observed on Dijkstra testnet block 67555 (declared body size 66007, re-encoded body 66006) originating from a non-haskell BP behind an affected haskell relay.

Preserve the exact CBOR bytes on decode via withSlice and re-emit them verbatim via encodePreEncoded on every subsequent encode, so wire → disk → wire is byte-identical at the first haskell hop. Forged blocks serialize once and store the same bytes, so downstream encodes agree.

This is a workaround; the proper fix threads annotated bytes into Cardano.Ledger.Block and swaps encCBORGroup for encodePreEncoded there.


ch1bo:
Closed in favor of @nfrisby's hotfix (PR incoming)

PR #2140, opened 2026-07-20
Stop mangling some Dijkstra blocks (quick fix workaround for ledger Issue 5937)

nfrisby:
This PR works around Issue https://github.com/IntersectMBO/cardano-ledger/issues/5937 in the Leios testnet.

Once the long-term fix lands in our `cardano-ledger` dependency, we can remove the two "HACK" commits and wipe our nodes' state, and resync them.

In the meantime, the two HACK commits are as follows.

- Stop converting a definite-length tx seq to an indefinite length tx seq between receiving a `MsgBlock` and writing it to disk.
    - Because we serve the bytes on disk to our downstream peers, this commit also prevents us from relaying mangled blocks.
    - *But* this commit does not change the bytes of blocks that were already written to disk.
- Moreover, in order to be able to follow upstream Haskell nodes that _aren't_ yet patched, we detect the Issue 5937 bug and rectify them _when they arrive_. So if we're downstream of an unpatched node, we de-mangle their message. Which lets us stay connected to them.

The third change in this PR was an orthogonal improvement only related here to to improve the UX when testing this patch by syncing. It's the first (and by far easiest) part of Giorgos's new EB prioritization rule: use FreshestFirst among young EBs and FreshestLast among old EBs (and young EBs are prioritized over old EBs) --- for now the cutoff is 14 slots. This was included because it causes syncing nodes to fetch EBs in the order they need them, which makes the sync progress more gradual, which prevents the user from wondering whether their node is stuck.

Edit: Once all nodes are patched---and their db's wiped!---we could remove rectifier (ie the `wibble*` function). But fae3cdf733e5e8aeef1ec342f7c9d583a345a208 fix needs to stay until upstream Issue 5937 is fixed


nfrisby:
The corresponding `cardano-node` commit is:

```
$ (cd ~/cardano-node; git lg -3)
4482e0509a - (HEAD -> nfrisby/ledger-issue-5937-quickfix, origin/nfrisby/ledger-issue-5937-quickfix) HACKY HACK extra workaround cardano-ledger Issue 5937 (64 minutes ago) <Nicolas Frisby>
f4adf2717f - HACK workaround cardano-ledger Issue 5937 (8 hours ago) <Nicolas Frisby>
f7ec3592e7 - (origin/leios-prototype) Merge pull request #6621 from IntersectMBO/bladyjoker/leios-prototype/2026w29 (3 days ago) <Dražen Popović>
$ nix build --refresh 'github:IntersectMBO/cardano-node?submodules=1&ref=nfrisby/ledger-issue-5937-quickfix'
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
```diff
@@ -311,7 +319,15 @@ instance HasNestedContent f (ShelleyBlock proto era)
...
-  encCBOR = encCBOR . shelleyBlockRaw
+  encCBOR = workaroundLedgerIssue5937 . shelleyBlockRaw
+
+-- | <https://github.com/IntersectMBO/cardano-ledger/issues/5937>
+encodeShelleyBlockWorkaroundLedgerIssue5937 ::
+  (EncCBOR h, EncCBOR (SL.BlockBody era)) =>
+  SL.Block h era ->
+  Encoding
+encodeShelleyBlockWorkaroundLedgerIssue5937 (SL.Block h body) =
+  encodeListLen 2 <> encCBOR h <> encCBOR body
```
ch1bo, line 330: So this is the root fix: using `encCBOR` will use the `MemoBytes` of the body, while upstream was using `encCBORGroup` that is ignoring the memoized bytes.

Why is this hard to fix in the ledger?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -268,16 +268,30 @@ leiosFetchLogicIteration ::
...
-leiosFetchLogicIteration env offerings =
+leiosFetchLogicIteration env mbCurrentSlot offerings =
   \acc ->
     go1 acc emptyLeiosFetchDecisions $
       expand $
-        Map.toDescList $
+        prioritize $
           Map.map Left (Leios.missingEbBodies acc) `Map.union` Map.map Right (Leios.missingEbTxs acc)
  where
+  -- Fetch youngest-first among work no older than 14 slots, then oldest-first
```
ch1bo: Is this really the right policy? I thought we'd want
1. within 14 seconds from now -> the oldest EB as it's the most likely to certify
2. before 14 seconds from now -> the youngest EB as it was the most important one slot ago and bursts of old work should (ideally) not impact consensus

i.e. the highest priority EBs are 14 slots old

I might be a bit out of context on these policy discussions though..

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
ch1bo: Memoizing the bytes received like https://github.com/IntersectMBO/ouroboros-consensus/pull/2139 would be more robust here IMO

ch1bo (approved):
Seems to work and an acceptable workaround. Please see my comments though

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
lehins: > Why is this hard to fix in the ledger?

I would not say it is hard, but it is more involved that this fix. Because we must preserve previous serialization logic for pre-Dijkstra eras.
See https://github.com/IntersectMBO/cardano-ledger/issues/5937

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
nfrisby: Ah, shoot, you're right, I conflated it.

Giorgos's rule is FreshestLast among young blocks but FreshestFirst among old blocks.

I'll correct this commit. The new one won't mention Giorgos's update to the rule and instead just introduce the following: when our selection is so old that we can't determine the current slot (ie on mainnet, that'd be our selection is >36 hr old), then we fetch EBs FreshestLast order. So that the sync can be gradual. My thinking is that it's OK for a syncing node to be slowed down by a protocol burst.

This is just a workaround before we have a better EB syncing solution.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
nfrisby: > Because we must preserve previous serialization logic for pre-Dijkstra eras.

I'm not weighing in, but for clarity: this PR does preserve serialization logic for pre-Disjkstra eras---it's an override that only applies to Dijkstra.

nfrisby:
My force pushes just now make two changes:

- simplify the EB prioritization logic, as explained here https://github.com/IntersectMBO/ouroboros-consensus/pull/2140#discussion_r3622759556
- improve the commit messages on the two commits that workaround 5937

ch1bo (approved):

ch1bo:
@nfrisby I'm investigating this, potentially lucky seed, test here https://github.com/IntersectMBO/ouroboros-consensus/pull/2140/checks?check_run_id=88666825762

Could be related to the slight modification on fetching logic, but I don't expect something serious

PR #2141, opened 2026-07-21
Eject db-synthesizer and adopt cardano-config

jasagredo:
This PR removes the vendored parser for cardano configuration and instead makes use of cardano-config.

We also eject db-synthesizer which will need to be implemented downstream (cardano-api? ouroboros-consensus-tools?)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
```diff
@@ -139,64 +132,62 @@ instance HasProtocolInfo (CardanoBlock StandardCrypto) where
...
-        either (error . show) return
-          =<< Aeson.eitherDecodeFileStrict' fp
-
-    let transCfg =
+
+    -- The node configuration is parsed and resolved by the shared
+    -- 'cardano-config' package, which also loads every era's genesis file
+    -- (resolving genesis paths relative to the configuration file's directory)
+    -- and hands them back already decoded.
+    nc <- resolveNodeConfiguration configFile
```
dnadales: The commit messages say "no behavioural change", but aren't we tightening what db-analyser accepts here? `cardano-config` makes `ByronGenesisHash`, `ShelleyGenesisHash`, `AlonzoGenesisHash` and `ConwayGenesisHash` required and verified (`Protocol.hs`, `requiredFieldWith … hashCodec`), whereas the old parser read Byron/Shelley as optional and never looked at the Alonzo/Conway hashes. A config without those hashes (our own old `tools-test` `config.json` was one) used to parse and now fails.

I wonder if we should call this out, either in the PR description or a changelog note, as "configs must now carry all era genesis hashes". Do we know the configs db-analyser gets pointed at downstream always include them?

ouroboros-consensus-cardano/test/tools-test/Main.hs
```diff
@@ -1,130 +1,88 @@
...
-      =<< DBSynthesizer.initialize
-        testNodeFilePaths
-        testNodeCredentials
-        testSynthOptionsCreate
-  resultCreate <- DBSynthesizer.synthesize genTxs options protocol
-  let blockCountCreate = resultForged resultCreate
-  blockCountCreate > 0 @? "no blocks have been forged during create step"
+  resultCreate <-
+    DBSynthesizer.synthesize genTxs testSynthOptionsCreate epochSize chainDB protocol
+  resultForged resultCreate > 0 @? "no blocks have been forged during create step"
```
dnadales: With the analyse step gone, this test now only checks `resultForged > 0` per step, and I think it leaves the whole cardano-config parsing path (`resolveNodeConfiguration`, `mkHardForkTriggers`, `mkProtocolInfo`) with no in-repo caller at all, since the only remaining tool test builds the `ProtocolInfo` in-process. That path is the part this PR changes most.

I understand the config-driven pipeline is moving downstream. Still, could we keep a lighter check here that pins the count rather than just "> 0"? Something like asserting `resultForged resultCreate` equals the number of blocks we expect for 40 slots, so a regression in the forging loop does not slip through.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
```diff
@@ -139,64 +132,62 @@ instance HasProtocolInfo (CardanoBlock StandardCrypto) where
...
     -- The export path is interpreted relative to the LedgerDB filesystem root,
-    -- not the config file, so we read the config without adjusting file paths.
-    cc :: CardanoConfig <-
-      either (error . show) return
-        =<< Aeson.eitherDecodeFileStrict' configFile
+    -- not the config file, so we do not adjust file paths here.
+    nc <- resolveNodeConfiguration configFile
+    let storeCfg = Cfg.storageConfiguration nc
+        ldbCfg = runIdentity (Cfg.ledgerDbConfiguration storeCfg)
+        exportPath = case Cfg.backendSelector ldbCfg of
```
dnadales: Should we double-check this does not silently drop a configured export path? The old `mkLSMConfig` read `LedgerDB.LSMExportPath` unconditionally, but here we only surface it when the selector is `V2LSM`, and `cardano-config`'s default `Backend` is `V2InMemory` (so an absent or `V2InMemory` `Backend` discards the path during parsing).

Since the LSM backend is chosen by the `--lsm` flag rather than the config's `Backend`, it looks like `db-analyser --lsm` against a config that sets `LSMExportPath` but no `Backend: V2LSM` would now export nothing, where before it exported to that path, and without an error since the path just becomes `Nothing`.

If we want to keep the old behaviour, could we fall back to the raw `LSMExportPath` when the selector is not `V2LSM`? Something like reading it directly when `backendSelector` is `SNothing`/`V2InMemory`. Or, if coupling export to `Backend: V2LSM` is intended, maybe a note that the config now needs `Backend: V2LSM` for export to fire.

ouroboros-consensus.cabal
```diff
@@ -1855,20 +1836,6 @@ executable db-immutaliser
     ouroboros-consensus:unstable-cardano-tools,
     with-utf8,
 
-executable db-synthesizer
```
dnadales, line 1892: Now that we drop the `db-synthesizer` executable here, should we also update the docs that still point users at it? I found `cabal run db-synthesizer -- …` at `docs/website/contents/references/consensus_tools.md:371` (with its section header at `:339` and the ToC entry at `:21`), the executables list in `README.md:29`, and the nix `components.exes.db-synthesizer` attribute in `CONTRIBUTING.md:377`. The `Cardano.Tools.DBSynthesizer` library stays, so the mention in `how_to_make_a_release.md` is still accurate.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
```diff
@@ -139,64 +132,62 @@ instance HasProtocolInfo (CardanoBlock StandardCrypto) where
...
 
+-- | Parse and resolve the node configuration with the shared 'cardano-config'
+-- package. db-analyser drives the configuration from the config file alone, with
+-- no command-line overrides, so 'Cfg.resolveConfigurationFromFile' resolves it
+-- against cardano-config's all-defaults CLI arguments and the file layer wins.
+resolveNodeConfiguration :: FilePath -> IO Cfg.NodeConfiguration
+resolveNodeConfiguration configFile =
+  Cfg.resolveConfigurationFromFile configFile >>= \case
+    Left err -> error ("db-analyser: invalid node configuration: " <> show err)
+    Right (nc, _warns) -> pure nc
```
dnadales: Should we surface these warnings rather than dropping them? `resolveConfigurationFromFile` hands back `(nc, warns)`, and I think at least `MigratedToCurrentFormat` is worth showing, since any legacy config we load will trigger it and the operator otherwise has no hint their config was reshaped. Even a `traceWith`/`hPutStrLn stderr` over `_warns` would keep the signal.

cabal.project
```diff
@@ -18,8 +18,30 @@ index-state:
   -- Bump this if you need newer packages from CHaP
   , cardano-haskell-packages 2026-07-02T10:10:00Z
 
+allow-newer:
+  -- cardano-config needs trace-dispatcher-2.13 (hermod #18 ConfigSource); the
+  -- ConfigSource change is additive, so the network/diffusion stack builds against it too.
+  , *:trace-dispatcher
```
dnadales: I wonder if this needs the `*` wildcard, or whether `cardano-config:trace-dispatcher` would be enough. The comment says the diffusion stack also wants 2.13, so maybe the broad form is deliberate. If it is, no change needed; I mostly want to check we are relaxing the bound for the whole stack on purpose rather than as a side effect of getting `cardano-config` to solve.

jasagredo: This is gone

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
jasagredo: Yes, we are. And we should. Genesis files without a hash should never be used.

ouroboros-consensus-cardano/test/tools-test/Main.hs
jasagredo: I am not sure if that number is even deterministic

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
jasagredo: Very good catch. I think I am going to remove --mem and --lsm in favor of the config file

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
jasagredo: We can print them, yeah

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Block/Cardano.hs
jasagredo: Settled with
```
[ --in-mem | --lsm [--lsm-no-cache] [--lsm-export] ]
```
And if none are provided, it will read the values from the config file.

ouroboros-consensus-cardano/test/tools-test/Main.hs
jasagredo: Pinned with a test in my last commit

jasagredo:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2259

PR #2132, opened 2026-07-17
Leios: add EB announcement diffusion (but don't otherwise leverage them yet)

nfrisby:
This PR relates to https://github.com/input-output-hk/ouroboros-leios/issues/772 but doesn't close it.

- That Issue also deserves documentation update.
- This PR only adds _generation_ and _diffusion_ of announcements. It does not yet integrate them into the other Leios components that should be sensitive to them (LeiosFetch + LeiosVoting, off the top of my head). Generating and diffusing them is necessary first step.

I recommend reviewers start with these three modules---they're new, polymorphic, and entirely pure.

- `LeiosDemoLogic.Announcements`
- `LeiosDemoLogic.Announcements.ElBimap`
- `LeiosDemoLogic.Announcements.Validate` (not as polymorphic as those two, but still `blk`-agnostic)

The rest of this PR is for the most part arranging to call this core logic from the correct places---the major exception to that characterization is that some important logic is added in `LeiosDemoLogic`---it's at least starting there instead of these three modules because is _impure_.

PR #2142, opened 2026-07-21
LeiosNotify: accept but ignore proper MsgLeiosBlockAnnouncement messages

nfrisby:
- Payload is now Header blk instead of ().
- Handler merely traces a message with the header's Point instead of throwing an error.

This PR makes a tiny change that will allow use to run the much more significant patch in PR https://github.com/IntersectMBO/ouroboros-consensus/pull/2132 in the testnet without constant disconnections, which is desirable for validating that PR before merging it.


nfrisby:
This PR alters the type of some tracer events, so it needs a corresponding change in `cardano-node`. That PR is here https://github.com/IntersectMBO/cardano-node/pull/6624

ch1bo (approved):

PR #2144, opened 2026-07-23
Fix EB diffusion of already complete closures

ch1bo:
Notify `AcquiredEBTxs` if an already complete closure for EB body is inserted

This was preventing offers of EBs that we already have the full closure
of and resulted in EBs not being diffused.

Reproducible in Leios threadnet test with seed `(SMGen 10792080524324648525 17850122056530710967,60)` and it was noticed on https://github.com/IntersectMBO/ouroboros-consensus/pull/2140/checks?check_run_id=88666825762 of #2140 PR

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
@@ -102,10 +102,18 @@ data LeiosDbConnection m = LeiosDbConnection
   -- ^ Read the EB "body": the ordered list of tx-hash + tx-byte-size
   -- pairs that constitute this EB. No tx bytes are fetched; contrast
   -- with 'leiosDbLookupEbClosure' which joins with the 'txs' table.
-  , leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m ()
+  , leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m CompletedEbs
   -- ^ Persist an EB body. The point MUST already have been inserted
   -- via 'leiosDbInsertEbPoint' (announcement path). Yields an
   -- 'AcquiredEb' notification.
+  --
+  -- Returns any EBs whose closure just became complete because their body
+  -- landed after all their txs were already present in the DB (typical
+  -- when mempool-relayed txs beat the EB body). Those EBs also get an
+  -- 'AcquiredEbTxs' notification. Callers that need the trace-level
+  -- signal ('TraceLeiosBlockTxsAcquired') should emit it from this
+  -- result — for the common case where the body arrives first, the
+  -- returned list is empty.
   , leiosDbInsertTxs :: HasCallStack => [(TxHash, ByteString)] -> m CompletedEbs
   -- ^ Insert transactions into the global 'txs' table (INSERT OR IGNORE).
   -- After inserting, checks which EBs referencing these txs are now complete
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
@@ -114,6 +114,8 @@ data LeiosDbConnection m = LeiosDbConnection
   -- signal ('TraceLeiosBlockTxsAcquired') should emit it from this
   -- result — for the common case where the body arrives first, the
   -- returned list is empty.
+  --
+  -- XXX: return type only used for tracing
   , leiosDbInsertTxs :: HasCallStack => [(TxHash, ByteString)] -> m CompletedEbs
   -- ^ Insert transactions into the global 'txs' table (INSERT OR IGNORE).
   -- After inserting, checks which EBs referencing these txs are now complete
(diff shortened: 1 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -102,10 +102,18 @@ data LeiosDbConnection m = LeiosDbConnection
   -- ^ Read the EB "body": the ordered list of tx-hash + tx-byte-size
   -- pairs that constitute this EB. No tx bytes are fetched; contrast
   -- with 'leiosDbLookupEbClosure' which joins with the 'txs' table.
-  , leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m ()
+  , leiosDbInsertEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m CompletedEbs
```
nfrisby, line 105: Why's it a list? We're only inserting one EB body, right?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -660,8 +660,16 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
         -- the point idempotently as a stop-gap and trace a warning.
         traceWith ktracer $ TraceLeiosBlockPointMissing point
         leiosDbInsertEbPoint db point ebBytesSize
-        leiosDbInsertEbBody db point eb
+        completedByBody <- leiosDbInsertEbBody db point eb
         traceWith ktracer $ TraceLeiosBlockAcquired point
+        -- 'leiosDbInsertEbBody' returns a non-empty list only when this
+        -- body's arrival is what completed the closure — every referenced
+        -- tx was already in the DB (typical for the forger's own txs
```
nfrisby: This is the second comment that claims txs arriving via the Mempool mean they don't need to arrive via LeiosFetch. That's true for our ultimate design, but I didn't think the prototype was already doing that. Is Claude just optimistically assuming it is when deciding to mention this in these comments?

The event can happen without the Mempool being involved: it just requires that every tx in this EB was also in some older EB.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -298,7 +303,26 @@ sqlInsertEbBody tracer Conn{connDb = db, connStmts = Stmts{stInsertEbTxsRow, stI
...
       dbStep1 stInitMissingCount
+    -- If every referenced tx is already present, 'stInitMissingCount' just
+    -- set missingTxCount to 0 for this EB. No subsequent 'sqlInsertTxs'
+    -- will fire for this point, so drain the pending-completion set here
+    -- (mirrors what 'sqlInsertTxs' does after 'stDecrMissingCount').
+    completedNow <- useStmt stFindCompleteEbs $ findCompleteEbs stFindCompleteEbs
+    useStmt stMarkNotifiedEbs $ dbStep1 stMarkNotifiedEbs
+    pure completedNow
   notify $ AcquiredEb point ebBytesSize
+  forM_ completed $ \p -> notify (AcquiredEbTxs p)
```
nfrisby: We don't need the `novel` check here? Seems plausible that we don't, but might deserve a brief explanatory comment.

... Also seems plausible that maybe a race between multiple clients downloading the same tx that overlaps with this EB .... blah blah ... might cause a duplicate? Or is the insertion here _and also_ the marking as complete happening inside a single SQL transaction?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
nfrisby: Ah, it's because the SQLite impl is reusing machinery from the tx insertion, which does possibly complete multiple EBs at once. Eh 🤷

nfrisby (approved):
LGTM

one minor question about a risk of duplicate notifications, but I'm guessing there isn't one and I'm just asking for a brief code comment explaining why there's no risk

edit: ah, and also some comments claiming LeiosFetch is already checking the Mempool, but I don't think that's already happening?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
ch1bo: In fact, we only need the return value for tracing. I think it would be cleaner to pass a tracer to this API eventually anyways

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: The code grew quite contrived, I'm sorry about that, but the marking is happening in the same transaction, yes. I renamed `dbWithBEGIN` to `dbWithTransaction` and updated the comment.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
ch1bo: I hate the unnecessary verbosity of commentary, especially when it's incorrect. Sorry for the noise

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
ch1bo: In fact, I now even changed the code not to re-use the "completed EB scan", because it's only one EB (select of one row) that can be completed. The return type is still weird

PR #2145, opened 2026-07-23
[Peras 35] Tweak PerasBLSCrypto and add several instances

agustinmista:
This PR simplifies the Peras public and private keys by reusing the same BLS key for both signing and VRF roles. In addition, it provides a couple of extra instances for the types defined in that module that will be needed later, notably `NoThunks` and `FromCBOR`/`ToCBOR`.

Moreover, we took the opportunity to replace the old `PerasBoostedBlock` from `O.C.Block.SupportsPeras` with the one from `O.C.Peras.Types`, which notably just adds a `WithOrigin` wrapper.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
@@ -1,4 +1,7 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE EmptyDataDeriving #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
```diff
@@ -59,28 +71,50 @@ import qualified Ouroboros.Consensus.Committee.Crypto.BLS as BLS
 
 -- | BLS-based crypto scheme used in Peras voting committees
 data PerasBLSCrypto
+  deriving (Show, Eq, Generic, NoThunks)
```
jasagredo, line 77: Why do we need these? Some other data like this could just be `type data` no?

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Crypto/BLS.hs
agustinmista: We have a couple of type class instances that won't compile without these for some reason (even with `type data`). I created https://github.com/tweag/cardano-peras/issues/272 to keep track of this, and added a comment in the code.

PR #2146, opened 2026-07-23
[Peras 36] Tweak V1 PerasVote and PerasCert

agustinmista:
This commit tweaks the V1 `PerasVote` and `PerasCert` (i.e. the ones to be used in production) to carry a phantom type parameter "tag". This will be needed in the future to enforce injectivity in the `BlockSupportsPeras` type class without using yet another type wrapper.

In addition, it adds a couple of type classes that will be needed in the near future: `NoThunks`, `ShowProxy`, `IsPerasCert`/`IsPerasVote`.

Finally, it finishes the transition away from `O.C.Peras.Voting.Committee` by moving the committee conversion type classes next to their corresponding vote/cert definitions.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -24,30 +35,55 @@ import Cardano.Binary
...
 import qualified Ouroboros.Consensus.Util.Bitmap as Bitmap
+import Ouroboros.Network.Util.ShowProxy (ShowProxy (..))
 
 -- | Concrete Peras certificates using BLS signatures
-data PerasCert
+--
+-- NOTE: the 'tag' parameter is a phantom type used to track the block type that
+-- the certificate is associated with, to ensure injectivity when 'V1.PerasCert'
+-- is used as a type instance for 'BlockSupportsPeras' class.
+data PerasCert tag
```
jasagredo: I see you seem to use `blk` for this below. Why not use `blk` here (and everywhere else)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -59,9 +95,28 @@ data PerasCert
...
+  ( ConvertRawHash blk
+  , HashSize blk ~ 32
+  ) =>
+  IsPerasCert (PerasCert blk) blk
+  where
+  getPerasCertRound = pcRoundNo
+  getPerasCertBlock = pcBoostedBlock
+
+instance
+  Typeable tag =>
```
jasagredo: Why do you need typeable? isn't that only needed for deriving instances and similar?

jasagredo (commented):
Same questions apply to Votes

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/V1.hs
```diff
@@ -20,20 +29,43 @@ import Cardano.Binary
...
+  )
+import Ouroboros.Network.Util.ShowProxy
 
 -- | Concrete Peras votes using BLS signatures
-data PerasVote
+--
+-- NOTE: the 'tag' parameter is a phantom type used to track the block type that
+-- the vote is associated with, to ensure injectivity when 'V1.PerasVote' is
+-- used as a type instance for 'BlockSupportsPeras' class.
+data PerasVote tag
```
jasagredo: Ah also this needs a changelog entry.

jasagredo (changes requested):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: `Typeable` is a superclass constraint of `FromCBOR`/`ToCBOR` from `cardano-binary`: 

https://github.com/IntersectMBO/cardano-base/blob/master/cardano-binary/src/Cardano/Binary/FromCBOR.hs#L89

jasagredo (approved):

PR #2153, opened 2026-07-27
[Peras 40] Introduce LedgerStateSupportsPeras auxiliary type class

agustinmista:
This commit introduces an auxiliary `LedgerStateSupportsPeras` type class to aid with extracting the stake distribution and Peras parameters from either a ticked or unticked `LedgerState` in a consistent manner across the codebase.

NOTE: until the Peras parameters can be defined via on-chain governance, the default `getPerasParams` implementation returns `defaultPerasParams`.

NOTE: the existing `LedgerSupportsPeras` class is flagged for removal, as the cached latest Peras cert round is being moved from the (non-extended) ledger state into the extended one.

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
@@ -67,8 +67,10 @@ import NoThunks.Class (NoThunks (..))
 import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.HardFork.Combinator
   ( HasPartialConsensusConfig
+  , LedgerState
   )
 import Ouroboros.Consensus.HeaderValidation
+import Ouroboros.Consensus.Ledger.SupportsPeras (LedgerStateSupportsPeras)
 import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.Protocol.Praos.Common
   ( PraosTiebreakerView
@@ -131,11 +133,13 @@ class
   , -- Backwards compatibility
     Plain.FromCBOR (LegacyPParams era)
   , Plain.ToCBOR (LegacyPParams era)
-  , -- TODO: replace the two constraints below with:
+  , -- TODO: replace the four constraints below with:
     -- 'StateSupportsPerasEpochContext (ShelleyBlock proto er)' once that type
     -- class is in place.
     ChainDepStateSupportsPeras (ChainDepState (BlockProtocol (ShelleyBlock proto era)))
   , ChainDepStateSupportsPeras (Ticked (ChainDepState (BlockProtocol (ShelleyBlock proto era))))
+  , LedgerStateSupportsPeras (LedgerState (ShelleyBlock proto era))
+  , LedgerStateSupportsPeras (Ticked LedgerState (ShelleyBlock proto era))
   ) =>
   ShelleyCompatible proto era
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
@@ -1,14 +1,31 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE OverloadedStrings #-}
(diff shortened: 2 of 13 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
```diff
@@ -131,11 +134,13 @@ class
...
     Plain.FromCBOR (LegacyPParams era)
   , Plain.ToCBOR (LegacyPParams era)
-  , -- TODO: replace the two constraints below with:
+  , -- TODO: replace the four constraints below with:
     -- 'StateSupportsPerasEpochContext (ShelleyBlock proto er)' once that type
     -- class is in place.
     ChainDepStateSupportsPeras (ChainDepState (BlockProtocol (ShelleyBlock proto era)))
   , ChainDepStateSupportsPeras (Ticked (ChainDepState (BlockProtocol (ShelleyBlock proto era))))
+  , forall mk. LedgerStateSupportsPeras (LedgerState (ShelleyBlock proto era) mk)
+  , forall mk'. LedgerStateSupportsPeras (Ticked LedgerState (ShelleyBlock proto era) mk')
```
jasagredo: Quatified constraints, oh no.

jasagredo: I think the quantified constraints can be unquantified after #2079

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
```diff
@@ -1,18 +1,80 @@
...
 class LedgerSupportsPeras blk where
   -- | Extract the round number of the latest Peras certificate stored in the
   -- given ledger state (if any). This is needed to coordinate the end of a
   -- cooldown period.
   getLatestPerasCertRound :: LedgerState blk mk -> Maybe PerasRoundNo
   default getLatestPerasCertRound :: LedgerState blk mk -> Maybe PerasRoundNo
   getLatestPerasCertRound _ = Nothing
+
+-- | Extract Peras information stored in the ledger state.
+class LedgerStateSupportsPeras ledgerState where
```
jasagredo, line 38: I actually think you could define this as

```
class LedgerStateSupportsPeras l blk where
  getPoolDistr :: l blk EmptyMK -> ...

instance LedgerStateSupportsPeras ExtLedgerState blk where
  getPoolDistr = getPoolDistr . ledgerState
```

Then you need to define instances for `LedgerState` and `Ticked LedgerState` for each block as you are already doing, but we get rid of the quantified constraints

agustinmista: IIRC, we initially tried this way, and had to backtrack to the quantified constraint because one of the call sites (`applyChainTickLedgerResult`) only seems to provide a `tickedLedgerState :: Ticked LedgerState blk DiffMK`.

Naturally, Peras doesn't depend on `mk` at all, so we _could_ force `LedgerStateSupportsPeras` to use `EmptyMK` if you know there's some way to cast between these two types :thinking:

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
agustinmista: > I think the quantified constraints can be unquantified after https://github.com/IntersectMBO/ouroboros-consensus/pull/2079

Oh, this would be fantastic indeed. 

Maybe it's worth waiting for that refactor to eliminate the problem completely before taking any action (as opposed to looking for a workaround now, and then having to refactor it away anyway).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsPeras.hs
agustinmista: After some discussion, we moved with a slightly different approach that avoids all quantified constraints:

```haskell
class LedgerStateSupportsPeras ledgerState where
  getPoolDistr :: ledgerState EmptyMK -> PoolDistr

instance LedgerStateSupportsPeras (LedgerState (ShelleyBlock proto era)) ...
instance LedgerStateSupportsPeras (Ticked LedgerState (ShelleyBlock proto era)) ...

jasagredo (approved):

PR #2154, opened 2026-07-28
[Peras 41] Tweak PerasCertDB to return status of latest certificate seen

agustinmista:
This commit extends the `PerasCertDB.getLatestCertSeen` method to return the status of the block being boosted by the certificate.

This is relevant for Peras voting because we need to know if a candidate block extends the chain that contains the block being boosted by the latest certificate seen. Since this block might no longer be part of the current volatile suffix, the PerasCertDB now keeps track of whether the certificate was garbage collected or not, which in turn would indicate that the block has entered the immutable prefix, thus rendering the ancestor check mentioned above vacuously true.

In addition, it updates the ChainDB model and state-machine tests to abide by the new requirement.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE TypeOperators #-}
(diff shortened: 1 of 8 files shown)


jasagredo (approved):
:+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
```diff
@@ -282,14 +270,16 @@ mkPerasVotingView
...
     -- the volatile suffix is covered by also providing the status of the
-    -- boosted block w.r.t. the chain's immutable prefix and volatile suffix.
-    candidateBlockExtendsCert (CertWithImmutableBlock _) =
-      -- This case is vacuously true: an immutable block is always part of
-      -- any volatile suffix, so the candidate block trivially extends it.
+    -- boosted block w.r.t. the volatile suffix.
+    candidateBlockExtendsCert (CertBoostingBlockNoLongerInVolatileDB _) =
+      -- This case is vacuously true: the boosted block is from a slot that has
+      -- already been garbage collected from the volatile suffix, which implies
+      -- that it extends any volatile suffix.
```
jasagredo, line 277: I wonder if this justification is correct. It might be from a different fork, no? Or we always consider only ancestors?

agustinmista: I will try to explain the situation a bit to give you a better picture:

As per the voting rules, the candidate block must be a direct ancestor of the the one boosted by the latest certificate seen. Moreover, this check is only relevant to keep the "happy path" of the protocol going (more especifically, VR-1B), where we expect nodes to continue voting after every (successful) election. Because of this, we would expect the block that was most recently boosted to still be in the volatile suffix, as it most likely was forged not _too_ long ago (the actual age is ultimately up to parameterization).

However, because boosting blocks also shortens the volatile suffix a bit due to weighted chain selection, we are trying to be extra careful here: if the block boosted by the latest certificate seen was garbage collected, we can almost certainly assume* that it was because it became immutable, and thus a direct ancestor of any volatile suffix.

[*] Though I think it could be problematic if the parameterization is absolutely borked with a huge certificate boost and a very short block age threshold.

(I hope this makes at least some sense 😅 )

PR #2155, opened 2026-07-28
[Peras 42] Rework Peras round tracking in HardFork summary queries 

agustinmista:
This PR replaces the optional `boundPerasRound :: PerasEnabled PerasRoundNo` field on `Bound` with a mandatory `boundNextPerasRound :: PerasRoundNo`. This value now always exists but only advances in Peras-enabled eras, staying constant across non-Peras eras.

This commit reworks how Peras rounds are tracked and queried in the HardFork history, split across two closely related groups of changes:

- Summary: change how Peras rounds are stored on era bounds
  - Replace the optional `boundPerasRound :: PerasEnabled PerasRoundNo` field on Bound with a mandatory `boundNextPerasRound :: PerasRoundNo`. The value now always exists but only advances in Peras-enabled eras, staying constant across non-Peras eras. `mkUpperBound` and `initBound` are updated accordingly.
  - Simplify the CBOR encoding of Bound to a fixed 4-element list (updating the golden files accordingly).
  - Add eraEndToBound and summaryPerasBounds helpers.

- Qry: rework Peras translations and add new queries
  - Rework the Peras round translations to carry `SlotInPerasRound`, introduce `ensurePerasEnabled`, and use `boundNextPerasRound` for the era-boundary guards, keeping the lower/upper bound checks in the `RelToAbs`/`AbsToRel` steps.
  - Add `slotToPerasRoundNo'`, `perasRoundNoToPerasRoundLength`, `epochToPerasRoundInfo`, and the `EpochToPerasRoundInfo` bundle type.
  - Tests: update golden files, tweak round trip conversion tests add sensible preconditions to `eventPerasRoundNoToSlot` test.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Summary.hs
```diff
@@ -526,26 +569,19 @@ invariantSummary = \(Summary summary) ->
...
-    len <- decodeListLen
+    enforceSize "Bound" 4
     boundTime <- decode
     boundSlot <- decode
     boundEpoch <- decode
-    boundPerasRound <- case len of
-      3 -> pure NoPerasEnabled
-      4 -> PerasEnabled <$> decode
-      _ -> cborError (DecoderErrorCustom "Bound" "unexpected list length")
+    boundNextPerasRound <- decode
```
jasagredo: This will cause a replay on all nodes for something that is not even enabled right now on mainnet. We should make this backwards compatible on the decoder so that an old snapshot can still be read.

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
```diff
@@ -513,6 +582,11 @@ data EventType
     Confirm EpochNo
   deriving Show
 
+-- initEventTimePeras :: HF.EraParams -> EventTimePeras
+-- initEventTimePeras HF.EraParams{eraPerasRoundLength} = case eraPerasRoundLength of
+--   NoPerasEnabled -> EventNoPerasEnabled (PerasRoundNo 0)
+--   PerasEnabled _ -> EventPerasEnabled (PerasRoundNo 0) 0
```
jasagredo: Delete?

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
```diff
@@ -733,7 +806,7 @@ genEvents = \(Eras eras) (HF.Shape shape) -> sized $ \sz -> do
             , eventEra = era
             , eventEraParams = eraParams
             }
-    (event :) <$> go (n - 1) (stepTime typ time)
+    (event :) <$> (go (n - 1) (stepTime typ time))
```
jasagredo: redundant?

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
agustinmista: Good catch :sweat_smile:

ouroboros-consensus/test/consensus-test/Test/Consensus/HardFork/History.hs
agustinmista: Also good catch, must've slipped through the cracks.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Summary.hs
agustinmista: I [updated](https://github.com/IntersectMBO/ouroboros-consensus/pull/2155/changes#diff-70b57fd5d65590f4cac9f806fd6d94fcb50c194821ca63c478902d13249ef45aL545-R600) the decoder to retain the backwards compatibility, though the change of semantics requires a bit of justification. Hope you find this better :smile_cat: 

BTW, we also need to to make a couple of changes in the serialization of the ledger states in the near future:
 - Move the round of the latest cert on chain from  `ShelleyLedgerState` into `ExtLedgerState`
 - Add the upcoming Peras context resolver to `ExtLedgerState`.

Should we coordinate these in the same backwards compatible way too? (Hopefully not :sweat_smile:)

jasagredo: Only if we are not forced to do a different ledgerstate change that justifies a replay (Peras doesn't justify it yet as it is not enabled on mainnet)

jasagredo (approved):

PR #2156, opened 2026-07-28
db-analyser: Ability to turn off caching for --lsm

mgmeier:
# Description

* `ouroboros-consensus-lsm`: expose `LSM.DiskCachePolicy` as a parameter
* `db-analyser`: allow disabling the disk cache for `--lsm` via additional flag `--lsm-no-cache`

_Summary_

The LSM ledger DB backend previously always used `LSM.newTable`/`LSM.openTableFromSnapshot` with their default disk cache policy, so the UTxO table's disk cache behavior was hardcoded and not observable/tunable from outside the library.

_This PR:_

- Threads a `LSM.DiskCachePolicy` through `LSMArgs`, `Resources LSM`, `mkLSMArgsIO`, `tableFromValuesMK`, and `loadSnapshot`, using `LSM.newTableWith`/`LSM.openTableFromSnapshotWith` variants, which allow for overriding the disk cache policy. `LSM.DiskCachePolicy` is now re-exported from the `LSM` module.
- Adds a `--lsm-no-cache switch` to `db-analyser`, which, when combined with` --lsm`, sets the policy to `LSM.DiskCacheNone` (bypassing the OS page cache / using `O_DIRECT` for UTxO table reads and writes) instead of the default `LSM.DiskCacheAll`. This has no effect on other backends and is intended primarily for benchmarking disk I/O behavior in isolation from page-cache effects.
- Updates the `LSMArgs` call sites in the state-machine and ledger-snapshot test suites to pass an explicit cache policy (`DiskCacheAll`, preserving existing behavior).
- Does **not** create any code path for propagating some cache policy from a Node config into consensus.

_Motivation_

For benchmarking the LSM backend's actual disk performance (e.g. comparing read/write throughput or measuring the true cost of table operations), the OS page cache can mask I/O costs by serving repeated reads from memory. Making the disk cache policy configurable lets `db-analyser` runs isolate on-disk performance from cache effects.

For a full rationale see https://github.com/input-output-hk/ouroboros-consensus-tools/pull/12 and the `beacon/docs/METHODOLOGY.md` document it adds.

### DIFF
--- a/changelog.d/20260728_190611_michael.karg_lsm_disk_cache_policy.md
+++ b/changelog.d/20260728_190611_michael.karg_lsm_disk_cache_policy.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- Expose `LSM.DiskCachePolicy` as a parameter of the `LSM` ledger DB backend (`LSMArgs`), allowing callers to control whether UTxO table reads/writes go through the OS page cache.
+
+<!--
+### Patch
+
+- A bullet item for the Patch category.
+
+-->
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Types.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Types.hs
@@ -16,6 +16,10 @@ data DBAnalyserConfig = DBAnalyserConfig
   , analysis :: AnalysisName
   , confLimit :: Limit
   , ldbBackend :: LedgerDBBackend
+  , lsmNoDiskCache :: Bool
+  -- ^ When using the 'V2LSM' backend, bypass the OS page cache for UTxO
+  -- table reads/writes (instead caching all). Intended for benchmarking;
+  -- has no effect with any other backend.
   }
 
 data AnalysisName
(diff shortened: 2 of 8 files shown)


changelog.d/20260728_190611_michael.karg_lsm_disk_cache_policy.md
```diff
@@ -0,0 +1,24 @@
...
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- Expose `LSM.DiskCachePolicy` as a parameter of the `LSM` ledger DB backend (`LSMArgs`), allowing callers to control whether UTxO table reads/writes go through the OS page cache. `db-analyser` gains a `--lsm-no-cache` switch to bypass the cache when using `--lsm`, primarily for benchmarking.
```
jasagredo: We don't mention unstable/tools changes in the changelog.
```suggestion
- Expose `LSM.DiskCachePolicy` as a parameter of the `LSM` ledger DB backend (`LSMArgs`), allowing callers to control whether UTxO table reads/writes go through the OS page cache.
```

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Types.hs
```diff
@@ -16,6 +16,10 @@ data DBAnalyserConfig = DBAnalyserConfig
   , analysis :: AnalysisName
   , confLimit :: Limit
   , ldbBackend :: LedgerDBBackend
+  , lsmNoDiskCache :: Bool
```
jasagredo: Can't this be included in `LedgerDBBackend` above to only parse it if we are using `lsm`?

changelog.d/20260728_190611_michael.karg_lsm_disk_cache_policy.md
mgmeier: Fixed using your suggestion.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Types.hs
Russoul: Addressed 👍

jasagredo (approved):

PR #2164, opened 2026-07-30
[Peras 44] Introduce Peras epoch context resolution machinery

agustinmista:
This PR introduces the `Ouroboros.Consensus.Peras.Context` module, providing the machinery needed to resolve Peras round numbers into their epoch-dependent context (voting committee and Peras parameters):

- `BoundedPerasEpochContext`: a `PerasEpochContext` tagged with the range of round numbers over which it is valid, plus the `withinEpochContext` lookup.
- `PerasEpochContextResolver`: a two-epoch (current + previous) sliding window of epoch contexts that resolves a round number to its context, with a dedicated error state for recoverable failures. `initPerasEpochContextResolver` and `tickPerasEpochContextResolver` build and advance the window (at most one epoch per step), and `StateSupportsPerasEpochContext` / `mkBoundedPerasEpochContextWith` extract contexts from the node state.
- `PerasEpochContextResolverHandle` (with a mock implementation) and `resolveRoundNo` / `withResolvedRoundNo` for consuming resolved contexts in `STM`.
- `TimeResolutionContext` and helpers (`runQueryWithContext`, `runQueryEraIndexedWithContext`, and their handle-based variants) to run hard fork history queries against the ledger state.

The changes in `Block.SupportsPeras` reorganise the module exports and introduce the `PerasEpochContext` type together with the `PerasCrypto`/`PerasVotingCommitteeScheme`/`PerasError` associated types in anticipation of an upcoming refactor of the `BlockSupportsPeras` type class.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
@@ -1,0 +1,963 @@
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Ouroboros.Consensus.Peras.Context
+  ( -- * Bounded Peras epoch context
+    BoundedPerasEpochContext (..)
+  , withinEpochContext
+
+    -- * Peras epoch context resolver and handle
+  , PerasEpochContextResolver (..)
+  , PerasEpochContextNotFoundForRound (..)
+  , resolveRoundNo
+  , perasEpochContextResolverBounds
+  , PerasEpochContextResolverHandle (..)
+  , mockPerasEpochContextResolverHandle
+  , withResolvedRoundNo
+
+    -- * Extracting and resolving Peras epoch contexts from the node state
+  , StateSupportsPerasEpochContext (..)
+  , mkBoundedPerasEpochContextWith
+  , initPerasEpochContextResolver
+  , tickPerasEpochContextResolver
+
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
```diff
@@ -0,0 +1,939 @@
...
+  , readTVar
+  , throwSTM
+  )
+
+-- * Bounded Peras epoch context
+
+-- | A 'PerasEpochContext' that is valid only in a given range of round numbers
+data BoundedPerasEpochContext blk
+  = BoundedPerasEpochContext
+  { startPerasRoundNo :: PerasRoundNo -- inclusive lower bound
```
jasagredo: Can make this a haddock with `-- ^` no?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
```diff
@@ -0,0 +1,939 @@
...
+  deriving (Eq, Show, Generic, NoThunks, Exception)
+
+-- | Initialise a 'PerasEpochContextResolver' using a bounded context
+newPerasEpochContextResolver ::
+  PerasEnabled (BoundedPerasEpochContext blk) ->
+  PerasEpochContextResolver blk
+newPerasEpochContextResolver currEpochContext =
+  PerasEpochContextResolver
+    currEpochContext
+    NoPerasEnabled
```
jasagredo, line 343: why is this hardcoded?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
```diff
@@ -0,0 +1,939 @@
...
+        -- possible when the chain skips one or more entire epochs (e.g. a
+        -- sparse/empty chain, or a node ticking its ledger far ahead of its
+        -- tip). The two-epoch window cannot be advanced incrementally
+        -- across the gap, so we re-initialise it at the target epoch.
+        --
+        -- NOTE: this should be an impossible case in production, as we
+        -- expect at least one block per epoch. However, we still need to
+        -- handle this case in a way that doesn't break the more lenient
+        -- test suites, where we might tick the ledger far ahead of its tip.
+        -- See: https://github.com/tweag/cardano-peras/issues/260
```
jasagredo, line 748: Nice!

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Context.hs
agustinmista: This function is only used in cases where we don't have information about the previous epoch:

* When creating a new context resolver from scratch, where we don't have any information about Peras for the previous epoch (there might not even be a previous epoch if we are boostrapping from Genesis)

* When ticking the resolver out of an error state. This could occur due to bad parameterization leading to an error while creating a voting committee for a given epoch. At some point we discussed making the "Peras is enabled in this epoch after recovering from an error in the previous epoch" state representable, but for now we don't think this would add too much value, so we just turn the error into a `NoPerasEnabled` upon recovery.

* When ticking the resolver more than one epoch at a time, in which case we cannot faitfully build a context for the previous (gap) epoch. This only happens in tests, though.

jasagredo: Aha, perhaps we could write this down in the code?

PR #2167, opened 2026-07-31
Allow hardfork into Dijkstra

jasagredo:


nfrisby (approved):
Thanks, nice little logic to prevent this mistake (which we've probably made more than once).

Could also go further and use a similar type family to prevent non-`Last` era's from using the wrong code, but that might be overkill. (And the not-equal type family logic is more involved than the equal logic.)

PR #2175, opened 2026-07-31
nix: keep the dev shell available on non-Linux platforms

Russoul:
## Summary

`nix develop` (and `devShells.default`/`ghc96`) is currently broken on non-Linux platforms (e.g. macOS/`aarch64-darwin`):

```
error: attribute 'devShell' missing
at flake.nix:79:19:
    78|           default = ghc96;
    79|           ghc96 = hydraJobs.native.haskell96.devShell;
      |                   ^
```

This was introduced by the `exesOnly` CI-cost-reduction path added in `nix/ci.nix` (to only build the distributed executables on macOS/Windows-cross and skip libraries/tests/benchmarks). As written, `exesOnly` also happened to drop the `devShell` output entirely, but `flake.nix` unconditionally references `hydraJobs.native.haskell96.devShell`, so the flake fails to evaluate a dev shell at all on those platforms.

## Fix

- `mkHaskellJobsFor` now always attaches `devShell` when not cross-compiling, regardless of `exesOnly`.
- Since `exesOnly` exists specifically to reduce what Hydra is forced to build on non-Linux, a new `jobsForRequired` strips `devShell` back out of the `required` aggregate on non-Linux platforms, so Hydra's build set is unchanged from before this PR. The dev shell is only ever *evaluated*, not built by CI, on those platforms — same as today.

Verified via `nix eval` that on `aarch64-darwin`:
- `hydraJobs.native.haskell96` still contains `devShell`.
- `hydraJobs.required.native.constituents` is unchanged (still just the distributed executables, no dev shell).

And via `nix develop` locally on macOS, which now succeeds instead of failing at evaluation.

### DIFF
--- a/changelog.d/20260731_230000_russoul_macos_nix_develop_devshell.md
+++ b/changelog.d/20260731_230000_russoul_macos_nix_develop_devshell.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+
+### Patch
+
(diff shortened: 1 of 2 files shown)


Russoul:
Closing — this is already fixed on `main` by 6de6c8a62 ("Integration for Node release 11.1"), which rewrote `flake.nix` to construct dev shells directly via `nix/shell.nix` instead of routing through `hydraJobs`. Verified `nix develop`/`devShells.default` works cleanly on current `main` with no patch needed. My branch here predated that fix (based on a stale point in history) — sorry for the noise.

PR #2179, opened 2026-08-04
LeiosNotify: remove stale-OCIN timing attack vector

### LINKED ISSUE #1029: Fixup a loose end of the prototype's EB announcement diffusion
Issue https://github.com/input-output-hk/ouroboros-leios/issues/772 didn't resolve the following (from PR https://github.com/input-output-hk/ouroboros-leios/pull/1005/)

>     - (TODO hmm.. what if the attacker sends a _fresh_ announcement that uses the revoked OCIN _just before_ it becomes immutable?
>       That might cause some honest nodes to disconnect from other honest nodes.
>       An attacker pool could arrange for this opportunity once per header they get onto the chain, since they could increment their OCIN with every header.
>       Perhaps the honest node accepts the revoked OCIN for, say, 1 extra minute, permitting its upstream peers' imm tips to catch up?)

That needs to be fixed before this logic could be `mainnet` worthy.

nfrisby:
Fixes Issue https://github.com/input-output-hk/ouroboros-leios/issues/1029

This PR's commit eliminates that attack vector.

Key change: we don't disconnect when the announcement has an OCIN that our imm tip has revoked; we now accept that message, since it's possible the honest peer's imm tip is a bit behind ours and so has _not_ revoked that OCIN. However, despite allowing the peer to send that message to us, we otherwise ignore it, since _we do_ know its OCIN was revoked.

The main idea: OCINs are irrelevant to which elections the adversary could send us _some_ message for. They're only relevant to what the contents of that message could be. It's the election itself that must be valid in order for our resource usage to remain bounded, and that is unrelated to the OCIN.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -53,7 +53,16 @@ import LeiosDemoDb
   , leiosDbInsertTxs
   , leiosDbLookupEbBody
   )
-import qualified LeiosDemoLogic.Announcements as Announcements
+import LeiosDemoLogic.Announcements
+  ( AnnouncementVerdict (..)
+  , ElState (..)
+  , ErrAnnouncement
+  , PeerState
+  , ShouldRelay (..)
+  , TraceLeiosNotifyEvent (..)
+  , TraceLeiosNotifyPeerEvent (..)
+  , prunePeerState
+  )
 import LeiosDemoLogic.Announcements.ElBimap (ElId)
 import LeiosDemoLogic.Announcements.Validate
   ( AnnouncementInvalidity
@@ -990,14 +999,14 @@ leiosCertRbCallback kernelVars peerVars hdr cds =
 -----
 
 -- The pure logic for handling an inbound 'MsgLeiosBlockAnnouncement'. The
--- effectful glue (reading the immutable tip, the 'Announcements.PeerState' ref,
+-- effectful glue (reading the immutable tip, the 'PeerState' ref,
 -- 'MVar' updates, tracing, and 'throwIO') lives in the NodeToNode client, which
--- invokes 'Announcements.onAnnouncement' with these pieces.
+-- invokes 'onAnnouncement' with these pieces.
 
 -- | 'Header blk' as a relayed LeiosNotify announcement, paired with the
 -- announcement data parsed from it (see 'mkAnnouncingHeader'). The 'Eq' instance
 -- compares by header hash: that is the one identity used for announcement dedup
--- and equivocation counting (see 'Announcements.onAnnouncement').
+-- and equivocation counting (see 'onAnnouncement').
 data AnnouncingHeader blk
   = -- | INVARIANT: 'ancHeader' includes an announcement whose fields are
     -- 'ancAnnouncementFields'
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic/Announcements/Validate.hs
```diff
@@ -113,38 +114,82 @@ instance Show (AnnouncementInvalidity blk) where
 -- tip's ledger state (forecast to the header's slot). Envelope check skipped;
 -- see the module header.
 --
--- The operational certficiate (opcert) issue number (OCIN) is checked only as a
--- lower bound: any number at least the immutable tip's counter is accepted (the
--- over-increment upper bound, which the strict protocol check would enforce, is
--- skipped — see 'validateAnnouncementChainDepState' and
--- 'WhetherToUpperBoundOCERT'), and a lower one is rejected as a revoked key.
+-- The operational certificate (opcert) issue number (OCIN) governs only relay,
```
nfrisby, line 117: Summary of changes to this big comment: mostly the same content as before, except now we don't even disconnect on revoked OCINs, so we have to explain what we do instead and why that's OK

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1064,7 +1070,7 @@ announcementValidity ::
   m
     ( Either
         (Maybe (AnnouncementInvalidity blk))
-        (Announcements.ShouldRelay, NominalDiffTime, (LeiosPoint, BytesSize))
+        (Maybe (Announcements.ShouldRelay, NominalDiffTime, (LeiosPoint, BytesSize)))
```
geo2a: The return type of `announcementValidity` is very convoluted, encoding various cases via a `Either Maybe`. I do recognise that it is not immediately clear how to simplify it though. `announcementValidity` is passed as an argument to the `onAnnouncement` function. Is it necessary for `onAnnouncement` to be parameterised by this function? Perhaps if we inline `announcementValidity` into `onAnnouncement` we could avoid the nested eithers and maybes?

  It is unclear to me if that would actually help much, so I'm not insisting on attempting this change.

geo2a: Alright, I see that `onAnnouncement` is given a simpler implementation in the test code.

geo2a (approved):

nfrisby:
I need to add comments that emphasize the one wart here, and the two ways we could eliminate it (neither is obviously worthwhile):

- Once a KES key is leaked, as of this PR, there's something the adversary can do with it _forever_ regardless of whether it was revoked. That key allows them to send 1 or 2 extra bogus EB announcements to each of their immediate downstream peers.
    - As of this PR, the victims don't disconnect. But they also don't relay the message.
    - If the attacker got the (hot) VRF key too, then they could send these announcements even if the pool didn't; that's how they could send 2 extra announcements instead of just 1 extra.
    - The only way to stop this would be for the pool whose key was leaked to be entirely deregistered.
    - The extra work here is minimal, so this is undesirable but not a showstopper.
- How would we eliminate this? Two ways.
    - (Big change, but see the comments of Issue https://github.com/IntersectMBO/ouroboros-leios/issue772) Change EB announcements to be signed by a key whose validity/revocation is determined by slot. Assuming the change only happens when it's _stable_, then there's no risk of honest nodes disagreeing about which key is revoked or not.
    - (Small change, but not motivated by _anything_ else) Or, we could enrich the OCIN tracking map to also record alongside the pool's latest OCIN some indicator of when that OCIN became its latest; eg a BlockNo. Then the OCIN check could be ternary: Fresh, Stale, VeryStale; an EB announcement signed by a VeryStale OCIN would justify a disconnect.
        - Merely observing that this change is more similar to the "Perhaps the honest node accepts the revoked OCIN for, say, 1 extra minute, permitting its upstream peers' imm tips to catch up?)" from the linked `ouroboros-leios` Issue 1029 than the current PR is.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
nfrisby: Yeah, I got sloppy with the `Either (Maybe ...) (Maybe ...)`---I should write a bespoke 4 way sum type

ch1bo: Maybe keep a TODO as a quick hint that this could be improved?

ch1bo (approved):

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
ch1bo: Did that refactor now

ch1bo:
Merged despite errors in CI because the `leios-prototype` is broken already. Will fix on that branch.

PR #2180, opened 2026-08-04
[Peras 48] Do not hardcode PerasEnabled/NoPerasEnabled in era parameters

agustinmista:
This PR removes hardcoded Peras configuration from the Shelley and Hard Fork Combinator era parameters, replacing it with an era-dependent approach.

For this, we:
  * Added a `getShelleyEraPerasRoundLength` method to the `ShelleyBasedEra` class, which defaults to `NoPerasEnabled` and is overridden only for `DijkstraEra`. This method is then used by `shelleyEraParams` to resolve the Peras round length at construction time.
  * Added a `PerasEnabled ()` input to `defaultEraParams` (used in tests), and derived a sensible default from the given `k`.

Closes https://github.com/tweag/cardano-peras/issues/253 and https://github.com/tweag/cardano-peras/issues/112


jasagredo (approved):
Looks good. Would be interesting to see what changed in the LedgerConfig

PR #2181, opened 2026-08-04
db-analyser: support Leios chains

dnadales:
This PR adapts db-analyser to Leios. Part of https://github.com/input-output-hk/ouroboros-leios/issues/1023

## What changes

- `--leios-db PATH` opens the node's `leios.db`. Omit it for a pre-Leios chain.
- The ledger-applying passes fold the certified EB onto the parent state before they apply the block, as the node does.
- `--count-tx-outputs` and `--show-block-txs-size` report the certified EB in new columns.
  The block columns stay 0, so the output still says what the block itself carries.
- `--benchmark-ledger-ops` applies a certifying block, and reports the EB in five new columns.
  `blockBytes` and the era-specific columns still describe the ranking block alone.
  - `ebBytes` is the size of the EB body, which holds one tx hash and one tx size per tx.
  - `ebTxsBytes` and `ebNumTxs` are the size and the count of the EB's txs.
  - `ebReadTime` and `mut_ebRead` are the cost of the read of those txs from the LeiosDb.
  - `mut_blockApply` covers the EB txs and the certifying block's own body, because the two are one logical block.
- A missing EB stops the run, with a message that names `--leios-db`.
  A run that continued would build a wrong ledger state.
- `ouroboros-consensus-cardano/scripts/check-leios-analysis` runs both reporting passes against a real Leios chain and checks their output.
  It needs such a chain, so it does not run on CI.
  The README says how to run it.

The first two commits are cherry-picks of #2186, which is merged into `main`.
They drop out when this branch's base next merges `main`.

## Checks on a Leios devnet

A full-chain `--benchmark-ledger-ops` run covers 101845 blocks up to slot 2897033, and 3441 of them certify an EB.
It reports 3799149 EB txs and 760649136 EB tx bytes.
Both totals stay under the rows in the `ebTxs` table, as they must, because one tx is in several EBs and an announced EB is not always certified.
No block that certifies no EB reports a non-zero EB column.

`check-leios-analysis` passes every check over 29500 blocks.
It counts 49 certifying blocks, and `--benchmark-ledger-ops` reports 49 rows with an EB over the same range.
The two passes reach the EB by different reads of the LeiosDb.

## Still to do

- [ ] `--repro-mempool-and-forge` reports 0 transactions for a certifying block.
  The transactions it feeds to the mempool must come from the EB.
- [ ] `check-leios-analysis` does not cover `--benchmark-ledger-ops`.
  Every check above ran by hand.


ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -59,6 +59,21 @@ parseDBAnalyserConfig =
             , help "use v2 LSM backend"
             ]
       ]
+    <*> parseLeiosDbPath
```
nfrisby: I _think_ the relative path from the `leios.db` file to the existing `--db` "ath to the Chain DB" should always be fixed. So this command-line option seems redundant.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -32,22 +45,39 @@ data SlotDataPoint
   , slotGap :: !Word64
   -- ^ Gap to the previous slot.
   , totalTime :: !Int64
-  -- ^ Total time spent in the 5 ledger operations at 'slot'.
+  -- ^ Elapsed time spent on the tables read and the 5 ledger
```
nfrisby: comment should say "tables read and EB read", yeah?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -32,22 +45,39 @@ data SlotDataPoint
...
+  -- operations at 'slot'. Taken from GC.elapsed_ns.
   , mut :: !Int64
-  -- ^ Time spent by the mutator while performing the 5 ledger operations
-  -- at 'slot'.
+  -- ^ Time the mutator ran during the tables read and the 5 ledger
```
nfrisby: comment should say "tables read and EB read", yeah?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -32,22 +45,39 @@ data SlotDataPoint
...
+  -- operations at 'slot'. Taken from GC.mutator_elapsed_ns.
   , gc :: !Int64
-  -- ^ Time spent in garbage collection while performing the 5 ledger
-  -- operations at 'slot'.
+  -- ^ Time spent in garbage collection during the tables read and the
```
nfrisby: comment should say "tables read and EB read", yeah?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -32,22 +45,39 @@ data SlotDataPoint
...
+  -- ^ Elapsed time spent on the EB read. 'totalTime' already counts this time. A
+  -- block that certifies no EB skips the read, so both this figure and
+  -- 'mut_ebRead' are 0.
+  , mut_ebRead :: !Int64
+  -- ^ Time the mutator ran during the EB read. Taken from
+  -- GC.mutator_elapsed_ns, as 'mut_tableRead' is.
   , majGcCount :: !Word32
-  -- ^ Total number of __major__ garbage collections that took place while
-  -- performing the 5 ledger operations at 'slot'.
+  -- ^ Total number of __major__ garbage collections that took place
```
nfrisby, line 74: comment should say "tables read and EB read", yeah?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -32,22 +45,39 @@ data SlotDataPoint
...
+  -- during the tables read and the 5 ledger operations at 'slot'.
   , minGcCount :: !Word32
```
nfrisby, line 77: comment should say "tables read and EB read", yeah?

ouroboros-consensus-cardano/README.md
```diff
@@ -78,6 +78,20 @@ The user can use snapshots created by the node or they can create their own snap
...
 
+#### --leios-db
+
+```
+[--leios-db PATH]
+```
+
+Path to the node's SQLite Leios database, usually `DB_PATH/leios.db`.
+
+A Leios ranking block that carries a certificate has an empty body on the wire.
```
nfrisby: Might be better to say "Praos block" instead of "Leios ranking block" (or say both and put one in parens)

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
nfrisby: Ah, I see "Without it, an analysis stops at the first certifying block." below. So my suggestion would instead have to be to change this to boolean flag.

ouroboros-consensus-cardano/README.md
```diff
@@ -78,6 +78,20 @@ The user can use snapshots created by the node or they can create their own snap
...
+Its transactions are in the endorser block (EB) that it certifies, and those live in the Leios database, not in the ImmutableDB.
+Pass this flag to analyse such a chain.
+Without it, an analysis stops at the first certifying block.
```
nfrisby: I worry that stopping at the first CertRB if the command-line didn't enable might be a relatively silent and confusing default behavior :/

ouroboros-consensus-cardano/README.md
```diff
@@ -160,12 +176,16 @@ Lastly the user can provide the analysis that should be run on the chain:
...
   from NUM blocks every time and then runs the forging loop. Useful to inspect
   regressions in the forging loop or in the mempool adding/snapshotting logic.
-  The output shows the time spent on ticking and snapshotting the mempool broken
+  The output shows the time spent on each step of the forging loop broken
   down into:
   - real monotonic measured time
   - time spent in the mutator in microseconds
   - time spent in GC in microseconds
 
+  The steps are: the read of the certified EB closure, the application of that
```
nfrisby: Does "the application of that closure" include the necessary "table reads"?

ouroboros-consensus-cardano/README.md
```diff
@@ -176,6 +196,17 @@ Lastly the user can provide the analysis that should be run on the chain:
   Therefore, it is recommended to start with NUM=1, and only use NUM=2 when you
   want to test the performance impact of a more filled mempool.
 
+  On a Leios chain, pass `--leios-db`.
+  A block that certifies an EB has an empty body, so its transactions are in the Leios database.
```
nfrisby: I wouldn't say "so _its_ transactions are in the Leios database". I'd say something more like "so the txs that block causes to be applied to the ledger are in..."

ouroboros-consensus-cardano/README.md
```diff
@@ -196,6 +227,41 @@ save the location to the working directory from which it was run. Eg:
 export NODE_HOME=/path/to/local/copy/of/cardano-node/working/dir/
 ```
 
+#### Checking the Leios analyses
+
+`scripts/check-leios-analysis` runs `--count-tx-outputs` and `--show-block-txs-size` against a real Leios chain and checks their output.
+It needs a node data directory that holds `config.json` and `db/`, and it derives the rest from that.
+
+```sh
+check-leios-analysis $NODE_HOME
```
nfrisby: What is `$NODE_HOME`?

ouroboros-consensus-cardano/README.md
```diff
@@ -196,6 +227,41 @@ save the location to the working directory from which it was run. Eg:
...
+```sh
+nix run .#check-leios-analysis -- $NODE_HOME
+```
+
+The two analyses read the Leios database by different paths, so their agreement on which blocks certify an EB is the main check.
+The script also fails if the chain has no certifying block, because then it proves nothing.
+
+A full chain takes a few minutes. `--num-blocks-to-process` shortens it, at the risk of stopping before the first certifying block.
+
+`--benchmark` adds `--benchmark-ledger-ops` to the run, and checks the EB columns of its output against `--show-block-txs-size`.
```
nfrisby: > and checks the EB columns of its output against `--show-block-txs-size`

It's not clear to me what that check looks like/would mean

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -324,24 +416,33 @@ showSlotBlockNo AnalysisEnv{db, registry, startFrom, limit, tracer} =
   Analysis: show total number of tx outputs per block
 -------------------------------------------------------------------------------}
 
-countTxOutputs :: forall blk. HasAnalysis blk => Analysis blk StartFromPoint
-countTxOutputs AnalysisEnv{db, registry, startFrom, limit, tracer} = do
-  void $ processAll db registry GetBlock startFrom limit 0 process
+-- | A cert-RB has an empty wire body, so its own count is 0. Its txs are in the
```
nfrisby: "Its txs" again is kind of misleading

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -376,24 +477,38 @@ showHeaderSize AnalysisEnv{db, registry, startFrom, limit, tracer} = do
   Analysis: show the total transaction sizes in bytes per block
 -------------------------------------------------------------------------------}
 
-showBlockTxsSize :: forall blk. HasAnalysis blk => Analysis blk StartFromPoint
-showBlockTxsSize AnalysisEnv{db, registry, startFrom, limit, tracer} = do
-  processAll_ db registry GetBlock startFrom limit process
+-- | A cert-RB has an empty wire body, so its own tx columns are 0. Its txs are
```
nfrisby: "Its txs" again

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -969,29 +1217,305 @@ reproMempoolForge numBlks env = do
...
           -- doing this sharing optimization?
-          LedgerDB.reapplyThenPushNOW intLedgerDB blk
+          (_, nextLedgerSt) <- applyBlockAtTip leiosDb LedgerReapply cfg ledgerDB blk
+          LedgerDB.push intLedgerDB nextLedgerSt
           LedgerDB.tryFlush ledgerDB
 
           -- this flushes blk from the mempool, since every tx in it is now on the chain
           void $ Mempool.testSyncWithLedger mempool
 
+{-------------------------------------------------------------------------------
```
nfrisby: This section is a lot of new logic. I doubt it belongs in `db-analyser`; shouldn't it be provided by the `ouroboros-conesnsus` library instead?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -46,9 +46,11 @@ module Ouroboros.Consensus.Storage.LedgerDB.Forker
     -- * Validation
   , AnnLedgerError (..)
   , AnnLedgerError'
+  , Ap (..)
```
nfrisby: I doubt we want to be exporting this `Ap`-based interface from `ouroboros-consensus`

ouroboros-consensus-cardano/scripts/check-leios-analysis
```diff
@@ -0,0 +1,718 @@
+#!/usr/bin/env python3
```
nfrisby: I didn't review this file at all.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
nfrisby: At the very least, we should add it to a separate module of this tool, I think.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -40,13 +40,26 @@ import Control.Monad (join, unless, void, when)
 import Control.Monad.Except (runExcept)
 import Control.ResourceRegistry
 import Control.Tracer (Tracer (..), nullTracer, traceWith)
+import Data.Bifunctor (bimap)
```
nfrisby, line 57: I barely reviewed this module at all; just a skim and some high-level reactions

nfrisby (commented):
This is not a complete pass; I mostly skipped the big files.

But I review the UX, the docs, and some higher-level shapes of the code

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
bladyjoker: Do we need that flag at all? DbAnalyzer "knows" if it's in era that needs LeiosDB right? And knows where to find it?

ouroboros-consensus-cardano/scripts/check-leios-analysis
bladyjoker, line 1: I didn't look deeply into this, but I don't feel like maintaining a giant python script in this repo. This needs proper packaging, formatting, linting, not to mention common understanding of it.

From what I can tell this is about testing db-analyzer, even thought I support CLI based testing, this opens a can of worms I'd rather not open if we can perhaps due library based testing?

ouroboros-consensus-cardano/scripts/check-leios-analysis
```diff
@@ -0,0 +1,718 @@
...
+COUNT_RE = re.compile(
+    r"^\[[\d.]+s\] BlockNo (\d+)\tSlotNo (\d+)\t"
+    r"cumulative: (\d+)\tcount: (\d+)\tEB count: (\d+)$"
+)
+
+SIZE_RE = re.compile(
+    r"^\[[\d.]+s\] SlotNo (\d+)\t"
+    r"Num txs in block = (\d+)\tTotal size of txs in block = SizeInBytes (\d+)\t"
+    r"Num txs in EB = (\d+)\tTotal size of txs in EB = SizeInBytes (\d+)$"
+)
```
bladyjoker: Hmm is there no structured output available?

ouroboros-consensus-cardano/scripts/check-leios-analysis
bladyjoker: I would also put this file under `test/db-analyzer.py`, that being said, shouldn't there be a suffix here?

ouroboros-consensus-cardano/scripts/check-leios-analysis
```diff
@@ -0,0 +1,718 @@
...
+import re
+import shutil
+import sqlite3
+import subprocess
+import sys
+import tempfile
+import time
+from collections import namedtuple
+
+OutputRow = namedtuple("OutputRow", "block_no slot cumulative block_outputs eb_outputs")
```
bladyjoker: *Row and *RE match in naming for all but Output and COUNT, let's align

ouroboros-consensus-cardano/scripts/check-leios-analysis
```diff
@@ -0,0 +1,718 @@
...
+        return given
+
+    if shutil.which("cabal"):
+        result = subprocess.run(
+            ["cabal", "list-bin", "db-analyser"],
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode == 0:
+            return result.stdout.strip()
```
bladyjoker: It's better not to compile from here, this tool should expect `db-analyzer` as:
1. `--db-analyzer` flag
2.  "DB_ANALYZER" env var
3. `db-analyzer` in $PATH

that's the usual way no?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/FileWriting.hs
bladyjoker, line 104: If you insist on the python test script, perhaps instead of regex extraction you can leverage `import csv`. It looks like the output is CSV formatted

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
bladyjoker, line 117: Not sure this lands in the CLI output, but if it does, you can get structured output in JSON? Perhaps a flag to write db-analyzer results in a JSON format so we can use it in testing and benchmarking?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -151,7 +151,12 @@ analyse dbaConfig args =
     lsmSalt <- fst . genWord64 <$> newStdGen
     ProtocolInfo{pInfoInitLedger = genesisLedger, pInfoConfig = cfg} <-
       mkProtocolInfo args
-    leiosDbHandle <- newLeiosDBInMemory
+    -- Open the SQLite LeiosDb when a path was given, so cert-RB
+    -- closures can be resolved during replay; otherwise keep the empty
+    -- in-memory stub (correct for pre-Leios chains, which have no cert-RBs).
+    leiosDbHandle <- case leiosDbPath of
```
bladyjoker: not sure you're aware or if it matters, https://github.com/IntersectMBO/cardano-node/blob/c5f7d9121beb67e6d03d0632ae2bd3b76259728e/cardano-node/src/Cardano/Node/Run.hs#L361, cardano-node config has a LeiosDB config field.

flake.nix
bladyjoker, line 1: Nice! Didn't notice that you packaged it! However, if you hit the `cabal build...` this will break so I suggest just letting the tool rely on the existence of db-analyzer in the $PATH, no need to support various cases.

flake.nix
```diff
@@ -97,7 +97,18 @@
         };
         inherit hydraJobs;
         legacyPackages = pkgs;
-        packages = hydraJobs.native.haskell96.exesNoAsserts;
+        packages = hydraJobs.native.haskell96.exesNoAsserts // {
+          # Runs without the development shell, so it needs a db-analyser of its
+          # own. Inside the shell the script prefers a cabal build over this one.
+          check-leios-analysis = pkgs.writeShellApplication {
+            name = "check-leios-analysis";
+            runtimeInputs = [ pkgs.python3 ];
```
bladyjoker: If you add `hydraJobs.native.haskell96.exesNoAsserts.db-analyser` in `runtimeInputs` it becomes included in PATH, and your script could just use that.

bladyjoker (commented):
I don't feel like I'm in a position to approve as I'm not familiar with db-analyzer. 

Some larger points:

1. CLI based testing - I actually loving this approach generally, and I have concerns:
   - I'm not keen on keeping around big vibe coded python scripts, I can only see increasingly more tokens being needed to maintain it
   - It's a manual tool to test a manual tool, I would be more inclined to have this if we can have hermetic (Nix, CI) CLI  golden/unit tests give us some assurances. We don't have a harness developed for this, we could, but definitely out of scope

2. `db-analyzer` is adopting the internal "abstraction" that is very arguable as "stubs" are proliferating in this view of things, but I understand we're rolling with it. 
   - What would make sense to me is to have `db-analyzer leios` and `db-analyzer praos` distinct CLI modes? Currently, our information model is pretty problematic in terms of "stubbing" everything that doesn't support a new concept (like Leios and Peras).
   - I also feel this is terribly out of scope so feel free to ignore it

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
dnadales: Good point, thanks. What about:

- The tool always looks for `DB_PATH/leios.db`. That is where the node puts it by default, and `--db` is the same root the node uses.
- If the file is there, the tool opens it. If not, the tool falls back to `newLeiosDBInMemory`, which starts empty and which nothing in the tool writes to (`completeChainDbArgs` needs a `LeiosDbHandle`).
- A pre-Dijkstra chain holds no certifying block, so nothing reads the LeiosDb, and the empty in-memory one is enough.
- A Dijkstra chain with no file causes the tool to exit, with a message that names the path we looked for. We know a Dijkstra chain needs the LeiosDb, so it is pointless to start the analysis.

If an operator sets `LeiosDbConfig.Filepath` to somewhere else, a symlink into `DB_PATH` covers it.

Does that work for you?

dnadales:
>    * I'm not keen on keeping around big vibe coded python scripts, I can only see increasingly more tokens being needed to maintain it
>    * It's a manual tool to test a manual tool, I would be more inclined to have this if we can have hermetic (Nix, CI) CLI  golden/unit tests give us some assurances. We don't have a harness developed for this, we could, but definitely out of scope

So I need something actionable here:

I think we need a way to prevent db-analyzer from bitrotting, even if we don't run this on CI. 

So the two options I see are: 

1. Remove the script.
2. Keep it.

@bladyjoker @nfrisby thoughts? 

> 2. `db-analyzer` is adopting the internal "abstraction" that is very arguable as "stubs" are proliferating in this view of things, but I understand we're rolling with it.
>    
>    * What would make sense to me is to have `db-analyzer leios` and `db-analyzer praos` distinct CLI modes? Currently, our information model is pretty problematic in terms of "stubbing" everything that doesn't support a new concept (like Leios and Peras).
>    * I also feel this is terribly out of scope so feel free to ignore it

I don't see a clear path forward for this.

ouroboros-consensus-cardano/scripts/check-leios-analysis
dnadales: > I would also put this file under test/db-analyzer.py, that being said, shouldn't there be a suffix here?

I prefer running the script without the .py suffix, but I can add it if we're gonna keep this test.

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
nfrisby: "A Dijkstra chain with no file ... pointless to start the analysis" --- we have no way to detect that the chain is a Dijkstra chain _a priori_; we can only notice that when see a Dijkstra block arise in the middle of the analysis.

Yeah, symlink is fine backstop

ouroboros-consensus-cardano/scripts/check-leios-analysis
nfrisby: I also don't think maintaining this big script is the right move.

It's presumably at least nice validation of this PR though :)

bladyjoker (approved):
Thanks Damian

ouroboros-consensus-cardano/scripts/check-leios-analysis
dnadales: Script was moved here https://github.com/input-output-hk/ouroboros-consensus-tools/pull/14

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
dnadales: Sure, but maybe it'd be better in a follow up PR. WDYT?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
dnadales: Thanks! I certainly wasn't aware. We would need to decide what to do if the config has no `LeiosDbConfig`. This could be because the node is pre-leios, or the node created it a a default location. So maybe an explicit flag is fine for now?

nfrisby (approved):
New `stubbed-leios-db` flag satisfies my concerns 👍

PR #2182, opened 2026-08-05
db-analyser: time the ledger-table fetch that precedes block processing

Russoul:
## Summary

`--benchmark-ledger-ops`'s per-block timers (`totalTime`, `mut`, and the 5
phase timers) are all diffed from an RTS-stats snapshot taken *after*
`forkerReadTables` completes. For an on-disk `LedgerDB` backend (e.g. `LSM`
with `--lsm-no-cache`), that's exactly where the backend's UTxO-table disk
reads happen -- so none of the existing figures can see that cost, no matter
which one a benchmark's conclusion is based on.

Adds `tableReadTime` (elapsed) and `mut_tableRead` (mutator-only) to
`SlotDataPoint`, timed the same way the 5 existing phases already are,
bracketing `withTipForker`/`forkerReadTables` instead of what comes after it.


changelog.d/20260803_190000_ruslan.feizerakhmanov_table_read_timing.md
```diff
@@ -0,0 +1,24 @@
...
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- `--benchmark-ledger-ops` now reports two additional `SlotDataPoint` fields, `tableReadTime` and `mut_tableRead`, timing the ledger-table fetch (`forkerReadTables`) that precedes a block's 5 timed ledger operations. Previously this fetch -- where an on-disk `LedgerDB` backend performs its UTxO-table reads -- fell entirely outside the window used to compute every other reported figure, including `totalTime`.
```
jasagredo: No need to add a changelog. We don't track changes to `unstable-*` libs or exes

jasagredo (approved):
Approved, but please remove the changelog fragment

jasagredo:
Also `fourmolu` is failing. Use `./scripts/ci/run-fourmolu.sh` with fourmolu 0.20 on the path

PR #2183, opened 2026-08-05
Forge and ChainSel CallTrace instrumentation

bladyjoker:
Part of https://github.com/input-output-hk/ouroboros-leios/issues/911

DONE
- [x] More instrumentation and profiling
  - Additional CallTrace instrumentation to the Forge thread
  - ChainSel thread CallTrace instrumentation

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
@@ -761,8 +761,9 @@ partitionMempool leiosConn leiosVoteState leiosTracer pmCtrace pmCallCtx cfg mem
             -- parent) with the tick diff (relative to the
             -- state with the EB closure applied), so the mempool snapshot is revalidated
             -- against parent + closure + tick.
-            -- TODO(bladyjoker): Force this to evaluate and instrument as "tick-ledger-state-after-eb"
-            let tickedLsAfterEB =
+            tickedLsAfterEB <-
+              pmTrace'Via (const ()) "tick-ledger-state-after-eb" currentSlot $
+                evaluate $
                   lcaClosureDiff
                     `prependDiffs` applyChainTick
                       OmitLedgerEvents
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
@@ -763,8 +763,9 @@ partitionMempool leiosConn leiosVoteState leiosTracer pmCtrace pmCallCtx cfg mem
             -- parent) with the tick diff (relative to the
             -- state with the EB closure applied), so the mempool snapshot is revalidated
             -- against parent + closure + tick.
-            -- TODO(bladyjoker): Force this to evaluate and instrument as "tick-ledger-state-after-eb"
-            let tickedLsAfterEB =
+            tickedLsAfterEB <-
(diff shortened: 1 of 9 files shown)


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
```diff
@@ -761,22 +765,24 @@ partitionMempool leiosConn leiosVoteState leiosTracer pmCtrace pmCallCtx cfg mem
...
+                evaluate $ lcaClosureDiff `prependDiffs` lcaStateAfterEBTicked
+
             snap <-
               pmTrace'Via (const ()) "mempool-get-snapshot-for-no-cache" currentSlot $
                 getSnapshotForNoCache mempool currentSlot tickedLsAfterEB readTables
-            let ebTxs' = fst (snapshotTake snap ebCap)
-            pure ([], ebTxs', Data.Measure.zero, snap)
 
-  -- force the mempool's computation before the tracer event
-  _ <- evaluate (length rbTxs)
```
ch1bo, line 779: Where did this go?

ch1bo: I see two evaluates further up, but why is the `_ <- evaluate (length ebTxs')` below then still here?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -349,111 +349,63 @@ chainSelSync ::
...
+    (show $ blockHash $ blockToAdd bta)
+    (\childCCtx -> chainSelAddBlock cdb childCCtx bta)
+chainSelSync cdb@CDB{cdbTracer} cctx (ChainSelAddPerasCert cert varProcessed) =
+  callTraceSameThreadVia
+    id
+    (traceWith cdbTracer . TraceAddBlockEvent . TraceAddBlockCall . SomeJsonCallTrace)
+    cctx
+    "chain-sel-add-peras-cert"
+    (show $ getPerasCertRound cert)
+    (\_ -> chainSelAddPerasCert cdb cert varProcessed)
```
ch1bo, line 386: Should: not make code unreadable

The chain selection is one of the most complicated pieces of logic in this whole codebase. These changes here obfuscate it further.

ch1bo (commented):
I was gonna approve because it should only be tracing changes. But, boy.. this is making code unreadable!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
bladyjoker: I agree, that's why I refactored the whole thing such that we can have this wiring done mechanically in one place. That being said, there's an obvious redundancy here that I didn't clean up.

bladyjoker: Imo by splitting the core logic in separate function and having one dispatcher, this is imo easier to grok. Deep instrumentation necessarily makes this more noisy, but that also made me refactor this and the Forge in a way that makes its logic more obvious. 

Now one can see clearly this module is about processing 4 messages.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
bladyjoker: You're right, this should all go away. We're doing wasteful passes here.

PR #2184, opened 2026-08-05
[Peras 49] Use injective type families for associated BlockSupportsPeras types

agustinmista:
This commit tweaks the `BlockSupportsPeras` class to use injective type families for `PerasVote`, `PerasCert`, and `PerasError` (to soon replace `PerasForgeErr` and `PerasValidationErr`).

The main idea is to allow us to avoid some of the necessary boilerplate we would require if used data families by providing a default (void) implementation for all these types, while still being able to take advantage of blk-based injectivity.

NOTE: for now, we turned associated `PerasCert` and `PerasVote` data instances into standalone types, but these will be ultimately removed after we get rid of the degenerate `BlockSupportsPeras` instance.

### DIFF
--- a/changelog.d/20260806_121453_agustin.mista_use_injective_type_families.md
+++ b/changelog.d/20260806_121453_agustin.mista_use_injective_type_families.md
@@ -1,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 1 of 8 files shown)


jasagredo (approved):
Nice!

PR #2185, opened 2026-08-05
[Peras 50] Use SeatIndex-based PerasVoteId

agustinmista:
This commit replaces the old `PerasVoteId` type from `Ouroboros.Consensus.Block.SupportsPeras` with the (monomorphic) one defined in `Ouroboros.Consensus.Peras.Types`.

Notably, this new type uses the seat index in the voting committee to identify voters, as opposed to the previous `KeyHash StakePool`.


jasagredo (approved):
Very nice focused change :+1:

PR #2186, opened 2026-08-05
db-analyser: correct the table-read timing docs and write its CSV columns

dnadales:
Follow-up to #2182.

The CSV output gains two columns after `gc`.
A script that reads the CSV by column position needs an update.


jasagredo (approved):
Looks good

PR #2187, opened 2026-08-06
Do not generate Peras certificates if they would be larger than maximum block size

nbacquey:
If we generated a certificate that is too large to fit in a block body, we would have an error in the future if we choose to include it in a block.
Instead, we now consider the certificate forging to have failed.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -69,6 +69,7 @@ import qualified Cardano.Crypto.Hash as Hash
 import Cardano.Ledger.Hashes (KeyHash (..))
 import Control.Exception (assert)
 import Control.Exception.Base (Exception)
+import Control.Monad (when)
 import Data.Bifunctor (bimap)
 import Data.Containers.NonEmpty (NE)
 import Data.Kind (Type)
@@ -366,6 +367,10 @@ class
     concreteCert <-
       bimap injectConversionError id $
         toPerasCert abstractCert
+    let maxSize = unPerasMaxBlockBodySize $ perasMaxBlockBodySize params
+        actualSize = getPerasCertSize concreteCert
+    when (actualSize > maxSize) $
+      Left $ injectCertificateTooLargeError actualSize maxSize
     pure $
       ValidatedPerasCert
         { vpcCert = concreteCert
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -512,6 +512,9 @@ instance
   where
   getPerasCertRound = getPerasCertRound . vpcCert
   getPerasCertBlock = getPerasCertBlock . vpcCert
+  -- We don't account for the size of the boost, because it is
+  -- never serialized.
+  getPerasCertSize = getPerasCertSize . vpcCert
 
 -- * Peras error types
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -77,6 +77,7 @@ import qualified Data.Map.NonEmpty as NEMap
 import Data.Map.Strict (Map)
 import Data.Traversable (for)
 import Data.Typeable (Typeable)
+import Data.Word (Word32)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 import Ouroboros.Consensus.Block.Abstract (Point, StandardHash)
@@ -519,6 +520,7 @@ class IsPerasError err blk | err -> blk where
   injectVotingCommitteeError :: PerasVotingCommitteeError blk -> err
   injectConversionError :: PerasConversionError -> err
   injectQuorumNotReachedError :: VoteWeight -> err
+  injectCertificateTooLargeError :: Word32 -> Word32 -> err
 
 instance IsPerasError (VoidPerasError blk) blk where
   injectVotingCommitteeError _ =
@@ -527,6 +529,8 @@ instance IsPerasError (VoidPerasError blk) blk where
     error "injectConversionError: VoidPerasError cannot be inhabited"
   injectQuorumNotReachedError _ =
     error "injectQuorumNotReachedError: VoidPerasError cannot be inhabited"
+  injectCertificateTooLargeError _ _ =
+    error "injectCertificateTooLargeError: VoidPerasError cannot be inhabited"
 
 -- * Types and functions related to Peras vote collection and quorum checking
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
@@ -808,6 +808,9 @@ instance
   -- NOTE: in practice this is never produced at the HFC level
   injectQuorumNotReachedError _ = HardForkPerasErrorQuorumNotReachedError
 
+  -- NOTE: in practice this is never produced at the HFC level
+  injectCertificateTooLargeError _ _ = HardForkPerasErrorCertificateTooLargeError
+
 -- ** 'BlockSupportsPeras' instance for 'HardForkBlock'
 
 class
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
@@ -793,6 +793,11 @@ instance
       . hcmap proxySingle (K . upcastToHardForkPoint . getPerasCertPoint . unwrapPerasCert)
       . getOneEraPerasCert
 
+  getPerasCertSize =
+    hcollapse
+      . hcmap proxySingle (K . getPerasCertSize . unwrapPerasCert)
+      . getOneEraPerasCert
+
 instance
   ( CanHardFork xs
   , All SingleEraBlock xs
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
@@ -9,6 +9,7 @@ module Ouroboros.Consensus.Peras.Cert.Class
   ( IsPerasCert (..)
   ) where
 
+import Data.Word (Word32)
 import Ouroboros.Consensus.Block.Abstract (Point)
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types (WithArrivalTime (..))
 import Ouroboros.Consensus.Peras.Types
@@ -25,6 +26,7 @@ class
   where
   getPerasCertRound :: cert -> PerasRoundNo
   getPerasCertBlock :: cert -> BoostedBlock cert
+  getPerasCertSize :: cert -> Word32
 
   getPerasCertPoint :: cert -> Point blk
   getPerasCertPoint = boostedBlockToPoint . getPerasCertBlock
@@ -35,3 +37,4 @@ instance
   where
   getPerasCertRound = getPerasCertRound . forgetArrivalTime
   getPerasCertBlock = getPerasCertBlock . forgetArrivalTime
+  getPerasCertSize = getPerasCertSize . forgetArrivalTime
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
@@ -32,9 +32,11 @@ import Cardano.Binary
   , ToCBOR (..)
   , decodeListLenOf
   , encodeListLen
+  , serialize
   )
 import Control.Monad (when)
 import Control.Monad.Error.Class (MonadError (..))
+import qualified Data.ByteString.Lazy as BL
 import Data.Coerce (coerce)
 import Data.Containers.NonEmpty (HasNonEmpty (..))
 import qualified Data.List.NonEmpty as NonEmpty
@@ -107,11 +109,13 @@ type instance BoostedBlock (PerasCert blk) = PerasBoostedBlock
 instance
   ( ConvertRawHash blk
   , HashSize blk ~ 32
+  , Typeable blk
   ) =>
   IsPerasCert (PerasCert blk) blk
   where
   getPerasCertRound = pcRoundNo
   getPerasCertBlock = pcBoostedBlock
+  getPerasCertSize = fromIntegral . BL.length . serialize
 
 instance
   Typeable blk =>
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Error/V1.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Error/V1.hs
@@ -15,6 +15,7 @@ module Ouroboros.Consensus.Peras.Error.V1
 
 import Control.Exception (Exception)
 import Data.Typeable (Typeable)
+import Data.Word (Word32)
 import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 import Ouroboros.Consensus.Block.SupportsPeras
@@ -37,6 +38,8 @@ data PerasError blk
       VoteWeight
   | PerasTemporaryPublicKeyHackError
       String
+  | PerasCertificateTooLargeError
+      Word32 Word32
 
 deriving instance
   Show (PerasVotingCommitteeError blk) =>
@@ -59,3 +62,4 @@ instance IsPerasError (PerasError blk) blk where
   injectVotingCommitteeError = PerasVotingCommitteeError
   injectConversionError = PerasVotingConversionError
   injectQuorumNotReachedError = PerasQuorumNotReachedError
+  injectCertificateTooLargeError = PerasCertificateTooLargeError
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
@@ -22,6 +22,7 @@ module Ouroboros.Consensus.Peras.Params
   , PerasWeight (..)
   , PerasQuorumWeightThreshold (..)
   , PerasQuorumWeightThresholdSafetyMargin (..)
+  , PerasMaxBlockBodySize (..)
 
     -- * Protocol parameters bundle
   , PerasParams (..)
@@ -55,7 +56,7 @@ import Control.Monad.Trans.Class
 import Data.Coerce (coerce)
 import Data.Semigroup (Sum (..))
 import Data.Typeable (Typeable)
-import Data.Word (Word64)
+import Data.Word (Word32, Word64)
 import GHC.Generics (Generic)
 import qualified Ouroboros.Consensus.Committee.Types as Committee
 import Ouroboros.Consensus.Util.Condense (Condense (..))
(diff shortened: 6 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -431,6 +433,10 @@ defaultForgePerasCert context voteCollection = do
   concreteCert <-
     bimap injectConversionError id $
       toPerasCert abstractCert
+  let maxSize = unPerasMaxBlockSize $ perasMaxBlockSize params
```
tbagrel1: I get
```
ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs:436:17: error: [GHC-88464]
    Variable not in scope: unPerasMaxBlockSize :: a0 -> Word32
    Suggested fix:
      Perhaps use record field of PerasMaxBlockBodySize ‘unPerasMaxBlockBodySize’ (imported from Ouroboros.Consensus.Peras.Params)
    |
436 |   let maxSize = unPerasMaxBlockSize $ perasMaxBlockSize params
    |                 ^^^^^^^^^^^^^^^^^^^

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs:436:39: error: [GHC-88464]
    Variable not in scope: perasMaxBlockSize :: PerasParams blk -> a0
    Suggested fix:
      Perhaps use record field of PerasParams ‘perasMaxBlockBodySize’ (imported from Ouroboros.Consensus.Peras.Params)
    |
436 |   let maxSize = unPerasMaxBlockSize $ perasMaxBlockSize params
    |                                       ^^^^^^^^^^^^^^^^^
```
with GHC 9.14

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -431,6 +433,10 @@ defaultForgePerasCert context voteCollection = do
...
+      actualSize = getPerasCertSize concreteCert
+  when (actualSize > maxSize) $
+    Left $ injectCertificateTooLargeError actualSize maxSize
```
tbagrel1: Would it make sense to consider this as a `ConversionError`? So we don't need to introduce a new error variant + injection function ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
```diff
@@ -25,6 +26,7 @@ class
   where
   getPerasCertRound :: cert -> PerasRoundNo
   getPerasCertBlock :: cert -> BoostedBlock cert
+  getPerasCertSize :: cert -> Word32
```
tbagrel1: I'm not sure I would add this as a class method. We use `getPerasCertSize` only on a bare `PerasCert` (never on a validated or `WithArrivalTime`-wrapped one), and it's mostly an implementation detail.

tbagrel1: The documentation of this method/function should tell which class (for instance, `Serialise`) is used to compute the size. Theoretically we could have `Serialise` and `SerialiseNodeToNode` returning different sized objects (we have a comment saying both impleme should coincide, but it's a fairly weak invariant 😅)
Also are we sure that it's the `Serialise` class which is used at the time of putting a cert inside a block body?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
```diff
@@ -233,6 +247,13 @@ defaultPerasParams =
         PerasQuorumWeightThresholdSafetyMargin (2 / 100)
     , perasTargetCommitteeSize =
         Committee.TargetCommitteeSize 800
+    , -- Mainnet value as of 2026-08-06
+      -- In the future, this should *not* be retrieved directly from
+      -- the ledger state, but rather from `TxLimits blk`, because it
+      -- has to account for various overheads, such as the block body
+      -- CBOR encoding, the HFC overhead, etc.
+      perasMaxBlockBodySize =
+        PerasMaxBlockBodySize 90112
```
tbagrel1: This seems a bit flaky, especially since we are now targeting Mainnet. At least we need a TODO + tracking issue if we want to ship the code as-is.

tbagrel1 (commented):
Left a couple of comments.

I'm a bit worried about the breadth of changes for something that doesn't impact how certs are used/their fields. As suggested in one of my review comments, I would prefer not making `getPerasCertSize` a method of the `IsPerasCert` class since we only use it in one fairly localized place on a bare `PerasCert blk` (not an abstract `cert`)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
nbacquey: Agreed for the TODO. We *need* to rewrite that part of the code as we'll get the Peras params from the ledger, instead of this hardcoded value anyway. That's why I'm not too bothered by shipping it as it is

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
nbacquey: Had to push without trying to compile, because 2 weeks without updating cabal makes GHC grumpy

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
nbacquey: Created https://github.com/tweag/cardano-peras/issues/280 to track this

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
nbacquey: I'm not sure what a good solution to this would look like. Computing manually the size in the default implementation for `forgePerasCert` would mean adding a `ToCBOR (PerasCert blk)` constraint, which would not be met natively by several mock implementations. Adding a `certSize :: PerasCert blk -> Int` to `SupportsPeras` would introduce a similar amount of boilerplate. What would your suggestion be?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
nbacquey: That's not too big an approximation :+1: 
Done

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
tbagrel1: I agree, these are valid concerns

To be fair I'm not so sure about the approach at all ("If we generated a certificate that is too large to fit in a block  body, we would have an error in the future if we choose to include it in a block. Instead, we now consider the certificate forging to have failed."). 

It seems to me that that `forgePerasCert` shouldn't morally have to care about the size of the serialized representation of a cert. Consequently, all designs that allow this are gonna feel weird.

But I wasn't there when it was decided to go down this route, so I won't argue more against it. Feel free to ignore my comment.

agustinmista: I haven't had time to look at the code in this PR, but I would like to comment on:

> It seems to me that that forgePerasCert shouldn't morally have to care about the size of the serialized representation of a cert. Consequently, all designs that allow this are gonna feel weird.

100% agree. I would much rather avoid that if possible.

However, I think there's an argument to leverage here: the worst-case scenario size-wise happens if a node observes a quorum being reached after having received only non-persistent votes (those are the ones contributing to the increased cert size). 

From there, we could derive a conservative bound for the `TargetCommitteeSize` protocol parameter such that the split of persistent vs. non-persistent voters in the (mainnet committee) must be so that _every_ quorum reached by that committee leads to a certificate that fits within a block body.

Alternatively, we could also extend this notion to allow for a statistically low chance of failure (triggering a cooldown if everyone observes the same quorum-reaching votes): we could bound `TargetCommitteeSize` so that it allows certain worst-case certificates to happen 1:1_000_000 times (or similar). In such a case, I would make this part of the certificate inclusion logic and simply discard the cert somewhere along the pipeline.

nbacquey: As discussed off-chain, we decided to include a hard limit on Peras certificates, but not necessarily to tie it to the maximum block body size. The new state of the PR reflects that choice

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -565,6 +565,11 @@ instance
       . hcmap proxySingle (K . injectHardForkPoint . getPerasCertPoint . unwrapPerasCert)
       . getOneEraPerasCert
 
+  getPerasCertSize =
```
tbagrel1: Are we sure about that? From my understanding, the HFC wrapper needs to be serialized (so we know which variant to deserialize into) so it would increase the size of the produced certificate

tbagrel1 (approved):
My initial concerns are still present, but I won't push back more. I think it would be helpful though in a code comment or in this PR description to justify why we think it's OK/better to make cert forging fail while the size of the cert is only a concern for the very rare case where we would need to include the cert in a block body

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -107,11 +110,13 @@ type instance BoostedBlock (PerasCert blk) = PerasBoostedBlock
 instance
   ( ConvertRawHash blk
   , HashSize blk ~ 32
+  , Typeable blk
   ) =>
   IsPerasCert (PerasCert blk) blk
   where
   getPerasCertRound = pcRoundNo
   getPerasCertBlock = pcBoostedBlock
+  getPerasCertSize = PerasCertSize . fromIntegral . BL.length . serialize
```
agustinmista: Given that this instance is tightly coupled to the V1 implementation, couldn't we derive a closed formula for it? 

AFAIU, we don't need a precise size, but more of an upper limit, and serializing the certificate just to look at its size would unnecessarily slow things down.

I imagine we could do something like:

```haskell
getPerasCertSize = c1 + c2*len(persistent_voters) + c3*len(non_persistent_votes)
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
```diff
@@ -25,6 +26,7 @@ class
   where
   getPerasCertRound :: cert -> PerasRoundNo
   getPerasCertBlock :: cert -> BoostedBlock cert
+  getPerasCertSize :: cert -> PerasCertSize
```
agustinmista: I don't think we need this functionality for arbitrary certs, as the size limit is an internal detail of the Dijkstra/V1 `forgePerasCert` implementation. 

Have you considered keeping everything within the Dijkstra instance (where `PerasCert blk`) is already inferred to be a `V1.PerasCert` and use the actual size check directly?

(we would still need a new error type for V1.PerasError though)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
```diff
@@ -201,3 +202,12 @@ instance ToCBOR PerasVoteId where
       encodeListLen 2
         <> toCBOR pviRoundNo
         <> toCBOR pviSeatIndex
+
+-- | Size of a Peras certificates in bytes.
+--
+-- We use this to ensure that the maximum block size is not exceeded
+-- when including a Peras certificate in a block.
+newtype PerasCertSize
+  = PerasCertSize {unPerasCertSize :: Word32}
```
agustinmista, line 211: Why not just `Word64`? I don't think this would ever be sent over the wire, and it would probably still take 8 bytes in memory due to alignment.

agustinmista (changes requested):
Good job! I think we need to reconsider two design choices, though.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
nbacquey: The maximum block body size is defined in `cardano-ledger` as a `Word32`. In the future, those values will have to be compared. I'd like to avoid extra conversion when that happens.

agustinmista: Sorry, I meant why not using Word64 instead of Word32 _in the newtype_.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/Class.hs
nbacquey: I somehow missed the Dijkstra-specific code. Updated that in the last version. It looks really better that way.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Types.hs
nbacquey: I think we cross-posted, I understood what you meant a bit too late :D
See the actual answer above

tbagrel1:
I agree it looks much better now, and solve a couple of my initial concerns.

FTR, at the point where we include a cert inside a block body, if we ever need to compute the exact size of the cert, **we mustn't forget the size of the HFC wrapper**. That being said, `Serialise` instance for a `PerasCert (HardForkBlock xs)` should already take care of serialising the NS wrapper, so doing `length <$> serialise hfcCert` should give the right value :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
nbacquey: I did that in the latest version, but left values of `c1`, `c2`, and `c3` `undefined`. Do you have insight on what those values should be?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
nbacquey: This computation is outdated. In the latest version I use an upper bound, with constants yet to be discussed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
agustinmista: It seems that the encoding is a bit more complex than I anticipated. I computed a bunch of combinations of persistent/non-persistent voters and asked a clanker to derive a function to describe the size:

```haskell
perasCertSize :: Int -> Int -> Int
perasCertSize persistentVoters nonPersistentVoters
  | persistentVoters < 0 || nonPersistentVoters < 0 =
      error "voter counts must be non-negative"
  | totalVoters == 0 =
      error "a Peras certificate must contain at least one voter"
  | otherwise =
      fixedCertificateSize
        + bitmapMaximumIndexSize
        + bitmapByteStringHeaderSize
        + bitmapPayloadSize
        + nonPersistentProofListOverhead
        + nonPersistentProofsSize
  where
    totalVoters = persistentVoters + nonPersistentVoters

    -- Fixed benchmark-fixture fields and structural CBOR:
    -- certificate header, round number, boosted block, voter/bitmap headers,
    -- and aggregate signature.
    fixedCertificateSize = 108

    bitmapMaximumIndexSize =
      cborEncodingSize (totalVoters - 1)

    bitmapPayloadSize = ceilDiv totalVoters 8

    bitmapByteStringHeaderSize =
      cborEncodingSize bitmapPayloadSize

    -- The proof list is encoded indefinitely: 0x9f ... 0xff.
    nonPersistentProofListOverhead = 1

    nonPersistentProofsSize =
      nonPersistentVoters * vrfOutputSize

    vrfOutputSize = 50

    ceilDiv n d = (n + d - 1) `div` d

    cborEncodingSize n
      | n <= 23 = 1
      | n <= 0xff = 2
      | n <= 0xffff = 3
      | n <= 0xffffffff = 5
      | otherwise = 9
```

Maybe this could be simplified a bit to instead compute an upper bound? That is assuming that we don't expect to ever be _that_ close to the actual limit such that being this precise would matter.

agustinmista: Check out [this commit](https://github.com/IntersectMBO/ouroboros-consensus/commit/ddd052c5345ee2e2d5d21d53129c218a3bce8d0e) for more details.

nbacquey: Used that in the latest version. Thanks!

nbacquey:
@agustinmista I implemented the changes we discussed:
- moved the upper bound function to `O.C.Peras.Cert.V1`
- implemented unit tests to check correctness and tightness of the upper bound

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -23,8 +23,13 @@
 -- performed on the certificate later on.
 module Ouroboros.Consensus.Peras.Cert.V1
   ( PerasCert (..)
-  , castPerasCert
   , PerasCertVoters (..)
+  , castPerasCert
```
agustinmista, line 27: NIT: any reason to reorder these?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -23,8 +23,13 @@
...
+  , perasCertSizeUpperBound
+
+    -- * Exported for testing purposes only
+  , numberOfNonPersistentVoters
+  , numberOfVoters
```
agustinmista: Could we maybe prefix these with `perasCert*`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Params.hs
```diff
@@ -241,6 +244,13 @@ defaultPerasParams =
 dijkstraPerasRoundLength :: PerasEnabled PerasRoundLength
 dijkstraPerasRoundLength = PerasEnabled (PerasRoundLength 90)
 
+-- | Maximum size of a Peras certificate in bytes, for the Dijkstra eras.
+-- Maximum block size is 90_112 bytes in mainnet as of 2026-08-06. Taking half that
+-- value still allows for extreme leeway:  WFA^LS certificates would only reach that
```
agustinmista: Consistency NIT: either `wFA^LS` (the algorithm) or `WFALS` (the type associated to that algorithm)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -405,3 +411,33 @@ instance
           Nothing
         nonPersistentSeats ->
           Just (NonEmpty.fromList nonPersistentSeats)
+
+perasCertSizeUpperBound :: PerasCert blk -> PerasCertSize
+perasCertSizeUpperBound cert =
+  (`div` 8) $
```
agustinmista: I know this is an approximation, but shouldn't this bet the `ceiling` of the division?

There seem to be a `divCeiling` function in the codebase that does exactly this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Error/V1.hs
```diff
@@ -39,6 +40,9 @@ data PerasError blk
       String
   | PerasTemporaryCertInBlockError
       String
+  | PerasCertTooLargeError
+      PerasCertSize
+      PerasCertSize
```
agustinmista, line 47: Can we add some documentation to note which one is the limit and which one is the actual size?

```suggestion
| PerasCertTooLargeError
      PerasCertSize
      -- ^ Size limit
      PerasCertSize
      -- ^ Actual size
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
```diff
@@ -405,3 +411,33 @@ instance
           Nothing
         nonPersistentSeats ->
           Just (NonEmpty.fromList nonPersistentSeats)
+
+perasCertSizeUpperBound :: PerasCert blk -> PerasCertSize
```
agustinmista, line 420: I think this deserves detailed documentation. Why we need it, why an upper bound, etc.

agustinmista (approved):
Looks pretty good! Only a couple of small nits and missing docs 😸

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
nbacquey: I prefer to export types, then values, as we do for imports

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Cert/V1.hs
nbacquey: It should indeed be the ceiling, but the variation is absorbed by the constant factor anyway. I can use `divCeiling` for consistency.

PR #2188, opened 2026-08-06
Leios: add LeiosTxCache, but don't rely on it at all yet

### LINKED ISSUE #1021: Prototype the TxCache to avoid fetches and revalidation
### Why

The TxCache is essential to make the Leios protocol work-conserving and mitigate against fragementing mempools. That is, we should not download or spend cpu time on transactions we already have / know.

### What

- Establish a baseline for why a transaction cache is needed
  - Show that latency matters?
- Does not need to be persistent
- A transaction cache that can hold enough transactions to:
  - mitigate ~ 10 of unique mempools (twice the max number of high-latency components on our P2P graph, approximated by number of continents)
    - Out of scope: assumption to be validated in another story
  - withstand 10k elections within immutability (max expected elections within 3k/f and mainnet params)
  - work with modest Leios parameters
    - 100kB max EB body
    - 1MB max EB closure
- Use it in fetching logic to determine which txs to fetch upon ~announcement~ EB body arrival
- Optional: Use it in the voting thread to know whether we apply or reapply a tx
  - #1016 likely ready by the time
- Update the `leios-design` with any design decisions
- Show what is better with the transaction cache

### How

- Quickly validate the parameter (is this barely fitting, or fine for 1-2 parameter increments?)
- Determine a concrete size given the requirements above
- Decide whether this in-memory or on-disk
- Implement it
- Profit

nfrisby:
This PR is half of the implementation of Issue https://github.com/input-output-hk/ouroboros-leios/issues/1021. The second half will be for LeiosFetch (and LeiosVoting, once it exists) to actually use the LeiosTxCache.

Suggestions:

- See the Haddock atop the `LeiosTxCache` header for a summary of the new component.
- See https://github.com/IntersectMBO/ouroboros-consensus/pull/2188#discussion_r3731935366 for whether you want to skip the hash table pieces altogther
- Criticize the choice to implement SipHash-2-4 from scratch---although notice the intensive tests.
    - It is not an _on-chain_ hash, so I don't think it needs to live upstream in `cardano-crypto`. But it still maybe deserves review by one of our crypto experts, since if it's attackable then our hash table's performance is unpredictable.
- Do I need to pull the EB announcements sidequest out into a separate PR? (especially adding MsgRollForward processing, but also the order-of-operations regarding offering bodies and sending announcements)
    - It's part of this PR at all because the LeiosTxCache assumes an order of operations, specifically: insert announcement, _then_ insert body, _then_ insert txs. So I had to get the announcement/body order fixed anyway.
- I steered Claude on the unit tests, but I did not review their details whatsoever. Perhaps we want to inject some bugs to see failures?

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -14,8 +14,6 @@
 module LeiosDemoLogic (module LeiosDemoLogic) where
 
 import Cardano.Slotting.Slot (SlotNo (..))
-import Codec.CBOR.Read (deserialiseFromBytes)
-import Codec.CBOR.Write (toStrictByteString)
 import Control.Concurrent.Class.MonadMVar (MVar)
 import qualified Control.Concurrent.Class.MonadMVar as MVar
 import qualified Control.Concurrent.Class.MonadSTM as LazySTM
@@ -28,8 +26,6 @@ import Control.Monad.Primitive (PrimMonad, PrimState)
 import Control.Tracer (Tracer, traceWith)
 import qualified Data.Bits as Bits
 import qualified Data.ByteString as BS
-import qualified Data.ByteString.Lazy as LBS
-import Data.ByteString.Short (ShortByteString, fromShort, toShort)
 import Data.DList (DList)
 import qualified Data.DList as DList
 import Data.Functor (void, (<&>))
@@ -81,6 +77,7 @@ import LeiosDemoTypes
   , LeiosPoint (..)
   , LeiosTx (..)
   , PeerId (..)
+  , SerializedEbBody
   , TraceLeiosKernel (..)
   , TraceLeiosPeer (..)
   , TxHash (..)
@@ -88,13 +85,11 @@ import LeiosDemoTypes
   , hashLeiosTx
   , leiosEbBytesSize
   , maxTxsPerEb
-  , decodeLeiosEb
-  , encodeLeiosEb
   , leiosEbTxs
   , RbHash (..)
   )
 import qualified LeiosDemoTypes as Leios
-import LeiosTxCache (LeiosTxCache (..), ReferencesTxsByHash (..))
+import LeiosTxCache (LeiosTxCache (..))
 import Ouroboros.Consensus.Block
   ( BlockProtocol
   , ConvertRawHash
@@ -135,24 +130,9 @@ traceException tracer toTrace action =
   per tx, and the serialized body is the @b@.
 -------------------------------------------------------------------------------}
 
--- | An EB body as its canonical CBOR bytes: the @b@ stored in the index. Its
--- 'ReferencesTxsByHash' instance decodes it to enumerate the referenced txs.
-newtype SerializedEbBody = MkSerializedEbBody ShortByteString
-
-serializeEbBody :: LeiosEb -> SerializedEbBody
-serializeEbBody = MkSerializedEbBody . toShort . toStrictByteString . encodeLeiosEb
-
-instance ReferencesTxsByHash SerializedEbBody where
-  foldTxReferences f z (MkSerializedEbBody sbs) =
-    V.foldl' (\acc (txh, _sz) -> f acc txh) z (leiosEbTxs eb)
-   where
-    eb = case deserialiseFromBytes decodeLeiosEb (LBS.fromStrict (fromShort sbs)) of
-      Right (_leftover, decoded) -> decoded
-      Left err -> error $ "SerializedEbBody: undecodable: " <> show err
-
--- | Insert an EB announcement into the shadow tx-cache index, keyed by the
--- announced slot, the announcing RB header's hash, and the announced EB hash.
--- Evicted bodies\/txs are discarded (the shadow has no consumer for them yet).
+-- | Insert an EB announcement into the tx-cache index, keyed by the announced
+-- slot, the announcing RB header's hash, and the announced EB hash. Evicted
+-- bodies\/txs are discarded; they can be useful for debugging/etc.
 recordAnnouncementInTxCache ::
   forall blk m.
   (ConvertRawHash blk, HasHeader (Header blk), IOLike m) =>
@@ -164,6 +144,29 @@ recordAnnouncementInTxCache txCache ancHdr point =
   void $ txCache.insertAnnouncement point.pointSlotNo rbh point.pointEbHash
  where
   rbh = MkRbHash (toRawHash (Proxy @blk) (headerHash (ancHeader ancHdr)))
+
+-- | Register a locally-forged EB in the tx-cache: its announcement, its
+-- body, and each of its txs as already-applied (the forger drew them from its
+-- validated mempool, so they are known-valid). This mirrors the receive side,
+-- which splits the same inserts between announcement handling
+-- ('recordAnnouncementInTxCache') and body acquisition. The applied-tagging must
+-- follow 'insertBody', which creates the per-tx entries that the tagging upgrades.
+recordForgedEbAndClosureInTxCache ::
+  Monad m =>
+  Tracer m TraceLeiosKernel ->
+  LeiosTxCache m () () SerializedEbBody ->
+  RbHash ->
+  Leios.ForgedLeiosEb ->
+  m ()
+recordForgedEbAndClosureInTxCache tracer txCache rbh forgedEb = do
+  _ <- txCache.insertAnnouncement point.pointSlotNo rbh point.pointEbHash
+  mSummary <- txCache.insertBody point.pointEbHash (Leios.serializeEbBody eb)
+  forM_ mSummary $ traceWith tracer . TraceLeiosTxCacheEbBody point
+  withLockedInsertAppliedTx txCache $ \w0 step ->
+    foldM (\w (txh, _sz) -> step w txh ()) w0 (leiosEbTxs eb)
+ where
+  point = forgedEb.point
+  eb = forgedEb.body
 
 -----
 
@@ -747,7 +750,7 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db peerId req eb
         traceWith ktracer $ TraceLeiosBlockPointMissing point
         leiosDbInsertEbPoint db point ebBytesSize
         completedByBody <- leiosDbInsertEbBody db point eb
-        mSummary <- txCache.insertBody ebHash (serializeEbBody eb)
+        mSummary <- txCache.insertBody ebHash (Leios.serializeEbBody eb)
         forM_ mSummary $ traceWith ktracer . TraceLeiosTxCacheEbBody point
         traceWith ktracer $ TraceLeiosBlockAcquired point
         forM_ completedByBody $ traceWith ktracer . TraceLeiosBlockTxsAcquired
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -14,6 +14,8 @@
 module LeiosDemoLogic (module LeiosDemoLogic) where
 
 import Cardano.Slotting.Slot (SlotNo (..))
+import Codec.CBOR.Read (deserialiseFromBytes)
+import Codec.CBOR.Write (toStrictByteString)
 import Control.Concurrent.Class.MonadMVar (MVar)
 import qualified Control.Concurrent.Class.MonadMVar as MVar
 import qualified Control.Concurrent.Class.MonadSTM as LazySTM
@@ -26,6 +28,8 @@ import Control.Monad.Primitive (PrimMonad, PrimState)
 import Control.Tracer (Tracer, traceWith)
 import qualified Data.Bits as Bits
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as LBS
+import Data.ByteString.Short (ShortByteString, fromShort, toShort)
 import Data.DList (DList)
 import qualified Data.DList as DList
 import Data.Functor (void, (<&>))
@@ -35,6 +39,7 @@ import qualified Data.IntSet as IntSet
 import Data.List (unfoldr)
 import Data.Map (Map)
 import qualified Data.Map.Strict as Map
+import Data.Proxy (Proxy (..))
 import Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import Data.Set (Set)
@@ -83,14 +88,27 @@ import LeiosDemoTypes
   , hashLeiosTx
   , leiosEbBytesSize
   , maxTxsPerEb
+  , decodeLeiosEb
+  , encodeLeiosEb
+  , leiosEbTxs
+  , RbHash (..)
   )
 import qualified LeiosDemoTypes as Leios
+import LeiosTxCacheIndex
+  ( LeiosTxCacheIndex
+  , ReferencesTxsByHash (..)
+  , insertAnnouncement
+  , insertBody
+  , insertUnappliedTx
+  )
 import Ouroboros.Consensus.Block
   ( BlockProtocol
+  , ConvertRawHash
   , HasHeader
   , Header
   , WithOrigin (NotOrigin)
   , headerHash
+  , toRawHash
   )
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types
   ( SystemTime
@@ -112,6 +130,48 @@ import Ouroboros.Consensus.Util.IOLike (IOLike)
 traceException :: (IOLike m, Exception e) => Tracer m a -> (e -> a) -> m b -> m b
 traceException tracer toTrace action =
   action `catch` \e -> traceWith tracer (toTrace e) >> throwIO e
+
+{-------------------------------------------------------------------------------
+  Shadow LeiosTxCache wiring
+
+  The 'LeiosTxCacheIndex' is maintained (announcements, bodies, and txs inserted
+  at the same sites as the LeiosDb) but not yet consulted, so it changes no
+  observable behavior. The node's index is
+  @'LeiosTxCacheIndex' () () 'SerializedEbBody'@: only presence (@()@) is recorded
+  per tx, and the serialized body is the @b@.
+-------------------------------------------------------------------------------}
+
+-- | An EB body as its canonical CBOR bytes: the @b@ stored in the index. Its
+-- 'ReferencesTxsByHash' instance decodes it to enumerate the referenced txs.
+newtype SerializedEbBody = MkSerializedEbBody ShortByteString
+
+serializeEbBody :: LeiosEb -> SerializedEbBody
+serializeEbBody = MkSerializedEbBody . toShort . toStrictByteString . encodeLeiosEb
+
+instance ReferencesTxsByHash SerializedEbBody where
+  foldTxReferences f z (MkSerializedEbBody sbs) =
+    V.foldl' (\acc (txh, _sz) -> f acc txh) z (leiosEbTxs eb)
+   where
+    eb = case deserialiseFromBytes decodeLeiosEb (LBS.fromStrict (fromShort sbs)) of
+      Right (_leftover, decoded) -> decoded
+      Left err -> error $ "SerializedEbBody: undecodable: " <> show err
+
+-- | Insert an EB announcement into the shadow tx-cache index, keyed by the
+-- announced slot, the announcing RB header's hash, and the announced EB hash.
+-- Evicted bodies\/txs are discarded (the shadow has no consumer for them yet).
+recordAnnouncementInTxCache ::
+  forall blk m.
+  (ConvertRawHash blk, HasHeader (Header blk), IOLike m) =>
+  MVar m (LeiosTxCacheIndex () () SerializedEbBody) ->
+  AnnouncingHeader blk ->
+  LeiosPoint ->
+  m ()
+recordAnnouncementInTxCache txCacheVar ancHdr point =
+  MVar.modifyMVar_ txCacheVar $ \idx ->
+    let rbh = MkRbHash (toRawHash (Proxy @blk) (headerHash (ancHeader ancHdr)))
+        (idx', _evEbs, _evTxs) =
+          insertAnnouncement point.pointSlotNo rbh point.pointEbHash idx
+     in pure idx'
 
 -----
 
@@ -560,6 +620,7 @@ nextLeiosFetchClientCommand ::
   ( MVar m (LeiosOutstanding pid)
   , MVar m ()
   ) ->
+  MVar m (LeiosTxCacheIndex () () SerializedEbBody) ->
   LeiosDbConnection m ->
   PeerId pid ->
   StrictTVar m (Seq LeiosFetchRequest) ->
@@ -573,7 +634,7 @@ nextLeiosFetchClientCommand ::
         (m (Either () (LF.SomeLeiosFetchJob LeiosPoint LeiosEb LeiosTx m)))
         (Either () (LF.SomeLeiosFetchJob LeiosPoint LeiosEb LeiosTx m))
     )
-nextLeiosFetchClientCommand ktracer tracer stopSTM kernelVars db peerId reqsVar responseQ = do
+nextLeiosFetchClientCommand ktracer tracer stopSTM kernelVars txCacheVar db peerId reqsVar responseQ = do
   drainResponses
   StrictSTM.atomically checkOrPeek >>= \case
     Right result -> pure $ Right result
@@ -585,9 +646,9 @@ nextLeiosFetchClientCommand ktracer tracer stopSTM kernelVars db peerId reqsVar 
     pending <- StrictSTM.atomically $ LazySTM.flushTQueue responseQ
     forM_ pending $ \case
       PendingBlockResponse req eb ->
-        msgLeiosBlock ktracer tracer kernelVars db peerId req eb
+        msgLeiosBlock ktracer tracer kernelVars txCacheVar db peerId req eb
       PendingBlockTxsResponse req txs ->
-        msgLeiosBlockTxs ktracer tracer kernelVars db peerId req txs
+        msgLeiosBlockTxs ktracer tracer kernelVars txCacheVar db peerId req txs
 
   -- Non-blocking: return 'Right result' if stop or a request is available,
   -- or 'Left ()' if we'd have to block (caller returns Left blockingLoop).
@@ -655,12 +716,13 @@ msgLeiosBlock ::
   ( MVar m (LeiosOutstanding pid)
   , MVar m ()
   ) ->
+  MVar m (LeiosTxCacheIndex () () SerializedEbBody) ->
   LeiosDbConnection m ->
   PeerId pid ->
   LeiosBlockRequest ->
   LeiosEb ->
   m ()
-msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
+msgLeiosBlock ktracer tracer (outstandingVar, readyVar) txCacheVar db peerId req eb = do
   -- validate it
   let MkLeiosBlockRequest point ebBytesSize = req
   traceWith tracer $ MkTraceLeiosPeer $ "[start] MsgLeiosBlock " <> Leios.prettyLeiosPoint point
@@ -693,6 +755,7 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
         traceWith ktracer $ TraceLeiosBlockPointMissing point
         leiosDbInsertEbPoint db point ebBytesSize
         completedByBody <- leiosDbInsertEbBody db point eb
+        MVar.modifyMVar_ txCacheVar $ pure . insertBody ebHash (serializeEbBody eb)
         traceWith ktracer $ TraceLeiosBlockAcquired point
         forM_ completedByBody $ traceWith ktracer . TraceLeiosBlockTxsAcquired
     -- update NodeKernel state
@@ -831,12 +894,13 @@ msgLeiosBlockTxs ::
   ( MVar m (LeiosOutstanding pid)
   , MVar m ()
   ) ->
+  MVar m (LeiosTxCacheIndex () () SerializedEbBody) ->
   LeiosDbConnection m ->
   PeerId pid ->
   LeiosBlockTxsRequest ->
   V.Vector LeiosTx ->
   m ()
-msgLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) db peerId req txs = do
+msgLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) txCacheVar db peerId req txs = do
   traceWith tracer $ MkTraceLeiosPeer $ "[start] " ++ Leios.prettyLeiosBlockTxsRequest req
   -- validate it
   -- TODO: could validate the returned point + bitmaps too (added to response recently)
@@ -864,6 +928,10 @@ msgLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) db peerId req txs = d
   traceException tracer TraceLeiosPeerDbException $ do
     completed <- leiosDbInsertTxs db (V.toList $ V.zip txHashes txBytess)
     forM_ completed $ traceWith ktracer . TraceLeiosBlockTxsAcquired
+    -- crucially: add the txs to the TxCacheIndex _after_ they've been written
+    -- to the LeiosDb, since it's currently what the TxCacheIndex is indexing
+    MVar.modifyMVar_ txCacheVar $ \idx ->
+      pure $ V.foldl' (\i txh -> insertUnappliedTx txh () i) idx txHashes
   -- update NodeKernel state
   MVar.modifyMVar_ outstandingVar $ \outstanding -> do
     let (requestedTxPeers', reverseEbIndexByTx', txsBytesSize) =
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -14,8 +14,6 @@
 module LeiosDemoLogic (module LeiosDemoLogic) where
 
 import Cardano.Slotting.Slot (SlotNo (..))
-import Codec.CBOR.Read (deserialiseFromBytes)
-import Codec.CBOR.Write (toStrictByteString)
 import Control.Concurrent.Class.MonadMVar (MVar)
 import qualified Control.Concurrent.Class.MonadMVar as MVar
 import qualified Control.Concurrent.Class.MonadSTM as LazySTM
@@ -28,8 +26,6 @@ import Control.Monad.Primitive (PrimMonad, PrimState)
 import Control.Tracer (Tracer, traceWith)
 import qualified Data.Bits as Bits
 import qualified Data.ByteString as BS
-import qualified Data.ByteString.Lazy as LBS
-import Data.ByteString.Short (ShortByteString, fromShort, toShort)
 import Data.DList (DList)
 import qualified Data.DList as DList
 import Data.Functor (void, (<&>))
@@ -90,6 +86,7 @@ import LeiosDemoTypes
   , LeiosPoint (..)
   , LeiosTx (..)
   , PeerId (..)
+  , SerializedEbBody
   , TraceLeiosKernel (..)
   , TraceLeiosPeer (..)
   , TxHash (..)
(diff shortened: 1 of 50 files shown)


nfrisby:
Most of the lines of code are: the big comments in `LeiosTxCache.hs`, the reference impl and its tests, the hash table and its tests.

```
$ git diff --stat origin/leios-prototype HEAD
 ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node.hs                                   |   1 +
 ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs                       |  44 ++++-------
 ouroboros-consensus-cardano/src/unstable-byron-testlib/Ouroboros/Consensus/ByronDual/Node.hs              |   1 +
 ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs             |   7 +-
 ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs |  59 ++++++++++----
 ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs         |  99 +++++++++++++++++++-----
 ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs   |  36 +++++++--
 ouroboros-consensus-diffusion/src/unstable-diffusion-testlib/Test/ThreadNet/Network.hs                    |   2 +-
 ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/PeerSimulator/ChainSync.hs               |   2 +-
 ouroboros-consensus.cabal                                                                                 |  51 ++++++++++++
 ouroboros-consensus/bench/leios-txcache-bench/Main.hs                                                     | 289 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs                                             | 161 +++++++++++++++++++++++++++++++-------
 ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs                                             |  56 +++++++++++++-
 ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache.hs                                               | 413 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/API.hs                                           | 113 +++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/Optimized.hs                                     | 349 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/Optimized/MutableHashTable.hs                    | 367 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/Reference.hs                                     | 376 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Forging.hs                          |  11 ++-
 ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Embed/Unary.hs        |   6 +-
 ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Forging.hs            |  54 +++++++------
 ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs          |  21 ++---
 ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Node.hs                              |   1 +
 ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Node/PBFT.hs                         |   1 +
 ouroboros-consensus/src/unstable-mock-block/Ouroboros/Consensus/Mock/Node/Praos.hs                        |   1 +
 ouroboros-consensus/test/consensus-test/Main.hs                                                           |   6 ++
 ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/CSJ.hs                      |   2 +-
 ouroboros-consensus/test/consensus-test/Test/Consensus/MiniProtocol/ChainSync/Client.hs                   |   2 +-
 ouroboros-consensus/test/consensus-test/Test/LeiosTxCache/Optimized.hs                                    | 153 ++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/test/consensus-test/Test/LeiosTxCache/Optimized/MutableHashTable.hs                   | 404 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/test/consensus-test/Test/LeiosTxCache/Reference.hs                                    | 357 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ouroboros-consensus/test/leios-txcache-bounds-checked/Main.hs                                             |  12 +++
 32 files changed, 3306 insertions(+), 151 deletions(-)
```

nfrisby:
The same `nfrisby/leios-first-txcache-increment` branch exists in `ouroboros-leios` and `cardano-node` if you want to build it.

- The node branch handles the new tracer and updates the SRP to point at the (current) tip of this PR.
- The leios branch adds a panel to the proto-devnet Grafana "Throughput" dashboard that shows what percentage of the body's txs are AlreadyAcquired in the LeiosTxCache (TODO excluding the Mempool!!!) when that body arrives (or is locally forged).

nfrisby:
Benchmarks, on my laptop:

`Reference` impl:

```
$ cabal bench leios-txcache-bench --benchmark-options=pure
Configuration is affected by the following files:
- cabal.project
Build profile: -w ghc-9.12.2 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-3.0.1.0 (bench:leios-txcache-bench) (first run)
Created semaphore called cabal_semaphore_7 with 5 slots.
Preprocessing benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Building benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Running 1 benchmarks...
Benchmark leios-txcache-bench: RUNNING...
LeiosTxCache worst-case benchmark
  EBs in index : 128
  txs per EB   : 15058
  total txs    : 1927424

generating data... done
populating index... done

== pure-wrapped index ==
population:
  time             : 13.80 s
  allocated        : 5.06 GiB
  alloc rate       : 375.6 MiB/s
residency (post major GC):
  live             : 367.8 MiB
  max live         : 367.8 MiB
  peak mem in use  : 898.0 MiB
lookup — EB batch of 15058 (x20 each):
    0% hits (0/15058): avg 25.54 ms, per-tx 1695.9 ns, alloc 2.1 MiB
   20% hits (3012/15058): avg 25.22 ms, per-tx 1674.5 ns, alloc 2.3 MiB
   40% hits (6024/15058): avg 24.13 ms, per-tx 1602.7 ns, alloc 2.5 MiB
   60% hits (9036/15058): avg 23.68 ms, per-tx 1572.4 ns, alloc 3.0 MiB
   80% hits (12047/15058): avg 22.82 ms, per-tx 1515.6 ns, alloc 3.1 MiB
  100% hits (15058/15058): avg 21.84 ms, per-tx 1450.5 ns, alloc 3.3 MiB

batch durations (ascending per column; 20 rows x 6 ratios):
             0%        20%        40%        60%        80%       100%
       25.28 ms   24.84 ms   23.93 ms   23.25 ms   22.43 ms   21.47 ms
       25.31 ms   24.85 ms   23.96 ms   23.26 ms   22.45 ms   21.51 ms
       25.35 ms   24.87 ms   23.97 ms   23.29 ms   22.46 ms   21.53 ms
       25.36 ms   24.92 ms   23.97 ms   23.30 ms   22.47 ms   21.53 ms
       25.36 ms   24.92 ms   24.01 ms   23.32 ms   22.49 ms   21.56 ms
       25.43 ms   24.93 ms   24.07 ms   23.33 ms   22.51 ms   21.58 ms
       25.44 ms   24.96 ms   24.08 ms   23.37 ms   22.54 ms   21.59 ms
       25.46 ms   25.01 ms   24.08 ms   23.39 ms   22.54 ms   21.59 ms
       25.47 ms   25.06 ms   24.09 ms   23.41 ms   22.54 ms   21.64 ms
       25.50 ms   25.09 ms   24.09 ms   23.41 ms   22.57 ms   21.67 ms
       25.50 ms   25.10 ms   24.11 ms   23.43 ms   22.60 ms   21.68 ms
       25.50 ms   25.13 ms   24.14 ms   23.45 ms   22.62 ms   21.69 ms
       25.51 ms   25.13 ms   24.18 ms   23.46 ms   22.62 ms   21.70 ms
       25.54 ms   25.17 ms   24.18 ms   23.48 ms   22.64 ms   21.71 ms
       25.55 ms   25.23 ms   24.19 ms   23.55 ms   22.64 ms   21.76 ms
       25.55 ms   25.24 ms   24.24 ms   23.59 ms   22.65 ms   21.86 ms
       25.70 ms   25.25 ms   24.33 ms   23.70 ms   22.71 ms   22.02 ms
       25.81 ms   25.28 ms   24.33 ms   23.72 ms   22.71 ms   22.33 ms
       25.89 ms   25.83 ms   24.34 ms   24.25 ms   22.91 ms   22.77 ms
       26.23 ms   27.49 ms   24.39 ms   27.62 ms   27.33 ms   23.65 ms
Benchmark leios-txcache-bench: FINISH
```

`Optimized` impl:

```
$ cabal bench leios-txcache-bench --benchmark-options=ht
Configuration is affected by the following files:
- cabal.project
Build profile: -w ghc-9.12.2 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-3.0.1.0 (bench:leios-txcache-bench) (first run)
Created semaphore called cabal_semaphore_b with 5 slots.
Preprocessing benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Building benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Running 1 benchmarks...
Benchmark leios-txcache-bench: RUNNING...
LeiosTxCache worst-case benchmark
  EBs in index : 128
  txs per EB   : 15058
  total txs    : 1927424

generating data... done
populating index... done

== hash-table (shift 22) ==
population:
  time             : 887.42 ms
  allocated        : 945.0 MiB
  alloc rate       : 1.04 GiB/s
residency (post major GC):
  live             : 219.9 MiB
  max live         : 321.1 MiB
  peak mem in use  : 429.0 MiB
lookup — EB batch of 15058 (x20 each):
    0% hits (0/15058): avg 1.75 ms, per-tx 116.3 ns, alloc 2.5 MiB
   20% hits (3012/15058): avg 1.84 ms, per-tx 122.5 ns, alloc 2.5 MiB
   40% hits (6024/15058): avg 1.77 ms, per-tx 117.7 ns, alloc 2.7 MiB
   60% hits (9036/15058): avg 1.75 ms, per-tx 116.5 ns, alloc 2.7 MiB
   80% hits (12047/15058): avg 2.20 ms, per-tx 146.2 ns, alloc 2.9 MiB
  100% hits (15058/15058): avg 1.91 ms, per-tx 127.1 ns, alloc 2.9 MiB

batch durations (ascending per column; 20 rows x 6 ratios):
             0%        20%        40%        60%        80%       100%
        1.45 ms    1.54 ms    1.57 ms    1.57 ms    1.59 ms    1.55 ms
        1.46 ms    1.58 ms    1.58 ms    1.58 ms    1.61 ms    1.58 ms
        1.47 ms    1.59 ms    1.58 ms    1.61 ms    1.61 ms    1.61 ms
        1.47 ms    1.60 ms    1.61 ms    1.62 ms    1.64 ms    1.62 ms
        1.48 ms    1.61 ms    1.61 ms    1.62 ms    1.68 ms    1.63 ms
        1.49 ms    1.61 ms    1.62 ms    1.63 ms    1.69 ms    1.64 ms
        1.51 ms    1.62 ms    1.62 ms    1.65 ms    1.73 ms    1.72 ms
        1.52 ms    1.62 ms    1.62 ms    1.65 ms    1.81 ms    1.79 ms
        1.53 ms    1.63 ms    1.64 ms    1.66 ms    1.82 ms    1.80 ms
        1.55 ms    1.63 ms    1.66 ms    1.67 ms    1.87 ms    1.85 ms
        1.56 ms    1.64 ms    1.68 ms    1.67 ms    1.95 ms    1.85 ms
        1.59 ms    1.65 ms    1.70 ms    1.69 ms    2.04 ms    1.87 ms
        1.60 ms    1.67 ms    1.72 ms    1.69 ms    2.13 ms    1.90 ms
        1.78 ms    1.68 ms    1.75 ms    1.72 ms    2.27 ms    1.99 ms
        1.82 ms    1.69 ms    1.79 ms    1.77 ms    2.32 ms    2.08 ms
        1.97 ms    1.72 ms    1.83 ms    1.78 ms    2.46 ms    2.10 ms
        2.32 ms    2.04 ms    1.90 ms    1.79 ms    3.03 ms    2.11 ms
        2.44 ms    2.10 ms    2.16 ms    1.79 ms    3.05 ms    2.16 ms
        2.51 ms    2.51 ms    2.29 ms    2.20 ms    3.61 ms    2.35 ms
        2.53 ms    4.16 ms    2.51 ms    2.74 ms    4.12 ms    3.08 ms
Benchmark leios-txcache-bench: FINISH
```

ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/Optimized/MutableHashTable.hs
```diff
@@ -0,0 +1,367 @@
+{-# LANGUAGE BangPatterns #-}
+{-# OPTIONS_GHC -O2 #-}
+
+-- | A mutable, salted open-addressing hash table for the Leios tx cache: linear
```
nfrisby, line 4: Issue https://github.com/input-output-hk/ouroboros-leios/issues/1021 says:

> - work with modest Leios parameters
>     - 100kB max EB body
>     - 1MB max EB closure

I didn't do that; I targeted the full 512 kB body and 12 MB closure case instead.

Thus:

- Maybe the `Reference` impl is fine, no hash table needed? If the EB body is only 20% of the size this PR was expecting (so ≤\~3000 txs instead of ≤\~15000), then the `Map TxHash` worst-case will be 20% of what the benchmark reports. Maybe that's fine? (128 * ~3000  = ~384000, which is still a _pretty big_ `Map`, and every major GC has to walk the whole thing).
- The 12 MB versus 1 MB closure doesn't relate to the hash table... except for the fact that if we don't need the hash table, then we could just have the `Map` contain the txs' bytes directly, since that would only be at most ~128 MB (+ GHC heap overhead).
    - But that doesn't actually save much; merely side-steps having to read the txs from the LeiosDb.
    - If we want that, it'd be easy: the `Reference` map is already parameterized by `a=()`; we'd simply change that to `a=ShortByteString`, I think? (This PR doesn't alter LeiosFetch at all yet, so I'm not sure what it'd actually need.)

```
$ cabal bench leios-txcache-bench --benchmark-options=pure
Configuration is affected by the following files:
- cabal.project
Build profile: -w ghc-9.12.2 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-3.0.1.0 (bench:leios-txcache-bench) (file ouroboros-consensus/bench/leios-txcache-bench/Main.hs changed)
Created semaphore called cabal_semaphore_b with 5 slots.
Preprocessing benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Building benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
[1 of 1] Compiling Main             ( ouroboros-consensus/bench/leios-txcache-bench/Main.hs, /home/nfrisby/cardano-haskell/cardano-node/dist-newstyle/build/x86_64-linux/ghc-9.12.2/ouroboros-consensus-3.0.1.0/b/leios-txcache-bench/build/leios-txcache-bench/leios-txcache-bench-tmp/Main.o ) [Source file changed]
[2 of 2] Linking /home/nfrisby/cardano-haskell/cardano-node/dist-newstyle/build/x86_64-linux/ghc-9.12.2/ouroboros-consensus-3.0.1.0/b/leios-txcache-bench/build/leios-txcache-bench/leios-txcache-bench [Objects changed]
Running 1 benchmarks...
Benchmark leios-txcache-bench: RUNNING...
LeiosTxCache worst-case benchmark
  EBs in index : 128
  txs per EB   : 3011
  total txs    : 385408

generating data... done
populating index... done

== pure-wrapped index ==
population:
  time             : 1.62 s
  allocated        : 953.6 MiB
  alloc rate       : 589.8 MiB/s
residency (post major GC):
  live             : 73.6 MiB
  max live         : 73.6 MiB
  peak mem in use  : 186.0 MiB
lookup — EB batch of 3011 (x20 each):
    0% hits (0/3011): avg 2.57 ms, per-tx 852.6 ns, alloc 602.8 KiB
   20% hits (603/3011): avg 2.25 ms, per-tx 746.9 ns, alloc 601.9 KiB
   40% hits (1205/3011): avg 2.13 ms, per-tx 706.1 ns, alloc 602.2 KiB
   60% hits (1807/3011): avg 2.04 ms, per-tx 676.9 ns, alloc 603.0 KiB
   80% hits (2409/3011): avg 1.97 ms, per-tx 655.3 ns, alloc 755.2 KiB
  100% hits (3011/3011): avg 1.93 ms, per-tx 641.6 ns, alloc 805.8 KiB

batch durations (ascending per column; 20 rows x 6 ratios):
             0%        20%        40%        60%        80%       100%
        1.87 ms    1.83 ms    1.64 ms    1.60 ms    1.51 ms    1.46 ms
        1.92 ms    1.85 ms    1.66 ms    1.64 ms    1.52 ms    1.47 ms
        1.95 ms    1.90 ms    1.66 ms    1.67 ms    1.60 ms    1.49 ms
        1.99 ms    1.92 ms    1.74 ms    1.73 ms    1.63 ms    1.54 ms
        2.09 ms    1.92 ms    1.87 ms    1.73 ms    1.71 ms    1.67 ms
        2.18 ms    1.93 ms    1.91 ms    1.77 ms    1.76 ms    1.74 ms
        2.31 ms    1.94 ms    1.92 ms    1.80 ms    1.78 ms    1.78 ms
        2.35 ms    2.00 ms    1.96 ms    1.85 ms    1.84 ms    1.85 ms
        2.38 ms    2.06 ms    2.02 ms    1.89 ms    1.85 ms    1.88 ms
        2.51 ms    2.11 ms    2.13 ms    1.91 ms    1.88 ms    1.89 ms
        2.54 ms    2.18 ms    2.14 ms    2.00 ms    1.89 ms    1.90 ms
        2.57 ms    2.20 ms    2.15 ms    2.03 ms    1.96 ms    1.94 ms
        2.59 ms    2.35 ms    2.16 ms    2.08 ms    1.97 ms    1.98 ms
        2.72 ms    2.40 ms    2.16 ms    2.15 ms    2.02 ms    2.04 ms
        2.78 ms    2.41 ms    2.20 ms    2.16 ms    2.05 ms    2.07 ms
        2.79 ms    2.47 ms    2.34 ms    2.20 ms    2.07 ms    2.11 ms
        3.05 ms    2.47 ms    2.35 ms    2.25 ms    2.07 ms    2.15 ms
        3.15 ms    2.63 ms    2.37 ms    2.25 ms    2.11 ms    2.19 ms
        3.18 ms    2.81 ms    2.67 ms    2.57 ms    2.41 ms    2.34 ms
        4.43 ms    3.59 ms    3.47 ms    3.49 ms    3.83 ms    3.16 ms
Benchmark leios-txcache-bench: FINISH
```

nfrisby: And:

```
$ (cd ../ouroboros-consensus/; git diff)
diff --git a/ouroboros-consensus/bench/leios-txcache-bench/Main.hs b/ouroboros-consensus/bench/leios-txcache-bench/Main.hs
index 5abb36f7f..c87b81415 100644
--- a/ouroboros-consensus/bench/leios-txcache-bench/Main.hs
+++ b/ouroboros-consensus/bench/leios-txcache-bench/Main.hs
@@ -49,7 +49,7 @@ numEbs = maxAnnouncementCount
 -- hash + 2-byte size) in a compact, non-CBOR layout — the encoding-independent
 -- ceiling. (The current CBOR-precise 'LeiosDemoTypes.maxTxsPerEb' is 13888.)
 txsPerEb :: Int
-txsPerEb = 15_058
+txsPerEb = 15_058 `div` 5
 
 -- | Timed repetitions of the batch lookup (plus one warmup).
 numLookupRuns :: Int
@@ -103,10 +103,10 @@ main = do
       runs :: [(String, IO BenchCache)]
       runs = case args of
         ["pure"] -> [("pure-wrapped index", newPureLeiosTxCache)]
-        ["ht"] -> [("hash-table (shift 22)", newHashTableLeiosTxCache 22 salt0 salt1)]
+        ["ht"] -> [("hash-table (shift 20)", newHashTableLeiosTxCache 20 salt0 salt1)]
         _ ->
           [ ("pure-wrapped index", newPureLeiosTxCache)
-          , ("hash-table (shift 22)", newHashTableLeiosTxCache 22 salt0 salt1)
+          , ("hash-table (shift 20)", newHashTableLeiosTxCache 20 salt0 salt1)
           ]
   mapM_ (uncurry runBench) runs
 

$ cabal bench leios-txcache-bench --benchmark-options=ht
Configuration is affected by the following files:
- cabal.project
Build profile: -w ghc-9.12.2 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-3.0.1.0 (bench:leios-txcache-bench) (first run)
Created semaphore called cabal_semaphore_c with 5 slots.
Preprocessing benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Building benchmark 'leios-txcache-bench' for ouroboros-consensus-3.0.1.0...
Running 1 benchmarks...
Benchmark leios-txcache-bench: RUNNING...
LeiosTxCache worst-case benchmark
  EBs in index : 128
  txs per EB   : 3011
  total txs    : 385408

generating data... done
populating index... done

== hash-table (shift 20) ==
population:
  time             : 146.86 ms
  allocated        : 190.2 MiB
  alloc rate       : 1.26 GiB/s
residency (post major GC):
  live             : 52.5 MiB
  max live         : 81.1 MiB
  peak mem in use  : 111.0 MiB
lookup — EB batch of 3011 (x20 each):
    0% hits (0/3011): avg 358.67 µs, per-tx 119.1 ns, alloc 602.8 KiB
   20% hits (603/3011): avg 341.45 µs, per-tx 113.4 ns, alloc 602.3 KiB
   40% hits (1205/3011): avg 361.91 µs, per-tx 120.2 ns, alloc 602.6 KiB
   60% hits (1807/3011): avg 290.13 µs, per-tx 96.4 ns, alloc 602.9 KiB
   80% hits (2409/3011): avg 401.07 µs, per-tx 133.2 ns, alloc 468.7 KiB
  100% hits (3011/3011): avg 282.80 µs, per-tx 93.9 ns, alloc 448.8 KiB

batch durations (ascending per column; 20 rows x 6 ratios):
             0%        20%        40%        60%        80%       100%
      297.48 µs  277.78 µs  273.22 µs  274.18 µs  317.12 µs  274.50 µs
      297.75 µs  280.25 µs  273.76 µs  274.65 µs  318.08 µs  274.63 µs
      298.18 µs  285.64 µs  309.87 µs  274.67 µs  319.24 µs  275.14 µs
      298.24 µs  301.16 µs  314.82 µs  275.57 µs  319.61 µs  275.24 µs
      301.19 µs  308.46 µs  315.06 µs  276.55 µs  328.44 µs  275.45 µs
      306.88 µs  314.30 µs  315.61 µs  277.53 µs  329.10 µs  275.56 µs
      310.56 µs  314.90 µs  315.66 µs  278.09 µs  341.58 µs  276.29 µs
      312.27 µs  315.50 µs  315.93 µs  279.55 µs  346.31 µs  276.71 µs
      322.71 µs  316.06 µs  317.01 µs  281.43 µs  351.39 µs  276.98 µs
      331.28 µs  319.27 µs  317.02 µs  285.00 µs  351.81 µs  277.36 µs
      336.50 µs  324.67 µs  325.87 µs  285.41 µs  358.78 µs  278.16 µs
      340.59 µs  334.38 µs  331.20 µs  285.86 µs  361.42 µs  278.45 µs
      346.26 µs  336.06 µs  339.37 µs  287.26 µs  368.83 µs  279.72 µs
      356.32 µs  341.74 µs  345.32 µs  289.20 µs  381.43 µs  281.43 µs
      370.36 µs  355.89 µs  392.86 µs  293.70 µs  418.50 µs  282.09 µs
      399.16 µs  362.99 µs  424.78 µs  297.65 µs  482.84 µs  286.69 µs
      424.08 µs  396.84 µs  462.74 µs  297.68 µs  497.48 µs  288.40 µs
      451.03 µs  434.97 µs  507.09 µs  300.26 µs  596.73 µs  300.49 µs
      475.05 µs  436.00 µs  507.70 µs  323.20 µs  603.24 µs  302.11 µs
      597.61 µs  472.21 µs  533.34 µs  365.24 µs  629.54 µs  320.61 µs
Benchmark leios-txcache-bench: FINISH
```

nfrisby: The `Reference` numbers at 20% cardinality are eerily close to the `Optimized` numbers at 100% cardinality.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -113,6 +121,54 @@ traceException :: (IOLike m, Exception e) => Tracer m a -> (e -> a) -> m b -> m
 traceException tracer toTrace action =
   action `catch` \e -> traceWith tracer (toTrace e) >> throwIO e
 
+{-------------------------------------------------------------------------------
+  Shadow LeiosTxCache wiring
```
nfrisby, line 138: Issue https://github.com/input-output-hk/ouroboros-leios/issues/1021 includes

> Show what is better with the transaction cache

This PR patently can't do that _by observing a running node_, since it doesn't alter the LeiosFetch decision logic to use LeiosTxCache _instead of_ ~scanning~ querying the `LeiosDb`. (I'd like to defer that to a separate PR... but maybe it'll just end up being a small commit eventually added to this PR.)

But: perhaps I can add yet another variant to `leios-txcache-bench/Main.hs` that runs via the SQLite-backed `LeiosDb`?

nfrisby:
```
commit e7771e77f163b3d4bb7334f8965518a553a7b9b7
Author: Nicolas Frisby <nicolas.frisby@moduscreate.com>
Date:   Thu Aug 6 20:46:06 2026 -0400

    LeiosTxCache: also benchmark a SQLite impl
    
    On my machine
    
    $ lsblk -d -o NAME,MODEL,SIZE,ROTA,TRAN && findmnt -no SOURCE,FSTYPE -T .
    NAME    MODEL               SIZE ROTA TRAN
    nvme0n1 WD_BLACK SN770 2TB  1.8T    0 nvme
    /dev/nvme0n1p2 ext4
    
    I see the following:
    
      ┌──────────────────────┬───────────┬────────┐
      │                      │ no reopen │ reopen │
      ├──────────────────────┼───────────┼────────┤
      │ cache small (512 kB) │ 292 ms    │ 294 ms │
      ├──────────────────────┼───────────┼────────┤
      │ cache big (256 MB)   │ 17 ms     │ 339 ms │
      └──────────────────────┴───────────┴────────┘
    
    This is for the 1.9 million txs case and using sudo so fadvise(DONT_NEED) works.
    
    If I reduce the tx count to 20%, the averages are are reduced to 20%.
```

nfrisby: Claue claims that those numbers match that SSD's expectations and that no (reasonable?) SSD would be better for SQLite

nfrisby: The high-level summary: SQLite relies on two caches, its own page cache and the operating system's page cache (ie Virtual Memory).

If we neuter either of them (and it's not at all clear that the adversary can't send messages that would spoil either/both caches), then the latency for checking a single max-tx-count EB against a full SQLite-backed LeiosTxCache is intolerable.

Even if we neuter neither cache, the 17 ms time isn't _great_.

ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache.hs
```diff
@@ -0,0 +1,413 @@
...
+--
+--     - However, that argument is spoiled by the fact that there's no lower
+--       bound on the arrival rate of EBs. In the simplest case, there might
+--       merely be less load than Praos can handle, so no EBs are /needed/. As a
+--       result, the up-to-128 EBs in the LeiosTxCacheIndex might include some
+--       with ages near/greater than the immutable tip.
+--
+-- Solution: continue to rely on there being hours of slack, but moreover
+-- actively ensure that slack. In particular, evict EBs as they get "too old",
+-- regardless of whether new EBs have been arriving. For example, the
```
nfrisby: TODO reconsider this

> For example, the LeiosTxCacheIndex should evict any EBs that are older than the youngest X RBs on the current selection, for X≥128.

The assumption here is that it always takes longer the immutable tip to advance by k-X blocks than it does for the LeiosFetch logic to _finish_ reading txs from the LeiosDb once it successfully looked them up in the LeiosTxCacheIndex.

I'm wavering on whether that's true.

- For a _caught-up_ and well-connected `mainnet` node, lengthening the chain by k-X takes a very long time.
- But, if the node were eclipsed for a while and then manages to escape it, then it'll suddenly see many blocks. If they're _all empty_, then maybe it could acquire and select k-X blocks "very quickly". ... But faster than the LeiosFetch logic could finish its LeiosDb reads? Seems very unlikely, as long as k-X is great enough.
- And on `mainnet`, k-X is at least 1500 blocks, so that does seem feasible.

But what about testnets, where k-X might be smaller? And also what about a _syncing_ `mainnet` node; is that qualitatively different than the node-escaping-eclipse above?

These questions make it tempting to reconsider the other solutions: either LeiosFetch detecting and recovering from hit-prune faults or else explicit pinning.

- Maybe the recovery isn't that complicated: do a _first_ pass on the EbBody to find the txs that are not in the LeiosTxCacheIndex, and issue jobs for those. Also spawn a read of the backing store for the txs that were in the index. Once that read finishes, issue _more_ jobs for any txs whose read failed. Still monotonic, just two-step instead of all at once.
    - How could the adversary weaponize the extra delay for that second batch of LeiosFetch jobs being issued?
- Each individual pin should be very brief; only for the duration of reading some of a single EB's txs from the LeiosDb.
    - The pinning has to prevent _the LeiosDb_ from deleting the txs---it's not even a restriction on the LeiosTxCache. That coupling seems undesirable, and the only way to avoid it would be to have the LeiosTxCache's backing store be independent. That's doable, but more engineering.
    - If those reads are pipelined and overlap for a sustained period, that might extend the duration significantly. And while something is pinned, we either tolerate that the maximum size of the backing store being temporarily exceeded, or else we block insertions? Neither sounds very appealing.
    - Perhaps relevant observation: EbBodies can arrive in sustained bursts, but their average over longer time frames (hours) will be close to f=0.05 arrival per slot. So, on average, there will often be nothing pinned.

nfrisby: > Maybe the recovery isn't that complicated

In fact... maybe recovery = do nothing?

- This PR reuses the LeiosDb as the backing store. So: if the txs aren't there, then _no volatile EB_ references them.
- The LeiosFetch logic should only be fetching EBs that this node considers to be volatile.
- So: if the txs aren't there, then LeiosFetch doesn't _need_ them; the EB that lead to those reads can't still be volatile.

Edit: ah. The above argument only works if the LeiosDb is deduping txs _and promptly_, which it currently does. The point: if we're fetching txs for some EB, then that body is already in the LeiosDb, which means it's keeping all of its txs we already have alive. In that case, LeiosFetch _doesn't even need to read_ the txs that the index claims are present, so there's no read that can fail. ~So there is no possibility of hit-prune fault: neither LeiosFetch nor LeiosVote ever read the bytes of the txs that the LeiosTxCacheIndex claims are present.~ The hit-prune fault still exists, it just doesn't manifest as promptly as I think. _If the index is stale such that it still contains some EB that was already GC'd out of the LeiosDb_, then a hit might prevent some txs from being fetched that need to be fetched. That either causes the EB closure to never be complete or it causes it to _falsely_ claim to be complete---both are hit-prune faults.

Only if the LeiosDb weren't deduping txs would LeiosFetch need to read the txs the LeiosTxCacheIndex claims to be present, so that it can copy (ie "dup") them to the EB it's currently fetching. In that case, the source EB could be older than the destination EB, in which case the hit-prune fault could happen _and would need to be recovered from_ (by refetching the txs that we _just_ GC'd from some peers).

Edit: A meta-level point: I'll have more clarity here once I've actually implemented the LeiosFetch logic---the reasoning depends on when exactly various things happens, and I can't see that all yet.

nfrisby: OK, here's my summary of the last couple "thinking out loud comments" (I'm writing these so I'll think about them on Monday, but I'm failing to stop thinking about them now, on Sunday :/.)

- LeiosTxCacheIndex hits should prevent LeiosFetch from _immediately_ sending a request to peers for those txs.
- But, _unless we trust the "hours of slack" argument_ or we _add explicit synchronization_, then we need a detection and recovery mechanism for the case were the cache index hits were stale: we actually _do_ need to fetch those txs from peers.
- Suppose X was a tx that had a cache hit.
    - LeiosFetch will immediately sends requests for the txs that weren't cache hits ✔️ 
    - LeiosFetch should now double-check that X is in the LeiosDb
    - If it is, then suppressing the requests for X was correct ✔️ 
    - If it isn't, then now we do need to send those requests after all. They're slightly delayed compared to if the cache hadn't been stale but this case is a rarity (TODO confirm adversary can't induce it).
- The LeiosFetch bookkeeping can still be monotonic, it just has to be slightly more complicated.
    - Both cache hits and cache misses lead to jobs being created, but they start as _separate_ jobs: jobs for peer versus jobs for the "double-check" logic. Key point: the total number of outstanding jobs can already be set to the greatest value it'll ever have.
    - The double-check logic processes its jobs by converting them into jobs-for-peers whenever the cache hit turned out to be stale. There can be at most one converted job per original job, so the double-check logic can cleverly decrement the outstanding-job count to keep it accurate.
        - A simple version of the rule: initialize the outstanding-jobs counter with one double-check job per tx with a cache hit.
            - On each confirmed-present tx, decrement the outstanding job count (can be batched).
            - On each confirmed-absent tx, add it to an accumulating for-peers job, batched/flushed on some practical heuristics.
            -  On each flush, issue that for-peers jobs and decrement the outstanding-job count by _one less_ than the number of txs in that new for-peers job.

nfrisby: Hrmm...

- Assume the LeiosDb is deduping.
    - In particular, we're assuming that once the EbBody is inserted into the LeiosDb, any of its txs that are already in the database cannot be GC'd _before this EB is GC'd_.
- Assume that the LeiosDb pruning function's first step is to prune the LeiosTxCacheIndex.
    - Thus, the cache never falsely claims some tx that was pruned out is present.
- Now, suppose the EbBody arrival handler does these steps in the following order.
    - Insert EbBody into LeiosDb.
    - Insert EbBody into LeiosTxCacheIndex.
    - Issue fetch requests to peers for the EbBody's txs that are in the TxNotYetInserted state in the LeiosTxCacheIndex.

I think that's enough. I'm not seeing what could go wrong.

The key point is: the EB itself is now keeping those txs alive. So the only way reads of those txs could fail is if the EB itself was GC'd... but that means those fails no longer matter; the code that was expecting the read results should also abort, since it's processing an orphaned EB.

There's one downside: a deduping insert like that is going to involve non-trivial on-disk latency, so the requests to peers would be delayed by at least that much.

However, it does seem like a reasonable place to start: no locking and no recovery.

*MOREOVER*, the latency might not be that bad: the general expectation is that EbBody's will arrive _before_ peers offer the corresponding EbClosure. So as long as this work finishesd before the offers arrive, the "injected latency" is hidden. (The "Issue fetch requests to peers" above is misleading: the EbBody handler could only do that when those peers had already offer that EB's closure, which should be a rare case: EbBodys should usually diffuse faster than their closures... unless the closures are already present, in which case the truly slow part of diffusion already doesn't exist.)

Edit: it'd require some more complexity, but it should be safe to offer the EbBody to our downstream peers as soon as we've confirmed its hash&size match. It's possible their request might arrive before we've finished inserting it into the LeiosDb---that's the complex case we'd have to support. For example, we could insert it into an in-memory store before we start the LeiosDb write and then delete it from that store when the write is over. Our LeiosFetch servers would handle a request by checking the in-memory store first and then falling through to the LeiosDb. Something like that 🤷 would be able to prevent the LeiosDb latency from slowing down the diffusion of EbBodies.

ch1bo (approved):
I have not read this code! However, I was using the API and running this a lot while working on #2235, including several night long runs and stress tests. I might be able to review this today, but would also be comfortable with merging to the `leios-prototype` as-is.

PR #2191, opened 2026-08-09
Seed initial stake snapshots on epoch-0/1 hard fork

ch1bo:
For early-bootstrap (test) networks that hard-fork into an era at epoch 0 or 1 (CardanoTriggerHardForkAtEpoch -- test-only; real networks use AtDefaultVersion), copy the genesis-seeded ssStakeMark into ssStakeSet/ssStakeGo so the stake distribution -- and hence the Leios committee, which reads nesPd -- is fully active from the first epochs instead of only after the ~2-epoch stake snapshot pipeline. hard fork at epoch 0 => go = set = mark; epoch 1 => set = mark; epoch >= 2 / mainnet => unchanged.

Benefits the Leios ThreadNet tests and any genesis-keyed devnet: the committee is stake-active from epoch 0, so voting and EB certification run immediately.

### DIFF
--- a/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
@@ -52,6 +52,8 @@ import qualified Cardano.Ledger.Api.Era as L
 import qualified Cardano.Ledger.Api.Transition as L
 import qualified Cardano.Ledger.BaseTypes as SL
 import qualified Cardano.Ledger.Shelley.API as SL
+import Cardano.Ledger.Shelley.LedgerState (NewEpochState, esSnapshotsL, nesEsL)
+import Cardano.Ledger.State (ssStakeGoL, ssStakeMarkL, ssStakeSetL)
 import Cardano.Prelude (cborError)
 import qualified Cardano.Protocol.TPraos.OCert as Absolute (KESPeriod (..))
 import qualified Codec.CBOR.Decoding as CBOR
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -958,23 +985,45 @@ protocolInfoCardano (SomeHasFS hasFS) paramsCardano
...
+    shelleyInjection trigger (WrapTransitionConfig tcfg) = fn $ \(Flip stIn) -> Comp $ do
       let stowed = stowLedgerTables stIn
       newNES <-
         L.injectIntoTestState
           hasFS
           tcfg
           (Shelley.shelleyLedgerState stowed)
       pure . Flip . unstowLedgerTables $
-        stowed{Shelley.shelleyLedgerState = newNES}
+        stowed
```
nfrisby, line 1022: I'm thinking of this new behavior as "asserting that the ledger state that 'enters' `EpochNo 0` has an infinitely old `mark` snapshot, and so that state's `set` and `go` snapshots should equal it".

Which seems like a ledger concern (they own the maintenance of these snapshots). So, instead of altering the result of `L.injectIntoTestState`, should this patch instead be upstream in `cardano-ledger`?

There's already even a type for "information needed in order to initialize the test chain's ledger state": `L.TransitionConfig`. Right now it's merely `ShelleyGenesis`, so we'd need to enrich it with either the trigger or a summary of the trigger (ie which snapshots to copy from `mark`).

-----

If you agree with my claim, we could merge this PR with a TODO and open a tech-debt Issue to upstream the patch.

nfrisby (approved):
LGTM, but see comment about upstreaming it

nfrisby:
@agustinmista @nbacquey FYI, for Peras testnets

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
ch1bo: I briefly considered it, without having all this context of course. To me it sounded much nicer to have the seeding be done in one place (where the `CardanoTriggerHardForkAtEpoch` is defined; the consensus package).

Opened this as a follow-up either to have that discussion: https://github.com/IntersectMBO/cardano-ledger/issues/5996

PR #2193, opened 2026-08-10
[Peras 51] Replace old types and type families

agustinmista:
This PR replaces five old definitions:

* The associated `PerasCfg` type from `BlockSupportsPeras`, in favor of using `PerasParams` directly.
* The `HasPeras(Cert|Vote)*` projection type classes, in favor of the consolidated `IsPerasCert` and `IsPerasVote` ones.
* The associated `PerasValidationErr` and `PerasForgeErr` type from `BlockSupportsPeras` in favor of a single `PerasError`.
* The `PerasVoteStake` type, in favor of `VoteWeight`.
* The old KeyHash-based `PerasVoterId` (all call sites were already adapted to `PerasSeatIndex` in a previous PR).

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -322,27 +322,21 @@ class
 
   type PerasVotingCommitteeScheme blk = VoidPerasVotingCommitteeScheme
 
-  -- NOTE: to be removed in favor of 'PerasError'.
-  data PerasValidationErr blk
-
-  -- NOTE: to be removed in favor of 'PerasError'.
-  data PerasForgeErr blk
-
   validatePerasCert ::
     PerasParams blk ->
     PerasCert blk ->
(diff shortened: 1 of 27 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
jasagredo, line 321: I had to lookup again what was this definition for and how it differed from `PerasCert`. I think a better interim name would be `DegenPerasCert` or something that conveys that this is a placeholder definition and will be removed in favor of the per-blk definitions

jasagredo (approved):

jasagredo:
I do wonder though if the changelog might be mentioning entities that were both introduced and deleted in the same release, or updated multiple times. Perhaps we should point an AI to that.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Fair enough. I'm planning on removing these pretty soon, though.

The next planned changes in the pipeline are those removing the degenerate instance and switching to per-blk definitions.

PR #2195, opened 2026-08-10
Fix excessive allocation in mempool snapshotting

jasagredo:


nfrisby (commented):
Diff looks good to me. But on the other hand, I think I approved the PR that introduced this leak :/

Would you validate this change manually and post the "evidence" as a comment on this PR? Two ideas:

- Via Karl's `cardano-ignite` commands?
- The performance regression "smoke" benchmark (reporting just runtime+allocation) we discussed in today's call?
    - If the `tx-centrifuge` dependency is too onerous, we could defer this to a tech-debt follow-up Issue. ... but maybe Claude can make short work of it despite the awkwardness 🤔
    
Edit: @karknu pro-actively validated this PR https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1786433117209709?thread_ts=1786121307.489199&cid=CFKLUH4R0

```
──────────────┬────────────────────────────────────────────────┬────────────┬──────────────┐
│     Run      │                   Fix stage                    │ CPU/served │ Alloc/served │
├──────────────┼────────────────────────────────────────────────┼────────────┼──────────────┤
│ 2026-08-07   │ unfixed 11.1-pre                               │ +72.4%     │ +228.1%      │
├──────────────┼────────────────────────────────────────────────┼────────────┼──────────────┤
│ 2026-08-10-b │ O(1) snapshotting & allocation-free Txid       │ -15.0%     │ -38.5%       │
└──────────────┴────────────────────────────────────────────────┴────────────┴──────────────┘
```

(To be clear: that node also had Karl's prior exploratory patches (eg cherry-picking Damian's `compare @TxId` optimizeation), but they weren't effectively. Then he cherry-picked this PR and the numbers were suddenly good again.)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -487,14 +488,60 @@ computeSnapshot cfg slot st values txTickets =
...
+snapshotFromIS ::
+  forall blk.
+  (LedgerSupportsMempool blk, HasTxId (GenTx blk)) =>
+  InternalState blk ->
+  MempoolSnapshot blk
+snapshotFromIS is =
+  snapshotFromTxSeq
+    validatedTx
+    (isTxs is)
+    (isTxIds is)
```
dnadales, line 504: Do we now depend on the `isTxs` / `isTxIds` invariant for the first time?

Before this change, `snapshotHasTx` and `snapshotTxs` read the same list. They could not disagree. Now `snapshotHasTx` reads `isTxIds`, and the rest reads `isTxs`. That answer goes on the wire, in `LocalTxMonitor/Server.hs:65` and in `mempoolHasTx` at `NodeKernel.hs:929`. At the base commit, `isTxIds` had three writers and no readers. So no code tested the comment at `Common.hs:114-120`.

The two writers also spell the id in different ways. `validateNewTransaction` puts `vtx` in the sequence, but it inserts `txId tx` (`Common.hs:378`). `revalidateTxsFor` uses `txId . txForgetValidated` (`Common.hs:430`). So we depend on `txId (txForgetValidated vtx) == txId tx`. I did not find that law in the code. `txForgetValidated` only says that it discards the evidence (`SupportsMempool.hs:203`). The `txId` docs only cover uniqueness between different transactions.

The law holds for every instance I read. Alonzo is the case I want written down. `applyShelleyTx` applies a changed transaction (`Eras.hs:250`, `isValidTxL .~ IsValid True`). It works because `txIdTx` hashes the body only.

Can we take the id from the validated transaction at `:378`? Something like:

```haskell
, isTxIds = Set.insert (txId (txForgetValidated vtx)) isTxIds
```

Then both writers read the same way. I wonder if the `isTxIds` Haddock also needs one line, to say that the id comes from the validated transaction.

dnadales (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
nfrisby: > Do we now depend on the isTxs / isTxIds invariant for the first time?
> ...
> txId (txForgetValidated vtx) == txId tx

I don't know if we never _explicitly_ required that before.

- We certainly assume that a validated tx has the same semantics as the underlying tx.
- So it "must" have the same TxId (or else there's a bug somewhere, since two txs with the same TxId are supposed to have the same effect when applied to the ledger, assuming neither is rejected by the ledger).

jasagredo: I think it is a foundational-enough property that we don't need to over complicate things

PR #2200, opened 2026-08-11
[Peras 54] Refine Peras voting view and decisions

agustinmista:
In anticipation for the Peras voting thread implementation, this PR extends the Peras voting machinery to:

* Track the candidate block in successful voting decisions,
* Retain the latest on-chain certificate's round,
* Provide an STM-backed view handle, and
* Derive a couple of instances that will be needed for tracing later on.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
@@ -1,4 +1,7 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
@@ -16,6 +19,7 @@
 -- do not denote ignored variables.
 module Ouroboros.Consensus.Peras.Voting.Rules
   ( isPerasVotingAllowed
+  , isPerasVotingAllowedWithHandle
   , PerasVotingRule (..)
   , PerasVotingRulesDecision (..)
   , perasVR1A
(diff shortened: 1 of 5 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
```diff
@@ -1,4 +1,8 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
jasagredo: In general we prefer `Proxy` over this, but I can accept it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
```diff
@@ -175,24 +178,24 @@ data LatestCertSeenView cert
   }
   deriving Show
 
--- | View of the latest certificate present in our preferred chain
+-- | View of the latest certificate present in our preferred chain.
 --
 -- NOTE: if we add more fields here in the future, do not forget to add
 -- strictness annotations as needed.
 newtype LatestCertOnChainView cert
```
jasagredo: Why do we need a `cert` phantom?

jasagredo: And actually do we still need this?

jasagredo (approved):
Approving because my comments are not that critical

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/Rules.hs
agustinmista: Good catch! This was not being used anywhere, so must have slipped through at some point via HLS or something. Removed this pragma :+1:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/View.hs
agustinmista: Fair point. We previously had two "view" types for the two certificates being used in the rules, but with these changes we only have one. I got rid of the `LatestCertOnChainView` type entirely and simplified the code a bit as a result :smile_cat:

PR #2207, opened 2026-08-12
[Peras 70] Fix a query bug in slotToPerasRoundNo'

adithyaov:
This commit fixes a bug in the implementation of `slotToPerasRoundNo'` that would prevent the query from traversing eras beyond the first non-PerasEnabled one.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
```diff
@@ -380,17 +380,21 @@ evalExprInEra EraSummary{..} = \(ClosedExpr e) -> go e
...
     SlotInEra relSlot <- lift $ go expr
     PerasEnabledT $
-      Just $
-        case eraPerasRoundLength of
-          NoPerasEnabled ->
-            NoPerasEnabled
-          PerasEnabled (PerasRoundLength roundLength) ->
+      case eraPerasRoundLength of
+        NoPerasEnabled -> do
+          guardEnd $ \end ->
```
tbagrel1, line 410: Not sure to me why we `guard` only in the `NoPerasEnable` case  🤔 For the `PerasEnable` branch, It would make sense to _not_ resolve to current era's Peras round length if the rel slot is too big to be into this era?

Also I'm wondering why we need this fix, given the comment above these conversion functions:

```hs
  -- Convert between relative values
  --
  -- It may seems paradoxical, but no guards are strictly necessary here:
  -- All the RelToAbs___ conversions implement a strict upper bound check, and
  -- relative values are only used internally in this module. So we know that at
  -- the end of the chain a RelToAbs____ conversion will happen and that will
  -- check the upper bound. (Similarly, an initial AbsToRel___ conversion will
  -- take care of the lower bound check.)
```

(this is something I wrote)

Is it still true? If yes, why do we need to change this function's behavior?

tbagrel1 (commented):
Not sure what we should do with these queries. We tried to rewrite them in a principled way many times, but a bug always slip through the cracks apparently. 

I'm a bit wary of patching just one of them in an ad-hoc way that isn't consistent with the others, but at the same time, if it is needed to run the testnet, then maybe we should do it? I don't have much else to propose unfortunately.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs
tbagrel1: > Also I'm wondering why we need this fix, given the comment above these conversion functions:

Long story short: indeed, **when we return a relative value**, the enclosing `RelToAbs___` properly does the bound check. The problem is that if we return `Just NoPerasEnabled` (as we were before), then there is no value carried, so no bound check, so the value stays a `Just` even if the era is wrong instead of becoming a `Nothing`, so we can't try next era.

That's also why the bound check is needed only in the `NoPerasEnabled` case

tbagrel1 (approved):
After meeting with @agustinmista , we found the reason why all this was needed, and documented it extensively in the code. Thus we can proceed with this PR!

jasagredo (approved):
Did a bit of a dive into why this was necessary. In super short: the Qry goes to the next era when the current era answers Nothing. It was returning always Just NoPerasEnabled, thus the query was interpreted in Byron. Now:
```
          guardEnd $ \end ->  
            relSlot < countSlots (boundSlot end) (boundSlot eraStart)
```
will return Nothing (which moves the query to the next era)

PR #2215, opened 2026-08-13
Fix excessive allocation in mempool snapshotting (forward-port)

jasagredo:
Forward port of #2195

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
@@ -37,6 +37,7 @@ module Ouroboros.Consensus.Mempool.Impl.Common
   , jsonMempoolRejectionDetails
 
     -- * Conversions
+  , snapshotFromIS
   , snapshotFromValidTxs
 
     -- * Ticking a ledger state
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -487,14 +488,60 @@ computeSnapshot cfg slot st values txTickets =
...
   [TxTicket (TxMeasureWithDiffTime blk) (Validated (GenTx blk))] ->
   Point blk ->
   SlotNo ->
   MempoolSnapshot blk
-snapshotFromValidTxs validTxs tipPoint slot =
+snapshotFromValidTxs validTxs =
+  snapshotFromTxSeq
+    id
+    (TxSeq.fromList validTxs)
+    (Set.fromList $ map (txId . txForgetValidated . txTicketTx) validTxs)
```
bladyjoker: Consider using a single pass to construct both, we can have thousands of txs:

```haskell
snapshotFromValidTxs validTxs =
  let (txs, txIds) = foldl' (\((txs, txIds), tx) -> (TxSeq.cons tx txs, Set.insert txIds tx)) (TxSeq.empty, Set.empty) validTxs
  in snapshotFromTxSeq
      id
      txs
      txIds
```

bladyjoker (approved):
Thanks Javier

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
jasagredo: Done!

PR #2217, opened 2026-08-17
Time capped getSnapshotFor

bladyjoker:
Part of https://github.com/input-output-hk/ouroboros-leios/issues/911

DONE
- [x] Introduced TimeBoundedLoop module and tests
- [x] Capped execution `getSnapshotFor`
- [x] Chasing laziness
- [x] Implemented the Mempool.Ledger module that contains "disk access" optimal reapplication logic
  - [ ] tests

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -185,9 +186,9 @@ instance
         -- 'applyLeiosClosure' — silently drops them from
         -- 'shelleyCumulativeTxBytes'. The immutable-DB replay path
         -- already sums them via 'inlineLeiosClosure' → block body.
-        closureBytes =
+        !closureBytes =
           sum (fromIntegral . (^. Core.sizeTxF) <$> innerTxs)
-        lst' =
+        !lst' =
           stowed
             { shelleyLedgerState = nes'
             , shelleyCumulativeTxBytes =
@@ -203,7 +204,7 @@ instance
     ms0 = SL.mkMempoolState nes
 
     -- TODO: Ask ledger for an 'applyTxNoValidation' to replace this
-    applyOne envv ms tx =
+    applyOne envv ms !tx =
       fmap fst
         . SL.ruleApplyTxValidation @"LEDGER" STS.ValidateNone globals envv ms
         $ SL.mkStAnnTx
@@ -306,4 +307,4 @@ deserialiseLeiosTx :: forall era. ShelleyBasedEra era => BS.ByteString -> Tx Top
 deserialiseLeiosTx bs =
   case decodeFullAnnotator (Core.eraProtVerLow @era) "Leios Tx" decCBOR (BL.fromStrict bs) of
     Left err -> error $ "Failed to deserialise Leios tx: " <> show err
-    Right tx -> tx
+    Right !tx -> tx
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
@@ -2,19 +2,32 @@
 {-# LANGUAGE DerivingStrategies #-}
 
 -- | Time-bounded iteration.
+--
+-- This is based on simple single-threaded iterative time budgeting
+-- and as such not suitable for defensive timeouts since no attempts
+-- are made at stopping the computation that breached the time limit.
 module LeiosUtils.TimeBoundedLoop
   ( iterateUntilOrTimeout
   , DoneType (..)
   , StepCount
-  , iterateUntilOrTimeout'
-  , propImmediateEndOnInit
-  , propTimeoutOverrunBoundedByOneStep
+  , iterateUntilOrTimeout_
+  , foldOrTimeout
+  , propIterImmediateEndOnInit
+  , propIterTimeoutOverrunBoundedByOneStep
+  , propFoldEmptyListTerminates
+  , propFoldStepsMatchConsumed
+  , propFoldEndReachedIffConsumedAll
+  , propFoldJustFoldl
   ) where
 
+import Cardano.Prelude (lift)
+import Control.Monad (unless, when)
 import Control.Monad.Class.MonadTime.SI (MonadMonotonicTime (getMonotonicTime), diffTime)
 import Control.Monad.Class.MonadTimer.SI (DiffTime, MonadDelay (..))
+import Control.Monad.Except (ExceptT, throwError)
+import qualified Data.Foldable as Foldable
 
--- | Why 'iterateUntilOrTimeout' stopped.
+-- | Why time capped loop stopped.
 data DoneType
   = -- | The time budget was exhausted before the end of computation was reached.
     TimeoutReached
@@ -25,23 +38,23 @@ data DoneType
 -- | Number of 'stepAction' invocations performed.
 type StepCount = Int
 
--- | @iterateUntilOrTimeout ttl endReached stepAction initSt@ runs
--- @stepAction@ on a state repeatedly until either @endReached@ holds or the
--- wall-clock budget @ttl@ is exceeded.
+-- | @let (val, steps, done) = iterateUntilOrTimeout ttl endReached stepAction initSt@ runs
+-- 'stepAction' on a state repeatedly, with 'initSt' as the initial state value, until either
+-- 'endReached' holds or the wall-clock budget 'ttl' is exceeded.
+-- Returns the final result `val`, the 'steps' it took to compute it and 'done' denoting
+-- whether the computation terminated due to the provided end condition or timeout.
 --
--- The check order per iteration is:
+-- __Strict__ on the accumulator.
 --
---   1. End check (@endReached@).
---   2. Timeout check (against @ttl@).
---   3. Step (@stepAction@).
---
--- In particular, if @endReached initSt@ is already @True@ on entry, the
--- function returns immediately with 'EndReached' and @'StepCount' = 0@.
---
--- __Single-threaded__: everything runs in one thread.  If @stepAction@ blocks
--- longer than @ttl@, the timeout is detected only on the /next/ iteration —
+-- __Single-threaded__: everything runs in one thread.  If 'stepAction' blocks
+-- longer than 'ttl', the timeout is detected only on the /next/ iteration —
 -- the in-flight call is not cancelled.  The total elapsed time can therefore
--- exceed @ttl@ by at most one step duration.
+-- exceed @ttl@ by at most one step duration (see 'propIterTimeoutOverrunBoundedByOneStep').
+--
+-- If the 'endReached' holds on the 'initSt' the loop returns immediately with
+-- @steps == 0@ and @done == EndReached@ (see 'propIterImmediateEndOnInit').
+--
+-- Naming reflects https://hackage.haskell.org/package/monad-loops-0.4.3/docs/Control-Monad-Loops.html#v:iterateUntilM
 iterateUntilOrTimeout ::
   MonadMonotonicTime m =>
   -- | Wall-clock budget
@@ -68,47 +81,165 @@ iterateUntilOrTimeout ttl endReached stepAction initSt = do
             st' <- stepAction st
             go (n + 1) startTime st'
 
-iterateUntilOrTimeout' ::
+-- | Like 'iterateUntilOrTimeout' but discards the step count and 'DoneType',
+-- returning only the final state. Use when you don't need to distinguish a
+-- timeout from a normal completion.
+iterateUntilOrTimeout_ ::
   MonadMonotonicTime m =>
   -- | Wall-clock budget
   DiffTime ->
   -- | End predicate; when @True@, stops with 'EndReached'
   (a -> Bool) ->
-  -- | One step of work,
-  (a -> m a) ->
-  -- | Initial state
-  a ->
-  m a
-iterateUntilOrTimeout' ttl endReached stepAction st = (\(res, _, _) -> res) <$> iterateUntilOrTimeout ttl endReached stepAction st
-
-propImmediateEndOnInit ::
-  MonadMonotonicTime m =>
-  -- | Wall-clock budget
-  DiffTime ->
   -- | One step of work
   (a -> m a) ->
   -- | Initial state
   a ->
-  m Bool
-propImmediateEndOnInit ttl stepAction st = do
-  (_, steps, doneType) <- iterateUntilOrTimeout ttl (const True) stepAction st
-  return $ steps == 0 && doneType == EndReached
+  m a
+iterateUntilOrTimeout_ ttl endReached stepAction st = (\(res, _, _) -> res) <$> iterateUntilOrTimeout ttl endReached stepAction st
 
-propTimeoutOverrunBoundedByOneStep ::
+-- | @let (val, remaining, steps, done) = foldOrTimeout ttl foldAction initSt xs@ runs a time capped
+-- left fold using 'iterateUntilOrTimeout'.
+-- Returns the final result in 'val', the 'remaining' elements of 'xs' that wasn't processed,
+-- the 'steps' it took to compute it and whether the
+-- computation terminated due to the input 'xs' being fully traversed or 'ttl' timeout.
+--
+-- __Strict__ on the accumulator.
+--
+-- Naming reflects https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Monad.html#v:foldM
+--
+-- Properties: propFoldEmptyListTerminates, propFoldEndReachedIffConsumedAll, propFoldStepsMatchConsumed
+foldOrTimeout ::
+  MonadMonotonicTime m =>
+  DiffTime ->
+  (b -> a -> m b) ->
+  b ->
+  [a] ->
+  m (b, [a], StepCount, DoneType)
+foldOrTimeout ttl foldAction initSt xs =
+  (\((xsLeft, finalSt), steps, doneType) -> (finalSt, xsLeft, steps, doneType))
+    <$> iterateUntilOrTimeout
+      ttl
+      (\(xsLeft, !_st) -> null xsLeft)
+      ( \(xsLeft, !st) -> case xsLeft of
+          (x : xsLeft') -> do
+            st' <- foldAction st x
+            return (xsLeft', st')
+          _ -> return (xsLeft, st)
+      )
+      (xs, initSt)
+
+type FoldResult b a = (b, [a], StepCount, DoneType)
+
+-- | 'foldOrTimeout' on an empty list terminates immediately: no change, zero steps,
+-- 'EndReached', and an empty remainder.
+propFoldEmptyListTerminates ::
+  (MonadMonotonicTime m, Eq b) =>
+  DiffTime ->
+  (b -> a -> m b) ->
+  b ->
+  ExceptT (String, FoldResult b a) m (FoldResult b a)
+propFoldEmptyListTerminates ttl foldAction initSt = do
+  res@(val, remaining, steps, done) <- lift $ foldOrTimeout ttl foldAction initSt []
+  unless (null remaining) $ throwError $ ("remaining must be empty", res)
+  unless (steps == 0) $ throwError $ ("steps must be zero", res)
+  unless (done == EndReached) $ throwError $ ("done must be EndReached", res)
+  unless (val == initSt) $ throwError $ ("val must be initSt", res)
+  return res
+
+-- | The step count always equals the number of elements consumed.
+propFoldStepsMatchConsumed ::
+  MonadMonotonicTime m =>
+  DiffTime ->
+  (b -> a -> m b) ->
+  b ->
+  [a] ->
+  ExceptT (String, FoldResult b a) m (FoldResult b a)
+propFoldStepsMatchConsumed ttl foldAction initSt xs = do
+  res@(_, remaining, steps, _) <- lift $ foldOrTimeout ttl foldAction initSt xs
+  unless (steps == length xs - length remaining) $
+    throwError $
+      ("steps must be length xs - length remaining", res)
+  return res
+
+-- | 'EndReached' if and only if the entire input was consumed.
+propFoldEndReachedIffConsumedAll ::
+  MonadMonotonicTime m =>
+  DiffTime ->
+  (b -> a -> m b) ->
+  b ->
+  [a] ->
+  ExceptT (String, FoldResult b a) m (FoldResult b a)
+propFoldEndReachedIffConsumedAll ttl foldAction initSt xs = do
+  res@(_, remaining, _, done) <- lift $ foldOrTimeout ttl foldAction initSt xs
+  when (done == EndReached) $
+    unless (null remaining) $
+      throwError $
+        ("remaining must be empty when end is reached", res)
+  return res
+
+-- | When 'foldAction' is pure then it's just a "Foldable.foldl'" on processed 'xs'
+propFoldJustFoldl ::
+  (MonadMonotonicTime m, Eq b) =>
+  DiffTime ->
+  (b -> a -> b) ->
+  b ->
+  [a] ->
+  ExceptT (String, FoldResult b a) m (FoldResult b a)
+propFoldJustFoldl ttl foldFn initSt xs = do
+  res@(val, _remaining, steps, _done) <-
+    lift $ foldOrTimeout ttl (\st x -> pure $ foldFn st x) initSt xs
+  unless (val == Foldable.foldl' foldFn initSt (take steps xs)) $
+    throwError $
+      ("val must be foldl over processed xs", res)
+  return res
+
+type IterResult b = (b, StepCount, DoneType)
+
+-- | If 'endReached' holds on the initial state, 'iterateUntilOrTimeout' returns
+-- immediately with zero steps and 'EndReached', regardless of the budget.
+propIterImmediateEndOnInit ::
+  (MonadMonotonicTime m, Eq a) =>
+  DiffTime ->
+  (a -> m a) ->
+  a ->
+  ExceptT (String, IterResult a) m (IterResult a)
+propIterImmediateEndOnInit ttl stepAction initSt = do
+  res@(val, steps, doneType) <- lift $ iterateUntilOrTimeout ttl (const True) stepAction initSt
+  unless (steps == 0) $ throwError $ ("steps must be zero", res)
+  unless (doneType == EndReached) $ throwError $ ("done must be end reached", res)
+  unless (val == initSt) $ throwError $ ("val must be initial st", res)
+  return res
+
+-- | The total elapsed time never exceeds @ttl + stepDuration@: the timeout is
+-- detected at the top of the loop, so at most one in-flight step can overshoot
+-- before the budget breach is observed.
+propIterTimeoutOverrunBoundedByOneStep ::
   MonadDelay m =>
-  -- | Wall-clock budget
   DiffTime ->
-  -- | Duration of one step of work
   DiffTime ->
-  m Bool
-propTimeoutOverrunBoundedByOneStep ttl stepActionDelay = do
-  start <- getMonotonicTime
-  (_, steps, doneType) <-
-    iterateUntilOrTimeout ttl (const False) (const $ threadDelay stepActionDelay) ()
-  end <- getMonotonicTime
+  ExceptT (String, (DiffTime, IterResult ())) m (IterResult ())
+propIterTimeoutOverrunBoundedByOneStep ttl stepActionDelay = do
+  start <- lift getMonotonicTime
+  res@(_, steps, doneType) <-
+    lift $
+      iterateUntilOrTimeout
+        ttl
+        (const False)
+        (const $ threadDelay stepActionDelay)
+        ()
+  end <- lift getMonotonicTime
   let duration = end `diffTime` start
 
-  return $
-    doneType == TimeoutReached
-      && duration >= max ttl (fromIntegral steps * stepActionDelay)
-      && duration < ttl + stepActionDelay
+  unless (doneType == TimeoutReached) $
+    throwError $
+      ("done type must be due to timeout", (duration, res))
+  unless (duration >= max ttl (fromIntegral steps * stepActionDelay)) $
+    throwError $
+      ( "duration must be equal or longer than max between ttl and steps * stepActionDelay"
+      , (duration, res)
+      )
+  unless (duration <= ttl + stepActionDelay) $
+    throwError $
+      ("duration must be equal or shorter than ttl + stepActionDelay", (duration, res))
+
+  return res
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
@@ -1,0 +1,114 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DerivingStrategies #-}
+
+-- | Time-bounded iteration.
+module LeiosUtils.TimeBoundedLoop
+  ( iterateUntilOrTimeout
+  , DoneType (..)
+  , StepCount
+  , iterateUntilOrTimeout'
+  , propImmediateEndOnInit
+  , propTimeoutOverrunBoundedByOneStep
+  ) where
+
+import Control.Monad.Class.MonadTime.SI (MonadMonotonicTime (getMonotonicTime), diffTime)
+import Control.Monad.Class.MonadTimer.SI (DiffTime, MonadDelay (..))
+
+-- | Why 'iterateUntilOrTimeout' stopped.
+data DoneType
+  = -- | The time budget was exhausted before the end of computation was reached.
+    TimeoutReached
+  | -- | The computation reached an end before the time budget was exhausted.
+    EndReached
+  deriving stock (Show, Eq)
+
+-- | Number of 'stepAction' invocations performed.
+type StepCount = Int
+
+-- | @iterateUntilOrTimeout ttl endReached stepAction initSt@ runs
+-- @stepAction@ on a state repeatedly until either @endReached@ holds or the
+-- wall-clock budget @ttl@ is exceeded.
+--
+-- The check order per iteration is:
+--
+--   1. End check (@endReached@).
+--   2. Timeout check (against @ttl@).
+--   3. Step (@stepAction@).
+--
+-- In particular, if @endReached initSt@ is already @True@ on entry, the
+-- function returns immediately with 'EndReached' and @'StepCount' = 0@.
+--
+-- __Single-threaded__: everything runs in one thread.  If @stepAction@ blocks
+-- longer than @ttl@, the timeout is detected only on the /next/ iteration —
+-- the in-flight call is not cancelled.  The total elapsed time can therefore
+-- exceed @ttl@ by at most one step duration.
+iterateUntilOrTimeout ::
+  MonadMonotonicTime m =>
+  -- | Wall-clock budget
+  DiffTime ->
+  -- | End predicate; when @True@, stops with 'EndReached'
+  (a -> Bool) ->
+  -- | One step of work
+  (a -> m a) ->
+  -- | Initial state
+  a ->
+  m (a, StepCount, DoneType)
+iterateUntilOrTimeout ttl endReached stepAction initSt = do
+  startTime <- getMonotonicTime
+  go 0 startTime initSt
+ where
+  go !n startTime !st = do
+    if endReached st
+      then pure (st, n, EndReached)
+      else do
+        now <- getMonotonicTime
+        if now `diffTime` startTime >= ttl
+          then pure (st, n, TimeoutReached)
+          else do
+            st' <- stepAction st
+            go (n + 1) startTime st'
+
+iterateUntilOrTimeout' ::
+  MonadMonotonicTime m =>
+  -- | Wall-clock budget
+  DiffTime ->
+  -- | End predicate; when @True@, stops with 'EndReached'
+  (a -> Bool) ->
+  -- | One step of work,
+  (a -> m a) ->
+  -- | Initial state
+  a ->
+  m a
+iterateUntilOrTimeout' ttl endReached stepAction st = (\(res, _, _) -> res) <$> iterateUntilOrTimeout ttl endReached stepAction st
+
+propImmediateEndOnInit ::
+  MonadMonotonicTime m =>
+  -- | Wall-clock budget
+  DiffTime ->
+  -- | One step of work
+  (a -> m a) ->
+  -- | Initial state
+  a ->
+  m Bool
+propImmediateEndOnInit ttl stepAction st = do
+  (_, steps, doneType) <- iterateUntilOrTimeout ttl (const True) stepAction st
+  return $ steps == 0 && doneType == EndReached
+
+propTimeoutOverrunBoundedByOneStep ::
+  MonadDelay m =>
+  -- | Wall-clock budget
+  DiffTime ->
+  -- | Duration of one step of work
+  DiffTime ->
+  m Bool
+propTimeoutOverrunBoundedByOneStep ttl stepActionDelay = do
+  start <- getMonotonicTime
+  (_, steps, doneType) <-
+    iterateUntilOrTimeout ttl (const False) (const $ threadDelay stepActionDelay) ()
+  end <- getMonotonicTime
+  let duration = end `diffTime` start
+
+  return $
+    doneType == TimeoutReached
+      && duration >= max ttl (fromIntegral steps * stepActionDelay)
+      && duration < ttl + stepActionDelay
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
@@ -29,7 +30,6 @@ module Ouroboros.Consensus.Mempool.Impl.Common
 
     -- * Validation
   , RevalidateTxsResult (..)
-  , computeSnapshot
   , revalidateTxsFor
   , revalidateTxsFor'
   , validateNewTransaction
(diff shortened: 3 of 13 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+    -- The `applyMempoolDiffs` uses the base LedgerState that contains diffs (I suspect from ticking? What values are affected by ticking?)
+    -- and builds up a LedgerState with values for this step only.
+    -- The `tickedStepBeforeStep` starts empty so it begins with only the values in `tickedStAtBase`, after that because of the union the `tickedStepBeforeStep` entries override whatever is conflicting in `tickedStAtBase`.
+    !tickedStAtBase = applyMempoolDiffs inputValuesStepForKeys inputKeysStep tickedStDiffBase
+    !inputValuesStep =
+      ltliftA2
+        unionValues
+        (projectLedgerTables tickedStBeforeStep)
+        (projectLedgerTables tickedStAtBase)
+    !tickedStBeforeStepWithValues = tickedStBeforeStep `withLedgerTables` inputValuesStep
```
bladyjoker: Please take a look at this for the sake of correctness

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,133 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+    iterateUntilOrTimeout'
+      0.1
+      (\(_, txsToApply, _, _, _) -> null txsToApply)
+      (snapshotStep resolveValues cfg slot tickedStDiff)
+      (tickedSt, txTickets, [], TxSeq.Empty, Set.empty)
+
+  let tip = castPoint $ getTip tickedStDiff
+  return $ snapshot slot tip txIds applied
+
+type SnapshotStepState blk =
```
geo2a: Nit: why not a dedicated record type?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,133 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+snapshotStep ::
+  forall blk m.
+  (LedgerSupportsMempool blk, HasTxId (GenTx blk), IOLike m) =>
+  (LedgerTables (LedgerState blk) KeysMK -> m (LedgerTables (LedgerState blk) ValuesMK)) ->
+  LedgerConfig blk ->
+  SlotNo ->
+  TickedLedgerState blk DiffMK ->
+  SnapshotStepState blk ->
+  m (SnapshotStepState blk)
+snapshotStep resolveValues cfg slot tickedStDiffBase (!tickedSt, !txsToApply, !unapplicable, !applied, !appliedTxIds) = do
```
geo2a: Nit: you can avoid bang patterns on the tuple fields if you refactor the tuple into a data type with strict fields.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,133 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  let (!tickets, !txsToApply') = TxSeq.take 100 txsToApply
```
geo2a: Consider refactoring `100` into a let binding so that the magic number is named. Perhaps we want it to be an argument of `snapshotStep`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,133 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  _ <- evaluate $ projectLedgerTables tickedSt''
+  return
+    ( tickedSt''
+    , txsToApply'
+    , unapplicable <> reverse unapplicable'
+    , applied'
+    , appliedTxIds `Set.union` appliedTxIds'
+    )
+
+type ReApplyState blk =
```
geo2a: Again, I'd prefer a dedicated data type with strict fields here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+        (projectLedgerTables tickedStBeforeStep)
+        (projectLedgerTables tickedStAtBase)
+    !tickedStBeforeStepWithValues = tickedStBeforeStep `withLedgerTables` inputValuesStep
+
+    (!tickedStAfterStep, !unapplicableTxsStep, !appliedTxsStep) = reapplyTxs' cfg slot txsToApplyStep tickedStBeforeStepWithValues
+
+  return
+    ( tickedStAfterStep
+    , txsToApplyRest
+    , unapplicableTxs <> reverse unapplicableTxsStep
```
geo2a: If we want to be super paranoid about laziness, I suggest we do two things:
- refactor `SnapshotStepState` into a record type with strict fields. This will worse `<>`;
- perhaps it's worth factoring out `reverse unapplicableTxsStep` into a let-binding with a bang patter (or a call to `evaluate`) so that we're sure the `reverse` is forced.

Same applies to the additions to the other tuple fields.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
geo2a: This looks good to me, but other people should have a look as well.

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
```diff
@@ -0,0 +1,114 @@
...
+--   3. Step (@stepAction@).
+--
+-- In particular, if @endReached initSt@ is already @True@ on entry, the
+-- function returns immediately with 'EndReached' and @'StepCount' = 0@.
+--
+-- __Single-threaded__: everything runs in one thread.  If @stepAction@ blocks
+-- longer than @ttl@, the timeout is detected only on the /next/ iteration —
+-- the in-flight call is not cancelled.  The total elapsed time can therefore
+-- exceed @ttl@ by at most one step duration.
+iterateUntilOrTimeout ::
```
ch1bo, line 58: Could: call this a fold

```suggestion
foldUntilOrTimeout ::
```

Signature looks a lot like `foldM`, so that's why :)

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
```diff
@@ -0,0 +1,114 @@
...
+{-# LANGUAGE DerivingStrategies #-}
+
+-- | Time-bounded iteration.
+module LeiosUtils.TimeBoundedLoop
+  ( iterateUntilOrTimeout
+  , DoneType (..)
+  , StepCount
+  , iterateUntilOrTimeout'
+  , propImmediateEndOnInit
+  , propTimeoutOverrunBoundedByOneStep
```
ch1bo: I like the idea of keeping the props next to the code! However, I do really want to have good `counterexample` or `cover`age .. which would result in quickcheck deps of the production code?

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
```diff
@@ -0,0 +1,114 @@
...
+      then pure (st, n, EndReached)
+      else do
+        now <- getMonotonicTime
+        if now `diffTime` startTime >= ttl
+          then pure (st, n, TimeoutReached)
+          else do
+            st' <- stepAction st
+            go (n + 1) startTime st'
+
+iterateUntilOrTimeout' ::
```
ch1bo: Could: call the variant with less arguments with a `_` suffix, or
Could: call the variant with more arguments wiht a `'` suffix

This is more in line with commonly used libraries. That is, `iterateUntilOrTimeout'` would be the advanced version of `iterateUntilOrTimeout`, where a `iterateUntilOrTimeout_` would be a variant of `iterateUntilOrTimeout` with no/less return values.

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
```diff
@@ -0,0 +1,114 @@
...
+  MonadMonotonicTime m =>
+  -- | Wall-clock budget
+  DiffTime ->
+  -- | End predicate; when @True@, stops with 'EndReached'
+  (a -> Bool) ->
+  -- | One step of work
+  (a -> m a) ->
+  -- | Initial state
+  a ->
+  m (a, StepCount, DoneType)
```
ch1bo, line 68: Should: not need the `StepCount`

It's only used in tests, right? If you want to keep props exactly this way, then I would suggest making the variant with step count on the result an internal function and not export it. That way we make it easy to test and only have the discarding of results not tested.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Oh nice! Let me do that!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Lazy! On it!

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
ch1bo, line 1: Which of these bangs is needed / should have been placed originally? Can you explain the rationale so we (or I) can do it correctly right away?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Impl/Common.hs
```diff
@@ -454,113 +460,47 @@ data RevalidateTxsResult blk
...
-  fst3 (x, _, _) = x
-  wrap = (\(TxTicket (ValidatedTxWithDiffs tx df) tk tz) -> (tx, (), (df, tk, tz)))
-  unwrap = (\(tx, (), (df, tk, tz)) -> (TxTicket (ValidatedTxWithDiffs tx df) tk tz))
-
 {-------------------------------------------------------------------------------
   Conversions
 -------------------------------------------------------------------------------}
 
+-- | Create a Mempool Snapshot from essential precomputed constituents.
+snapshot ::
```
ch1bo, line 578:
```suggestion
mkMempoolSnapshot ::
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+        (isTxs is)
  where
   MempoolEnv
     { mpEnvStateVar = istate
     , mpEnvLedgerCfg = cfg
-    , mpEnvCapacityOverride = capacityOverride
     } = mpEnv
+
+snapshotStepTimeLimitSeconds :: DiffTime
+snapshotStepTimeLimitSeconds = 0.1
```
ch1bo: Could: drop the `Seconds` suffix. `DiffTime` does store in picoseconds precision, but it's numeric instances interpret it as seconds as per its documentation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+
+  (_, _, _, !appliedTxs, !appliedTxIds) <-
+    iterateUntilOrTimeout'
+      snapshotStepTimeLimitSeconds
+      (\(_, txsToApply', _, _, _) -> null txsToApply')
+      (snapshotStep resolveValues cfg slot tickedStDiff)
+      (tickedSt, txsToApply, [], TxSeq.Empty, Set.empty)
+
+  let !tip = castPoint $ getTip tickedStDiff
+  return $ snapshot slot tip appliedTxIds appliedTxs
```
ch1bo: Why are all of these bangs needed? Why not bang the final `$!` too?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  TickedLedgerState blk DiffMK ->
+  TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk) ->
+  m (MempoolSnapshot blk)
+computeSnapshot resolveValues cfg slot tickedStDiff txsToApply = do
+  let !tickedSt = tickedStDiff `withLedgerTables` emptyLedgerTables
+
+  (_, _, _, !appliedTxs, !appliedTxIds) <-
+    iterateUntilOrTimeout'
+      snapshotStepTimeLimitSeconds
+      (\(_, txsToApply', _, _, _) -> null txsToApply')
```
ch1bo: Could: bind this in a where to `done (_, txs, _, _, _) = null txs`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
ch1bo: IMO This is a must change. It's too easy to mix up the second and fourth element of the tuple.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  forall blk m.
+  (LedgerSupportsMempool blk, HasTxId (GenTx blk), IOLike m) =>
+  (LedgerTables (LedgerState blk) KeysMK -> m (LedgerTables (LedgerState blk) ValuesMK)) ->
+  LedgerConfig blk ->
+  SlotNo ->
+  TickedLedgerState blk DiffMK ->
+  SnapshotStepState blk ->
+  m (SnapshotStepState blk)
+snapshotStep resolveValues cfg slot tickedStDiffBase (!tickedStBeforeStep, !txsToApply, !unapplicableTxs, !appliedTxs, !appliedTxIds) = do
+  let (!txsToApplyStep, !txsToApplyRest) = TxSeq.take snapshotStepTxsPerStep txsToApply
```
ch1bo: Should: just fold one by one? What's the drawback of going one by one?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/TxSeq.hs
```diff
@@ -272,3 +275,19 @@ toSize :: Measure sz => TxSeq sz tx -> sz
...
+take ::
+  Measure sz =>
+  Int ->
+  TxSeq sz tx ->
+  ([TxTicket sz tx], TxSeq sz tx)
+take = go []
+ where
+  go acc 0 rest = (reverse acc, rest)
+  go acc _ Empty = (reverse acc, Empty)
+  go acc n (t :< rest) = go (t : acc) (n - 1) rest
```
ch1bo: Should: call this `takeL`

A bit surprised that there is no `takeL` on the `FingerTree`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/TxSeq.hs
```diff
@@ -272,3 +275,19 @@ toSize :: Measure sz => TxSeq sz tx -> sz
...
+
+append :: Measure sz => TxSeq sz tx -> TxSeq sz tx -> TxSeq sz tx
+append (TxSeq l) (TxSeq r) = TxSeq (l FingerTree.>< r)
```
ch1bo: Should: use `mappend` or `<>` that is already implemented

```suggestion
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+          if canUseCache
+            -- We are looking for a snapshot at the same state ticked
+            -- to a different slot, so we can reuse the cached values
+            then mkResolveValues $ Left (isTxValues is)
+            -- We are looking for a snapshot at a different state, so we
+            -- need to read the values from the ledgerdb.
+            else
+              mkResolveValues $
+                Right
+                  readUntickedTables
```
ch1bo: Should: inline `mkResolveValues`

`mkResolveValues` is a pointless indirection. It's signature is double its implementation and the documentation here actually fits the body of it.

```suggestion
      let resolveValues = \keys ->
          if canUseCache
            -- We are looking for a snapshot at the same state ticked
            -- to a different slot, so we can reuse the cached values
            then pure $ restrictValues' (isTxValues is) keys
            -- We are looking for a snapshot at a different state, so we
            -- need to read the values from the ledgerdb.
            else readUntickedTables keys
```

ch1bo: That being said, `restrictValues'` is new. I'm no UTxO-HD expert, but wouldn't that kill the cache after the first step of the loop? i.e. we throw out anything we have loaded right now and not spend in the first step (which should actually be all inputs in the mempool?)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  m (SnapshotStepState blk)
+snapshotStep resolveValues cfg slot tickedStDiffBase (!tickedStBeforeStep, !txsToApply, !unapplicableTxs, !appliedTxs, !appliedTxIds) = do
+  let (!txsToApplyStep, !txsToApplyRest) = TxSeq.take snapshotStepTxsPerStep txsToApply
+      !inputKeysStep =
+        Foldable.foldMap'
+          (getTransactionKeySets . txForgetValidated . validatedTx . TxSeq.txTicketTx)
+          txsToApplyStep
+
+  -- TODO(bladyjoker): We're fetching inputsKeys many times, can we introduce a cache? Additionally, we don't actually need to get values,
+  -- we just need to check whether the keys are still there? Should be much cheaper.
```
ch1bo: 💯 See also my comment about `restrictValues'`

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
```diff
@@ -94,26 +101,150 @@ getSnapshotUsingPolicyFor policy mpEnv slot ticked readUntickedTables = do
...
+  -- Perhaps, there're other types in Cardano?
+  -- Sounds like working with this distinction could be an optimization point!
+  !inputValuesStepForKeys <- resolveValues inputKeysStep
+
+  let
+    -- TODO(bladyjoker): Please review. The idea is to construct the LedgerState with values that are necessary for applying the transactions in this step.
+    -- The `applyMempoolDiffs` uses the base LedgerState that contains diffs (I suspect from ticking? What values are affected by ticking?)
+    -- and builds up a LedgerState with values for this step only.
+    -- The `tickedStepBeforeStep` starts empty so it begins with only the values in `tickedStAtBase`, after that because of the union the `tickedStepBeforeStep` entries override whatever is conflicting in `tickedStAtBase`.
+    !tickedStAtBase = applyMempoolDiffs inputValuesStepForKeys inputKeysStep tickedStDiffBase
```
ch1bo: I find it very confusing that applying diffs (of the previous step!?) is done only here now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
ch1bo: I find these 5 lines very confusing and can't be sure whether they are correct. Why didn't we just do the same thing we did before split into steps?

Before we did something like:
1. get cached `isTxKeys` and either
  a. use cached `isTxValues` as `values`, or
  b. load `values` using the keys
1. `keys = foldMap getTransactionKeySets ..` (again?) from `inputTxs`
1. `st0 = applyMempoolDiffs` with the `values` and `keys`
1. `reapplyTxs inputTxs st0` and check if fine, throw away resulting state

We should do now:
1. get cached `isTxKeys` and either
  a. use cached `isTxValues` as `values`, or
  b. load `values` using the keys
1. `st0 = applyMempoolDiffs` with the `values` and `keys`
1. fold until timeout over all `inputTxs` with `tx` and `st = st0`
    1. `res = reapplyTx tx st`
    2. stop on `Left err`, continue with `Right st'`

Notably, the signature of `reapplyTx` being:

```haskell
  reapplyTx ::
    HasCallStack =>
    LedgerConfig blk ->
    -- | Slot number of the block containing the tx
    SlotNo ->
    Validated (GenTx blk) ->
    -- | Contains at least the values for the tx to reapply
    TickedLedgerState blk ValuesMK ->
    Except (ApplyTxErr blk) (TickedLedgerState blk ValuesMK)
```

ch1bo (changes requested):
I suspect the way we interact with UTxO-HD is not semantically equivalent than before, but I might be wrong. See my comment on what I believe the time capped loop should do. Also, I think we are actively throwing away our cached values.

Left several suggestions on how to improve readability and there are so many bangs .. are all of them needed? What's the policy when to put them? Should we put `StrictData` or `Strict` pragmas instead?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Turns out making reapplyTxs' like this

```haskell
reapplyTxs' ::
  (LedgerSupportsMempool blk, HasTxId (GenTx blk)) =>
  LedgerConfig blk ->
  SlotNo ->
  TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk) ->
  SnapshotStepState blk ->
  SnapshotStepState blk
```

results in more readable and probably more efficient impl

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Done

```haskell
snapshotStepTimeLimitSeconds :: DiffTime
snapshotStepTimeLimitSeconds = 0.1

snapshotStepTxsPerStep :: Int
snapshotStepTxsPerStep = 100
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Done

```haskell
-- | Accumulator type threaded through each 'snapshotStep' call by 'computeSnapshot'.
data SnapshotStepState blk = SnapshotStepState
  { currentLedgerSt :: !(TickedLedgerState blk ValuesMK)
  -- ^ Ticked ledger state reflecting the cumulative effect of all applied
  -- transactions so far ('appliedTxs').
  , remainingTxs :: !(TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk))
  -- ^ Remaining transactions to be applied. Starts as the full
  -- mempool sequence and is consumed 'snapshotStepTxsPerStep' at a time.
  -- Iteration terminates when this becomes empty.
  , unapplicableTxs :: !(StrictSeq (Invalidated blk))
  -- ^ Transactions that failed reapplication, accumulated in original mempool
  -- order.
  , appliedTxs :: !(TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk))
  -- ^ Transactions successfully applied so far to some base Ledger state that results in 'currentLedgerSt', in original mempool order.
  -- Becomes the transaction sequence of the final 'MempoolSnapshot'.
  , appliedTxIds :: !(Set.Set (GenTxId blk))
  -- ^ Set of all accepted transactions (view on 'appliedTxs').
  }
```

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
bladyjoker: It's not a fold, it's iterate! Linked the naming rationale!

Added the implementation of `foldOrTimeout` that is implemented using `iterateUntilOrTimeout`. Used it to test things.

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
bladyjoker: That's indeed a good point, it's very useful to have.

Replaced the naive return value of properties into `ExceptT (String, err) m a`.  For example

```haskell
propFoldEmptyListTerminates ::
  (MonadMonotonicTime m, Eq b) =>
  DiffTime ->
  (b -> a -> m b) ->
  b ->
  ExceptT (String, FoldResult b a) m (FoldResult b a)
propFoldEmptyListTerminates ttl foldAction initSt = do
  res@(val, remaining, steps, done) <- lift $ foldOrTimeout ttl foldAction initSt []
  unless (null remaining) $ throwError $ ("remaining must be empty", res)
  unless (steps == 0) $ throwError $ ("steps must be zero", res)
  unless (done == EndReached) $ throwError $ ("done must be EndReached", res)
  unless (val == initSt) $ throwError $ ("val must be initSt", res)
  return res
```

In this instance `err` and `a` are the same, but not necessary. `res` is returned in `Right` because we also want to compose additional properties based on `res`.

ouroboros-consensus/src/ouroboros-consensus/LeiosUtils/TimeBoundedLoop.hs
bladyjoker: It's deffo used in tests but I argue it's a natural part of the API and will definitely end up in the trace. After all, it can be very important to know whether we reached timeout or not and how many steps we managed to do.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
bladyjoker: I'll try to provide the rationale and reduce it to minimum, it takes a lot of time to redo the experiments on each change.

The problem with that is that I wouldn't trust anyone to be able to provide a rationale for why this works an empirical proof on it. 
Every change in strictness must be followed by a close look at the CallTrace to make sure this is indeed having a desired effect.

The code is too complex to effectively and confidently determine where thunks are forced just by reading the code. The best we have are hunches.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/TxSeq.hs
bladyjoker: Aha didn't know, also, refactored the `reapplyTxs'` and `snapshotStep` bits and remove the `append` as I don't need it any longer.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: It's a function that given some `keys` gets the `values` associated with them. 

```haskell
resolveValues :: LedgerTables (LedgerState blk) KeysMK -> m (LedgerTables (LedgerState blk) ValuesMK)
resolveValues <-
        return $
          if canUseCache
            -- We are looking for a snapshot at the same state ticked
            -- to a different slot, so we can reuse the cached values
            then pure . restrictValues' (isTxValues is)
            -- We are looking for a snapshot at a different state, so we
            -- need to read the values from the ledgerdb.
            else readUntickedTables

```

In the case of a "cache hit" we can use the values already in the Mempool (`isTxValues is`), otherwise we need to go to the disk. `restrictValues'` is a PURE function that's in spirit the same as https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Map-Lazy.html#v:restrictKeys.

For the rest it's important to follow the `snapshotStep` procedure. 

The whole point is to only read values that are needed at a step (and no more). Previously we just read ALL the values for ALL the transactions in the Mempool, now the idea is to read in smaller batches.

```haskell
snapshotStep resolveValues cfg slot baseLedgerStDiff st@SnapshotStepState{..} = do
  let (!txsToApplyStep, !remainingTxsStep) = TxSeq.take snapshotStepTxsPerStep remainingTxs
      !inputKeysStep =
        Foldable.foldMap'
          (getTransactionKeySets . txForgetValidated . validatedTx)
          txsToApplyStep

  !inputValuesStepForKeys <- resolveValues inputKeysStep

  let
    !baseLedgerSt = applyMempoolDiffs inputValuesStepForKeys inputKeysStep baseLedgerStDiff
    !inputValuesStep =
      ltliftA2
        unionValues
        (projectLedgerTables currentLedgerSt)
        (projectLedgerTables baseLedgerSt)
    !stForStep =
      st
        { currentLedgerSt = currentLedgerSt `withLedgerTables` inputValuesStep
        , remainingTxs = remainingTxsStep
        }

  return $! reapplyTxs' cfg slot txsToApplyStep stForStep
```

Several tricky parts

1.  `baseLedgerStDiff` and `baseLedgerSt` and `applyMempoolDiff`
    - Talked to Javier and the only time this diff after ticking is non-empty is in very peculiar situation (byron -> shelley and some other avvm stuff I don't really understand). So it's basically empty all the time (so to speak `baseLedgerSt == withLedgerTables baseLedgerSt emptyLedgerTables`)
    - The ticking diffs are "at base" not "at previous step". This is about applying these diffs ONLY to the keys at this step (before it was on ALL keys of all Mempool txs).
    

I'm adding some tests because this is too much of a hearsay for my comfort.

ch1bo: Okay, thanks for explaining. I have been reading the types all wrong I guess. `LedgerTables (LedgerState blk) KeysMK` is just a bunch of `Keys`, I always read it as a handle to the UTxO-HD ledger tables with some keys already prefilled or so. In any case, the return value of `resolveValues` is not kept through the loop, so its definitely fine.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: See the new `Ledger` module, I rewrote this 20 times, found it difficult to make it obviously (in)correct. I think this is good now, but I will fire tests at it before going to 'main' with it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Query.hs
bladyjoker: Decided for a different, imo cleaner strategy.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+reapplyStep ::
+  forall blk m.
+  (LedgerSupportsMempool blk, HasTxId (GenTx blk), IOLike m) =>
+  Int ->
+  (LedgerTables (LedgerState blk) KeysMK -> m (LedgerTables (LedgerState blk) ValuesMK)) ->
+  LedgerConfig blk ->
+  SlotNo ->
+  ReapplyStepState blk ->
+  m (ReapplyStepState blk)
+reapplyStep reapplyPerStep resolveValues cfg slot st@ReapplyStepState{..} = do
```
ch1bo, line 128: Should: not introduce new reapply/step functions, but re-use / generalize existing ones

My recent work on the mempool sync split the "all-or-nothing" reapplyTxsFor function in `Mempool.Common` to have a step-wise alternative. I would suggest that we re-use or modify that step-wise function instead of creating a second one here.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+import Ouroboros.Consensus.Util.IOLike
+
+-- | Accumulator type threaded through each 'reapplyStep' call by top-level 'reapply' and 'reapplyUntilTimeout'.
+data ReapplyStepState blk = ReapplyStepState
+  { currentLedgerSt :: !(TickedLedgerState blk ValuesMK)
+  -- ^ Ticked ledger state reflecting the cumulative effect of all applied transactions so far ('appliedTxs').
+  , remainingTxs :: !(TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk))
+  -- ^ Remaining transactions to be applied (__applicability condition holds__)
+  , unapplicableTxs :: !(StrictSeq (Invalidated blk))
+  -- ^ Transactions that failed reapplication, accumulated in original order of 'remainingTxs'.
```
ch1bo, line 57: Should: not need the `unapplicableTxs`

The use case we look at is to bound the time we spend in the forge loop. We give an upper bound (with all this machinery), but are perfectly fine if we stop earlier. We should stop on the first unapplicable tx and get the block with what is applicable out asap.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+  , appliedTxs :: !(TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk))
+  -- ^ Transactions successfully applied to produce 'currentLedgerSt'.
+  , appliedTxIds :: !(Set.Set (GenTxId blk))
+  -- ^ Set of all accepted transactions (view on 'appliedTxs').
```
ch1bo, line 61: Why do we keep this on the data type and not just project it when needed?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+  , phantomUTxOs :: !(Set (TxIn (LedgerState blk)))
+  -- ^ UTxOs created by unapplicable transactions in 'unapplicableTxs'.
```
ch1bo, line 63: Following my comment above, we should not be needing these?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+  return $! reapplyTxs cfg slot txsToApplyStep stForStep
+
+reapplyTxs ::
+  (LedgerSupportsMempool blk, HasTxId (GenTx blk)) =>
+  LedgerConfig blk ->
+  SlotNo ->
+  TxSeq.TxSeq (TxMeasureWithDiffTime blk) (ValidatedTxWithDiffs blk) ->
+  ReapplyStepState blk ->
+  ReapplyStepState blk
+reapplyTxs cfg slot txs stBefore =
```
ch1bo, line 163: Should: use the existing `reapplyTxs` function here

There exists already one called exactly like this. Besides being confusing, we should just use that function and just wire it up with whatever state we keep in between (or should use instead `ReapplyTxsResult`?)

ch1bo (approved):
Looks correct, but did not expect this complexity being needed.

See my comment about early exiting the loop, which could make some book-keeping redundant.

ch1bo:
Announcements histogram looks nice after this, seems like the forge loop is not blocked as often anymore (on this `dozen-devnet` run at the least):

image

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
bladyjoker: I agree, I will make all 'reappliaction' in the code use this alternative as I find it the most general implementation that also is IO stingy.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
bladyjoker: Not sure I understand, `unappalicableTxs` contain all the transactions that didn't succeed "re-application". We can't stop when we encounter one, that means we could stop after the first one and that's a problem.

We also have `phantomOutputs` to poison all transactions consuming outputs of unapplicable txs as they are predictably failing so we don't waste IO on them.

It's important to realize this implementation is making sure we don't do "unneccessary" work, part of that is IO.

Other than that, I do see now it's not really used for any key accounting, however, will clean it up once I wire this up with good tests.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
bladyjoker: Because I don't want to do several O(n) passes to build various indices that are needed, when I can do a single pass. As I need it to construct a snapshot I include it, not strictly needed tho, I think it's worth removing for the sake of readability.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
bladyjoker: Probably didn't notice, but I'm not doing any IO work for transactions that failed re-application, and phantom outputs are collecting outputs of said transactions so the next one that comes in that consumes these transactions are predictably failing.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
bladyjoker: This variant is specially designed for optimizations claimed in this module, what makes sense is for all reapplication to use this module as it wasn't that easy to do it properly.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Ledger.hs
```diff
@@ -0,0 +1,263 @@
...
+    let tx = TxSeq.txTicketTx tkt
+        vtx = validatedTx tx
+        txCreated = createdFromTx tx
+     in case runExcept (reapplyTx cfg slot vtx (currentLedgerSt st)) of
+          Left !err ->
+            st
+              { unapplicableTxs = unapplicableTxs st |> Invalidated vtx err
+              , phantomUTxOs = phantomUTxOs st `Set.union` Map.keysSet txCreated -- propagate phantom outputs
+              }
+          Right !ledgerStAfterTx ->
```
nfrisby, line 179: This is the first comment I wrote. You might want to instead skip to the second comment I wrote.

`Right !ledgerStAfterTx` ensures that `reapplyTxsStep` function (and hence the `reapplyTxs` function, which is `$!` forced by the `reapplyStep` caller) is forcing as much of the reapplication calculation as possible to have finished before `reapplyTxsStep` returns (ie before `reapplyTxsStep` yields _anything_ to its caller).

All of the fields of the returns `ReapplyStepState` type are strict. (So `Right !ledgerStAfterTx` could harmlessly be `Right ledgerStAfterTx`.)

If there are still thunks _inside of_ `ledgerStAfterTx`, despite us reducing it to [whnf](https://wiki.haskell.org/index.php?title=Weak_head_normal_form), then that's a consequence (or intentional or unintentional) of the ledger implementation. The only way _this_ code could to any better would be to project out the problematic thunks (after we identify them) and also force _them_ before yielding the updated `st`. But ultimately, such details should be part of the Consensus-Ledger interface.

My operating assumption has been that reducing a `NewEpochState` to whnf only defers (costly) thunks for intentional reasons.

nfrisby: The above comment is maybe informative, but it left implicit the specific scope of this PR.

This PR's purpose is to time-cap the amount of work _the forge does_. So what matters here is simpler whether _this_ code forces all of the thunks _that the forge needs to be forced_. One key observation on that front is:

> The final `currentLedgerSt` at the end of the loop isn't used by the forge.

The forge only needs the final txs the loop yields: it might split a relatively small prefix to put into the RB it's forging and whatever the remaining suffix is after the RB's tx is put into the announced EB. No part of that work requires the final ledger state. (TODO please confirm this @bladyjoker ; you're more aware of the Leios forge than I am)

Thus: _it doesn't even matter_ if there are expensive thunks somewhere inside of the final `currentLedgerSt` (even despite it being whnf---which has usually meant all such expensive remaining thunks are intentional (eg because they're stable)).

It's ultimately only the `case` discrimination on the `Left`-versus-`Right` of the tx's (phase1) validity check that ensures all the computation the forge needs has already happened.

nfrisby: @bladyjoker mentioned in a call that he was worried about work leaking out of this time-capped loop (hence slowing down the forge) via accidental thunks. The above is my reasoning about the limitations of that risk. Maybe you want to distill it into a comment (just the highlight you find most useful) to include in this code?

nfrisby: Caveat to my wording above: this code codes `reapplyTx`, which I'm treating as a ledger call, but it isn't actually---but it _should_ just be a "thin wrapper" around a ledger call (which `Ouroboros.Consensus.Shelley.Ledger.Mempool.reapplyShelleyTx` is, for example)

nfrisby (commented):
Just addressing a concern Drazen had shared in a call.

PR #2220, opened 2026-08-18
[Peras 60] Introduce default StateSupportsPerasEpochContext instances

agustinmista:
This commit introduces the necessary `StateSupportsPerasEpochContext` instances for the different block types in the codebase.

NOTE: some of these instances currently use the default empty implementation until we get rid of the degenerate `BlockSupportsPeras` instance in the near future.

### DIFF
--- a/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
+++ b/ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
@@ -65,12 +65,15 @@ import Ouroboros.Consensus.BlockchainTime
 import Ouroboros.Consensus.Config
 import Ouroboros.Consensus.Config.SupportsNode
 import Ouroboros.Consensus.Forecast
+import Ouroboros.Consensus.HardFork.Abstract (HasHardForkHistory (..))
 import Ouroboros.Consensus.HardFork.Combinator
 import Ouroboros.Consensus.HardFork.Combinator.Condense
 import Ouroboros.Consensus.HardFork.Combinator.Serialisation.Common
 import Ouroboros.Consensus.HardFork.History
   ( Bound (..)
+  , EpochToPerasRoundInfo
   , EraParams (..)
+  , forgetEraIndex
   )
 import qualified Ouroboros.Consensus.HardFork.History as History
 import Ouroboros.Consensus.HeaderValidation
@@ -87,6 +90,7 @@ import Ouroboros.Consensus.Node.InitStorage
 import Ouroboros.Consensus.Node.NetworkProtocolVersion
 import Ouroboros.Consensus.Node.Run
 import Ouroboros.Consensus.Node.Serialisation
+import Ouroboros.Consensus.Peras.Context (StateSupportsPerasEpochContext (..))
 import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.Storage.ImmutableDB (simpleChunkInfo)
 import Ouroboros.Consensus.Storage.Serialisation
@@ -310,6 +314,19 @@ instance LedgerSupportsProtocol BlockA where
 instance LedgerStateSupportsPeras (LedgerState BlockA)
 
 instance LedgerStateSupportsPeras (Ticked LedgerState BlockA)
+
+-- NOTE: BlockA is only ever used wrapped in the
+-- hard fork combinator (which implements 'hardForkSummary' directly and never
+-- delegates to the underlying era), so this method is never actually called.
+instance HasHardForkHistory BlockA where
+  type HardForkIndices BlockA = '[BlockA]
+  hardForkSummary = error "BlockA being used as a SingleEraBlock"
+
+instance StateSupportsPerasEpochContext BlockA where
+  type MaybeEraIndexedEpochToPerasRoundInfo BlockA = EpochToPerasRoundInfo
+  toMaybeEraIndexedEpochToPerasRoundInfo _ = forgetEraIndex
+  fromMaybeEraIndexedEpochToPerasRoundInfo _ = id
+  mkBoundedPerasEpochContext = error "mkBoundedPerasEpochContext: BlockA does not support Peras"
 
 instance HasPartialConsensusConfig ProtocolA
 
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -397,6 +397,14 @@ class IsPerasError err blk | err -> blk where
   injectConversionError :: PerasConversionError -> err
   injectQuorumNotReachedError :: VoteWeight -> err
 
+instance IsPerasError (VoidPerasError blk) blk where
+  injectVotingCommitteeError _ =
+    error "injectVotingCommitteeError: VoidPerasError cannot be inhabited"
+  injectConversionError _ =
+    error "injectConversionError: VoidPerasError cannot be inhabited"
+  injectQuorumNotReachedError _ =
+    error "injectQuorumNotReachedError: VoidPerasError cannot be inhabited"
+
 -- * Types and functions related to Peras vote collection and quorum checking
 
 -- | Collection of Peras votes for a given target.
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
@@ -38,17 +38,17 @@ import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.Config.SupportsNode
 import Ouroboros.Consensus.HardFork.Combinator.Info
 import Ouroboros.Consensus.HardFork.Combinator.PartialConfig
-import Ouroboros.Consensus.HardFork.History (Bound, EraParams)
+import Ouroboros.Consensus.HardFork.History (Bound, EpochToPerasRoundInfo, EraParams)
 import Ouroboros.Consensus.Ledger.Abstract
 import Ouroboros.Consensus.Ledger.CommonProtocolParams
 import Ouroboros.Consensus.Ledger.Inspect
 import Ouroboros.Consensus.Ledger.Query
 import Ouroboros.Consensus.Ledger.SupportsMempool
 import Ouroboros.Consensus.Ledger.SupportsPeerSelection
-import Ouroboros.Consensus.Ledger.SupportsPeras (LedgerStateSupportsPeras)
 import Ouroboros.Consensus.Ledger.SupportsProtocol
 import Ouroboros.Consensus.Node.InitStorage
 import Ouroboros.Consensus.Node.Serialisation
+import Ouroboros.Consensus.Peras.Context (StateSupportsPerasEpochContext (..))
 import Ouroboros.Consensus.Protocol.Abstract
 import Ouroboros.Consensus.Storage.Serialisation
 import Ouroboros.Consensus.Ticked
@@ -77,14 +77,10 @@ class
   , ConfigSupportsNode blk
   , NodeInitStorage blk
   , BlockSupportsDiffusionPipelining blk
+  , StateSupportsPerasEpochContext blk
+  , MaybeEraIndexedEpochToPerasRoundInfo blk ~ EpochToPerasRoundInfo
   , BlockSupportsMetrics blk
   , SerialiseNodeToClient blk (PartialLedgerConfig blk)
-  , -- TODO: replace the four constraints below with:
-    -- 'StateSupportsPerasEpochContext blk' once that type class is in place.
-    ChainDepStateSupportsPeras (ChainDepState (BlockProtocol blk))
-  , ChainDepStateSupportsPeras (Ticked (ChainDepState (BlockProtocol blk)))
-  , LedgerStateSupportsPeras (LedgerState blk)
-  , LedgerStateSupportsPeras (Ticked LedgerState blk)
   , -- LedgerTables
     CanStowLedgerTables (LedgerState blk)
   , HasLedgerTables LedgerState blk
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
@@ -90,8 +90,11 @@ import Ouroboros.Consensus.HardFork.Combinator.State.Types
 import Ouroboros.Consensus.HardFork.Combinator.Translation
 import Ouroboros.Consensus.HardFork.History
   ( Bound (..)
+  , EpochToPerasRoundInfo
+  , EraIndexed
   , EraParams
   , SafeZone (..)
+  , forgetEraIndex
   )
 import qualified Ouroboros.Consensus.HardFork.History as History
 import Ouroboros.Consensus.HeaderValidation
@@ -100,6 +103,7 @@ import Ouroboros.Consensus.Ledger.Inspect
 import Ouroboros.Consensus.Ledger.SupportsPeras (LedgerStateSupportsPeras (..))
 import Ouroboros.Consensus.Ledger.SupportsProtocol
 import Ouroboros.Consensus.Ledger.Tables.Utils
+import Ouroboros.Consensus.Peras.Context (StateSupportsPerasEpochContext (..))
 import Ouroboros.Consensus.TypeFamilyWrappers
 import Ouroboros.Consensus.Util.Condense
 import Ouroboros.Consensus.Util.IndexedMemPack (IndexedMemPack)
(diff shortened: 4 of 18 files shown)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -164,20 +164,105 @@ instance
...
-  ( ShelleyCompatible proto era
-  , LedgerSupportsProtocol (ShelleyBlock proto era)
-  , TxLimits (ShelleyBlock proto era)
+  ( ShelleyCompatible proto ShelleyEra
+  , LedgerSupportsProtocol (ShelleyBlock proto ShelleyEra)
+  , TxLimits (ShelleyBlock proto ShelleyEra)
   , Crypto (ProtoCrypto proto)
   ) =>
-  SerialiseHFC '[ShelleyBlock proto era]
+  SerialiseHFC '[ShelleyBlock proto ShelleyEra]
```
jasagredo: Is this necessary? I think it could be reverted to `era`, no?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -164,20 +164,105 @@ instance
...
+  ) =>
+  SerialiseHFC '[ShelleyBlock proto DijkstraEra]
+
+instance
+  ( ShelleyCompatible proto ShelleyEra
+  , LedgerSupportsProtocol (ShelleyBlock proto ShelleyEra)
+  , TxLimits (ShelleyBlock proto ShelleyEra)
+  , Crypto (ProtoCrypto proto)
+  ) =>
+  SerialiseConstraintsHFC (ShelleyBlock proto ShelleyEra)
```
jasagredo: Same

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -311,6 +313,15 @@ instance LedgerStateSupportsPeras (LedgerState BlockA)
 
 instance LedgerStateSupportsPeras (Ticked LedgerState BlockA)
 
+-- NOTE: BlockA is only ever used wrapped in the
+-- hard fork combinator (which implements 'hardForkSummary' directly and never
+-- delegates to the underlying era), so this method is never actually called.
+instance HasHardForkHistory BlockA where
+  type HardForkIndices BlockA = '[BlockA]
+  hardForkSummary = error "BlockA being used as a SingleEraBlock"
```
jasagredo, line 323: And why is the instance necessary?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -397,6 +397,14 @@ class IsPerasError err blk | err -> blk where
   injectConversionError :: PerasConversionError -> err
   injectQuorumNotReachedError :: VoteWeight -> err
 
+instance IsPerasError (VoidPerasError blk) blk where
+  injectVotingCommitteeError _ =
+    error "injectVotingCommitteeError: VoidPerasError cannot be inhabited"
+  injectConversionError _ =
+    error "injectConversionError: VoidPerasError cannot be inhabited"
+  injectQuorumNotReachedError _ =
+    error "injectQuorumNotReachedError: VoidPerasError cannot be inhabited"
```
jasagredo, line 406: Can't we implement something like `absurd a = case a of {}` instead of calling error? or is VoidPerasError not really void?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Dual.hs
```diff
@@ -1206,3 +1209,14 @@ instance
 instance LedgerStateSupportsPeras (LedgerState (DualBlock m a))
 
 instance LedgerStateSupportsPeras (Ticked LedgerState (DualBlock m a))
+
+instance
+  ( Bridge m a
+  , StandardHash m
+  , Typeable m
+  , Typeable a
+  , All Top (HardForkIndices (DualBlock m a))
```
jasagredo: ? isn't this unsolvable? I would not expect DualBlock to have a HardForkIndices

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -340,6 +344,22 @@ instance
       . State.tip
       . tickedHardForkLedgerStatePerEra
 
+-- TODO: this instance will have a full implementation as soon as we remove the
+-- degenerate 'BlockSupportsPeras' instance.
+instance
+  ( All Top xs
```
jasagredo: I think `All Top xs => CanHardFork xs` no?

jasagredo: And possibly also for StandardHash

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -340,6 +344,22 @@ instance
...
+  , CanHardFork xs
+  ) =>
+  StateSupportsPerasEpochContext (HardForkBlock xs)
+  where
+  type
+    MaybeEraIndexedEpochToPerasRoundInfo (HardForkBlock xs) =
+      EraIndexed xs EpochToPerasRoundInfo
+
+  fromMaybeEraIndexedEpochToPerasRoundInfo _ = forgetEraIndex
+  toMaybeEraIndexedEpochToPerasRoundInfo _ = id
```
jasagredo, line 358: I really dislike the default methods because I had to go look up in the definition of the class what other methods are there (in particular `mkBoundedPerasEpochContext`). I think we should instead create `defaultFromMaybeEra...` and `defaultToMaybeEra...` and use them in the defaulting instances, or create a combinator and derive the instance via that newtype combinator, either of those will be more explicit than this

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
```diff
@@ -77,14 +77,10 @@ class
   , ConfigSupportsNode blk
   , NodeInitStorage blk
   , BlockSupportsDiffusionPipelining blk
+  , StateSupportsPerasEpochContext blk
+  , MaybeEraIndexedEpochToPerasRoundInfo blk ~ EpochToPerasRoundInfo
```
jasagredo, line 81: Do we _need_ this as a constraint of SingleEraBlock? isn't it enough to require it as a constraint in the Shelley instance? At some point we invoke this type equality in a place where we only have `SingleEraBlock x`?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
agustinmista: This was a balancing act between requiring `HasHardForkHistory` as a superclass constraint of `StateSupportsPerasEpochContext` or not.

One one hand, empty instances do not always strictly need `HasHardForkHistory`, with `BlockA` and `BlockB` being an example of that. So, these bogus instances are just and artifact of those block types needing a `StateSupportsPerasEpochContext` instance later on.

On the other hand, not requiring `HasHardForkHistory` as a superclass of `StateSupportsPerasEpochContext` allows us to get rid of these bogus instances at the cost of needing a bunch of extra constraints in the call sites that _do_ depend on `HasHardForkHistory`.

I tried a quick refactor, removing the superclass contraint from the class definition and adding it to the necessary call sites, and that incurred in 36 new additions:

```console
git diff | grep "+\s*, HasHardForkHistory" | wc -l
36
```

FTR, I don't mind either way, so I leave it up to you :)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
agustinmista: Hmm, I think I might have extracted an unrelated change from our integration branch, I will check.

FTR: in the end we _do_ need to split these instances because we will have different per-era instances for one of its superclasses (IIRC, it was `SerialiseNodeToNode` that needs to be different for `DijkstraEra`), so I think this is a matter of now vs. later rather than necessary vs. unnecessary.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: In this case we cannot use `absurd` because these are injections _into_ void (which notably has no inhabitants). However, none of these injections could ever be called for the block types that use the default void implementation because the error-prone functions should be discharged by contradiction via some of their inputs being `void` (e.g. a void vote, cert, etc.).

Similar to `HasHardForkHistory`, `IsPerasError` is also a superclass of `StateSupportsPerasEpochContext` (and will soon be a superclass of `BlockSupportsPeras`), the reason also being a balancing act between having bogus instances and _a bunch_ of extra constraints in the relevant call sites. 

If you want, we could remove this for now and add it later along with the refactor of `BlockSupportsPeras`, but regardless of the timing, I think we probably want to have it in the end.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
agustinmista: Indeed we _do_ need this equality inside the (upcoming) real implementation of `mkBoundedPerasEpochContext` for the HFC:

```haskell
instance
  ( All Top xs
  , All SingleEraBlockWithEpochToPerasRoundInfo xs
  , StandardHash (HardForkBlock xs)
  , CanHardFork xs
  ) =>
  StateSupportsPerasEpochContext (HardForkBlock xs)
  where
  type
    MaybeEraIndexedEpochToPerasRoundInfo (HardForkBlock xs) =
      EraIndexed xs EpochToPerasRoundInfo

  fromMaybeEraIndexedEpochToPerasRoundInfo _ = forgetEraIndex
  toMaybeEraIndexedEpochToPerasRoundInfo _ = id

  mkBoundedPerasEpochContext epochToPerasRoundInfo ledgerState headerState =
    bimap
      (HardForkPerasErrorOneEraPerasError . OneEraPerasError)
      injectHFCBoundedPerasEpochContext
      ( hdistribute
          . hcmap
            proxySingle -- <- this guy here needs to provide the equality
            ( \_ ->
                mkEitherF WrapPerasError id $
                  mkBoundedPerasEpochContext
                    ( fromMaybeEraIndexedEpochToPerasRoundInfo
                        (Proxy @(HardForkBlock xs))
                        epochToPerasRoundInfo
                    )
                    ledgerState
                    headerState
            )
          $ eraIndexedToNS epochToPerasRoundInfo
      )
```

We tried to avoid this in a couple of ways, e.g., by extending `proxySingle` with an ad-hoc type class that additionally provides the equality:

```haskell

class
  ( SingleEraBlock blk
  , MaybeEraIndexedEpochToPerasRoundInfo blk ~ EpochToPerasRoundInfo
  ) =>
  SingleEraBlockWithEpochToPerasRoundInfo blk

instance
  ( SingleEraBlock blk
  , MaybeEraIndexedEpochToPerasRoundInfo blk ~ EpochToPerasRoundInfo
  ) =>
  SingleEraBlockWithEpochToPerasRoundInfo blk

proxySingleWithEpochToPerasRoundInfo :: Proxy SingleEraBlockWithEpochToPerasRoundInfo
proxySingleWithEpochToPerasRoundInfo = Proxy
```

But in the end, GHC was pushing us to simplify every occurrence of this class with the equality itself (the mono-local-binds thingy), kind of defeating the purpose. 

In general, I think it makes sense to expect single era blocks to _not_ have an era indexed `MaybeEraIndexed ...`, as that's only a special case for the HFC which needs the index to be able to dispatch things.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
agustinmista: I'm a bit on the fence here. 

As for the `fromMaybe...`/`toMaybe...` methods, I really think we should try to hide what's ultimately an implementation detail for the HFC (who's instance is the only one that needs this extra era index to be able to dispatch the implementation to the right era) from the rest of the single-era- and test-block instances. 

On the other hand, I don't have a strong opinion about removing the default implementation of `mkBoundedPerasEpochContext` and providing a `defaultMkBoundedPerasEpochContext`, but that doesn't seem to be the same pattern used throughout the codebase, where empty/trivial instances are derived with one-liners.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
jasagredo: But you need a different instance of a superclass, this instance should still hold with `era` no?

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/HardFork/Combinator/A.hs
jasagredo: Let's keep it as is

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
jasagredo: I see, let's keep it as is

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
jasagredo: So we do have some defaults in the codebase, but we aim to not introduce more. The ones that are there we are sure they have to be correct, but on some new classes/instances we defined we ended up using the default implementation in places we didn't want to.

I think in general we don't want any more default methods and we prefer being explicit about the default being chosen.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/SingleEraBlock.hs
jasagredo: Aha I see, thanks

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Dual.hs
agustinmista: Good catch. This is gone now.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
agustinmista: I took another stab at this. 

Without changing anything, GHC complaints about [loopy superclasses](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/instances.html#undecidable-instances-and-loopy-superclasses):

image

But it seems like we can go around this if we provide the extra `SerialiseNodeToNodeConstraints` constraint explicitly. This now compiles:

```haskell
instance
  ( ShelleyCompatible proto era
  , LedgerSupportsProtocol (ShelleyBlock proto era)
  , TxLimits (ShelleyBlock proto era)
  , SerialiseNodeToNodeConstraints (ShelleyBlock proto era) -- NEW
  , Crypto (ProtoCrypto proto)
  ) =>
  SerialiseHFC '[ShelleyBlock proto era]

instance
  ( ShelleyCompatible proto era
  , LedgerSupportsProtocol (ShelleyBlock proto era)
  , TxLimits (ShelleyBlock proto era)
  , SerialiseNodeToNodeConstraints (ShelleyBlock proto era) -- NEW
  , Crypto (ProtoCrypto proto)
  ) =>
  SerialiseConstraintsHFC (ShelleyBlock proto era)
```

In any case. I'm going to remove this hunk and leave it for a different PR.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
agustinmista: Alright, I took a stab at this again: using derivingVia doesn't seem very feasible given the amount of associated types up in the chain that GHC is trying to coerce under the hood.

Here is a [commit](https://github.com/IntersectMBO/ouroboros-consensus/commit/a1e8d9d534690b6b5731de6a1275509fa088f674) removing the default implementations and replacing them with per-instance ones. Let me know if you prefer that and I will incorporate those changes here.

On a separate note, our [refactored](https://github.com/IntersectMBO/ouroboros-consensus/pull/2067/changes#diff-2bbdaf0087231e9007ba4e6553f77f2715d57884339b3eb8daa3aeea0ebbf7ca) `BlockSupportsPeras` type class relies heavily on default instances to avoid needing a substantial amount of boilerplate. If you think this would be an issue, please let me know and I will refactor those into separate definitions before continuing.

agustinmista:
Alright, I removed the default type class implementations. We should be good to go as soon as it passes CI :crossed_fingers::smile_cat::crossed_fingers:

jasagredo (approved):

PR #2222, opened 2026-08-19
Pin cardano-ledger to #5992

agustinmista:
Pins `cardano-ledger` to https://github.com/IntersectMBO/cardano-ledger/pull/5992 via an SRP.

This is needed for Peras/Leios to be able to extract the BLS keys that were added to the `PoolDistr` data structure.


agustinmista:
CCing @gdeest

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -509,7 +509,18 @@ instance
...
+              SL.IndividualPoolStake
+                { SL.individualPoolStake = SL.qripsStake qrips
+                , SL.individualTotalPoolStake = SL.qripsTotalStake qrips
+                , SL.individualPoolStakeVrf = SL.qripsVrf qrips
+                , SL.individualPoolStakeBls = SL.qripsBls qrips
+                }
+         in SL.PoolDistr
+              { SL.unPoolDistr = fmap toIndividualPoolStake (SL.qrpdDistr result)
+              , SL.pdTotalActiveStake = SL.qrpdTotalActiveStake result
+              }
```
jasagredo: I think the intention in the Ledger side is to provide the queries on their code, not having to reconstruct types in ouroboros-consensus, no? 

cc @aniketd

agustinmista: That sounds reasonable.

FTR: one direction is [already covered](https://github.com/IntersectMBO/cardano-ledger/blob/2211ea1259502191363096e7d4a32544082156da/libs/cardano-ledger-api/src/Cardano/Ledger/Api/State/Query.hs#L759) in ledger:

```haskell
toQueryResultIndividualPoolStake :: IndividualPoolStake -> QueryResultIndividualPoolStake
toQueryResultPoolDistr :: PoolDistr -> QueryResultPoolDistr
```

But we couldn't find anything going in the other direction yet.

ch1bo: It seems like their intention was to have dedicated types to be used on the queries to keep some external stability while changing `PoolDistr` internally? A consistent way to wire this up would be to just switch to their new dedicated types here.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -509,7 +509,18 @@ instance
       QueryStakePoolDefaultVote stakePool ->
         SL.queryStakePoolDefaultVote st stakePool
       GetPoolDistr2 mPoolIds ->
-        SL.querySetSnapshotStakePoolDistr st (maybe Set.empty id mPoolIds)
```
ch1bo, line 512: Should: just use the upstream data type

This is fine because it is byte identical in the encoding for past eras (according to ledger team)

ch1bo (approved):
Looks good, but we can avoid the back-translation.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
aniketd: > I think the intention in the Ledger side is to provide the queries on their code, not having to reconstruct types in ouroboros-consensus, no?

Yes, the `QueryResult*` types (more of them to come soon) are intended to provide some stability to downstream. Types like `IndividualPoolState` and `PoolDistr` are best considered internal to `ledger` as they can potentially change.

> But we couldn't find anything going in the other direction yet.

Yes, this is intentional. We want to avoid promoting the use of types used internally for the ledger state outside of ledger, as that increases unnecessary coupling, and then any changes in those representations end up needing a large integration effort.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
aniketd: Yes, the `QueryResult*` types preserve the original serialisation so as not to break protocol.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
ch1bo: Should: use the `QueryResult*` type here then

This would avoid the construction / mapping here altogether and ledger provides byte-level compatibility between the types.

ch1bo (approved):
We can avoid the translation, otherwise fine from my side (I'm not a maintainer of this repo though)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query/Types.hs
ch1bo, line 73: Is this still used? Is the `PoolDistr c` copy still used?

agustinmista: It seems that both the (deprecated?) `GetPoolDistr` and `GetStakeDistribution` queries are still using the old `PoolDistr c`.

Not sure what's the plan for those, but I do feel though that ironing them out might warrant its own separate PR :sweat_smile:

ch1bo: I'm fine with another PR for that purpose. Now is just the last time we get notified by the compiler that we need to do something about it :)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query/Types.hs
```diff
@@ -71,3 +74,16 @@ fromLedgerPoolDistr pd =
...
+  IndividualPoolStake
+    { individualPoolStake = SL.qripsStake ips
+    , individualPoolStakeVrf = SL.fromVRFVerKeyHash $ SL.qripsVrf ips
+    }
+
+fromQueryResultPoolDistr :: SL.QueryResultPoolDistr -> PoolDistr c
+fromQueryResultPoolDistr pd =
+  PoolDistr
+    { unPoolDistr = Map.map fromQueryResultIndividualPoolStake $ SL.qrpdDistr pd
+    }
```
jasagredo, line 89: This will be gone on #2265 , which I will rebase on top of this PR

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -448,7 +448,7 @@ instance
       GetStakeSnapshots mPoolIds ->
         SL.queryStakeSnapshots st mPoolIds
       GetPoolDistr mPoolIds ->
-        fromLedgerPoolDistr $ answerPureBlockQuery cfg (GetPoolDistr2 mPoolIds) ext
+        fromQueryResultPoolDistr $ answerPureBlockQuery cfg (GetPoolDistr2 mPoolIds) ext
```
jasagredo, line 451: Same

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
```diff
@@ -109,13 +108,14 @@ type instance BlockProtocol (ShelleyBlock proto era) = proto
...
   , ShelleyProtocol proto
-  , EncCBORGroup (SL.BlockBody era)
   , -- Header constraints
     Eq (ShelleyProtocolHeader proto)
   , Show (ShelleyProtocolHeader proto)
   , NoThunks (ShelleyProtocolHeader proto)
   , EncCBOR (ShelleyProtocolHeader proto)
   , DecCBOR (Annotator (ShelleyProtocolHeader proto))
+  , EncCBOR (SL.Block (ShelleyProtocolHeader proto) era)
+  , DecCBOR (Annotator (SL.Block (ShelleyProtocolHeader proto) era))
```
jasagredo: Perhaps we could group these in a separate section like the `-- Header constraints` above. This would be something like `-- Block constraints`. Not super important

ouroboros-consensus-cardano/test/tools-test/disk/config/dijkstra-genesis.json
```diff
@@ -2,5 +2,6 @@
   "maxRefScriptSizePerBlock": 1048576,
   "maxRefScriptSizePerTx": 204800,
   "refScriptCostStride": 25600,
-  "refScriptCostMultiplier": 1.2
+  "refScriptCostMultiplier": 1.2,
+  "minPoolMargin": 0.015
```
jasagredo, line 6: I don't know where are you getting this particular value from, but I guess it doesn't matter until Dijkstra is in effect

cabal.project
```diff
@@ -20,6 +20,31 @@ index-state:
 
 packages: .
 
+-- Temp: pin cardano-ledger to PR #5992
```
jasagredo:
```suggestion
-- Temp: pin cardano-ledger to PR IntersectMBO/cardano-ledger#5992
```

Like the title of the PR, the hyperlink on github resolves to 5992 here in ouroboros-consensus, so better to make this explicit just in case.

jasagredo (approved):
Some minor comments. I will rebase #2265 on top of this PR.

ouroboros-consensus-cardano/test/tools-test/disk/config/dijkstra-genesis.json
agustinmista: This is indeed a bit arbitrary. 

FTR: I took it from the [DijkstraGenesis example](https://github.com/IntersectMBO/cardano-ledger/blob/2978a29891c15e0e1d82cb21553b28e28924c214/eras/dijkstra/impl/testlib/Test/Cardano/Ledger/Dijkstra/Examples.hs#L173) in Ledger.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Block.hs
agustinmista: Addressed in 06c5211a27dca44756ba48070d78123d450070b1

cabal.project
agustinmista: Addressed in 06c5211a27dca44756ba48070d78123d450070b1

agustinmista:
Hey @jasagredo, thanks for the review!

I pushed some fixes to 06c5211a27dca44756ba48070d78123d450070b1 to simplify the future rebase of #2265, but let me know if you prefer me to squash them up into the existing commits.

PR #2223, opened 2026-08-19
Immutable/volatile split of LeiosDB + garbage collection

### LINKED ISSUE #969: LeiosDb volatile/immutable split + GC
### Why

### What

* Out of scope
    * Compressing tx blobs

geo2a:
Implements https://github.com/input-output-hk/ouroboros-leios/issues/969

`cardano-node` PR: https://github.com/IntersectMBO/cardano-node/pull/6656

This PR splits the contents of LeiosDB into volatile and immutable partitions, and periodically marking EBs as immutable once their referencing RB becomes immutable and garbage collection orphaned volatile EBs.

An EB is marked as immutable via the `immutable` flag column of the `ebs` table. All transactions (rows of the `txs` table) referenced by this EB via the `ebTxs` table become immutable as well, but there are no explicit marker in those tables.

The PR also adds observability of the volatile/immutable EB counts an the GC rates, which is wired into a Graphana dashboard which looks like this:

run-13-08-2026-k2160-v03-leiosdb-board

## Benchmarking

The PR adds the `leios-gc-bench` benchmark, which works both synthetic data and with a real `leios.db`. I've generated some input data by running the devnet overnight to test the performance:

```
Totals over 50 cycles:
  marked as immutable ebs/ebTxs/txs = 83/203314/130147
  evicted ebs/ebTxs/txs = 0/0/0
  insert collisions     = 0
  mark as immutable : min=20 ns  avg=42 ms  max=884 ms
  gc                : min=383 μs  avg=241 ms  max=12.049219837 s
```

The duration of the first immutable marking and GC are an outlier, which Claude attributes to cold OS page cache. The outliers are clearly visible on the following scatter plot (x axis is log scale):

gc-bench-plot-k2160

When running the benchmark with synthetic data, the outliers don't show up, probably because the sqlite database is written into before the start of the measurement:

```
Totals over 50 cycles:
  marked as immutable ebs/ebTxs/txs = 750/150000/150000
  evicted ebs/ebTxs/txs = 750/150000/150000
  insert collisions     = 0
  mark as immutable : min=38 ms  avg=43 ms  max=52 ms
  gc                : min=18 ms  avg=26 ms  max=257 ms
```
gc-bench-plot-synth

## Implementation details

One particular peculiarity that I had to iron out was the use of the `unsafe` FFI calls to the SQLite bindings. The immutalisation or GC operations need to run in a `safe` FFI call so that they are run on a separate GHC RTS capability and do not block the node from doing other things. This needs further investigation, together with auditing other FFI calls and making a weighted decision on which mechanism to better for every one of them.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
@@ -38,6 +38,7 @@ import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Builder as BB
 import qualified Data.ByteString.Lazy as BSL
+import Data.IORef (IORef, atomicModifyIORef', newIORef)
 import Data.Int (Int64)
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
@@ -102,16 +103,14 @@ newLeiosDBSQLite tracer dbPath = do
   notificationChan <- atomically newBroadcastTChan
   -- start a thread to sample the sizes of the volatile LeiosDB partition
   startVolatileStatsSampler tracer dbPath
-  -- One root call context per maintenance operation, held for the handle's
-  -- lifetime, so span ids stay unique across the calls of one node run. The
-  -- names label the ChainDB background threads that drive each operation.
   gcRootCtx <- rootCallCtx "leiosdb-gc"
   copyRootCtx <- rootCallCtx "leiosdb-copy"
+  noGCYetDoneRef <- newIORef True -- True until the first GC of this handle
   pure $
     LeiosDbHandle
       { subscribeEbNotifications =
           atomically (dupTChan notificationChan)
-      , leiosDbGarbageCollect = sqlGarbageCollect tracer gcRootCtx dbPath
+      , leiosDbGarbageCollect = sqlGarbageCollect tracer gcRootCtx dbPath noGCYetDoneRef
       , leiosDbMarkAsImmutable = sqlMarkAsImmutable tracer copyRootCtx dbPath
       , open = openSQLiteConnection tracer dbPath notificationChan
       }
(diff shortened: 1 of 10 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -222,6 +466,42 @@ openSQLiteConnection tracer dbPath notificationChan = do
...
 
+-- | Open an SQLite connection, setting the shared PRAGMAs plus the idempotent 'sql_schema'.
+openRawConnection :: HasCallStack => FilePath -> IO DB.Database
+openRawConnection dbPath = do
+  db <- open2 (fromString dbPath) [SQLOpenReadWrite, SQLOpenCreate] SQLVFSDefault
+  traverse_ (dbExec db) $
+    [ "pragma journal_mode = WAL;"
+    , "pragma synchronous = normal;"
+    , "pragma page_size = 32768;"
+    , "pragma mmap_size = 268435500;"
```
ch1bo, line 477: A claude instance mentioned that page size cannot be change after setting journal mode to WAL and it was all along at the default (`4096`). We should probably double check whether we really want the 32KiB

```suggestion
    [ "pragma synchronous = normal;"
    , "pragma page_size = 32768;"
    , "pragma mmap_size = 268435500;"
    , "pragma journal_mode = WAL;"
```

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -90,19 +101,244 @@ newLeiosDBSQLiteFromEnv tracer = do
...
+                withStmt db sql_gc_clear_candidates dbStep1Safe
+              let nTxsTotal = nTxs + nTxsFromFirstGC
+              when (nEbs > 0 || nEbTxs > 0 || nTxsTotal > 0) $
+                execWithInt64x3
+                  db
+                  sql_update_volatile_stats
+                  ( negate (fromIntegral nEbs)
+                  , negate (fromIntegral nEbTxs)
+                  , negate (fromIntegral nTxsTotal)
+                  )
```
ch1bo, line 226: Why are we writing metadata (counts of table rows) into the database itself?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -503,40 +787,89 @@ sqlLookupEbClosure conn ebHash =
...
+      "CREATE TABLE IF NOT EXISTS leiosDbStats ("
+    , "  id INTEGER PRIMARY KEY CHECK (id = 0),"
+    , "  volatileEbs INTEGER NOT NULL,"
+    , "  volatileEbTxs INTEGER NOT NULL,"
+    , "  volatileTxs INTEGER NOT NULL,"
+    , "  immutableEbs INTEGER NOT NULL,"
+    , "  immutableEbTxs INTEGER NOT NULL,"
+    , "  immutableTxs INTEGER NOT NULL"
+    , ");"
+    , -- TODO(geo2a): why exactly is this needed? Initialize the stats with the existing data on node restart?
```
ch1bo, line 835: Didn't you put it? :D

Basically the same question as I had above.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -503,40 +787,89 @@ sqlLookupEbClosure conn ebHash =
...
   unlines
-    [ "CREATE TABLE ebs ("
+    [ "CREATE TABLE IF NOT EXISTS ebs ("
     , "  ebSlot INTEGER NOT NULL,"
     , "  ebHashBytes BLOB NOT NULL,"
     , "  ebBytesSize INTEGER NOT NULL,"
     , -- NULL = body not downloaded, >0 = txs missing, 0 = just completed, <0 = notified
       "  missingTxCount INTEGER,"
+    , -- 1 = the immutable chain references this EB
+      "  immutable INTEGER NOT NULL DEFAULT 0,"
```
ch1bo, line 806: Must: not re-use the same table for the immutable partition

The whole goal is to make queries scanning the table or recomputing indices not needing to process the immutable entries. A separate table is the absolute minimum requirement we need. A separate file is what we actually want, to allow SPOs to keep the immutable DB on a potentially slower disk.

ch1bo (changes requested):

geo2a:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2261

PR #2225, opened 2026-08-21
Leios: move vote validation outside the STM transaction

ramsay-t:
## Problem

`addVote` runs the BLS pairing (`validateLeiosVote`) **inside** the `atomically` block that inserts into the shared vote state. Under gossip load every concurrently committing inbound vote aborts the transaction, and every retry re-executes the pairing. On a block producer this holds the node's **own, fully-ready vote** (EB closure downloaded and checked at ~+1.5s) for 5–10 seconds before casting — frequently missing the voting window.

## Change

Dedup peek (duplicates exit with zero crypto) → validate in plain IO (a retry can never re-run the pairing) → short insert transaction that re-checks membership. Worst case introduced: one redundant verification per *concurrently received* duplicate — bounded by receipts, not retries. `aggregateLeiosCert` deliberately stays inside (fires ~once per certified RB); the pre-existing committee-epoch TODO is unchanged and should be reviewed separately.

## Field evidence — reversible A/B/A on a live Musashi BP

Same node (NITEN, 4 vCPU KVM without ADX, `-N2`), same network, same measurement (closure-acquired → own-vote-cast per EB, from node.log timestamps):

| period | binary | votes | p90 | max | holds >5s |
|---|---|---|---|---|---|
| ~21 h | stock | 869 | 5.91s | 10.08s | **114** |
| ~18 h | patched | 901 | 2.14s | 3.12s | **0** |
| 30 min | stock (rollback) | 63 | 6.13s | 10.26s | **8** |
| post-restore | patched | 8+ | 2.51s | 2.51s | **0** |

Specimen from the rollback window: closure acquired at announce+1.0s, own vote cast at announce+11.1s — held 10.1s with everything in hand. The patched windows carried the heavier sustained load (574k inbound vote acquisitions over 18h).

A minimal, dependency-free reproduction of the retry re-execution (dedup-read + expensive step inside `atomically`, execution counter: 1 thread = 1.00×, 4 threads = 2.15×, 10 threads = 1.72×) and the full 1,841-vote dataset + measurement script are available in the field report (Musashi Slack thread).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
@@ -83,57 +83,65 @@ newLeiosVoteState getCommittee = do
   pointStates <- atomically $ newTVar (Map.empty :: Map RbHash PointState)
   pure
     LeiosVoteState
-      { addVote = \vote -> atomically $ do
-          seen <- readTVar seenVotes
-          if Set.member vote seen
+      { addVote = \vote -> do
+          -- Validate outside the transaction: the BLS pairing is ms-scale, and
+          -- inside 'atomically' every conflicting commit re-ran it. Worst case
+          -- now is one redundant verification per concurrently-received duplicate.
+          alreadySeen <- atomically $ Set.member vote <$> readTVar seenVotes
+          if alreadySeen
             then pure AlreadyKnown
             else do
               -- TODO: disallow votes from different epoch (than the committee is).
               -- Could use slot numbers or put epoch into votes to distinguish?
-              getCommittee >>= \case
+              mCommittee <- atomically getCommittee
+              case mCommittee of
                 Nothing -> pure NoCommittee
                 Just committee ->
                   case validateLeiosVote committee vote of
                     Left reason -> pure $ VoteInvalid reason
-                    Right weight -> do
-                      writeTVar seenVotes $! Set.insert vote seen
-                      writeTChan votesChan vote
+                    Right weight -> atomically $ do
+                      seen <- readTVar seenVotes
+                      if Set.member vote seen
+                        then pure AlreadyKnown
+                        else do
+                          writeTVar seenVotes $! Set.insert vote seen
+                          writeTChan votesChan vote
 
-                      -- FIXME: This code is not only ugly, but we need to also
-                      -- keep track of which committee the cert is for. We shall
-                      -- only use the cert (return on queryCert) if we are in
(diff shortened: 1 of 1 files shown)


ch1bo (approved):
Nice catch. Can you ask your agent how much of a difference this makes in actual "time to release" a vote? The measured delta between acquiring and voting is not as helpful because it mixes the receival into the measure. A clearer answer could be the exact time of `LeiosVoted` compared with the `earliestVoteSlot`, that is slot of announcement + 3 * L_hdr (= `3s` after announcment creation)

ramsay-t:
*(Claude here — Ramsay's agent, posting via his account by his request.)*

Good framing — computed exactly that from the same dataset (`LeiosVoted` − `earliestVoteSlot`, with `earliestVoteSlot` = announce slot + 3·L_hdr, L_hdr = 1s):

| period | n | p50 | p90 | p99 | max | released ≤0.5s after gate |
|---|---|---|---|---|---|---|
| stock | 869 | 0.28s | **5.48s** | 7.99s | 9.64s | 55% |
| patched | 901 | 0.19s | **0.88s** | 4.59s | 5.59s | 86% |
| stock (rollback) | 63 | 0.19s | **6.92s** | 10.34s | 10.34s | 59% |
| patched (restored) | 8 | 0.16s | 3.24s | 3.24s | 3.24s | 88% |

So the median release was always gate-dominated — the difference is entirely in the tail: on stock, the p90 vote releases **5.5–7s after the window legally opens** (the contention cost), while patched brings that to **~0.9s** and lifts the released-within-0.5s share from ~55–59% to 86–88%.

One caveat this metric needs: the residual patched tail (p99 4.6s) is closures arriving *after* the gate opened — in every such case the vote follows the closure by <1s, so it's receive-side latency, outside this patch's scope (it's the announcement-diffusion issue discussed in the Musashi thread).

Per-vote data for all four periods is in the field-report CSV if you want to slice it differently.

ramsay-t:
*(Claude again, via Ramsay's account.)*

CI note for reviewers: the remaining **9.14.1** failure is inherited from the base branch — it dies compiling `unstable-consensus-testlib/Test/Util/Serialisation/Roundtrip.hs:665` (`withMaxSuccess` deprecation promoted to error under the newer QuickCheck), a file this PR doesn't touch. The `leios-prototype` tip fails the same job without this commit (see the branch's run history). 9.12.2 passed on re-run after the rebase; hlint is green after d52f358. Everything this PR *can* turn green is green.

PR #2226, opened 2026-08-21
[Peras 61] Add default BlockSupportPeras implementation helpers 

agustinmista:
This PR introduces a default implementation for the upcoming methods in the refactored `BlockSupportPeras` type class:

* `defaultForgePerasVoteIfEligible`
* `defaultVerifyPerasVote`
* `defaultForgePerasCert`
* `defaultVerifyPerasCert`

NOTE: these currently have redundant constraints that had to be suppressed temporarily until we no longer have a degenerate `BlockSupportPeras` instance.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
@@ -12,6 +12,8 @@
 {-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+-- Disabled until we get rid of the degenerate 'BlockSupportsPeras' instance.
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
 module Ouroboros.Consensus.Block.SupportsPeras
   ( -- * Voting committee types for Peras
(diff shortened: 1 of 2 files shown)


jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -242,6 +251,141 @@ class
...
+      (PerasCrypto blk)
+      (PerasVotingCommitteeScheme blk)
+  ) =>
+  PerasEpochContext blk ->
+  PerasVoteCollectionWithQuorum blk ->
+  Either (PerasError blk) (ValidatedPerasCert blk)
+defaultForgePerasCert context voteCollection = do
+  let params = pecParams context
+  abstractVoteCollection <-
+    bimap injectConversionError id $
```
jasagredo, line 339: Isn't this combinator exactly `Data.Bifunctor.first :: (a -> b) -> p a c -> p b c`

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
agustinmista: Yes, it is! We tend to write it this way mostly because we find it less confusing than using the poorly named `first`/`second`, and in many cases we actually want to map both sides, so we just use `bimap` in all cases.

PR #2227, opened 2026-08-21
Decouple LedgerDB garbage collection from snapshotting

jasagredo:
As the thread delay for snapshotting was blocking the snapshotting thread, we
would accumulate states in the LedgerDB until the snapshot was taken. We could
have done this earlier but we didn't notice. This change just runs that code in
a separate thread, allowing garbage collection to proceed.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
@@ -191,6 +191,7 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
       )
       (Query.getAnyKnownBlock immutableDB volatileDB)
       ledgerDbGetVolatileSuffix
+      (Args.cdbsRegistry cdbSpecificArgs)
 
   lift $ do
     traceWith tracer $ TraceOpenEvent OpenedLgrDB
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
@@ -191,7 +191,6 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
       )
       (Query.getAnyKnownBlock immutableDB volatileDB)
       ledgerDbGetVolatileSuffix
-      (Args.cdbsRegistry cdbSpecificArgs)
 
   lift $ do
     traceWith tracer $ TraceOpenEvent OpenedLgrDB
@@ -323,7 +322,12 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
                 Background.garbageCollectBlocks e slot
                 Background.garbageCollectPeras e slot
                 LedgerDB.garbageCollect (cdbLedgerDB e) slot
-            , intTryTakeSnapshot = getEnv2 h $ LedgerDB.tryTakeSnapshot . cdbLedgerDB
+            , intTryTakeSnapshot = getEnv h $ \e -> do
+                -- 'tryTakeSnapshot' only enqueues a request for the snapshots;
+                -- as the background thread that would serve it is usually not
+                -- running in the tests using this, serve it ourselves.
+                LedgerDB.tryTakeSnapshot (cdbLedgerDB e)
+                Background.tryServeSnapshotRequestNow e
             , intAddBlockRunner = getEnv h (Background.addBlockRunner addBlockTestFuse)
             , intKillBgThreads = varKillBgThreads
             }
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
@@ -21,6 +21,10 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.Background
     -- * Copying blocks from the VolatileDB to the ImmutableDB
   , copyToImmutableDB
 
+    -- * Writing LedgerDB snapshots
+  , snapshotRunner
+  , tryServeSnapshotRequestNow
+
     -- * Executing garbage collection
   , garbageCollectBlocks
   , garbageCollectPeras
@@ -120,6 +124,10 @@ launchBgTasks cdb@CDB{..} = do
     launch "ChainDB.copyToImmutableDBRunner" $
       copyToImmutableDBRunner cdb ledgerDbTasksTrigger gcSchedule
 
+  !snapshotThread <-
+    launch "ChainDB.snapshotRunner" $
+      snapshotRunner cdb
+
   atomically $
     writeTVar cdbKillBgThreads $
       sequence_
@@ -127,6 +135,7 @@ launchBgTasks cdb@CDB{..} = do
         , cancelThread ledgerDbMaintenanceThread
         , gcThread
         , copyToImmutableDBThread
+        , snapshotThread
         ]
  where
   launch :: String -> m Void -> m (m ())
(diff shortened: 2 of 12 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -339,58 +345,90 @@ implTryTakeSnapshot ::
...
+  -- then discards.
+  mayStart <- atomically $ do
+    inFlight <- readTVar (ldbSnapshotInFlight env)
+    if inFlight
+      then pure False
+      else do
+        -- Claim the slot; released by 'releaseInFlight'.
+        writeTVar (ldbSnapshotInFlight env) True
+        pure True
+  Monad.when mayStart $ flip onException releaseInFlight go
```
bladyjoker: Hmm, thinking maybe it would be simpler to manage with a `TBQueue` here? You're main loop in "Snapshot" thread just reads requests in the `requestsQ` and processes them sequentially.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -339,58 +345,90 @@ implTryTakeSnapshot ::
...
+        (slot,) <$> (duplicateStateRef $ anchorHandle $ snd $ prune pruneStrat lseq)
+
+    -- look at the list of the ledger tables handles from the previous step and take the snapshots
+    case NonEmpty.nonEmpty handles of
+      -- Nothing was selected, so no request is in flight after all.
+      Nothing -> releaseInFlight
+      -- Deferred so that the delay does not block the caller.
+      Just nonEmptyHandles ->
+        Monad.void $
+          forkLinkedThread (ldbRegistry env) "LedgerDB.takeSnapshot" $
```
bladyjoker: Thinking maybe it makes sense to include this thread in NodeKernel. With a `TBQueue` this module can just dispatch a requests and implement the harness that takes care of delays and whatnot.

bladyjoker: The dispatch also is then "clearly" non blocking.

bladyjoker: Unless we filled the queue, which makes me ask, how is the situation managed now when you get much more requests than you are willing to handle? Just dropping them?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -339,58 +345,90 @@ implTryTakeSnapshot ::
...
+            flip finally releaseInFlight $ do
+              copyBlocks
+
+              delayBeforeSnapshotting <- getRandomDelay (onDiskSnapshotDelayRange (ldbSnapshotPolicy env))
+              traceWith (LedgerDBSnapshotEvent >$< ldbTracer env) $
+                SnapshotRequestDelayed now delayBeforeSnapshotting (NonEmpty.map fst nonEmptyHandles)
+              threadDelay delayBeforeSnapshotting
```
bladyjoker: Not sure how, if this could be part of the "request dispatch harness" so it doesn't intervene here. From what I can tell we're trying to artificially manage the rate at which snapshots can be taken?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -339,58 +345,90 @@ implTryTakeSnapshot ::
...
+  releaseInFlight :: m ()
+  releaseInFlight = atomically $ writeTVar (ldbSnapshotInFlight env) False
+
+  go :: m ()
+  go = do
+    now <- getMonotonicTime
+    timeSinceLastSnapshot <- do
+      mLastSnapshotRequested <- readTVarIO $ ldbLastSuccessfulSnapshotRequestedAt env
+      for mLastSnapshotRequested $ \lastSnapshotRequested -> do
+        pure $ now `diffTime` lastSnapshotRequested
```
bladyjoker, line 361: Thinking how to separate this song and dance of managing delays in a way that's not interleaved with main logic.

If you manage to come up with a function that is easy to test that would be sweet. Smtn like this to run in a separate thread.

```haskell
processWithRandomDelay :: DiffTime -> TBQueue a -> (a -> m ()) -> m ()
processWithRandomDelay maxDelay reqQ act = do
  forever $ do
    delay <- getRandomDelay maxDelay
    threadDelay delay
    req <- readTBQueue reqQ
    act req
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -339,58 +345,90 @@ implTryTakeSnapshot ::
...
+              copyBlocks
+
+              delayBeforeSnapshotting <- getRandomDelay (onDiskSnapshotDelayRange (ldbSnapshotPolicy env))
+              traceWith (LedgerDBSnapshotEvent >$< ldbTracer env) $
+                SnapshotRequestDelayed now delayBeforeSnapshotting (NonEmpty.map fst nonEmptyHandles)
+              threadDelay delayBeforeSnapshotting
+
+              for_ nonEmptyHandles $ \(_, h) -> do
+                Monad.void $ takeSnapshot snapManager Nothing h
+                Monad.void $ close . tables $ h
```
bladyjoker: async this?

bladyjoker (commented):
I'm not familiar enough to talk about correctness here.

So I will simply comment on the correctness agnostic points.

My main points are that "client" (caller) and "server" (actual snapshotter) can be decoupled with `TBQueue` and similar, and even introduce a new "LedgerDB.takeSnapshot" thread in NodeKernel that processes snapshot requests in sequence.

I find that delay management could also be further separated from the main logic of taking snapshots, which would enable us to even test these in separation hopefully.

That being said, I see this is about backports, so I sense I'm late to the party and will leave someone qualified to comment on correctness rather than code structure.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -314,26 +323,107 @@ triggerLedgerDbTasks (LedgerDbTasksTrigger varSt) =
...
+
+serveSnapshotRequest ::
+  forall m blk.
+  IOLike m =>
+  ChainDbEnv m blk ->
+  -- | Run before writing the snapshots; in a running node this copies the
+  -- immutable blocks from the VolatileDB to the ImmutableDB, as the LedgerDB
+  -- assumes that all snapshots correspond to immutable blocks.
+  m () ->
+  (LedgerDB.SnapshotDelayRange -> m DiffTime) ->
```
bladyjoker: Why is `getRandomDelay` passed around as argument?

Since you're passing around `CDB` that's everything you need to construct `getRandomDelay` exactly at the call site.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -314,26 +323,107 @@ triggerLedgerDbTasks (LedgerDbTasksTrigger varSt) =
...
+  , HasHeader blk
+  , GetHeader blk
+  , HasCallStack
+  ) =>
+  ChainDbEnv m blk ->
+  m Void
+snapshotRunner cdb@CDB{..} =
+  forever $
+    LedgerDB.withSnapshotRequest (LedgerDB.snapshotRequestQueue cdbLedgerDB) $
+      serveSnapshotRequest cdb (void $ copyToImmutableDB cdb) (mkRandomSnapshotDelay cdb)
```
bladyjoker: Here too, `cdb` is sufficient to create both `copyBlocks` and `getRandomDelay` right? So why not just create it in `serveSnapshotRequest`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -314,26 +323,107 @@ triggerLedgerDbTasks (LedgerDbTasksTrigger varSt) =
...
+  ChainDbEnv m blk ->
+  -- | Run before writing the snapshots; in a running node this copies the
+  -- immutable blocks from the VolatileDB to the ImmutableDB, as the LedgerDB
+  -- assumes that all snapshots correspond to immutable blocks.
+  m () ->
+  (LedgerDB.SnapshotDelayRange -> m DiffTime) ->
+  LedgerDB.SnapshotRequest m ->
+  m ()
+serveSnapshotRequest CDB{cdbTracer} copyBlocks getRandomDelay req = do
+  copyBlocks
```
bladyjoker: This can just be
```haskell
copyToImmutableDB cdb

delayBeforeSnapshotting <- mkRandomSnapshotDelay cdb (LedgerDB.snapshotRequestDelayRange req)
```

and we get rid of `copyBlocks` and `getRandomDelay` arguments

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -322,7 +322,11 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
                 Background.garbageCollectBlocks e slot
                 Background.garbageCollectPeras e slot
                 LedgerDB.garbageCollect (cdbLedgerDB e) slot
-            , intTryTakeSnapshot = getEnv2 h $ LedgerDB.tryTakeSnapshot . cdbLedgerDB
+            , intTryTakeSnapshot = getEnv2 h $ \e copyBlocks getRandomDelay -> do
+                LedgerDB.tryTakeSnapshot (cdbLedgerDB e)
+                -- The background snapshotting thread is usually not running in
+                -- the tests that use this, so serve the request ourselves.
+                Background.tryServeSnapshotRequest e copyBlocks getRandomDelay
```
bladyjoker: I don't understand this, can you explain?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -328,69 +331,88 @@ implGarbageCollect env slotNo = do
...
+    -- serving the queue, which will take the snapshots
+    whenJust (NonEmpty.nonEmpty handles) $ \nonEmptyHandles -> do
+      enqueued <-
+        tryEnqueueSnapshotRequest
+          queue
+          SnapshotRequest
+            { snapshotRequestedAt = now
+            , snapshotRequestSlots = NonEmpty.map fst nonEmptyHandles
+            , snapshotRequestDelayRange = onDiskSnapshotDelayRange (ldbSnapshotPolicy env)
+            , writeSnapshots = write now nonEmptyHandles
```
bladyjoker, line 394: Why is `writeSnapshots` a thing? Why isn't `write` defined and used in `ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -328,69 +331,88 @@ implGarbageCollect env slotNo = do
...
+    -- hand the ledger tables handles from the previous step over to the thread
+    -- serving the queue, which will take the snapshots
+    whenJust (NonEmpty.nonEmpty handles) $ \nonEmptyHandles -> do
+      enqueued <-
+        tryEnqueueSnapshotRequest
+          queue
+          SnapshotRequest
+            { snapshotRequestedAt = now
+            , snapshotRequestSlots = NonEmpty.map fst nonEmptyHandles
+            , snapshotRequestDelayRange = onDiskSnapshotDelayRange (ldbSnapshotPolicy env)
```
bladyjoker: `snapshotRequestDelayRange` is the same across all requests? If so, it's probably misplaced how it's supplied.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
```diff
@@ -328,69 +331,88 @@ implGarbageCollect env slotNo = do
...
     pure (close, ())
 
+-- | Decide whether snapshots should be taken and, if so, enqueue a request for
+-- them on the given queue.
+--
+-- Selecting and duplicating the ledger tables handles happens here, on the
+-- calling thread, because it has to see the 'LedgerSeq' before the caller goes
+-- on to garbage-collect it. Everything after that -- in particular the
+-- randomised delay, which is measured in minutes -- happens in whichever
+-- thread serves the queue, so that this one is free to carry on.
```
bladyjoker, line 342: Not sure I understand why most of the request processing logic is here and not in the thread? Was expecting all of that basically in the snapshotRunner

bladyjoker (commented):
I think my main question is why is most of the logic in the "client" and not the "server" (`snapshotRunner`)?

Th core API is `tryTakeSnapshot :: m ()`, it's not accepting a list of slots for example. So I believe that having a request like

```haskell
data SnapshotRequest m = SnapshotRequest
  { snapshotRequestedAt :: !Time
  }
```
is sufficient?

Currently the "server" is not really using any of the fields and just invoking `writeSnapshots req`. Which is a signal something is odd. Instead, if you'd move all logic from V2 module into Background module, it'd be a more straighforward flow, no `  inFlight <- snapshotRequestInFlight queue`, no `Monad.unless enqueued $ closeHandles nonEmptyHandles`, just a forever loop that reads requests and processes them.

Clients dispatch the request, the server thread reads requests, processes them with delay and that's it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: We are now using a 1-queue

bladyjoker (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Yes, we are now using a 1-queue, and drop the snapshots if they come concurrently

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: This has to come just after the snapshotting

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
jasagredo: Indeed, good point

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
jasagredo: Indeed, creating it in the server directly

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
jasagredo: Indeed

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
jasagredo: This is just for tests, the internal components are to be used by state machine and others, which don't have background threads, so we take the snapshot in place

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Because we would need to expose the types for the handles to the Background module and I'd rather avoid that

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Indeed

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/V2.hs
jasagredo: Because we would need to expose a lot of things to the server, which I would rather not do

PR #2232, opened 2026-08-24
[Peras 62] Rewire default BlockSupportsPeras to use mocked types

agustinmista:
This PR:

- Moves `MockPeras*` types `unstable-consensus-testlib` to the main `ouroboros-consensus` library.
- Reworks the degenerate `BlockSupportsPeras` instance to use `MockPeras*` types. This change supplies non-void associated types ahead of the later `PerasState` integration (next PR in this stack).
- Removes the old local `PerasVote'` and `PerasCert'` types and their associated instances.

NOTE: most of the breaking changes in the test suite are "final", as we intend to keep using `MockPeras*` types for most tests after removing the degenerate `BlockSupportsPeras` instance.

NOTE: node-to-node serialisation instances live in `O.C.Node.Serialisation` rather than beside the mock types solely to avoid a module cycle. This is strictly temporary.


agustinmista:
Quick question @jasagredo: would you add a changelog to this PR? 

My instinct is that we could skip it given that this is either:
* A temporary change for production blocks (those will be rewired to `VoidPeras*` or `V1.Peras*` types, or
* A change affecting test suites.

jasagredo (approved):

jasagredo:
I agree this needs no changelog @agustinmista precisely because of the reasons you mention.

PR #2233, opened 2026-08-25
LeiosDemoDb: BEGIN IMMEDIATE for writing transactions

### LINKED ISSUE #969: LeiosDb volatile/immutable split + GC

ch1bo:
Should fix the last remaining occurrences of `LeiosDbException ErrorBusy`

A bare BEGIN is BEGIN DEFERRED, so a transaction that reads before it writes has to upgrade its lock. In WAL mode that upgrade fails with SQLITE_BUSY_SNAPSHOT whenever another connection committed in between, and that status is neither serviced by the busy handler nor retryable at the statement level: the transaction's snapshot is stale for good, so re-stepping the same statement can never succeed. The retry loops in withDie and dbStepInsert spin to exhaustion and then throw, which arrives as ExceptionInLinkedThread and takes the node down.

Observed on a twelve-node devnet under load: 297 LeiosDbException ErrorBusy across all twelve nodes, every single one from the same statement, the stMarkNotifiedEbs UPDATE in sqlInsertTxs. That transaction inserts only txs not already persisted, so under backlog the inserts are frequently empty, the SELECT loop takes the snapshot, and that UPDATE is the first write, hence the upgrade. It never failed when inserts had happened, because the write lock was already held.

Taking the lock at BEGIN makes that upgrade impossible, so the only place a writer can now be told BUSY is the BEGIN itself.

Splits the helper so readers keep BEGIN DEFERRED and do not exclude each other, and only the three writers take the lock up front.

Also reduces the maximum retries to `1000`, leading to an error after exhausting retries within ~50s instead of `~5000s`.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
@@ -269,7 +269,7 @@ sqlLookupEbBody conn ebHash =
 
 sqlInsertEbPoint :: Conn -> LeiosPoint -> BytesSize -> IO ()
 sqlInsertEbPoint conn point ebBytesSize =
-  dbWithTransaction db $ useStmt stmt $ do
+  dbWithWriteTransaction db $ useStmt stmt $ do
     dbBindInt64 stmt 1 (fromIntegral $ unSlotNo point.pointSlotNo)
     dbBindBlob stmt 2 point.pointEbHash.ebHashBytes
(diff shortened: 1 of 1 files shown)


nfrisby (approved):
Looks to me like a reasonable fix to attempt

geo2a (approved):

ch1bo:
More fixes following this in https://github.com/IntersectMBO/ouroboros-consensus/pull/2245

PR #2235, opened 2026-08-25
Leios prototype: EB validation in vote logic

### LINKED ISSUE #1016: Prototype: EB validation

ch1bo:
Resolves and validates EB closures in `runLeiosVoting`.

This is using the `TxCache` from https://github.com/IntersectMBO/ouroboros-consensus/pull/2188 to determine whether we need to `applyTx` or `reapplyTx` via the existing `LedgerSupportsMempool` API.

As this change is heavily restructuring the Leios voting loop, I realized that we incorrectly used the `currentSlot` to determine whether voting timed out -> this is wrong! The current slot (derived via `ChainDB.getCurrentLedger`) is only ever advancing when the tip advances. Thus, the "time" was lagging quite a lot. Changed the semantics determine onset and deadline of voting period using the wall clock (via `slotToWallclock`).

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -370,14 +370,21 @@ prop_leios seed =
       & counterexample ("announced-and-certified slots: " <> show announcedAndCertifiedSlots)
 
   -- In an honest net every acquired closure must apply against the announcing
-  -- RB's ledger state. A rejection here means the voting thread validates
-  -- against the wrong state or the wrong slot, not that anything was invalid.
+  -- RB's ledger state, because 'partitionMempool' cuts both the RB's and the
+  -- EB's transactions out of a single mempool snapshot: the snapshot is
+  -- internally consistent, so its transactions apply in sequence against the
+  -- state it was taken for, and the announcing RB /is/ that state plus its own
+  -- share of the split. The EB's share therefore applies against the RB's
+  -- post-state by construction. So a rejection here cannot mean a transaction
+  -- was bad; it means the voting thread validated against the wrong state or
+  -- the wrong slot.
   propClosuresValidate =
     null closureRejections
       & counterexample "a voter rejected an honestly-produced EB's transactions"
       & prettyCounterexampleList "rejections" 240 closureRejections
       -- The reapply-vs-apply split: how much full validation the LeiosTxCache
-      -- spared us. A forger's own EB should be entirely reapplied.
+      -- spared us. Tabulated only for visibility -- nothing here requires any
+      -- particular ratio.
       & tabulate
         "EB closure validation (reapplied/total)"
         [show reapplied <> "/" <> show (reapplied + applied) | (reapplied, applied) <- validatedSplits]
@@ -817,13 +824,14 @@ endorseInvalidTx adversary nid tni
 -- behaviour we care about. This tx is unsigned for the same reason — the
 -- signature is irrelevant to the check that has to catch it.
 --
--- Forced, because 'ForgedLeiosEb' and the forge path run 'NoThunks'
--- invariants over what they are handed.
 -- 'assumeValidatedClosureTx' is how the adversary lies: it stamps the tx as
 -- mempool-validated so the forge will endorse it, which is exactly the claim a
 -- voter is now supposed to disbelieve.
 invalidEbTx :: Validated (GenTx (CardanoBlock StandardCrypto))
-invalidEbTx = assumeValidatedClosureTx (GenTxDijkstra (mkShelleyTx (force tx)))
+invalidEbTx =
+  -- 'force'd because 'ForgedLeiosEb' and the forge path run 'NoThunks'
+  -- invariants over what they are handed, as the mempool does in 'respendTx'.
+  assumeValidatedClosureTx (GenTxDijkstra (mkShelleyTx (force tx)))
  where
   tx :: Tx TopTx DijkstraEra
   tx = mkBasicTx (mkBasicTxBody & inputsTxBodyL .~ Set.singleton phantomInput)
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -74,6 +74,7 @@ import LeiosDemoTypes
   ( LeiosNotVotedReason (..)
   , LeiosPoint (..)
   , LeiosVote (..)
+  , RbHash (..)
   , TraceLeiosKernel (..)
   , hashLeiosEb
   , minCertificationGap
@@ -304,9 +305,10 @@ prop_leios seed =
         -- counts shrugs off the spurious 'TraceLeiosVoteAcquired's emitted
         -- by 'NodeToNode' for relay-redelivered votes that 'addVote'
         -- already classified 'AlreadyKnown'.
-        Set.size acquiredVotePairs === numNodes * Set.size castVotes
+        Set.size votePairsToDiffuse === numNodes * Set.size votesToDiffuse
           & counterexample "created votes not diffused"
           & counterexample ("cast votes: " <> show (Set.size castVotes))
+          & counterexample ("votes required to diffuse: " <> show (Set.size votesToDiffuse))
           & counterexample
             ( "acquired (node, vote) pairs: "
                 <> show (Set.size acquiredVotePairs)
@@ -329,6 +331,23 @@ prop_leios seed =
 
   -- Which EB slot each announcing RB hash announced, so that a vote can be
   -- dated: 'LeiosVote' carries only the hash it signed.
+  -- Which EB each announcing RB hash announced, so that a vote can be dated:
+  -- 'LeiosVote' carries only the hash it signed.
+  announcedPoints :: Map RbHash LeiosPoint
+  announcedPoints = Map.fromList . flip mapMaybe leiosTraces $ \case
+    TraceLeiosBlockAnnounced{announcingRbHashBytes, announcedEbPoint} ->
+      Just (MkRbHash announcingRbHashBytes, announcedEbPoint)
+    _ -> Nothing
+
+  -- Same requirement the EB diffusion properties make, one step further along:
+  -- a vote is required to have diffused iff its EB had 'minCertificationGap'
+  -- slots left to propagate. A vote cast near the end of the sim can still be
+  -- in flight when it stops, which says nothing about diffusion.
+  votesToDiffuse =
+    flip Set.filter castVotes $ \vote ->
+      maybe False diffusionRequired (Map.lookup vote.announcingRbHash announcedPoints)
+
+  votePairsToDiffuse = Set.filter ((`Set.member` votesToDiffuse) . snd) acquiredVotePairs
 
   -- Distinct (node, vote) pairs from 'TraceLeiosVoteAcquired'. A vote
   -- received multiple times by the same node (e.g. relayed back via a
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -329,8 +329,6 @@ prop_leios seed =
     TraceLeiosVoted{vote} -> Just vote
     _ -> Nothing
 
-  -- Which EB slot each announcing RB hash announced, so that a vote can be
-  -- dated: 'LeiosVote' carries only the hash it signed.
   -- Which EB each announcing RB hash announced, so that a vote can be dated:
   -- 'LeiosVote' carries only the hash it signed.
   announcedPoints :: Map RbHash LeiosPoint
@@ -775,6 +773,7 @@ prop_leios_invalid_eb seed
     ChainTipDoesNotAnnounce -> "ChainTipDoesNotAnnounce"
     TooLate -> "TooLate"
     NotOnCommittee -> "NotOnCommittee"
+    VoteRejected{} -> "VoteRejected"
 
 -- | The EB points certified by a chain's CertRBs. A CertRB carries no
 -- announcement of its own for the EB it certifies; the announcement lives on
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -71,7 +71,8 @@ import LeiosDemoDb
   , withLeiosDb
   )
 import LeiosDemoTypes
-  ( LeiosPoint (..)
+  ( LeiosNotVotedReason (..)
+  , LeiosPoint (..)
   , LeiosVote (..)
   , TraceLeiosKernel (..)
   , hashLeiosEb
@@ -79,8 +80,13 @@ import LeiosDemoTypes
   , prettyEbHash
   , prettyLeiosPoint
   )
-import Lens.Micro ((%~), (^.))
+import Lens.Micro ((%~), (.~), (^.))
 import Ouroboros.Consensus.Block (SlotNo (..), blockSlot, getHeader)
+import Ouroboros.Consensus.Block.Forging
+  ( BlockForging (..)
+  , ForgeBlockArgs (..)
+  , MkBlockForging (..)
+  )
 import Ouroboros.Consensus.Cardano
   ( CardanoBlock
   , Nonce (NeutralNonce)
@@ -88,7 +94,11 @@ import Ouroboros.Consensus.Cardano
   , ProtocolParamsShelleyBased (..)
   , ShelleyGenesis (..)
   )
-import Ouroboros.Consensus.Cardano.Block (pattern BlockDijkstra, pattern LedgerStateDijkstra)
+import Ouroboros.Consensus.Cardano.Block
+  ( pattern BlockDijkstra
+  , pattern GenTxDijkstra
+  , pattern LedgerStateDijkstra
+  )
 import Ouroboros.Consensus.Cardano.Node (CardanoProtocolParams (..), protocolInfoCardano)
 import Ouroboros.Consensus.Config (SecurityParam (..), TopLevelConfig, configLedger)
 import Ouroboros.Consensus.HeaderValidation (headerStateChainDep)
@@ -103,7 +113,7 @@ import Ouroboros.Consensus.Ledger.Extended
   , ExtLedgerState (..)
   , ledgerState
   )
-import Ouroboros.Consensus.Ledger.SupportsMempool (extractTxs)
+import Ouroboros.Consensus.Ledger.SupportsMempool (GenTx, extractTxs)
 import Ouroboros.Consensus.Ledger.Tables.MapKind (EmptyMK, ValuesMK)
 import Ouroboros.Consensus.Ledger.Tables.Utils (applyDiffs, forgetLedgerTables)
 import Ouroboros.Consensus.Mempool (TraceEventMempool (..))
@@ -115,6 +125,7 @@ import Ouroboros.Consensus.Shelley.Ledger.Ledger
   , shelleyLedgerState
   , shelleyLedgerTip
   )
+import Ouroboros.Consensus.Shelley.Ledger.Mempool (mkShelleyTx)
 import Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
 import Ouroboros.Consensus.Storage.LedgerDB (ResolveLeiosBlock (..))
 import qualified Ouroboros.Network.Mock.Chain as Chain
@@ -133,6 +144,7 @@ import Test.QuickCheck
   , choose
   , conjoin
   , counterexample
+  , discard
   , forAll
   , ioProperty
   , property
@@ -188,6 +200,8 @@ tests =
         testProperty "basic functionality" prop_leios
     , adjustQuickCheckTests (`div` 10) $
         testProperty "late join" prop_leios_late_join
+    , adjustQuickCheckTests (`div` 10) $
+        testProperty "invalid endorsed tx is not certified" prop_leios_invalid_eb
     ]
 
 -- | Verify a suite of basic Leios ThreadNet invariants in a single run:
@@ -310,6 +324,9 @@ prop_leios seed =
   castVotes = Set.fromList . flip mapMaybe leiosTraces $ \case
     TraceLeiosVoted{vote} -> Just vote
     _ -> Nothing
+
+  -- Which EB slot each announcing RB hash announced, so that a vote can be
+  -- dated: 'LeiosVote' carries only the hash it signed.
 
   -- Distinct (node, vote) pairs from 'TraceLeiosVoteAcquired'. A vote
   -- received multiple times by the same node (e.g. relayed back via a
@@ -573,6 +590,214 @@ prop_leios_late_join seed =
  where
   numSlots = 200 :: Word64
 
+-- | An EB whose closure cannot apply must never be certified: the committee is
+-- supposed to validate the endorsed transactions before signing a vote.
+--
+-- 4 nodes, 200 slots, node 1 adversarial. Only the EBs it announces carry the
+-- bogus tx — its RBs stay honest — so the chain keeps growing and the honest
+-- nodes' own EBs still certify, which is what makes 'poisonedNeverCertified'
+-- more than a statement about a stalled network.
+prop_leios_invalid_eb :: Seed -> Property
+prop_leios_invalid_eb seed
+  -- An unlucky seed where the adversary never led early enough to announce an
+  -- EB with time to diffuse. There is nothing to conclude either way, so don't
+  -- count the run rather than passing it vacuously.
+  | Set.null poisonedPointsToDiffuse = discard
+  -- Whether any honest EB reaches quorum is seed-dependent: votes scatter
+  -- across forks, so a 4-node run can finish with nothing certified at all.
+  -- Such a run cannot tell "the poisoned EB was rejected" apart from "nothing
+  -- was certified", so it is no evidence either -- discard rather than fail.
+  | Set.null honestCertifiedPoints = discard
+  | otherwise =
+      conjoin
+        [ adversaryWasEffective
+            & counterexample "[failed] adversaryWasEffective"
+        , poisonedNeverCertified
+            & counterexample "[failed] poisonedNeverCertified"
+        , poisonedNeverValidated
+            & counterexample "[failed] poisonedNeverValidated"
+        , validationDidRun
+            & counterexample "[failed] validationDidRun"
+        , isNothing testOutput.exceptionThrown
+            & counterexample "[failed] test threw an exception"
+            & prettyCounterexampleList "all traces" 120 (show <$> traces)
+        ]
+        & tabulate "poisoned EB turned down on" poisonedNotVotedReasons
+ where
+  adversary = CoreNodeId 1
+
+  numCoreNodes = NumCoreNodes 4
+
+  numSlots = 200 :: Word64
+
+  (testOutput, _) =
+    runThreadNet'
+      seed
+      (NumSlots numSlots)
+      numCoreNodes
+      (trivialNodeJoinPlan numCoreNodes)
+      (endorseInvalidTx adversary)
+
+  traces = testOutput.allTraces
+
+  -- 'TraceLeiosBlockForged' is emitted by the forger only, so the emitting
+  -- node id attributes each EB to whoever made it.
+  forgedBy nid = Set.fromList $ flip mapMaybe traces $ \case
+    FromNode nid' (FromLeios TraceLeiosBlockForged{slot, eb})
+      | nid' == nid ->
+          Just (MkLeiosPoint slot (hashLeiosEb eb))
+    _ -> Nothing
+
+  poisonedPoints = forgedBy adversary
+
+  -- Only EBs forged early enough to have 'minCertificationGap' slots left can
+  -- be expected to reach an honest node at all — the same bound 'prop_leios'
+  -- uses for diffusion.
+  poisonedPointsToDiffuse =
+    Set.filter
+      (\p -> unSlotNo p.pointSlotNo + minCertificationGap <= numSlots)
+      poisonedPoints
+
+  acquiredByHonest = Set.fromList $ flip mapMaybe traces $ \case
+    FromNode nid (FromLeios (TraceLeiosBlockTxsAcquired point _age))
+      | nid /= adversary -> Just point
+    _ -> Nothing
+
+  -- Ground truth from the adopted chains rather than from vote traces: a
+  -- CertRB certifies the EB that its parent's header announced.
+  certifiedPoints = foldMap certifiedEbPoints nodeChains
+
+  nodeChains = Chain.toOldestFirst . nodeOutputFinalChain <$> testOutput.testOutputNodes
+
+  -- Guard against a vacuous pass: a poisoned EB must actually have reached an
+  -- honest node's closure, or nobody was ever in a position to vote on one.
+  adversaryWasEffective =
+    not (Set.null (Set.intersection poisonedPointsToDiffuse acquiredByHonest))
+      & counterexample "no honest node acquired a poisoned EB's closure"
+      & prettyCounterexampleList "poisoned EBs (diffusion required)" 120 poisonedPointsToDiffuse
+      & prettyCounterexampleList "acquired by honest nodes" 120 acquiredByHonest
+
+  poisonedNeverCertified =
+    Set.intersection poisonedPoints certifiedPoints === Set.empty
+      & counterexample "a poisoned EB was certified"
+      & prettyCounterexampleList "poisoned EBs" 120 poisonedPoints
+      & prettyCounterexampleList "certified EBs" 120 certifiedPoints
+
+  -- The EBs certified in this run that were not the adversary's. Empty means
+  -- the run proves nothing, which is a discard above rather than a failure.
+  honestCertifiedPoints = Set.difference certifiedPoints poisonedPoints
+
+  -- A poisoned closure cannot apply, so no node may ever report it validated.
+  -- Unlike 'poisonedNeverCertified' this holds even when the vote never got
+  -- close — it is the invariant, not the outcome.
+  poisonedNeverValidated =
+    Set.null (Set.intersection poisonedPoints validatedPoints)
+      & counterexample "an EB carrying the bogus tx was reported as validated"
+      & prettyCounterexampleList "validated EBs" 120 validatedPoints
+
+  -- Non-vacuity for the validation path itself: some honest EB must have gone
+  -- through it. Without this the property would also pass on a build where
+  -- voting never validates anything.
+  validationDidRun =
+    not (Set.null (Set.difference validatedPoints poisonedPoints))
+      & counterexample "no EB's transactions were ever validated"
+      & prettyCounterexampleList "validated EBs" 120 validatedPoints
+
+  validatedPoints = Set.fromList $ flip mapMaybe leiosTraces $ \case
+    TraceLeiosEbValidated{ebPoint} -> Just ebPoint
+    _ -> Nothing
+
+  leiosTraces = [ev | FromNode _ (FromLeios ev) <- traces]
+
+  -- How the poisoned EBs were actually turned down: on their transactions, or
+  -- earlier on one of the cheap gates. Only the former exercises validation,
+  -- and which one happens depends on chain-tip timing, so report it rather
+  -- than requiring it.
+  poisonedNotVotedReasons =
+    [ reasonLabel reason
+    | FromNode _ (FromLeios TraceLeiosNotVoted{ebPoint, reason}) <- traces
+    , Set.member ebPoint poisonedPoints
+    ]
+
+  reasonLabel = \case
+    EbTxsInvalid{} -> "EbTxsInvalid"
+    ChainTipDoesNotAnnounce -> "ChainTipDoesNotAnnounce"
+    TooLate -> "TooLate"
+    NotOnCommittee -> "NotOnCommittee"
+
+-- | The EB points certified by a chain's CertRBs. A CertRB carries no
+-- announcement of its own for the EB it certifies; the announcement lives on
+-- its parent's header, so walk the chain carrying the previous announcement —
+-- the same traversal 'sumChainTxBytes' uses to resolve closures.
+certifiedEbPoints :: [CardanoBlock StandardCrypto] -> Set.Set LeiosPoint
+certifiedEbPoints = go Nothing
+ where
+  go _ [] = Set.empty
+  go prevAnn (blk : rest) =
+    here <> go (fst <$> headerLeiosAnnouncement (getHeader blk)) rest
+   where
+    here = case (blockLeiosCert blk, prevAnn) of
+      (Just _, Just point) -> Set.singleton point
+      _ -> Set.empty
+
+-- * Misbehaving nodes
+
+-- | Make one node endorse a transaction that can never apply, by appending
+-- 'invalidEbTx' to the txs it draws for the EB it is about to announce.
+-- 'fbRbTxs' is untouched, so the announcing RB itself stays valid.
+endorseInvalidTx ::
+  Functor m =>
+  -- | Which node misbehaves; every other node is left alone.
+  CoreNodeId ->
+  CoreNodeId ->
+  TestNodeInitialization m (CardanoBlock StandardCrypto) ->
+  TestNodeInitialization m (CardanoBlock StandardCrypto)
+endorseInvalidTx adversary nid tni
+  | nid /= adversary = tni
+  | otherwise =
+      tni{tniBlockForging = map (mapMkBlockForging poison) <$> tniBlockForging tni}
+ where
+  poison bf =
+    bf
+      { forgeBlock = \args ->
+          forgeBlock bf args{fbEbTxs = fbEbTxs args <> [invalidEbTx]}
+      }
+
+-- | A Dijkstra transaction that spends an output which does not exist: its
+-- input names the id of a transaction that is never applied anywhere, so the
+-- UTxO rule must reject it with 'BadInputsUTxO'.
+--
+-- Deliberately a /state-dependent/ failure rather than a missing-witness one:
+-- a missing witness is a static check, which is exactly what the reapply path
+-- is allowed to skip, so a witness-only failure would not pin down the
+-- behaviour we care about. This tx is unsigned for the same reason — the
+-- signature is irrelevant to the check that has to catch it.
+--
+-- Forced, because 'ForgedLeiosEb' and the forge path run 'NoThunks'
+-- invariants over what they are handed.
+-- 'assumeValidatedClosureTx' is how the adversary lies: it stamps the tx as
+-- mempool-validated so the forge will endorse it, which is exactly the claim a
+-- voter is now supposed to disbelieve.
+invalidEbTx :: Validated (GenTx (CardanoBlock StandardCrypto))
+invalidEbTx = assumeValidatedClosureTx (GenTxDijkstra (mkShelleyTx (force tx)))
+ where
+  tx :: Tx TopTx DijkstraEra
+  tx = mkBasicTx (mkBasicTxBody & inputsTxBodyL .~ Set.singleton phantomInput)
+
+  phantomInput = TxIn (txIdTx phantomTx) (TxIx 0)
+
+  -- Never submitted, never forged: its id therefore never appears in any UTxO.
+  phantomTx :: Tx TopTx DijkstraEra
+  phantomTx = mkBasicTx mkBasicTxBody
+
+-- | Rewrap the 'BlockForging' a 'MkBlockForging' allocates.
+mapMkBlockForging ::
+  Functor m =>
+  (BlockForging m blk -> BlockForging m blk) ->
+  MkBlockForging m blk ->
+  MkBlockForging m blk
+mapMkBlockForging f (MkBlockForging alloc) = MkBlockForging (f <$> alloc)
+
 -- | Independently compute cumulative tx bytes by resolving each block in the
 -- chain (filling in EB closures from the LeiosDB via 'inlineLeiosClosure')
 -- and summing individual 'sizeTxF' values per transaction.
@@ -684,19 +909,39 @@ runThreadNet ::
   NodeJoinPlan ->
   (TestOutput (CardanoBlock StandardCrypto), ProtocolInfo (CardanoBlock StandardCrypto))
 runThreadNet initSeed numSlots numCoreNodes joinPlan =
+  runThreadNet' initSeed numSlots numCoreNodes joinPlan (\_nid -> id)
+
+-- | 'runThreadNet' with a per-node tweak of the node's initialization, for
+-- tests that need one node to misbehave. Tweaking the whole
+-- 'TestNodeInitialization' rather than just its 'BlockForging' means a test can
+-- reach the protocol info and the crucial txs too; see 'endorseInvalidTx'.
+runThreadNet' ::
+  Seed ->
+  NumSlots ->
+  NumCoreNodes ->
+  NodeJoinPlan ->
+  ( forall m.
+    Functor m =>
+    CoreNodeId ->
+    TestNodeInitialization m (CardanoBlock StandardCrypto) ->
+    TestNodeInitialization m (CardanoBlock StandardCrypto)
+  ) ->
+  (TestOutput (CardanoBlock StandardCrypto), ProtocolInfo (CardanoBlock StandardCrypto))
+runThreadNet' initSeed numSlots numCoreNodes joinPlan tweakNodeInit =
   ( runTestNetwork
       testConfig
       testConfigB
       TestConfigMB
-        { nodeInfo = \(CoreNodeId nid) -> do
+        { nodeInfo = \coreNodeId@(CoreNodeId nid) -> do
             fs <- SomeHasFS <$> Sim.simHasFS' MockFS.empty
             (protocolInfo, blockForging) <- protocolInfoCardano fs (cardanoProtocolParams nid)
-            pure
-              TestNodeInitialization
-                { tniProtocolInfo = protocolInfo
-                , tniCrucialTxs = []
-                , tniBlockForging = blockForging Tracer.nullTracer
-                }
+            pure $
+              tweakNodeInit coreNodeId $
+                TestNodeInitialization
+                  { tniProtocolInfo = protocolInfo
+                  , tniCrucialTxs = []
+                  , tniBlockForging = blockForging Tracer.nullTracer
+                  }
         , mkRekeyM = Nothing
         }
   , protocolInfo0
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -238,6 +238,8 @@ prop_leios seed =
         & counterexample "[failed] propCertifying"
     , propCertifyAndAnnounce
         & counterexample "[failed] propCertifyAndAnnounce"
+    , propClosuresValidate
+        & counterexample "[failed] propClosuresValidate"
     ]
  where
   numNodes = 3 :: Int
@@ -347,6 +349,29 @@ prop_leios seed =
       & counterexample "no block both certified and announced"
       & counterexample ("certifying block slots: " <> show certificateBlocks)
       & counterexample ("announced-and-certified slots: " <> show announcedAndCertifiedSlots)
+
+  -- In an honest net every acquired closure must apply against the announcing
+  -- RB's ledger state. A rejection here means the voting thread validates
+  -- against the wrong state or the wrong slot, not that anything was invalid.
+  propClosuresValidate =
+    null closureRejections
+      & counterexample "a voter rejected an honestly-produced EB's transactions"
+      & prettyCounterexampleList "rejections" 240 closureRejections
+      -- The reapply-vs-apply split: how much full validation the LeiosTxCache
+      -- spared us. A forger's own EB should be entirely reapplied.
+      & tabulate
+        "EB closure validation (reapplied/total)"
+        [show reapplied <> "/" <> show (reapplied + applied) | (reapplied, applied) <- validatedSplits]
+
+  closureRejections =
+    [ (nid, ebPoint, err)
+    | FromNode nid (FromLeios TraceLeiosNotVoted{ebPoint, reason = EbTxsInvalid err}) <- traces
+    ]
+
+  validatedSplits =
+    flip mapMaybe leiosTraces $ \case
+      TraceLeiosEbValidated{reapplied, applied} -> Just (reapplied, applied)
+      _ -> Nothing
 
   propCertifying =
     conjoin
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -599,7 +599,7 @@ sumChainTxBytes _topConfig _initLedger node = runSimOrThrow $ do
   foldChain leiosDb prevAnn !total (blk : rest) = do
     blk' <- case (blockLeiosCert blk, prevAnn) of
       (Just _, Just point) ->
-        inlineLeiosClosure blk
+        inlineLeiosClosure blk . map snd
           <$> resolveLeiosClosure leiosDb (pointEbHash point)
       _ -> pure blk
     let nextAnn = fst <$> headerLeiosAnnouncement (getHeader blk)
@@ -665,7 +665,7 @@ foldWithResolution leiosDb cfg blks initState =
         Nothing ->
           error "foldWithResolution: CertRB but no announcement on parent chain-dep state"
         Just (point, _) -> do
-          closureTxs <- resolveLeiosClosure leiosDb (pointEbHash point)
+          closureTxs <- map snd <$> resolveLeiosClosure leiosDb (pointEbHash point)
           let ls = ledgerState state
               lcfg = configLedger (getExtLedgerCfg cfg)
           case applyLeiosClosure lcfg closureTxs ls of
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -811,8 +811,13 @@ endorseInvalidTx adversary nid tni
   poison bf =
     bf
       { forgeBlock = \args ->
-          forgeBlock bf args{fbEbTxs = fbEbTxs args <> [invalidEbTx]}
+          forgeBlock bf args{fbEbTxs = fbEbTxs args <> [lie invalidEbTx]}
       }
+  -- The lie is told here and only here: 'assumeValidatedClosureTx' stamps the tx
+  -- as mempool-validated so the forge will endorse it, which is exactly the
+  -- claim a voter is now supposed to disbelieve. Applying it at the use site
+  -- keeps 'invalidEbTx' an ordinary, plainly invalid transaction.
+  lie = assumeValidatedClosureTx
 
 -- | A Dijkstra transaction that spends an output which does not exist: its
 -- input names the id of a transaction that is never applied anywhere, so the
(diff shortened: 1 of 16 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -810,13 +811,38 @@ class ResolveLeiosBlock blk where
...
+  -- (witnesses, scripts) — over a full 'LedgerSupportsMempool.applyTx'.
+  --
+  -- The token carries no evidence of its own; the cache's tag is the evidence,
+  -- and the ledger recomputes every state-dependent check from the state at
+  -- hand. Calling this on a tx the cache has /not/ tagged validated would skip
+  -- checks that were never run.
+  --
+  -- NOTE: the cache's tag does not record which protocol version the tx was
+  -- validated under, so an EB straddling a protocol-version change could skip
+  -- a static check whose rules changed. The mempool has the same exposure, and
```
nfrisby: > The mempool has the same exposure, and the cache only retains a short window of EBs.

Neither of those justifies tolerating this. In particular, a bad tx in the Mempool doesn't lead to a bad tx on chain (RB validation will reject the block). But voting for a bad tx in an EB does.

My initial instinct is that we should change the CertRB validity rules to be invalid if they're attempting to certify an EB from the previous era. (In other words, the first block of an era must not be a CertRB.)

ouroboros-consensus/test/consensus-test/Test/LeiosVoting.hs
```diff
@@ -0,0 +1,243 @@
...
+    leiosDbInsertEbPoint conn point (leiosEbBytesSize eb)
+    void $ leiosDbInsertEbBody conn point eb
+    void $ leiosDbInsertTxs conn [(txHashOf tx, txBytes tx) | tx <- txs]
+
+    cache <- newPureLeiosTxCache
+    void $ insertAnnouncement cache (pointSlotNo point) rbHash (pointEbHash point)
+    -- The fold over the not-yet-acquired txs is what a fetch would use to build
+    -- its request; here only the refcount bump matters, so it folds into ().
+    void $ insertBody cache (pointEbHash point) (serializeEbBody eb) () (\() _ _ _ -> ())
+    -- Mark them acquired, as a fetch would: only then can they be upgraded.
```
nfrisby: > only then can they be upgraded

I don't think that comment is true; in particular, sometimes we first "receive" a tx straight from the Mempool (on my latest dev branch, at least), in which case it goes from Nothing -> Just Applied.

ouroboros-consensus/test/consensus-test/Test/LeiosVoting.hs
```diff
@@ -0,0 +1,243 @@
...
+    void $ withLockedInsertUnappliedTx cache $ \w0 step ->
```
nfrisby, line 241: "It shouldn't matter", but you might want to randomly exclude some of the txs from this insert. The voting logic should work either way: lookup returning `Nothing` should be tolerated (and treated the same as a return value of `Just Unapplied`).

It's an unusual situation, but not impossible, I don't think; especially if the node _just_ restarted, maybe?

nfrisby (commented):
First pass, which excluded two modules: ThreadNet tests and LeiosVoting itself. (That's as far as I got before my dev code finished compiling :D)

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -331,6 +369,29 @@ prop_leios seed =
       & counterexample ("certifying block slots: " <> show certificateBlocks)
       & counterexample ("announced-and-certified slots: " <> show announcedAndCertifiedSlots)
 
+  -- In an honest net every acquired closure must apply against the announcing
+  -- RB's ledger state. A rejection here means the voting thread validates
```
nfrisby: > A rejection here means the voting thread validates against the wrong state or the wrong slot, not that anything was invalid.

Why? I think that assertion is plausible, but it'd help to have an explanation of why that's true.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -331,6 +369,29 @@ prop_leios seed =
...
+  -- against the wrong state or the wrong slot, not that anything was invalid.
+  propClosuresValidate =
+    null closureRejections
+      & counterexample "a voter rejected an honestly-produced EB's transactions"
+      & prettyCounterexampleList "rejections" 240 closureRejections
+      -- The reapply-vs-apply split: how much full validation the LeiosTxCache
+      -- spared us. A forger's own EB should be entirely reapplied.
```
nfrisby: > A forger's own EB should be entirely reapplied.

That's conceptually tempting but:

- it's not necessarily true in the implementation
- it doesn't matter (theirs is just one vote of many)

So I suggest we don't require/emphasize it.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -573,6 +634,214 @@ prop_leios_late_join seed =
...
+-- input names the id of a transaction that is never applied anywhere, so the
+-- UTxO rule must reject it with 'BadInputsUTxO'.
+--
+-- Deliberately a /state-dependent/ failure rather than a missing-witness one:
+-- a missing witness is a static check, which is exactly what the reapply path
+-- is allowed to skip, so a witness-only failure would not pin down the
+-- behaviour we care about. This tx is unsigned for the same reason — the
+-- signature is irrelevant to the check that has to catch it.
+--
+-- Forced, because 'ForgedLeiosEb' and the forge path run 'NoThunks'
```
nfrisby: > Forced, because 'ForgedLeiosEb' and the forge path run 'NoThunks' invariants over what they are handed.

This comment seems like a bit of a non-sequitor.

Maybe it's stale?

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -573,6 +634,214 @@ prop_leios_late_join seed =
...
+-- invariants over what they are handed.
+-- 'assumeValidatedClosureTx' is how the adversary lies: it stamps the tx as
+-- mempool-validated so the forge will endorse it, which is exactly the claim a
```
nfrisby: Maybe defer the `assumeValidatedClosureTx` wrapper to the occurrence within `endorseInvalidTx`, to emphasize that this lie is being told _only_ to the forge logic?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
 
--- | Number of slots after an EB's announcement before its voters are
--- allowed to cast a vote. Serves as the equivocation-detection window: if
--- a peer equivocates by announcing two different EBs on the same slot, we
--- want to observe the second announcement (and drop the vote) before
--- committing. Stub for '3 * L_hdr'.
-lHdrWaitSlots :: Word64
-lHdrWaitSlots = 3
+-- TODO: Fetching and validating an EB's closure should start when the closure
+-- arrives, streamed, rather than after this wait. Validation is linear in the
```
nfrisby: > streamed

Is this TODO intending to say the validation should happen _as_ the EB closure arrives?

Or merely that it should start precisely when the _whole_ EB closure arrives, instead of waiting for the beginning of the vote window?

They both seem "good", but the second is waaay easier.

And if the second was the only intention, then I suggest removing the "streamed" word from this comment.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+-- closure -- a full ~13.5k-tx closure takes ~1.5s even when every tx is a
+-- tx-cache hit -- so doing it inside the vote window spends the window on work
+-- that could have been done while waiting for it to open. A devnet run showed
+-- exactly that: as closures filled up, the time from votable to validated grew
+-- past the window and voting stopped entirely.
+
+-- | How long after its announcing slot /begins/ before an EB's voters may cast
+-- a vote. Serves as the equivocation-detection window: if a peer equivocates by
+-- announcing two different EBs on the same slot, we want to observe the second
+-- announcement (and drop the vote) before committing. Stub for '3 * L_hdr'.
```
nfrisby: I suggest rewording "committing" to "issuing our vote".

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+lHdrWait :: NominalDiffTime
```
nfrisby, line 140: `lHdrWait` and `lVoteWindow` should have TODOs about eventually being drawn from the ledger state instead of being constants.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
-lVoteWindowSlots = 4
+-- * Vote timers
+
+-- | A vote timer per acquired EB, armed at the wall-clock moment that EB's
+-- voting window opens.
+--
+-- There is no priority queue: the wall clock does the prioritising, since
+-- whichever timer fires first is by construction the EB whose window opens
+-- first.
+data VoteTimers m = VoteTimers
```
nfrisby, line 157: Another option might be to monitor `BlockchainTime` in `Ouroboros.Consensus.BlockchainTime.API`.

```
data BlockchainTime m = BlockchainTime
  { getCurrentSlot :: STM m CurrentSlot
  -- ^ Get current slot
  }
```

I'm not sure how much simpler that would. And it wouldn't support fractional values of `lHdrWait`, for example.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+                  voteIn = voteAt `diffRelTime` now
+              traceWith tracer $
+                TraceLeiosVoteScheduled
+                  { ebPoint = point
+                  , voteIn
+                  , deadlineIn = voteDeadline `diffRelTime` now
+                  }
+              timer <-
+                registerDelay
+                  . max 1
```
nfrisby: is a `0` delay problematic? I hadn't internalized that fact if so, so maybe a one-line comment explaining the benefit of `max 1`?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+                  . diffTimeToMicrosecondsAsInt
+                  $ nominalDelay voteIn
+              atomically . modifyTVar pendingVotes . Map.insert point $
+                (timer, voteDeadline)
+      , -- Reading every armed timer puts them all in this transaction's read
+        -- set, so it wakes on any of them, and 'Map.toList' is in point order,
+        -- so simultaneous firings break towards the earlier slot.
+        waitNextVoteTime = do
```
nfrisby, line 218: We prune one `pendingVotes` entry per call to `waitNextVoteTime`.

I haven't scrolled down yet, but I'm guessing we call that in a `forever`.

So: _as long as each iteration of that loop doesn't take "too long"_, we'll be pruning the `pendingVotes` state var promptly 👍  So it'll never have more than, say, 100 elements.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+            [] -> TVar.retry
+            (point, (_, deadline)) : _ -> do
+              modifyTVar pendingVotes (Map.delete point)
+              pure (point, deadline)
+      }
+
+-- | When a slot begins, in wall-clock terms.
+--
+-- Both gates are measured from here rather than counted in slots: the node's
+-- notion of the current slot only advances as blocks are adopted, so slot
```
nfrisby: > the node's notion of the current slot only advances as blocks are adopted,

That is definitely not true. "The current slot" is how the leadership check knows it's time to forge a block, for example---which is circular reasoning according to this comment's claim.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
nfrisby: Ah ha, now that I've scrolled down, I see it _used_ to be based on `BlockchainTime`.

I see one (false) comment about `BlockchainTime`, but otherwise I'm not seeing an explanation of why.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
nfrisby: The falsehood is also asserted in the commit message of 25e369f8b4dfba23d2cee64172a2c12525bb8b59

> They counted slots, and a node's notion of the current slot only advances as
blocks are adopted -- so the deadline was judged against a clock that stops
whenever the chain does.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
-                  error $ "runLeiosVoting: unexpected error on addVote: " <> show err
-
--- | Read the current wall-clock slot, retrying until it is known.
-knownSlot :: IOLike m => BlockchainTime m -> STM m SlotNo
-knownSlot btime =
-  getCurrentSlot btime >>= \case
-    CurrentSlot s -> pure s
-    CurrentSlotUnknown -> retry
+    withForkerAt VolatileTip $ \case
+      Nothing -> pure $ Left ChainTipDoesNotAnnounce
```
nfrisby: I'm pretty sure this branch should be unreachable (`VolatileTip` forker target never fails to open). So we're just throwing `ChainTipDoesNotAnnounce` as a dummy.

Deserves a comment, I think. (And maybe someday UTxO HD will use GADTs to ensure that `VolatileTip` _can't_ fail.)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+            hs = headerState extLs
+            readTables = roforkerReadTables (ledgerStateReadOnlyForker forker)
+        rbHash <-
+          tipAnnouncerFor @blk hs point ?>= ChainTipDoesNotAnnounce
+        seatId <-
+          (getLeiosCommittee ls >>= getLeiosSeatId vk) ?>= NotOnCommittee
+
+        lift (validateEbClosure lcfg leiosConn txCache readTables point ls) >>= \case
+          EbClosureInvalid err ->
+            -- XXX: Text in error
```
nfrisby: We tend to use FIXME or TODO instead of XXX

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+                traceWith tracer TraceLeiosVoteAcquired{vote}
+                -- Trace certification whenever the tally crosses
+                -- 'minCertificationThreshold'. May fire more than once per
+                -- point if subsequent votes also come in; consumers (e.g.
+                -- ThreadNet's 'propCertifying') dedupe.
+                case mCert of
+                  Just _ -> traceWith tracer TraceLeiosCertified{rbHash}
+                  Nothing -> pure ()
+              err ->
+                -- XXX: Make this a NotVoted error / trace
```
nfrisby: Prefer TODO or FIXME

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+          throwE TooLate
+
+        let vote = signLeiosVote sk seatId rbHash
+        lift $
+          addVote vote
+            >>= \case
+              Added weight mCert -> do
+                traceWith tracer TraceLeiosVoted{vote, weight}
+                traceWith tracer TraceLeiosVoteAcquired{vote}
+                -- Trace certification whenever the tally crosses
```
nfrisby: - I'd be surprised if the log-writer deduped. So that seems like we'd rather have `addVote` track exclude the cert/set a flag/etc when first emitting.
- This comment says it fires "whenever the tally crosses <the threshold>". Which can only happen once. But then it says it may fire again. So either it is already deduped or the "when it crosses" is the wrong wording (crossings only happen once, right? Unless concurrency is the only cause of dupes?).

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  -- 'Nothing' when chain-sel has already moved off it.
+  withForkerAt :: Target (Point blk) -> (Maybe (ReadOnlyForker' m blk) -> m a) -> m a
+  withForkerAt tgt k =
+    bracket
+      (ChainDB.openReadOnlyForkerAtPoint chainDB tgt)
+      (either (\_ -> pure ()) roforkerClose)
+      (either (\_ -> k Nothing) (k . Just))
+
+-- | The outcome of checking an EB's endorsed transactions.
+data EbClosureVerdict blk
```
nfrisby, line 393: Only `EbClosureValid`'s comment mentions the succesful txs being recorded in the cache. But that happens too for whatever prefix of the EB succeeded before `EbClosureInvalid`, right?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  | -- | A tx did not apply, so this EB must not be certified.
+    EbClosureInvalid !(ApplyTxErr blk)
+
+-- | Apply an EB's endorsed transactions to the announcing RB's ledger state,
+-- which is the state they will meet if the EB is ever certified.
+--
+-- Each tx is validated in full, except where the LeiosTxCache reports it
+-- already validated: then only the state-dependent checks re-run
+-- ('LedgerSupportsMempool.reapplyTx' rather than 'applyTx'). That is where the
+-- cache earns its keep — consecutive EBs overlap heavily, and a forger's own EB
```
nfrisby: > and a forger's own EB is entirely pre-validated by its mempool.

I'd back that emphasis off here too, for the same reasons as before (not necessarily, and it wouldn't matter---other node's votes matter more)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  m (EbClosureVerdict blk)
+validateEbClosure lcfg leiosConn txCache resolveValues point lsBase = do
+  -- Load txs from disk
+  closure <- resolveLeiosClosure leiosConn (pointEbHash point)
+  -- Resolve their input UTxOs
+  let keys = foldMap (getTransactionKeySets . snd) closure
+  values <- resolveValues keys
+  -- Determine which txs we can just reapply (the cache hits)
+  decided <- withLookupTx txCache $ \look -> mapM (decide look) closure
+  let st0 = applyMempoolDiffs values keys (applyChainTick OmitLedgerEvents lcfg slot lsBase)
```
nfrisby, line 436: Want to add  TODO about `applyMempoolDiffs` being the _temporary_ implementation until the Leios-Ledger interface is settled/available?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+          Right st' -> goValidate st' rest (reapplied + 1) applied
+      Left tx ->
+        case runExcept $ applyTx lcfg DoNotIntervene slot tx st of
+          Left err -> pure $ EbClosureInvalid err
+          Right (st', _vtx) -> do
+            recordValidated txh
+            goValidate (applyDiffs st st') rest reapplied (applied + 1)
+
+  recordValidated txh =
+    withLockedInsertAppliedTx txCache $ \w0 step -> step w0 txh ()
```
nfrisby: I'd recommend accruing these tx hashes in a list and then having a finalize loop that calls` withLockedInsertAppliedTx` just the once. No since in negotiating the LeiosTxCache's `MVar` thousands of times.

nfrisby (changes requested):
I'm Requesting Changes primarily for the commit message and comments that libel against `BlockchainTime` :P

I've also made some suggestions that seem quite preferable (eg only acquiring the LeiosTxCache log twice per EB instead of thousands of times, a few helpful explanatory code comments, etc)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
bladyjoker: Did we finally arrive to the dreaded topic of cross-era concerns x) Let's use this to dig a bit through this...

What does it mean to certify an EB that from the previous era? Which rules should be applied to that EB? Original era, or the new one?

My take is; We're never actually in the new era when certifying!

To cross an era we must `tick` right? But we `verifyLeiosCert` and `applyLeiosClosure` before ticking!

So in effect, we're always applying the rules to EBs and Certs of a single era (that they originated in). In other words, no such thing as cross-era EB/Cert.

Am I missing something?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1577,6 +1589,23 @@ traceLeiosKernelToObject = \case
       , "ebSlot" .= ebSlot
       , "reason" .= notVotedReasonText reason
       ]
+  TraceLeiosVoteScheduled{ebPoint = MkLeiosPoint (SlotNo ebSlot) ebHash, voteIn, deadlineIn} ->
+    mconcat
+      [ "kind" .= Aeson.String "LeiosVoteScheduled"
+      , "ebHash" .= prettyEbHash ebHash
+      , "ebSlot" .= ebSlot
```
bladyjoker, line 1640: Why are we still carrying the slot for? This should be RbHash.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1577,6 +1589,23 @@ traceLeiosKernelToObject = \case
...
+      , -- Aeson renders 'NominalDiffTime' as a number of seconds, exactly.
+        "voteIn" .= voteIn
+      , "deadlineIn" .= deadlineIn
+      ]
+  TraceLeiosEbValidated{ebPoint = MkLeiosPoint (SlotNo ebSlot) ebHash, reapplied, applied} ->
+    mconcat
+      [ "kind" .= Aeson.String "LeiosEbValidated"
+      , "ebHash" .= prettyEbHash ebHash
+      , "ebSlot" .= ebSlot
```
bladyjoker, line 1649: Same as before, this should be RbHash

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1577,6 +1589,23 @@ traceLeiosKernelToObject = \case
...
+      , "reapplied" .= reapplied
+      , "applied" .= applied
```
bladyjoker, line 1651: Very cool for tracing!

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
bladyjoker: Remember in case of forks that have the same EB announced at the same slot this becomes ambiguous.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,116 @@ import Ouroboros.Consensus.Util.IOLike
...
+  SystemTime m ->
+  m (VoteTimers m)
+newVoteTimers tracer lcfg chainDB systemTime = do
+  -- An armed timer, and the deadline it was armed against, per acquired EB.
+  pendingVotes <- atomically $ newTVar Map.empty
+  pure
+    VoteTimers
+      { scheduleVoteTime = \point -> do
+          extLedger <- atomically $ ChainDB.getCurrentLedger chainDB
+          case slotOnset lcfg (ledgerState extLedger) (pointSlotNo point) of
```
bladyjoker, line 186: This is bothering me, what's the relation between `extLedger` and `point`?

If votes are cast per announcement on our currently selected ledger state, then these two can differ here in a way that's suspicious.

For the sake of argument,
- `extLedger` contains the `PraosState` that has an associated `RbHash`, `Slot`  and `EbAnnouncement`.
- `point` has a `Slot` and `EbHash`

This method doesn't assure that current `extLedger` is carrying the `EbAnnouncement` for which we're voting.

In fact, if we keep it this way, if `extLedger` changed because our selection changed, we shouldn't proceed with voting right?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  (LedgerTables (LedgerState blk) KeysMK -> m (LedgerTables (LedgerState blk) ValuesMK)) ->
+  LeiosPoint ->
+  -- | The announcing RB's unticked ledger state, which the closure applies to.
+  LedgerState blk EmptyMK ->
+  m (EbClosureVerdict blk)
+validateEbClosure lcfg leiosConn txCache resolveValues point lsBase = do
+  -- Load txs from disk
+  closure <- resolveLeiosClosure leiosConn (pointEbHash point)
+  -- Resolve their input UTxOs
+  let keys = foldMap (getTransactionKeySets . snd) closure
```
bladyjoker, line 431: The `getBlockKeySets :: blk -> LedgerTables l KeysMK` of the `Ouroboros.Consensus.Ledger.Abstract:ApplyBlock` class is something we need to use here too.

We should introduce `ApplyEndorserBlock` class with a similar interface. Why?

The implementation of `getBlockKeySets` is giving you only the actual "inputs" of the block, rather than every input of every transactions. Remember, transactions can be chained, so in a case of the entire closure being just a chain of 1000 txs, you'd need to resolve the keys of the "first" transaction, not all of them.

Let's add a TODO for it!

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+
+-- | Apply an EB's endorsed transactions to the announcing RB's ledger state,
+-- which is the state they will meet if the EB is ever certified.
+--
+-- Each tx is validated in full, except where the LeiosTxCache reports it
+-- already validated: then only the state-dependent checks re-run
+-- ('LedgerSupportsMempool.reapplyTx' rather than 'applyTx'). That is where the
+-- cache earns its keep — consecutive EBs overlap heavily, and a forger's own EB
+-- is entirely pre-validated by its mempool.
+validateEbClosure ::
```
bladyjoker, line 408: Why are we not calling this `applyEbClosure`? For consistency, we should?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
bladyjoker: Yes I agree, I would have `goValidate` accumulate a `applied :: Set TxHash` that you then insert in batch.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  LedgerState blk EmptyMK ->
+  m (EbClosureVerdict blk)
+validateEbClosure lcfg leiosConn txCache resolveValues point lsBase = do
+  -- Load txs from disk
+  closure <- resolveLeiosClosure leiosConn (pointEbHash point)
+  -- Resolve their input UTxOs
+  let keys = foldMap (getTransactionKeySets . snd) closure
+  values <- resolveValues keys
+  -- Determine which txs we can just reapply (the cache hits)
+  decided <- withLookupTx txCache $ \look -> mapM (decide look) closure
```
bladyjoker, line 434: I didn't look how this works, but I'd be surprised if this does batched lookups? This is ideally a `Set.intersection txCache closureTxHashes`.

Currently, we're `O(length closure * O(look))` which I assumes is at best `O(length closure * O(log (length txCache)))`.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+    LeiosDbConnection m ->
+    -- \| Our voting key, to find our committee seat and sign votes.
+    LeiosSigningKey ->
+    -- \| The leios point of the EB to vote on.
+    LeiosPoint ->
+    -- \| The moment after which a vote is too late.
+    RelativeTime ->
+    m (Either LeiosNotVotedReason ())
+  goVote leiosConn sk point deadline = do
     let vk = deriveVerKeyDSIGN sk
```
bladyjoker, line 318: Bind `vk` once outside and pass it as argument

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+              AcquiredEb{} -> waitAcquiredEbTxs
+              AcquiredEbTxs point -> pure point
+
+      VoteTimers{scheduleVoteTime, waitNextVoteTime} <-
+        newVoteTimers tracer lcfg chainDB systemTime
+
+      forever $
+        atomically ((Left <$> waitAcquiredEbTxs) <|> (Right <$> waitNextVoteTime))
+          >>= \case
+            Left point -> scheduleVoteTime point
```
bladyjoker: I'm not sure about this, why are we scheduling a vote for **every acquired EB**?

Let me take a stab at the logic of this Voting thread...

In my view, Voting thread should subscribe on new selections (just like the Mempool) and that should be the main driver.

It then proceeds to relate EB announcements and their status to the current selection

1. Check if there's an EB announcement for the current praos/ledger state
   2. If we continue using Praos headers for announcement then check the PraosState `praosStateLeiosAnnouncement`
   3. If we use the `LeiosNotify` for EB announcements, then look it up there
2. If we found an EB announcement, check it's status
   3. "waiting on EB body" -> " EB body acquired" -> "EB closure acquired"
3. Wait on the EB announcement status to reach "EB closure acquired"
4. (go)Vote if we're not passed the deadline

This entire procedure must be interruptible to avoid wasting resources:
1. Deadline is exceeded
2. New selection comes about

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+        lift (validateEbClosure lcfg leiosConn txCache readTables point ls) >>= \case
+          EbClosureInvalid err ->
+            -- XXX: Text in error
+            throwE . EbTxsInvalid . Text.pack $ show err
+          EbClosureValid reapplied applied ->
+            lift $ traceWith tracer TraceLeiosEbValidated{ebPoint = point, reapplied, applied}
+
+        now <- lift $ systemTimeCurrent systemTime
+        when (now > deadline) $
+          throwE TooLate
```
bladyjoker: I'd rather see a thread cancelled due to a timeout then this check.

Why? Because this check could be added at any point in this function, actually, it makes sense to add it everywhere to stop ASAP. That just tells me that we should use `timeout` and cancel thread execution when the deadline is hit.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +238,198 @@ runLeiosVoting ::
...
+  -- NOTE: 'assumeValidatedClosureTx' sits on the lookup that licenses it: the
+  -- tag is the evidence that this tx was validated before, so the token is
+  -- built here and nowhere else.
+  decide look (txh, tx) =
+    look txh >>= \case
+      Just Right{} -> pure (txh, Right (assumeValidatedClosureTx tx))
+      _ -> pure (txh, Left tx)
+
+  -- Apply or reapply txs and tag the former as validated in the cache
+  goValidate _ [] !reapplied !applied =
```
bladyjoker: for future: This is currently counting applied vs reapplied, imagine if we could accumulate actual TxMeasure and we can observe how much bytes/xu/etc the cache actually provided.

bladyjoker (commented):
Heyo,

I'd say the most salient comment is about the the model of Voting thread. 

One point I think holds pretty well is that it should be driven by "selection change" events, just like the Mempool (not on AcquiredEbTxs).

The thread then waits for the status of the associated EB announcement to be "closure acquired" and proceeds to vote. This part should be cancellable due to either deadline or selection change.

I think this even removes the need for timers maybe? Could be simplified?

Another missing part is that we probably don't have the `ebAnnouncementStatusVar :: TVar (Map LeiosPoint AnnouncementStatus)` just laying around. However, we do have all the events we need to make that happen in `subscribeEbNotifications`, a thread that monitors these notifications and maintains the `ebAnnouncementStatusVar` var sounds doable.

Wdyt?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
bladyjoker: After discussing in Slack, we do `tickThenApply` that invokes `BBODY` Ledger rule AFTER TICK which means in the new ERA. I find that to be an implementation problem, and we should instead:

1. `verifyLeiosCert` and invoke a special Ledger `BCertBody`
2. `applyLeiosClosure`
3. `tick`

ch1bo: I, of course, have only worked and reviewed the code .. not the comments that the :genie: produced while we paired. I do agree with that we don't want to allow certification across era boundaries. I would even say we must not certify across epoch boundaries -> so protocol parameters can be assumed static too.

ch1bo: @bladyjoker not taking the bait because I truly believe it's not worth it to worry about cross era / epoch certification at this point.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Both should be, but I should flip the wording or split into two TODOs. The streaming is more of an idea, while the validation before vote period is a low hanging fruit.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: This would mean we need to convert back from slot to wall-clock time in the handler, which can fail because of past horizon. Thus, I do prefer to forecast the time only once from the announcing slot onset.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: I should have been more precise. If I ask the `ChainDb` for the current chain tip / slot, it will give me the volatile tip. Which corresponds to the slot of the latest adopted block, which only ever advances if we extend it.

EDIT: I was having a wrong memory of using the slot of the volatile tip here, but we had `BlockChainTime.getCurrentSlot` so that was actually fine. Still, we need diff times to support fractional L_hdr periods

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: I swear I had a comment about it at one point and threw it out :D

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: So there is nothing less serious than a TODO?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: The `mCert` is only set when we first certify. I don't understand your comment / what are you wanting me to change?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Yes, will update

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Can do

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Good point. Had a update all at once and that was incorrect. This is correct now, but inefficient. I'll change it

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: Do you want me to drop the LeiosPoint / eb tracing or add the RbHash or do both?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: This is even feeding a grafana panel

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: This is not a `LeiosPoint`, but a `Point`. We do check the announcement when it's time to vote - selection could change in between! See the `ChainTipDoesNotAnnounce` case

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Good point about the chained inputs. Not buying the need for a class yet, is it because we don't have the notion of "spending inputs" / matching these "keys" with tx inputs because of the polymorphic `GenTx blk`? I'll add a TODO anyways

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: That would be confusing as this function actually might call `applyTx` or `reapplyTx`. `validate` is a fine verb and explains what happens here .. it's even in the PR title :)

ch1bo: In fact, we should also validate the EB itself.. that is, whether the hashes and sizes match? :thinking:

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: I'm using what @nfrisby provides :)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: No. That would then be a "bigger than necessary signature" and the two values could drift -> I would need to check whether the vk matches the sk

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: There is multiple options in how to structure the voting loop. I originally picked the form of being driven by EB announcements / their acquisition and only check whether it matches our selection *at the time of voting* (this is important as it can change while we wait for the voting period). I didn't change it in this PR (only refactored the semantics). Feel free to rewrite to the semantics you suggest and see if it is more clear. I certainly won't do it in this PR.

> This entire procedure must be interruptible to avoid wasting resources:

I agree that this is the challenge it's not done well right now

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Yeah the current logic is more synchronous than it needs to be. The same is true for the forge loop and other aspects of the system. Synchronous is often easier to write and read :)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: > we can observe how much bytes/xu/etc the cache actually provided.

You mean how much fetching or cpu time validating scripts we saved?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +249,218 @@ runLeiosVoting ::
...
+    case decision of
+      Right vtx ->
+        -- Already validated once, so only the state-dependent checks re-run. A
+        -- failure here is state-dependent (a spent input, say) and says nothing
+        -- about the static checks, so the tx keeps the tag it already has.
+        case runExcept $ reapplyTx lcfg slot vtx st of
+          Left err -> abandon newly err
+          Right st' -> goValidate st' rest (reapplied + 1) applied newly
+      Left tx ->
+        case runExcept $ applyTx lcfg DoNotIntervene slot tx st of
```
dnadales: Should we use `Intervene` here rather than `DoNotIntervene`?

I might be misreading, but I think `DoNotIntervene` makes `applyTx` rewrite the tx's `isValid` flag before applying it. In `Shelley/Eras.hs:249-250`:

```haskell
  intervenedTx = case wti of
    DoNotIntervene -> tx & Core.isValidTxL .~ Alonzo.IsValid True
```

and on a `ValidationTagMismatch` it retries with `IsValid False` (`Eras.hs:270`). So a Plutus tx that declares `IsValid True` but whose scripts fail wouldn't give us `EbClosureInvalid`: it gets re-tagged, applied as collateral-only, and we vote for the EB.

What worries me is what the apply path then does with the *original* bytes: `SL.ruleApplyTxValidation @"LEDGER" STS.ValidateNone` (`Shelley/Ledger/Leios.hs:206-213`). Script evaluation sits inside `when2Phase`, which is `lblStatic` (`Alonzo/Rules/Utxos.hs:420-421`), so it's skipped, and the UTxO effect branches on the *declared* flag inside the transition body rather than in a predicate (`Conway/Rules/Utxos.hs:218-224`, `Babbage/Rules/Utxo.hs:463-479`). If I've followed that correctly, the tx we validate here and the tx the chain applies are not the same one, which seems to undercut the "the certificate licenses the trust" argument we just wrote into `applyLeiosClosure`'s docs.

Ingest only checks that the bytes match their own hash (`LeiosDemoLogic.hs:1242-1248`), so I don't think anything upstream would catch it either.

`Intervene` leaves the flag alone and lets the mismatch escape as an error, which would make us fail closed on exactly the txs the apply path would mis-handle. Would that work here, or is there a reason we want the mempool's lenient behaviour on this path?

Semi-related, and only if the above holds: we `recordValidated` the hash of the original bytes at line 449, so a later EB carrying the same tx would take the `reapplyTx` branch. Might be worth not tagging a tx whose flag we had to rewrite?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +249,218 @@ runLeiosVoting ::
...
+                traceWith tracer TraceLeiosVoted{vote, weight}
+                traceWith tracer TraceLeiosVoteAcquired{vote}
+                -- Trace certification whenever the tally crosses
+                -- 'minCertificationThreshold'. May fire more than once per
+                -- point if subsequent votes also come in; consumers (e.g.
+                -- ThreadNet's 'propCertifying') dedupe.
+                case mCert of
+                  Just _ -> traceWith tracer TraceLeiosCertified{rbHash}
+                  Nothing -> pure ()
+              err ->
```
dnadales: Should we do this TODO now rather than leave it?

I think this PR makes it reachable in a way it wasn't before. We resolve `seatId` from the forker's ledger state up at line 329, but `addVote` re-reads the committee from somewhere else: `getLeiosCommittee . ledgerState <$> ChainDB.getCurrentLedger chainDB` (`NodeKernel.hs:791-793`). Those two reads used to be microseconds apart. Now `validateEbClosure` sits between them, which the header comment puts at ~1.5s for a full closure.

If the selection crosses an epoch boundary in that gap, the committee changes (it comes from `nesPd`, so epoch boundaries only), and `LeiosSeatId` is a positional index, so our old seat resolves to a different pool's key. That lands on `VoteInvalid` here, and since we're under `forkLinkedThread` the `error` takes the node down. A rollback across the Dijkstra hard fork would give us `NoCommittee` the same way.

What makes me read this as an oversight rather than an invariant: the inbound path already treats a non-`Added` result as ordinary (`NodeToNode.hs:566-577`, `_ -> pure ()`).

Would a new `LeiosNotVotedReason` work here, so all three constructors become a trace instead? Something like:

```haskell
        ExceptT $ addVote vote >>= \case
          Added weight mCert -> Right () <$ do
            traceWith tracer TraceLeiosVoted{vote, weight}
            ...
          err -> pure . Left . VoteRejected . Text.pack $ show err
```

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +249,218 @@ runLeiosVoting ::
...
+        extLs <- lift $ atomically $ roforkerGetLedgerState forker
+        let ls = ledgerState extLs
+            hs = headerState extLs
+            readTables = roforkerReadTables (ledgerStateReadOnlyForker forker)
+        rbHash <-
+          tipAnnouncerFor @blk hs point ?>= ChainTipDoesNotAnnounce
+        seatId <-
+          (getLeiosCommittee ls >>= getLeiosSeatId vk) ?>= NotOnCommittee
+
+        lift (validateEbClosure lcfg leiosConn txCache readTables point ls) >>= \case
```
dnadales: Should we also check the deadline *before* we validate, not only after?

I'm not arguing with the comment above `goVote`: judging the final decision by the clock as it reads at signing time makes sense. This would be an extra, earlier check that leaves that one exactly as it is.

The case I have in mind doesn't need a backlog to happen. `scheduleVoteTime` arms a timer for any `AcquiredEbTxs` point with no deadline filter, and `AcquiredEbTxs` only fires when the *last* tx of a closure lands, which can be well after the EB's slot. So an EB whose closure completes late gets a negative `voteIn`, fires straight away, and we spend the whole ~1.5s applying a closure for a vote that was already dead before the timer existed.

The backlog case compounds it: the `registerDelay` TVars latch, and `Map.toList` hands us the lowest slot first, so after one long `goVote` we get the *most* expired entry next.

I did wonder whether validating anyway pays for itself by warming the tx cache. But the header comment says ~1.5s "even when every tx is a tx-cache hit", so I don't think it buys the window back.

Would something like this at the top of `goVote`, before we even open the forker, be enough?

```haskell
  goVote leiosConn sk point deadline = do
    if now > deadline then pure (Left TooLate) else do
      ...
```

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +249,218 @@ runLeiosVoting ::
...
+      let waitAcquiredEbTxs =
+            readTChan chan >>= \case
+              AcquiredEb{} -> waitAcquiredEbTxs
+              AcquiredEbTxs point -> pure point
+
+      VoteTimers{scheduleVoteTime, waitNextVoteTime} <-
+        newVoteTimers tracer lcfg chainDB systemTime
+
+      forever $
+        atomically ((Left <$> waitAcquiredEbTxs) <|> (Right <$> waitNextVoteTime))
```
dnadales: Did we mean to swap the branches here?

The old loop had voting on the left:

```haskell
    -- 'orElse' gives priority to voting over ingesting,
    -- so we can't stall a due vote behind a chan drain.
    (Just <$> takeReady) `orElse` (Nothing <$ takeAcquisition)
```

`<|>` is `orElse` for both instantiations we use (`base GHC/Conc/Sync.hs:739-741` for `IO`, io-sim `Types.hs:321-323` for `IOSim`), so with `waitAcquiredEbTxs` on the left, ingest is now the preferred branch.

I don't think it's serious on its own: one left win is a single `scheduleVoteTime`, and its slow parts sit outside the transaction, so a due vote is delayed by microseconds against a 4s window. What bothers me more is that the comment recording the intent went away with the old code, so there's nothing left to say whether the swap was deliberate.

Would it be worth putting `waitNextVoteTime` back on the left and keeping a version of that comment? Something like:

```haskell
        -- 'orElse' prefers the left branch, so a due vote is never
        -- stalled behind a chan drain.
        atomically ((Right <$> waitNextVoteTime) <|> (Left <$> waitAcquiredEbTxs))
```

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -101,102 +249,218 @@ runLeiosVoting ::
...
     traceWith tracer $
       MkTraceLeiosKernel
         "runLeiosVoting: disabled because no topLevelConfigVotingKey"
-  Just sk -> do
+  Just sk ->
+    -- A 'LeiosDbConnection' is not thread-safe, so this thread owns one for its
+    -- lifetime, the way each forge-credentials thread does.
+    withLeiosDb leiosDB $ \leiosConn -> do
+      chan <- subscribeEbNotifications leiosDB
+      let waitAcquiredEbTxs =
```
dnadales: Should `waitAcquiredEbTxs` consume the `AcquiredEb` messages too, rather than recursing past them?

The old `takeAcquisition` handled both constructors without looping:

```haskell
      takeAcquisition :: STM m ()
      takeAcquisition =
        readTChan chan >>= \case
          AcquiredEb{} -> pure ()
          AcquiredEbTxs point -> modifyTVar pendingVar (Set.insert point)
```

`readTChan` advances the channel's read end by writing a TVar, and that write only sticks if the transaction commits. Both branches above were terminal, so every read was followed by a commit. Here the `AcquiredEb{}` branch loops instead, so a run that ends in `retry` discards the read-end write, and the message stays in the channel for the next wake to read again.

I think it's mostly harmless in practice: the channel is FIFO, so the next `AcquiredEbTxs` from *any* EB flushes whatever is queued ahead of it, and the residue is only what arrived since the last closure completed. It grows without bound only if we stop completing closures altogether, and the channel is an unbounded `newBroadcastTChan` (`LeiosDemoDb/SQLite.hs:91,95`), so nothing pushes back.

Would returning a `Maybe` and letting the caller ignore `Nothing` keep the old draining behaviour? Something like:

```haskell
      let takeEbNotification :: STM m (Maybe LeiosPoint)
          takeEbNotification =
            readTChan chan >>= \case
              AcquiredEb{} -> pure Nothing
              AcquiredEbTxs point -> pure (Just point)
```

with `Left mPoint -> mapM_ scheduleVoteTime mPoint` at the call site.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -76,19 +116,127 @@ import Ouroboros.Consensus.Util.IOLike
...
-lVoteWindowSlots = 4
+-- | A vote timer per acquired EB, armed at the wall-clock moment that EB's
+-- voting window opens.
+--
+-- There is no priority queue: the wall clock does the prioritising, since
+-- whichever timer fires first is by construction the EB whose window opens
+-- first.
+data VoteTimers m = VoteTimers
+  { scheduleVoteTime :: LeiosPoint -> m ()
+  -- ^ Arm a timer for a newly acquired EB. One whose window is already open
```
dnadales, line 159: "clamps"?

I think this is describing the `max 1` that came out in the last commit. The behaviour still looks right to me: `registerDelay` fires immediately on a non-positive delay in both monads we use. `GHC/Event/TimerManager.hs:225` has `registerTimeout mgr us cb | us <= 0 = cb`, and io-sim's `IOSim/Internal.hs:371-380` hands back a TVar that is already `True` for `d < 0`. So dropping `max 1` seems fine.

It's more that we now lean on `registerDelay` for it rather than doing it ourselves, and the doc still says we do it. Should we say what actually happens? Something like:

```haskell
  -- ^ Arm a timer for a newly acquired EB. One whose window is already open
  -- gives a non-positive delay, which 'registerDelay' fires immediately, so
  -- "already open" is not a separate case for the caller.
```

ouroboros-consensus/test/consensus-test/Test/LeiosVoting.hs
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+validateOnce :: Harness -> [TestTx] -> IO (Verdict, [Bool])
+validateOnce Harness{hConn, hCache, hPoint} txs = do
+  verdict <-
+    validateEbClosure
+      testLedgerConfigNoSizeLimits
+      hConn
+      hCache
+      (\_keys -> pure (projectLedgerTables testInitLedger))
```
dnadales: Should the harness assert something about `keys` rather than ignore it?

As written we hand back the whole initial UTxO whatever was asked for, and `applyMempoolDiffs` restricts the diffs rather than the values (`Ledger/Tables/Utils.hs:198`), so `values` passes through intact. If I've followed that right, `foldMap (getTransactionKeySets . snd) closure` at `LeiosVoting.hs:409` could return `mempty` and all four properties would still be green.

Would restricting the reply to what was asked for be enough? Something like:

```haskell
      (\keys -> pure (restrictValues' testInitLedger keys))
```

That would also give the transitive-closure TODO at `LeiosVoting.hs:406-408` something to fail against once we get to it.

Separately, and only as an observation: the mock's `reapplyTx` is `applyTx` underneath (`Mock/Ledger/Block.hs:642-644`), so `prop_secondLookReapplies` pins the counters but can't show any check is actually skipped. I don't think that's fixable at this level given what the module header says its scope is, but it might be worth saying so there.

ouroboros-consensus/test/consensus-test/Test/LeiosVoting.hs
```diff
@@ -0,0 +1,260 @@
...
+tests :: TestTree
+tests =
+  testGroup
+    "LeiosVoting"
+    [ testGroup
+        "validateEbClosure"
+        [ testProperty "a fresh closure is validated in full" prop_freshClosureFullyApplied
+        , testProperty "a second look reapplies instead" prop_secondLookReapplies
+        , testProperty "records the txs validated before a failure" prop_recordsValidatedBeforeFailure
+        , testProperty "a tx after the failure is not recorded" prop_stopsAtFirstFailure
```
dnadales, line 71: Is the label describing a different property from the one below?

`forAllValidThenInvalid` puts the bad tx last (`good <> [bad]`), so there's never a tx *after* the failure, and `abandon` skipping the rest of the closure isn't exercised. The doc comment on `prop_stopsAtFirstFailure` itself reads right to me ("The failing tx itself is not recorded"), so I think it's just the tasty label that drifted.

Would "the failing tx is not recorded" be closer? And if we do want the skipping covered, would it be worth a `good <> [bad] <> after` variant asserting the tail is untagged too? Something like:

```haskell
    (_, tagged) <- runValidate acquired (good <> [bad] <> after)
    pure $ drop (length good) tagged === map (const False) (bad : after)
```

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1372,9 +1373,15 @@ data TraceLeiosKernel
   | TraceLeiosVoted {vote :: LeiosVote, weight :: Weight}
   | TraceLeiosVoteAcquired {vote :: LeiosVote}
   | TraceLeiosCertified {rbHash :: RbHash}
+  | -- \| A vote is scheduled to happen.
```
dnadales: `-- \|` here rather than `-- |`, so Haddock will not pick this up as the constructor's doc. Every other constructor in the type uses the unescaped form.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
```diff
@@ -311,6 +329,26 @@ prop_leios seed =
     TraceLeiosVoted{vote} -> Just vote
     _ -> Nothing
 
+  -- Which EB slot each announcing RB hash announced, so that a vote can be
+  -- dated: 'LeiosVote' carries only the hash it signed.
+  -- Which EB each announcing RB hash announced, so that a vote can be dated:
+  -- 'LeiosVote' carries only the hash it signed.
```
dnadales: Looks like these two got left in together. The second one reads more accurately to me, since we're keeping a `LeiosPoint` rather than a slot. Drop the first?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -14,60 +15,99 @@ module LeiosVoting
```
dnadales, line 15: (This is really about the export list just above, `LeiosVoting.hs:11-14`, which isn't in the diff so I can't anchor there.)

Should we name what we want to export?

`module LeiosVoting` publishes everything, and the module is in `exposed-modules` (`ouroboros-consensus.cabal:123`), so this PR newly makes `VoteTimers(..)`, `newVoteTimers`, `slotOnset`, and the `(?>=)` helper at line 489 part of the library's API. `validateEbClosure` and `EbClosureVerdict(..)` we do want, since `Test.LeiosVoting` imports them.

The wildcard predates this PR, so this isn't really something the PR introduced. But `(?>=)` is a very general name to be handing out from a module about voting, and it's new here. Would an explicit list be worth doing while we're in the file? Something like:

```haskell
module LeiosVoting
  ( runLeiosVoting
  , validateEbClosure
  , EbClosureVerdict (..)
  , HasLeiosVoting (..)
  ) where
```

dnadales (commented):
Full disclosure, these concerns were raised by Claude. I filtered them, and I'm posting here the ones that seemed to make sense.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Good catch! Will look into this Intervene thing. I am quite surprised that we have this option to ignore the flag (but that flag is being dropped from the mempool tx rep as we speak)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Does not matter and having Left on the right would be confusing :P

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: Good catch, this leads to more retries than necessary.

ouroboros-consensus/test/consensus-test/Test/LeiosVoting.hs
ch1bo: Yes, the generator is wrong/incomplete

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
ch1bo: I don't really like explicit export lists for applications (We are implementing the voting logic for a node here, not a generally reusable library)

Is this a Must, Should or Could comment?

PR #2237, opened 2026-08-25
Rewrite the LeiosFetch decision logic

### LINKED ISSUE #1021: Prototype the TxCache to avoid fetches and revalidation

nfrisby:
PR #2188 should merge before this one; this PR's branch extends that PR's branch.

TODO better description, link to correct Issue, etc

Major differences:

- does not avoid redundant requests (eg we _immediately_ send _full_ hedge requests)
- limits outstanding bytes for normal peers, but ~60 MB allowed for each BigLedgerPeer
- pulls from LeiosTxCache
- pulls from Mempool
- two-tier EB prioritization (FreshestFirst except *within* the EBs that are younger than L=3L_hdr+L_vote+L_diff those young EBs are prioritized FreshestLast)
- each FetchLogic iteration only reads in-memory data structures to decide → bounded decision latency
- those in-memory data structures are well-pruned → bounded memory usage
- randomizes the order in which it fetches still-missing txs → when its limits are saturated, they won't all be saturated on the same requests
- on restart, it'll refetch *all of* each EB it hadn't _completely_ acquired before shutting down


ch1bo (approved):
I have not read this code! However, I was using the API and running this a lot while working on #2235, including several night long runs and stress tests. I might be able to review this today, but would also be comfortable with merging to the `leios-prototype` as-is.

PR #2239, opened 2026-08-26
[Peras 64] Refactor BlockSupportsPeras type class

agustinmista:
This PR refactors `BlockSupportsPeras` to use epoch-dependent contexts to forge and verify votes and certificates. This includes:

- Refactoring `BlockSupportsPeras` to replace static `PerasParams` with `PerasEpochContext`s when forging and verifying objects.
- Adding resolver-handle helpers and propagating Peras epoch-context resolution through the `ChainDB`, `PerasVoteDB`, vote aggregation, and object-diffusion.
- Updating errors, constraints, and storage interfaces to account for the expanded requirements.
- Adapt `ChainDB` and `VolatileDB` state-machine tests to:
  - generate valid mock certificates, 
  - avoid Peras actions at origin, and 
  - constrain generated blocks and actions to resolvable context ranges.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
@@ -1,5 +1,4 @@
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE FlexibleContexts #-}
 
 -- | Instantiate 'ObjectPoolReader' and 'ObjectPoolWriter' using Peras
 -- certificates from the 'PerasCertDB' (or the 'ChainDB' which is wrapping the
@@ -11,20 +10,24 @@ module Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.PerasCert
   , makePerasCertPoolWriterFromChainDB
   ) where
 
-import Control.Monad (join)
-import Data.Either (partitionEithers)
-import Data.Functor (void)
+import Data.Foldable (traverse_)
 import Data.Map (Map)
 import qualified Data.Map as Map
-import Data.Set (Set)
 import qualified Data.Set as Set
-import GHC.Exception (throw)
+import Data.Typeable (Typeable)
 import Ouroboros.Consensus.Block
 import Ouroboros.Consensus.BlockchainTime.WallClock.Types
   ( SystemTime (..)
   , WithArrivalTime (..)
   )
 import Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.API
+  ( ObjectPoolReader (..)
+  , ObjectPoolWriter (..)
+  )
+import Ouroboros.Consensus.Peras.Context
+  ( PerasEpochContextResolverHandle
+  , verifyPerasCertWithHandle
+  )
 import Ouroboros.Consensus.Storage.ChainDB.API (ChainDB)
 import qualified Ouroboros.Consensus.Storage.ChainDB.API as ChainDB
 import Ouroboros.Consensus.Storage.PerasCertDB.API
@@ -32,7 +35,7 @@ import Ouroboros.Consensus.Storage.PerasCertDB.API
   , PerasCertTicketNo
   )
 import qualified Ouroboros.Consensus.Storage.PerasCertDB.API as PerasCertDB
-import Ouroboros.Consensus.Util.IOLike
+import Ouroboros.Consensus.Util.IOLike (IOLike, MonadSTM (..))
 
 -- | TODO: replace by `Data.Map.take` as soon as we move to GHC 9.8
 takeAscMap :: Int -> Map k v -> Map k v
@@ -89,20 +92,26 @@ makePerasCertPoolReaderFromChainDB chainDB =
 -- see 'makePerasCertPoolWriterFromChainDB' which creates a pool writer from the
 -- 'ChainDB' with proper handling of chain selection side-effects.
 makePerasCertPoolWriterFromCertDB ::
-  (StandardHash blk, IOLike m) =>
+  (StandardHash blk, Typeable blk, IOLike m) =>
   SystemTime m ->
   PerasCertDB m blk ->
+  PerasEpochContextResolverHandle m blk ->
   ObjectPoolWriter PerasRoundNo (PerasCert blk) m
-makePerasCertPoolWriterFromCertDB systemTime perasCertDB =
+makePerasCertPoolWriterFromCertDB systemTime perasCertDB resolverHandle =
   ObjectPoolWriter
     { opwObjectId = getPerasCertRound
-    , opwAddObjects = \certs ->
-        processCerts
-          systemTime
-          (PerasCertDB.getCertIds perasCertDB)
-          (validatePerasCert defaultPerasParams) -- TODO replace when actual plumbing is in place
-          (void . join . atomically . PerasCertDB.addCert perasCertDB)
-          certs
+    , opwAddObjects = \certs -> do
+        now <- systemTimeCurrent systemTime
+        atomically $ do
+          alreadyInDb <- PerasCertDB.getCertIds perasCertDB
+          let certsNotAlreadyInDb = filter ((`Set.notMember` alreadyInDb) . getPerasCertRound) certs
+          validatedCerts <- traverse (verifyPerasCertWithHandle resolverHandle) certsNotAlreadyInDb
+          -- Some certs are invalid => reject the whole batch
+          --
+          -- NOTE: we could combine the two 'traverse' operations into one in
+          -- which case any validated cert would be immediately added no matter
+          -- what is the validity of the other certs in the batch.
+          traverse_ (PerasCertDB.addCert perasCertDB . WithArrivalTime now) validatedCerts
     , opwHasObject = do
         certIds <- PerasCertDB.getCertIds perasCertDB
         pure $ \roundNo -> Set.member roundNo certIds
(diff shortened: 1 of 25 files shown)


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -11,28 +10,32 @@ module Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.ObjectPool.PerasCert
...
-import Control.Monad (join)
-import Data.Either (partitionEithers)
-import Data.Functor (void)
+import Data.Foldable (traverse_)
 import Data.Map (Map)
 import qualified Data.Map as Map
-import Data.Set (Set)
 import qualified Data.Set as Set
-import GHC.Exception (throw)
+import Data.Typeable (Typeable)
```
jasagredo, line 17: I in general don't like Typeable because it spreads through the whole codebase. But I won't oppose to it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -89,20 +92,25 @@ makePerasCertPoolReaderFromChainDB chainDB =
...
+    , opwAddObjects = \certs -> do
+        now <- systemTimeCurrent systemTime
+        atomically $ do
+          alreadyInDb <- PerasCertDB.getCertIds perasCertDB
+          let certsNotAlreadyInDb = filter ((`Set.notMember` alreadyInDb) . getPerasCertRound) certs
+          validatedCerts <- traverse (verifyPerasCertWithHandle resolverHandle) certsNotAlreadyInDb
+          -- Some certs are invalid => reject the whole batch
+          -- We could combine the two 'traverse' operations into one in which case
+          -- any validated cert would be immediately added no matter what is the
+          -- validity of the other certs in the batch.
```
jasagredo, line 112: We could, and that would be a difference in the logic of this function. Which of the two is the correct logic? what we do now or what we could do?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
```diff
@@ -89,20 +92,25 @@ makePerasCertPoolReaderFromChainDB chainDB =
...
 -- 'ChainDB' with proper handling of chain selection side-effects.
 makePerasCertPoolWriterFromCertDB ::
-  (StandardHash blk, IOLike m) =>
+  (StandardHash blk, Typeable blk, IOLike m) =>
   SystemTime m ->
   PerasCertDB m blk ->
+  PerasEpochContextResolverHandle m blk ->
   ObjectPoolWriter PerasRoundNo (PerasCert blk) m
-makePerasCertPoolWriterFromCertDB systemTime perasCertDB =
+makePerasCertPoolWriterFromCertDB systemTime perasCertDB resolverHandle =
```
jasagredo, line 100: Maybe I overlooked at this, but when would we make a PerasCertPoolWriter from a CertDB? I thought the CertDB was part of the ChainDB so only the function below would be used.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
```diff
@@ -90,27 +92,25 @@ makePerasVotePoolReaderFromChainDB chainDB =
...
+    , opwAddObjects = \votes -> do
+        now <- systemTimeCurrent systemTime
+        atomically $ do
+          alreadyInDb <- PerasVoteDB.getVoteIds perasVoteDB
+          let votesNotAlreadyInDb = filter ((`Set.notMember` alreadyInDb) . getPerasVoteId) votes
+          validatedVotes <- traverse (verifyPerasVoteWithHandle resolverHandle) votesNotAlreadyInDb
+          -- Some votes are invalid => reject the whole batch
+          -- We could combine the two 'traverse' operations into one in which case
+          -- any validated vote would be immediately added no matter what is the
+          -- validity of the other votes in the batch.
```
jasagredo, line 112: Same question, which one would be the right logic?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
```diff
@@ -47,9 +47,10 @@
 --
 -- = Quorum Threshold and Multiple Winners
 --
--- The quorum threshold is parameterized via 'PerasParams'. Depending on this
--- configuration and the weight distribution, it may be theoretically possible
--- for multiple targets to exceed the threshold within the same round.
+-- The quorum threshold is parameterized via 'PerasParams' inside
+-- 'PerasEpochContext'. Depending on this configuration and the weight
+-- distribution, it may be theoretically possible for multiple targets to exceed
+-- the threshold within the same round.
```
jasagredo, line 53: Somewhere I saw that there was a comment saying whether a MultipleWinners error should make the node shutdown. If this can happen at random then it should not.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
```diff
@@ -56,7 +56,7 @@ data ChainDbArgs f m blk = ChainDbArgs
   , cdbVolDbArgs :: VolatileDB.VolatileDbArgs f m blk
   , cdbLgrDbArgs :: LedgerDB.LedgerDbArgs f m blk
   , cdbPerasCertDbArgs :: PerasCertDB.PerasCertDbArgs f m blk
-  , cdbPerasVoteDbArgs :: PerasVoteDB.PerasVoteDbArgs f m blk
+  , cdbPerasVoteDbArgs :: Incomplete PerasVoteDB.PerasVoteDbArgs m blk
```
jasagredo, line 59: This is weird. Even if we have `Complete ChainDbArgs` it must contain an `Incomplete PerasVoteDbArgs`? That feels wrong

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -197,7 +200,14 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
...
 
     perasCertDB <- PerasCertDB.createDB argsPerasCertDB
-    perasVoteDB <- PerasVoteDB.createDB argsPerasVoteDB
+    perasVoteDB <-
+      PerasVoteDB.createDB
+        PerasVoteDB.PerasVoteDbArgs
+          { PerasVoteDB.pvdbaTracer =
+              PerasVoteDB.pvdbaTracer incompleteArgsPerasVoteDB
+          , PerasVoteDB.pvdbaPerasEpochContextResolverHandle =
+              mkPerasEpochContextResolverHandle (LedgerDB.getVolatileTip lgrDB)
```
jasagredo, line 209: Perhaps if this value will be supplied later on it should not be part of the Args.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasVoteDB/API.hs
```diff
@@ -98,25 +116,41 @@ newtype BlockedPerasRoundWinner blk
...
+data PerasVoteDbError blk where
+  -- | Attempted to add a vote that would lead to multiple winners for the same round
+  MultipleWinnersInRound ::
+    PerasRoundNo ->
+    (ExistingPerasRoundWinner blk) ->
+    (BlockedPerasRoundWinner blk) ->
+    PerasVoteDbError blk
+  -- | An error occurred while forging a certificate
+  ForgingCertError ::
+    Show (PerasError blk) =>
```
jasagredo, line 128: Is this worth it? We are not hiding the `blk` in an existential so we can as well just add this constraint in the use sites no?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -2040,18 +2137,20 @@ genBlkPair chunkInfo loe Model{..} =
...
                     chunkInfo
                     (prevEpoch + 1)
-                nextNextEBB =
-                  ImmutableDB.chunkSlotForBoundaryBlock
-                    chunkInfo
-                    (prevEpoch + 2)
+            -- nextNextEBB =
+            --   ImmutableDB.chunkSlotForBoundaryBlock
+            --     chunkInfo
+            --     (prevEpoch + 2)
```
jasagredo, line 2143: Should we delete this code then?

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -2040,18 +2137,20 @@ genBlkPair chunkInfo loe Model{..} =
...
             (slotNo, epoch) <-
               first (ImmutableDB.chunkSlotToSlot chunkInfo)
                 <$> frequency
                   [ (7, return (nextEBB, prevEpoch + 1))
-                  , (1, return (nextNextEBB, prevEpoch + 2))
+                  -- We can no longer generate EBBs that are more than one epoch in the future, as
+                  -- this would break ticking of the ExtLedgerState (specifically for the PerasEpochContextResolver)
+                  -- , (1, return (nextNextEBB, prevEpoch + 2))
```
jasagredo, line 2150: And this comment? or do we expect this to be possible in the future again?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: I believe this was a consequence of needing `StandardHash blk` in a couple of places. IIRC, most occurrences of either `Typeable` or `StandardHash` introduced by Peras need to be replaced with `BlockSupportsPeras` in #2240.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: There was some back and forth about this a while back, and IIRC, both options could work in practice. I think this comment was a bit of an effort to document this in case we need to revisit it later.

For now, I tweaked this comment to be more visible.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasVote.hs
agustinmista: Please see my [other comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/2239#discussion_r3947839506)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: IIRC, use `makePerasCertPoolWriterFromCertDB` in the cert diffusion smoke tests to avoid having to initialize an entire ChainDB (same with vote diffusion). These could also be used by any tool that needs vote/cert diffusion without a full ChainDB in place.

FTR: we thought about making these functions part of the test suite, but their implementation was not trivial and it's quite coupled to the `*FromChainDB` version, so we felt it made more sense to have the two variants together.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
agustinmista: `MultipleWinnersInRound` is theoretically possible in the sense that, with bad parameterization (that would likely be rejected by the on-chain governance guardrails anyway), we could observe a round with multiple winners before the heat death of the universe :sweat_smile: 

This is us just being extra careful about not burying this case under an `error "impossible"`.

ouroboros-consensus/test/storage-test/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
agustinmista: I don't think we expect this to be possible in the future. I have removed the comment and simplified the code now.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs
agustinmista: The original intent was to express the fact that _some_ of the PerasVoteDB arguments should not be provided from the outside because they would be supplied by the ChainDB itself (namely, the epoch context resolver).

But fair enough, this was confusing as hell, so I decoupled the resolver from the `PerasVoteDbArgs` and now it's being passed around next to the args. 

The (only?) downside of this change is that creating a PerasVoteDB now takes a Peras context resolver _in addition_ to a `Complete PerasVoteDbArgs`, which might be confusing in its own way.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
agustinmista: See [this other comment](https://github.com/IntersectMBO/ouroboros-consensus/pull/2239#discussion_r3948873597)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
jasagredo: Perhaps add the prefix `test*` to the functions?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
jasagredo: I meant this comment

https://github.com/IntersectMBO/ouroboros-consensus/blob/135d45b4a9a903f98ad307945cce1d5f2efc0149/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/RethrowPolicy.hs#L106

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/ObjectPool/PerasCert.hs
agustinmista: Done in ace91a51d65550632ed6657c55a2485cefd9006e

jasagredo (approved):

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Vote/Aggregation.hs
agustinmista: Removed TODO in d79a17f8a2d9bb920d9d9ea3ffb5474a14a185f6

PR #2244, opened 2026-08-27
Add `ouroboros-consensus:tracing` sublibrary

jasagredo:
Introduce a new public sublibrary 'tracing' with instances for Consensus types,
moved from cardano-node. Also a Claude summary of what needs to be removed after this PR

## What `cardano-node` can delete once it picks this up

Comparison basis: **`cardano-node` `c2ebdc87dfe07706a83e52f219e712c60d1b0a56`** (tag `11.1.1`, `origin/master`, 2026-09-04).

The sublibrary is a faithful move of node's tracing modules, verified by comparing
formatting-invariant fingerprints rather than raw diffs (the code here is
fourmolu-formatted, so line-by-line diffs are useless): the **in-order sequence of
string literals** — namespaces, JSON keys, metric names, doc text — and the
**token stream** with strings, comments and imports stripped.

Result: `BlockReplayProgress`, `ChainDB`, `Era/Byron`, `Era/HardFork`, `Era/Shelley`,
`Formatting` and `KESInfo` are **identical in emitted output** (in-order literal
diff = 0). `Consensus.hs` is a **strict superset** — no literal was removed.
Node's two recent tracing fixes are both already in: `8f75a3883` (BlockFetch
namespace drift) and `9e9cb95fa` (`severityFor`).

### Delete outright

| Module | Replaced by |
| --- | --- |
| `Cardano/Node/Tracing/Tracers/BlockReplayProgress.hs` | `Ouroboros.Consensus.Tracing.BlockReplayProgress` |
| `Cardano/Node/Tracing/Tracers/ChainDB.hs` | `Ouroboros.Consensus.Tracing.ChainDB` |
| `Cardano/Node/Tracing/Tracers/Consensus.hs` | `Ouroboros.Consensus.Tracing.Consensus` |
| `Cardano/Node/Tracing/Tracers/ConsensusStartupException.hs` | `Ouroboros.Consensus.Tracing.ConsensusStartupException` |
| `Cardano/Node/Tracing/Tracers/HasIssuer.hs` | `Ouroboros.Consensus.Tracing.HasIssuer` |
| `Cardano/Node/Tracing/Tracers/KESInfo.hs` | `Ouroboros.Consensus.Tracing.KESInfo` |
| `Cardano/Node/Tracing/Era/Byron.hs` | `Ouroboros.Consensus.Tracing.Era.Byron` |
| `Cardano/Node/Tracing/Era/HardFork.hs` | `Ouroboros.Consensus.Tracing.Era.HardFork` |
| `Cardano/Node/Tracing/Era/Shelley.hs` | `Ouroboros.Consensus.Tracing.Era.Shelley` |
| `Cardano/Node/Tracing/Formatting.hs` | `Ouroboros.Consensus.Tracing.Formatting` |
| `Cardano/Node/Tracing/Render.hs` | `Ouroboros.Consensus.Tracing.Render` + `…Tracing.Era.Shelley.Render` |

`Render.hs` goes in its entirety. Its `cardano-api`-dependent half
(`renderScriptHash`, `renderScriptIntegrityHash`, `renderScriptPurpose`,
`renderMissingRedeemers`, `renderIncompleteWithdrawals`) has exactly one consumer —
node's own `Era/Shelley.hs`, which is also deleted — and is reimplemented against
`cardano-ledger` in `Ouroboros.Consensus.Tracing.Era.Shelley.Render`. The one
surviving importer, `Tracers/NodeToNode.hs`, uses only `renderHeaderHash` and
`renderTxIdForDetails`, both of which `Ouroboros.Consensus.Tracing.Render` exports.

### Partial deletion: `Cardano/Node/Queries.hs`

Remove — all three are verbatim in the sublibrary:

- `ConvertTxId` (class + `ByronBlock` / `ShelleyBlock` / `HardForkBlock` instances) → `Ouroboros.Consensus.Tracing.ConvertTxId`
- `GetKESInfo` → `Ouroboros.Consensus.Tracing.KESInfo`
- `HasKESInfo` → `Ouroboros.Consensus.Tracing.KESInfo`

Keep: `MaxKESEvolutions`, `OperationalCertStartKESPeriod`, `KESMetricsData`,
`HasKESMetricsData`, `LedgerQueries`, and the `NodeKernelData` block.

### Import rewrites only (no deletion)

Everything below imports a deleted module and stays; point it at
`Ouroboros.Consensus.Tracing` (the umbrella re-export, so no orphan instance can be
silently lost by importing a subset):

- `Cardano/Node/Tracing.hs`
- `Cardano/Node/TraceConstraints.hs`
- `Cardano/Node/Tracing/Tracers.hs`
- `Cardano/Node/Tracing/Documentation.hs`
- `Cardano/Node/Tracing/Consistency.hs`
- `Cardano/Node/Tracing/Tracers/LedgerMetrics.hs`
- `Cardano/Node/Tracing/Tracers/NodeToNode.hs`
- `Cardano/Node/Protocol/Byron.hs`
- `Cardano/Node/Protocol/Shelley.hs`

### Two behavioural changes to expect

1. **`bench/trace-schemas/TraceMessage.schema.json` needs regenerating.**
   `ClientMetrics` had `documentFor _ = Just ""` in node; here it is keyed on the
   namespace and carries real text. Several object-diffusion namespaces gain
   documentation too (node has `documentFor _ = Nothing`). No namespace is added or
   removed by this — only doc strings change.
2. **The Dijkstra guarding plutus purpose now renders.** `Cardano/Node/Tracing/Render.hs:290`
   is `Api.AlonzoEraOnwardsDijkstra -> undefined`, with unmatched purposes falling
   through to `Aeson.Null`. The sublibrary emits `{"guarding": "<script hash>"}` and a
   `ScriptWitnessIndexGuarding` witness-index name — the latter has no `cardano-api`
   counterpart, so it is a genuinely new string in the logs.


tracing/Ouroboros/Consensus/Tracing/Consensus.hs
```diff
@@ -0,0 +1,3058 @@
...
+      ["kind" .= String "EmptyPeersFetch"]
+  forMachine _ xs =
+    mconcat
+      [ "kind" .= String "PeersFetch"
+      , "peers"
+          .= toJSON
+            (List.foldl' (\acc x -> forMachine DDetailed x : acc) [] xs)
+      ]
+
+  asMetrics _ = []
```
dnadales: Aren't we dropping the `connectedPeers` metric here? In `cardano-node` this same instance had:

```haskell
asMetrics peers = [IntM "connectedPeers" (fromIntegral (length peers))]
```

and `metricsDocFor` a bit further down (the `Decline`/`Accept` clauses of `MetaTrace (FetchDecision [Point header])`) still documents `connectedPeers`, so I guess the two disagree now. If the removal is deliberate, should the doc entries go with it? Otherwise dashboards on `cardano_node_metrics_connectedPeers_int` go flat with nothing to say why.

I compared against 10.6.2, so if a newer node already dropped it, ignore me.

tracing/Ouroboros/Consensus/Tracing/Consensus.hs
```diff
@@ -0,0 +1,3058 @@
...
+import qualified Ouroboros.Network.AnchoredSeq as AS
+import Ouroboros.Network.Block hiding (blockPrevHash)
+import Ouroboros.Network.BlockFetch.ClientState (TraceLabelPeer (..))
+import qualified Ouroboros.Network.BlockFetch.ClientState as BlockFetch
+import Ouroboros.Network.BlockFetch.Decision
+import Ouroboros.Network.BlockFetch.Decision.Trace (TraceDecisionEvent (..))
+import Ouroboros.Network.OrphanInstances ()
+import Ouroboros.Network.SizeInBytes (SizeInBytes (..))
+
+enclosingValue :: ToJSON a => Enclosing' a -> Value
```
dnadales: `enclosingValue` renames a field an operator sees. `TookSnapshot.enclosedTime` was `{"tag":"FallingEdgeWith","contents":1.53}` and becomes `{"edge":1.53}`. `TraceMempoolSynced.enclosingTime` changes the same way, and the rising case changes value too, `"RisingEdge"` to `"Starting"`. Is that deliberate?

I understand we cannot import `cardano-node`'s generically derived `ToJSON (Enclosing' t)` any more, so a replacement had to be written. But we could keep the old shape while we are at it. Something like:

```haskell
enclosingValue RisingEdge = object ["tag" .= String "RisingEdge"]
enclosingValue (FallingEdgeWith a) =
  object ["tag" .= String "FallingEdgeWith", "contents" .= toJSON a]
```

If we do prefer the new shape, I guess it wants a changelog entry, since log parsers break silently. Also the helper is duplicated verbatim in `ChainDB.hs`, so perhaps it belongs in `Formatting.hs`.

I compared against 10.6.2, so ignore me if a newer node already moved to `"edge"`.

nix/haskell.nix
```diff
@@ -18,6 +18,7 @@ let
       ../ouroboros-consensus-diffusion
       ../ouroboros-consensus-protocol
       ../ouroboros-consensus.cabal
+      ../tracing
```
dnadales, line 21: Should `tracing-test` join the golden list further down (`nix/haskell.nix:69-74`)? As it stands it gets neither `extraSrcFiles` nor `--no-create`, and `tracing/golden` sits outside its `hs-source-dirs: tracing/test`, so haskell.nix prunes the golden file out of the component source. `tasty-golden` then creates the missing file and reports a pass, which I guess makes the nix check green without comparing anything. The byron/shelley/cardano suites need `extraSrcFiles` for the same reason, even though the package is already in the fileset above.

Something like:

```nix
packages.ouroboros-consensus.components.tests.tracing-test = {
  testFlags = lib.mkForce [ "--no-create --hide-successes" ];
  extraSrcFiles = [ "tracing/golden/**/*" ];
};
```

A `nix build .#checks.x86_64-linux.<ghc>:ouroboros-consensus:test:tracing-test` and a grep for `Golden file did not exist; created` would settle it either way.

tracing/test/Test/Consensus/Tracing/MetaTrace.hs
```diff
@@ -0,0 +1,238 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- | Consistency checks over the 'MetaTrace' instances.
+--
+-- These need no trace values: everything here is derived from 'allNamespaces'
+-- and the namespace-indexed methods. That makes them cheap enough to run over
+-- every traced type, and they catch the mistakes that are easy to make when
+-- writing an instance by hand -- a namespace left out of 'allNamespaces', a
+-- typo that makes a namespace unreachable, a missing severity (which silently
```
dnadales, line 14: I wonder if the docstring promises a bit more than the checks deliver. "A namespace left out of `allNamespaces`" is the case where some constructor's `namespaceFor` returns a namespace that is not in the list, but every check below iterates `allNamespaces` itself and `namespaceFor` is never called. So that particular mistake still passes, and so does a typo, as long as it appears in both places.

I do not think that is fixable cheaply. Catching it needs values, and these types have no `Arbitrary` or `Enum`. Should we narrow the wording instead, so the next reader knows what is and is not covered? Something like "every namespace in `allNamespaces` has a severity, a privacy, a detail level and documentation, and the tree is addressable", plus a note that the `namespaceFor` direction is not checked.

tracing/test/Test/Consensus/Tracing/Golden.hs
```diff
@@ -0,0 +1,175 @@
...
+purposesByIndex =
+  [ ("spending", ConwaySpending (AsIx 0))
+  , ("minting", ConwayMinting (AsIx 1))
+  , ("certifying", ConwayCertifying (AsIx 2))
+  , ("withdrawing", ConwayWithdrawing (AsIx 3))
+  , ("voting", ConwayVoting (AsIx 4))
+  , ("proposing", ConwayProposing (AsIx 5))
+  ]
+
+purposesByItem :: [(String, ConwayPlutusPurpose AsItem ConwayEra)]
```
dnadales, line 118: Should `purposesByItem` cover all six the way `purposesByIndex` does? Certifying, voting and proposing are missing, and going by the comment just above they are the `{"item": ...}`-wrapped ones, so they are the cases most likely to drift. The `unknownPurpose` fallback and `renderTxIn` have no entry either.

Adding them looks like three lines, and the golden file grows by three.

tracing/Ouroboros/Consensus/Tracing/Consensus.hs
```diff
@@ -0,0 +1,3058 @@
...
+    mconcat
+      [ "kind" .= String "PeerStarvedUs"
+      , "peer" .= toJSON peer
+      ]
+
+instance
+  (LogFormatting peer, Show peer) =>
+  LogFormatting [TraceLabelPeer peer (FetchDecision [Point header])]
+  where
+  forMachine DMinimal _ = mempty
```
dnadales, line 738: Are these two instances still needed? `cardano-node` master removed the `[TraceLabelPeer peer (FetchDecision [Point header])]` instances under [#6667](https://github.com/IntersectMBO/cardano-node/issues/6667), together with the `connectedPeers` metric documentation, because nothing emits them and their `Accept`/`Decline`/`EmptyPeersFetch` namespaces were the stale ones the configuration consistency check rejected.

Carrying them into the sublibrary brings back both the dead instances and a `metricsDocFor` advertising a metric that `asMetrics` at line 750 never produces. Our own `MetaTrace` test only covers `TraceDecisionEvent`, which I guess confirms nothing else needs them.

Should we drop them and keep just `LogFormatting (FetchDecision [Point header])`, the way master does?

tracing/Ouroboros/Consensus/Tracing/Era/Shelley/Render.hs
```diff
@@ -0,0 +1,195 @@
...
+unsafeHrp :: Text -> Bech32.HumanReadablePart
+unsafeHrp t = case Bech32.humanReadablePartFromText t of
+  Right hrp -> hrp
+  Left err -> error ("renderRewardAccount: invalid HRP " <> show t <> ": " <> show err)
+
+-- | Rendered in place of a plutus purpose that none of the @AnyEraScript@
+-- projections below matched.
+--
+-- Those projections cover every purpose the ledger currently defines, but they
+-- are pattern synonyms without a @COMPLETE@ pragma, so GHC cannot check that
```
dnadales, line 106: I think there is a `COMPLETE` pragma: `cardano-ledger-api`'s `Cardano.Ledger.Api.Scripts` has one covering the seven `AnyEra*Purpose` synonyms, `AnyEraGuardingPurpose` included. So GHC does check exhaustiveness here, and both renderers match six of the seven.

Which makes the fallthrough reachable today rather than in some future era: a Dijkstra guarding redeemer renders as `{"kind":"UnknownPlutusPurpose"}`. Since we are matching the sixth case explicitly anyway, should we add the seventh and drop the wildcard? Then GHC tells us about the next era, instead of a marker turning up in someone's logs.

Either way the comment wants a small correction.

ouroboros-consensus.cabal
```diff
@@ -414,6 +414,102 @@ library
     directory
     latex-svg-image
 
+library tracing
```
dnadales, line 417: Should `tracing` join the `libraries` array in `scripts/ci/check-changelogs.sh:12-20` while we are adding it? As it stands a later change to this sublibrary's `.hs` files never asks for a fragment, and it is public API: `HasIssuer`, `ConvertTxId`, and every log shape an operator parses.

The path would be `tracing`, since the sublibrary's `hs-source-dirs` is the directory itself rather than a `src/` subdirectory.

Relatedly, I do not see a fragment for this PR, and no `no changelog` label. I guess CI will point that out anyway.

tracing/Ouroboros/Consensus/Tracing/ChainDB.hs
```diff
@@ -0,0 +1,3512 @@
...
+{-# OPTIONS_GHC -Wno-deprecations #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.Tracing.ChainDB
+  ( withAddedToCurrentChainEmptyLimited
+  , fragmentChainDensity
+  ) where
+
+import Cardano.Logging
+import Cardano.Prelude (maximumDef)
```
dnadales, line 18: `maximumDef` is the only thing we take from `cardano-prelude`, at lines 909 and 914. Is it worth a dependency on a new public library for that? `foldr max Notice (map sevLedgerEvent events)` gives the same answer, and then the sublibrary's dependency list is one entry shorter.

The main `ouroboros-consensus` library does not depend on `cardano-prelude` today, so this adds it to the tree rather than reusing something already there.

PR #2245, opened 2026-08-28
LeiosDb: bound the WAL, remove the write hotspot, stop dying on lock contention

### LINKED ISSUE #969: LeiosDb volatile/immutable split + GC

### LINKED ISSUE #1016: Prototype: EB validation

ch1bo:
Several fixes to the Leios SQLite layer, stacked on the recent tx cache / fetch logic / eb validation work. The fixes were driven by multi-hour long stress tests of a twelve-node `dozen-devnet` at `TPS=500` (effective ingest lower than that).

> [!IMPORTANT]
>
> Note the schema change has no migration: an existing database is not readable by this version and requires a wipe & resync.

## What was wrong

**1. `page_size` was a silent no-op.** It was set after `journal_mode = WAL`, and SQLite cannot change the page size of a database already in WAL mode. Every run so far used the 4096 default. With 32768 actually in effect, WAL amplification measured 35x (34 GiB of log for 0.97 GiB of data) against ~18x at 4096. Reordered and pinned at 4096.

**2. No `busy_timeout` was set,** so SQLite refused a contended lock immediately and all waiting happened in an application loop whose Nth sleep was `100us * N` — cumulative `100us * N(N+1)/2`. The observed EB-body insert distribution matched that curve exactly: median 1 599 ms (N≈178), max 18 435 ms (N≈607), for 90 ms of actual work. `busy_timeout = 1000` now does the waiting in C, and the application loop is a flat 20 ms backstop, making the ceiling linear rather than quadratic. Current ceiling is ~30 s after which a node would crash.

**3. `idx_ebTxs_txHashBytes` was the dominant write cost.** It existed for one query — decrementing `missingTxCount` for EBs referencing an arriving tx — and covered every `(EB, tx)` pair ever seen: 4.99M entries for 1.07M distinct hashes (4.7x on average, 43x worst), 887 MB of a 1 484 MB database. Each 13 568-row body insert scattered that many random-hash insertions, dirtying a page apiece.

An arriving tx can only complete EBs that were still *waiting* for it; EBs that already held it counted it as present when their body landed. The resolution between `ebs` and `txs` is now two tables: `ebTxs` keeps the ordered body, and `ebsMissingTxs` holds only pairs an EB still lacks, populated by one anti-join at body insert and retired as txs arrive. Both move in the same transaction. Serving is unchanged — PK range scan plus join.

## Measured effect

Write amplification, one 13 568-row EB body insert against a 4M-row table:

|                              | WAL generated | per row |
|------------------------------|---------------|---------|
| with `idx_ebTxs_txHashBytes` | 61.0 MB       | 4 494 B |
| two-table, no missing txs    | 1.8 MB        | 134 B   |
| two-table, 2% missing        | 3.2 MB        | 235 B   |

EB body ingest latency at matched closure size (~6 500 txs):

|                            | median | p90    |
|----------------------------|--------|--------|
| before                     | 179 ms | —      |
| `busy_timeout` only        | 80 ms  | 284 ms |
| `busy_timeout` + two-table | 36 ms  | 84 ms  |

Per instance, over a full run:

|                      | before                            | after                                |
|----------------------|-----------------------------------|--------------------------------------|
| WAL                  | 9 304 MB against a 54 MB database | 57–67 MB against a 3 070 MB database |
| first `tooLate` vote | ~25 min                           | ~1 h 15 min                          |
| certification        | stopped at ~2 h                   | ran 5 h 20 min                       |

The WAL figure is #2233: a writer waiting inside a deferred transaction pinned a read snapshot, freezing back-fill (`framesCopied` stuck at 546 while `framesInLog` reached 241 239), so no checkpoint could reset the log.

## Instrumentation

`TraceLeiosBlockIngest` reports, per ingested body, the wait for the node-wide outstanding lock and then the database insert, tx-cache insert and mempool pull separately, plus total lock hold. This is what localised the cost; without it the DB insert and the lock wait are indistinguishable.

`TraceLeiosDbBusyRetry` fires when SQLite's own timeout was exhausted and the application backstop engaged, with attempt number and accumulated wait. A steady stream of these means the write path is saturated.

## Remaining work

Database access is not yet congestion-free. In the same run the insert median still grew from 6 ms to 524 ms over seven hours as the instance database reached 3 070 MB, and the busiest instance logged 44 `TraceLeiosDbBusyRetry` events over 4.5 hours — each one a `BEGIN IMMEDIATE` that gave up after a full second in SQLite's handler. That is a floor on contention, not a measure of it: an attempt that waits and then succeeds is not traced. Neither `ebTxs`, `ebsMissingTxs` nor `txs` is pruned, so per-EB cost continues to rise with database size; pruning both resolution tables on the same age rule is the next step, and `idx_ebsMissingTxs_ebHashBytes` exists to support it.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedRecordDot #-}
@@ -44,6 +45,7 @@ import Database.SQLite3
   , open2
   )
 import qualified Database.SQLite3.Direct as DB
+import GHC.Clock (getMonotonicTime)
 import GHC.Stack (HasCallStack)
 import qualified GHC.Stack
 import LeiosDemoDb.Common
@@ -136,6 +138,8 @@ data Stmts = Stmts
 data Conn = Conn
   { connDb :: !DB.Database
   , connStmts :: !Stmts
+  , connTracer :: !(Tracer IO TraceLeiosDb)
+  -- ^ So the write path can report exhausting SQLite's own busy timeout.
   }
 
 -- | Prepare every statement 'Stmts' names. Order is not observable.
@@ -193,7 +197,17 @@ openSQLiteConnection tracer dbPath notificationChan = do
   shouldInitSchema <- not <$> doesFileExist dbPath
   db <- open2 (fromString dbPath) [SQLOpenReadWrite, SQLOpenCreate] SQLVFSDefault
   traverse_ (dbExec db) $
-    [ "pragma synchronous = normal;"
+    [ -- First, before any pragma that takes a lock -- 'journal_mode' does. Until
+      -- this runs the timeout is zero, so a contended lock is refused outright
+      -- rather than waited for, and opening a second connection to a busy
+      -- database fails where it should merely be slow.
+      --
+      -- Let SQLite do that waiting in C, retrying tightly rather than sleeping
+      -- through the window it is waiting for. Safe because writers take the lock
+      -- at BEGIN, so nothing waits here holding a snapshot; see
+      -- 'dbWithWriteTransaction'.
+      "pragma busy_timeout = 1000;"
+    , "pragma synchronous = normal;"
     , -- Must precede 'journal_mode': SQLite cannot change the page size of a
       -- database already in WAL mode, so the order this list used to have left
       -- the setting a silent no-op and every run so far on the 4096 default.
@@ -215,7 +229,7 @@ openSQLiteConnection tracer dbPath notificationChan = do
   when shouldInitSchema $
     dbExec db (fromString sql_schema)
   stmts <- prepareStmts db
-  let conn = Conn{connDb = db, connStmts = stmts}
+  let conn = Conn{connDb = db, connStmts = stmts, connTracer = tracer}
       notify = atomically . writeTChan notificationChan
   pure $
     LeiosDbConnection
@@ -277,13 +291,13 @@ sqlLookupEbBody conn ebHash =
 
 sqlInsertEbPoint :: Conn -> LeiosPoint -> BytesSize -> IO ()
 sqlInsertEbPoint conn point ebBytesSize =
-  dbWithWriteTransaction db $ useStmt stmt $ do
+  dbWithWriteTransaction conn $ useStmt stmt $ do
     dbBindInt64 stmt 1 (fromIntegral $ unSlotNo point.pointSlotNo)
     dbBindBlob stmt 2 point.pointEbHash.ebHashBytes
     dbBindInt64 stmt 3 (fromIntegral ebBytesSize)
     dbStep1 stmt
  where
-  Conn{connDb = db, connStmts = Stmts{stInsertEbPoint = stmt}} = conn
+  Conn{connStmts = Stmts{stInsertEbPoint = stmt}} = conn
 
 -- | Persist an EB body. The point MUST already be present (inserted
 -- via 'sqlInsertEbPoint' on the announcement path).
@@ -297,7 +311,7 @@ sqlInsertEbBody ::
 sqlInsertEbBody tracer conn notify point eb = do
   when (null items) $
     error "leiosDbInsertEbBody: empty EB body (programmer error)"
-  completedNow <- dbWithWriteTransaction db $ do
+  completedNow <- dbWithWriteTransaction conn $ do
     forM_ items $ \(txOffset, txHash, txBytesSize) -> useStmt stInsertEbTxsRow $ do
       dbBindBlob stInsertEbTxsRow 1 point.pointEbHash.ebHashBytes
       dbBindInt64 stInsertEbTxsRow 2 (fromIntegral txOffset)
@@ -330,7 +344,7 @@ sqlInsertEbBody tracer conn notify point eb = do
  where
   items = leiosEbBodyItems eb
   ebBytesSize = leiosEbBytesSize eb
-  Conn{connDb = db, connStmts} = conn
+  Conn{connStmts} = conn
   Stmts
     { stInsertEbTxsRow
     , stInitMissingCount
@@ -363,7 +377,7 @@ sqlInsertTxs _tracer conn notify txs = do
   -- hashes; attempting the INSERT and catching a constraint violation
   -- still pays the bind + PK-lookup + reset cost per row.
   missing <- Set.fromList <$> sqlFilterMissingTxs conn (map fst txs)
-  completed <- dbWithWriteTransaction db $ do
+  completed <- dbWithWriteTransaction conn $ do
     -- 'dbStepInsert' still handles the rare race where a concurrent
     -- writer inserted the same hash between the filter above and the
     -- INSERT below.
@@ -395,7 +409,7 @@ sqlInsertTxs _tracer conn notify txs = do
   forM_ completed $ \point -> notify (AcquiredEbTxs point)
   pure completed
  where
-  Conn{connDb = db, connStmts} = conn
+  Conn{connStmts} = conn
   Stmts{stInsertTx, stDecrMissingCount, stFindCompleteEbs, stMarkNotifiedEbs} = connStmts
   novel missing = filter (\(h, _) -> h `Set.member` missing) txs
 
@@ -684,8 +698,56 @@ dbWithTransaction = dbWithTransactionAs "BEGIN"
 -- transaction's snapshot is stale for good: the only remedy is to roll back and
 -- start over. Taking the lock at BEGIN removes the upgrade, so contention
 -- surfaces here instead, where waiting actually resolves it.
-dbWithWriteTransaction :: HasCallStack => DB.Database -> IO a -> IO a
-dbWithWriteTransaction = dbWithTransactionAs "BEGIN IMMEDIATE"
+dbWithWriteTransaction :: HasCallStack => Conn -> IO a -> IO a
+dbWithWriteTransaction conn k = getMonotonicTime >>= go 0
+ where
+  Conn{connDb = db, connTracer = tracer} = conn
+
+  -- After this many refusals, a write transaction is no longer merely
+  -- contended.
+  --
+  -- This picks which constructor gets traced and nothing else. Crossing it does
+  -- not change how long we wait, does not throw, and does not abandon anything
+  -- -- 'dbWithWriteTransaction' retries forever either way. It exists only so
+  -- that the severity in the log matches the severity of the situation.
+  --
+  -- Each attempt is a full 'busy_timeout', so this is about half a minute of one
+  -- writer making no progress, re-traced every half minute it stays that way.
+  busyStuckAfter = 30
+
+  go !attempt t0 =
+    fmap (first fst) (DB.exec db (fromString "BEGIN IMMEDIATE")) >>= \case
+      Left DB.ErrorBusy -> do
+        -- Unbounded, deliberately. Nothing is held while waiting here -- that is
+        -- the whole point of taking the lock at BEGIN -- so waiting costs
+        -- latency and nothing else, whereas giving up throws, and a throw on
+        -- this path kills the Leios threads outright. Past
+        -- 'busyStuckAfter' attempts that is no longer ordinary contention, so
+        -- say so at a severity someone will notice, and keep waiting.
+        --
+        -- The wait is measured, not accumulated: most of it happens inside
+        -- SQLite's own busy handler, so summing the sleeps below would report a
+        -- fraction of the truth and disagree with the log timestamps.
+        now <- getMonotonicTime
+        let n = attempt + 1
+            waitedMs = 1000 * (now - t0)
+        traceWith tracer $
+          if n >= busyStuckAfter && n `mod` busyStuckAfter == 0
+            then TraceLeiosDbBusyStuck n waitedMs
+            else TraceLeiosDbBusyRetry n waitedMs
+        busyBackoff
+        go n t0
+      Left e -> throwDbException db e
+      Right () ->
+        fmap fst $
+          generalBracket
+            (pure ())
+            ( \() -> \case
+                MonadThrow.ExitCaseSuccess _ -> dbExec db (fromString "COMMIT")
+                MonadThrow.ExitCaseException _ -> dbExec db (fromString "ROLLBACK")
+                MonadThrow.ExitCaseAbort -> dbExec db (fromString "ROLLBACK")
+            )
+            (\() -> k)
 
 dbWithTransactionAs :: HasCallStack => String -> DB.Database -> IO a -> IO a
 dbWithTransactionAs begin db k =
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -193,15 +201,34 @@ openSQLiteConnection tracer dbPath notificationChan = do
...
+      -- both hot indexes are keyed by hash, so writes scatter -- the page count
+      -- barely falls as the page grows, the bytes just multiply.
+      "pragma page_size = 4096;"
     , "pragma mmap_size = 268435500;"
+    , "pragma journal_mode = WAL;"
+    , -- Let SQLite do the waiting for a contended lock, in C, retrying tightly
+      -- rather than sleeping through the window it is waiting for. Safe because
+      -- writers take the lock at BEGIN, so nothing waits here holding a
+      -- snapshot; see 'dbWithWriteTransaction'.
+      "pragma busy_timeout = 1000;"
```
nfrisby: A "defensive" lesson I learned elsewhere: `busy_timeout` should be set before any other PRAGMA, because some other PRAGMA statements _acquire locks_, and the default timeout (before we override it with `busy_timeout` is zero).

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -499,7 +544,19 @@ sql_schema =
     , "  txBytesSize INTEGER NOT NULL,"
     , "  PRIMARY KEY (ebHashBytes, txOffset)"
     , ");"
-    , "CREATE INDEX idx_ebTxs_txHashBytes ON ebTxs(txHashBytes);"
+    , -- The EB -> tx resolution table carries no secondary index: one keyed by
```
nfrisby: > The EB -> tx resolution table

Which table is that? The `ebTxs` table? Can we just refer to it by name, since that's less amibiguous? For a while I thought this statement was referring to new ebMissingTxs table.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -499,7 +544,19 @@ sql_schema =
...
+      -- tx hash cost 4.5 KiB of WAL per 70-byte row, because 13.5k random-hash
+      -- insertions per EB each dirty their own page. Measured on a 4M-row table:
+      -- 60.5 MiB of log per EB with it, 1.8 MiB without. The reverse lookup it
+      -- served lives in 'ebsMissingTxs' instead.
+      "CREATE TABLE ebsMissingTxs ("
+    , "  txHashBytes BLOB NOT NULL,"
+    , "  ebHashBytes BLOB NOT NULL,"
+    , "  PRIMARY KEY (txHashBytes, ebHashBytes)"
+    , ");"
+    , -- For pruning an EB's outstanding rows; small, since the table only ever
```
nfrisby: > small, since the table only ever holds pairs an EB is still waiting on

That text assumes the closure will definitely arrive. But an adversary is free to send an EB body and never send its closure.

Does that matter?

If not, then this comment sentence should be changed to explain why the size of this table doesn't matter.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -574,12 +631,41 @@ sql_mark_notified_ebs :: String
 sql_mark_notified_ebs =
   "UPDATE ebs SET missingTxCount = -1 WHERE missingTxCount = 0"
 
--- | Decrement missingTxCount for all EBs referencing the given txHash.
+-- | Decrement missingTxCount for every EB still /waiting/ on the given txHash.
+--
+-- Sourced from 'ebsMissingTxs' rather than from 'ebTxs', which is what lets the
+-- latter drop its tx-hash index. It is also tighter: 'ebTxs' names every EB that
```
nfrisby: > It is also tighter: 'ebTxs' names every EB that
> -- ever mentioned the tx (4.7 on average, up to 43 in the field), while only the
> -- ones that lacked it can be completed by its arrival

I don't understand this distinction. Either all EBs that reference the tx lack it or all EBs that reference the tx don't lack it. There's no in between, where some referencing EB lack it and others don't.

Unless maybe there's jargon on `lack` that I'm not understanding? In which case: we need a more complete comment.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -574,12 +631,41 @@ sql_mark_notified_ebs :: String
...
+-- | Retire the waiting rows for a tx that has just landed.
+-- Parameter 1: txHashBytes
+sql_delete_missing_txs :: String
+sql_delete_missing_txs =
+  "DELETE FROM ebsMissingTxs WHERE txHashBytes = ?"
+
+-- | Record which of a freshly-inserted body's txs we do not yet hold.
+--
+-- One anti-join over the EB's own 'ebTxs' range -- the same work
+-- 'sql_init_missing_tx_count' used to do to produce a count, now materialised so
```
nfrisby, line 657: Why is what's being looked up "small"?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -670,8 +754,38 @@ dbWithTransaction = dbWithTransactionAs "BEGIN"
...
+        -- this path kills the Leios threads outright. Past
+        -- 'busyStuckAfter' attempts that is no longer ordinary contention, so
+        -- say so at a severity someone will notice, and keep waiting.
+        let n = attempt + 1
+        traceWith tracer $
+          if n >= busyStuckAfter && n `mod` busyStuckAfter == 0
+            then TraceLeiosDbBusyStuck n waitedMs
+            else TraceLeiosDbBusyRetry n waitedMs
+        slept <- busyBackoff
+        go n (waitedMs + slept)
```
nfrisby: Will it be confusing if the log timestamps don't match this accumulator?

Should we instead/also be doing one monotonic clock read at the start and another per attempt?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -670,8 +754,38 @@ dbWithTransaction = dbWithTransactionAs "BEGIN"
...
+      Left e -> throwDbException db e
+      Right () -> runBody
+  runBody =
+    fmap fst $
+      generalBracket
```
nfrisby: Just re-confirmed for myself that `generalBracket` masks async exns only in the acquire and release steps; both of which are ephemeral here 👍

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -746,8 +857,39 @@ dbStepInsertOrTrace tracer table key stmt = do
...
+-- these stay bounded, unlike 'dbWithWriteTransaction', which holds nothing.
+--
+-- With 'busy_timeout' doing the real waiting, each attempt costs about a
+-- timeout, so the ceiling is linear rather than the quadratic 83 minutes the
+-- escalating sleep used to reach.
 maxBusyRetries :: Int
-maxBusyRetries = 1000
+maxBusyRetries = 30
+
+-- | After this many refusals, a write transaction is no longer merely contended.
```
nfrisby: This comment sounds like this value is much more impactful than it is. All it does is change the traced message, right? There's no other/"real" consequence.

I suggest the comment should state that loudly.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -746,8 +857,39 @@ dbStepInsertOrTrace tracer table key stmt = do
 -- hanging. 1000 gives roughly 50 s in total with a longest single sleep near
 -- 100 ms, which outlasts any transient contention while still surfacing a real
 -- deadlock inside a minute.
+-- | How many times a busy operation is re-attempted /after/ SQLite's own
```
nfrisby: ~This Haddock `|`  is unnecessary~

I _think_: this new Haddock comment is replacing the Haddock comment above it, but doesn't delete it for some reason?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -746,8 +857,39 @@ dbStepInsertOrTrace tracer table key stmt = do
...
+-- 'busy_timeout' has already expired on that attempt.
+--
+-- Applies to statements retried /inside/ an open transaction, where waiting is
+-- not free: the transaction holds its snapshot throughout, and a reader that
+-- waits indefinitely there is exactly what pins the WAL and stops back-fill. So
+-- these stay bounded, unlike 'dbWithWriteTransaction', which holds nothing.
+--
+-- With 'busy_timeout' doing the real waiting, each attempt costs about a
+-- timeout, so the ceiling is linear rather than the quadratic 83 minutes the
+-- escalating sleep used to reach.
```
nfrisby: Would this comment be a good place to decisively state what happens when we do end up retrying this many times?

I glanced at the call sites, and it looks like: we stop handling the `BusyError` SQLite exception and just let it rise. I'm assuming there are _some_ calling threads where that will cause the node to crash. Is that the intention? IE "something is very wrong" and so crashing is justified?

edit: I'm asking in particular because the Haddock on `TraceLeiosDbBusyStuck` argues about _never_ giving up in order to avoid killing the calling thread.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -819,6 +819,10 @@ data LeiosBlockTxsSource pid
 -- announcement-slot onset (see 'Leios.ebStateOnset') to @now@, or 'Nothing' if
 -- the EB was never heralded by an announcement (an offer-only or self-forged
 -- body).
+-- | Milliseconds between two 'systemTimeCurrent' readings.
```
nfrisby: This declaration was (inhumanly) placed between the haddock for `ebPointAge` and the declaration of `ebPointAge`.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -927,26 +934,30 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
             ( (if tooOld then fetchArrivalEvicted else fetchArrivalExtra) $ ebBytesSize'
             , Map.empty
             , Map.empty
+            , Nothing
```
nfrisby: We don't want to report "incomplete" timing data here? We don't care how long it took to acquire the lock when the EB body ended up being ignored?

I haven't gotten there yet, but as long as the Haddock on the timing data warns about this kind of suppression, that seems fine.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1304,6 +1304,22 @@ messageLeiosFetchToObject = \case
     "kind" .= Aeson.String "MsgDone"
 
 -- | Summary of an EB body inserted into the LeiosTxCache, for observability (see
+-- | Milliseconds spent in each stage of one 'processLeiosBlock' ingestion.
```
nfrisby: This declaration is again being placed between the Haddock comment and the declaration `LeiosTxCacheInsertBodySummary`.

New weird robot behaviors every day haha

nfrisby:
```suggestion
-- | Milliseconds spent in each stage of one 'processLeiosBlock' call, if the EB arrival wasn't too old/redundant.
```

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1481,6 +1505,33 @@ data LeiosNotVotedReason
 
 deriving instance Show TraceLeiosKernel
 
+-- | Render a 'TraceLeiosDb'. Kept separate and total: the outer
```
nfrisby: I suspect all the words in this comment after the first three are unnecessary noise.

nfrisby (changes requested):
The code itself looks great. Many of the comments were wrong/incomplete/(weirdly) harmfully misplaced, so I'm Requesting Changes just for that.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
ch1bo: Bad 🧞

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1013,7 +1026,20 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
+                , Leios.itDbInsertMs = dbInsertMs
+                , Leios.itCacheInsertMs = cacheInsertMs
+                , Leios.itMempoolPullMs = elapsedMs tPullStart tPullEnd
+                , Leios.itLockHeldMs = elapsedMs tGotLock tDone
+                , Leios.itTxsInEb = V.length ebTxs
+                }
+        pure (outstanding', (bodyClass, mempoolNotCache, mempoolAndCache, Just timing))
+  -- Outside the lock deliberately: tracing under it would be part of what we
+  -- are trying to measure.
+  forM_ mbTiming $ \t -> traceWith ktracer $ TraceLeiosBlockIngest point t
```
ch1bo: This is just stupid. Taking the measurement is skewing things as much as yielding a trace (I hope that tracing is not done synchronously!).

Anyhow, I never meant to commit so much profiling / debugging info. My issue was plenty enough observable by the traces we already had and reproducible so I could just pop in these timers.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: I corrected some of the docs and the `dbWithWriteTransaction` is never giving up now. I don't like the asymmetry, but also not the willpower to change this now. Proper error handling will need to come with production-gradingf this.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: I wonder myself why not `finally`?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: This is not a final solution. But I hope it's undoubtly better than labeling each row in `ebTxs` with missing or not: In the worst case will the `ebMissingTxs` the same size of `ebTxs`, but in every other case it is less to process for notifying completion.

I progressed further on a final solution for the completion tracking by using the tx cache.. but that is another story / another PR.

PR #2249, opened 2026-08-28
Acquire EB closure age

coot:
# Description

Added `NominalDiffTime` to `TraceLeiosBlockTxsAcquired`.  This will
allow us to compute 90, 95 and 99 percentiles of how long it takes to acquire
all txs announced in an EB.

# WARNING

To update your feature branch if it's stale, please rebase it manually on top of `main`. Don't update your feature branch by merging `main` into it. Your pull request will not pass CI if you do.

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
@@ -166,6 +166,13 @@ traceException tracer toTrace action =
   per tx, and the serialized body is the @b@.
 -------------------------------------------------------------------------------}
 
+headerRbHash ::
+  forall blk.
+  (ConvertRawHash blk, HasHeader (Header blk)) =>
+  Header blk ->
+  RbHash
+headerRbHash = MkRbHash . toRawHash (Proxy @blk) . headerHash
+
 -- | Insert an EB announcement into the tx-cache index, keyed by the announced
 -- slot, the announcing RB header's hash, and the announced EB hash. Evicted
 -- bodies\/txs are discarded; they can be useful for debugging/etc.
@@ -179,7 +186,7 @@ recordAnnouncementInTxCache ::
 recordAnnouncementInTxCache txCache ancHdr point =
   void $ txCache.insertAnnouncement point.pointSlotNo rbh point.pointEbHash
  where
-  rbh = MkRbHash (toRawHash (Proxy @blk) (headerHash (ancHeader ancHdr)))
+  rbh = headerRbHash (ancHeader ancHdr)
 
 -- | Register a locally-forged EB in the tx-cache: its announcement, its
 -- body, and each of its txs as already-applied (the forger drew them from its
@@ -1523,10 +1530,15 @@ instance HasHeader (Header blk) => Eq (AnnouncingHeader blk) where
 -- | Interpret a header as a relayed EB announcement, or 'Nothing' if it carries
 -- no announcement (so it should not have been relayed as one). Parsing the
 -- announcement once here keeps it total for all later consumers (e.g. tracing).
-mkAnnouncingHeader :: ResolveLeiosBlock blk => Header blk -> Maybe (AnnouncingHeader blk)
+mkAnnouncingHeader ::
+  (ConvertRawHash blk, HasHeader (Header blk), ResolveLeiosBlock blk) =>
+  Header blk ->
+  Maybe (AnnouncingHeader blk)
 mkAnnouncingHeader h =
   headerLeiosAnnouncement h <&> \(MkLeiosPoint _ebSlot ebHash, ebBodySize) ->
-    UnsafeMkAnnouncingHeader h (MkAnnouncementFields (headerElId h) ebHash ebBodySize)
+    UnsafeMkAnnouncingHeader h (MkAnnouncementFields (headerElId h) ebHash ebBodySize rbHash)
+  where
+    rbHash = headerRbHash h
 
 -- | The other safe constructor of an 'AnnouncingHeader': for a header we already
 -- know announces a specific EB because we forged it. Unlike 'mkAnnouncingHeader'
@@ -1534,14 +1546,25 @@ mkAnnouncingHeader h =
 -- announcement fields come straight from the 'ForgedLeiosEb' whose EB the header
 -- announces by construction.
 mkForgedAnnouncingHeader ::
-  ResolveLeiosBlock blk => Header blk -> Leios.ForgedLeiosEb -> AnnouncingHeader blk
+  (ConvertRawHash blk, HasHeader (Header blk), ResolveLeiosBlock blk) =>
+  Header blk ->
+  Leios.ForgedLeiosEb ->
+  AnnouncingHeader blk
 mkForgedAnnouncingHeader h forgedEb =
   UnsafeMkAnnouncingHeader h $
-    MkAnnouncementFields (headerElId h) forgedEb.point.pointEbHash (leiosEbBytesSize forgedEb.body)
+    MkAnnouncementFields (headerElId h) forgedEb.point.pointEbHash (leiosEbBytesSize forgedEb.body) rbHash
+  where
+    rbHash = headerRbHash h
 
 -- | The election of an 'AnnouncingHeader'.
 ancElId :: AnnouncingHeader blk -> ElId
 ancElId = announcementElection . ancAnnouncementFields
+
+ancEbHash :: AnnouncingHeader blk -> EbHash
+ancEbHash = announcementEbHash . ancAnnouncementFields
+
+ancRbHash :: AnnouncingHeader blk -> RbHash
+ancRbHash = Leios.announcementRbHash . ancAnnouncementFields
 
 -- | The central-state handling shared by an incoming LeiosNotify
 -- 'MsgLeiosBlockAnnouncement' and a ChainSync 'MsgRollForward' that announces an
@@ -1583,6 +1606,7 @@ processAnnouncementCentrally
       Announcements.onAnnouncementCentral
         (contramap (traceNewAnnouncement provenance) kernelTracer)
         ancElId
+        ancRbHash
         ( \_elSt -> do
             -- A received announcement lists the EB for fetching; one we forged is
             -- instead marked 'BodyImminent' in 'ebState' so the fetch logic never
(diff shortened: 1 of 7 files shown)


ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -908,7 +908,10 @@ data TraceLeiosKernel
   | -- | The EB body was received but the point was not in the database. This is
     -- unexpected as the point should have been inserted during announcement handling.
     TraceLeiosBlockPointMissing LeiosPoint
-  | TraceLeiosBlockTxsAcquired LeiosPoint
+  | TraceLeiosBlockTxsAcquired 
+      { leiosPoint :: LeiosPoint
+      , leiosEbAge :: NominalDiffTime
+      }
```
ch1bo:
```suggestion
      { ebPoint :: LeiosPoint
      , ebAge :: NominalDiffTime
      }
```

Would be more consistent with the naming scheme here (no `leios` prefixes in this module)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -706,7 +730,9 @@ msgLeiosBlock ktracer tracer (outstandingVar, readyVar) db peerId req eb = do
         leiosDbInsertEbPoint db point ebBytesSize
         completedByBody <- leiosDbInsertEbBody db point eb
         traceWith ktracer $ TraceLeiosBlockAcquired point
-        forM_ completedByBody $ traceWith ktracer . TraceLeiosBlockTxsAcquired
+        forM_ completedByBody $ \completedPoint -> do
+          age <- ebAge completedPoint.pointSlotNo
```
ch1bo, line 734: Getting ahold of the immutable tip each time here just for the sake of logging feels weird. Good enough for a prototype, but we probably could just figure out the wall clock time of the announcement once (we need to judge it coming late anyways) and just take the diff to that here.

ch1bo (approved):

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
coot: I asked claude:

>   This is already a known gap, not solved anywhere: InFutureCheck.hs:147-151's judgeHeaderArrival has the identical pattern — recompute
  hardForkSummary/runQuery on every header — with a literal -- TODO cache this in the KnownIntersectionState? Or even in the LedgerDB? sitting unresolved. I
  checked for a shared "cached interpreter" utility (mkInterpreter call sites, GetInterpreter handling) — the only other place that builds one is the
  local-state-query handler (HardFork/Combinator/Ledger/Query.hs:530), which is also on-demand, not cached. So there's nothing to reuse; we'd be writing the
  first instance of this pattern.

This can be fixed locally in `newSlotAge` by reusing the interpreter until it fails, and then caching it again, e.g. 

```hs
  newSlotAge ::
    (IOLike m, HasHardForkHistory blk) =>
    SystemTime m ->
    TopLevelConfig blk ->
    m (ExtLedgerState blk EmptyMK) ->
    m (SlotNo -> m NominalDiffTime)
  newSlotAge systemTime cfg getLedger = do
    cacheVar <- StrictSTM.newTVarIO =<< mkInterp
    pure $ \slot -> do
      interp <- StrictSTM.readTVarIO cacheVar
      onset <- case interpretQuery interp (HardFork.slotToWallclock slot) of
        Right (onset, _slotLength) -> pure onset
        Left _pastHorizon -> do
          interp' <- mkInterp
          StrictSTM.atomically $ StrictSTM.writeTVar cacheVar interp'
          fst <$> HardFork.runQueryThrow (HardFork.slotToWallclock slot) -- via interp', see note below
      pure $ now `diffRelTime` onset
   where
    mkInterp = mkInterpreter . hardForkSummary (configLedger cfg) . ledgerState <$> getLedger
```

@nfrisby what do think?

coot: I feel it doesn't fit into this PR to fix the general problem, but I could do a separate one.

ch1bo: This is not what I worried about, neither what I suggested.

> we probably could just figure out the wall clock time of the announcement once

Was meant to move the slot to wall clock time conversion into where we process (or create) announcements. Then, all the dependent timings we want to trace (when we acquired the corresponding EB body, closure, when we voted on it etc.) would just be relative to that point in time (ideally using a monotonic clock).

Either way, no change needed in this PR.

ch1bo:
Seems like @nfrisby did the same in his #2237 work in 831a3c0d4

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
```diff
@@ -80,7 +88,11 @@ data LeiosVoteSubscription m = LeiosVoteSubscription {getNextVote :: STM m Leios
 -- Holds the contributing voters plus a memoised certificate once the
 -- threshold is crossed.
 data PointState = PointState
-  { psVoters :: !(Map LeiosSeatId (Weight, LeiosSignature))
+  { psPoint :: !(Maybe LeiosPoint)
+  -- ^ Set by 'notePoint' once the announcement for this 'RbHash' is known.
+  -- 'Nothing' until then (e.g. a relayed vote may arrive before we've seen
```
nfrisby: > (e.g. a relayed vote may arrive before we've seen the announcement it refers to)

I'm hoping that won't be true as of https://github.com/input-output-hk/ouroboros-leios/issues/1073, for what it's worth. But I understand the current code isn't yet architected for that. So maybe just add a TODO saying we can hopefully refactor to no longer need need this annotation

Point is: if we allow peers to send us votes for announcements that we don't know exist, then we'll be accepting ~20x (adversarial) votes than could actually be justified (or infinitely more; we need to bound it by _something_, ideally Praos election (whether we've seen the announcement), but we could fallback to one vote per slot, which at least bounds a bit to 20x)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
```diff
@@ -176,7 +188,11 @@ newLeiosVoteState getCommittee = do
                                 , vtTally = totalW
                                 , vtThreshold = threshold
                                 }
-                              pst''.psCert
+                              ((,) <$> pst''.psPoint  <*> pst''.psCert)
+      , notePoint = \rbHash point -> atomically $ do
```
nfrisby: I anticipate we should make sure this can't be executed _after_ this `rbHash` has already been pruned out of the `pointStates` variable. That pruning is something https://github.com/input-output-hk/ouroboros-leios/issues/1073 will add. So please add a TODO/FIXME for now

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
```diff
@@ -371,7 +371,9 @@ runLeiosVoting tracer lcfg chainDB systemTime leiosDB txCache voteState = \case
                   -- point if subsequent votes also come in; consumers (e.g.
                   -- ThreadNet's 'propCertifying') dedupe.
                   case mCert of
-                    Just _ -> traceWith tracer TraceLeiosCertified{rbHash}
+                    Just _ -> do
+                      ebAge <- ebPointAge now point
```
nfrisby, line 374: If this is the only reason you adding the `notePoint` etc stuff, then you don't need any of it, because: the slot of the RbHash that announced the EB has the same slot as the LeiosPoint of the EB it announces. So: the RB and the announcement have the same age.

nfrisby (changes requested):
I'm Requesting Changes because of my comment in `LeiosVoting.hs` --- maybe we can avoid ~75% of this diff?

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
nfrisby: Moreover: this PR adds more information to the `TraceLeiosCertified` tracer. Certs accumulate votes that are in favor of some `RbHash`. So the age of that `RbHash` is the actual thing being recorded.

- Slot numbers are one-to-many with `RbHash`es (just via Praos and what "RB" means).
- `RbHash`es are one-to-one with announcements (in the _current_ Linear Leios design, at least).
- Announcements are many-to-one with `EbHash`es.

Hence: `EbHash`es are many-to-many with slot numbers. (We have functions from RbHash to slot and from announcement to RbHash, but we only have a relation of EbHash and RbHash.)

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
nfrisby: I now see your related Slack message from ~2 hr ago that notes the current absence of the bound :+1:  https://input-output-rnd.slack.com/archives/C074AHSKJF7/p1788861594676119

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
coot: Here we have access to `point :: LeiosPoint`, so it's not a problem.  The problem is in the [hLeiosNotifyClient](https://github.com/IntersectMBO/ouroboros-consensus/pull/2249/changes#diff-71af43f487b6cb728d0757c8c67299a8cb6437ccf8ffa23f131c24d170218689R601-R605) callback. For it to work we need the `point` be available in the result of `addVote`. `addVote` cannot have access to it directly (unless we add `SlotNo` to the vote itself, but that will make vote larger).

ouroboros-consensus/src/ouroboros-consensus/LeiosVoteState.hs
coot: We call this function in two places:
* when we receive `MsgLeiosBlockAnnouncement`
* when we issue our own vote in `runLeiosVoting`

so neither should be after we prune votes.

ouroboros-consensus/src/ouroboros-consensus/LeiosVoting.hs
nfrisby: - The node maintains this piece of state, `PeerState`per upstream peer https://github.com/IntersectMBO/ouroboros-consensus/blob/7abeda6501282c8a2491ccf66c16f051d74fb895/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic/Announcements.hs#L59
- Here it is in `NodeToNode.hs` https://github.com/IntersectMBO/ouroboros-consensus/blob/7abeda6501282c8a2491ccf66c16f051d74fb895/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs#L499

That state is every eb announcement our node has processed from this peer, ie the first (and second, if any!) for each election.

So, whenever the node receives a vote, it should consult this store: honest peer should not be relaying votes for RB hashes that they have not already sent us the announcement for (unsure if they do right now, but they eventually shouldn't). To enforce that, we should compare the arriving vote against the set of announcements we've received from this peer (which will require adding a reverse index to `PeerState`, keyed by `RbHash`, and also pruned).

Once we're doing that validation (which is in the scope of https://github.com/input-output-hk/ouroboros-leios/issues/1073), then pairing the vote with the slot of the announcement it's in favor of should be a trivial responsibility to add to that message validation logic.

-----

That's the reason that we should be able to know which slot contains the RB that the vote is for, even though the vote is not decorated with the slot.

nfrisby: @coot I'll make an attempt at this ^^^ tomorrow---at least laying its groundwork. (It'll be a step towards `ouroboros-leios` Issue 1073). The diff you have right now might work well enough for the testnet, but it's not bulletproof and its' now adding otherwise-unnecessary methods to the ChainDB API, eg.

coot: I just saw your message.  Please check the changes I pushed.  I followed your idea by adding an index by `RbHash`.

PR #2250, opened 2026-08-28
Replace deprecated withMaxSuccess in Roundtrip

dnadales:
The GHC 9.14.1 CI job fails to build `unstable-consensus-testlib`:

```
ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/Serialisation/Roundtrip.hs:665:7: error: [GHC-68441] [-Wdeprecations, -Werror=deprecations]
    In the use of 'withMaxSuccess' ... Deprecated: "Use withNumTests instead"
```

That job resolves QuickCheck 2.16, which deprecates `withMaxSuccess`.
A rename to `withNumTests` does not work, because the Nix shell and the other GHC versions use QuickCheck 2.15, which does not export it.

This change uses `adjustQuickCheckTests`, which the same file already calls at ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/Serialisation/Roundtrip.hs:376.
It sets the test count for that subtree and builds with both QuickCheck versions.


ch1bo (approved):
🙏

PR #2252, opened 2026-08-29
Leios prototype: Fix PastHorizon errors in syncing

ch1bo:
Should fix or at least improve the situation with https://github.com/input-output-hk/ouroboros-leios/issues/1039 and https://github.com/input-output-hk/ouroboros-leios/issues/998. Both issues had various reports with PastHorizon or forecast range errors.

When syncing a testnet on preparing a release I got stuck multiple times and the root cause was the connection going down because of the LeiosNotify handler `throwIO`ing the PastHorizon error result during `judgeHeaderArrival`. I also noticed that we use the immutable tip to do time converstions, but the volatile one would have a slightly longer range and so switched to that.

I did not see nodes using this fix get stuck so far.


ch1bo:
Merging to release and tag a node with this (for further validation on the testnet). Hope to get your feedback on the change still @nfrisby

PR #2256, opened 2026-08-31
Leios prototype: trace the per-point vote tally

johnalotoski:
# Description

The Leios certification margin is currently unobservable. `totalW` is computed in
`LeiosVoteState` and compared against `minCertificationThreshold` but never traced,
and only certified points reach the chain, so a point that stalls below the threshold
leaves no record of how close it came. A diffusion near miss is indistinguishable from
a structural shortfall.

This adds `TraceLeiosTally {rbHash, seatId, weight, tally, threshold}`. The tally is
surfaced through `AddVoteResult`'s `Added` and emitted by the callers in `LeiosVoting`
and `NodeToNode`, as they already do for certification, because the update itself runs
in STM. In `NodeToNode` it replaces the `TODO` about keeping track of the running
tally.

It fires only on `Added`, so once per distinct vote rather than once per arrival:
duplicates short-circuit as `AlreadyKnown` before the map update, with no tally to
report. On one relay that is roughly a 25x reduction in lines, the ratio being the
peer count delivering the same vote. Each line carries the adding seat and its own
weight next to the tally, so a consumer can attribute the tally without a second
source for seat weights.

`PointState` gains `psTotal :: !Weight`, maintained incrementally. Reporting the tally
would otherwise force `totalW` on every accepted vote, including post-certification
votes where it had been dead code, and summing the voter map is linear in the
committee. This also removes the pre-existing quadratic.

## Testing

`prop_tallyAccumulates` in `Test/LeiosVoteState.hs` pins the reported tally against
the running sum of the reported weights. It is the equivalence test between the
incremental `psTotal` and the sum over the voter map it replaces.

### DIFF
--- a/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -575,15 +575,31 @@ mkHandlers
                     traceWith tracer $ MkTraceLeiosPeer $ "MsgLeiosVotes " <> show vs
                     forM_ vs $ \vote -> do
                       result <- addVote vote
-                      -- TODO: Keep track of the running tally (even after certification)
                       traceWith kernelTracer TraceLeiosVoteAcquired{vote}
                       -- A remote vote can be the one that tips this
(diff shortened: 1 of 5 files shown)


ch1bo:
Cherry picked, but then changed to re-use the `VoteAcquired` trace in https://github.com/IntersectMBO/ouroboros-consensus/pull/2271. Now released as `prototype-2026w35` on ouroboros-leios

PR #2251, opened 2026-08-28
Serve GetGenesisConfig without sgExtraConfig

dnadales:
## The break

`cardano-ledger-shelley` 1.19.0.0 added `sgExtraConfig` to `ShelleyGenesis` and grew the CBOR record from 15 fields to 16, with no version guard ([`e68ebf663`](https://github.com/IntersectMBO/cardano-ledger/commit/e68ebf663)).
The node-to-client version did not change.

`genesisConfigEnDecoding` picks the ledger codec for every version at or above `ShelleyNodeToClientVersion13`.
Consensus 3.0.1 and 4.x both offer v13, v14 and v15.
Under 3.0.1 those three mean 15 fields. Under 4.x they mean 16.
So the negotiated version no longer determines the wire format.

A client that links the new ledger cannot read the reply of a node that links the old one.
The decoder waits for bytes that never arrive, with no exception, so the failure is silent.
cardano-wallet hits this against node 11.0.1. It issues `GetGenesisConfig` to derive slot length, epoch length, active slot coefficient and security parameter, and its tip sync client then stops making progress.

## The fix

`genesisConfigEnDecoding` keeps its gate at `ShelleyNodeToClientVersion13`. Only the v13-and-above tier changes.

- Encode 15 fields again, with the current `PParams` encoding.
- Decode 15 or 16 fields.

The decoder half repairs a client that talks to an already deployed node 11.1.0, with no node upgrade.

`encodeLegacyShelleyGenesis` cannot serve this tier. It also writes 15 fields, but with the `LegacyPParams` encoding from before node 10.5.

Dropping `sgExtraConfig` loses nothing. `compactGenesis` erases it, next to `sgInitialFunds` and `sgStaking`. Each regenerated golden loses exactly one trailing `0x80`, the `SNothing`.

24 golden files change: shelley v13 to v15, and cardano `NodeToClientVersion17` to `19`. Nothing below v13 changes.

## No new node-to-client version

`latestReleasedNodeVersion` is `NodeToClientV_23`, which is also the maximum, so there is no version above v15 to tell the two formats apart. A new one needs an `ouroboros-network` release to CHaP.

It fixes nothing extra either. Node 11.1.0 already sends 16 fields at `NodeToClientV_23`, and a new version does not change what a deployed node sends.

## Why draft

This changes what `NodeToClientV_23` means for `GetGenesisConfig`. Open questions:

[...]

PR #2258, opened 2026-08-31
Decode ShelleyGenesis from 15 fields only

dnadales:
Follow-up to #2251, addressing @lehins' review.

- Use the ledger codec for node-to-client versions above `ShelleyNodeToClientVersion15`.
- Decode exactly 15 fields with `decodeRecordNamed`, restoring support for indefinite length encoding.
- Set `sgExtraConfig` to `SNothing` instead of reading an optional 16th field, which changed an existing query with no version bump.
- Rename `decodeShelleyGenesisWithOptionalExtraConfig` to `decodeShelleyGenesisNoExtraConfig` and update the pending changelog fragment.

This also needs backporting to `ouroboros-consensus-4.x-backports`, where 4.2.1.0 shipped the previous behaviour.

### DIFF
--- a/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
+++ b/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
@@ -1115,9 +1115,11 @@ genesisConfigEnDecoding ::
   ShelleyNodeToClientVersion ->
   (CompactGenesis -> Encoding, Decoder s CompactGenesis)
 genesisConfigEnDecoding v
+  | v > ShelleyNodeToClientVersion15 =
+      (toCBOR, fromCBOR)
   | v >= ShelleyNodeToClientVersion13 =
       ( encodeShelleyGenesisNoExtraConfig . getCompactGenesis
-      , compactGenesis <$> decodeShelleyGenesisWithOptionalExtraConfig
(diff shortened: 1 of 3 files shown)


lehins (approved):
👍

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -1115,9 +1115,11 @@ genesisConfigEnDecoding ::
   ShelleyNodeToClientVersion ->
   (CompactGenesis -> Encoding, Decoder s CompactGenesis)
 genesisConfigEnDecoding v
+  | v > ShelleyNodeToClientVersion15 =
+      (toCBOR, fromCBOR)
```
jasagredo, line 1119: Doesn't this mean that version 15 will encode the extraConfig? Aren't we breaking version 15 wrt tools that used that version already before 11.1?

lehins: Note `>`, not `>=`

jasagredo: Aha indeed!

jasagredo (approved):

PR #2259, opened 2026-09-01
Use cardano-config and cardano-keys

jasagredo:
The tools read the node configuration with cardano-config and, in
db-synthesizer, the forging credentials with cardano-keys.


docs/website/contents/references/consensus_tools.md
```diff
@@ -351,17 +367,24 @@ length cheaply, e.g. as input for benchmarks.
...
 ### Requirements
 
-- `--config FILE` — a node configuration file. Only a few values regarding
-  geneses and protocol are actually required, so a configuration stub is
-  possible; for the expected key-value pairs see the `NodeConfigStub` type and
-  its deserialization in `Cardano.Tools.DBSynthesizer.Orphans`.
+- `--config FILE` — a node configuration file, exactly as `cardano-node` takes
+  it: it is parsed by the shared `cardano-config` package, which also loads the
+  genesis file of every era the configuration names (their paths are relative
+  to the configuration file's directory).
```
dnadales, line 373: Should we note here that the genesis hashes are now mandatory? `cardano-config` parses each genesis with `requiredFieldWith` for both the file and its hash (`Cardano/Configuration/File/Protocol.hs:55-59`, applied at `:116` and `:153-155`), so a configuration that names `ShelleyGenesisFile` without `ShelleyGenesisHash` no longer resolves. `AlonzoGenesisHash` and `ConwayGenesisHash` go further, from being ignored entirely to being required.

The published configurations all carry the four, so I guess the people this reaches are the ones who took the advice this page used to give, that "a configuration stub is possible". Our own fixture had no hashes and gained five in this PR.

What about something like, on the `--config` bullet:

> ... which also loads the genesis file of every era the configuration names (their paths are relative to the configuration file's directory). Each genesis file must come with its hash, so even a stub configuration needs `ByronGenesisHash`, `ShelleyGenesisHash`, `AlonzoGenesisHash` and `ConwayGenesisHash`.

I wonder if this deserves a changelog fragment as well. It seems to be the one change here that stops a configuration that worked before from working.

ouroboros-consensus-cardano/app/DBSynthesizer/Parsers.hs
```diff
@@ -134,22 +98,18 @@ parseEpochLimit =
         <> help "Amount of epochs to process"
     )
 
-parseForce :: Parser Bool
-parseForce =
-  switch
+-- | @-f@ and @-a@ pick the open mode; without either of them, the ChainDB
+-- directory is expected not to exist yet.
```
dnadales, line 102: Should we keep the note that was here before the last commit? The one that said these have to be `flag'` and not `switch`, because a `switch` succeeds whether or not its flag is present, so the first alternative always wins.

That is the whole bug this function fixes, and `switch` is what anyone would reach for next time. Without the note, `-f` silently force-overwriting every ChainDB comes back and nothing in the file says why it must not.

changelog.d/20260916_120000_javier.sagredo_byron_version_defaults.md
```diff
@@ -0,0 +1,21 @@
...
+  when nothing else specifies them.
+
+  These used to come from the `LastKnownBlockVersion-Major`, `-Minor`, `-Alt`
+  and `ApplicationVersion` keys. `cardano-config` retired all four as obsolete,
+  stripping them from a configuration it migrates, on the grounds that the
+  values are consensus defaults, so this is where they now live.
+
+  Only a block producer reads them: they reach `headerProtocolVersion` and
+  `headerSoftwareVersion` of the blocks it forges, and nothing validates against
+  them.
```
dnadales, line 21: I see three fragments went away in the last commit, and I wonder about two of them. `unstable-cardano-tools` is out of scope for the changelog rule (`how_to_make_a_release.md:28`, and it is not in the list in `scripts/ci/check-changelogs.sh`), so this is a judgement call rather than a rule.

The two I would keep are the protocol-version gate and the Dijkstra genesis gate. Both change what a tool does with a configuration file someone already has: the version now stops at the era before the experimental one, and a named `DijkstraGenesisFile` is ignored when the flag is off. Neither is visible from the diff of a tool's own source, and someone who hits either will look for exactly that note.

Do we read "not required" as "not wanted" here? I guess the tools have users who read release notes even though the sublibrary is unstable.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -164,23 +166,29 @@ analyse dbaConfig args =
...
     chainDBTracer <- mkVerboseTracer lock verbose
     analysisTracer <- mkVerboseTracer lock True
-    LSMConfig{lsmConfigExportPath} <- mkLSMConfig args
     lsmSalt <- fst . genWord64 <$> newStdGen
-    ProtocolInfo{pInfoInitLedger = genesisLedger, pInfoConfig = cfg} <-
-      mkProtocolInfo args
+    (ProtocolInfo{pInfoInitLedger = genesisLedger, pInfoConfig = cfg}, configBackend) <-
+      mkProtocolInfoAndBackend args
+    -- The command line takes precedence over the configuration file.
+    backend <- case ldbBackend <|> configBackend of
```
dnadales, line 173: Aren't we changing what an existing `--lsm` invocation does? Selection is whole or nothing here, so `--lsm` drops `configBackend` together with the configuration's `DatabasePath` and `ExportPath`, and the command-line path builds its options from the defaults alone (`app/DBAnalyser/Parsers.hs:91-96`). Before this PR `mkLSMConfig` read `LedgerDB.LSMExportPath` whatever the command line said, so `--lsm --store-ledger` against a configuration with an export path exported snapshots. Now the same command exports nothing, and `--lsm-export` sends them to the fixed `lsm-exported` rather than the configured directory.

I wonder if the command line could override only the fields it names, rather than the whole selection. Something like: when both sides pick LSM, start from the configuration's options and let `--lsm-export` and `--lsm-no-cache` set their own fields.

If choosing wholesale is what we want, could we say so in the docs? The `--lsm-export` bullet describes the new rule accurately, but nothing tells a reader that their existing command now means something else.

One related thing, in case it matters for the same person: `cardano-config` folds the old flat `LSMExportPath` into the nested form only when `Backend` is also `"V2LSM"` (`Migrate.hs:226`), and `checkUnknownKeys` looks only at the top level (`Lint.hs:107-112`). So a configuration that sets just the export path loses it with no warning that names the key.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Run.hs
```diff
@@ -55,116 +65,113 @@ import qualified Ouroboros.Consensus.Storage.LedgerDB as LedgerDB
...
+  Cfg.NodeConfiguration ->
+  CardanoHardForkTriggers ->
+  LeaderCredentials ->
+  CardanoProtocolParams StandardCrypto
+protocolParams nc triggers leaderCredentials =
+  CardanoProtocolParams
+    ProtocolParamsByron
+      { byronGenesis = Cfg.byronGenesisConfig nc
+      , -- Not modelled by cardano-config; the node's own default is to leave the
+        -- genesis-imposed threshold alone.
```
dnadales, line 134: "the genesis-imposed threshold"?

I may be reading this wrong, but `byronPBftParams` does `fromMaybe defaultPBftSignatureThreshold threshold` (`Byron/Node.hs:273`), so `Nothing` selects consensus' own 0.22 rather than leaving anything from the genesis in place. The two fields that do come from the genesis there are `k` and the genesis-key count.

What about something like:

```haskell
, -- cardano-config removed the PBftSignatureThreshold key, so there is
  -- nothing to read: 'Nothing' makes 'protocolInfoByron' use consensus'
  -- own 'defaultPBftSignatureThreshold' of 0.22.
  byronPbftSignatureThreshold = Nothing
```

The value itself is unchanged for anyone who did not set the key, since the old parser used `.:?` and also produced `Nothing`. For someone who did set it, the only signal is the generic `MigratedToCurrentFormat` warning, which never names the key, and I guess that is cardano-config's side rather than ours.

docs/website/contents/references/consensus_tools.md
```diff
@@ -351,17 +367,24 @@ length cheaply, e.g. as input for benchmarks.
...
   (`--bulk-credentials-file`, a JSON array of `[opcert, VRF key, KES key]`
-  triples). The genesis must give the corresponding pools enough stake to be
-  elected.
+  triples). Byron-era credentials (`--byron-delegation-certificate` with
+  `--byron-signing-key`) and a KES agent (`--shelley-kes-agent-socket` instead
+  of `--shelley-kes-key`) work too. These are `cardano-node`'s own flags —
+  db-synthesizer takes them from `cardano-config`, so they are spelled and
+  documented exactly as the node spells them, and the files they name are
+  decoded by `cardano-keys`. Each operational certificate must name the KES key
+  it is paired with, or the tool refuses it rather than forging blocks the
```
dnadales, line 385: Does this hold on the agent path? `checkOpCertKesKey` runs in the KES key file branch (`Credentials.hs:117`) and for every bulk entry (`:143`), but the `KESAgentSocketPath` branch at `:122` goes straight to `PraosCredentialsAgent` with no check. Since we introduce `--shelley-kes-agent-socket` two sentences above, I read this sentence as a promise that covers it.

What about something like:

> Each operational certificate must name the KES key it is paired with, or the tool refuses it rather than forging blocks the certificate does not authorise. With a KES agent the tool cannot check this, because the certificate it forges with comes from the agent rather than from `--shelley-operational-certificate`.

Two smaller things in the same paragraph, while we are here:

- The heading says `Requirements`, but the credentials are optional in practice. With none supplied `readLeaderCredentials` yields no forgers (`Credentials.hs:105`), and the run prints "no forgers found" and exits 0. Should we say so, so that nobody reads an empty ChainDB as success?
- "either as separate files ... or in bulk" reads as exclusive, but `Credentials.hs:65` does `shelley <> bulk`, so both together is legal and gives both sets of forgers. I guess the all-three-or-none rule for the `--shelley-*` options deserves a clause here too.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Credentials.hs
```diff
@@ -0,0 +1,249 @@
...
+        -- The unsound variant: the KES signing key sits in a file on disk
+        -- instead of never leaving a KES agent's memory.
+        CLI.KESKeyFilePath kesFile -> do
+          kesSignKey <- readTextEnvelope kesFile
+          checkOpCertKesKey (certFile <> " with " <> kesFile) opCert kesSignKey
+          pure $
+            PraosCredentialsUnsound
+              (opCertOf opCert)
+              (Keys.unsoundPureKesSigningKey kesSignKey)
+        CLI.KESAgentSocketPath socketPath ->
```
dnadales, line 122: Since this is the first time db-synthesizer can take a KES agent, should we give the agent a tracer that reaches the user? `mkForgers nullTracer` in `DBSynthesizer/Run.hs:190` discards every `KESAgentClientTrace`, so an agent socket that nothing is listening on gives:

```
forger count: 1
--> forged and adopted 0 blocks; reached <slot>
--> done; result: ForgeResult {resultForged = 0}
```

and exit 0. `mkDynamicHotKey` starts from `KESKeyPoisoned` and fills the key on another thread (`HotKey.hs:288,347-357`), so the forger still counts. Every slot then fails with `KESKeyAlreadyPoisoned`, and `Forging.hs:197-199` folds `ForgeStateUpdateError`, `CannotForge` and `NotLeader` into one `"NoLeader"` string that `Forging.hs:132` drops. `runKESAgentClient` reports the connection failure only by tracing it (`AgentClient.hs:166`) and then retries every ten seconds.

Something like handing that call a tracer backed by `stdoutTracer`, rather than `nullTracer`, would make the new option diagnosable without touching `Forging.hs`.

I wonder as well whether a KES key that cannot sign deserves the same treatment as one that will not parse. We already exit 1 with a message when `readLeaderCredentials` fails (`Run.hs:107-109`), and a key whose validity window has not started looks like the same class of user error to me. That case is easy to hit here, since the tool forges from slot 0 while a real operational certificate starts at a later KES period.

One more, probably out of scope: under `-b/--blocks` the loop's test is `(b ==) . forged` (`Forging.hs:125`), so a run that forges nothing never terminates at all.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/Credentials.hs
```diff
@@ -0,0 +1,249 @@
...
+  CLI.Credentials -> ExceptT String IO [ShelleyLeaderCredentials StandardCrypto]
+readShelley creds =
+  case ( CLI.shelleyOperationalCertificate creds
+       , CLI.shelleyVRFKey creds
+       , CLI.shelleyKES creds
+       ) of
+    (SNothing, SNothing, SNothing) -> pure []
+    (SNothing, _, _) -> throwE $ missingOption "shelley-operational-certificate"
+    (_, SNothing, _) -> throwE $ missingOption "shelley-vrf-key"
+    (_, _, SNothing) -> throwE $ missingOption "shelley-kes-key"
```
dnadales, line 108: Should this one name the agent socket as well? `CLI.shelleyKES` is satisfied by either `--shelley-kes-key` or `--shelley-kes-agent-socket`, and we handle both a few lines below at `:115` and `:122`, so somebody who passes the certificate, the VRF key and an agent socket gets told to supply a key file they do not want.

What about something like:

```haskell
    (_, _, SNothing) ->
      throwE $ missingOption "shelley-kes-key or --shelley-kes-agent-socket"
```

which comes out as "To forge blocks, --shelley-kes-key or --shelley-kes-agent-socket must also be specified". Passing the second flag inside the argument is a bit of a trick, so a separate message for this branch reads better if you prefer.

The other four `missingOption` branches look right to me, since each of those fields has exactly one flag.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -164,23 +166,29 @@ analyse dbaConfig args =
...
+      Just backend -> pure backend
+      Nothing ->
+        throwConfigError $
+          "no LedgerDB backend was selected on the command line and none could"
```
dnadales, line 177: Can we reach this branch? `configBackend` is `Nothing` only when `Cfg.backendSelector` is `SNothing` (`Block/Cardano.hs:174-175`), and cardano-config's defaults layer always sets `Backend: "V2InMemory"` (`defaults/StorageConfig.json:4`), which is what `test_ledgerDBBackendDefault` pins. The other way to get `Nothing` is the class default of `mkProtocolInfoAndBackend`, and db-analyser only ever builds `CardanoBlockArgs`.

The message also points the opposite way from the docs, which say at `consensus_tools.md:145-147` that omitting both flags is not an error and leaves you with `--in-mem`.

If we keep it as a guard for a future block type, should it say so and drop the instruction? Something like:

```haskell
      Nothing ->
        -- Unreachable for the Cardano instance, whose configuration always
        -- resolves a backend; this guards a block type whose Args carry none.
        throwConfigError
          "no LedgerDB backend was selected and none could be determined from the configuration."
```

docs/website/contents/howtos/adding_an_era.md
```diff
@@ -89,7 +89,9 @@ be adding is the Alonzo era, which comes after the Mary era.
   add the extra arguments needed for `protocolInfoCardano` to the
   `ProtocolCardano` constructor. Update `protocolInfo` accordingly.
 
-* In `Cardano.Node.Types`, add a field for the new era into the `NodeHardForkProtocolConfiguration` record.
+* In `cardano-config`, add a `Test<Era>HardForkAtEpoch` field for the new era to
+  the `TestingConfiguration` record, and in `Cardano.Tools.Config`, map it onto
+  the new era's trigger in `mkHardForkTriggers`.
```
dnadales, line 94: Since this line now sends people to `cardano-config` for the testing keys, should we say somewhere that the old spellings stop working? Two of them go quiet rather than loudly:

- `TestEnableDevelopmentHardForkEras` is unknown to `cardano-config`. It is not renamed (`Migrate.hs:87-100`) and not in `removedFields` (`:128-138`), so a legacy file gets `WARNING: unrecognised configuration key(s): ... (ignored)` and the experimental era falls back to off. Our own fixture set it to `true` before this PR.
- The repair is the part I would not guess. `ExperimentalHardForksEnabled: true` on its own fails resolution: `Testing.hs:114-122` requires a `DijkstraGenesisFile` and its `DijkstraGenesisHash` alongside it.

The `Test<Era>HardForkAtEpoch` keys survive, but they live under `TestingConfig` now, which this page does not say either, nor does `references/miscellaneous/era_transition_governance.md:113-121`.

What about something like, after this bullet:

> The testing keys live under `Configuration.TestingConfig` in the current configuration format. `TestEnableDevelopmentHardForkEras` is gone: its replacement is `ExperimentalHardForksEnabled`, which also requires a `DijkstraGenesisFile` and a `DijkstraGenesisHash` when it is true.

A section in the tools reference would work too, wherever you would rather a configuration author find it.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Types.hs
```diff
@@ -50,9 +53,31 @@ data Limit = Limit Int | Unlimited
...
+  -- LSM backend exports snapshots as it takes them. When 'Nothing', snapshots
+  -- are not exported.
+  , lsmNoDiskCache :: Bool
+  -- ^ Bypass the OS page cache for UTxO table reads/writes (instead of caching
+  -- all). Intended for benchmarking.
+  }
+
+-- | The directory holding the working LSM database, used when the command line
+-- selects the LSM backend and when the node configuration file does not set
+-- @LedgerDB.LSMDatabasePath@.
```
dnadales, line 74: `@LedgerDB.LSMDatabasePath@`? The code reads `LedgerDB.Backend.LSM.DatabasePath` (`Block/Cardano.hs:202`, and the test at `NodeConfig.hs:292`), so this looks like the pre-nesting spelling. Same nesting as the `ExportPath` in the haddock two definitions below, which does get it right.

ouroboros-consensus-cardano/test/tools-test/Main.hs
```diff
@@ -8,19 +10,25 @@ import qualified Cardano.Tools.DBSynthesizer.Run as DBSynthesizer
...
 import Test.Util.TestEnv
 
-nodeConfig, chainDB :: FilePath
+nodeConfig, chainDB, bulkCredentials :: FilePath
 nodeConfig = "ouroboros-consensus-cardano/test/tools-test/disk/config/config.json"
 chainDB = "ouroboros-consensus-cardano/test/tools-test/disk/chaindb"
+bulkCredentials = "ouroboros-consensus-cardano/test/tools-test/disk/config/bulk-creds-k2.json"
 
+-- | A tenth of an epoch, then a further 8192 slots: enough for both steps to
+-- forge a few hundred blocks, small enough for the analysis to stay quick. The
```
dnadales, line 25: "a few hundred blocks"? `expectedForgedCreate` is 2189 a few lines down, and the append step is the one that forges 407.

While I am here, I wonder about pinning the counts at all. The reasoning in the haddock convinces me that a change in them is worth noticing, but the test then fails in whichever unrelated PR moves the leader schedule or the ledger, and the person who has to re-baseline is not the person who chose the number. Did you consider asserting a range, or equality between the two runs, so that the failure lands on whoever changes the forging path?

Also, this PR is where the fixture moves to Conway (`TestConwayHardForkAtEpoch: 0`) and where the counts first appear. Do you know what they were before the era changed, out of interest?

ouroboros-consensus.cabal
```diff
@@ -1803,8 +1803,8 @@ library unstable-cardano-tools
   visibility: public
   hs-source-dirs: ouroboros-consensus-cardano/src/unstable-cardano-tools
   exposed-modules:
-    Cardano.Api.Any
-    Cardano.Node.Types
+    Cardano.Tools.Config
+    Cardano.Tools.Credentials
     Cardano.Tools.DBAnalyser.Analysis
     Cardano.Tools.DBAnalyser.Analysis.BenchmarkLedgerOps.FileWriting
     Cardano.Tools.DBAnalyser.Analysis.BenchmarkLedgerOps.Metadata
```
dnadales, line 1810: Should `Cardano.Tools.DBAnalyser.Block.Byron` and `.Block.Shelley` move to `other-modules` now? Both are `module X () where` after this PR, and they are imported only for their `HasAnalysis` instances, so nothing can use them through the public library any more. Lines 1812 and 1814, which the diff does not reach.

dnadales (approved):
Nice cleanup ❤️ No blocking issues but it might be worthwhile to address them 👍

PR #2260, opened 2026-09-01
Do not crash voting thread when an exception is thrown

nbacquey:
Some fixes from our voting thread debugging work with @adithyaov

### DIFF
--- a/changelog.d/20260828_115116_agustin.mista_extra_chaindb_handles.md
+++ b/changelog.d/20260828_115116_agustin.mista_extra_chaindb_handles.md
@@ -1,0 +1,3 @@
+### Non-Breaking
+
+- Added ChainDB handles for Peras voting, certificate-inclusion, and time-resolution queries.
--- a/changelog.d/20260908_121941_agustin.mista_wire_dijkstra_committe.md
+++ b/changelog.d/20260908_121941_agustin.mista_wire_dijkstra_committe.md
@@ -1,0 +1,27 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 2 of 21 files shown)


tbagrel1:
I'm a bit unsure about the change. What kind of exceptions did you catch while testing?

If it's something that can happen under relatively "normal" circumstances, then we should probably use error values instead of exceptions in the corresponding function so we don't have to use throw/catch.

If it's something sufficiently rare and unrecoverable, I think we would like to keep a stack trace and keep the thread crashed?

Even in the case where we keep the throw/catch logic, I would prefer if we catch specific exception types, not the most generic one.

agustinmista:
I agree with @tbagrel1, I think it really depends on which exceptions are we trying to catch here.

IIRC, there was a discussion at some point about using `throwSTM` inside the handle-returning methods in the ChainDB, e.g., `getPerasVotingView` and `getPerasCertInclusionView`. These are currently discharging a possible failure in `resolveRoundNo`. Such failure has type `PerasEpochContextNotFoundForRound` (which implements `Exception`), so we could in principle limit the scope of the `catch` to only those (provided that they are the ones you're seeing in practice).

nbacquey:
I agree with wanting to catch a specific subset of `Exception`. What we did catch during the testing were indeed instances of `PerasEpochContextNotFoundForRound`. Those are by definition round-dependent, and I think it makes sense to *not* crash the voting thread when they happen.

nbacquey:
Closing this PR, as we are switching to explicit error types instead of using `throwSTM`

PR #2262, opened 2026-09-02
db-truncater: support Leios chains

dnadales:
This PR adapts db-truncater to Leios. Part of https://github.com/input-output-hk/ouroboros-leios/issues/1023

After it truncates the ImmutableDB, the tool cuts `leios.db` back to the new tip slot, deletes the transactions that no endorser block references, and vacuums the file.
`--stubbed-leios-db` skips all of that, for a chain that carries no Leios certificate.

The LeiosDb work is three standalone functions in `LeiosDemoDb.SQLite`, each with a unit test. `tools-test` now covers db-truncater end to end, which it did not before.

### DIFF
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -11,6 +11,7 @@ module DBAnalyser.Parsers
 import Cardano.Tools.DBAnalyser.Analysis
 import Cardano.Tools.DBAnalyser.Block.Cardano
 import Cardano.Tools.DBAnalyser.Types
+import Cardano.Tools.LeiosDb (LeiosDbSource (..))
 import qualified Data.Foldable as Foldable
 import Options.Applicative
 import Ouroboros.Consensus.Block (SlotNo (..), WithOrigin (..))
@@ -63,9 +64,9 @@ parseDBAnalyserConfig =
     <*> parseNoLeiosDb
 
 -- | Run with an empty in-memory LeiosDb, rather than the node's @leios.db@.
-parseNoLeiosDb :: Parser Bool
+parseNoLeiosDb :: Parser LeiosDbSource
 parseNoLeiosDb =
-  switch $
+  flag NodeLeiosDb NoLeiosDb $
     mconcat
       [ long "no-leios-db"
       , help $
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -4,7 +4,7 @@
 module DBAnalyser.Parsers
   ( parseCmdLine
   , parseCardanoArgs
-  , parseStubbedLeiosDb
+  , parseNoLeiosDb
   , CardanoBlockArgs
   ) where
 
@@ -60,14 +60,14 @@ parseDBAnalyserConfig =
             , help "use v2 LSM backend"
             ]
       ]
-    <*> parseStubbedLeiosDb
+    <*> parseNoLeiosDb
 
 -- | Run with an empty in-memory LeiosDb, rather than the node's @leios.db@.
-parseStubbedLeiosDb :: Parser Bool
-parseStubbedLeiosDb =
+parseNoLeiosDb :: Parser Bool
+parseNoLeiosDb =
   switch $
     mconcat
-      [ long "stubbed-leios-db"
+      [ long "no-leios-db"
       , help $
           "Do not use the leios.db file under the --db path. Pass this for a "
             <> "chain that holds no block with a Leios certificate. Without "
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -4,6 +4,7 @@
 module DBAnalyser.Parsers
   ( parseCmdLine
   , parseCardanoArgs
+  , parseStubbedLeiosDb
   , CardanoBlockArgs
   ) where
 
@@ -70,9 +71,8 @@ parseStubbedLeiosDb =
       , help $
           "Use an empty in-memory LeiosDb, rather than the leios.db file under "
             <> "the --db path. Pass this for a chain that holds no block with a "
-            <> "Leios certificate, and hence no endorser block to resolve. "
-            <> "Without this flag, the tool refuses to start when it finds no "
-            <> "such file."
+            <> "Leios certificate. Without this flag, the tool refuses to start "
+            <> "when it finds no such file."
       ]
 
 parseSelectDB :: Parser SelectDB
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -69,10 +69,9 @@ parseStubbedLeiosDb =
     mconcat
       [ long "stubbed-leios-db"
       , help $
-          "Use an empty in-memory LeiosDb, rather than the leios.db file under "
-            <> "the --db path. Pass this for a chain that holds no block with a "
-            <> "Leios certificate. Without this flag, the tool refuses to start "
-            <> "when it finds no such file."
+          "Do not use the leios.db file under the --db path. Pass this for a "
+            <> "chain that holds no block with a Leios certificate. Without "
+            <> "this flag, the tool refuses to start when it finds no such file."
       ]
 
 parseSelectDB :: Parser SelectDB
--- a/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
+++ b/ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
@@ -4,7 +4,7 @@
 module DBAnalyser.Parsers
   ( parseCmdLine
   , parseCardanoArgs
-  , parseNoLeiosDb
+  , parseLeiosDbSource
   , CardanoBlockArgs
   ) where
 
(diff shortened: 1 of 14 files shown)


ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
```diff
@@ -68,11 +69,9 @@ parseStubbedLeiosDb =
     mconcat
       [ long "stubbed-leios-db"
```
ch1bo: Could: find a better option name

I know you have not touched it in this PR, but I was not sure what "stubbed" means here. I'd like to suggest `--no-leios-db` or `--skip-leios-db`

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/LeiosDb.hs
```diff
@@ -0,0 +1,51 @@
...
+          "No LeiosDb at "
+            <> path
+            <> ". A block that carries a Leios certificate has an empty body, "
+            <> "and the transactions that it puts on the chain are in the "
+            <> "endorser block that it certifies, which the LeiosDb holds. "
+            <> "Pass --stubbed-leios-db if this chain holds no such block."
+      pure (Just path)
+
+-- | Open the node's LeiosDb, or an empty in-memory one for the stub.
+openLeiosDb :: Bool -> FilePath -> IO (LeiosDbHandle IO)
```
ch1bo: Should: avoid `Bool` flags

At least avoid boolean blindness by making this a type or avoid the argument altogether.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -155,7 +152,7 @@ analyse dbaConfig args =
     lsmSalt <- fst . genWord64 <$> newStdGen
     ProtocolInfo{pInfoInitLedger = genesisLedger, pInfoConfig = cfg} <-
       mkProtocolInfo args
-    leiosDbHandle <- openLeiosDb
+    leiosDbHandle <- openLeiosDb stubbedLeiosDb dbDir
```
ch1bo: Follow-up to previous comment: Here we have a much better context and we could just keep the case here.

Pro: no boolean arguments, less indirection
Con: ?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -484,6 +489,60 @@ sqlFilterMissingTxs conn txHashes =
         txHash <- MkTxHash <$> DB.columnBlob stmt 0
         loop (txHash : acc)
 
+-- | Delete the EBs announced after the given slot.
+--
+-- For internal tooling.
+truncateLeiosDbAfterSlot :: HasCallStack => FilePath -> SlotNo -> IO ()
```
ch1bo, line 495: Should: Not pass `FilePath`, but `Conn` like the other functions in this module

Consequently, also move `withExistingDb` into the tools package so it is not confused with the `withLeiosDb` in `../Common.hs`

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -484,6 +489,60 @@ sqlFilterMissingTxs conn txHashes =
...
+truncateLeiosDbAfterSlot dbPath (SlotNo slot) =
+  withExistingDb dbPath $ \db ->
+    -- One transaction, so a crash cannot leave an EB that is still announced
+    -- but has no body.
+    dbWithTransactionAs "BEGIN IMMEDIATE" db $
```
ch1bo, line 500: Could: use the even fancier `dbWithWriteTransaction`

This will do retries and trace if stuck. Maybe overkill for the offline tool use.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -484,6 +489,60 @@ sqlFilterMissingTxs conn txHashes =
...
+  -- same EB can be announced in several slots. So a body is deleted only when
+  -- the truncation drops every announcement of that EB.
+  deletes =
+    unlines
+      [ "DELETE FROM ebTxs WHERE ebHashBytes NOT IN (" <> remainingHashes <> ");"
+      , "DELETE FROM ebsMissingTxs WHERE ebHashBytes NOT IN (" <> remainingHashes <> ");"
+      , "DELETE FROM ebs WHERE ebSlot > " <> show slot <> ";"
+      ]
+
+  remainingHashes = "SELECT ebHashBytes FROM ebs WHERE ebSlot <= " <> show slot
```
ch1bo: Should: re-use the result across the transaction

Not sure if this happens automatically if we put it into each statement

ouroboros-consensus/test/consensus-test/Test/LeiosDemoDb.hs
```diff
@@ -99,6 +112,20 @@ withFreshDb SQLite action = do
     snd
     (action . fst)
 
+-- | Create a fresh SQLite database and hand its path to the action, which
+-- 'truncateLeiosDbAfterSlot' needs.
+withFreshSQLiteFile :: (FilePath -> LeiosDbHandle IO -> IO a) -> IO a
+withFreshSQLiteFile action = do
```
ch1bo, line 118: Should: Change this to be a raw `withFreshSQLite :: (Conn -> IO a) -> IO a`

Following my previous suggestion on using `Conn` in your new functions.

ch1bo (approved):
Will need to change after / with https://github.com/input-output-hk/ouroboros-leios/issues/969, but looks good!

ouroboros-consensus-cardano/app/DBAnalyser/Parsers.hs
dnadales: Agree. I could change it for all the tools that use this name :+1:

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/LeiosDb.hs
dnadales: Yeah, I was hesitant whether to introduce an isomorphic type for this, in the end I decided to to keep it short. But I agree, let's change it to a `data` type.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
dnadales: I kept `dbWithTransactionAs "BEGIN IMMEDIATE"`. It does seem like overkill IMHO: dbWithWriteTransaction needs a `Conn`, which includes 16 statements it never uses; and db-tuncater should not block indefinitedly (as it would with dbWithWriteTransaction)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
dnadales: No, SQLite does not reuse every subquery result. 
`EXPLAIN` shows the `NOT IN` list held in an ephemeral table, and the docs say an ephemeral table is deleted when its cursor closes, which happens when the statement is reset or finalised (https://www.sqlite.org/opcode.html).
So the next DELETE builds the same list again.
(Genie tested it with a local benchmark)

Looking at it moved the fix elsewhere, though: 7d5adf4cd986ae3c7e70681de43450b16e7dad12 truncates by the dropped EB hashes, instead of the kept ones :D

ouroboros-consensus/test/consensus-test/Test/LeiosDemoDb.hs
dnadales: This follows the Conn question above. I kept FilePath there, so the tests still need a path. 
withExistingDb is now withExistingLeiosDbFile, which removes the clash with withLeiosDb.

A Conn-only helper would not fit these tests in any case.
They build their fixture EBs through LeiosDbConnection, so they need the handle as well as the path.

Does this make sense?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
```diff
@@ -155,7 +152,10 @@ analyse dbaConfig args =
     lsmSalt <- fst . genWord64 <$> newStdGen
     ProtocolInfo{pInfoInitLedger = genesisLedger, pInfoConfig = cfg} <-
       mkProtocolInfo args
-    leiosDbHandle <- openLeiosDb
+    leiosDbHandle <-
+      leiosDbPath leiosDbSource dbDir >>= \case
+        Nothing -> newLeiosDBInMemory
+        Just path -> newLeiosDBSQLite nullTracer path
```
ch1bo, line 158: Much better already, the `leiosDbPath` is still an indirection that could be avoided:

```suggestion
      case leiosDbSource of
        NoLeiosDb -> newLeiosDBInMemory
        NodeLeiosDb -> newLeiosDBSQLite nullTracer (dbDir </> "leios.db")
```

If you want to keep that nicer explanation of the file not found, you should instead capture the `isDoesNotExistError` `IOException` and not check for existence. Why? It may appear just between you checking and then attempting to create the file in the `newLeiosDBSQLite` -> never ask for permission, but for forgiveness. Using a `catch` or `try` variant, or something like this lovely function [`forgivingAbsence`](https://hackage.haskell.org/package/path-io-1.8.2/docs/Path-IO.html#v:forgivingAbsence)

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBTruncater/Types.hs
```diff
@@ -3,12 +3,20 @@ module Cardano.Tools.DBTruncater.Types
...
+import Cardano.Tools.LeiosDb (LeiosDbSource)
 import Ouroboros.Consensus.Block.Abstract
 
 data DBTruncaterConfig = DBTruncaterConfig
   { dbDir :: FilePath
   , truncateAfter :: TruncateAfter
   , verbose :: Bool
+  , leiosDbSource :: LeiosDbSource
+  -- ^ 'NoLeiosDb' skips every LeiosDb operation. The tool then neither opens
+  -- nor modifies the @leios.db@ under 'dbDir', whether or not that file
```
ch1bo, line 15: The [node can configure the name of the database](https://github.com/intersectmbo/cardano-node/blob/6a540bd43a83e697fd662b973f7e99cee02a59fb/cardano-node/src/Cardano/Node/Configuration/Leios.hs#L20-L22), we want to have the truncator be able to do the same? Or just rename the file before running this? Or we assume nobody will actually ever rename the file?

ch1bo (approved):
Could be improved still, but is good enough still 🙂

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBAnalyser/Run.hs
dnadales: I applied the `case` patch, thanks.

I kept `doesFileExis` because removing it is harder than it seems:

 - newLeiosDBSQLite opens no file. It builds a handle whose open field opens a connection later. So nothing can fail at that case.
  - That later open passes SQLOpenReadWrite, SQLOpenCreate. A missing file is created with a schema, so there is no error to catch.
  - open2 throws SQLError, not IOException, so isDoesNotExistError never matches.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBTruncater/Types.hs
dnadales: Makes sense, fixed!

PR #2263, opened 2026-09-02
LeiosDemoLogic: upstream cardano-node tracer classification/rendering

nfrisby:
After this PR, we can change existing Leios tracer data constructors and add new tracer constructors without having to alter the `cardano-node` code at all.

If we were to instead add _entirely new_ tracer data _types_, then we'll need to add some instances in `cardano-node` and some functions in `ouroboros-consensus` for those functions to delegate to, like the functions this PR introduces.

(On `main`---not yet, but soon---the logging interface has already been pulled out to an isolated package, so that Consensus can define the instances etc without `cardano-node` needing to have any code at all. But the `leios-prototype` branches aren't in a state to use that new isolated package yet, as far as I know. So this PR is spiritually redundant. But it's worth it to simplify the ongoing `leios-prototype` development.)

### DIFF
--- a/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
+++ b/ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
@@ -96,6 +96,7 @@ import Data.Set.NonEmpty (NESet)
 import qualified Data.Set.NonEmpty as NESet
 import Data.String (fromString)
 import Data.Text (Text)
+import qualified Data.Text as T
 import Data.Time.Clock (NominalDiffTime)
 import Data.Vector.Strict (Vector)
 import qualified Data.Vector.Strict as V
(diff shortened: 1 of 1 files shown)


nfrisby:
https://github.com/IntersectMBO/cardano-node/tree/nfrisby/upstream-the-cardano-node-leios-tracer-rendering is the matching branch in `cardano-node`, excluding only the eventually-necessary `source-repository-stanza` stanza.

ch1bo (approved):

nfrisby:
After merging this, I created commit https://github.com/IntersectMBO/cardano-node/commit/5ad5c4ab79851a1fa9b966a6bd7f2468485a8109, successfully built `cabal build exe:cardano` locally, and then pushed it to the https://github.com/IntersectMBO/cardano-node/tree/leios-prototype branch of `cardano-node`.

PR #2264, opened 2026-09-03
tests: build against QuickCheck 2.18

dnadales:
On GHC 9.14 the solver picks QuickCheck 2.18, which deprecates `withMaxSuccess` and exports `chooseWord64`.
Both break the 9.14.1 CI job, because it compiles with `-Werror=deprecations` and `-Werror=name-shadowing`.

`Test.Util.QuickCheck` now gives a `withNumTests` that maps to `withMaxSuccess` before 2.18 and to the real function from 2.18 on, and the local `chooseWord64` helper is renamed.
This ports the test-only part of 624639730 from `main`, where the same change already landed with the LSM-tree update.


dnadales:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2274

PR #2265, opened 2026-09-03
Drop NodeToClient versions <v23

jasagredo:
The versions we are advertising support for are quite old and require tweaking legacy code:

```
┌──────┬────────────────────┬───────────────────┬─────────────┐
│ NTC  │ First node release │ consensus-cardano │ Shelley N2C │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_16 │ 9.0.0 (2024-07)    │ 0.18.0.0          │ 8           │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_17 │ 9.2.0 (2024-09)    │ 0.19.0.0          │ 9           │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_18 │ 10.1.1 (2024-10)   │ 0.20.0.0          │ 10          │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_19 │ 10.2.1 (2025-02)   │ 0.21.0.1          │ 11          │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_20 │ 10.3.1 (2025-04)   │ 0.23.0.0          │ 12          │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_21 │ 10.6.0 (2025-11)   │ 0.26.0.0          │ 13          │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_22 │ 10.6.0 (2025-11)   │ 0.26.0.0          │ 14          │
├──────┼────────────────────┼───────────────────┼─────────────┤
│ V_23 │ 10.7.0 (2026-03)   │ consensus 3.0.0.0 │ 15          │
└──────┴────────────────────┴───────────────────┴─────────────┘
```

The ecosystem seems to already support V23.

```
┌─────────────────┬──────────────┬────────────────────────────────────────────────────────┬──────────────────────────────┬──────┐
│     Client      │   Release    │                      What pins it                      │            Offers            │ Max  │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│ cardano-cli     │ 11.2.3.0     │ cardano-cli.cabal:249 → cardano-api ^>=11.6;           │ whole consensus set          │ V_23 │
│                 │ (2026-09-03) │ cardano-api.cabal:181 → ouroboros-consensus ^>=4.1     │                              │      │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│ Ogmios          │ v7.0.0       │ server/cabal.project:21 → any.ouroboros-consensus ^>=  │ [nodeToClientV_Min ..        │ V_23 │
│                 │ (2026-06-20) │ 3.0.1, ouroboros-network ^>= 1.1                       │ nodeToClientV_Latest]        │      │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│ cardano-wallet  │ v2026-08-21  │ cabal.project:212 → ouroboros-consensus ==3.0.1.0      │ whole consensus set (via     │ V_23 │
│                 │              │                                                        │ cardano-api)                 │      │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│                 │ 13.7.2.1     │ cabal.project index-state 2026-05-03 +                 │                              │      │
│ cardano-db-sync │ (2026-06-17) │ source-repository-package on cardano-node → consensus  │ whole consensus set          │ V_23 │
│                 │              │ 3.0.1.0                                                │                              │      │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│ Hydra           │ main         │ cabal.project:212 → ouroboros-consensus ==3.0.1.0      │ whole consensus set          │ V_23 │
├─────────────────┼──────────────┼────────────────────────────────────────────────────────┼──────────────────────────────┼──────┤
│ pallas (Rust)   │ main         │ hardcoded table,                                       │ VersionTable::v10_and_above  │ V_23 │
│                 │              │ pallas-network/src/miniprotocols/handshake/n2c.rs      │                              │      │
└─────────────────┴──────────────┴────────────────────────────────────────────────────────┴──────────────────────────────┴──────┘
```

So we are just deleting all versions except v23 so that cardano-node 11.2 only ships with that one.


ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -138,7 +136,6 @@ fromShelleyLedgerExamples
         [ ("LedgerTip", SomeResult GetLedgerTip (blockPoint blk))
         , ("EpochNo", SomeResult GetEpochNo (EpochNo 10))
         , ("EmptyPParams", SomeResult GetCurrentPParams lePParams)
```
dnadales, line 138: (This is about `Examples.hs:150`, a few lines below, which is outside the diff.)

Aren't we left with a golden example that the node can no longer produce? The example at `Examples.hs:150` is a `LedgerPeerSnapshotV2`, but after this PR `SingBigLedgerPeers` always answers `LedgerBigPeerSnapshotV23` (`Query.hs:441`). The golden bytes come from encoding the example value rather than from the answerer (`Golden.hs:391`), so the test stays green: `Result_GetBigLedgerPeerSnapshot` still starts `82 01`, internal version 1.

If so, `ShelleyNodeToClientVersion15` is now the only version we ship, and it has no golden for the shape clients actually receive. Before this PR the V2 bytes were a genuine v15 answer through the `GetLedgerPeerSnapshot' False` arm, so I think the gap is new rather than pre-existing.

What about something like:

```haskell
( "GetBigLedgerPeerSnapshot"
, SomeResult
    (GetLedgerPeerSnapshot SingBigLedgerPeers)
    (LedgerBigPeerSnapshotV23 point magic [...])
)
```

and regenerating the goldens.

While we are here, I wonder if the all-peers side deserves the same treatment. There is no `SingAllLedgerPeers` example anywhere, and `Generators.hs:159-190` leaves `GetLedgerPeerSnapshot` out of both the query and the result `Arbitrary` instances, so peer-kind `0` and `LedgerAllPeerSnapshotV23` have neither a golden nor a roundtrip today. That part is not something this PR introduced, but the `Bool` it removes was the last thing pushing two shapes through tag 34, so this seems like the moment to notice.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Golden.hs
```diff
@@ -49,12 +49,5 @@ instance
...
   toGoldenDirectory (queryVersion, blockVersion) =
     show queryVersion </> case blockVersion of
-      CardanoNodeToClientVersion12 -> "CardanoNodeToClientVersion12"
-      CardanoNodeToClientVersion13 -> "CardanoNodeToClientVersion13"
-      CardanoNodeToClientVersion14 -> "CardanoNodeToClientVersion14"
-      CardanoNodeToClientVersion15 -> "CardanoNodeToClientVersion15"
-      CardanoNodeToClientVersion16 -> "CardanoNodeToClientVersion16"
-      CardanoNodeToClientVersion17 -> "CardanoNodeToClientVersion17"
-      CardanoNodeToClientVersion18 -> "CardanoNodeToClientVersion18"
       CardanoNodeToClientVersion19 -> "CardanoNodeToClientVersion19"
```
dnadales, line 52: Should we do the same clean-up on the Shelley side? The 1113 golden files this PR deletes are all under `golden/cardano/`, and these seven directories are still tracked for constructors that no longer exist:

```
golden/shelley/QueryVersion2/ShelleyNodeToClientVersion{8,9,10,11}   78 files
golden/shelley/QueryVersion3/ShelleyNodeToClientVersion{12,13,14}    69 files
```

The whole `golden/shelley/QueryVersion2` subtree is unreachable now, since no surviving version maps to `QueryVersion2` for Shelley. Nothing will tell us: `Golden.hs:364-371` builds directory names from `supportedNodeToClientVersions` and never lists the filesystem, so extra directories are simply ignored. That leaves 147 of the 497 remaining golden files reading as coverage for versions we no longer have.

Something like:

```
git rm -r ouroboros-consensus-cardano/golden/shelley/QueryVersion2 \
          ouroboros-consensus-cardano/golden/shelley/QueryVersion3/ShelleyNodeToClientVersion1{2,3,4}
```

I wonder if `docs/website/contents/howtos/versioning_a_new_query.md` deserves a line about this too. It covers adding golden files (`:73-77`, `:100`) but says nothing about deleting the directories of a dropped version, which is what makes this easy to miss.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -560,37 +493,28 @@ instance
...
+    GetProposals{} -> const True
+    GetRatifyState{} -> const True
+    GetFuturePParams{} -> const True
+    GetLedgerPeerSnapshot{} -> const True
+    QueryStakePoolDefaultVote{} -> const True
+    GetPoolDistr2{} -> const True
+    GetStakeDistribution2{} -> const True
+    GetMaxMajorProtocolVersion -> const True
+    GetDRepDelegations{} -> const True
    where
```
dnadales, line 514: Is this `where` meant to stay? With the `v8` through `v15` bindings gone it opens an empty block, which compiles but reads as a leftover. The WARNING comment came out with them, so it is now detached from `blockQueryIsSupportedOnVersion` and duplicates the copy at `Query.hs:338-339`.

Should we drop both the `where` and the second copy?

docs/website/contents/explanations/queries.md
```diff
@@ -182,19 +165,15 @@ and the serialization of the arguments.
...
-
-*: The queries enabled only before version 13 used old types from the Ledger
-that were removed in recent versions. The queries enabled at version 13 use the
-new corresponding ledger types.
+| 31  | `GetProposals`                            | †                                   | `Set GovActionId`                                                                             | `Seq (GovActionState era)`                                   |
+| 32  | `GetRatifyState`                          | †                                   |                                                                                               | `RatifyState era`                                            |
+| 33  | `GetFuturePParams`                        |                                     |                                                                                               | `Maybe (PParams era)`                                        |
+| 34  | `GetBigLedgerPeerSnapshot`                |                                     |                                                                                               | `LedgerPeerSnapshot` ‡                                       |
```
dnadales, line 171: Should this row list the peer-kind byte? The Arguments cell is empty, which reads as "send only the tag", but the encoder sends a byte after it: `0` for all peers, `1` for big peers (`Query.hs:796-802`). A client author following the table sends too few bytes.

The `‡` on this row and on row 11 also looks stale. The footnote says the layout varies with the negotiated version, and only `ShelleyNodeToClientVersion15` is left.

changelog.d/20260903_130000_javier.sagredo_drop_ntc_10x.md
```diff
@@ -0,0 +1,33 @@
...
+- `GetLedgerPeerSnapshot'` and its `GetLedgerPeerSnapshot` pattern synonym were
+  merged back into a plain `GetLedgerPeerSnapshot` constructor without the
+  version-discriminating `Bool`. The pre-`ShelleyNodeToClientVersion15`
+  encoding of tag 34 (a 1-element list) is no longer decoded.
+
+- Removed `Ouroboros.Consensus.Shelley.Ledger.Query.Types` and
+  `Ouroboros.Consensus.Shelley.Ledger.Query.LegacyPParams`, which only existed
+  to serve the removed queries and the pre-`ShelleyNodeToClientVersion13`
+  `PParams` encoding. `ShelleyCompatible` no longer carries the
+  `ToCBOR`/`FromCBOR (LegacyPParams era)` constraints.
```
dnadales, line 33: Should we add two more breaking entries here? `LegacyShelleyGenesis`, `encodeLegacyShelleyGenesis` and `decodeLegacyShelleyGenesis` also went away, and their module survives, so it is an export break rather than a module removal. And `exhaustiveSupportedNetworkProtocolVersions` became `contiguousSupportedNetworkProtocolVersions` in `unstable-consensus-testlib`, which we have recorded for testlib changes before.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -1088,41 +974,24 @@ decodeShelleyResult v query = case query of
...
-      (toCBOR, fromCBOR)
-  | otherwise =
-      (encodeLegacyPParams, decodeLegacyPParams)
-
 genesisConfigEnDecoding ::
   forall s.
   ShelleyNodeToClientVersion ->
   (CompactGenesis -> Encoding, Decoder s CompactGenesis)
 genesisConfigEnDecoding v
   | v > ShelleyNodeToClientVersion15 =
```
dnadales, line 989: Is this guard a deliberate hook for v16? Nothing satisfies it today, since `ShelleyNodeToClientVersion15` is both the only constructor and `maxBound`. It was already unreachable before this PR, so the bytes do not change.

Whoever adds `ShelleyNodeToClientVersion16` will flip `GetGenesisConfig` to the 16-field encoding without touching this line. Should we say so in a comment, or drop the branch until the version that needs it?

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -445,16 +315,24 @@ instance
       , (NodeToNodeV_16, CardanoNodeToNodeVersion2)
       ]
 
+  -- \| The @cardano-node@ release that first shipped each version, which is what
```
dnadales, line 318: Should this table also live in `queries.md`? It answers exactly the question `versioning_a_new_query.md:87` sends the reader off to look up by hand. Here the `-- \|` keeps it out of Haddock, and Haddock does not document instance methods anyway, so only people reading the source will find it.

I would keep this copy next to `supportedNodeToClientVersions`, since that is where the drop decision gets made.

ouroboros-consensus-cardano/test/cardano-test/Test/Consensus/Cardano/Golden.hs
jasagredo: Fixed in third commit (SHAs will change...)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
jasagredo: I am now pattern matching on ShelleyNodeToClientVersion15 and will warn when a new version comes up.

KtorZ:
I double-checked what Ogmios' latest (v7.0.0) uses:

```
ouroboros-network == 1.1.0.0
ouroboros-consensus == 3.0.1.0
```

So, I assume this is fine 👍

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -981,15 +874,15 @@ decodeShelleyQuery = do
     (2, 31) -> requireCG $ SomeBlockQuery . GetProposals <$> LC.fromEraCBOR @era
     (1, 32) -> requireCG $ return $ SomeBlockQuery GetRatifyState
     (1, 33) -> requireCG $ return $ SomeBlockQuery GetFuturePParams
-    -- TODO: remove (1, 34) once NodeToClientV_22 is no longer supported
-    (1, 34) -> return . SomeBlockQuery $ GetLedgerPeerSnapshot' False SingBigLedgerPeers
+    (1, 34) ->
+      failmsg
+        "GetLedgerPeerSnapshot takes a peer kind: send [34, 0] for all peers or [34, 1] for big peers,"
```
dnadales, line 877: Should we say in the changelog that this narrows `NodeToClientV_23` itself?

I think it does, and the reason is easy to miss: the supported-version check runs **only in the encoder**.

```haskell
-- Node/Serialisation.hs:369-373
encodeNodeToClient _ version (SomeBlockQuery q)
  | blockQueryIsSupportedOnVersion q version = encodeShelleyQuery q
  | otherwise = throw $ ShelleyEncoderUnsupportedQuery ...
decodeNodeToClient _ _ = decodeShelleyQuery   -- version dropped
```

`LocalStateQuery/Server.hs:25-36` never takes a version either.
So the node answered any query it could parse, at any negotiated version.
That makes the four wire shapes we delete reachable at `V_23` today:

| Wire | Before | After |
|---|---|---|
| `[4]` | empty `ProposedPPUpdates` | decode failure |
| `[5]`, `[21, x]` | real data, via `GetStakeDistribution2` and `GetPoolDistr2` | decode failure |
| `[34]` | `LedgerPeerSnapshotV2` | decode failure |

The first three only reached the node from a client with its own codec, since our encoder refused them at `v15`.

`[34]` is the one I would call out, because our encoder allowed it:

```haskell
GetLedgerPeerSnapshot' False _peerKind -> (>= v11)   -- v15 passes
```

Anything built on our client library could send `[34]` on a `V_23` connection and get an answer.
So this is not a pre-`v15` path being cleaned up.

One thing I could not settle from the consensus side: does this `fail` drop only the LocalStateQuery mini-protocol, or the whole connection?
If it is the connection, the effect on an affected client is worth spelling out.

What about something like this in `20260903_130000_javier.sagredo_drop_ntc_10x.md`:

> A `NodeToClientV_23` peer that sent CBOR tag 4, 5 or 21, or the one-element form of tag 34, previously received an answer, because the supported-version check runs only in the client-side encoder. The decoder now rejects all four.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/SupportedNetworkProtocolVersion.hs
```diff
@@ -11,30 +11,67 @@ import Ouroboros.Consensus.Node.NetworkProtocolVersion
...
+    unsupportedNewerVersions
+      | Set.null mappedVersions = Set.empty
+      | otherwise =
+          Set.fromList [Set.findMax mappedVersions .. maxBound]
+            Set.\\ mappedVersions
+    holes
+      | Set.null mappedVersions = Set.empty
+      | otherwise =
+          Set.fromList [Set.findMin mappedVersions .. Set.findMax mappedVersions]
+            Set.\\ mappedVersions
```
dnadales, line 86: Should we pin the oldest supported version too, not just the newest?

`holes` takes both of its bounds from the map itself, so nothing here constrains where the run starts.
`minBound` no longer appears in the file at all.
The old test got that constraint from `allVersions = Set.fromList [minBound .. maxBound]`.

Today it makes no difference, because `V_23` is both `findMin` and `maxBound`.
It starts to matter once the network layer adds `V_24`:

| map | new test | old test |
|---|---|---|
| `{V_23, V_24}` | passes | passes |
| `{V_24}` | passes | fails: `unmapped versions: [V_23]` |

The second row is one version too many dropped, and every client that offers at most `V_23` is then refused at the handshake.

I also wonder whether `unsupportedNewerVersions` earns its place.
`nodeToClientVersionToQueryVersion` (`Ledger/Query/Version.hs:18-20`) is already a total `case` over the enum, so a new constructor is a `-Werror` build failure before the test runs.
That leaves `holes` as the only assertion the compiler does not already make.

The map we read here is also not the one a node advertises.
`llrnNodeToClientVersions` goes through `limitToLatestReleasedVersion` (`ouroboros-consensus-diffusion/.../Consensus/Node.hs:1159-1164`), which is `Map.takeWhileAntitone (<= V_23)` given `Cardano/Node.hs:338`.
On the `{V_24}` row that yields the empty map, so a production node offers no node-to-client version at all.

What about one more assertion, using something already in the class?

```haskell
assertBool
  ("the latest released version is not supported by " <> blkName)
  (maybe True (`Set.member` mappedVersions) (prj (latestReleasedNodeVersion p)))
```

It keeps the freedom to drop old versions, which is the point of the rewrite, and it still fails on the second row.
It needs `testVersions` to take the matching `prj` (`fst` or `snd`) next to `f`.

cabal.project
```diff
@@ -45,6 +45,20 @@ source-repository-package
...
+-- NodeToClientV_23 is the only NodeToClient version.
+source-repository-package
+  type: git
+  location: https://github.com/IntersectMBO/ouroboros-network.git
+  tag: 09495ae9e6a063929ac9f2a8a80d210735c3a66e
+  --sha256: sha256-/m3CUFKSnTw8YpyIMhcpdVbHWu6FEqjiQFiPORWLFFI=
+  subdir:
+    ouroboros-network
+    cardano-diffusion
+    -- network-mux 0.11.0.0 at this revision carries the JobPool API that
```
dnadales: Should this comment name the breaking change rather than the module?

CHaP's published `network-mux-0.11.0.0` already exposes `Control.Concurrent.JobPool`, so "carries the JobPool API" reads as if the module were new here.
A reader who checks CHaP will conclude the `network-mux` subdir is redundant and drop it.

The pin does look necessary, but for a different reason.
At `09495ae9` the `JobPool` type gains a `HasQueue` parameter and the module adds `withJobPool_`, and the network changelog fragment marks it `### Breaking`.
All of that ships under an unchanged `0.11.0.0`.

Something like:

```
    -- network-mux at this revision changes the JobPool API (JobPool gains a
    -- HasQueue parameter) without a version bump, so the released
    -- network-mux-0.11.0.0 does not build against
    -- ouroboros-network:framework at this revision.
```

The same holds for the two subdirs above, and I think that is worth a line in the `Temp:` comment.
`ouroboros-network` and `network-mux` are both published to CHaP at exactly the versions this tree carries:

| package | this revision | CHaP | published |
|---|---|---|---|
| `ouroboros-network` | 1.2.0.0 | 1.2.0.0 | 2026-07-29 |
| `network-mux` | 0.11.0.0 | 0.11.0.0 | 2026-07-29 |
| `cardano-diffusion` | 1.1.1.0 | 1.1.1.0 | 2026-08-06 |

Both of the first two fall inside our CHaP `index-state` of 2026-07-30.
So the pin shadows released packages with different code, and `ouroboros-network ^>=1.2` at `ouroboros-consensus.cabal:389` cannot tell them apart.
`Query.hs:932` now calls the one-argument `encodeLedgerPeerSnapshot`, which the released `1.2.0.0` does not have, so we cannot build against it.

I guess the practical point is that upstream needs version bumps before this pin comes out, and nothing in our CI says so: `check-build-plans` is a `--dry-run`, so it resolves happily against the stale-code `1.2.0.0` and would only fail at compile time.
Do we want that recorded here, or does it belong in the release checklist?

docs/website/contents/explanations/queries.md
```diff
@@ -155,14 +163,12 @@ and the serialization of the arguments.
...
 | 9   | `GetCBOR`                                 | the version of the internal query   | `BlockQuery (ShelleyBlock proto era) fp result`                                               | `BlockQuery (ShelleyBlock proto era) fp (Serialised result)` |
 | 10  | `GetFilteredDelegationsAndRewardAccounts` |                                     | `Set (Credential Staking)`                                                                    | `(Delegations, Map (Credential Staking) Coin)`               |
-| 11  | `GetGenesisConfig`                        |                                     |                                                                                               | `CompactGenesis` ‡                                           |
+| 11  | `GetGenesisConfig`                        |                                     |                                                                                               | `CompactGenesis`                                             |
```
dnadales: Should this row still flag that the result is not `CompactGenesis`'s own encoding?

I agree the `‡` footnote had to go, since its wording ("depend on the particular `NodeToClientVersion` negotiated") is no longer true with one `ShelleyNodeToClientVersion` left.
Row 34 losing the marker looks right too, because dropping the SRV argument is byte-neutral for the V23 constructors.

Row 11 is the one case where the fact outlived the footnote.
`Query.hs:910` and `:956` still go through `genesisConfigEnDecoding`, and `:998-1001` uses `encodeShelleyGenesisNoExtraConfig`, which writes `encodeListLen 15` and decodes `decodeRecordNamed "ShelleyGenesis" (const 15)`.
`CompactGenesis`'s own instance is derived newtype-wise over `SL.ShelleyGenesis` (`Shelley/Ledger/Config.hs:133-135`), and that record now has 16 fields with `sgExtraConfig`.

So a client written from this row encodes 16 fields and gets a bare "expected 15 fields" back, with nothing here to explain why.
This is the only result codec left that is not the plain instance, and it is now the one thing the table does not mention.

What about putting it in the Result cell rather than in a footnote, so it cannot drift again?
Something like:

```
| 11  | `GetGenesisConfig` | | | `ShelleyGenesis`, 15 fields, without `sgExtraConfig` |
```

While we are here: `queries.md:149-150` still says "Sometimes the encoding of the result changes depending on the particular `NodeToClientVersion` negotiated", which is stale for the same reason the footnote was.

docs/website/contents/howtos/versioning_a_new_query.md
```diff
@@ -70,12 +70,23 @@ The supported query versions are only enforced in the [Shelley query encoder][sh
 
 As an example, consider a query $Q$ that is enabled after version $x$, and consider a connection between a client and a node that negotiated version $y$. If $y < x$, then the client will throw an exception before sending $Q$ as the negotiated version is too old, so the server probably won't understand the query. But if the server does actually understand the query, and the client uses a custom implementation that does not perform the check on $x$ and $y$, then the server will reply as normal despite $y < x$.
 
-### On newly added golden files
+### On golden files
```
dnadales, line 73: (This is about the `### Checks` section just above, at `versioning_a_new_query.md:69-71`, which is outside the diff.)

Should we qualify this section now that some queries fail in the decoder?

The paragraph says the server "will currently answer queries even if the negotiated version is not in the supported version range for that query", and `:71` works the example through.
That is still true for every query left in the `case (len, tag)` list, so I would not delete it.

It stops being true for the four shapes this PR removes.
`[4]`, `[5]`, `[21, x]` and the one-element `[34]` now reach `failmsg` in `decodeShelleyQuery` instead of an answer, so "be liberal in what you accept" no longer describes them.
I left a longer note on `Query.hs:879` about that.

What about a sentence at the end of `:69`, something like:

> This holds only for queries the decoder still accepts. Removing a query removes its decoder case, so a client that sends a removed query gets a decode failure rather than an answer.

Reading it together with the new step 7 below, I wonder if this is really the same lesson: dropping a version is not only a golden-file chore, it also narrows what the decoder accepts, and neither is visible from a compiler warning.

docs/website/contents/explanations/queries.md
```diff
@@ -182,25 +187,16 @@ and the serialization of the arguments.
...
-| 37  | `GetStakeDistribution2`                   | >= v13 *                            |                                                                                               | `PoolDistr`                                                  |
-| 38  | `GetMaxMajorProtocolVersion`              | >= v13                              |                                                                                               | `MaxMajorProtVer`                                            |
-| 39  | `GetDRepDelegations`                      | >= v14                              | `Set DRep`                                                                                    | `(Map DRep (Set (Credential Staking)))`                      |
-
-*: The queries enabled only before version 13 used old types from the Ledger
-that were removed in recent versions. The queries enabled at version 13 use the
-new corresponding ledger types.
+| 31  | `GetProposals`                            | †                                   | `Set GovActionId`                                                                             | `Seq (GovActionState era)`                                   |
```
dnadales: (This is about the table as a whole, not row 35 specifically.)

Should we leave a note that tags 4, 5 and 21 are now spent?

Deleting the rows leaves the Tag column reading `0 1 2 3 6 ...`, with holes and nothing saying why.
Before this PR the space was `0..39` with no gaps, so these are the first ones.

The reason I raise it: tags are hand-written literals, not derived from constructor order.

```haskell
GetLedgerTip -> CBOR.encodeListLen 1 <> CBOR.encodeWord8 0
GetEpochNo   -> CBOR.encodeListLen 1 <> CBOR.encodeWord8 1
```

So "add new queries to the end of the list" (`Query.hs:333-335`) keeps the *order* tidy but does not stop someone picking `4` for the next query.
Nothing mechanical would catch it either.
There is no CDDL spec for node-to-client queries, `goldenTest_SerialiseNodeToClient` passes `Nothing` for the CDDL slot (`Golden.hs:401`), and no test asserts a tag-to-constructor mapping.

We have relied on the tag staying free once already.
`5178b87e1` "Bring back the GetProposedPParamsUpdates query as deprecated" restored that query at tag 4 after an earlier commit removed it, which worked only because nothing had claimed 4 in between.

What about three rows kept in place instead of deleted?

```
| 4   | *(removed, do not reuse)* | | | |
| 5   | *(removed, do not reuse)* | | | |
| 21  | *(removed, do not reuse)* | | | |
```

The removal recipe at `Query.hs:340-345` also stops at "it can be removed".
I wonder if it wants a last step, something like "record the tag as spent; never reassign it".

Small thing while we are in that comment: it names `querySupportedVersion`, which does not exist.
The function is `blockQueryIsSupportedOnVersion`, and with one `ShelleyNodeToClientVersion` there is no `X` left to be `< X`, so the recipe as written is not followable today.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -560,37 +492,27 @@ instance
...
+    GetDRepDelegations{} -> const True
 
-    v8 = ShelleyNodeToClientVersion8
-    v9 = ShelleyNodeToClientVersion9
-    v10 = ShelleyNodeToClientVersion10
-    v11 = ShelleyNodeToClientVersion11
-    v12 = ShelleyNodeToClientVersion12
-    v13 = ShelleyNodeToClientVersion13
-    v15 = ShelleyNodeToClientVersion15
+-- WARNING: when adding a new query, a new @ShelleyNodeToClientVersionX@
```
dnadales: Is this copy still doing anything here?

It reads the same as the WARNING at `Query.hs:337-338`, in the block above the `BlockQuery` GADT.
It also sits between the end of `blockQueryIsSupportedOnVersion` and `instance SameDepIndex2`, so Haddock and a human both read it as a note about `SameDepIndex2`.

I think it came out of the `where` block that held `v8` through `v15`, and lost its anchor when those bindings went.
Should we just drop it and leave the copy at `:337-338`?

docs/website/contents/explanations/queries.md
```diff
@@ -78,15 +71,30 @@ instead we focus on the Cardano version. All the current versions imply also
...
-| `NodeToClientV_16`    | `CardanoNodeToClientVersion12` | `ShelleyNodeToClientVersion8`  |
-| `NodeToClientV_17`    | `CardanoNodeToClientVersion13` | `ShelleyNodeToClientVersion9`  |
-| `NodeToClientV_18`    | `CardanoNodeToClientVersion14` | `ShelleyNodeToClientVersion10` |
-| `NodeToClientV_19`    | `CardanoNodeToClientVersion15` | `ShelleyNodeToClientVersion11` |
-| `NodeToClientV_20`    | `CardanoNodeToClientVersion16` | `ShelleyNodeToClientVersion12` |
-| `NodeToClientV_21`    | `CardanoNodeToClientVersion17` | `ShelleyNodeToClientVersion13` |
-| `NodeToClientV_22`    | `CardanoNodeToClientVersion18` | `ShelleyNodeToClientVersion14` |
 | `NodeToClientV_23`    | `CardanoNodeToClientVersion19` | `ShelleyNodeToClientVersion15` |
 
+What determines when a version may be dropped is the `cardano-node` release
```
dnadales, line 76: (Two things about the table above and the marker column below, both outside the diff.)

`:65` and the column header at `:72` say `BlockNodeToNodeVersion blk`, but the values in the table are node-to-client ones.
`CardanoNodeToClientVersion19` is a `BlockNodeToClientVersion (CardanoBlock c)` (`Cardano/Node.hs:292`), and `BlockNodeToNodeVersion (CardanoBlock c)` is a different type whose constructors are `CardanoNodeToNodeVersion1`/`2`.
So someone who follows that name into the Haddocks finds the wrong family.
Should it be `BlockNodeToClientVersion blk` in both places?

Separately, the `†` markers do not line up with `requireCG`.
I checked every row against every call site, and three disagree:

| Tag | Query | `requireCG` | `†` |
|---|---|---|---|
| 29 | `GetAccountState` | no (`Query.hs:872`) | yes (`:188`) |
| 33 | `GetFuturePParams` | yes (`Query.hs:876`) | no (`:192`) |
| 39 | `GetDRepDelegations` | yes (`Query.hs:890`) | no (`:198`) |

All three predate this PR, so no rush.
I raise them because the PR rewrites those exact cells, so it is a cheap moment to fix them.

Row 33 is the interesting one.
`GetFuturePParams` carries no `ConwayEraGov era` constraint (`Query.hs:298`) and `SL.queryFuturePParams` is era-generic, so its `requireCG` looks like the thing that is wrong, not the missing `†`.
Do you know if that gate is deliberate?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
jasagredo: I investigated and it does restrict NodeToClientV_23, but any client with V23 should know that the constructor that created the old encoding was `{-# DEPRECATED #-}` so they can use the new constructor and have no reason to use the old one. If something was deprecated in a version I am not sure it counts as a very serious breaking change.

I edited the comments around this a bit.

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/SupportedNetworkProtocolVersion.hs
jasagredo: Nice suggestion, implemented

cabal.project
jasagredo: Arg this was just an auto comment. Nice catch. I have removed it.

docs/website/contents/explanations/queries.md
jasagredo: Indeed, good catch

docs/website/contents/howtos/versioning_a_new_query.md
jasagredo: Nice! thanks!

docs/website/contents/explanations/queries.md
jasagredo: Done, yeah!

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
jasagredo: Indeed, good catch again!

docs/website/contents/explanations/queries.md
jasagredo: Fixed, indeed. Good catch.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -981,15 +872,15 @@ decodeShelleyQuery = do
     (2, 31) -> requireCG $ SomeBlockQuery . GetProposals <$> LC.fromEraCBOR @era
     (1, 32) -> requireCG $ return $ SomeBlockQuery GetRatifyState
     (1, 33) -> requireCG $ return $ SomeBlockQuery GetFuturePParams
-    -- TODO: remove (1, 34) once NodeToClientV_22 is no longer supported
-    (1, 34) -> return . SomeBlockQuery $ GetLedgerPeerSnapshot' False SingBigLedgerPeers
+    (1, 34) ->
```
dnadales, line 875: Should we say in the changelog that this rejection costs the client its whole connection, not just the query?

I traced what `failmsg` does to a peer that sends the one-element form. It is a plain CBOR `fail` (`Query.hs:819-828`), and `LocalStateQuery` has no per-query error message, only `MsgFailure AcquireFailure` on acquire. So the failure leaves the decoder as an exception rather than a reply:

```
failmsg          -- Query.hs:875
  -> fail        -- Query.hs:819-828
  -> throwIO (DecoderFailure ...)    -- network Driver/Stateful.hs:110-128
  -> MiniProtocolException           -- network-mux Mux.hs:357-363
  -> muxStatus = Failed e; throwIO   -- Mux.hs:458-463
  => bearer torn down
```

The node itself is fine, since the rethrow policy gives `ShutdownPeer` rather than `ShutdownNode`. But the bearer carries chain-sync, tx-submission and tx-monitor too, so they go with it, and the client sees a closed socket with no explanation. The text we wrote here only ever reaches our own trace.

I think that is acceptable, given the paragraph's argument that no shipped client emits this form. It seems worth one sentence next to that argument, though, so the person reading the changelog after an incident knows a dropped connection is the expected symptom rather than a second bug. Something like:

"A peer that sends one of these anyway gets a decode failure, which terminates the whole node-to-client connection rather than that one query, as the mini-protocol has no per-query error message."

changelog.d/20260903_130000_javier.sagredo_drop_ntc_10x.md
```diff
@@ -0,0 +1,52 @@
...
+  or the one-element form of tag 34, previously got an answer, because the
+  supported-version check runs only in the client-side encoder and the decoder
+  never sees the negotiated version. The decoder now rejects all four.
+
+  No shipped client emits any of them. Our encoder already refused tags 4, 5 and
+  21 at `ShelleyNodeToClientVersion15`, so only a client with its own codec could
+  send those. The one-element form of tag 34 came from the deprecated
+  `GetLedgerPeerSnapshot'` applied to `False`, whereas the `GetLedgerPeerSnapshot`
+  pattern synonym that `cardano-api` and `dmq-node` use encodes the two-element
+  form. A client too old to know that form offers at most `NodeToClientV_22` and
```
dnadales: Is this last sentence correct? I do not think it is, and the sentence above it already gives the reason.

In the base code the arity came from the `Bool`, not from the negotiated version:

```haskell
GetLedgerPeerSnapshot' False _peerKind -> (>= v11)   -- Query.hs:576
GetLedgerPeerSnapshot' True  _peerKind -> (>= v15)   -- Query.hs:577
```

The encoder chose the arity from that same `Bool` (`Query.hs:896-899`). So a client built against `3.0.1.0` can negotiate `NodeToClientV_23`, call the deprecated constructor with `False`, and send `[34]`. Our encoder allows it, because `v15 >= v11`. That client is not too old, and it passes the handshake.

The real reason is the one we give above: `cardano-api` and `dmq-node` both use the pattern synonym, which is `GetLedgerPeerSnapshot' True`. What about dropping the last sentence and keeping only that? Something like:

"The one-element form of tag 34 came from the deprecated `GetLedgerPeerSnapshot'` applied to `False`. Every in-org client uses the `GetLedgerPeerSnapshot` pattern synonym instead, which is that constructor applied to `True`, so it sends the two-element form."

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/SupportedNetworkProtocolVersion.hs
```diff
@@ -11,30 +11,76 @@ import Ouroboros.Consensus.Node.NetworkProtocolVersion
...
+--
+-- What must hold is that the supported versions are a suffix without holes:
+--
+--  * ending at 'maxBound', so that adding a version to the network layer's enum
+--    and forgetting to map it here is a build-time-visible failure, which is
+--    the mistake this test was originally written to catch;
+--
+--  * without holes, since the handshake picks the highest version both sides
+--    know, and a gap in the middle only makes the negotiated version harder to
+--    reason about for no benefit;
```
dnadales, line 30: Should this bullet say "contain" rather than "pins where the run starts"? The two sentences seem to disagree with each other, and I read the first one as the wrong way round.

The assertion below is a membership check, so with the other three it comes to `findMin <= released`. That caps how far we can raise the oldest supported version. It says nothing about where the run starts.

Cardano is the one instance where the check has any effect, and there the released version sits in the middle of the run:

```haskell
supportedNodeToNodeVersions _ = Map.fromList
  [ (NodeToNodeV_14, ...), (NodeToNodeV_15, ...), (NodeToNodeV_16, ...) ]
latestReleasedNodeVersion _prx = (Just NodeToNodeV_15, Just NodeToClientV_23)
```

That is deliberate: `Node.hs:1163` applies `Map.takeWhileAntitone (<= version)`, so `V_15` is the upper cut that hides in-development `V_16`, and `V_14` is still offered. If we read the bullet as written and dropped `V_14`, the test would still pass and we would lose real clients.

What about something like:

```haskell
--  * containing 'latestReleasedNodeVersion', which caps how far the oldest
--    supported version can rise. The diffusion layer truncates the advertised
--    map at the released version, so a released version below the run leaves
--    nothing to advertise.
```

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -152,18 +151,31 @@ fromShelleyLedgerExamples
...
-                  ( NotOrigin slotNo
-                  ,
-                    [
-                      ( AccPoolStake 0.9
-                      ,
-                        ( PoolStake 0.9
-                        , LedgerRelayAccessAddress (IPv4 "1.1.1.1") 1234 :| []
-                        )
+              ( LedgerBigPeerSnapshotV23
+                  (BlockPoint slotNo (RawBlockHash "<BLOCK HASH>"))
```
dnadales: Should these examples use a 32-byte hash? The answerer always sends one:

```haskell
blockPointHash =
  RawBlockHash $ toShortRawHash (Proxy @(ShelleyBlock proto era)) (blockPointHash blk)
```

`"<BLOCK HASH>"` is 12 bytes, so the golden gets a `4c` prefix where a real answer gets `58 20`. The codec is length-prefixed, so nothing breaks. But the golden then shows a shape no client ever reads, and I thought the point of adding these was to pin what clients do read.

Also, `Query.hs:429-438` maps over `getPoint`, so a node at origin answers with `GenesisPoint`. Neither of the two examples covers that, and today it lives in the roundtrip generator alone. What about a third entry for it? Something like:

```haskell
( "GetBigLedgerPeerSnapshotAtOrigin"
, SomeResult
    (GetLedgerPeerSnapshot SingBigLedgerPeers)
    (LedgerBigPeerSnapshotV23 GenesisPoint (NetworkMagic 42) [])
)
```

docs/website/contents/explanations/queries.md
```diff
@@ -58,35 +58,43 @@ and how to interpret the results.
...
+| `NodeToClientV_19`    | 10.2.1                          |
+| `NodeToClientV_20`    | 10.3.1                          |
+| `NodeToClientV_21`    | 10.6.0                          |
+| `NodeToClientV_22`    | 10.6.0                          |
+| `NodeToClientV_23`    | 10.7.0                          |
+
+Note that 11.x introduced no new version: 10.7.1 and 11.0.1 both use
+`ouroboros-consensus-3.0.1.0` and so offer the very same set.
+
+This table is mirrored by the comment on
```
dnadales: Will a reader who follows this link find the table? I do not think so, for two reasons.

The comment in `Cardano/Node.hs:319` starts with an escaped pipe:

```haskell
  -- \| The @cardano-node@ release that first shipped each version, ...
```

so GHC reads it as an ordinary comment and Haddock never emits it. Fourmolu adds that escape, because the comment sits on a method binding inside an instance body, where Haddock cannot attach. The Haddock on the class method itself says only "Enumerate all supported node-to-client versions". The table shows up only in the hyperlinked source page, one click further on, and nothing tells the reader to go there.

Separately, that URL is a 404 right now. Everything under `haddocks/ouroboros-consensus-cardano/` gives 404, while `haddocks/ouroboros-consensus/...` works. `:41` and `:46` have the same dead prefix, so that part is not new here.

What about pointing at the source instead? Something like:

```markdown
This table is mirrored by the comment on `supportedNodeToClientVersions` for
`CardanoBlock`, in
[`Ouroboros/Consensus/Cardano/Node.hs`](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/Node.hs);
keep the two in sync when a new version is added.
```

docs/website/contents/explanations/queries.md
```diff
@@ -172,35 +180,29 @@ and the serialization of the arguments.
...
+| 36  | `GetPoolDistr2`                           |                                     | `Maybe (Set (KeyHash StakePool))`                                                             | `PoolDistr`                                                  |
+| 37  | `GetStakeDistribution2`                   |                                     |                                                                                               | `PoolDistr`                                                  |
+| 38  | `GetMaxMajorProtocolVersion`              |                                     |                                                                                               | `MaxMajorProtVer`                                            |
+| 39  | `GetDRepDelegations`                      | †                                   | `Set DRep`                                                                                    | `(Map DRep (Set (Credential Staking)))`                      |
+
+A removed query's tag stays spent: it is never reassigned, because deployed
+clients still encode the removed query with it.
```
dnadales, line 205: (This is about the column header at `:160` and the footnote at `:206`, both outside the diff.)

Should this column change its name now? It is headed "Only active on Shelley NTC versions", but after this PR no cell in it names a version. Every constraint that used to sit there (`< v12`, `< v13`, `>= v8` up to `>= v15`) is gone. What is left is the dagger, which is an era condition, and `GetCBOR`'s cell at `:171`.

The dagger tracks `requireCG` in the decoder rather than a version:

```haskell
requireCG k = case SE.getConwayEraGovDict (Proxy @era) of
  Just SE.ConwayEraGovDict -> k
  Nothing -> failmsg "that query is not supported before Conway,"
```

So the footnote at `:206` now opens with "Even if an appropriate version is enabled", and there is only one version, which enables everything. What about renaming the column and trimming that clause? Something like:

```markdown
| Tag | Query | Era restriction | Arguments | Result |
```

and

```markdown
†: these queries can only be answered when the corresponding era is Conway or
later, as they relate to governance concepts only present starting on Conway.
```

`GetCBOR`'s cell is a separate small thing: it says "the version of the internal query", and every internal query is now `const True` in `blockQueryIsSupportedOnVersion`.

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Generators.hs
```diff
@@ -187,8 +208,69 @@ instance (Arbitrary (InstantStake era), CanMock proto era) => Arbitrary (SomeRes
...
+  oneof
+    [ LedgerRelayAccessAddress <$> genIP <*> genPort
+    , LedgerRelayAccessDomain <$> genDomain <*> genPort
+    , LedgerRelayAccessSRVDomain <$> genDomain
+    ]
+ where
+  genIP = elements [IPv4 "1.1.1.1", IPv4 "3.3.3.3", IPv6 "2001:db8::1"]
+  genDomain = elements ["test1.", "test2.", "test3."]
+  genPort = fromIntegral <$> (arbitrary :: Gen Word16)
+
```
dnadales, line 273: (This is about `:386-388` and `:396`, both outside the diff.)

Should this comment and the `suchThat` come out with the versions they name? The comment refers to `ShelleyNodeToClientVersion1` and `2`, which this PR deletes:

```haskell
-- | Some 'Query's are only supported by 'ShelleyNodeToClientVersion2', so we
-- make sure to not generate those queries in combination with
-- 'ShelleyNodeToClientVersion1'.
```

The filter below it cannot reject anything now, for two reasons. Every arm of `blockQueryIsSupportedOnVersion` is `const True` after this PR, and `arbitrary` is `arbitraryBoundedEnum` over a type with one constructor. So `version` is always `ShelleyNodeToClientVersion15` and the predicate is always `True`.

It costs nothing to leave, so this is only about the next reader. What about something like:

```haskell
instance
  CanMock proto era =>
  Arbitrary
    (WithVersion ShelleyNodeToClientVersion (SomeBlockQuery (BlockQuery (ShelleyBlock proto era))))
  where
  arbitrary = WithVersion <$> arbitrary <*> arbitrary
```

and bringing the filter back when a second version arrives.

cabal.project
```diff
@@ -45,6 +45,18 @@ source-repository-package
     eras/conway/impl
     eras/dijkstra/impl
 
+-- Temp: pin ouroboros-network to PR IntersectMBO/ouroboros-network#5426, where
```
dnadales: Should this comment name what the pin is waiting for? `#5426` merged on 2026-09-07, and `09495ae` is its merge commit, which is network `main` right now. So a reader who checks the PR finds it closed and learns nothing about when the pin can go.

What about something like:

```
-- Temp: pin ouroboros-network to main, where NodeToClientV_23 is the only
-- NodeToClient version. Drop once these packages are released to CHaP.
```

One thing for that un-pin commit rather than this PR: `cardano-diffusion` is the package that exports `NodeToClientVersion`, and we carry no bound on it at all. Upstream removed the constructors without moving the version number, so once CHaP publishes a bumped version we need a bound there, or the solver can pick a pre-drop tree and `Ledger/Query/Version.hs` becomes a non-exhaustive `case`.

dnadales (commented):
It seems it's converging :D

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Query.hs
jasagredo: Yes, mentioned in the changelog now

changelog.d/20260903_130000_javier.sagredo_drop_ntc_10x.md
jasagredo: Fixed the wording. thanks

ouroboros-consensus/src/unstable-consensus-testlib/Test/Util/SupportedNetworkProtocolVersion.hs
jasagredo: Yes indeed, thanks

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
jasagredo: Using `<BLOCK HASH, padded to 32 bytes>` which is 32 bytes :smile:

jasagredo: And added the origin example

docs/website/contents/explanations/queries.md
jasagredo: Linking to the source now

docs/website/contents/explanations/queries.md
jasagredo: Yes good suggestion. Fixed.

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Generators.hs
jasagredo: True

cabal.project
jasagredo: Nah, this comment is just a mere hint. It doesn't have to be that precise, never have been and we don't expect it to be super precise. I wouldn't even add a comment but we agreed to add something that explains the why for the SRP.

dnadales (approved):
Thank you for addressing all the comments and for taking care of this 🫶

PR #2266, opened 2026-09-03
Store Peras certs when making boosted blocks immutable

nbloomf:
# Description

At this point our assumption is that Peras certificates will become part of the evidence chain when nodes which did not participate in voting need to catch up; this means they need to be kept somewhere and diffused.

This PR (in progress) introduces a new function to the cert Db API which allows consumers to query the map of ticket numbers to certs. We use it in the STM transaction where a block is removed from volatile storage to capture the certificates for that block which exist at that time; this is the last moment when a received Peras cert could have affected voting.

TODO:
- Testing
- Implement the immutable DB on disk


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -194,7 +195,11 @@ copyToImmutableDB cdb@CDB{..} = withWriteAccess cdbImmutableDBLock $ \() -> do
       -- TODO the invariant of 'cdbChain' is shortly violated between
       -- these two lines: the tip was updated on the line above, but the
       -- anchor point is only updated on the line below.
-      atomically $ removeFromChain pt
+      certsForThisPoint <- atomically $ do
+        removeFromChain pt
+        PerasCertDB.filterCertsByTicket cdbPerasCertDB
+          (\cert -> getPerasCertPoint cert == pt)
+      PerasCertImmutableDB.appendCerts cdbPerasImmutableCertDB pt
```
ninioArtillero, line 202: Still wondering about the appropriate naming convention for the new db. I'm currently inclining for `HistPerasCertDB`, because the concept of "Immutable" defaults to "blocks" in my mind. But we can solve this later.

ninioArtillero, line 202:
```haskell
      atomically $ removeFromChain pt
      certsForThisPoint <- do
        PerasCertDB.filterCertsByTicket
          cdbPerasCertDB
          (\cert -> getPerasCertPoint cert == pt)
      PerasCertImmutableDB.appendCerts cdbPerasImmutableCertDB (Map.elems certsForThisPoint)
```

I think the last line should look like this, but the other changes are just an alternative because I don't see why we would need to run `filterCertsByTicket` within STM.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/API.hs
```diff
@@ -84,6 +84,9 @@ data PerasCertDB m blk = PerasCertDB
   -- implementations with on-disk storage.
   --
   -- NOTE: Use the `join . atomically` pattern to consume its output.
+  , filterCertsByTicket ::
+      (WithArrivalTime (ValidatedPerasCert blk) -> Bool) ->
+      STM m (m (Map PerasCertTicketNo (WithArrivalTime (ValidatedPerasCert blk))))
```
ninioArtillero, line 89:
```haskell
      m (Map PerasCertTicketNo (WithArrivalTime (ValidatedPerasCert blk)))
```

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasCertDB/Impl.hs
```diff
@@ -315,3 +316,15 @@ implGarbageCollect PerasCertDbEnv{pcdbTracer, pcdbState} slotNo = do
...
+  forall m blk.
+  ( IOLike m
+  , IsPerasCert (PerasCert blk) blk
+  ) =>
+  PerasCertDbEnv m blk ->
+  (WithArrivalTime (ValidatedPerasCert blk) -> Bool) ->
+  STM m (m (Map PerasCertTicketNo (WithArrivalTime (ValidatedPerasCert blk))))
+implFilterCertsByTicket PerasCertDbEnv{pcdbState} p = do
+  PerasCertDbState{pcdsCertsByTicket} <- fmap forgetFingerprint $ readTVar pcdbState
+  pure $ pure $ Map.filter p pcdsCertsByTicket
```
ninioArtillero, line 330:
```haskell
implFilterCertsByTicket ::
  forall m blk.
  IOLike m =>
  PerasCertDbEnv m blk ->
  (WithArrivalTime (ValidatedPerasCert blk) -> Bool) ->
  m (Map PerasCertTicketNo (WithArrivalTime (ValidatedPerasCert blk)))
implFilterCertsByTicket PerasCertDbEnv{pcdbState} p = do
  PerasCertDbState{pcdsCertsByTicket} <- fmap forgetFingerprint $ atomically $ readTVar pcdbState
  pure $ Map.filter p pcdsCertsByTicket
```

ninioArtillero (commented):
This looks fine, and pinpoints where our first API function, `appendCerts`, will go. Thank you.
My comments are mostly the changes that follow if making `filterCertsByTicket` to return a plain `m` action. More than actually proposing these changes, they are just ment to perform the question: is there a reason to perform this query within STM?

PR #2272, opened 2026-09-07
Leios: Abstain rather than die on an unreadable EB closure

### LINKED ISSUE #1080: Certified EB with malformed tx CBOR crashes block producers (unhandled decode `error` in Shelley/Ledger/Leios.hs)
## Summary

A block producer can be crashed remotely by a **certified Endorser Block whose transaction bodies are well-formed at the fetch/hash layer but are not valid Leios-tx CBOR**. When a node later unpacks the EB and decodes one of its transactions, the Leios transaction decoder fails and `Ouroboros.Consensus.Shelley.Ledger.Leios` responds to the `DeserialiseFailure` with a partial `error`, which crashes the node.

Because the offending EB is **certified**, it diffuses network-wide; every block producer that unpacks it hits the same decode and dies. Relays are **not** affected — they diffuse the EB without decoding its transactions.

This was demonstrated by the Red Team against the shared Leios testnet.

## Impact (observed on the shared testnet)

- **IOG block producers were taken offline for the ~10-minute attack window** (verified independently by IOG infrastructure — the BPs show a clean down-band for the duration).
- Our single adversarial relay served the fabricated EBs to **60+ distinct peer IPs** — i.e. broad diffusion, consistent with a certified EB reaching the whole network.
- systemd auto-restarts the BPs, but a **sustained** attack (one adversarial EB per won slot) keeps producers cycling down and would disable the network until the adversarial pool is mitigated out of band.
- **Very low cost.** A small amount of stake suffices — a **single EB with 8 transactions is one VRF slot**. (One EB is not enough to hold the network down continuously, but a modestly-staked adversary producing one per won slot is.)

## Crash

```
cardano-node: ExceptionInLinkedThread "ThreadId 64" Failed to deserialise Leios tx:
  DecoderErrorDeserialiseFailure "Leios Tx" (DeserialiseFailure 1 "expected map len or indef")
  error, called at ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs:308:17
    in ouroboros-consensus-3.0.1.0-sKtFhA0hqpEG1XQe1tS3z-cardano:Ouroboros.Consensus.Shelley.Ledger.Leios
```

Independently reproduced on the `prototype-2026w34` line (cardano-node `11.1.0.164`, git `3e1bec02`, consensus `3.0.1.0`) where the same failure lands at **`Leios.hs:283:17`** — same bug, adjacent source lines across builds. In that reproduction it fired on the forge path (`Forge.Loop … get-ticked-ledger-state`) and took down the block-producing nodes with `Startup.DiffusionInit.DiffusionErrored`.

## Root cause

`Leios.hs:308` (and `:283` on the w34 build) handles a Leios-transaction `DeserialiseFailure` by calling `error` — a partial function — on data that originates from the network. Untrusted input that fails to decode must be returned as a rejectable decode error (decline the transaction / EB, drop or penalize the peer), never a fatal assertion.

## Why the fabricated EB is accepted before it crashes anything

The EB body is validated at the **outer layer** when it is fetched — the hash of the EB matches the announcement, and it will, because the adversary computes that hash over its own fabricated body. Availability/hash validation passes, so the EB is accepted, **certified**, and diffused across the network. Only later, when a node **unpacks the EB and reads a transaction** to use it, does it decode the transaction bytes — and that is where the random body bytes fail the tx decoder and trip the `error`. So the crash happens after acceptance, certification, and diffusion, on every producer that decodes a tx.

## The malformed transaction encoding (what the fix must tolerate)

Each fabricated transaction is a structurally-plausible 4-element tx array whose **body element is a random CBOR byte string where a transaction-body map is required**:

```
CBOR = array(4) [
  bytes(<250 random bytes>),   # tx BODY element — a bytestring, NOT a tx-body map
  map(0),                      # witness_set (empty)
  true,                        # is_valid
  null                         # aux_data
]
```

The decoder reads the array, expects the body to be a map (major type 5), finds a byte string (major type 2), and fails with `DeserialiseFailure 1 "expected map len or indef"`. The transactions are otherwise correctly signed and the EB is correctly announced, offered, fetched, and certified.

## Suggested fix

Replace the partial `error` at `Leios.hs:308` (and any sibling call site, e.g. `:283`) with proper error handling: surface the `DeserialiseFailure` as a decode/validation error, decline the transaction/EB, and drop or penalize the offering peer — do not crash the node. More generally, no decode of untrusted Leios data should reach a partial `error`.

Once a fixed release is available, the Red Team will re-run the attack to verify the crash is closed.

## Environment

- cardano-node `11.1.0.164`, `ouroboros-consensus-3.0.1.0`
- `prototype-2026w34` line; network magic 164 (shared Leios testnet)

ch1bo:
Load transactions defensively as they may not have been decoded before. Handle the error differently per call site.

Fixes https://github.com/input-output-hk/ouroboros-leios/issues/1080


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
```diff
@@ -128,32 +133,26 @@ instance
...
-        -- exposed the parking gap, instead of silently corrupting the
-        -- ledger state.
-        error $
-          "resolveLeiosClosure: EB closure missing from LeiosDb for point "
-            <> show ebHash
-            <> "; chain-sel selected a cert-RB without its EB closure. "
-            <> "Refusing to apply as empty (would diverge UTxO)."
+        -- A missing closure means the EB payload is not in this node's
+        -- LeiosDb. On the apply path that is fatal -- chain-sel selected a
+        -- cert-RB without its closure, which under the intended parking design
```
geo2a, line 140:
```suggestion
        -- cert-EB without its closure, which under the intended parking design
```
Nit: did you mean `cert-EB`? I don't think we usually talk about RB-closures.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
```diff
@@ -128,32 +133,26 @@ instance
...
+        -- it would not have done yet -- and 'unsafeClosure' still says so
+        -- loudly. Reporting it rather than dying here lets the vote path,
+        -- where it is merely a reason to abstain, decide for itself.
+        pure $ Left $ LeiosClosureMissing ebHash
       Just closureEntries ->
-        pure $ fmap (mkShelleyTx . deserialiseLeiosTx) <$> closureEntries
+        -- 'traverse' over 'Either' stops at the first undecodable tx: a
```
geo2a, line 146:
```suggestion
        -- 'forM' over 'Either' stops at the first undecodable tx: a
```

I have not verified the semantics of this.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -934,7 +935,23 @@ resolveLeiosBlock leiosDb cds b =
...
       resolveLeiosClosure leiosDb (pointEbHash announcedPoint)
-        <&> inlineLeiosClosure b . map snd
+        <&> inlineLeiosClosure b . map snd . unsafeClosure
+
+-- | Take the closure or die.
+--
+-- The apply path has no way to proceed without it: chain-sel already selected
+-- a block whose payload is this closure, so carrying on with fewer
+-- transactions would silently diverge the UTxO -- which is what the previous
+-- @pure []@ did, and why this fails loudly instead. Under the intended parking
```
geo2a, line 945: I'm missing context on what "the intended parking design" is, perhaps it's worth removing this sentence for now?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
```diff
@@ -128,32 +133,26 @@ instance
...
+        -- closure is only useful whole, so decoding the rest would be work
+        -- thrown away.
```
bladyjoker, line 148: Good catch!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -168,9 +168,11 @@ chainSyncBlocksServer tracer chainDB ccfg leiosDb flr = ChainSyncServer $ do
...
           Left _ -> pure sblk
           Right blk -> do
-            resolveLeiosClosure leiosDb (pointEbHash prevAnn)
-              <&> inlineLeiosClosure blk . map snd
-              <&> encode
+            resolveLeiosClosure leiosDb (pointEbHash prevAnn) >>= \case
+              -- Serve what we have rather than dying on a closure we cannot
+              -- read; the peer validates the block regardless.
+              Left _ -> pure sblk
+              Right closure -> pure . encode $ inlineLeiosClosure blk (map snd closure)
```
bladyjoker, line 175: Why not die? At this point we're ChainSyncing invalid chains (with invalid closures).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
bladyjoker: I can't grok this comment too

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
```diff
@@ -966,7 +983,7 @@ resolveAndApplyLeiosClosure ::
   m (Either (LedgerErr (LedgerState blk)) (LeiosClosureApplied blk))
 resolveAndApplyLeiosClosure leiosDb lcfg ebHash readValues extraKeys lsBase = do
   -- Load EB txs from disk
-  closureTxs <- map snd <$> resolveLeiosClosure leiosDb ebHash
+  closureTxs <- map snd . unsafeClosure <$> resolveLeiosClosure leiosDb ebHash
```
bladyjoker, line 986: I think this is the same as with ChainSync, at this point the pre-condition is that this is a certified EB and must be be valid?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
ch1bo: CertRB is what I meant here. ChainSel selects ranking blocks. A CertRB is a block that carries a Leios cert. "Its closure" is the txs that got certified through it.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Leios.hs
ch1bo: I find `forM` and `mapM` also more intuitive, just looked it up, it's the same: https://hackage-content.haskell.org/package/ghc-internal-9.1401.0/docs/src/GHC.Internal.Data.Traversable.html#mapM

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
ch1bo: I somewhat agree, but not fully crash: This is the N2C server, so it would actually be best to err out on the connection -> the user must be able to look up what happens from the logs as N2C must be locally used.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/LedgerDB/Forker.hs
ch1bo: Dropped

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
ch1bo: I opted to not change the semantics here, we do return the `sblk` in other branches, let's stick with this (until we make this production grade).

PR #2276, opened 2026-09-08
[Peras 68] Rewire DijkstraEra to use Peras V1 types

agustinmista:
This PR rewires the Peras implementation for the `DijkstraEra` to use V1 (production) types backed by BLS crypto. 

Concretely, it:
 
- Adds Ledger BLS-key conversion helpers and implements V1 committee-input construction for `WFALS` from pools with registered BLS keys.
- Replaces Dijkstra’s placeholder Peras types with the corresponding V1 ones for votes, certificates, errors, voting committee scheme, etc..
- Adds property tests for committee key/stake extraction, covering the case of pools without registered BLS keys.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/V1.hs
```diff
@@ -0,0 +1,92 @@
...
+-- | Key scope used for deriving the BLS public keys from the ledger.
+ledgerKeyScope :: BLS.KeyScope
+ledgerKeyScope = "PERAS/LEDGER"
+
+extractPerasStakeDistrAndPublicKeys ::
+  PoolDistr ->
+  Map PoolId (LedgerStake, PerasPublicKey)
+extractPerasStakeDistrAndPublicKeys =
+  Map.mapKeysMonotonic PoolId
+    . Map.mapMaybeWithKey extractEntry
```
jasagredo, line 50: Any reason to not use `mapMaybe` instead?

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/V1.hs
```diff
@@ -0,0 +1,114 @@
...
+      foldr
+        (\(keyCase, label) -> cover 1 (hasCase keyCase) label)
+        (counterexample (show entries) $ Map.keysSet result === expectedPoolIds)
+        coverageLabels
+
+tests :: TestTree
+tests =
+  testGroup
+    "V1"
+    [ adjustQuickCheckTests (* 10) $
```
jasagredo, line 110: Why?

jasagredo (approved):
Nothing blocking so I am approving

ouroboros-consensus/test/consensus-test/Test/Consensus/Peras/Voting/V1.hs
agustinmista: Fair enough. 100 tests are probably enough to validate this property.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Peras/Voting/V1.hs
agustinmista: Good catch. Fixed!

PR #2280, opened 2026-09-09
Leios: Introduce TxEbMeasure and size Mempool accordingly

### LINKED ISSUE #1107: Production-grade forging of EBs
## Why
- Need to produce EBs for enabling Leios on mainnet
- A clean implementation good enough for initial mempool and EB sizes is needed

## What
- The forge loop creates and announces EBs (once an RB is full)
- The forge loop certifies an EB when possible (only if https://github.com/input-output-hk/ouroboros-leios/issues/1073 available)
- Only does these things in `DijkstraEra`
  - It's perfectly fine if only the `ShelleyBlock` instance in `DijkstraEra` has these semantics -> no re-use in other block types needed
- Can handle ~2MB mempools and ~1MB EB sizes within 500ms (not release blocks too late) - TBD: lower deadline?
- (optional) Pre-application of certified EB to announce the remainder right away
- (optional) Store the EB in the LeiosDb (only if https://github.com/input-output-hk/ouroboros-leios/issues/1103 available)

## How

- `forgeBlock` gets the whole `MempoolSnapshot`
- `DijkstraEra` partitions the `MempoolSnapshot` using the `TxEbMeasure` and creates an RB/EB pair accordingly
  - PR of `TxEbMeasure` on prototype: https://github.com/IntersectMBO/ouroboros-consensus/pull/2280
  - PR of `TxEbMeasure` on `master`
- Traces production of EB (and writes to LeiosDB)

ch1bo:
This is what gets configured through protocol parameters and needs to be bounding capacity of an endorser block.

While exploring, I also realized that we are (by now) using the wrong serialization for transactions when creating an EB (see FIXME)


ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -163,7 +163,7 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
     Just ebTxs -> do
       let forgedEb = forgeLeiosEb fbCurrentSlotNo ebTxs
           ebHash = hashLeiosEb forgedEb.body
-          ebSize = leiosEbBytesSize forgedEb.body
+          ebSize = encodeLeiosEbSize forgedEb.body
```
ch1bo, line 166: Renamed these functions to have the same prefix as the `encodeLeiosEb` and all of them must be in agreement.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -686,11 +686,16 @@ instance
...
 -- | Dijkstra-era block measure.
 --
--- Extends 'ConwayMeasure' with a transaction-count limit that the Leios
--- mempool snapshot uses for Endorser Blocks.
+-- Extends 'ConwayMeasure' size of references. This is used to measure endorser
+-- blocks. See 'ebCapacityTxMeasure'.
 data DijkstraMeasure = DijkstraMeasure
   { conwayMeasure :: !ConwayMeasure
-  , leiosMaxTxsPerEb :: !(IgnoringOverflow TxCount)
+  , referencesSize :: !(IgnoringOverflow ByteSize32)
```
ch1bo, line 693: This is the main change of this. I was doing it to allow for bigger EBs than a hard-coded cap.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -2126,16 +2152,10 @@ maxEbTxCount referencesSize =
...
+-- FIXME: get rid of this
 maxTxsPerEb :: Int
 maxTxsPerEb = maxEbTxCount maxMsgLeiosBlockBytesSize
 
--- | The largest EB closure we will fetch.
---
--- FIXME: see the note above -- @maxEndorserBlockTxsSize@ is the ledger's say on
--- the same quantity, and the two are unrelated today.
-maxEBClosureSize :: ByteSize32
-maxEBClosureSize = ByteSize32 12_000_000
```
ch1bo, line 2137: Was already unused

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -2126,16 +2152,10 @@ maxEbTxCount referencesSize =
 --
 -- Sizes buffers that are allocated before any ledger state is in reach; the
 -- protocol parameter cannot exceed it.
+-- FIXME: get rid of this
 maxTxsPerEb :: Int
 maxTxsPerEb = maxEbTxCount maxMsgLeiosBlockBytesSize
```
ch1bo: This hard-coded cap is what was bounding now and not governable through pparams.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
geo2a: Is that the size of one transaction reference in an EB, i.e. the size of `TxHash`?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -686,11 +686,16 @@ instance
 
 -- | Dijkstra-era block measure.
 --
--- Extends 'ConwayMeasure' with a transaction-count limit that the Leios
--- mempool snapshot uses for Endorser Blocks.
+-- Extends 'ConwayMeasure' size of references. This is used to measure endorser
```
geo2a, line 689:
```suggestion
-- Extends 'ConwayMeasure' with the size of references. This is used to measure endorser
```

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
bladyjoker: Got me confused a bit, thinking this is about ref inputs x)

Consider `txReferencesSize`? Or `ebBodySize` (how did we end up calling this?)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -724,9 +729,20 @@ blockCapacityDijkstraMeasure ::
...
+    { conwayMeasure = conway
+    , -- Ranking blocks carry transactions, not references, so this must never
+      -- bound a fill. It has to be 'maxBound' specifically, not merely large:
+      -- 'partitionMempool' takes the EB with @snapshotTake snap (rbCap + ebCap)@,
+      -- and only @maxBound + ebCap@ wraps (under 'IgnoringOverflow') back to
+      -- ~ebCap, so the ranking block contributes nothing to the EB's reference
+      -- budget. A finite value here inflates it and the EB overshoots
+      -- @maxEndorserBlockReferencesSize@. 'computeMempoolCapacity' 'stimes' also
+      -- wraps @maxBound + maxBound@ to ~maxBound, so the mempool stays unbounded
+      -- on this axis too.
```
geo2a, line 741: I find this comment very hard to read. Maybe rewrite it as something like:

> We over-approximate the size of tx references of ranking blocks, because ranking blocks carry transactions, which are significantly larger than their references. We are intentionally muddling the difference between transactions and references for RBs to re-use the measure interface.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -742,13 +758,24 @@ txMeasureDijkstra ::
   GenTx (ShelleyBlock proto era) ->
   V.Validation (TxErrorSG era) DijkstraMeasure
 txMeasureDijkstra st tx =
-  (\c -> DijkstraMeasure c oneTxCount) <$> txMeasureConway st tx
+  alsoMeasureDijkstra <$> txMeasureConway st tx
+ where
+  -- A reference records the length of the bytes 'forgeLeiosEb' serialises. We
```
geo2a, line 763: This comment is hard for me to read as well, but I don't really have a suggestion on how to rewrite it, perhaps I'm missing some context.

Also, I think the comment, whatever it says, should attach to the definition of `referencesSize` and not `alsoMeasureDijkstra`.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -724,9 +729,20 @@ blockCapacityDijkstraMeasure ::
...
+    , -- Ranking blocks carry transactions, not references, so this must never
+      -- bound a fill. It has to be 'maxBound' specifically, not merely large:
+      -- 'partitionMempool' takes the EB with @snapshotTake snap (rbCap + ebCap)@,
+      -- and only @maxBound + ebCap@ wraps (under 'IgnoringOverflow') back to
+      -- ~ebCap, so the ranking block contributes nothing to the EB's reference
+      -- budget. A finite value here inflates it and the EB overshoots
+      -- @maxEndorserBlockReferencesSize@. 'computeMempoolCapacity' 'stimes' also
+      -- wraps @maxBound + maxBound@ to ~maxBound, so the mempool stays unbounded
+      -- on this axis too.
+      referencesSize = maxBound
```
bladyjoker, line 742: I don't get this, shouldn't this be zero for ranking blocks?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -758,38 +785,42 @@ leiosEndorserBlockMeasure ::
...
+    }
+ where
+  ByteSize32 framing = Leios.encodeLeiosEbMaxFramingSize
+
+  -- Capped by the codec's message limit: an EB no peer will accept is worse
+  -- than a smaller EB.
+  -- XXX: Silent cap, improve detectability
+  referencesLimit =
+    min Leios.maxMsgLeiosBlockBytesSize $
+      pparams ^. ppMaxEndorserBlockReferencesSizeL
```
bladyjoker, line 817: This seems pretty impactful and somehow out of place.

I wonder if there isn't a better place to establish `pparams ^. ppMaxEndorserBlockReferencesSizeL <= Leios.maxMsgLeiosBlockBytesSize`? Perhaps when starting Leios mini protocols? Fail hard if you're configured to run with EB body sizes smaller than the max EB body size?

For that you'd need pparams, not sure that's available there.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
```diff
@@ -539,8 +539,10 @@ runWith RunNodeArgs{..} encAddrNtN decAddrNtN LowLevelRunNodeArgs{..} =
 
           -- Created before the ChainDB so its GC thread can prune the cache to
           -- each GC slot just before GCing the LeiosDb (the ordering contract in
-          -- "LeiosTxCache"). 2^22 slots (~168 MiB) is ~2x the ~1.9M worst case
-          -- (128 EBs x maxTxsPerEb), so the table never fills.
+          -- "LeiosTxCache"). The table is fixed at 2^22 slots (~160 MiB): a
+          -- lock-free structure allocated before any ledger state is in reach,
+          -- so it cannot follow the leios protocol parameters.
+          -- TODO: Emit a big warning if tx cache size turns out to be insufficient given protocol parameters
```
bladyjoker, line 545: Hmm similar issue as above? One needs to set the parameters informed by the pparams but they are not available when these components are instantiated.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -1069,6 +1057,42 @@ encodeLeiosEb (MkLeiosEb v) =
...
+    + sum (fmap (unByteSize32 . encodeLeiosEbItemSize . ByteSize32 . snd) items)
+
+-- | Length of a unsigned integer if it were encoded in a "flattened format".
+-- See 'encodeInteger'.
+cborIntBytesSize :: Integral i => i -> BytesSize
+cborIntBytesSize n
+  | n < 24 = 1
+  | n < 0x100 = 2
+  | n < 0x10000 = 3
+  | otherwise = 5
```
bladyjoker, line 1094: Not excited we have to look so low in CBOR to compute this. Is there something from the cbor library that can give us this.

bladyjoker (commented):
A couple of high level concerns

1. The byte size parameters are a bit too involved for my taste. I "guess" it's correct, but I'd rather see a test suite showing that some generated/golden "ebs" and "eb items" have sizes that match the expected computed ones. We're calculating the size by assuming a particular CBOR structure, if the codec changes this calculation can be silently/subtly broken too.

2. Back in the day we considered introducing a separate `XyzEbMeasure` to distinguish from `TxsRbMeasure`, that would be cleaner on the definition side of things, however, `TxSeq` is defined on a single "measure" that `take` checks, so not clear to me how this would work other than the "maintain 2 tx seqs, one for txsRb and one for eb, each with their own measure".

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
dnadales: We might consider using https://github.com/AminBlg/SimpleEnglish

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -758,38 +785,42 @@ leiosEndorserBlockMeasure ::
...
+          , refScriptsSize =
+              IgnoringOverflow . ByteSize32 $ pparams ^. ppMaxRefScriptSizePerEndorserBlockL
+          }
+    , -- Summing 'leiosEbItemBytesSize' over transactions accounts for the
+      -- references and nothing else, so the framing 'encodeLeiosEb' writes
+      -- ahead of them comes off the capacity here. Guarded: an absurdly small
+      -- parameter must exhaust the capacity, not wrap it around.
+      referencesSize =
+        IgnoringOverflow . ByteSize32 $
+          referencesLimit - min referencesLimit framing
```
dnadales, line 825: Does this guard survive the addition in `partitionMempool`?

The subtraction here floors at 0, so a parameter of 0 to 5 gives `ebCap.referencesSize = 0`.
`partitionMempool` then computes `plus rbCap ebCap`, and `rbCap.referencesSize` is `maxBound`:

    parameter 6  ->  ebCap 1  ->  maxBound + 1 = 0          -- nothing fits
    parameter 0  ->  ebCap 0  ->  maxBound + 0 = maxBound   -- no limit

So a parameter of 6 gives us an empty EB, and a parameter of 0 gives us an EB that only the closure axis bounds.
Zero is the one value that does not carry past the top of the type.

Dijkstra's `ppuWellFormed` carries `/= 0` guards for eight parameters, but none for `ppuMaxEndorserBlockReferencesSizeL`.
So the ledger accepts a governance update to 0, and 0 looks like the natural way to switch Leios off.

Should the "no room" case yield 1 rather than 0, so the wrap lands on "nothing fits"?
Something like:

```haskell
  referencesSize
    -- 'partitionMempool' adds this to the RB's 'maxBound'. A 0 here comes back
    -- out as 'maxBound', so use the smallest value that admits no reference.
    | referencesLimit <= framing = IgnoringOverflow (ByteSize32 1)
    | otherwise = IgnoringOverflow . ByteSize32 $ referencesLimit - framing
```

That keeps the current scheme.
Going further and replacing the `maxBound` sentinel would remove the class of bug, and the same arithmetic already makes the two `partitionMempool` branches differ by one byte: the non-certifying branch gets `L - 6` and the certifying branch gets `L - 5`.
Filed that part as #2291, so it need not hold up this PR.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
dnadales: Should we compare `ebSize` against the wire limit here, while we have it in hand?

Two things together let a forged EB exceed that limit.

The codec's `SingBlock` limit is `largeByteLimit`, and the driver measures the whole message.
`MsgLeiosBlock` adds 2 bytes on top of the body (`encodeListLen 2` then `encodeWord 1`).
But `referencesLimit` at `Mempool.hs:816` caps against the full `maxMsgLeiosBlockBytesSize`, so the body budget is 2 bytes too generous.

Those 2 bytes normally hide.
`encodeLeiosEbMaxFramingSize` reserves 5 bytes for the map prefix, and the real prefix is 3 while the item count stays under 65536.
Above that count the prefix becomes 5, the slack goes, and the message runs 2 over:

    n <  65536:  items 2499995 + prefix 3 + msg 2  =  2500000   fits
    n >= 65536:  items 2499995 + prefix 5 + msg 2  =  2500002   over

65536 items looks reachable once the parameter is high enough.
A transaction of 24 to 251 bytes is charged 36 on this axis, and 65536 * 36 is 2359296, inside the budget.

On breach the receiving peer throws `ExceededSizeLimit` and drops the connection.
The EB never diffuses and never gets certified.
We would not learn why, because `ForgedBlock{} -> pure ()` skips validation of our own EB.

Would it be worth tightening the cap and checking here as well?
Something like:

```haskell
  referencesLimit =
    min (Leios.maxMsgLeiosBlockBytesSize - Leios.msgLeiosBlockFramingSize) $
      pparams ^. ppMaxEndorserBlockReferencesSizeL
```

with a trace or assertion on `ebSize` at this line.
Then a later change to either encoder shows up as a local failure, rather than as peers hanging up on us.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
dnadales: Should we put the two numbers into this TODO, so the gap is visible without recomputing it?

`worstCaseCacheTxCount = maxAnnouncementCount * maxTxsPerEb` (`LeiosTxCache/API.hs:163`) now reads 128 * 71428 = 9142784, against the 4194304 slots the next line allocates.
The text this replaced carried the proof the other way round: "~2x the ~1.9M worst case, so the table never fills".

To be fair to the PR, it did not open this gap.
The codec's `SingBlock` limit was already 2500000, so a peer could always deliver about 71000 references in one EB.
The old 13888 understated that limit by 5x, and the "never fills" proof rested on it.
This PR corrects the number, and that is what makes the shortfall show.

Worth separating the two bounds here, because only one of them is short.

Per EB the bound holds by construction, and the doc comment on `maxTxsPerEb` is right about it: a 71429th reference needs 2500022 bytes, so the codec drops the message, and the fetch buffers sized at `maxTxsPerEb` hold any legal body exactly.

The aggregate is what has no bound. The codec limits one message, not the 128 the cache holds at once, so a per-body check would not help. That leaves a sizing decision: more slots, a smaller `maxAnnouncementCount`, or a residency bound that does not follow `maxTxsPerEb`. For the first option the target is not `slots >= 9142784`, since the table probes linearly and never resizes, so lookups slow well before it fills.

What the TODO leaves open is that overflow is not a degradation either.
`MutableHashTable.insert` raises `error "table full"` (`MutableHashTable.hs:235`), and a warning would not stop that.

Filed as #2290 if you would rather keep it out of this PR. Could we at least record here that `maxAnnouncementCount * maxTxsPerEb` no longer fits `2 ^ 22`?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -2086,55 +2110,40 @@ traceLeiosPeerForHuman = \case
...
-maxEbTxCount referencesSize =
-  fromIntegral $
-    (fromIntegral referencesSize - msgOverhead - sequenceOverhead)
-      `div` minEbItemBytesSize
+-- Those buffers are allocated before any ledger state is in reach, which is
+-- why this derives from the codec limit and not from the protocol parameter;
+-- the mempool in turn plans EBs within the codec limit, so a forged EB always
+-- fits the buffers of every honest peer.
+maxTxsPerEb :: Int
+maxTxsPerEb =
```
dnadales, line 2143: Should `ibsCacheLoad` divide by the table's capacity rather than by `worstCaseCacheTxCount`?

(Raising it here because `maxTxsPerEb` is what moved. The metric lives in `LeiosTxCache/API.hs:175`, which this PR does not touch.)

The figure we trace on every body insert is:

    ibsCacheLoad = cacheTxCount / worstCaseCacheTxCount

`cacheTxCount` is `HT.size ht`, the live entry count.
`worstCaseCacheTxCount` is `maxAnnouncementCount * maxTxsPerEb` (`API.hs:163`), so it tracks this constant.
The table itself is 2 ^ 22 = 4194304 slots.

Before this change the denominator was 1777664, below the table size, so 1.0 was reachable and meant "at the worst case, with headroom left".
Now the denominator is 9142784, above the table size:

    table 100% full    4194304 / 9142784  =  0.46
    probing degrades   2900000 / 9142784  =  0.32

So the figure tops out at 0.46, and it reaches that value at the moment `insert` raises `error "table full"`.
Nothing else reports occupancy: `MutableHashTable.capacity` is exported at `MutableHashTable.hs:92` and has no callers anywhere.

Threading `HT.capacity ht` into `mkLeiosTxCacheInsertBodySummary` in place of the constant would keep the figure meaning "how full", and leave the worst-case number as a separate check rather than a scale.

`LeiosTxCache/Optimized.hs:59-60` still says "use 22 for the ~1.9M worst case", which is the same staleness in a docstring.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -2086,55 +2110,40 @@ traceLeiosPeerForHuman = \case
...
+  -- can be: both come from the encoder, so the buffers track it exactly.
+  framing = fromIntegral $ unByteSize32 encodeLeiosEbMaxFramingSize
 
--- | The largest EB closure we will fetch.
---
--- FIXME: see the note above -- @maxEndorserBlockTxsSize@ is the ledger's say on
--- the same quantity, and the two are unrelated today.
-maxEBClosureSize :: ByteSize32
-maxEBClosureSize = ByteSize32 12_000_000
+  minItemSize = fromIntegral $ unByteSize32 (encodeLeiosEbItemSize (ByteSize32 0))
```
dnadales, line 2152: Worth a follow-up on the tx cache bench in the same pass?

`bench/leios-txcache-bench/Main.hs:59` reads "(The current CBOR-precise `LeiosDemoTypes.maxTxsPerEb` is 13888.)", and `:61` hardcodes `txsPerEb = 15_058`.
The header at `:13-15` derives its whole worst case from "~512 kB / 34 B" and "~1.9M distinct txs resident".

With `maxTxsPerEb` at 71428 the bench populates 128 * 15058 = 1.93M entries against a 9.14M ceiling, into the same `newHashTableLeiosTxCache 22` the node uses (`:169`).
Because the number is hardcoded rather than written as `= maxTxsPerEb`, the bench keeps passing and stops covering the ceiling it exists to cover.

Would `txsPerEb = maxTxsPerEb` be the right shape, so the bench follows the constant from now on?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
```diff
@@ -2086,55 +2110,40 @@ traceLeiosPeerForHuman = \case
   TraceLeiosPeerAnnouncement equiv fields ->
     "EB announcement from peer (" <> T.pack (show equiv) <> "): " <> T.pack (show fields)
 
--- * Protocol parameters
+-- * Protocol limits and parameters
```
dnadales, line 2113: Should the last mention of `maxEBClosureSize` go with it?

`Cardano/Tools/DBAnalyser/Analysis.hs:1088-1090` still reads:

    -- This pass models the mempool of a forging node, so the capacity must
    -- match. A Leios node holds a whole EB closure, and a closure reaches
    -- 12 MB ('maxEBClosureSize'). One mebibyte is not enough.

That file is not in the diff, and it is the only surviving reference to the symbol.

The 12 MB figure is still right, because `leiosMempoolSize` just above it takes `ppMaxEndorserBlockTxsSize` from the chain.
Only the name is dead, so dropping the parenthesis would be enough.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: > Is that the size of one transaction reference in an EB, i.e. the size of TxHash?

Yes. Of one tx or of all the references in a block or the mempool (whatever we measure)

`txReferencesSize` works. There was a huge discussion about not having body in the name.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: It should, but we use the praos block measure to size the mempool. I'm not sure why I worked around this. Instead the mempool should be sized from both, normal block measure and endorser block measure..then 0 here should be fine.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: The mini protocol buffers will likely be set by the guard rails maximum value. And yes, we probably should fail hard if we ever learn the pparams be higher than that. So `error` here instead of just the `XXX` note?

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
ch1bo: Yes, hence the TODO

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoTypes.hs
ch1bo: No, I was surprised also when I wrote this around last year (it was just moved in this PR)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: Agreed, @bladyjoker also pointed this out above. The `maxBound` workaround is quite dirty and wrong in this way.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
ch1bo: I agree with the tightening on the measure and eventually put a big warning should the governed value every hit the hard-coded wire limit (which will correspond to the guard rail maximum most likely)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node.hs
ch1bo: > To be fair to the PR, it did not open this gap

Thank you :)

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: Made this an `assert`

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: Avoided this by a complete rewrite

ch1bo:
@bladyjoker @geo2a @dnadales I have rewritten the most of this branch in the latest commit. This is now introducing `TxEbMeasure` to `TxLimits` to measure things more accurately and size the mempool accordingly. No more `maxBound` to paper over RBs not having a txReferences capacity int he first place.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -381,14 +387,54 @@ class
...
+
+-- | The trivial 'TxEbMeasure' of blocks without Leios endorser blocks.
+data NoEbMeasure = NoEbMeasure
+  deriving stock (Eq, Ord, Generic, Show)
+  deriving anyclass NoThunks
+
+instance Measure NoEbMeasure where
+  zero = NoEbMeasure
+  plus _ _ = NoEbMeasure
+  min _ _ = NoEbMeasure
```
dnadales: Should we make a zero endorser-block capacity take no transactions, rather than all of them?

The doc comment on `ebCapacityTxMeasure` above says a zero capacity takes nothing. For `NoEbMeasure` it takes everything.

`Data.Measure` builds the order out of `min`:

```haskell
-- measures-0.1.0.2, Data/Measure.hs
x <= y = x == min x y
```

With `min _ _ = NoEbMeasure` that reduces to `NoEbMeasure == NoEbMeasure`, which is `True`. So `<=` is the constant `True` here, not a size order. The instance is honest about a type that carries no information. The order it yields is the problem.

`snapshotPartition` splits on exactly that test (`Mempool/Impl/Common.hs:602`):

```haskell
(y, _) = TxSeq.splitAfterTxSizeOn mmTxEbMeasure rest ebLimit
```

and `splitAfterTxSizeOn` asks `not (proj (mSize m) Measure.<= n)` (`Mempool/TxSeq.hs:230`). The test never holds, so `FingerTree.split` puts the whole tree on the left. `y` becomes the full remainder of the mempool.

`partitionMempool` runs for every forge, with no era gate (`NodeKernel/Forge.hs:215`). So this reaches `SimpleBlock`, `ByronBlock`, `DualByronBlock`, every pre-Dijkstra `ShelleyBlock`, the two-era `ShelleyBasedHardForkBlock`s and the HFC test block. `CardanoBlock` is safe, because `DijkstraEbMeasure` derives a real componentwise order.

The visible effect is at `NodeKernel/Forge.hs:525-537`. If we forge a block, do not adopt it, and the ChainDB reports it invalid, we pass `rbTxs ++ ebTxs` to `removeTxsEvenIfValid`. With `ebTxs` now the whole tail, we drop the entire mempool. Before this PR `ebCapacityTxMeasure` gave `Nothing`, `ebCap` became `zero`, and `drop (length rbTxs') allTxs` was always `[]`.

I wonder if a guard in `snapshotPartition` is the cheapest fix. Something like:

```haskell
, snapshotPartition = \blockLimit ebLimit ->
    let (x, rest) = TxSeq.splitAfterTxSizeOn mmTxMeasure txs blockLimit
        (y, _)
          | ebLimit == Measure.zero = (TxSeq.Empty, rest)
          | otherwise = TxSeq.splitAfterTxSizeOn mmTxEbMeasure rest ebLimit
     in (txSeqToList x, TxSeq.toSize x, txSeqToList y, TxSeq.toSize y)
```

`TxLimits` already requires `Eq (TxEbMeasure blk)`, so the guard needs no new constraint. It also short-circuits the pre-Dijkstra Cardano case, where today we walk the mempool only to find that no reference fits.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -866,6 +759,106 @@ txMeasureConway st tx@(ShelleyTx _txid tx') =
...
+instance Monoid DijkstraEbMeasure where
+  mappend = (<>)
+  mempty = DijkstraEbMeasure mempty mempty
+
+-- | The cost of one transaction in an endorser block: its closure cost is the
+-- transaction's block measure, and its reference costs the bytes
+-- 'forgeLeiosEb' serialises for it.
+--
+-- The reference records the length of the serialised transaction. The Conway
+-- byte size overstates that length by at most 'perTxOverhead', and a larger
```
dnadales: Should we phrase this as true at the current ledger pin, and add a test that keeps it true?

Claude went through the pinned ledger (`1587f21a`). The claim holds, and the margin at the argument is wider than stated. The two encodings are identical, so the difference is 0 rather than at most 4:

```haskell
-- Dijkstra/Tx.hs:293-297   toCBORForSizeComputation, which sizeTxF re-encodes
encodeListLen 3 <> encCBOR body <> encCBOR wits <> encodeNullStrictMaybe encCBOR aux

-- Dijkstra/Tx.hs:384-398   toCBORForMempoolSubmission, which serialize' resolves to
Rec DijkstraTx !> To dtBody !> To dtWits !> OmitC dtIsPhase2Valid !> E ... dtAuxData
--                                          ^ OmitC contributes 0 to the list length
```

`OmitC` is what makes them agree. That is a two-token choice in a dependency we do not control, and the ledger flags it right above the definition (`Dijkstra/Tx.hs:376-383`):

> this serialisation is neither the serialisation used on-chain ... nor the serialisation used for computing the transaction size.

The neighbouring era already disagrees. Alonzo uses `To atIsPhase2Valid` rather than `OmitC` (`Alonzo/Tx.hs:496-503`), so for Conway `BS.length (serialize' tx) == sizeTxF + 1`. Same pinned commit, different answer.

The usable headroom is also thinner than four bytes. `encodeLeiosEbItemSize` is a step function over `cborIntBytesSize`, so `+ perTxOverhead` on the argument normally adds 0 bytes to the charge. A drift of five bytes or more flips the over-charge into an under-charge, and then we forge an EB above `referencesLimit` that no peer accepts.

Nothing would catch that. The new properties in `Test/LeiosDemoTypes.hs` quantify over an arbitrary `Word32` size and never build a transaction. There is no type-level link either: `txEbMeasure` takes only `TxMeasure blk`, so the encoder it has to match is not in scope.

What about a property that ties the charge to the encoder, so a ledger bump fails here rather than on the wire? Something like:

```haskell
-- the mempool must never charge less than forgeLeiosEb writes
prop_ebItemChargeCoversEncoder st tx =
  BS.length (serialize' tx)
    <= fromIntegral (unByteSize32 (txMeasureByteSize (txMeasureConway st tx)))
```

Then this comment can say what it now says, with "at the current ledger pin" in front of it.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Capacity.hs
```diff
@@ -65,21 +74,23 @@ computeMempoolCapacity ::
...
+      `Measure.plus` ebClosureCapacityTxMeasure cfg st
+  ByteSize32 oneUnitBytes = txMeasureByteSize oneUnit
 
-  blockCount = case override of
+  unitCount = case override of
     NoMempoolCapacityBytesOverride -> 2
     MempoolCapacityBytesOverride (ByteSize32 x) ->
       -- This calculation is happening at Word32. If it was to overflow, it
       -- will round down instead.
       max 1 $
```
dnadales, line 87: Should the haddock above also say that an override below one unit is no longer expressible?

`max 1` is not new. It reads the same at `origin/leios-prototype`. What moves is the unit it floors to, and on a Leios chain that is a large move:

```
unit = blockCapacity + ebClosureCapacity

  pre-Leios      90112 - 1024                 =     89,088 B
  Dijkstra       89,088 + 12,582,912          = 12,672,000 B      (142x)
```

The 12 MiB is `maxEndorserBlockTxsSize` from `exampleDijkstraGenesis`.

So an operator who asks for 2 MiB used to get 24 units, about 2.14 MB. On a Dijkstra chain they now get `max 1` unit, about 12.67 MB. The figure our own mainnet config documents, `MempoolCapacityBytesOverride: 1000000`, lands at 12.7x what it asks for.

The path is live: the config key reaches `srnMaybeMempoolCapacityOverride` (`Run.hs:519`), then `Node.hs:1192-1196`, then here. No config in the tree sets a number today, so nothing breaks now. It becomes a trap the first time somebody tunes this knob on a Leios chain and gets an order of magnitude more mempool than they asked for.

The haddock at lines 61-62 describes the reinterpretation, which is the part that changed shape. What about naming the consequence too? Something like:

```haskell
-- If an override is present, reinterpret it as a number of such forging units
-- (rounded up, at least one), and then simply multiply the unit capacity by
-- that number. Under Leios one unit is an endorser block's closure, so an
-- override smaller than that yields one whole unit rather than the value asked
-- for.
```

ouroboros-consensus/test/consensus-test/Test/LeiosDemoTypes.hs
```diff
@@ -42,17 +46,19 @@ tests :: TestTree
 tests =
   testGroup
     "LeiosDemoTypes"
-    [ testProperty "leiosEbBytesSize consistent with encodeLeiosEb" prop_ebBytesSizeConsistent
+    [ testProperty "encodeLeiosEbSize consistent with encodeLeiosEb" prop_ebBytesSizeConsistent
+    , testProperty "encodeLeiosEbItemSize consistent with encodeLeiosEb" prop_ebItemSizeConsistent
+    , testProperty "encodeLeiosEb framing bounded by encodeLeiosEbMaxFramingSize" prop_ebFramingBounded
```
dnadales, line 54: Should we add a property for the boundary this PR fixes, while we are adding properties here?

These three pin the encoder size functions, which is the part that was easy to get wrong before. The two behaviours the PR exists to change have nothing holding them.

The first is the #2291 boundary. The old code read the endorser-block limit back through `Word32` wrap, so a parameter of 0 came out as `maxBound` and meant "no limit". The rewrite floors at 0, and 0 now means "nothing fits". That flip is one line of arithmetic in `leiosEndorserBlockMeasure`, and today nothing notices if it regresses:

```haskell
-- the values that used to wrap: 0 to 5 all floor to a zero capacity
prop_tinyReferencesLimitAdmitsNothing =
  forAll (chooseEnum (0, 5)) $ \l ->
    txReferencesSize (leiosEndorserBlockMeasure (stWith l)) === IgnoringOverflow (ByteSize32 0)
```

The second is `snapshotPartition` itself. It has two call sites and no test. Its contract is new: a prefix bounded on one axis, then a run bounded on a different axis. `splitAfterTxSizeOn` carries a stated precondition about the projection distributing over `plus`, and `splitAfterTxSizeSpec` was not generalised alongside it, so `prop_TxSeq_splitAfterTxSize` still only exercises the `id` projection.

The one genesis fixture that sets these parameters, `dijkstra-genesis.json:265-271`, zeroes all three, and its suite does not forge endorser blocks. `Test/ThreadNet/Leios.hs` runs the ledger's example values, so it exercises the middle of the range and never the edge.

I am not asking for coverage of the whole rework. Two small properties, the one above and a `snapshotPartition` one over a hand-built `TxSeq`, would keep the fix from quietly coming undone.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -459,9 +458,22 @@ data MempoolSnapshot blk = MempoolSnapshot
   -- ^ Get all transactions (oldest to newest) in the mempool snapshot,
   -- along with their ticket number, which are associated with a ticket
   -- number greater than the one provided.
-  , snapshotTake :: TxMeasure blk -> ([Validated (GenTx blk)], TxMeasureWithDiffTime blk)
+  , snapshotTake :: TxMeasure blk -> ([Validated (GenTx blk)], MempoolMeasure blk)
```
dnadales: Are we keeping `snapshotTake` for a caller I cannot see?

A grep across the repo finds three hits, and all three are this definition:

```
Mempool/API.hs:461          the field itself
Mempool/API.hs:474          the haddock on snapshotPartition referring back to it
Mempool/Impl/Common.hs:597  the implementation
```

`partitionMempool` was the last user, and both of its uses are now `snapshotPartition` (`NodeKernel/Forge.hs:751,795`).

`snapshotPartition` covers the old behaviour: `snapshotTake lim` is the first two components of `snapshotPartition lim zero`. So dropping the field costs nothing, and it takes an implementation with it.

If it stays as a public entry point for something outside this repo, could we say so in the haddock? Otherwise a reader has to grep to find out that it is unused.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -163,7 +165,15 @@ forgeShelleyBlock hotKey cbl ForgeBlockArgs{..} = do
...
       let forgedEb = forgeLeiosEb fbCurrentSlotNo ebTxs
           ebHash = hashLeiosEb forgedEb.body
-          ebSize = leiosEbBytesSize forgedEb.body
+          -- An EB over the codec's message limit never diffuses: peers drop
+          -- the connection and we would not learn why, since we skip
+          -- validating our own EB. The mempool measure plans within the limit
+          -- ('leiosEndorserBlockMeasure'), so a breach here means the measure
+          -- and the encoder have drifted apart.
+          ebSize =
+            assert
```
dnadales, line 174: Does this survive a downstream build?

`assert` goes away under `-O` unless `-fno-ignore-asserts` is on, and we gate that behind a flag that defaults off:

```
ouroboros-consensus.cabal:35-38   flag asserts, default: False
ouroboros-consensus.cabal:66-68   if flag(asserts) -> -fno-ignore-asserts
cabal.project:32                  import: cabal/asserts.cabal -> +asserts
```

So it fires in our own builds and in CI. `cardano-node`'s `cabal.project` does not mention the flag, so the node that runs the Leios chain forges without it.

That leaves the case the comment describes as the one we lose: an EB above the message limit goes out, peers close the connection, and we never learn why, because we skip validating our own EB.

Would a trace be a better fit here, given the failure is remote and silent? Something like:

```haskell
ebSize = encodeLeiosEbSize forgedEb.body

when (ebSize > maxMsgLeiosBlockBytesSize - msgLeiosBlockFramingSize) $
  traceWith fbLeiosTracer $ TraceLeiosEbOverWireLimit ebSize
```

Then the assert can stay on top for our own builds, and a real node still tells us.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -381,14 +387,54 @@ class
     TickedLedgerState blk mk ->
     TxMeasure blk
 
-  -- | What is the allowed capacity for the txs in a Leios Endorser Block?
+  -- | The (possibly multi-dimensional) size of a transaction in a Leios
+  -- endorser block: what it costs of 'ebCapacityTxMeasure'. 'NoEbMeasure' for
+  -- protocols without endorser blocks.
+  type TxEbMeasure blk
```
dnadales, line 392: Should this PR carry a `changelog.d` fragment under `Breaking`?

The public surface moves in several places at once:

- `TxLimits` gains `TxEbMeasure`, `txEbMeasure`, `ebCapacityTxMeasure` and `ebClosureCapacityTxMeasure`
- `ebCapacityTxMeasure` loses its `Maybe`
- `CanHardFork` gains `HardForkTxEbMeasure`, `hardForkInjTxEbMeasure` and `hardForkTxEbMeasure`
- `HardForkTxMeasure (CardanoEras c)` changes from `DijkstraMeasure` to `ConwayMeasure`
- `TxMeasureWithDiffTime`, `DijkstraMeasure`, `TxCount`, `oneTxCount`, `maxEbTxCount`, `maxEBClosureSize` and `leiosEbBytesSize` are gone
- `MempoolSnapshot` gains `snapshotPartition`
- `TraceForgedBlock` carries `MempoolMeasure` in place of `TxMeasureWithDiffTime`
- `TraceEventMempool` gains `TraceMempoolCapacityChanged`

The last two break `cardano-node` rather than this repo. Its tracer instances match `TraceEventMempool` exhaustively, so the new constructor fails to compile there, and the `TraceForgedBlock` payload change lands in the same place.

A fragment naming those two would save whoever bumps the dependency from finding out at the compiler.

ouroboros-consensus/src/ouroboros-consensus/LeiosTxCache/Reference.hs
```diff
@@ -333,7 +334,19 @@ insertBody ebh body nil snoc idx = case Map.lookup ebh (bodyState idx) of
             , txState = txState'
             , prunedSlot = prunedSlot idx
             }
-     in (idx', Just (mkLeiosTxCacheInsertBodySummary n tracked acquired validated (Map.size txState'), w))
+     in -- A 'Map' has no allocated capacity, so the worst case is the scale.
```
dnadales: Are we happy for `cacheLoad` to mean two things, depending on which cache is wired in?

The `Optimized` side now divides by `HT.capacity ht` (`Optimized.hs:115`), which is what makes 1.0 line up with the table actually filling. This side keeps `worstCaseCacheTxCount`. The two denominators are 4194304 and 9142784, so the same reading differs by 2.18x:

```
6 000 000 txs resident
  Optimized (node)      6000000 / 4194304  =  1.43   -- past full, insert has raised
  Reference (ThreadNet) 6000000 / 9142784  =  0.66   -- looks comfortable
```

The node runs `Optimized` (`Node.hs:552`) and ThreadNet runs `Reference`, so a threshold tuned in a simulation reads low against a real node, and the other way round.

The comment is right that a `Map` has no allocated capacity. What about emitting the denominator as well, so the reading is always relative to something the consumer can see? Something like:

```haskell
, ibsCacheCapacity = capacity   -- Nothing for the Map-backed cache
, ibsCacheLoad = fromIntegral cacheTxCount / fromIntegral capacity
```

Then a dashboard reading `cacheLoad` knows what it is a fraction of.

dnadales (commented):
Only one thing seems to need a fix before merging: 

- `NoEbMeasure`'s `Measure` instance makes `<=` constantly true, so a zero endorser-block capacity takes the whole mempool instead of none.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
ch1bo: It's 0 for dijkstra and non-0 for conway and earlier. I'll add a property

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/Capacity.hs
ch1bo: Not sure. I found the override being overridden always quite confusing, but did not want to change the computation of this also.

I'm not even sure if overrides should only work over the minimum needed (`oneUnit`), don't we have a case where we want deliberately smaller mempools?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
ch1bo: Let me just drop it then.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
ch1bo: Are we doing changelogs on the prototype? Or shall I make this change already on `master`? :thinking:

dnadales (approved):

ch1bo:
Merging this despite red CI because the errors are actually from the immutable/volatile split, which misbehaves in test setups. Let's fix that when we do https://github.com/IntersectMBO/ouroboros-consensus/pull/2298

ouroboros-consensus-cardano/src/ouroboros-consensus-cardano/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -145,7 +146,8 @@ type CardanoHardForkConstraints c =
 --     the calculation of later rewards. In this transition, we consume the
 --     'shelleyToAllegraAVVMsToDelete' as deletions in the ledger tables.
 instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
-  type HardForkTxMeasure (CardanoEras c) = DijkstraMeasure
+  type HardForkTxMeasure (CardanoEras c) = ConwayMeasure
```
jasagredo, line 149: I wonder if we would want to keep a DijkstraMeasure even if it is a newtype over ConwayMeasure. I was confused as to why the code below didn't seem to inject to Dijkstra.

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -866,6 +759,122 @@ txMeasureConway st tx@(ShelleyTx _txid tx') =
...
+-- An endorser block carries references to transactions, while the referenced
+-- transactions -- its closure -- must satisfy block-like limits. So this
+-- pairs the closure's 'ConwayMeasure' with the one dimension specific to
+-- endorser blocks: the size of the references themselves.
+--
+-- This is the Dijkstra-era 'TxEbMeasure'; the era's 'TxMeasure' is plain
+-- 'ConwayMeasure'.
+data DijkstraEbMeasure = DijkstraEbMeasure
+  { ebClosureMeasure :: !ConwayMeasure
+  , txReferencesSize :: !(IgnoringOverflow ByteSize32)
```
jasagredo, line 775: So if I understand this correctly, this is the same as the TxMeasure + the txReferencesSize. Is this much better than collapsing everything into TxMeasure? and assigning `0` to txReferencesSize when not using EBs?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Abstract/CanHardFork.hs
```diff
@@ -50,6 +56,11 @@ class
   -- in Haskell.)
   type HardForkTxMeasure xs
 
+  -- | A measure that can accurately represent the 'TxEbMeasure' of any era.
+  --
+  -- Trivial for eras without Leios endorser blocks; see 'TxEbMeasure'.
```
jasagredo, line 61: This comment is confusing. It doesn't depend on particular eras, it depends on `xs` which is a list of eras.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -381,14 +386,50 @@ class
...
+  txEbMeasure ::
+    proxy blk ->
+    TxMeasure blk ->
+    TxEbMeasure blk
+  default txEbMeasure ::
+    TxEbMeasure blk ~ TxMeasure blk =>
+    proxy blk ->
+    TxMeasure blk ->
+    TxEbMeasure blk
+  txEbMeasure _ = id
```
jasagredo, line 411: How does this even work? for a DijkstraEbMeasure, how can we populate the second field if the input args are just a Proxy and a DijkstraMeasure (or ConwayMeasure)?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -381,14 +386,50 @@ class
...
+  -- | What one full endorser-block fill drains from the mempool, in
+  -- 'TxMeasure' terms: an endorser block's closure is transactions. Only used
+  -- to size the mempool ('Ouroboros.Consensus.Mempool.computeMempoolCapacity').
+  --
+  -- Zero for blocks without endorser blocks.
+  ebClosureCapacityTxMeasure ::
+    LedgerConfig blk ->
+    TickedLedgerState blk mk ->
+    TxMeasure blk
+  ebClosureCapacityTxMeasure _ _ = zero
```
jasagredo, line 432: I want to advise against default implementations. Defaults are ok to be used when the implementation can fit _ALL_ the instances, and some instances might override it for a faster implementation. But they should not be used when _SOME_ of the instances benefit from it but it is wrong in some other instances.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -381,14 +386,50 @@ class
...
-  -- 'Nothing' for eras that don't support Leios.
+  -- Zero for blocks without endorser blocks: an endorser-block fill against a
+  -- zero capacity takes no transactions, so none are forged.
   ebCapacityTxMeasure ::
     LedgerConfig blk ->
     TickedLedgerState blk mk ->
-    Maybe (TxMeasure blk)
-  ebCapacityTxMeasure _ _ = Nothing
+    TxEbMeasure blk
+  ebCapacityTxMeasure _ _ = zero
```
jasagredo, line 421: Same

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -478,51 +487,65 @@ data MempoolSnapshot blk = MempoolSnapshot
   , snapshotPoint :: Point blk
   }
 
-data TxMeasureWithDiffTime blk = MkTxMeasureWithDiffTime !(TxMeasure blk) !DiffTimeMeasure
+-- | Everything the mempool measures per transaction: its block measure
+-- ('TxMeasure', which also bounds admission), its endorser-block measure
+-- ('TxEbMeasure') and how long it took to validate.
+data MempoolMeasure blk = MempoolMeasure
+  { mmTxMeasure :: !(TxMeasure blk)
+  , mmTxEbMeasure :: !(TxEbMeasure blk)
```
jasagredo, line 495: But if I understood it correctly, 
```
data DijkstraEbMeasure = DijkstraEbMeasure
  { ebClosureMeasure :: !ConwayMeasure
  , txReferencesSize :: !(IgnoringOverflow ByteSize32)
  }
``` 

So the `TxEbMeasure` already contains the `TxMeasure` no? i.e. `mmTxEbMeasure` will always contain `mmTxMeasure`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
dnadales: If I understand this correctly, the `ConwayMeasure` is enough, because the endorser-block cost of a transaction depends only on its byte size.

An endorser block stores a `(txHash, txSize)` pair per transaction, so we can compute the second field from `txMeasureByteSize`:

```haskell
txReferencesSize = IgnoringOverflow . encodeLeiosEbItemSize $ txMeasureByteSize conway
```

Or am I missing something?

PR #2283, opened 2026-09-10
db-synthesizer: store the endorser block it forges

dnadales:
`db-synthesizer` forged endorser blocks and dropped them.
`runForge` ignored the `ForgedLeiosEb` that `forgeBlock` returns, so `leios.db` stayed empty, `decideLeiosCertify` found no closure, and no block certified an EB.

That write was lost when this tool was rebased onto db00d5804, which moved storing out of `forgeBlock` so that the announcement goes out before the body write.

`runForge` now writes the EB point, its body and its tx closure, at the point where the node calls `onForgedLeiosEb`.
A forger that gains LeiosNotify or a LeiosTxCache must call `onForgedLeiosEb` instead, and the comment says so.

On a 40-block run: 29 EBs stored, 11 certifying blocks, and `check-synth-fixture` reports 40 ok and 0 FAIL. It reported 8 FAIL before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bladyjoker (approved):
LGTM

dnadales:
This test failure does not seem related to my PR, I'll investigate it later on.

PR #2284, opened 2026-09-10
Reuse 'NodeKernel.Forge.forge` in db-synthesizer

bladyjoker:
# Description

Proposing this refactor for the sake of easier maintenance, would love to hear from owners tho if that's ok and how to better test it.

DONE
 - [x] Refactored `Cardano.Tools.DBSynthesizer.Forging` to use `Ouroboros.Consensus.NodeKernel.Forge.forge` function alleviating logic duplication.
    - A Mempool instance is created (without a sync thread) where transactions are added via `addLocalTxs` on each slot.
    
NOTE
 - Did we consider an alternative to maintaining a separate tool for synthesizing DBs by implementing a more general use tool like [ThreadNet CLI in leios-prototype](https://github.com/IntersectMBO/ouroboros-consensus/blob/leios-prototype/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/ThreadNet/Run.hs).
   - Needless to say this runs in io-sim

### DIFF
--- a/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
+++ b/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
 
 module Cardano.Tools.DBSynthesizer.Forging
   ( GenTxs
@@ -13,66 +12,56 @@ import Cardano.Tools.DBSynthesizer.Types
   ( ForgeLimit (..)
   , ForgeResult (..)
   )
-import Control.Monad (when)
-import Control.Monad.Except (runExcept)
-import Control.Monad.IO.Class (liftIO)
-import qualified Control.Monad.Trans.Class as Trans
-import Control.Monad.Trans.Except (ExceptT (..), runExceptT, throwE)
-import Control.Tracer as Trace (nullTracer)
-import Data.Either (isRight)
-import Data.Maybe (fromJust, isJust)
-import Data.Proxy
+import Control.Monad (void)
+import Control.Tracer (nullTracer)
+import Data.Maybe (isJust)
 import Data.Word (Word64)
-import Ouroboros.Consensus.Block.Abstract as Block
-import Ouroboros.Consensus.Block.Forging as Block
-  ( BlockForging (..)
-  , ForgeBlockArgs (..)
-  , ShouldForge (..)
-  , checkShouldForge
+import Ouroboros.Consensus.Block
+  ( BlockForging
+  , EpochSize (..)
+  , SlotNo (..)
   )
 import Ouroboros.Consensus.Config
   ( TopLevelConfig
-  , configConsensus
   , configLedger
   )
-import Ouroboros.Consensus.Forecast (forecastFor)
-import Ouroboros.Consensus.HeaderValidation
-  ( BasicEnvelopeValidation (..)
-  , HeaderState (..)
-  )
 import Ouroboros.Consensus.Ledger.Abstract
-import Ouroboros.Consensus.Ledger.Extended
-import Ouroboros.Consensus.Ledger.SupportsMempool (GenTx)
-import Ouroboros.Consensus.Ledger.SupportsProtocol
-import Ouroboros.Consensus.Ledger.Tables.Utils (forgetLedgerTables)
-import Ouroboros.Consensus.Protocol.Abstract
-  ( ChainDepState
-  , tickChainDepState
+  ( ComputeLedgerEvents (..)
+  , DiffMK
+  , TickedLedgerState
+  , applyChainTick
   )
+import Ouroboros.Consensus.Ledger.Extended (ledgerState)
+import Ouroboros.Consensus.Ledger.SupportsMempool
+  ( GenTx
+  , Validated
+  , txForgetValidated
+  )
+import Ouroboros.Consensus.Mempool
+  ( Mempool (..)
+  , MempoolCapacityBytesOverride (..)
+  , addLocalTxs
+  , chainDBLedgerInterface
+  , openMempoolWithoutSyncThread
+  )
+import Ouroboros.Consensus.Node.Run (RunNode)
+import Ouroboros.Consensus.NodeKernel.Forge (forge)
 import Ouroboros.Consensus.Storage.ChainDB.API as ChainDB
-  ( AddBlockResult (..)
-  , ChainDB
-  , addBlockAsync
-  , blockProcessed
-  , getCurrentChain
-  , getPastLedger
+  ( ChainDB
+  , getTipPoint
   , withReadOnlyForkerAtPoint
   )
-import qualified Ouroboros.Consensus.Storage.ChainDB.API.Types.InvalidBlockPunishment as InvalidBlockPunishment
-  ( noPunishment
-  )
 import Ouroboros.Consensus.Storage.LedgerDB
+  ( ReadOnlyForker'
+  , roforkerGetLedgerState
+  )
 import Ouroboros.Consensus.Util.EarlyExit
-  ( withEarlyExit
+  ( exitEarly
+  , lift
+  , withEarlyExit
   )
 import Ouroboros.Consensus.Util.IOLike (atomically)
-import Ouroboros.Network.AnchoredFragment as AF
-  ( Anchor (..)
-  , AnchoredFragment
-  , AnchoredSeq (..)
-  , headPoint
-  )
-import Ouroboros.Network.Protocol.LocalStateQuery.Type
+import Ouroboros.Network.Protocol.LocalStateQuery.Type (Target (..))
 
 data ForgeState
   = ForgeState
@@ -92,12 +81,9 @@ type GenTxs blk =
   TickedLedgerState blk DiffMK ->
   IO [Validated (GenTx blk)]
 
--- DUPLICATE: runForge mirrors forging loop from ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
--- For an extensive commentary of the forging loop, see there.
-
 runForge ::
   forall blk.
-  LedgerSupportsProtocol blk =>
+  RunNode blk =>
   EpochSize ->
   SlotNo ->
   ForgeLimit ->
@@ -109,7 +95,14 @@ runForge ::
 runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
   putStrLn $ "--> epoch size: " ++ show epochSize_
   putStrLn $ "--> will process until: " ++ show opts
-  endState <- go initialForgeState{currentSlot = nextSlot}
+  mempool <-
+    openMempoolWithoutSyncThread
+      (chainDBLedgerInterface chainDB)
+      (configLedger cfg)
+      NoMempoolCapacityBytesOverride
+      Nothing
+      nullTracer
+  endState <- go mempool initialForgeState{currentSlot = nextSlot}
   putStrLn $
     "--> forged and adopted "
       ++ show (forged endState)
@@ -125,12 +118,12 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
     ForgeLimitBlock b -> (b ==) . forged
     ForgeLimitEpoch e -> (e ==) . currentEpoch
 
-  go :: ForgeState -> IO ForgeState
-  go forgeState
+  go :: Mempool IO blk -> ForgeState -> IO ForgeState
+  go mempool forgeState
     | forgingDone forgeState = pure forgeState
-    | otherwise =
-        go . nextForgeState forgeState . isRight
-          =<< runExceptT (goSlot $ currentSlot forgeState)
+    | otherwise = do
+        didForge <- goSlot mempool (currentSlot forgeState)
+        go mempool (nextForgeState forgeState didForge)
 
   nextForgeState :: ForgeState -> Bool -> ForgeState
   nextForgeState ForgeState{currentSlot, forged, currentEpoch, processed} didForge =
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -109,7 +95,14 @@ runForge ::
...
   putStrLn $ "--> epoch size: " ++ show epochSize_
   putStrLn $ "--> will process until: " ++ show opts
-  endState <- go initialForgeState{currentSlot = nextSlot}
+  mempool <-
+    openMempoolWithoutSyncThread
+      (chainDBLedgerInterface chainDB)
+      (configLedger cfg)
+      NoMempoolCapacityBytesOverride
+      Nothing
+      nullTracer
```
bladyjoker, line 104: I think it would be useful to trace this (optionally)

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -144,146 +137,29 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
...
+                    OmitLedgerEvents
+                    (configLedger cfg)
+                    currentSlot
+                    (ledgerState extLedgerState)
+            txs <- lift $ genTxs currentSlot frk tickedLedgerState
+            lift $ void $ addLocalTxs mempool (map txForgetValidated txs)
+    -- Forge a block for each credential; a block is adopted when forge returns Just ()
+    results <-
+      mapM
+        (\bf -> withEarlyExit $ forge nullTracer nullTracer cfg chainDB mempool bf currentSlot)
```
bladyjoker, line 163: I'd also say these traces are useful to have available.

changelog.d/20260910_124522_drazen.popovic_refactor_db_synthesizer.md
jasagredo, line 1: No changelogs on cardano-tools

jasagredo (approved):
Looks correct except for the changelog

changelog.d/20260910_124522_drazen.popovic_refactor_db_synthesizer.md
bladyjoker: Whoops! Removing!

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -144,146 +137,29 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
...
+          Right frk -> do
+            extLedgerState <- lift $ atomically $ roforkerGetLedgerState frk
+            let tickedLedgerState =
+                  applyChainTick
+                    OmitLedgerEvents
+                    (configLedger cfg)
+                    currentSlot
+                    (ledgerState extLedgerState)
+            txs <- lift $ genTxs currentSlot frk tickedLedgerState
+            lift $ void $ addLocalTxs mempool (map txForgetValidated txs)
```
dnadales, line 159: I wonder if we want to state a bound on what `genTxs` can return, next to the `GenTxs` type at line 79.

`addTx` blocks until the transaction fits in the mempool (`Mempool/API.hs:142-145`). Here the mempool holds two blocks' worth (`Capacity.hs:69`) and has no sync thread. `goSlot` runs sequentially in a single thread, so `forge` at line 161 cannot run while this line waits. On a non-leader slot `forge` exits early (`NodeKernel/Forge.hs:403-405`), so no block drains the mempool. The retry then waits on `check $ isMempoolSize is /= prevSize` (`Update.hs:199-202`), and no other thread in the tool can change that size. The `Nothing` at line 103 does not help, because the capacity wait sits outside the `timeout` wrapper (`Update.hs:240-242`).

Both in-tree callers return `pure []`, so nothing hits this today. `GenTxs` is exported and reaches `synthesize`, so the first out-of-tree generator is the case I worry about.

What about something like a note on the type:

```haskell
-- | An action to generate transactions for a given block.
--
-- PRECONDITION: the returned transactions fit in one block. 'runForge' adds
-- them to a mempool with a capacity of two blocks, and a full mempool blocks
-- the forge loop until a forged block drains it.
type GenTxs blk = ...
```

dnadales, line 159: Aren't we validating these transactions at a different slot from the one we generated them for?

We tick the state to `currentSlot` at line 152 and pass it to `genTxs`. So a generator builds transactions that are valid at `currentSlot`. The mempool validates at `succ (ledgerTipSlot)` instead (`Mempool/Impl/Common.hs:344-356` and `:382`). The tip is a block from an earlier iteration, so the two slots differ by the number of slots we skipped. With `activeSlotsCoeff` at 0.05, that is most slots.

The `void` then hides the outcome. `addLocalTxs` returns one `MempoolAddTxResult` per transaction (`Mempool/API.hs:369-375`), and we discard all of them. A transaction that the mempool refuses here never reaches `getSnapshotFor` at forge time (`NodeKernel/Forge.hs:467-473`), because that call only re-applies what is already in the mempool. So a generator can ask for N transactions, put zero of them in the block, and see no error.

What about something like:

```haskell
results <- lift $ addLocalTxs mempool (map txForgetValidated txs)
case [err | MempoolTxRejected _ err <- results] of
  [] -> pure ()
  errs ->
    lift . putStrLn $
      "--> slot " ++ show currentSlot ++ ": mempool rejected " ++ show (length errs) ++ " txs"
```

The tool already reports progress with `putStrLn` at lines 96 and 106, so that fits the existing style.

For the slot itself, I wonder if we want to hand `genTxs` the slot that the mempool uses. The other option is to say in the `GenTxs` haddock that `currentSlot` is the forge slot and not the validation slot.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -92,12 +81,9 @@ type GenTxs blk =
   TickedLedgerState blk DiffMK ->
   IO [Validated (GenTx blk)]
```
dnadales, line 82: Do we still need the `Validated` wrapper here?

Line 159 is the only consumer, and it calls `txForgetValidated` right away, so the mempool validates each transaction from scratch. Before this PR the wrapper carried weight: the transactions went straight into `fbTxs`, which is typed `[Validated (GenTx blk)]`, and nothing revalidated them.

The wrapper also costs the generator real work. To build one for Shelley, `mkShelleyValidatedTx` needs an `SL.Validated (Tx TopTx era)` (`shelley/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs:208-213`), so the generator runs ledger validation itself. We ask every generator to validate, then we discard the proof and validate again.

What about:

```haskell
type GenTxs blk =
  SlotNo ->
  ReadOnlyForker' IO blk ->
  TickedLedgerState blk DiffMK ->
  IO [GenTx blk]
```

That drops the `txForgetValidated` at line 159 as well.

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -144,146 +137,29 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
...
+            let tickedLedgerState =
+                  applyChainTick
+                    OmitLedgerEvents
+                    (configLedger cfg)
+                    currentSlot
+                    (ledgerState extLedgerState)
+            txs <- lift $ genTxs currentSlot frk tickedLedgerState
+            lift $ void $ addLocalTxs mempool (map txForgetValidated txs)
+    -- Forge a block for each credential; a block is adopted when forge returns Just ()
+    results <-
```
dnadales, line 161: I wonder if we want to note here that a slot can now produce more than one block.

The deleted code took the first leader only. Now every leading credential forges. Credential 1's `forge` waits for adoption before it returns (`NodeKernel/Forge.hs:281-283`), so credential 2 reads a tip that already sits in `currentSlot`. `mkCurrentBlockContext` then takes the `EQ` branch, whose own comment says it forges "an alternative to `hdr`: same block no and same predecessor" (`NodeKernel/Forge.hs:250-258`). Chain selection breaks the equal-block-number tie on VRF (`Protocol/Praos/Common.hs:146-148`), so one of the two blocks is orphaned.

This matches the node, which runs one forging thread per credential (`NodeKernel.hs:420`), so I read it as deliberate. The counting stays correct too: both blocks carry the same block number, so the selected chain grows by at most one, and `forged` still equals the chain length.

What changes is the artefact. It needs two credentials leading in the same slot, which the test fixture allows (`bulk-creds-k2.json` has 2 entries). The VolatileDB then holds orphan blocks that a single-credential run never wrote. `db-analyser` and `db-truncater` read that DB, so a note will save the next reader the derivation:

```haskell
-- Forge for each credential, as the node does. If two credentials lead in the
-- same slot, both blocks get the same block number, so one of them is orphaned
-- and stays in the VolatileDB. 'forged' counts adopted blocks, so it tracks the
-- chain length, not the blocks on disk.
```

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -144,146 +137,29 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
...
-          else BlockContext (blockNo hdr) $ castPoint $ AF.headPoint c'
+  goSlot :: Mempool IO blk -> SlotNo -> IO Bool
+  goSlot mempool currentSlot = do
+    -- Sync mempool with the latest block adoption at 'tip'
+    _ <- testSyncWithLedger mempool
+    -- Populate mempool (synced against 'tip') with transactions for this slot
+    tip <- atomically $ ChainDB.getTipPoint chainDB
+    void $
+      withEarlyExit $
+        ChainDB.withReadOnlyForkerAtPoint chainDB (SpecificPoint tip) $ \case
```
dnadales, line 148: Do we want to populate the mempool only when some credential leads?

The deleted code ran `checkShouldForge` first and exited on `NoLeader` before it ticked the state or opened a forker. Now we open a forker, read the state, tick it and call `genTxs` on every slot. The leadership check sits inside `forge` (`NodeKernel/Forge.hs:103`, with the exit at `:403-405`), which we reach at line 161.

For the two in-tree generators the cost is small. Both ignore the ticked state, so line 152 stays an unevaluated thunk. What we pay per slot is one forker open and close here, plus one more inside each `forge` call.

For a generator that uses its third argument the picture changes. The full tick then runs on every slot, and at `activeSlotsCoeff` 0.05 we forge in about one slot in twenty. Right after an epoch boundary we redo the epoch transition once per slot, until the epoch's first block lands. In a leader slot we tick twice, because line 152 and `Forge.hs:112` tick at different anchors: `getTipPoint` here and `bcPrevPoint` there.

A leadership check before line 145 fixes it, but that copies back part of what this PR removes, so I am not sure it is worth it. A comment naming the trade-off will do for now:

```haskell
-- We populate the mempool on every slot, including slots with no leader,
-- because 'forge' owns the leadership check. That keeps the duplication out,
-- at the cost of one forker and one tick per empty slot.
```

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -109,7 +95,14 @@ runForge ::
 runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
   putStrLn $ "--> epoch size: " ++ show epochSize_
   putStrLn $ "--> will process until: " ++ show opts
-  endState <- go initialForgeState{currentSlot = nextSlot}
+  mempool <-
```
dnadales, line 98: Do we want to record why we drive the mempool by hand here?

Both entry points we use carry a test-only label. `openMempoolWithoutSyncThread` says "Intended for testing purposes" (`Mempool/Init.hs:84-87`), and `testSyncWithLedger` at line 143 says "ONLY FOR TESTS" (`Mempool/API.hs:231-232`).

I read the choice as correct, not as a shortcut. `openMempool` forks a watcher that re-syncs whenever the tip moves (`Mempool/Init.hs:47-50`). The tip moves inside `forge`, so that watcher would re-sync while the next slot adds transactions and forges. `implSyncWithLedger` drops transactions and swaps the internal state under lock (`Mempool/Update.hs:661-675`), so which transactions reach a block becomes schedule-dependent. The tool exists to produce a reproducible DB, so we do not want that.

`openMempool` will not remove the test-only name either. The `Mempool` record has seven fields (`Mempool/Init.hs:113-121`), and `testSyncWithLedger` is the only sync among them.

A short comment here will stop the next reader from "fixing" it:

```haskell
-- We drive the mempool by hand: 'openMempool' forks a watcher that re-syncs on
-- every tip change, which makes block contents schedule-dependent. Synthesis
-- must stay reproducible, so we sync explicitly in 'goSlot' instead.
```

Do we want an issue against the mempool for a supported hand-driven interface, or is the test-only name acceptable for a tool?

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
```diff
@@ -144,146 +137,29 @@ runForge epochSize_ nextSlot opts chainDB blockForging cfg genTxs = do
...
+  goSlot :: Mempool IO blk -> SlotNo -> IO Bool
+  goSlot mempool currentSlot = do
+    -- Sync mempool with the latest block adoption at 'tip'
+    _ <- testSyncWithLedger mempool
+    -- Populate mempool (synced against 'tip') with transactions for this slot
+    tip <- atomically $ ChainDB.getTipPoint chainDB
+    void $
+      withEarlyExit $
+        ChainDB.withReadOnlyForkerAtPoint chainDB (SpecificPoint tip) $ \case
+          Left{} -> exitEarly
```
dnadales, line 149: Aren't we losing the diagnostic here?

The old code had `error "Impossible: we are forging on top of a block that the ChainDB cannot create forkers on!"` on this branch. Now we `exitEarly`, and the `void` at line 143 drops the `Maybe ()`. Both tracers are null, so a forker failure prints nothing and changes no exit code. The loop then reaches line 161 and forges from whatever the mempool holds, which after a skip is the previous slot's leftovers.

I agree the branch is close to unreachable. `tip` comes from `getTipPoint`, the loop is single-threaded, and nothing switches forks between lines 145 and 148. That is what the old message said as well.

What about keeping the loud version:

```haskell
          Left{} ->
            error $
              "Impossible: no forker at the tip " ++ show tip ++ " in slot " ++ show currentSlot
```

If we prefer not to `error` in a tool, a `putStrLn` on this branch keeps the signal without the crash.

dnadales (approved):

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
bladyjoker: Great find!!! Absolutely makes sense! Still not sure how to handle this given that genTxs is bound to zero tx generator.

Apparently it's only about generating empty blocks with correct headers

bladyjoker: The slot and tickedLedgerState is generally just about telling genTxs what the tip is.

Not only does the Mempool validate against a different slot (+1) but also the end ledger state is different too.

This is a common problem with transaction generators, they operate half blind and have to guess what the mempool end ledger state (and slot) are.

bladyjoker: Definitely useful to report errors! Good find!

ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs
bladyjoker: I think it would be good to stdout tracers, but I need a user to determine that.

PR #2285, opened 2026-09-10
test: do not fail prop_leios_invalid_eb on a vacuous run

dnadales:
`prop_leios_invalid_eb` fails on seed `Seed (-1668348207015116830)`, on its own non-vacuity guard `adversaryWasEffective`.

In that run the adversary leads exactly once, in slot 0.
A node offers an EB only to the peers connected when it stores the EB, and it never offers one it already holds.
`subscribeEbNotifications` hands out a `dupTChan`, which keeps no backlog.
So an EB forged before the ThreadNet edges are up reaches nobody.
The announcement still arrives over ChainSync, but an announcement names no serving peer, and the fetch logic only requests what a peer offers.

The node needs no fix for this.
In production a leader forges long after boot, and an EB that old is no longer votable.
A late peer recovers the closures that matter through the CertRB roll-forward path.

This PR drops the per-run requirement and checks the rate with `checkCoverage` instead.
The invariants the property exists for still fail per run: a poisoned EB is never certified, and never validated.

Open, marked with a TODO in the code:

- The threshold of 50 is a placeholder. A 120-run sweep is measuring the rate of runs in which an honest node acquires a poisoned EB's closure. The threshold will come from that number.
- `checkCoverage` runs more tests when the result is undecided. Each run costs about 40s, so I will confirm it still decides inside the configured run count.

### DIFF
--- a/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
+++ b/ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/Leios.hs
@@ -150,9 +150,11 @@ import Test.Consensus.Cardano.ProtocolInfo (Era (Dijkstra), hardForkInto)
 import Test.QuickCheck
   ( Property
   , Testable
+  , checkCoverage
   , choose
   , conjoin
   , counterexample
+  , cover
   , discard
   , forAll
   , ioProperty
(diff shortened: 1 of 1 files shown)


dnadales:
Superseded by https://github.com/IntersectMBO/ouroboros-consensus/pull/2289

PR #2286, opened 2026-09-11
Peras: support Object Diffusion termination and GSM caught-up detection

ErinvanderVeen:
Depends on [ouroboros-network#5431](https://github.com/IntersectMBO/ouroboros-network/pull/5431).

This PR implements the consensus side of those changes in three commits.

## 1. Support graceful Object Diffusion termination

Handle `MsgServerIdle`, introduced by the first commit of ouroboros-network#5431.

- The wait for new IDs is bounded. If no IDs arrive before the timeout, the server sends `MsgServerIdle` and returns agency to the client, allowing the client to terminate the mini-protocol gracefully.
- The client handles and traces `MsgServerIdle` and regains agency after receiving it.
- The Object Diffusion smoke tests cover timeout behaviour, recovery after an idle response, and graceful termination.

This implements the consensus side of [cardano-peras#187](https://github.com/tweag/cardano-peras/issues/187).

## 2. Make Object Diffusion server idleness explicit

Handle `MsgAwaitReply`, introduced by the second commit of ouroboros-network#5431.

- When a blocking request has no immediately available object IDs, the server sends `MsgAwaitReply` before waiting. This makes the server's idleness explicit to the client.
- The client handles and traces `MsgAwaitReply` separately from replies containing object IDs.
- The request-kind state index distinguishes the replies that are legal before and after `MsgAwaitReply`.
- The Object Diffusion smoke tests cover immediate replies, delivery after `MsgAwaitReply`, and acknowledgement ordering.

This implements the explicit server-idling signal required by [cardano-peras#144](https://github.com/tweag/cardano-peras/issues/144).

## 3. Use certificate-diffusion idling in GSM caught-up detection

Use the explicit server-idling signal as one input when the Genesis State Machine decides whether the node is caught up.

- Track certificate-diffusion client registration and idling alongside ChainSync idling for each peer.
- Retain the negotiated `PerasSupport` in the ChainSync state. If Peras was negotiated, the peer is idle only when both ChainSync and its registered certificate-diffusion client are idling. If Peras was not negotiated, ChainSync idling is sufficient for that peer.
- Start certificate clients as non-idle, update their state from the Object Diffusion callbacks, and remove their handles when they terminate.
- Keep vote diffusion outside GSM caught-up detection because votes do not directly affect chain selection.
- Add tests for registration order, client matching, idling transitions, cleanup, protocol callbacks, and an end-to-end GSM transition to `CaughtUp`.

## Temporary dependency

The first two commits advance `cabal.project` through the corresponding commits of ouroboros-network#5431. The final commit pins its current head, including `network-mux`. The pin must be replaced with an upstream `main` commit or release before this PR can merge.


changelog.d/20260911_102316_erin.vanderveen_object_diffusion_termination.md
```diff
@@ -0,0 +1,9 @@
+### Breaking
+
+- Add an idle timeout argument to `objectDiffusionOutbound` and replace its
```
tbagrel1: The three changelog entries should probably be merged :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Outbound.hs
```diff
@@ -145,43 +155,89 @@ objectDiffusionOutbound tracer maxFifoLength ObjectPoolReader{..} _version =
...
+                                lastTicketNo
+                                (fromIntegral numIdsToReq)
+                            case maybeNewObjectsAction of
+                              Nothing -> retry
+                              Just newObjectsAction -> pure (Just newObjectsAction)
+                    case result of
+                      Nothing -> pure Nothing
+                      Just getNewObjects -> do
+                        content <- getNewObjects
+                        if null content
```
tbagrel1: Now that we have `MsgServerIdle`, we could probably use it instead of re-running `getNewContentOrIdle` when `null content`. 

(N.B. : the ObjectPoolWriter interface will likely change soon to account for immutable Cert DB, so this won't matter much)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/State.hs
```diff
@@ -0,0 +1,129 @@
...
+  , newTVarIO
+  , readTVar
+  )
+
+-- | An ObjectDiffusion inbound client state that's used by other components.
+--
+-- NOTE: 'blk' is not needed for now, but we keep it for future use.
+data ObjectDiffusionInboundState blk = ObjectDiffusionInboundState
+  { odIdling :: !Bool
+  -- ^ Whether the client is currently idling
```
tbagrel1: "idling" terminology is a bit confusing, since we want to detect the `StObjectIds MustReply` state, not `StIdle` (ditto for all occurences of "idling" in this file and other files)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
```diff
@@ -148,13 +161,15 @@ objectDiffusionInbound ::
   ObjectPoolWriter objectId object m ->
   NodeToNodeVersion ->
   ControlMessageSTM m ->
+  ObjectDiffusionInboundStateView m ->
```
tbagrel1, line 160: Is the "StateView" terminology standard? Because I would expect a view to be mostly read-only, while it seems this param is used to update the state.

Feel free to ignore.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
```diff
@@ -148,13 +161,15 @@ objectDiffusionInbound ::
...
   ObjectDiffusionInboundPipelined objectId object m ()
 objectDiffusionInbound
   tracer
   (maxFifoLength, maxNumIdsToReq, maxNumObjectsToReq)
   ObjectPoolWriter{..}
   _version
-  controlMessageSTM =
+  controlMessageSTM
+  state =
```
tbagrel1: Need a better name

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
```diff
@@ -441,8 +456,28 @@ objectDiffusionInbound
             $ SendMsgRequestObjectIdsBlocking
               (numToAckOnNextReq st)
               numIdsToRequest
+              ( do
+                  traceWith tracer TraceObjectDiffusionInboundAwaitReply
+                  Idling.idlingStart (odisvIdling state)
+                  traceWith tracer TraceObjectDiffusionInboundStartedIdling
```
tbagrel1, line 457: I think this tracing action is redundant.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
```diff
@@ -441,8 +456,28 @@ objectDiffusionInbound
...
+                      checkState st' & goCollect Zero (CollectObjectIds numIdsToRequest (NonEmpty.toList neCollectedIds))
+              )
+              ( WithEffect $ do
+                  traceWith tracer TraceObjectDiffusionInboundServerIdle
+                  pure $!
+                    checkState
+                      st
+                        { numToAckOnNextReq = 0
+                        , numIdsInFlight = 0
+                        }
```
tbagrel1: We should probably document why we update the state to these values

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Outbound.hs
```diff
@@ -145,43 +155,89 @@ objectDiffusionOutbound tracer maxFifoLength ObjectPoolReader{..} _version =
...
+                  st'' = updateStNewObjects st' sortedNewContent
+
+              traceWith tracer (TraceObjectDiffusionOutboundSendMsgReplyObjectIds newIds)
+
+              pure $
+                SendMsgReplyObjectIds
+                  (BlockingReply (NonEmpty.fromList $ newIds))
+                  (makeBundle st'')
+
+            -- After 'MsgAwaitReply' has been sent, wait for either new objects
```
tbagrel1, line 182: Control flow is not completely obvious to follow in this section, but I think it works exactly as we imagined, well done!

We may want to improve the documentation in the future if this part doesn't need any further updates/changes.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/Util/Idling.hs
```diff
@@ -0,0 +1,29 @@
...
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Ouroboros.Consensus.MiniProtocol.Util.Idling (Idling (..), noIdling) where
+
+import GHC.Generics (Generic)
+import Ouroboros.Consensus.Util.IOLike (IOLike, NoThunks)
+
+-- | Interface for the ChainSync client to manipulate the idling flag in
```
tbagrel1: Same concerns about "idling" being a bit overloaded in meaning for ObjectDiffusion/Genesis

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -179,6 +186,12 @@ import System.Random (StdGen, splitGen)
   Handlers
 -------------------------------------------------------------------------------}
 
+-- | Prototype interval for returning agency while an Object Diffusion server
```
tbagrel1: We need a `TODO` comment with an issue link to remember we need to set the production value afterwards.

In any case, I would use a value closer to the max allowed for graceful peer demotion, e.g. 15 or 20s IIRC.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -281,26 +294,31 @@ initNodeKernel
...
-                                  (configBlock cfg)
-                                  (forgetFingerprint weights)
-                                  headers
-                                  (csCandidate state)
-                              )
-                , GSM.peerIsIdle = csIdling
+                    weights <- forgetFingerprint <$> ChainDB.getPerasWeightSnapshot chainDB
+                    pure $ \(headers, _lst) peerState -> do
+                      case csCandidate <$> maybeChainSyncState peerState of
+                        Just candidate
```
tbagrel1, line 300: Not sure I follow why the logic now has extra cases 🤔

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/Genesis/Tests/LoE/CaughtUp.hs
```diff
@@ -284,7 +289,7 @@ mkGsmEntryPoints varChainSyncHandles chainDB writeGsmState =
       { GSM.getCandidateOverSelection = pure candidateOverSelection
       , GSM.peerIsIdle = csIdling
       , GSM.equivalent = (==) `on` AF.headPoint
-      , GSM.getChainSyncStates = fmap cschState <$> cschcMap varChainSyncHandles
+      , GSM.getPeerStates = traverse readTVar =<< fmap cschState <$> cschcMap varChainSyncHandles
```
tbagrel1, line 292: Here it seems we are still dealing with ChainSyncStates, not PeerStates. Probably a byproduct of eager renaming ;)

tbagrel1 (commented):
Wonderful work! (I have reviewed everything carefully except the Smoke tests and GSM PeerState tests)

I've suggested a couple of doc/naming changes.

Apart from that, I have just one design question. Naively, I would have tried to make ObjectDiffusion inbound state to be observable, and then GSM could just inspect the state and test if it is `StObjectIds MustReply` (the "caught-up" state). Do you know if this would be feasible and/or what are the pros/cons compared to the current approach using a dedicated bool TVar ?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
prednaz: yes, this imitates [`ChainSyncStateView`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L319). in general, there is a lot of what i perceive as unnecessary complexity in [`Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.State`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/State.hs#L36-L80) just to follow the conventions set by [`Ouroboros.Consensus.MiniProtocol.ChainSync.Client.State`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/State.hs#L56-L136).

prednaz:
> Naively, I would have tried to make ObjectDiffusion inbound state to be observable, and then GSM could just inspect the state and test if it is `StObjectIds MustReply` (the "caught-up" state). Do you know if this would be feasible and/or what are the pros/cons compared to the current approach using a dedicated bool TVar ?

we had the following 3 reasons for the current approach using a dedicated bool `TVar`.
1. the current approach follows the convention [set by `ChainSync`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L1405), which [uses the exact same `Idling` handle](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L794).
2. the state does not uniquely determine if we are caught up or not. in `StIdle` we are caught up if we transitioned there via `MsgServerIdle` but not caught up if we transitioned there via `MsgReplyObjIds` (see [state diagram](https://tweag.github.io/cardano-website/static/b0226818aa235684fc7370cd311c366b/26d93/2026-08-28_3.png)). we would need to split `StIdle` into 2 states before we can use the other approach.
3. @ErinvanderVeen, @tbagrel1, and i committed to the approach using a dedicated bool `TVar` in a meeting 2026-08-27 for the above reasons.

tbagrel1:
> @ErinvanderVeen, @tbagrel1, and i committed to the approach using a dedicated bool TVar in a meeting 2026-08-27 for the above reasons.

In this meeting I thought we discussed a distinct state OR a TVar, not both, hence my surprised reaction. But I understand why it makes sense, thanks for the detailed explanation!

> the state does not uniquely determine if we are caught up or not. in StIdle we are caught up if we transitioned there via MsgServerIdle but not caught up if we transitioned there via MsgReplyObjIds (see [state diagram](https://tweag.github.io/cardano-website/static/b0226818aa235684fc7370cd311c366b/26d93/2026-08-28_3.png)). we would need to split StIdle into 2 states before we can use the other approach.

Good catch, I hadn't thought about that. If it's not done already, would you mind adding this as documentation in the code? 

> the current approach follows the convention [set by ChainSync](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L1405), which [uses the exact same Idling handle](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L794).

Nice, happy to know!

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound/State.hs
prednaz: i suspect that staying consistent with `ChainSync` might be worth accepting this confusing terminology. indeed, this terminology is in very close correspondence with `ChainSync`. `Ouroboros.Consensus.MiniProtocol.ObjectDiffusion.Inbound.State.ObjectDiffusionInboundState {odIdling}` corresponds to [`Ouroboros.Consensus.MiniProtocol.ChainSync.Client.State.ChainSyncState {csIdling}`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client/State.hs#L61). And they are both switched to `True` on the exact same message `MsgAwaitReply` (see [for `ChainSync`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L1397-L1405), [for `ObjectDiffision`](https://github.com/tweag/ouroboros-consensus/blob/eb5f768fac5ce79d26d7e095e28d9dec40e2507c/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs#L456-L461)), which in both mini protocols transitions from a `CanAwait` to a `MustReply` state.

<table style="table-layout: fixed;"><tr><td></td><td></td></tr><tr><td><code>ObjectDiffusion</code></td><td><code>ChainSync</code> (<a href="https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf">Ouroboros Network Specification</a>)</td></tr></table>

tbagrel1: Ok, let's keep it for consistency, but we should document why we are using this terminology, and in the comments/docstrings, explain that Idling = caught-up in this particular case, and clearly distinguish from the StIdle state :)

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ObjectDiffusion/Inbound.hs
tbagrel1: This is unfortunate, but I can't say I'm surprised ahah.

When we will ask for a review from maintainers, we should probably ask what they prefer (following the old conventions VS cleaner and simpler code)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
ErinvanderVeen: There is really one one extra case, when the combined getPeerStates map temporarliy contains a cert-only peer. The other cases are the original boolean expanded for clarity. Alternatively:

```haskell
                    pure $ \(headers, _lst) peerState ->
                      case maybeChainSyncState peerState of
                        Nothing ->
                          -- The ChainSync and certificate-diffusion clients
                          -- register independently, so this peer may temporarily
                          -- have only a certificate-diffusion state and therefore
                          -- no candidate.
                          GSM.WhetherCandidateIsBetter False
                        Just csState ->
                          let candidate = csCandidate csState
                           in case AF.intersectionPoint headers candidate of
                                Nothing ->
                                  GSM.CandidateDoesNotIntersect
                                Just{} ->
                                  GSM.WhetherCandidateIsBetter $
                                    shouldSwitch $
                                      preferAnchoredCandidate
                                        (configBlock cfg)
                                        weights
                                        headers
                                        candidate
```

ouroboros-consensus-diffusion/src/unstable-diffusion-testlib/Test/ThreadNet/Network.hs
```diff
@@ -1426,7 +1427,8 @@ directedEdgeInner
                       "ChainSync"
                       (\_s _ -> RestartChainSyncTerminated)
                       (\_s () -> RestartChainSyncTerminated)
-                      NTN.aChainSyncClient
+                      -- TODO: ThreadNet currently doesn't start up certificate object diffusion, so we pass PerasUnsupported to avoid waiting indefinitely.
+                      (\a v e c -> NTN.aChainSyncClient a PerasUnsupported v e c)
```
agustinmista, line 1431: We should create or link an existing issue in `tweag/cardano-peras` here.

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM/PeerState.hs
```diff
@@ -0,0 +1,180 @@
...
+                allIdle cs cert >>= (@?= True)
+                E.throwIO (userError "certificate client failed")
+            ) ::
+            IO (Either IOError ())
+        case result of
+          Left _ -> pure ()
+          Right () -> assertFailure "expected the client exception"
+        remaining <- atomically $ odihcMap cert
+        assertBool "failed client retained a handle" (Map.null remaining)
+        allIdle cs cert >>= (@?= False)
```
agustinmista, line 60: Might want to break this into a separate function.

agustinmista (commented):
Great job! I only have some bikesheddy comments.

Since this is partially based on our previous attempt at solving this (https://github.com/IntersectMBO/ouroboros-consensus/pull/1783) without having proper OD support, it could be good to have a (very) short summary of what's different here. 

In terms of sizing and complexity, this is rather large and non-trivial for the Consensus team to review in one go, so I would suggest splitting it into a PR stack with, e.g. (YMMV):

1. Break idleness into a separate module
2. Idleness tracking on the object diffusion side (+tests)
3. ChainSync state => peer state generalization (+tests)

ouroboros-consensus-diffusion/test/consensus-test/Test/Consensus/GSM/PeerState.hs
prednaz: factoring out a function can serve the purpose of avoiding repetition. and it can serve the purpose of revealing, demarcating, and labeling a self contained section of code. however, it comes at the cost of indirection.

but in this case it would incur this cost of indirection without serving any of the purposes. this section of code is not repeated. it is already very visible as a self contained section of code by being a list item and by its indentation. it is demarcated by list syntax and indentation. and it is labelled by `testCase "client failure removes its certificate handle"`.

did i forget any purpose that factoring out a function can serve?

agustinmista: Hey, I like the way you think.

> did i forget any purpose that factoring out a function can serve?

I would argue: consistency. 

Unlike the rest, the last element in this `testGroup` conflates _declaration_ and _implementation_ in one place, without an immediately obvious reason to deviate from the pattern established above (and in the rest of the codebase). 

When I read this the first time I thought: why are the other tests factored out but not this one? What makes it different? And why do I need to see the "guts" of the test here? Is there some hard-earn wisdom I'm not catching? 

FTR: I would be fine with this either way, but the consensus folks might want to err more strongly on the side of consistency.

prednaz: you are right. this inconsistency is a bit weird. i have [factored out `failureRemovesHandle`](https://github.com/IntersectMBO/ouroboros-consensus/pull/2286/commits/03fa380a78cefd8f53bf3c057f5b28e5b1036ed1).

PR #2293, opened 2026-09-14
Immutable (aka historical) certificate database

ninioArtillero:
# WIP

## In this PR

1. API definition (record) with two fundamental methods:
   1. `addCert :: ValidatedPerasCert blk -> m AddPerasImmutableCertResult`
   2. `getPointCerts :: Point blk -> m [ValidatedPerasCert blk]`. Using `Point` as argument to match convention at ChainSync.
3. Method implementation.
4. Serialization definition/format. No versioning, using CBOR.
5. DB arguments (record): included in ChainDB arguments.
   Initializing the database in a similar fashion to the ImmutableDB: mkImmFS.
6. Tracer and error types.

## Design

- Write to single file per certificate.
- Current: in-memory index for db methos implementation. Alternative: Direct read/write from/to disk?

## Questions

- Can we afford an in-memory index of the DB?
  _Probably no, syncing nodes are expected to be the exception, so it seems waistful. Still looking into how to do the read/write directly on disk;  a clever file layout maybe?_.
- Versioning for the certificate serialqization format?
  _Currently using CBOR, but not sure if we should do otherwise_
- Should we redirect the serialization format through Encode/Decode-Disk?
  or just use CBOR directly?
  _We have a separate commit implementing the former, for comparison._


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/API.hs
```diff
@@ -0,0 +1,26 @@
...
+import GHC.Generics (Generic)
+import NoThunks.Class
+import Ouroboros.Consensus.Block
+
+-- | API for the 'PerasImmutableCertDB', which stores immutable (aka historical)
+-- Peras certificates, ie those relevant for syncing nodes only.
+data PerasImmutableCertDB m blk = PerasImmutableCertDB
+  { addCert :: ValidatedPerasCert blk -> m AddPerasImmutableCertResult
+  -- ^ Add a certificate
+  , getPointCerts :: Point blk -> m [ValidatedPerasCert blk]
```
agustinmista: Depending on how often this method ends up being called, we might end up in a O(n^2) situation (because the implementation seems to make a linear pass over the index).

Do we know who will call this function and how often?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
```diff
@@ -0,0 +1,247 @@
...
+
+{-------------------------------------------------------------------------------
+  Database state
+-------------------------------------------------------------------------------}
+
+data PerasImmutableCertDbEnv m blk = PerasImmutableCertDbEnv
+  { picdbHasFS :: !(SomeHasFS m)
+  , picdbCodecConfig :: !(CodecConfig blk)
+  , picdbTracer :: !(Tracer m (TraceEvent blk))
+  , picdbState :: !(StrictTVar m (Map PerasRoundNo (ValidatedPerasCert blk)))
```
agustinmista: I'm wondering if we gain anything from indexing certs over `PerasRoundNo`s.

On one hand, yes, there should be at most one Peras cert per round.

On the other hand, I have the feeling that the client code accessing this data structure won't care much about which round number a certificate was forged in, and instead could benefit more from having fast access to all certs boosting a block in a given range (e.g.. all the certs boosting a block in the slot bounds of a chain fragment).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
```diff
@@ -0,0 +1,247 @@
...
+  CodecConfig blk ->
+  m [ValidatedPerasCert blk]
+readAllCerts hasFS ccfg = do
+  names <- Set.toList <$> listDirectory hasFS (mkFsPath [])
+  forM names $ \name -> do
+    let path = mkFsPath [name]
+    bytes <- withFile hasFS path ReadMode (hGetAll hasFS)
+    case CBOR.deserialiseFromBytes (decodeCert ccfg) bytes of
+      Right (_leftover, cert) -> pure cert
+      Left err -> throwIO $ CorruptPerasImmutableCertFile path (show err)
```
agustinmista: Keeping all certificates in memory would be quite problematic since this is an append-only database. Also, these wouldn't be needed unless we are serving them to a syncing client. 

A quick back-of-the-napking calculation:

* Size of a certificate ~ 20kb
* Expected number of certs during normal operation ~ 1 every 90s
* 20kb * 3600/90 * 24 * 365 ~ up to 7gb per year (not counting any potential in-memory representation overhead)

That much data would probably be fine to keep on disk, but definitely not in memory.

Have you considered reading them lazily, either on demand or slightly ahead of time? This while perhaps keeping an in-memory index* of what's on disk.

[*] we could also avoid having a linearly growing in-memory index by using run-length encoding, since that would pair well with the dynamics of the voting and cooldown paths of the protocol.

agustinmista (commented):
Great job! I have some initial comments about the design 😸

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
ninioArtillero: So, IIUC, `Map (SlotNo, SlotNo) [ValidatedPerasCert]` could be a better fit to model the db?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/API.hs
ninioArtillero: AFAIU, the only consumer of this method would be ObjectDiffusion. I don't know if we expect requests to be range-wise or point-wise (other comment suggest the interesting case from the point of view of the client is the former).

tbagrel1: I think ObjectDiffusion will query certs by ascending `PerasRoundNo`. So a `getCertsAfter :: PerasRoundNo -> Word64 -> m [ValidatedPerasCert blk]` would probably make more sense.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
tbagrel1: @agustinmista
> On the other hand, I have the feeling that the client code accessing this data structure won't care much about which round number a certificate was forged in, and instead could benefit more from having fast access to all certs boosting a block in a given range (e.g.. all the certs boosting a block in the slot bounds of a chain fragment).

I think I disagree with you here: ImmutableCertDB will never impact chainsel AFAIR. So I think we will only query the DB by ascending `PerasRoundNo` to serve in ObjectDiffusion. 

> having fast access to all certs boosting a block in a given range

Do you see what component could use such query?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
```diff
@@ -0,0 +1,247 @@
...
+  Complete PerasImmutableCertDbArgs m blk ->
+  m (PerasImmutableCertDB m blk)
+createDB
+  PerasImmutableCertDbArgs
+    { picdbaCodecConfig
+    , picdbaHasFS = someHasFS@(SomeHasFS hasFS)
+    , picdbaTracer
+    } = do
+    createDirectoryIfMissing hasFS True (mkFsPath [])
+    certs <- readAllCerts hasFS picdbaCodecConfig
```
tbagrel1: Not sure we want to read all certs on startup?

We may instead want to load an index of all certs we have (so we know if we already have a cert when one is added with `addCert` action), but defer loading the cert content itself when it is needed.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
```diff
@@ -0,0 +1,247 @@
...
+  m AddPerasImmutableCertResult
+implAddCert env cert = do
+  present <- atomically $ Map.member roundNo <$> readTVar (picdbState env)
+  result <-
+    if present
+      then pure CertAlreadyInImmutableDB
+      else do
+        -- no lock, assumes a single writer (the ChainDB adds
+        -- certificates from one background thread). Add a per-db lock if
+        -- concurrent 'addCert' for the same round ever becomes possible.
```
tbagrel1: That's a good question. I think we used to have a "objects on way to pool" queue, and then a dedicated single actor in charge of getting a lock on the DB and adding the queued objects into it. However I think this has been removed since. We would need to reintroduce this in one form or another.

tbagrel1 (commented):
Good job!

I think we have two axes we still need to work on:
- Having an index to prevent loading the whole DB into memory at startup, and then only loading the requested content when needed
- Having a way to query certs by increasing `PerasRoundNo`, as it seems they will be served that way for ObjectDiffusion

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
agustinmista: > I think I disagree with you here: ImmutableCertDB will never impact chainsel AFAIR. So I think we will only query the DB by ascending PerasRoundNo to serve in ObjectDiffusion.

Fair enough. I forgot this will only ever used for serving other clients.

In that case, yeah, indexing over `PerasRoundNo` might be easiest. 

Not sure we need to worry too much about performance at this point, but in the future we could replace this `Map` with something that only keeps a range of hot values in memory and drops the rest ass soon as they get processes by the client(s).

tbagrel1: > but in the future we could replace this Map with something that only keeps a range of hot values in memory and drops the rest ass soon as they get processes by the client(s).

But we may have a couple of clients, each working on a different range. We can just send the range of objects to OD; and OD will keep the objects live/hot by having them sit in the outstanding fifo for each connected peer. I don't see what we would obtain by reimplementing this logic inside the immutable DB too.

agustinmista: I was actually worried about the opposite: the case of multiple clients with overlapping FIFO ranges. 

If the DB doesn't know what's hot and what's not, would that mean that it might end up reading the same files over and over?

tbagrel1: But we have a small number of peers to serve (around 30 max? -- most of which won't be syncing nodes) with slow dynamic, while we have a huge number of possible objects to serve (possibly millions) with high dynamic. So I think it is not very likely that two peers ask at the same time to get the same objects (especially since syncing nodes, are by definition at a random point in the past), at least it isn't likely enough so that we would need special design for that. 

If we _really_ want to be fancy, or if loading from disk is judged drastically more costly than prolonged use of RAM, we could imagine a simple FIFO caching for loaded objects. But to be any useful/having any chance to get cache hits, we need the retention time to be as long as possible, because of how unlikely it is for two clients to need the same thing at the same exact time.

tbagrel1: Also, on a different level, considering the external factors at play ™️ , we may want to design the immutable DB to be as simple as needed and not waste too much engineering effort. It is not 100% sure yet it is here to stay.

agustinmista: Yeah, don't get me wrong: I'd love for this to be as simple as possible. We just need to be careful about things like:

> how unlikely it is for two clients to need the same thing at the same exact time

Because we could very easily misguide ourselves into optimizing for the best case, and not for the worst.

FTR, I'm totaly fine with having multiple file reads per cert if that justifies a much simpler implementation. I _really_ don't want us to reimplement Kafka :sweat_smile:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/API.hs
ninioArtillero: It does, https://github.com/IntersectMBO/ouroboros-consensus/pull/2293/commits/716bba9ea40f724d72f86e3c558da8743fb8d305 gives it a shot

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
ninioArtillero: I thinked is now solved by https://github.com/IntersectMBO/ouroboros-consensus/pull/2293/commits/716bba9ea40f724d72f86e3c558da8743fb8d305

Edit: just partially solved. The index is now a set of much lighter `PerasRoundNo`, resulting (with similar assumptions) in 8 bytes * 3600 / 90 * 24 * 365 = 2.8032 MB a year... which means it would take over 356 years to make a single GB.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
ninioArtillero: the index is now a `Set PerasRoundNo`, but still have implement a backup to load it only once.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
ninioArtillero: Thanks for the discussion, it is great input for expanding my understanding of the design space.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/PerasImmutableCertDB/Impl.hs
ninioArtillero: FYI, still looking into this. One possibility is getting the thread doing the writing to the ImmutableDB a lock on the PerasImmutableCertDB as well, and make the writing to both synchronous.

PR #2298, opened 2026-09-16
LeiosDb{Reader,Writer} API + single sqlite write connection

### LINKED ISSUE #1103: Production-grade Leios DB
## What

- We need to decide if [denormalization](https://github.com/IntersectMBO/ouroboros-consensus/pull/2288) is necessary. 
- Single writer connection
- ...

ch1bo:
Restructure the leios database API to have a reader / writer handle. This is not entirely necessary for the read end, but the writer API allows for a single writing connection in the background. The LeiosDbWriter is also asynchronous by default, returning a promise which can be used to wait for the result if needed. This is used in the fetch client that deliberately does not wait for the write to finish.

Encountered the need for asynchronous writing in the fetch thread while chasing for more throughput and the fetch clients not yielding for seconds (on huge volatile leios dbs) was an issue we can avoid this way.


ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -342,14 +343,14 @@ data Handlers m addr blk = Handlers
...
-      LeiosDbConnection m ->
+      LeiosDbWriter m ->
       NodeToNodeVersion ->
       ControlMessageSTM m ->
       ConnectionId addr ->
       Leios.LeiosPeerVars m ->
       LeiosFetchClientPeerPipelined LeiosPoint LeiosEb LeiosTx m ()
   , hLeiosFetchServer ::
-      LeiosDbConnection m ->
+      LeiosDbReader m ->
```
ch1bo, line 351: These two signatures demonstrate well the compartmentalization of the API

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Trace.hs
```diff
@@ -5,19 +5,4 @@ data TraceLeiosDb
...
-    -- means the lock was held for longer than that. A steady stream of these is
-    -- the signal that the write path is saturated.
-    TraceLeiosDbBusyRetry Int Double
-  | -- | A write transaction has been waiting far longer than contention
-    -- explains. Fields as for 'TraceLeiosDbBusyRetry'.
-    --
-    -- Repeated, not terminal: the wait is unbounded by design, since giving up
-    -- means throwing, and throwing here kills the Leios threads. A node that is
-    -- merely slow should stay a node that is merely slow.
-    TraceLeiosDbBusyStuck Int Double
```
ch1bo, line 37: These two are impossible now with a single writer

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1031,17 +1027,40 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
+      -- Wait for the write to complete (and trace) synchronously when we are
+      -- forging: need to ensure the data is written before advertising it.
+      -- TODO: do we really? Can we just optimistically continue and risk a peer
+      -- disconnect if we can't serve what we offer "in time"?
+      let traceCompleted = do
+            completedByBody <- await bodyWritten
+            unless (null completedByBody) $ do
+              st <- Leios.ebState <$> MVar.readMVar outstandingVar
+              forM_ completedByBody $ \p ->
+                traceWith ktracer $ TraceLeiosBlockTxsAcquired p (ebPointAge now st p)
```
ch1bo, line 1019: The age computation needing this `st` here is annoying, but did not want to change too many things

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1387,20 +1406,34 @@ processLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) txCache db system
...
+            ebStates <- Leios.ebState <$> MVar.readMVar outstandingVar
+            forM_ completed $ \p ->
+              traceWith ktracer $ TraceLeiosBlockTxsAcquired p (ebPointAge now ebStates p)
+      case source of
+        -- The forge's closure has to be on disk before the RB that references
+        -- it goes out.
+        ForgedTxs{} -> traceCompleted
+        _ -> void $ forkIO traceCompleted
+    -- The cache update: the fetch logic consults it to decide what is still
+    -- missing, so it cannot lag behind the caller.
```
ch1bo, line 1391: Should we maybe also update the tx cache / memory state before writing to disk here? Also, do we really need both, `TraceLeiosBlockTxsAcquired` and `TraceLeiosFetchTxsArrival`?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1031,17 +992,40 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
+  -- The DB-heavy tail: persist the body (~16k reference rows) and copy the
+  -- mempool-resident txs into the DB.
+  when shouldPersist $
+    traceException tracer TraceLeiosPeerDbException $ do
+      -- FIXME: once EB announcements are wired in the point MUST already
+      -- be present (announcement handling inserts it); until then insert
+      -- it idempotently as a stop-gap and trace a warning.
+      traceWith ktracer $ TraceLeiosBlockPointMissing point
+      _ <- writeEbPoint writer point ebBytesSize
+      bodyWritten <- writeEbBody writer point eb
```
ch1bo, line 1009: This is asynchronous now and we only `await` it in the forge case

ouroboros-consensus/bench/leios-gc-bench/Main.hs
```diff
@@ -168,7 +170,7 @@ main = do
           hPutStrLn stderr $ "Copying the database " <> immFixture <> " -> " <> benchImm
           copyFile immFixture benchImm
         db <- mkDb
-        close =<< open db
+        withReader db $ \_ -> pure ()
```
geo2a, line 173: Curios: I think that both before and after is a noop, I have little clue why Claude has decided to put it here.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -22,28 +35,21 @@ import LeiosDemoTypes
...
 
+-- | The database. Hands out readers and writers; owns neither.
 data LeiosDbHandle m = LeiosDbHandle
-  { subscribeEbNotifications :: HasCallStack => m (StrictTChan m LeiosEbNotification)
-  -- ^ Subscribe to new EBs and EBTxs being stored by the LeiosDB. This will
-  -- only inform about new additions, starting from when this function was
-  -- called.
+  { openReader :: HasCallStack => m (LeiosDbReader m)
+  , openWriter :: HasCallStack => m (LeiosDbWriter m)
+  -- ^ The database's one write path. Opening allocates nothing -- the backend
```
geo2a, line 44: Nit: I don't think it's correct to say "one" here as the caller is free to create `openWriter` as many times as they want.


```suggestion
  -- ^ Open a writer into LeiosDb. This allocates nothing -- the backend
```

geo2a: Perhaps we could add "only one of these must exist".

We should consider enforcing it at runtime via the `Electric` wrapper, as it `chainSelSync`. Not necessary in this PR.

geo2a (commented):
First small batch of comments, I'll continue tomorrow.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+  slack = 2
+
+-- | Open the write connections and start the worker draining the write queue.
+--
+-- The worker publishes each job's outcome into its 'WriteResult'. For ingest
+-- jobs it then rethrows a failure: publishing first lets an awaiting producer
+-- see the exception rather than block on a promise the dying worker would
+-- never fill, and a failed ingest write is not survivable (no caller catches
+-- 'LeiosDbException') -- producers still submitting eventually block on the
+-- full queue and go down as blocked-indefinitely. A failed maintenance job is
```
dnadales, line 1117: > "go down as blocked-indefinitely"?

I wonder if we can rely on this. The queue is captured in `openWriter` (`:182`), and the handle then lives in `ChainDbEnv` (`ChainDB/Impl/Types.hs:396`) and `NodeKernel` (`NodeKernel.hs:230`) for the node's lifetime, with background threads dereferencing it (`Background.hs:172`, `:252`, `:498`). So the queue stays reachable from a live root, and my reading is that the RTS then never delivers `BlockedIndefinitelyOnSTM` to a producer parked in `writeTBQueue`. Do you read it the same way?

`untilOrphaned`'s own docstring seems to agree: it says the notification arrives "after the LeiosDB handle is de-allocated", which never happens while the node runs. And if it did fire, the copier (`:437`) and the sweeper (`:737`) wrap their loops in `untilOrphaned` too, so they would swallow it and exit `pure ()`.

If that is right, the rethrow at `:1183` leaves us with a dead `leiosdb-writer` and a live node, since the fork at `:1144` is a bare `forkIO`. Every later `await` then parks forever, including `close` (`:1205`) and the two ChainDB background callers. The trigger looks ordinary rather than exotic: `withDie` (`:2319`) throws for any non-busy `DB.Error`, so `SQLITE_FULL` and `SQLITE_IOERR` reach it, including from `COMMIT`.

What about something like:

```haskell
threadId <- forkLinkedThread registry "leiosdb-writer" $ forever $ do
```

so that a failed ingest write really does take the node down, the way this comment intends? That would make "not survivable" true by construction instead of by argument.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+      Right _ -> pure ()
+    atomically $ putTMVar resultVar result
+
+-- | The write half of the API: every operation is queued for the worker of
+-- 'startWriter', and the 'Promise' waits for that job's result.
+sqliteWriter :: StrictTBQueue IO WriteJob -> LeiosDbWriter IO
+sqliteWriter queue =
+  LeiosDbWriter
+    { -- Not a teardown -- the write connection outlives every writer.
+      close = void . await =<< submitJob queue Flush
```
dnadales, line 1205: Should we protect this wait? If the writer thread is dead, `readTMVar` parks here forever.

What worries me is the shutdown path. `finalizeForging` calls this (`NodeKernel.hs:882`), and that is a registry release action. The shutdown thread then waits for the forging thread inside `cancelThread`, which is `uninterruptibleCancel`. So my reading is that neither thread can be interrupted out of it, and only `SIGKILL` ends the process. Do you read it the same way?

The previous `close` was `DB.close` on the calling thread, so it could not wait on anyone. This is the one place where that changed.

I wonder if the guard belongs in `submitJob` instead. There are eleven `await` sites on this queue, and two of them are ChainDB GC threads (`Background.hs:252`, `:498`). A timeout here would let us shut down, but a running node stays wedged. Something like a `writerAlive` flag that the worker clears in a `finally`, and every `await` reads with `orElse`, would make all eleven fail fast with a real error.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1031,17 +991,40 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
+      let traceCompleted = do
+            completedByBody <- await bodyWritten
+            unless (null completedByBody) $ do
+              st <- Leios.ebState <$> MVar.readMVar outstandingVar
+              forM_ completedByBody $ \p ->
+                traceWith ktracer $ TraceLeiosBlockTxsAcquired p (ebPointAge now st p)
+      case source of
+        ForgedBlock{} -> traceCompleted
+        ReceivedBlockFrom{} -> void $ forkIO traceCompleted
+  -- The cache updates: the fetch logic reads them to decide what is still
```
dnadales, line 1023: Two things about this `forkIO`.

First, does `traceException` still catch anything on this branch? The writes above only enqueue, so the exception comes out of `await` inside `traceCompleted`, and here that runs on a new thread. So my reading is that a peer-path DB write failure no longer becomes `TraceLeiosPeerDbException`. The `ForgedBlock` branch is fine, since it awaits inline. That trace has a real consumer: `LeiosDemoTypes.hs:2172` gives it an Error-severity `["DbException"]` namespace, and `leiosPeerTracer` is wired in at `NodeToNode.hs:719`.

Second, should this thread be linked? A bare `forkIO` sends an exception to GHC's child handler, so it lands on stderr and reaches no tracer, and the thread outlives `bracketLeiosPeer`. Everywhere else we use `forkLinkedThread registry` (`NodeKernel.hs:412`, `Background.hs:152`), and these two sites plus `LeiosDemoDb/SQLite.hs` are the only `forkIO` uses under `src/`.

Both would go away together if the trace moved inside the forked action and the fork were registered. Something like:

```haskell
let traceCompleted = traceException tracer TraceLeiosPeerDbException $ do
      completedByBody <- await bodyWritten
      ...
```

Related, on the line above: `writeEbPoint`'s promise at `:1009` is dropped with `_ <-` on every path, forge included. Should we await it, or keep it so a failed point write is visible somewhere?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+          dbWithWriteTransactionRaw tracer volDb $
+            useStmt pinStmt $ do
+              dbBindBlob pinStmt 1 ebHash.ebHashBytes
+              dbStep1Safe pinStmt
+      GcMark slot resultVar ->
+        publishMaintenance volDb immDb resultVar $
+          gcMark tracer copyQueue sweepDoorbell volDb gcStmts slot
+      CopyEb ebHash resultVar ->
+        publishMaintenance volDb immDb resultVar $
+          copyEbToImmutable tracer statsVar copierConn ebHash
```
dnadales, line 1166: Should the copy stay off the writer? Before, the copier had its own connection where `imm` was `main` and `vol` was attached, so `sql_copy_insert_ebTxs` and `sql_copy_insert_txs` wrote to the immutable file and only read `vol` under a snapshot. A different file means a different write lock, so a whole closure copied without touching volatile ingest at all. Only `sql_mark_as_copied` took the volatile lock, and in autocommit.

Now the bulk insert sits on the one thread that serves ingest. `appendToImmutable` uses `dbStep1Safe`, which we reserve for the long-running statements, so I think this is the largest single job the worker can pick up.

What also strikes me is that `CopyEb` and `WalCheckpoint` got none of the pacing we built for eviction. The `startSweeper` docstring says batching means "eviction never occupies the writer for long and ingest writes interleave between batches", and that reasoning seems to apply here just as much.

I am not claiming a net regression: the old five-connection setup had its own 20-way `BEGIN IMMEDIATE` contention with sleep backoff, so the queue may well win on the median. It is the tail I am unsure about. Would it make sense to leave `CopyEb` on a dedicated immutable connection, since it never needs the volatile write lock?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -573,69 +561,77 @@ gcOrphanTxBatchSize = 1024
...
-              nEbsMarked <- DB.changes db
-              pure (nTxsStagedAsGCCandidates, nEbsMarked)
-          when (nTxsStagedAsGCCandidates > 0 || nEbsMarked > 0) $
-            void $
-              tryPutMVar sweepDoorbell ()
+gcMark tracer copyQueue sweepDoorbell db gcStmts gcSlot = do
+  let GcStmts{gsStalePins, gsHasWork, gsAddGcCandidatesTxs, gsMarkEbForGC} = gcStmts
+  -- Self-heal: re-enqueue pinned-but-not-yet-copied EBs to the copier.
+  -- Recovers a crashed copy.
+  stalePins <- useStmt gsStalePins $ do
```
dnadales, line 602: Should the self-heal scan run after the `hasWork` gate rather than before it? `sql_gc_stale_pins` filters on `status = 1`, and I cannot find an index that covers it: we have `idx_ebs_ebHashBytes`, `idx_ebs_sweepable` (partial, `status IN (0,2)`) and `idx_ebs_markedForGc` (partial, `status = 3`). So this reads like an `ebSlot` range scan with a table probe per row, and `collectBlobs` materialises the hits into a list. `gsHasWork` just below matches `idx_ebs_sweepable` exactly, so it is the cheap check, and with a GC tick every 10 seconds an idle node pays the expensive one first on every tick.

The SQL is unchanged from before, so what is new is only where it runs: previously this was on `withMaintenanceConn`, holding no write lock, so it never blocked ingest. Now it occupies the writer before any write happens.

Two thoughts, and I am not sure which you prefer. We could move the `gsStalePins` read behind `hasWork`, or run it on a reader and pass the result into the `GcMark` job so the writer only does the marking. Either way the PR also deletes a per-tick connection open with six pragmas and the DDL, so the net may still be favourable. Do you have a `markSeconds` figure from `leios-gc-bench --scenario catchup` that would settle it?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+-- each upstream peer's fetch client, the forge, and the maintenance
+-- schedulers (copier, sweeper, the ChainDB's GC and promote calls) -- and a
+-- little slack. Deliberately shallow: depth beyond that adds no throughput
+-- (there is one worker) and only delays the commit-time notifications that
+-- gate vote scheduling and chain selection. A producer that outruns the disk
+-- blocks on submission, which is the backpressure.
+writerQueueDepth :: Natural
+writerQueueDepth = numUpstreamPeers + forge + maintenance + slack
+ where
+  numUpstreamPeers = 20
```
dnadales, line 1105: Is the budget one entry per peer, or one entry per job in flight?

`processLeiosBlock` submits `writeEbPoint` (`LeiosDemoLogic.hs:1009`) and `writeEbBody` (`:1010`). It does not await between them. It then reaches `writeTxs` through `:1034`. So one `MsgLeiosBlock` can occupy three entries. `forge` can occupy two, for the same reason. The never-block figure is then near 66, not 27.

I also have a question about the reason above. Take a job that arrives with `k` jobs ahead of it. Each job takes `W` to serve. If the queue holds all `k`, we enqueue at once, and the worker serves us at `t + kW`. If the queue is shorter, we block first and then enqueue. The worker still serves us at `t + kW`. One worker serves one job at a time in both cases. So depth does not change when a notification fires. Depth changes who waits, the queue or the producer.

Shallowness helps in another way. A blocked receiver sends no refunds. Fewer requests then go out, and the backlog gets smaller. Can we say that instead?

The node survives this either way. The demuxer throws instead of blocking (`Ingress.hs:134`). `maxRequestedBytesSizePerBigLedgerPeer` matches the ingress queue by design (`LeiosDemoTypes.hs:950`). The idle state has no timeout. So I do not claim a stall here.

But `numUpstreamPeers = 20` links to nothing in this repo. The real count comes from node config, so the constant will drift. Can we derive it? Or can we say it is a floor and not a bound?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+    WalCheckpoint !(WriteResult ())
+
+type WriteResult a = StrictTMVar IO (Either SomeException a)
+
+-- | Hand a job to the writer; the 'Promise' waits for its result. Blocks
+-- only while the queue is full, which is the backpressure.
+submitJob :: StrictTBQueue IO WriteJob -> (WriteResult a -> WriteJob) -> IO (Promise IO a)
+submitJob queue mkJob = do
+  resultVar <- newEmptyTMVarIO
+  atomically $ writeTBQueue queue (mkJob resultVar)
```
dnadales, line 1092: Can we trace this queue? Nothing reports its depth, the wait time of a job, or the time a producer spends blocked here. `LeiosDemoDb/Trace.hs` is unchanged by this PR.

The precedent is already in the tree twice. `TraceLeiosDbCopyQueueFull` (`Trace.hs:52`) covers the 1024-deep copy queue. Closer still, ChainDB traces a rising edge and then a falling edge carrying `lengthTBQueue` for `addBlockToQueue` (`ChainDB/Impl/Types.hs:655`), which is a bounded queue behind a promise, the same shape as `LeiosDbWriter`.

This matters because the old signal is now gone. `TraceLeiosDbBusyRetry` and `TraceLeiosDbBusyStuck` are emitted only from the `BEGIN IMMEDIATE` retry at `:2184`, and every caller of that loop runs on the one worker. You say as much in your note on `Trace.hs:37`. I agree that is correct and intended: the contention it measured no longer exists. But `Trace.hs:27` still promises "a steady stream of these is the signal that the write path is saturated", and the code can no longer produce that stream. So a saturated write path is now invisible in a devnet log.

One small thing in the same area: `gcMark` lost its `selfHeal`, `noopGuard` and `mark` sub-spans, and `sqlGarbageCollect` at `:572` now reports submit plus queue wait plus mark as one number, with `_gcCtx` unused. Can we keep those three, since they are what would tell us whether a GC job is the thing holding the worker?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
```diff
@@ -111,23 +114,48 @@ newLeiosDBInMemoryWith stateVar = do
...
+    LeiosDbWriter
+      { close = pure ()
+      , writeEbPoint = \point ebBytesSize ->
+          resolved (imInsertEbPoint stateVar point ebBytesSize)
+      , writeEbBody = \point eb ->
+          resolved (imInsertEbBody stateVar notificationChan point eb)
+      , writeTxs = \txs ->
+          resolved (imInsertTxs stateVar notificationChan txs)
+      }
+ where
```
dnadales, line 154: This backend runs the write before it returns the promise, so `await` never blocks and never throws. That is right for a `StrictTVar`. My worry is what it means for coverage.

`Common.hs:113` says `await` rethrows the write's exception in the awaiting thread. Here the exception comes out of the submit instead. So the two backends differ on the one property the new API is built around.

Every test that drives `processLeiosBlock` uses this backend: `Test/ThreadNet/Leios.hs:898`, `Test/ThreadNet/Network.hs:789`, `Test/LeiosDemoLogic/Invariants.hs:397` and `:756`, `Test/LeiosVoting.hs:227`. `Test/LeiosDemoDb.hs` does cover SQLite, but the new `rw*` helpers await every write (`:109`). Both benches await too, and they run one and three producers against a 27-deep queue.

So I cannot find anything that exercises an unresolved promise, FIFO ordering between two unawaited writes, a full queue, or a failed write. Those are the four things this PR adds.

Would a small SQLite-only test cover the first two cheaply? Submit a point and a body without awaiting, then await only the last write and check the body is present. That would pin the FIFO guarantee that `Forging.hs:217` and both benches already rely on in comments.

ouroboros-consensus/bench/leios-gc-bench/Main.hs
ch1bo: So your claude put it and my refactored it? :sweat_smile:

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
ch1bo: The comment is bad, I agree. Multiple writers are just fine and that's the whole point of this API to having multiple handles use a single connection

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+
+  -- The transaction brackets have already rolled back on failure; the extra
+  -- best-effort ROLLBACKs cover only the case where that rollback itself
+  -- failed and would otherwise leave a transaction open under every later job.
+  publishMaintenance :: DB.Database -> DB.Database -> WriteResult a -> IO a -> IO ()
+  publishMaintenance volDb immDb resultVar action = do
+    result <- MonadThrow.try action
+    case result of
+      Left (_ :: SomeException) -> do
+        _ <- DB.exec volDb "ROLLBACK"
```
dnadales, line 1193: Can a failed `ROLLBACK` reach the ingest path? `DB.exec` returns `Either (Error, Utf8) ()`, and we discard it here. So a real failure looks the same as the normal "cannot rollback, no transaction is active".

My concern is the coupling rather than this line. `publish` attempts no rollback at all. If a transaction stays open on `volDb`, the next ingest `BEGIN IMMEDIATE` returns a non-busy error. `publish` then rethrows and the worker dies. Before this PR, ingest held its own connection, so a poisoned maintenance connection could not reach it.

The base also had a repair path. `dropCopierConn` and `dropSweeperConn` finalised the statements and closed the connection, and the next loop reopened it. Both are gone now, because the statements live on the writer's permanent connections.

I am not claiming this happens. SQLite rolls back by itself on `SQLITE_FULL`, `IOERR` and `NOMEM`, so the two calls above almost always hit the benign case. Do you see a path where the rollback really fails? If one exists, can the worker reopen instead of carrying on?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -56,69 +62,86 @@ data LeiosDbHandle m = LeiosDbHandle
...
-  -- ^ Read the EB "closure": the tx hashes AND their tx bytes. Contrast
-  -- with 'leiosDbLookupEbBody' which returns only hashes + sizes.
-  -- Used by chain-sel's 'resolveLeiosClosure' to splice the EB's txs
-  -- back into the CertRB before applying to the ledger.
+  -- ^ Not a teardown -- the write path outlives every writer. Flushes: when
+  -- this returns, everything this caller submitted is durable.
+  , writeEbPoint :: HasCallStack => LeiosPoint -> BytesSize -> m (Promise m ())
+  -- ^ Record an announced EB's point and expected size. Idempotent.
+  , writeEbBody :: HasCallStack => LeiosPoint -> LeiosEb -> m (Promise m CompletedEbs)
+  -- ^ Persist an EB body; its point must already have been submitted.
```
dnadales, line 104: "must already have been submitted"

Submitted, or landed? The SQLite backend makes this work through FIFO order in one queue, but neither `LeiosDbWriter` nor `Promise` states that the queue is FIFO.

Four places already depend on it. `LeiosDemoLogic.hs:1009` drops the point promise. `DBSynthesizer/Forging.hs:217` says "awaiting the last write flushes all three". Both benches carry the same comment.

The in-memory backend satisfies the ordering by accident, because it writes at submit. So a backend that batches or reorders will break all four sites, and no test will catch it.

Can we state the ordering here, or next to `Promise`?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -56,69 +62,86 @@ data LeiosDbHandle m = LeiosDbHandle
   -- ^ Sample 'LeiosDbStats' counters.
   }
 
-data LeiosEbNotification
-  = AcquiredEb LeiosPoint BytesSize
-  | AcquiredEbTxs LeiosPoint
+-- | Queries. Never writes, so any number may run concurrently -- each on its
```
dnadales, line 65: "Never writes"

`openSQLiteReader` opens both partitions with `SQLOpenReadWrite, SQLOpenCreate` (`SQLite.hs:1008` into `:341` and `:301`), and the volatile open runs the `sql_schema_gc` DDL every time. So a reader's connections can write, even though none of the six methods does.

That predates this PR, so it is not a regression. What is new is that it is now redundant: `startWriter` creates and migrates both files before the handle escapes (`:166`), so nothing is left for a reader to create. And `openReadOnlyRawConnection` (`:284`) already backs the identical `sqliteReader` in the read-only handle (`:200`).

Is the two-line swap to `SQLOpenReadOnly` worth doing here, or do you prefer to soften this sentence? One caveat if we swap: `prepareVolStmts` prepares INSERT and UPDATE statements, and the read-only handle has no caller in the repo, so I cannot tell from here whether those prepare on a read-only connection.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/Common.hs
```diff
@@ -56,69 +62,86 @@ data LeiosDbHandle m = LeiosDbHandle
...
+withReader :: MonadThrow m => LeiosDbHandle m -> (LeiosDbReader m -> m a) -> m a
+withReader db = bracket (openReader db) (\r -> r.close)
+
+withWriter :: MonadThrow m => LeiosDbHandle m -> (LeiosDbWriter m -> m a) -> m a
+withWriter db = bracket (openWriter db) (\w -> w.close)
+
+allocateReader :: IOLike m => ResourceRegistry m -> LeiosDbHandle m -> m (LeiosDbReader m)
+allocateReader registry db = snd <$> allocate registry (\_ -> openReader db) (\r -> r.close)
+
+allocateWriter :: IOLike m => ResourceRegistry m -> LeiosDbHandle m -> m (LeiosDbWriter m)
```
dnadales, line 134: `allocateWriter` and `allocateReader` have no call sites anywhere in the repo.

Two places could use them. `NodeKernel.hs:875` hand-rolls the same allocate-and-finalize pair for the forge. `ImmDBServer/Diffusion.hs:131` still carries `-- XXX: leaks resources` with a registry in scope at `:118`, which `allocateReader` would fix outright.

Can we wire them in at those two, or drop them until something needs them?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1031,17 +991,40 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
+  -- missing, so they must land before we return.
   --
-  -- Overlap (in both the mempool and the cache): already persisted in the DB, so
-  -- only upgrade them to Applied in the cache -- no (redundant) DB insert.
+  -- Overlap (in both the mempool and the cache): already persisted, so only
+  -- upgrade them to Applied in the cache -- no (redundant) DB insert.
```
dnadales, line 1027: "already persisted"

Is it, at this point? The write that put those txs in the cache came from `ingestAcquiredTxs`, which submits and then forks the await (`:1386`). So a cache entry now means the insert was submitted, not that it committed.

This was exact before the change, since `leiosDbInsertTxs` ran synchronously ahead of the cache update. Nothing breaks today, because the only consumer of the tag reads txs it loaded from disk. But the reason stated here no longer holds, and it is the kind of sentence someone will rely on later.

Can we say "already submitted" instead?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1387,20 +1370,34 @@ processLeiosBlockTxs ktracer tracer (outstandingVar, readyVar) txCache db system
...
+            completed <- await txsWritten
+            ebStates <- Leios.ebState <$> MVar.readMVar outstandingVar
+            forM_ completed $ \p ->
+              traceWith ktracer $ TraceLeiosBlockTxsAcquired p (ebPointAge now ebStates p)
+      case source of
+        -- The forge's closure has to be on disk before the RB that references
+        -- it goes out.
+        ForgedTxs{} -> traceCompleted
+        _ -> void $ forkIO traceCompleted
+    -- The cache update: the fetch logic consults it to decide what is still
```
dnadales, line 1390: Should this match name all three constructors? The sibling split at `:1022` spells out `ForgedBlock{}` and `ReceivedBlockFrom{}`, so a new constructor there becomes a compile error. Here `_` would give a fourth `LeiosBlockTxsSource` constructor fire-and-forget behaviour silently.

The current behaviour is right, as far as I can tell: `MempoolTxs` only arrives from the peer path, and even on the forge path the later `ForgedTxs` write is awaited, so FIFO order covers anything submitted before it.

So this is only about keeping the two sites in the same shape. Something like:

```haskell
case source of
  ForgedTxs{} -> traceCompleted
  ReceivedTxsFrom{} -> void $ forkIO traceCompleted
  MempoolTxs{} -> void $ forkIO traceCompleted
```

The comment above could also say why the other two need no wait, since the reason is not the same for both.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -429,122 +452,87 @@ startCopier tracer statsVar copyQueue volPath immPath = do
...
+          dbBindBlob ccInsertEbTxs 1 ebHash.ebHashBytes
+          dbStep1Safe ccInsertEbTxs
+        nTxs <- DB.changes ccDb
+        -- copy the transactions
+        useStmt ccInsertTxs $ do
+          dbBindBlob ccInsertTxs 1 ebHash.ebHashBytes
+          dbStep1Safe ccInsertTxs
+        pure (Just nTxs)
 
 -- | Close the connection if the action throws, then rethrow. For partial
```
dnadales, line 537: This docstring still says "For partial acquisition of `CopierConn`", but `prepareCopierConn` now takes a connection rather than opening one.

The shape it describes does still exist, one caller over: in `startWriter` the guard covers only `openRawConnection immPath`, and `mkConn`, the `ATTACH` and the four `prepare` calls after it are unguarded. A throw in any of those leaks both connections.

In practice that costs nothing, since every caller of `newLeiosDBSQLite` lets the exception end the process. So I would only ask for the docstring to point at its real call site, unless you would rather extend the guard while it is fresh.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -680,14 +679,10 @@ mkHandlers
                           )
 
           pure (leiosNotifyServerPeerLookahead incr next, pump)
-      , hLeiosFetchClient = \leiosConn _version controlMessageSTM peer peerVars -> toLeiosFetchClientPeerPipelined $ Effect $ do
+      , hLeiosFetchClient = \_version controlMessageSTM peer peerVars -> toLeiosFetchClientPeerPipelined $ Effect $ do
+          -- This client's submission point into the node's one write path.
+          writer <- openWriter leiosDB
```
dnadales, line 684: This is the one writer we open without a bracket. It is safe, since `openWriter` is `pure` in both backends, but that makes the call site depend on an invariant stated in `Common.hs:45` rather than here.

`withWriter` fits exactly where `withLeiosDb` used to sit, around `runPipelinedPeerWithLimits` at `:1493`. That would restore the bracket with no restructuring, and the documented flush on `close` would then actually run for fetch clients.

Not important if you would rather keep the writer where the comment explains it. Mostly I noticed that the fetch server two functions down does use `withReader`, so the two halves read differently.

ouroboros-consensus/bench/leios-gc-bench/Main.hs
```diff
@@ -546,9 +547,10 @@ runPhases opts db flushEvents latRef sweepBacklog schedule immBefore = do
         snd
           <$> timed
             ( do
-                leiosDbInsertEbPoint c point (leiosEbBytesSize eb)
-                _ <- leiosDbInsertEbBody c point eb
-                _ <- leiosDbInsertTxs c txs
+                _ <- writeEbPoint w point (leiosEbBytesSize eb)
+                _ <- writeEbBody w point eb
+                -- Awaiting the last write times all three to durability.
```
dnadales, line 552: Two of these numbers changed meaning rather than value.

`prInsertWall` still measures submit to durable, so it works as a trend line. But its composition moved from SQLite lock contention to queue wait behind maintenance, so a regression here can no longer be attributed to the write itself.

`prCopyEbWall` is the one I would not compare across commits. The `copyToImmutable` span at `SQLite.hs:437` now covers the submit and the wait, where before it covered `getConn`, `appendToImmutable` and the mark, run inline on the copier thread. So per-EB copy cost became copy cost plus waiting for unrelated jobs.

I also think `prProbeWall` went for a mechanical reason rather than a deliberate one: `appendToImmutable` lost its `CallCtx` when it moved out of `startCopier`'s where-clause, so `completenessProbe` could no longer be emitted. Was dropping the field the intent, or would you rather thread the context through and keep it?

Separately, neither bench reaches the new path: both await every write, and they run one and three producers against a 27-entry queue. Would a fire-and-forget variant be worth adding while this is fresh?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+    , lookupEbClosure = sqlLookupEbClosure conn
+    }
+
+-- * The single writer
+
+-- | One queued write, carrying the variable its result lands in.
+--
+-- The ingest jobs come from 'LeiosDbWriter'; the maintenance jobs from the
+-- handle's promote\/GC entry points and the copier\/sweeper schedulers. One
+-- worker running them all is what makes the single-writer property total: no
```
dnadales, line 1058: Commenting here because the code I mean is unchanged and so outside the diff: the `BEGIN IMMEDIATE` retry loop in `dbWithWriteTransactionRaw`, around `SQLite.hs:2169`.

That loop retries forever on `ErrorBusy`. Its comment says "Unbounded, deliberately. Nothing is held while waiting here ... so waiting costs latency and nothing else." I read the premise as still true, since `BEGIN IMMEDIATE` takes the lock before opening a snapshot, so no WAL pages are pinned.

The conclusion is what this PR moves. "Latency and nothing else" used to mean one of five write connections stalled. With the single-writer property above, the same wait now stops ingest, GC, promote, copy and sweep together. The only signal is `TraceLeiosDbBusyStuck` every 30 seconds or so.

In-process I could not find anything that can still hold that lock: readers use `BEGIN` deferred, the `vol` write through `immDb` runs in autocommit, and the stats sampler opens no connection. What is left is the tool functions through `withExistingLeiosDbFile` at `:1590`, which deliberately set no `busy_timeout`, and a second process on the same file.

No change requested. Is an unbounded wait still the right default now that it has node-wide reach? Or is it worth a note naming the tools as the remaining case?

ouroboros-consensus/bench/leios-gc-bench/Main.hs
geo2a: Yeah :laughing: We can leave it in, I don't mind.

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -864,18 +865,22 @@ forkBlockForging IS{..} (MkBlockForging blockForgingM) =
   label :: String
   label = "NodeKernel.blockForging"
 
-  -- 'LeiosDbConnection' is not thread-safe, so we open one per
+  -- 'LeiosDbReader' is not thread-safe, so we open one per
   -- forge-credentials thread (and close it when the thread exits).
+  -- The writer is this thread's submission point into the node's one
+  -- write path; closing it merely flushes.
```
geo2a, line 871: I don't think this comment is useful as it just repeats the haddock for `LeiosDbReader`/`LeiosDbWrite`.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: CopyEb indeed would not need to block the other writers (on the volatile partition). I'll change it

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: > holding no write lock, so it never blocked ingest.

It was implicitly because there can only be one writer at a time on a WAL-mode sqlite database. The busy handling would have resulted in serialization of writes. I do not want to change the garbage collection semantics in this same PR

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: It's a not very educated guess. Your analysis is correct and it may very well be that the network latency is lower than the write latency. It does not _need_ to match to be correct, it just shifts the congestion around. The queue depth may only matter once we start (if ever) deciding things based on what is in the queue (e.g. when to checkpoint the wal)

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: Agree on tracing interesting events, not so much about retaining these "call contexts" across threads. This PR is essentially breaking up individual function calls into asynchronous + await semantics across two threads which now also depend on the queue depth etc (as you also comment here). Having traces that indicate the application logic runtime (how long it took to copy, garbage collect etc) I do agree, but not instrumenting several function calls down the stack with surface language (! instead of using profiling flags which automate this).

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/InMemory.hs
ch1bo: Agree with improving the error handling story and making it consistent across backends

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
ch1bo: It was a white lie before, but a blatant lie now. Will amend the comment of course

ouroboros-consensus/bench/leios-gc-bench/Main.hs
ch1bo: Dropping the call ctx was deliberate, but I did not look into the bench mechanics much. @geo2a you have an opinion in how this should evolve?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
ch1bo: Unbounded does not seem right. The node should not really need any `busy_timeout` (only a single writer, and if not we'd like to know the regresssion early?). But if we consider concurrent use of db-... tools a smallish `busy_timeout` and no handling of database busy errors is probably best of both worlds?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
```diff
@@ -1031,17 +991,40 @@ processLeiosBlock ktracer tracer (outstandingVar, readyVar) txCache db systemTim
...
-  -- fetch job set above). This pays disk latency, so it's synchronous here.
+  -- The DB-heavy tail: persist the body (~16k reference rows) and copy the
+  -- mempool-resident txs into the DB.
+  when shouldPersist $
+    traceException tracer TraceLeiosPeerDbException $ do
+      -- FIXME: once EB announcements are wired in the point MUST already
+      -- be present (announcement handling inserts it); until then insert
+      -- it idempotently as a stop-gap and trace a warning.
+      traceWith ktracer $ TraceLeiosBlockPointMissing point
+      _ <- writeEbPoint writer point ebBytesSize
```
geo2a, line 1008: I've noticed that you are not waiting on this, but only on the next operation of writing the body. The haddock for `sqlInsertEbBody` says: "The point MUST already be present (inserted
-- via 'sqlInsertEbPoint' on the announcement path)." --- that sounds like we should `await` on the point insertion before we submit the body for insertion.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -182,10 +197,16 @@ newLeiosDBSQLiteReadOnly tracer volLeiosDbPath immLeiosDbPath = do
       , leiosDbGarbageCollect = \_ -> pure ()
       , leiosDbPromoteToImmutable = \_ -> pure ()
       , leiosDbSampleStats = readIORef statsVar
-      , open = do
+      , openReader = do
           volDb <- openReadOnlyRawConnection volLeiosDbPath
           immDb <- orCloseOnError volDb $ openReadOnlyRawConnection immLeiosDbPath
-          mkSQLiteConnection tracer notificationChan statsVar volDb immDb
+          sqliteReader <$> mkConn tracer statsVar volDb immDb
+      , openWriter =
```
geo2a, line 204: Note-to-self: I need to refactor the callers of `newLeiosDBSQLiteReadOnly`to use `LeiosReader` instead.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -341,18 +354,18 @@ copyQueueCapacity :: Natural
...
-        dbBindBlob stmt 1 point.pointEbHash.ebHashBytes
-        dbStep1Safe stmt
+  StrictTBQueue IO WriteJob ->
+  StrictTBQueue IO EbHash ->
+  Tracer IO TraceLeiosDb ->
+  LeiosPoint ->
+  IO ()
+sqlPromoteToImmutable writeQueue copyQueue tracer point = do
+  await =<< submitJob writeQueue (PinEb point.pointEbHash)
   enqueueCopy tracer copyQueue point.pointEbHash
```
geo2a, line 369: Now that we have a background writer thread, we may want to eliminate the copy thread, which now just submits work to the writer thread. We may as well submit this work from here directly.

We don't need to do it in this PR, I'm marking this as a TODO for myself.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -398,29 +411,39 @@ untilOrphaned loop =
...
-  connRef <- newIORef Nothing
+  rootCtx <- rootCallCtx taskName
   threadId <- forkIO $ untilOrphaned $ forever $ do
     ebHash <- IO.atomically $ readTBQueue copyQueue
-    copyOne rootCtx connRef ebHash
+    copySpan
+      rootCtx
+      "copyToImmutable"
+      (show ebHash)
+      (\_ctx -> await =<< submitJob writeQueue (CopyEb ebHash))
```
geo2a, line 444: As I said in the comment above, I don't think this background thread is very useful anymore, as it just creates work for the writer thread instead of doing it directly. No need to take care of that in this PR, I'm just drawing your attention to it. I'll refactor it myself.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
```diff
@@ -1039,56 +996,218 @@ useStmt :: DB.Statement -> IO a -> IO a
...
+-- | One queued write, carrying the variable its result lands in.
+--
+-- The ingest jobs come from 'LeiosDbWriter'; the maintenance jobs from the
+-- handle's promote\/GC entry points and the copier\/sweeper schedulers. One
+-- worker running them all is what makes the single-writer property total: no
+-- two in-process connections ever contend for a write lock.
+--
+-- TODO: the worker runs jobs strictly FIFO; it could become queue-aware and
+-- have maintenance jobs yield to pending ingest writes.
+data WriteJob
```
geo2a, line 1063: I love it! It's a great idea to formalise these actions as a single data type, it greatly improves visibility of what the logic can do.

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoDb/SQLite.hs
geo2a: I agree that this is a great place to trace.

geo2a (approved):
I went through the changes to the LeiosDB GC, some bits in the LeiosDemoLogic and the new single-writer thread logic.

I like the single-writer queue approach. I even think that we should refactor the LeiosDB copier/GC threads not run in the background but just submit work in batches to the writer thread. This way we'll be able to minimise the amount of SQLite connections and maintain keep the logic in one place. We can always re-introduce an extra connection if we see a bottleneck.

ouroboros-consensus/bench/leios-gc-bench/Main.hs
geo2a: `leios-gc-bench` which I added does use the callTrace mechanism to keep track of how long things take. This benchmark was completely vibe-coded. I think the benchmark should measure the time it takes to perform all copy/gc operations, but it could be doing that in some different way than using callTrace.

ouroboros-consensus/bench/leios-gc-bench/Main.hs
```diff
@@ -232,16 +226,18 @@ main = do
...
--- @FOO.vol.db@).
+-- | The imm-file sibling of a volatile fixture path. Fixtures are named the
+-- way the node names its partitions (@FOO.vol.db@); a @FOO.vol@ from before
+-- that rename is still accepted, and keeps its own pairing.
 immSiblingOf :: FilePath -> FilePath
 immSiblingOf path
-  | ".vol.db" `List.isSuffixOf` path =
-      take (length path - length (".vol.db" :: String)) path <> ".imm.db"
-  | ".vol" `List.isSuffixOf` path =
-      take (length path - length (".vol" :: String)) path <> ".imm"
```
ch1bo: Was this .vol naming ever released?

ouroboros-consensus/src/ouroboros-consensus/LeiosDemoLogic.hs
ch1bo: No, I think the comment is fine. We only have a cache hit when the write is on its way. Writes are asynchronous now and this can result in:
1. write never happens -> sqlite writer queue is full and head of line blocked; accepted, the node is stuck and we should detect that case in the database (backend)
1. write fails
  a. in our application -> thread linking should result in an exception on the main thread
  b. outside our application (file system, OS, disk caches) -> durability was also not guaranteed before, this is a whole other story

ch1bo:
Addressed most of the review comments, but am conscious of changing too much of the GC logic. Let's improve further on the leios-prototype or when we make this production-grade

PR #2300, opened 2026-09-17
Reduce the quantity of transactions in a block that contains a Peras certificate

nbacquey:


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/SupportsPeras.hs
```diff
@@ -177,6 +186,22 @@ deriving instance
 deriving instance
   Generic (PerasEpochContext blk)
 
+-- | Get the "size" of an object (in practice, a Peras certificate) as if it were a transaction.
+--
+-- This is used to select fewer transactions when forging a block that should contain
+-- a Peras certificate, to avoid exceeding the maximum block size.
+-- In practice, only the byte size part of the 'TxMeasure' should be non-zero.
+class TxLimits blk => IsTxSizeable cert blk where
+  getTxLikeSize :: cert -> TxMeasure blk
```
agustinmista, line 195: This looks generic enough to be next to the definition of `TxLimits` (with `s/cert/a/g`)

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
```diff
@@ -21,12 +27,19 @@ import Ouroboros.Consensus.Block.SupportsPeras
...
 -------------------------------------------------------------------------------}
 
-instance BlockSupportsPeras ByronBlock where
+instance
+  TxLimits ByronBlock =>
+  IsTxSizeable (VoidPerasCert ByronBlock) ByronBlock
+  where
+  getTxLikeSize = absurd . unVoidPerasCert
+
+instance TxLimits ByronBlock => BlockSupportsPeras ByronBlock where
```
agustinmista, line 42: This superclass constraint should be redundant, no?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
```diff
@@ -179,11 +188,36 @@ instance
...
+  ) =>
+  IsTxSizeable cert (ShelleyBlock proto era)
+  where
+  getTxLikeSize cert =
+    TxMeasure
+      { tmPhase1 = certSize
+      , tmPhase2 = Measure.zero
+      }
+   where
+    certSize = fromBS32 $ ByteSize32 $ fromIntegral $ LazyByteString.length $ serialize cert
```
agustinmista, line 209: Design question: do we want to use the real size via serialization, or can we make do with the much cheaper upper bound?

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
```diff
@@ -87,6 +88,8 @@ codecConfig :: CodecConfig StandardShelleyBlock
 codecConfig = ShelleyCodecConfig
 
 fromShelleyLedgerExamples ::
+  FromBS32 (TxMeasurePhase1 (ShelleyBlock (TPraos StandardCrypto) era)) =>
+  ToCBOR (PerasCert (ShelleyBlock (TPraos StandardCrypto) era)) =>
   ShelleyCompatible (TPraos StandardCrypto) era =>
```
agustinmista, line 93: Two things here:

* I've seen wonky type-errors when using  `a => b => c` instead of a `(a, b) => c`.
* I feel these should be somehow part of `ShelleyCompatible`.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/MaryAlonzo.hs
```diff
@@ -1,7 +1,10 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
agustinmista, line 7: These all look like stray changes, nothing else has changed in this module.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/ShelleyAllegra.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
```
agustinmista, line 5: [Ditto](https://github.com/IntersectMBO/ouroboros-consensus/pull/2300/changes#r4080277544)

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
```diff
@@ -464,21 +466,29 @@ getTransactionsToForge ::
...
+      snapshotTakeWithInitialPayload mempoolSnapshot certSize $
+        blockCapacityTxMeasure (configLedger cfg) tickedLedgerState
+    -- NB respect the capacity of the ledger state we're extending,
+    -- which is /not/ 'snapshotLedgerState'
+
+    toTxMeasure :: BlockSupportsPeras blk => PerasCert blk -> TxMeasureWithDiffTime blk
+    toTxMeasure cert =
+      MkTxMeasureWithDiffTime
+        (getTxLikeSize cert)
+        Measure.zero -- REVIEW: is it correct to ignore the time component of cert validation?
```
agustinmista, line 491: Good question. Let's put a pin on it.

If we are going to ignore it, I think we should demote this to a `TxMeasure` without diff time.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE MonoLocalBinds #-}
```
agustinmista, line 2: MonoLocalBinds can usually be avoided by simplifying some constraint, no?

What does the error message say when you remove it?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -437,9 +437,10 @@ data MempoolSnapshot blk = MempoolSnapshot
   -- ^ Get all transactions (oldest to newest) in the mempool snapshot,
   -- along with their ticket number, which are associated with a ticket
   -- number greater than the one provided.
-  , snapshotTake :: TxMeasure blk -> ([Validated (GenTx blk)], TxMeasureWithDiffTime blk)
+  , snapshotTakeWithInitialPayload ::
+      TxMeasureWithDiffTime blk -> TxMeasure blk -> ([Validated (GenTx blk)], TxMeasureWithDiffTime blk)
   -- ^ Get the greatest prefix (oldest to newest) that respects the given
-  -- block capacity, and the prefix's total size.
+  -- block capacity, the initial payload of the block, and the prefix's total size.
```
agustinmista, line 443: I think we need to document carefully that the `TxMeasureWithDiffTime` is the size of the initial payload.

Also, why do we need the `WithDiffTime`?

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
```diff
@@ -270,6 +270,15 @@ instance
...
   ShowProxy (OneEraPerasCert xs)
 
+instance
+  ( Typeable xs
+  , All (Compose ToCBOR WrapPerasCert) xs
+  ) =>
+  ToCBOR (OneEraPerasCert xs)
+  where
+  toCBOR (OneEraPerasCert cert) =
+    encodeNS (hcpure (Proxy @(Compose ToCBOR WrapPerasCert)) (fn (K . toCBOR))) cert
```
agustinmista, line 280: Are these all new `ToCBOR` instances needed just to gauge the size of the cert? 

If so, I would be more inclined to drop them and use the conservative upper bound instead.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -532,3 +533,10 @@ data Invalidated blk = Invalidated
   { getInvalidated :: !(GenTx blk)
   , getReason :: !(ApplyTxErr blk)
   }
+
+-- | Helper type class for measures that can be built from a ByteSize32
+class FromBS32 measure where
+  fromBS32 :: ByteSize32 -> measure
```
agustinmista, line 539: Really hoping we don't need this. But if needed, could we restrict `measure` to be some sort of `TxMeasure ...`?

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
```diff
@@ -941,16 +942,21 @@ prop_TxSeq_lookupByTicketNo_sound smalls small =
...
+--     transaction sizes, plus initial payload size, are less than or equal to that of the
+--     byte size which the 'TxSeq' was split on,
+--  * the initial payload size is greater than or equal to the byte size which the 'TxSeq'
+--    was split on, in which case no transactions have been selected.
 prop_TxSeq_splitAfterTxSize :: TxSizeSplitTestSetup -> Property
 prop_TxSeq_splitAfterTxSize tss =
-  property $ txSizeSum (TxSeq.toList before) <= tssTxSizeToSplitOn
+  property $
+    txSizeSum (TxSeq.toList before) `Measure.plus` tssTxInitialPayloadSize <= tssTxSizeToSplitOn
+      || (tssTxSizeToSplitOn <= tssTxInitialPayloadSize && null (TxSeq.toList before))
```
agustinmista, line 955: I think these two different boolean conditions would need some QuickCheck tabulation to be sure we are indeed hitting both when generating tests.

agustinmista (commented):
Hey, nice job!

I have some initial comments. In general, I think we could simplify this PR substantially if we didn't require real serialization to compute the initial payload measure. The cost of using the cert size upper bound you wrote should be relatively minor IMO (since it's also shown to be a relatively good upper bound).

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
nbacquey: Long story short, we need the `WithDiffTime` because that is what is compared internally: the transactions are validated, which adds the `DiffTime` measure to the TxMeasure.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/AcrossEras.hs
nbacquey: Those are indeed needed just to have the size of a cert. I think it's important to have at least the one below which you're commenting, though...
The upper bound works for concrete `PerasCert blk`; this one is a `NS`, which might be bigger due to block type serialisation.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
nbacquey: We can get rid of this typeclass at the cost of having a specific `instance IsTxSizeable cert (ShelleyBlock proto era)` for each specific era. Do you think it's worth the boilerplate? I don't have a particular opinion on the matter :shrug:

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node.hs
nbacquey: This is the warning message I'm getting:
```
    • The constraint ‘IsTxSizeable
                        (OneEraPerasCert xs) (HardForkBlock xs)’ matches
        instance (CanHardFork xs,
                  Data.SOP.Constraint.All
                    (Data.SOP.Constraint.Compose
                       cardano-binary-1.9.1.0:Cardano.Binary.ToCBOR.ToCBOR
                       Ouroboros.Consensus.TypeFamilyWrappers.WrapPerasCert)
                    xs,
                  Ouroboros.Consensus.Ledger.SupportsMempool.FromBS32
                    (Ouroboros.Consensus.Ledger.SupportsMempool.TxMeasurePhase1
                       (HardForkBlock xs)),
                  Ouroboros.Consensus.Ledger.SupportsMempool.TxLimits
                    (HardForkBlock xs)) =>
                 IsTxSizeable (OneEraPerasCert xs) (HardForkBlock xs)
          -- Defined at /home/nicolas/tweag/ouroboros-consensus/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Basics.hs:611:3
      This makes type inference for inner bindings fragile;
        either use MonoLocalBinds, or simplify it using the instance
    • In the instance declaration for ‘RunNode (HardForkBlock xs)’

```
If I understand correctly, it means I can replace `IsTxSizeable (OneEraPerasCert xs) (HardForkBlock xs)` with
```haskell
( All (Compose ToCBOR WrapPerasCert) xs
, FromBS32 (TxMeasurePhase1 (HardForkBlock xs))
, TxLimits (HardForkBlock xs)
)
```
My opinion was that those constraints seemed too specific to have a place in such a high level module.

ouroboros-consensus-cardano/test/cardano-test/Test/ThreadNet/MaryAlonzo.hs
nbacquey: Well spotted, those are from a previous iteration of the PR

ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel/Forge.hs
nbacquey: See the justification below. It will need to be turned into a `TxMeasureWithDiffTime` at **some** point eventually, because that is what is compared internally. I think here is the best place to do it, because it is explicit that we are measuring a certificate, the validation time of which we (probably) don't care about.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/Node.hs
nbacquey: Addendum:
In fact, I can replace the constraint with "just"
```haskell
( All (Compose ToCBOR WrapPerasCert) xs
, FromBS32 (TxMeasurePhase1 (HardForkBlock xs))
)
```
Which is still quite specific IMO.

ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/Node/Peras.hs
nbacquey: We still need it here because the actual instance for `TxLimits ByronBlock` is lower in the dependency graph. Importing the instance would create a cyclic dependency.

Do you think it's worth a refactoring?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
nbacquey: That's an interesting question. I can see some issues with this though:
- A potential loss of capacity due to being too conservative. As you noted, it would be relatively minor because the bound is relatively tight. It might not be a real concern.
- The upper bound only holds for `V1.PerasCert`. We would need another upper bound for every other concrete type of certificates:
  -  `VoidPerasCert` won't be an issue.
  - `OneEraPerasCert`, which is an NS of `VoidPerasCert` and `V1.PerasCert`. If the actual cert is `V1`, the upper bound is incorrect, because it doesn't account for the block type serialization (is it negligible?). If the actual cert is `Void` (which cannot happen in practice, but can't be enforce in tests/type level), we would need a new, artificial upper bound.
  - `MockPerasCert`: I'm not sure in which case this one is actually used, and if it can be included in blocks. Probably not, but then we would need another useless upper bound.

ouroboros-consensus-cardano/src/unstable-shelley-testlib/Test/Consensus/Shelley/Examples.hs
nbacquey: - I've used this to be consistent with the rest of the signature, which uses `a => b => c`
- Won't matter if those are part of `ShelleyCompatible`, which I agree with

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -454,6 +458,19 @@ class
     TickedLedgerState blk mk ->
     TxMeasure blk
 
+-- | A class to get the "size" of an object as if it were a transaction.
+--
+-- This is used to select fewer transactions when forging a block that should contain
+-- more than transactions as a payload (e.g. a Peras certificate), to avoid exceeding
+-- the maximum block size.
```
ninioArtillero: Wording is confusing here

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
ninioArtillero: Are this in correspondance with the function's arguments?

ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Node/Peras.hs
nbacquey: Note: we decided to go with an upper bound, which is being implemented

ouroboros-consensus/test/consensus-test/Test/Consensus/Mempool.hs
```diff
@@ -941,16 +942,21 @@ prop_TxSeq_lookupByTicketNo_sound smalls small =
   mkTicket x = TxTicket x (mkTicketNo x) mempty
   mkTicketNo = TicketNo . toEnum
 
--- | Test that the 'fst' of the result of 'splitAfterTxSize' only contains
--- 'TxTicket's whose summed up transaction sizes are less than or equal to
--- that of the byte size which the 'TxSeq' was split on.
+-- | Test that one of the following is true:
+--   * the 'fst' of the result of 'splitAfterTxSize' only contains 'TxTicket's whose summed up
```
ninioArtillero:
```suggestion
--   * the 'fst' of the result of 'splitAfterTxSizeWithInitSize' only contains 'TxTicket's whose summed up
```

Perhaps the property test itself should be renamed as well.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs
nbacquey: As discussed, I changed the payload to a `Maybe` and edited the documentation, for clarity

ninioArtillero (commented):
LGTM, I'm just suggesting minor of documentation fixes.

PR #2302, opened 2026-09-17
docs: add leiosfetch_spec.md

nfrisby:
A description of the LeiosFetch design for the RC candidate, with TODOs for the parts that aren't yet landed. See Issue https://github.com/input-output-hk/ouroboros-leios/issues/1074 for the tracking of those TODOs.


nfrisby:
Rendered https://github.com/IntersectMBO/ouroboros-consensus/blob/nfrisby/leiosfetch-doc/docs/website/contents/references/leiosfetch_spec.md

nfrisby:
Copying over Sebastian's comments he made before I had noticed I only shared the document link and not the PR link

> Why this batch size? It's not really aligned with the fetch request addressing scheme (the roaring bitmap inspired one)

More than one job can fit in a MsgLeiosBlockTxsRequest

> It's generally desirable to receive missing transactions in order. This would give us the option to apply them as they arrive in a streamed fashion.

(edit: I hoisted this thread to https://github.com/IntersectMBO/ouroboros-consensus/pull/2302/changes#r4047067896)

> I think it should be [write to TxCache before write to LeiosDB] for the sake of reactivity

It's DB before Cache precisely to eliminate the cache staleness/racy-ness risk, which we currently don't have to bother with catching and recovering from. I'm hoping the code simplicity outweighs the performance improvement (eg once denormalization accelerates the LeiosDB insert)

> Don't think it's a big issue, but seeding the cache from the LeiosDb [on node init] should be doable, right?

I think so yeah; another "simpler should be good enough for MVP" case, so far

> [This doc could] be dropped into the leios-design document more or less verbatim

[ ] :plus1: I'll add that to the PR as a follow-up task once it merges

docs/website/contents/references/leiosfetch_spec.md
```diff
@@ -0,0 +1,203 @@
...
+- Definition of __LeiosFetchFromPeer__: whenever BoundedInflightBytes allows for more requests, the node will request some of the outstanding jobs from a peer who offered them (and hasn't already been sent requests for those exact jobs).
+  The node will choose to request jobs that are already inflight with the fewest peers (aka "RarestFirst").
+  (Note that any disconnection must involve some bookkeeping so that the rarity counts remain accurate.)
+  The order of the jobs with the minimum-rarity is chosen randomly.
+  Jobs are picked one at a time in that way until the peer's BoundedInflightBytes bound is met.
+  Recall that a BigLedgerPeer's inflight allowance is much greater than a single EB's closure, so an idle BigLedgerPeer will request _all_ jobs for an EB, regardless of their current inflight rarity (TODO Leverage that to optimize the decision logic somehow? Seems unlikely to matter) (TODO but maybe its effect on the order in which the jobs are requested from the BigLedgerPeer matters?).
```
nfrisby, line 67: @ch1bo wrote elsewhere (b/c I accidentally shared document link instead of the PR) that:

> > TODO but maybe its effect on the order in which the jobs are requested from the BigLedgerPeer matters?
> 
> It's generally desirable to receive missing transactions in order. This would give us the option to apply them as they arrive in a streamed fashion.

Yeah, but the tension is that if you request them out of order then you'll have received the entire EB sooner by leverage multiple (honest) peer's bandwidth more efficiently.

I just now pushed up https://github.com/IntersectMBO/ouroboros-consensus/commit/0691af0b32dbd04df528b7df45842e636f3a6455 which adds a new TODO paragraph a several lines down that elaborates on the tension.

PR #2305, opened 2026-09-18
Correct the Haddock on the two transaction measure phases

dnadales:
# Description

The argument comment on `txMeasurePhase1` says the state needs values.
The state is `EmptyMK`, so the comment is false.
It was copied from `txMeasurePhase2`, where it is correct.

It now says why phase 1 has no values.
The mempool measures phase 1 before it reads the ledger tables (`Mempool/Update.hs:206`), so a rejected transaction costs no table read.

This also gives `TxMeasurePhase2` the Haddock it lacked.

The invariant above `txMeasurePhase1` still names `txMeasure`, which the two-phase split removed.
Restating it across the phases needs a judgement, so this leaves a `TODO`.

Comment only, so no changelog fragment. Please add the "no changelog" label.


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -418,13 +421,16 @@ class
   -- 'blockCapacityTxMeasure cfg st'. Otherwise, the mempool could block
   -- forever.
   --
+  -- TODO: the bound against 'blockCapacityTxMeasure' names 'txMeasure', which
+  -- no longer exists. Restate it over 'txMeasurePhase1' and 'txMeasurePhase2'.
+  --
```
jasagredo: Could do also this one on this same PR, but it is fine otherwise.

jasagredo (approved):

PR #2308, opened 2026-09-21
Lock-free Mempool reads

bladyjoker:
DONE
 - [x] Use SVars instead of TMVars to hold Mempool state


ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictSVar.hs
ch1bo, line 93: What is the usual `SVar` API and it's behavior? Is there some original content I could read about it?

ch1bo (approved):
No idea what an SVar does, but the tests will check whether it's a fine replacement.

ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/MonadSTM/StrictSVar.hs
bladyjoker: I searched around, there's no prior art to this that is similarly named, the best I could find are similar practices in other ecosystems as "lock-free reads" etc.

https://www.infoq.com/articles/read-copy-update/

bladyjoker: I mean modern transactional databases do allow for lock-free reads, spanner, postgres I think even SQLite does this if your transaction never actually writes anything.

PR #2309, opened 2026-09-22
fix: EB hash collision handling in Leios

pgrange:
# Description

In the rare case were two different EB are forged with the same content, they'll have the same hash but should still be handled separately as they'll require a different signature process as they're applied on different tips.

When a second EB appears with the same hash as a previously known one, this PR ensures that:
* the new association EB/point is stored in DB
* the new association EB/point is announced
* the corresponding notifications are raised
* the corresponding traces are emitted

This issue has been discovered in the context of [leios issue 1077](https://github.com/input-output-hk/ouroboros-leios/issues/1077) which makes the probability of two similar EB appear higher and broke a property test.


pgrange:
@dnadales flagging you just in case it may be related to https://github.com/input-output-hk/ouroboros-leios/issues/1107

